PR #85, opened 2019-09-27
Tidy up exceptions to be more accurate.

ikelos:
Hiya, here's my attempt at going through all the exceptions we throw.  Note, this is just fixing up the exceptions and explicitly does not change any that we catch, so if you see one that you expect to get caught but doesn't, please shout.

Also if there's any you disagree with, or don't think are worthwhile) or any I've missed, please shout too...  5:)

I'm a little nervous about making a change of this size with potentially subtle and difficult to spot issues, so this is likely to go into 3.0.0_beta2, but still worth reviewing/checking if you can.  5:)

### DIFF
--- a/volatility/framework/layers/vmware.py
+++ b/volatility/framework/layers/vmware.py
@@ -5,10 +5,14 @@
 import struct
 from typing import Any, Dict, List, Optional
 
-from volatility.framework import interfaces, constants
+from volatility.framework import interfaces, constants, exceptions
 from volatility.framework.configuration import requirements
 from volatility.framework.layers import physical, segmented, resources
 from volatility.framework.symbols import native
+
+
+class VmwareFormatException(exceptions.LayerException):
+    """Thrown when an error occurs with the underlying Crash file format."""
 
 
 class VmwareLayer(segmented.SegmentedLayer):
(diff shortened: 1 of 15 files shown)


atcuno:
Looks good to me. I like the very specific errors

volatility/framework/layers/vmware.py
```diff
@@ -5,12 +5,16 @@
...
 
-from volatility.framework import interfaces, constants
+from volatility.framework import interfaces, constants, exceptions
 from volatility.framework.configuration import requirements
 from volatility.framework.layers import physical, segmented, resources
 from volatility.framework.symbols import native
 
 
+class VmwareFormatException(exceptions.LayerException):
+    """Thrown when an error occurs with the underlying Crash file format."""
```
iMHLv2: This says Crash file format, but should say VMware file format.

iMHLv2:
All looks good, with the exception (hehehe!!) of the Crash vs VMware comment I left above. I think the more specific exceptions make sense. I didn't do a thorough search for statements specifically catching the old exceptions, but I imagine they'll appear quickly and can be fixed easily.

ikelos:
Cool, good spot, thanks!  I've fixed up the typo.  If you're happy to have it merged and we just let exceptions get caught and patched fairly sharply, then I'll rebase all this...  5:)

volatility/framework/layers/vmware.py
ikelos: Good spot, thanks, I've fixed this now.  5:)

PR #89, opened 2019-10-17
Update README from txt to markdown so the preview looks better on github

Yohannfra:


ikelos:
Thanks for your contribution!  It looks good and should be easy to merge, but as the first external pull request we've had you get the pleasure of us figuring out our review process (so thank you for starting us off with such an easy one!).  5:)

One suggestion I'd make would be to include the removal of the original readme file, so the change is self contained.  Don't worry too much, we can take it as is, but it's easier for the small number of reviewers if requests are complete standalone units.

We'll try to get the process sorted sometime today, and hopefully get this all merged shortly after... 5:)

Yohannfra:
I just add the deletion of the old README file, hope your prefer it that way. 
Thanks for your quick review :)

thebeanogamer:
Would wrapping the commands in backticks look better? ` ``` `

To use line 51 as an example, would it look better like this, rather than just plain text?
```$ python vol.py —f /home/user/samples/stuxnet.vmem windows.info```

thebeanogamer:
Also (and this is obviously just my opinion, more than happy to be told I'm wrong), would it render better without the manual addition of `<br>`s? It is deviating from the idea of matching readme.txt as closely as possible, but equally, it could help with rendering either on devices smaller than the lines or substantially bigger, as GitHub will automatically insert line breaks to make best use of the space.

Example:
image

thebeanogamer:
@Yohannfra Could you please give me push to the fork this is on? There's a few minor things that I think might be worth changing, but it's probably easier for me to just do them then to try and explain them. Of course, if you'd rather, I can just describe stuff and you can decide if it's worth changing. Still just my opinion at the end of the day.

PR #100, opened 2019-10-23
Check progress_callback is not None before call

xabiugarte:
Just like in https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/pdbconv.py#L921


ikelos:
Good spot, thanks!  5:)  This should be fairly quick to get merged.  I'll try to get to it later today...  5:)

PR #105, opened 2019-10-23
layers: Improve scanning of layered chunks

ikelos:
Previously, for segmented layers, no chunk larger than a full segment
was scanned.  Now, we combine contiguous segments up to
scanner.chunk_size (or until a non-present segment) in order to scan it.

This may be slightly slower (by concatting the data together) but gives
results of hits across page boundaries (the point of the scanning
capability), whereas previously each page was scanned individually.

### DIFF
--- a/volatility/framework/interfaces/layers.py
+++ b/volatility/framework/interfaces/layers.py
@@ -458,22 +458,33 @@ class TranslationLayerInterface(DataLayerInterface, metaclass = ABCMeta):
         """Essentially, for paged systems we take a bunch of pages and chunk them up into scanner.page_size or
         as large a chunk as possible (if there are gaps)."""
         for (section_start, section_length) in sections:
+            # For each section, split it into scan size chunks
             for chunk_start in range(section_start, section_start + section_length, scanner.chunk_size):
+                # Shorten it, if we're at the end of the section
                 chunk_length = min(section_start + section_length - chunk_start, scanner.chunk_size + scanner.overlap)
 
+                # Prev offset keeps track of the end of the previous subchunk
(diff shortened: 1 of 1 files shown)


ikelos:
Hiya, this is a reasonably important fix.  At the moment things like the intel layer get scanned a page at a time (4096 bytes each block passed to the scanner).  This means we miss anything that lives on the boundaries.  What we expected was for it to read as much data as possible (16Mb + 1 page) and then scan again started 16Mb later.

This should now implement that, up until the next invalid page (in which case we won't be missing a boundary), so we simply return whatever we've got when we hit an invalid page.

I've confirmed this does find more results than without (whilst dealing with issue #100).  If one (or more) of you could check through this make sure it makes sense and looks good, then I can get it merged.  I haven't figured out the exact speed impact, I suspect there will be a slight slow down due to building the chunk of data in memory, but I don't know how significant it is (and it certainly isn't to the point of being unusable).  This will affect most *every* layer that does scanning at the moment, other than flat files.  It will likely impact poolscan most, since that does the biggest scanning we do.  It's necessary though, since at the moment we miss hits between page boundaries (ie, our old method was not really better than grep)...

ikelos:
Hmmm, probably best to hold off on this, an accidental commit led to buildbot building it all and apparently it broke pretty much everything, so I need to do a lot of investigation into it.  5:S  Sorry for the false start!

ikelos:
Ok, sorry for the spam, think I've resolved the issue so the review is back on.  5;)

atcuno:
This looks good. I tested the code separately to make sure all the overlap was working as advertised.

Issue #101, opened 2019-10-23
yarascan.YaraScan doesn't work


doomedraven:
memdump and yara will be shared in private

works just fine on vol2
```
vol.py -f X.dmp --profile=Win7SP1x86 yarascan -y X.yar
Volatility Foundation Volatility Framework 2.6.1
Rule: X
<removed>
```

```
python3 vol.py -vvvvvvv -f memdump.dmp yarascan.YaraScan --yara-file test.yar
Volatility 3 Framework 1.0.0-beta.1
INFO     root        : Volatility plugins path: ['/home/X/volatility3/volatility/plugins', '/home/X/volatility3/volatility/framework/plugins']
INFO     root        : Volatility symbols path: ['/home/X/volatility3/volatility/symbols', '/home/X/volatility3/volatility/framework/symbols']
Level 6  volatility.framework: Importing from the following paths: /home/X/volatility3/volatility/plugins, /home/X/volatility3/volatility/framework/plugins
DEBUG    volatility.framework: Importing module: volatility.plugins.windows.statistics
DEBUG    volatility.framework: Importing module: volatility.plugins.windows.registry.certificates
DEBUG    volatility.framework: Importing module: volatility.plugins.yarascan
DEBUG    volatility.framework: Importing module: volatility.plugins.layerwriter
DEBUG    volatility.framework: Importing module: volatility.plugins.timeliner
DEBUG    volatility.framework: Importing module: volatility.plugins.configwriter
DEBUG    volatility.framework: Importing module: volatility.plugins.mac.pstree
DEBUG    volatility.framework: Importing module: volatility.plugins.mac.lsmod
DEBUG    volatility.framework: Importing module: volatility.plugins.mac.lsof
DEBUG    volatility.framework: Importing module: volatility.plugins.mac.pslist
DEBUG    volatility.framework: Importing module: volatility.plugins.mac.bash
DEBUG    volatility.framework: Importing module: volatility.plugins.mac.psaux
DEBUG    volatility.framework: Importing module: volatility.plugins.mac.check_sysctl
DEBUG    volatility.framework: Importing module: volatility.plugins.mac.proc_maps
DEBUG    volatility.framework: Importing module: volatility.plugins.mac.netstat
DEBUG    volatility.framework: Importing module: volatility.plugins.mac.ifconfig
DEBUG    volatility.framework: Importing module: volatility.plugins.mac.malfind
[...]

PR #106, opened 2019-10-23
vadyarascan: Ensure we scan the right layer

ikelos:
This patch ensures we scan the vad segments within the correct layer
(the process layer, not the kernel layer) and lists the pid in the
output.

Just for confirmation, asking @iMHLv2 for a quick review.  I'll commit this on the 31st October if I don't hear any objections...  5;)


ikelos:
As a note, this should (partially) fix issue #101.  The scanning engine would also need improvement, but that's a bigger task, and lives in pull request #105.

doomedraven:
thanks verified that it works, i have applied both 106 and 105 prs

PR #110, opened 2019-10-25
Fix on eprocess is_valid

xabiugarte:
It seems that Windows symbols sometimes represent the DirectoryTableBase in KPROCESS as an array, others as an unsigned long, others as an unsigned long long. This fix deals with these discrepancies in the is_valid method for the EPROCESS class, just like it is done in the add_process_layer method.

### DIFF
--- a/volatility/framework/symbols/windows/extensions/__init__.py
+++ b/volatility/framework/symbols/windows/extensions/__init__.py
@@ -696,7 +696,7 @@ class EPROCESS(generic.GenericIntelProcess, ExecutiveObject):
                 return False
 
             # check for all 0s besides the PCID entries
-            if self.Pcb.DirectoryTableBase & ~0xfff == 0:
+            if self.Pcb.DirectoryTableBase[0] & ~0xfff == 0:
                 return False
 
             ## TODO: we can also add the thread Flink and Blink tests if necessary
--- a/volatility/framework/symbols/windows/extensions/__init__.py
+++ b/volatility/framework/symbols/windows/extensions/__init__.py
@@ -694,7 +694,7 @@ class EPROCESS(generic.GenericIntelProcess, ExecutiveObject):
 
             # check for all 0s besides the PCID entries
             if isinstance(self.Pcb.DirectoryTableBase, objects.Array):
-                dtb = self.Pcb.DirectoryTableBase.cast("unsigned long long")
+                dtb = self.Pcb.DirectoryTableBase.cast("pointer")
             else:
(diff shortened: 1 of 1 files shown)


volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -692,11 +692,17 @@ def is_valid(self) -> bool:
             if self.UniqueProcessId % 4 != 0:
                 return False
 
-            if self.Pcb.DirectoryTableBase == 0:
+            # check for all 0s besides the PCID entries
+            if isinstance(self.Pcb.DirectoryTableBase, objects.Array):
+                dtb = self.Pcb.DirectoryTableBase.cast("unsigned long long")
```
ikelos: Looks good, I wonder whether just always casting it to the right value is worthwhile, but I think this way is better for two reasons:
a) technically `unsigned long long` may not be defined by the windows symbol table (strange, but can happen)
b) always casting something will have a (extremely small) performance impact, so probably better to the isinstance to make sure first.

I can't figure out if it's better to unify this and the other existing code, but if so where would it live?  5:S  Hmmm, probably fine for now, but if we need to do it a third time, we should definitely put it in a single place...  5:)

ikelos:
Hiya, sorry for the delay, we just had a bit more of a discussion about this particular patch and a couple points were raised:
* If the memory image is 32-bit, then this might end up reading 64-bits (depending on the type of `unsigned long long`) which could mess up table translations.
* If the pdb for that particular windows image doesn't (for some completely unknown, unlikely reason) define `unsigned long long` or defines it to be some other size we're not expecting, that could also lead to trouble.

So, we were hoping you'd be able to keep it so that it only casts arrays (we considered doing it all the time and figured that's a bad idea), but also change the casting to a `pointer` type, which should always be masked to the right size for an address (which the DTB should always be) regardless of whether it's 64-bit or 32-bit.

We're keen for contributors to make the changes so that they learn how/why we do things, to keep us from getting loaded down and so that their work gets the commit message (rather than looking like one of us made the fix) however we're happy if you'd like us to do it too, just let us know.  If we don't hear back from you before December 3rd, we'll probably merge your work and then commit a change to use pointers, but we'd like you to get the credit.  5:)

xabiugarte:
Sorry for the long wait, this last commit implements what you mentioned on the last comment.

Thanks for the patience!

ikelos:
Awesome, thanks for fixing this up!  5:)

PR #112, opened 2019-10-25
Add init_order_modules and mem_order_modules to EPROCESS class

xabiugarte:
This PR adds functions init_order_modules and mem_order_modules, equivalent to load_order_modules for the other 2 linked lists present in the PEB. I don't know if these where intentionally excluded, but since the equivalents were present in volatility2 and I was using them in PyREBox, I just added them to the class.

### DIFF
--- a/volatility/framework/symbols/windows/extensions/__init__.py
+++ b/volatility/framework/symbols/windows/extensions/__init__.py
@@ -706,6 +706,7 @@ class EPROCESS(generic.GenericIntelProcess, ExecutiveObject):
 
         return True
 
+
     def add_process_layer(self, config_prefix: str = None, preferred_name: str = None):
         """Constructs a new layer based on the process's DirectoryTableBase."""
 
@@ -729,9 +730,8 @@ class EPROCESS(generic.GenericIntelProcess, ExecutiveObject):
         # Add the constructed layer and return the name
         return self._add_process_layer(self._context, dtb, config_prefix, preferred_name)
 
-    def load_order_modules(self) -> Iterable[int]:
-        """Generator for DLLs in the order that they were loaded."""
-
+    def get_peb(self) -> interfaces.objects.ObjectInterface:
+        """Constructs a PEB object"""
         if constants.BANG not in self.vol.type_name:
             raise ValueError("Invalid symbol table name syntax (no {} found)".format(constants.BANG))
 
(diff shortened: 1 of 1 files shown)


ikelos:
Thanks for the patches, I don't think they were intentionally left out, I'm just not sure we needed them yet, but I'll check with the rest of the team.  Unfortunately it may take us a while to review these, but we'll get to them as soon as we can...  5:)

ikelos:
Also, as a slight note, it feels like a lot of the work getting the Peb in the right layer could be merged into a single method and then shared by all three methods?  Would you be happy to mock up a `def get_peb(self) -> interfaces.objects.ObjectInterface` method and convert all three to use it please?

xabiugarte:
Hi @ikelos, you are totally right. I just created that get_peb() function. Thanks!

ikelos:
Awesome, thanks!  Looking good, just one last point from me which would be the docstrings. Would you be able to update those to describe the differences between the functions (or what each one does more specifically)?  After that I think I'm good to approve it, and if @iMHLv2 is happy with it then we can merge it!  5:)

iMHLv2:
@xabiugarte Thanks for the patches. For the pydocs, you could probably just say:

* `load_order_modules`: "Generator for DLLs in the order that they were loaded." (no changes needed, that's what it says now)

* `mem_order_modules`: "Generator for DLLs in the order they appear in memory."

* `init_order_modules`: "Generator for DLLs in the order they were initialized."

Also instead of indicating `Iterable[int]` I think the 3 methods should be using `Iterable[interfaces.objects.ObjectInterface]` since they're yielding instances of objects rather than integers. 

I know this wasn't in the original `load_order_modules` method, but based on a recent addition (see commit 3df5e995) calls to `self.add_process_layer()` should be done inside a try/except in case the DTB for a process isn't valid anymore (for example, calling one of these methods on a terminated process). 

As @ikelos mentioned, we didn't intentionally leave those methods out, we just hadn't gotten to it yet, so thanks for taking the initiative!

xabiugarte:
Hi, I have added the changes you proposed. 
Regarding the try/except for self.add_process_layer(), the function just returns (None) (https://github.com/volatilityfoundation/volatility3/pull/112/files#diff-3bb5ba1663a0509a8200255de9fa3d16R740), in the same way it returns in case the Peb is not a valid address for the layer. That said, I don't know if it should be raising/reraising some exception instead... What do you think?

ikelos:
Good spot!  You've actually highlighted a slight inconsistency we need to iron out, in that the operating systems differ (windows returns a string or raises, linux and mac catch some exceptions but sometimes return None), and we need to figure out how we want to deal with it, so best hold there while we figure it out, but thanks for making the other changes!  5:)

ikelos:
Just for information, this is probably dependent upon #123.

ikelos:
Ok, so the decision we've gone for is to raise exceptions (rather than potentially raise exceptions *and* return None), so ideally we'd like to ditch anything that might cause it to return `None`.  The idea is not to have to `try/except` but then also have to  check `if is not None`.

That means you can either catch the exceptions and return one that's more meaningful, or just let them bubble up.  I think probably bubbling up makes the most sense (and gives the most data about the cause of the problem), but happy for you to catch and change the exception if you think it makes sense...

Once we get that cleared up, I think it's good for merging.  5:)

xabiugarte:
Hi, I have updated the pull request following the latest comments. Let me know if that's okay according to what you decided.

Thanks!

ikelos:
Thanks very much, we'll try to get this reviewed by the end of the week (depending on how busy people are)...  5:)

Issue #97, opened 2019-10-19
Support ELF file format images


koromodako:
First of all, many thanks for this release, I have been waiting for it for a long time :)

I'll try to be as precise as possible:

- The version of Volatility you're using: `v1.0.0-beta.1-10-g27a291cf`
- The operating system used to run Volatility: `Ubuntu 19.04`
- The version of Python used to run Volatility: `python3/disco,now 3.7.3-1 amd64`
- The suspected operating system of the memory sample: 

`Windows 7 SP1 x64` which can be analyzed with `volatility2` profile called `Win7SP1x64`

- The complete command line you used to run Volatility: 

```
> $ volatility3 -s volatility3-symbols -f win7sp1x64.dmp windows.statistics.Statistics
Volatility 3 Framework 1.0.0-beta.1
Progress:    0.00		Scanning FileLayer using PageMapScanner
Unsatisfied requirement plugins.Statistics.primary: Memory layer for the kernel

A translation layer requirement was not fulfilled.  Please verify that:
	A file was provided to create this layer (by -f, --single-location or by config)
	The file exists and is readable
	The necessary symbols are present and identified by volatility
Unable to validate the plugin requirements: ['plugins.Statistics.primary']
```

When I run it with `-vvv` I observe a debug message saying:

```
DEBUG    volatility.framework.automagic.windows: Self-referential pointer not in well-known location, moving to recent windows heuristic
```

I took a quick look at the code and it seems to mean that `automagic` does not
match anything when scanning for DTB because I won't see a debug message saying:

```
DEBUG    volatility.framework.automagic.windows: DTB was found at: 0x......
```

Have you tested `volatility3` on a Windows 7 dump ? Do you need me to perform more tests ?

ikelos:
Hiya, thanks for doing the introductory diagnostics for us.  If it isn't finding the DTB, that's a bit of an issue.  Does vol 2 tell you where it is?  If so, you can write it into a configuration file to force vol 3 to know where it is, but I'd like to get a little more information if possible.  Could you please attach a copy of the `-vvv` output so we can look through it please?

In answer to your other questions, yes, we've tested it on many different images ranging from XP through to Windows 10 (and several server versions on the way).  Windows 7 is included in that, but as you pointed out, if it can't identify the DTB then this is before it even tries to do the windows part, it's still on the Intel part.

koromodako:
@ikelos, here is `volatility2` output:
```
[...]

PR #113, opened 2019-10-27
Elf64 support

ikelos:
Ok, this seems to have stabilized and seems to be working ok.  It should resolve issue #97.  I'd appreciate some eyes on it, particularly those that worked with the previous incarnation.  Let's go with another 7 days, so if I don't hear anything then I'll merge this on the 3rd November.  5:)


doomedraven:
amazing <3

PR #115, opened 2019-10-27
simplify vad dump to be used by other plugins

doomedraven:


volatility/framework/plugins/windows/vaddump.py
```diff
@@ -36,14 +36,33 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                 requirements.PluginRequirement(name = 'vadinfo', plugin = vadinfo.VadInfo, version = (1, 0, 0)),
                 ]
 
+    @staticmethod
```
ikelos: One convention is that we tend to use `classmethod` rather than `staticmethod`, but I can't think of anything from the class that such an extensible method (not this one in particular, but any in the future) might want to use?  I have the nagging feeling I did try `staticmethod` once but stuck with `classmethod`, but I honestly don't recall why, so I won't push too hard on this one if you think `staticmethod` makes more sense...  5:)

volatility/framework/plugins/windows/vaddump.py
```diff
@@ -36,14 +36,33 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+    def vad_dump(vad, proc_layer):
```
ikelos: Another convention is that we usually pass `context` and `layer_name` rather than passing in a full layer.  I think this is because if the method ever needs to be extended in the future, having the context means you've got everything, whereas having just the layer makes it slightly more troublesome to get other things you might need (such as symbol tables, etc).  I think since this is core code, and we need to be careful of interfaces changing in the future, we should try to be consistent, so this is a change I think has to happen.

I'd expect the final signature to be something along the lines of:
```
def vad_dump(cls, context: interfaces.context.ContextInterface, layer_name: str, vad: interfaces.objects.ObjectInterface) -> bytes:
```

doomedraven:
im totally fine with both requested changes, code updated with them

doomedraven:
great, thanks for checking it

PR #116, opened 2019-10-28
Framework info plugin

ikelos:
Add in a little plugin to show the descendants of the core bits of volatility.  It might make diagnostics a little easier (and equates to volatility --info).  It should not require a `-f` parameter` so can be run simply as:
```
python vol.py frameworkinfo
```


ikelos:
Hmmm, can't tell if it should be a core plugin, or another example for people to write their own under `volatility.plugins`?  Any thoughts?  I'm starting to lean towards it being a core plugin?

ikelos:
Ok, I'm happier this is a core plugin, so I'll merge it now...

PR #121, opened 2019-11-02
pdbconv: Avoid more name-collisions

ikelos:
Previously we allow name-collisions for types with name <anonymouos-tag>
so this has now been resolved, as well as descending too deeply down the
type to tree determine the size of array occasionally.

Another reasonably deep fix, but quite straight forward.  It will eventually require regeneration of all the windows symbol tables, but using the previous ones should have minimal impact (they're usually deeply down a type tree, and for a type that wasn't given a name).

No timeline set on this one (although, probably a month mebbe?) because things this foundational make me want them checked more than I want them committed...  5:)

### DIFF
--- a/volatility/framework/symbols/windows/pdbconv.py
+++ b/volatility/framework/symbols/windows/pdbconv.py
@@ -366,8 +366,9 @@ class PdbReader:
             offset += length_len
             output, consumed = self.consume_type(module, offset, length)
             leaf_type, name, value = output
-            if name == '<unnamed-tag>' or name == '__unnamed':
-                name = '__unnamed_' + hex(len(self.types) + 0x1000)[2:]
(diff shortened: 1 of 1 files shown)


ikelos:
I have a strong aversion to `while True`; from a program design perspective it's just an easy way to slip up and form an infinite loop.  The example is the same as `while not <condition>:` and I far prefer that over the example.  If there's another way that looks neater without risking infinite loops, I'm up for it.

I agree the loop variable is a little ugly (and requires initialization), but the following wouldn't work:

`while not element_type > 0x1000:`

because later on we do bomb out (directly after we've set element_type to be greater than 0x1000), so that's why I went with the loop variable.  Happy to rename it, or restructure the code if there's a way that would look better?

PR #123, opened 2019-11-03
Is valid exception catching

ikelos:
This makes the layer `is_valid` method a little more robust.

Coverts all calls to is_valid to boolean checks (any exceptions thrown now should likely still be thrown and/or should be caught by the is_valid method).

Adds some lengths in where the liable length is known.

Removes checks that are already wrapped in a try/except call (there's no point doing the check if we're still going to catch an exception for something else).

This feels like it could potentially trip something up, so not putting a merge deadline on this.

### DIFF
--- a/volatility/framework/layers/registry.py
+++ b/volatility/framework/layers/registry.py
@@ -227,13 +227,14 @@ class RegistryHive(linear.LinearlyMappedLayer):
 
     def is_valid(self, offset: int, length: int = 1) -> bool:
         """Returns a boolean based on whether the offset is valid or not."""
-        # TODO: Fix me
-
-        # Pass this to the lower layers for now
-        return all([
-            self.context.layers[layer].is_valid(offset, length)
-            for (_, offset, length, layer) in self.mapping(offset, length)
-        ])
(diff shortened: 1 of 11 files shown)


volatility/framework/layers/registry.py
```diff
@@ -227,8 +227,14 @@ def dependencies(self) -> List[str]:
...
         """Returns a boolean based on whether the offset is valid or not."""
-        # TODO: Fix me
-        return True
+        try:
+            # Pass this to the lower layers for now
+            return all([
+                self.context.layers[layer].is_valid(offset, length)
+                for (_, offset, length, layer) in self.mapping(offset, length)
+            ])
+        except exceptions.InvalidAddressException:
```
atcuno: Is there a reason this one catches an exception still?

ikelos: Yep, the mapping call itself may still throw an `InvalidAddressException`.  We've also got a case where `is_valid` can sometimes call `mapping` so we don't often let `mapping` use `is_valid`, so I think this catch still needs to be there.

Issue #109, opened 2019-10-24
Standardize debug log messages for smear and ensure used in existing code places


atcuno:
To fufill this issue requires two tasks:

- [ ] Create a standardized debug log message format for issues where smear is encountered. Ideally this message would include the address (offset) and attempted operation that encountered the smear

- [ ] Go through the currently written plugins and extensions to use this template and ensure all needed places have the debug messages emitted

ikelos:
Actually, the plugins should still be raising exceptions that would cause them to stop, the UI should catch the exception message and raise it more diplomatically.  This is important in case volshell and/or another application wants to do something different based on the exceptions.

Going through the plugins/extensions is only necessary where they *can* continue on an error but currently don't (which should be rare), so it's really the first issue we'll be concentrating on...

ikelos:
Ok, there's a mock-up of improved exception responses, in the branch called issue109-standardized-exceptions.  If it looks ok, I'll make it a PR, we can review it and go from there.  If there's issues with the way forward then we can just ditch the branch.  5:)

ikelos:
Ok, the resolution for this was merged, so I'm gonna close it off.  5:)

PR #124, opened 2019-11-03
Issue109 standardized exceptions

ikelos:
Ensure that if an exception gets thrown and not caught, that it's displayed to the user in the CLI with appropriate suggestions.

I'm reasonably happy with how this is laid out, and the other improvements to exceptions.  @atcuno I know you were interested in ensuring suitable warnings, please take a look at this.  I'll merge it by the 14th (so, after our next meeting) if I haven't had any feedback...  5:)


ikelos:
Should fix #109.

atcuno:
I like it! Having the easy if/else switches based on exception type and the caused_by allowing arbitrary messages should make it easy to maintain and add on more if ever needed

PR #90, opened 2019-10-18
Add Missing Word to Readme

thebeanogamer:
Add a missing `vol.py` from the command `python -f <imagepath> windows.info` in README.txt. If the README is being re-done by #89, this would need to be changed there as well.

PR #126, opened 2019-11-04
Markdown all the things

thebeanogamer:
If it can be markdown, it will be markdown! This _should_ all be compliant.

Also includes #90.

On @ikelos' advice, I've left the email address alone.

Happy to change whatever.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ The latest generated copy of the documentation can be found at: <https://volatil
 
 ## Licensing and Copyright
 
-Copyright © 2007-2019 Volatility Foundation
+Copyright (C) 2007-2019 Volatility Foundation
 
 All Rights Reserved
 
--- a/README.md
+++ b/README.md
@@ -7,106 +7,108 @@ visibility into the runtime state of the system. The framework is intended
 to introduce people to the techniques and complexities associated with
 extracting digital artifacts from volatile memory samples and provide a
 platform for further work into this exciting area of research.
-<br>
+
 In 2019, the Volatility Foundation released a complete rewrite of the
 framework, Volatility 3. The project was intended to address many of the
 technical and performance challenges associated with the original
 code base that became apparent over the previous 10 years. Another benefit
 of the rewrite is that Volatility 3 could be released under a custom
 license that was more aligned with the goals of the Volatility community,
-the Volatility Software License (VSL). See the LICENSE file for more details.
+the Volatility Software License (VSL). See the [LICENSE](LICENSE.txt) file for more details.
 
 ## Requirements
 
-- Python 3.5.3 or later. http://www.python.org
-- Pefile 2017.8.1 or later. https://pypi.org/project/pefile/
+- Python 3.5.3 or later. <https://www.python.org>
+- Pefile 2017.8.1 or later. <https://pypi.org/project/pefile/>
 
 ## Optional Dependencies
 
-- yara-python 3.8.0 or later. https://github.com/VirusTotal/yara-python
-- capstone 3.0.0 or later. https://www.capstone-engine.org/download.html
+- yara-python 3.8.0 or later. <https://github.com/VirusTotal/yara-python>
+- capstone 3.0.0 or later. <https://www.capstone-engine.org/download.html>
 
 ## Downloading Volatility
 
-The latest stable version of Volatility will always be the master <br>
-branch of the GitHub repository. You can get the latest version of <br>
-the code using the following command: <br>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,116 +7,119 @@ visibility into the runtime state of the system. The framework is intended
...
     ```shell
-    $ python vol.py —f /home/user/samples/stuxnet.vmem windows.info
+    python3 vol.py —f /home/user/samples/stuxnet.vmem windows.info
     ```
-4. Run some other plugins. The -f or —-single-location is not strictly
+
+4. Run some other plugins. The `-f` or `—-single-location` is not strictly
 required, but most plugins expect a single sample. Some also
-require/accept other options.  Run "python vol.py <plugin> -h"
+require/accept other options.  Run `python3 vol.py <plugin> -h`
```
ikelos: This is somewhat limiting, not all OSes use `python3` and as python2 is about to end-of-life, I suspect that a number of OSes may do a change and/or symlink to run python3 when people type python, so I'd prefer to leave this as is...

README.md
```diff
@@ -7,116 +7,119 @@ visibility into the runtime state of the system. The framework is intended
...
 3. To get more information on a Windows memory sample and to make sure
 Volatility supports that sample type, run
-'python -f <imagepath> windows.info’
+`python3 vol.py -f <imagepath> windows.info`
 
    Example:
+
     ```shell
-    $ python vol.py —f /home/user/samples/stuxnet.vmem windows.info
+    python3 vol.py —f /home/user/samples/stuxnet.vmem windows.info
```
ikelos: See below

README.md
```diff
@@ -7,116 +7,119 @@ visibility into the runtime state of the system. The framework is intended
...
+2. See available options:
+
+    ```shell
+    python3 vol.py —h
+    ```
 
 3. To get more information on a Windows memory sample and to make sure
 Volatility supports that sample type, run
-'python -f <imagepath> windows.info’
+`python3 vol.py -f <imagepath> windows.info`
```
ikelos: See below

README.md
```diff
@@ -7,116 +7,119 @@ visibility into the runtime state of the system. The framework is intended
...
 
+    ```shell
     git clone https://github.com/volatilityfoundation/volatility3.git
+    ```
 
-2. To see available options, run "python vol.py -h"
+2. See available options:
+
+    ```shell
+    python3 vol.py —h
```
ikelos: See below (for some reason, I review upwards?)

README.md
```diff
@@ -7,116 +7,119 @@ visibility into the runtime state of the system. The framework is intended
...
+The framework is documented through doc strings and can be built using sphinx.
 
-The latest generated copy of the documentation can be found at: <br>
-https://volatility3.readthedocs.io/en/latest/ <br>
+The latest generated copy of the documentation can be found at: <https://volatility3.readthedocs.io/en/latest/>
 
 ## Licensing and Copyright
 
-Copyright (C) 2007-2019 Volatility Foundation
+Copyright © 2007-2019 Volatility Foundation
```
ikelos: In case people do read this in a text editor, I'd feel happier if this were `(C)` rather than `©`.

thebeanogamer: Yeah, that's fair enough. I'll swap it back.

README.md
thebeanogamer: So my logic was that `python` can be ambiguous, but I've always seen `python3` mean Python 3, however, that is me mostly hiding in a corner with a big Ubuntu sign, so if that's not the case elsewhere it might be better avoided.

I've not actually tried running under Python 2. Does it error graciously or does it scream exceptions?

README.md
thebeanogamer: Done

README.md
thebeanogamer: 🤷‍♂ Your opinion that matters, not mine.

ikelos: We could add a starting check, but as I say, python2 official end of life is less than 2 months away.  I know it won't all change overnight, but I suspect that the python symlink will move from defaulting to python2 to defaulting to python3.  You're right, python3 does usually mean explicitly run under python3, so I could be persuaded on this.  It's either that or adding a check into vol.py before importing the CLI, which seems a little daft, so I guess I'll let it stay...  5:P

thebeanogamer: _Shakes fist at the idea of questions with no clear answer_

thebeanogamer:
Hehe

thebeanogamer:
🙌

PR #132, opened 2019-11-09
Move poolscan typecheck

### LINKED ISSUE #131: Implement poolscan checks for object validity and verify all constraints are accurate
image

vol2'result is normal
vol3'result translate txt --> this picture
vol3'result print --> closed ploblem 2hours ago picture

trouble process is iexplore.exe and rundll32.exe
why this problem?

ikelos:
This shifts reorganizes the pool header code and moves the check for an object's header type matching the pool header's `object_type` from the pool header's `get_object` to the calling plugin's pool traversal code.

It should be noted that this changes the pool header code, and we'll need to bump something somewhere to tell plugins whether they can or can't use that code.  The main version is really intended for the UI to check, having a single plugin_api version would require all other plugins to bump their required versions before they'd be interoperable with this version.  Breaking down the components so that each has its own API number would be difficult to define what's part of which component.  That's a separate discussion that needs to happen though.

This will need some work on the constraints to add necessary Index constraints (if any) and to decide whether to skip the type testing or not (such as for processes).

Somewhat resolves #131.

### DIFF
--- a/volatility/framework/plugins/windows/poolscanner.py
+++ b/volatility/framework/plugins/windows/poolscanner.py
@@ -13,7 +13,6 @@ from volatility.framework.layers import scanners
 from volatility.framework.renderers import format_hints
 from volatility.framework.symbols import intermed
 from volatility.framework.symbols.windows import extensions
-from volatility.framework.symbols.windows.extensions import pool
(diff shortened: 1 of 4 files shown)


ikelos:
This is dependent on #110

PR #137, opened 2019-11-13
CLI: Don't output the title for structured output.

ikelos:
This ensures that renderers like CSV and the upcoming JSON renderer don't output the title line in stdout, but that all other plugins do output it as a normal part of the output.

### DIFF
--- a/volatility/cli/__init__.py
+++ b/volatility/cli/__init__.py
@@ -77,7 +77,6 @@ class CommandLine(interfaces.plugins.FileConsumerInterface):
     def run(self):
         """Executes the command line module, taking the system arguments,
         determining the plugin to run and then running it."""
-        sys.stdout.write("Volatility 3 Framework {}\n".format(constants.PACKAGE_VERSION))
 
         volatility.framework.require_interface_version(1, 0, 0)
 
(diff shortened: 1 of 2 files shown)


npetroni:
These changes look good.

It's somewhat orthogonal to this change, but we should probably not emit console control characters when the file descriptor is not a tty. During testing, I noticed that the string `Formatting...^M` gets written to `stdout` for the `pretty` renderer.  Many text tools such as `head` and `cat` will interpret those characters and emulate the behavior, but it's cleaner to omit them.

```
$ python vol.py -r pretty -f image.raw windows.pslist  > out.txt
$ hexdump -C out.txt | head -n 10
00000000  56 6f 6c 61 74 69 6c 69  74 79 20 33 20 46 72 61  |Volatility 3 Fra|
00000010  6d 65 77 6f 72 6b 20 31  2e 30 2e 30 2d 62 65 74  |mework 1.0.0-bet|
00000020  61 2e 31 0a 46 6f 72 6d  61 74 74 69 6e 67 2e 2e  |a.1.Formatting..|
00000030  2e 0d 20 20 7c 20 20 50  49 44 20 7c 20 50 50 49  |..  |  PID | PPI|
00000040  44 20 7c 20 20 20 49 6d  61 67 65 46 69 6c 65 4e  |D |   ImageFileN|
00000050  61 6d 65 20 7c 20 20 4f  66 66 73 65 74 28 56 29  |ame |  Offset(V)|
00000060  20 7c 20 54 68 72 65 61  64 73 20 7c 20 48 61 6e  | | Threads | Han|
00000070  64 6c 65 73 20 7c 20 53  65 73 73 69 6f 6e 49 64  |dles | SessionId|
00000080  20 7c 20 57 6f 77 36 34  20 7c 20 20 20 20 20 20  | | Wow64 |      |
00000090  20 20 20 20 20 20 20 20  20 20 20 20 43 72 65 61  |            Crea|
```

ikelos:
Hmmm, so we were actually write `\r` in order that the carriage was returned and the line was written over once actual data is returned.  I think changing that to sys.stderr makes sense too, since it's more diagnostic for the program and not part of the actual output...

PR #146, opened 2019-12-03
Add in JSON renderer

ikelos:
This renderer will output a formatted/pretty-printed JSON output file.  It contains a nested tree structure (in particular for tree output) and makes use of a special `__children` field to do so.  We don't currently have strict checking on column names not starting with `__` but it seems like a remote possibility so for now I'm not including any special casing for it.

I'll probably merge this on the 7th unless someone shouts...

### DIFF
--- a/volatility/cli/text_renderer.py
+++ b/volatility/cli/text_renderer.py
@@ -296,6 +296,7 @@ class JsonRenderer(CLIRenderer):
     }
 
     name = 'JSON'
+    structured_output = True
 
     def get_render_options(self) -> List[RenderOption]:
         pass
--- a/volatility/cli/text_renderer.py
+++ b/volatility/cli/text_renderer.py
@@ -3,13 +3,15 @@
 #
 
 import datetime
+import json
 import logging
 import random
 import string
 import sys
 from functools import wraps
-from typing import Any, List, Tuple, Dict
+from typing import Any, List, Tuple, Dict, Optional
 
+from volatility.framework.interfaces.renderers import RenderOption
 from volatility.framework.renderers import format_hints
 
 vollog = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


ikelos:
Here's an example output JSON file for a tree view which therefore has children:

[out.json.txt](https://github.com/volatilityfoundation/volatility3/files/3922892/out.json.txt)

My concern with not outputting `__children` is that it just shifts the problem for the sake of humans seeing an empty value.  Without the `__children` field, every processor of the JSON will then need to verify the field is there or handle errors, rather than just accepting an empty list (ie, no need to special case if we include it every time).

I've no experience of JSON lines, and I don't know what browsers support them, but I know all browsers support standard JSON.  The other things is, would we then output each row and include a depth (ie, how vol2 JSON renderer used to do it, which I think we got negative feedback about) or output a row that can then have children (which for tree will blow the file size up anyway)?  I'm not sure what advantages JSON lines offers over stock JSON?

I don't have a use case in mind, just a lot of people seem to want a machine readable output that's easy to process and JSON seems fairly ubiquitous...

I've added the structured_output change, good spot, thanks!  5:)

ikelos:
Ok, this is on hold until we get some actual consumers of JSON to tell us whether this format is what they want (or whether they want JSON lines).  So if you're reading this, and that sounds like you, please shout!  We want to hear what you've got to say!  5:D

ikelos:
What's item 2 (member names)?

npetroni:
> What's item 2 (member names)?

JSON consumers might care about the key names that are used for fields. For example, in `windows.pslist` one key will have `"Offset(V)"` as the name, which is a bit non-standard.

npetroni:
Is it intentional that times have spaces at the end?
```
"CreateTime": "2005-06-25 16:57:59.000000 "
```

ikelos:
Hmmm, for member names, they're really column names, and the renderer doesn't have anything else to go off?  We could do sanitization of characters, but there's probably not a great deal more we could do to them?

As for the create times, the JSON module doesn't serial datetimes, so we convert it to a string using a [format string](https://github.com/volatilityfoundation/volatility3/pull/146/files#diff-927abf79440adef63cc25b2255edb998R292) which due to the %Z at the end (which is empty if there's no timezone attached to the timestamp) creates the space.  I believe I was aiming for [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) format, but I may have missed.  5;)  Happy to go with whatever format, but may need someone else to make the decision...  5;)

ikelos:
I've also added a JSONL formatter and realized that the JSON we were outputting was designed for humans to read because we were pretty printing it.  Happy to ditch one if we don't end up using it, or one becomes king, but as you've said, we need guidance.  5;)

ikelos:
Ok, we now actually push it out in JSON ISO 8601 format, rather than something designed to be human readable.

I don't think sanitizing the JSON key names will be useful, because it's the only thing clients will have for display, meaning that either every consumer needs to know exactly what plugin the output is from or the only thing it will be able to display to the user is the sanitized column name (which may not longer be human relevant).  We could architect something more complex to pass metadata (such as plugin, version, version of vol, runtime, stats, etc) as well as the treegrid, but then that's likely to foul up JSON lines, which would send the entire tree as one of the lines.  Still open for suggestions, but I'm not sure we'll get suggestions from a PR rather than from code in the tree during beta period.  As such I'm going to suggest another week for comments, reset the clock if I get any, but otherwise merge this on the 15th for testing.

PR #148, opened 2019-12-04
Linux - update handling of kernels that do not have loadable module s…

atcuno:
…upport

### DIFF
--- a/volatility/framework/plugins/linux/lsmod.py
+++ b/volatility/framework/plugins/linux/lsmod.py
@@ -4,16 +4,19 @@
 """A module containing a collection of plugins that produce data typically
 found in Linux's /proc file system."""
 
+import logging
+
 from typing import List, Generator, Iterable
 
 from volatility.framework import contexts
-from volatility.framework import renderers, constants, interfaces
+from volatility.framework import exceptions, renderers, constants, interfaces
 from volatility.framework.automagic import linux
 from volatility.framework.configuration import requirements
 from volatility.framework.interfaces import plugins
 from volatility.framework.objects import utility
 from volatility.framework.renderers import format_hints
 
+vollog = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


volatility/framework/plugins/linux/lsmod.py
```diff
@@ -44,6 +47,12 @@ def list_modules(cls, context: interfaces.context.ContextInterface, layer_name:
 
         vmlinux = contexts.Module(context, vmlinux_symbols, layer_name, 0)
 
+        try:
+            vmlinux.get_type("module")
+        except exceptions.SymbolError:
+            vollog.debug("The required symbol 'module' is not present in symbol table. Please check that kernel modules are enabled for the system under analysis.") 
+            return
```
ikelos: So as discussed, I'd prefer not to catch the exception inside this method, but leave it to callers to deal with.

ikelos: If we catch around the for loop in the `_generator` method, then that means callers can distinguish between error states and empty lists.

volatility/framework/plugins/linux/lsmod.py
```diff
@@ -44,6 +47,9 @@ def list_modules(cls, context: interfaces.context.ContextInterface, layer_name:
 
         vmlinux = contexts.Module(context, vmlinux_symbols, layer_name, 0)
 
+        # this will cause an exception if module support is not compiled into the kernel
```
ikelos: If the line below is there specifically to trip the exception early, this comment should say it, otherwise it seems unnecessary (because the code will either trip or not during the course of its run?)

atcuno: good call. check again :)

ikelos:
Perfect, all merged, thanks!  5:)

PR #149, opened 2019-12-04
Svcscan minor tweak

ikelos:
Arg, so I forget to check the counterpoint, the other validation (greater than 16299) should probably also say that it *doesn't* have that member.  Also, the metadata check (for next time we do a big rebuild of the library) will have been incorrect.  Sorry to throw down another review, but I figure you'll have samples that can verify this more easily than I do...  5:S

This also removes the redundant fallback checks for `ObHeaderCookie`.


iMHLv2:
@ikelos This should be good now after a56394e0bb329e9f2cbc4da6c8cdb84d8c59c0ab

iMHLv2:
I'm closing this and re-submitting as https://github.com/volatilityfoundation/volatility3/pull/180. The newer one will be cleaned up and also avoid merge conflicts.

iMHLv2:
I'll leave `svcscan-minor-tweak` for now, but it should be safe to delete that branch any time.

PR #152, opened 2019-12-04
Objects: Simplify get_symbol_table to get_symbol_table_name.

ikelos:
The only place we were using this was to get the table to look up its name and then use it to construct a new object (based on the previous object's table and a new type name) at a different offset.  This would cause us to do an unnecessary table lookup each time, when what we really wanted (the table name) had already been determined.  This simplifies the API a little to just return the symbol's name.

Options for the future include:
* Changing the function to construct a new object based on the table of the previous
* Removing this as an unnecessary extravagence
* Adding additional convenience functions to further split the `.vol.type_name` down
* Breaking this up at object create to ensure `type_name` is always stored separately to `symbol_table_name`, but this would be reasonably disturbing to the codebase (a lot of changes everywhere, which would need to be made carefully)

For now, this seems like a quick solution to the issue.

### DIFF
--- a/volatility/framework/interfaces/objects.py
+++ b/volatility/framework/interfaces/objects.py
@@ -128,17 +128,19 @@ class ObjectInterface(metaclass = ABCMeta):
         """Writes the new value into the format at the offset the object
         currently resides at."""
 
-    def get_symbol_table(self) -> 'interfaces.symbols.SymbolTableInterface':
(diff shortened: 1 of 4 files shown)


ikelos:
Also note, this is technically an API breaking change, but given we're still in the beta period I've chosen to not bump the API version.

PR #153, opened 2019-12-08
Quickscanning

ikelos:
Ok, so this mostly just pulls segments from the virtmap plugin, but it's given reasonable speed ups (around 50%).

One this this does change is not to exclude the "Unused" section, since I've found kernels that had valuable data in areas that were marked that way.  I don't know whether our marking was wrong, or whether the kernel was just doing weird things, but using `--quick` will definitely restrict findings that come from outside the allocated memory ranges.

@iMHLv2 if you can shed any light on what's going on with the Unused stuff, and/or why our virtmap code doesn't work for certain windows versions (from around 2016) that would be much appreciated!  5:)


iMHLv2:
By the way, I started looking at this before the call. The older Windows 10 PDBs have `_MI_VISIBLE_STATE` but without the required `SystemVaRegions` member. There do not appear to be other members there in its place with a similar function. Furthermore, the older PDBs do not have `_MI_SYSTEM_VA_ASSIGNMENT` at all, which is the type for `SystemVaRegions`. Next step would be to see if Windbg's `!poolfind` and `!vm` plugins work against memory dumps from those older systems (and if so, how they enumerate the regions).

ikelos:
Cool, can I leave that with you to do then and let me know back here please?

PR #162, opened 2019-12-26
VadDump: Fix incorrect variable out_of_range in vad_dump

Lyx09:
This PR fixes an incorrect variable initialization in the VadDump module.

The out_of_range variable should be `vad.get_end()` instead of `vad_get_start() + vad.get_end()`. The typo must've come up when porting this module from volatility2 to volatility3 which originally was `out_of_range = vad.Start + vad.Length `.
This is causing the output files to be a lot heavier and containing extra unwanted data.
Therefore I just fixed this variable.


ikelos:
Thanks, this looks good but I'll pass it by our VAD expert to double check...  5;)

iMHLv2:
Looks good, thanks for the fix.

PR #164, opened 2020-01-02
Adding pypykatz plugin to parse LSASS process

skelsec:
The plugin serves as an interface between pypykatz and volatility3.
The pypykatz (just like mimikatz) extracts stored secrets in the LSASS process.


ikelos:
Hmmm, so there's really not a lot of code here, it just calls out to separate code that's stored in the pypykatz package?  Given that and the fact that it's MIT licensed, I doubt we'd be able to include it in the core and it doesn't seem like it makes much sense?

I'd suggest packaging it as a standalone plugin (that actually has the guts of the code within the plugin, ideally as a reusable `classmethod` that accepts a context and whatever other parameters it may need.  That would allow other plugins to cross-rely on it, and would leave you completely in control of the plugin versioning and everything.

Also, so that other people can require the plugin in their own plugins and not get into versioning issues if the plugin's API changes, I'd define a `_version` variable that's a 3-tuple following the semver semantics (you can see an example of this in `pslist` and you've actually used the dependency requirement in this plugin code itself).  You can provide instructions for either where to put the plugin, or for using the `-p` plugin path parameter, but it really doesn't make much sense to add such a small shim into the package.

If you've got strong reason for wanting this to be in the core codebase then let us know, but otherwise I'm likely to close this off in a couple of days I'm afraid...

doomedraven:
oh this is pretty nice, thanks

skelsec:
As discussed on slack, I'm closing this PR and wait until an official decision is made where to put 3rd party plugins.

PR #165, opened 2020-01-03
Mac quick pslist fix

atcuno:
@ikelos 

I am fine with the patch the way it is since task structures won't cross page boundaries, but that won't always be the case.

Could you please respond before merging with the proper (if any) way to get the size of an object like I thought .size used to do.


atcuno:
This had the old commits too. Going to redo properly

PR #166, opened 2020-01-03
Fix is_valid call to remove size

atcuno:
@ikelos 

I am fine with the patch the way it is since task structures won't cross page boundaries, but that won't always be the case.

Could you please respond before merging with the proper (if any) way to get the size of an object like I thought .size used to do

### DIFF
--- a/volatility/framework/plugins/mac/pslist.py
+++ b/volatility/framework/plugins/mac/pslist.py
@@ -87,7 +87,7 @@ class PsList(interfaces.plugins.PluginInterface):
             else:
                 seen[proc.vol.offset] = 1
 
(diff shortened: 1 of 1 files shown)


ikelos:
Yep, it sounds like you're waiting for #159 to get reviewed and merged.  5;)  Once that's in place this pull request should be unnecessary.  5:)  Happy for you to give it a test and let me know if it resolves the issue?

atcuno:
Ah nice. It is actually unnecessary to check .size here since the allocations won't cross page boundaries, but definitely will be needed in other places. So you can either just merge this or delete it. Either is fine with me.

ikelos:
I'll go with delete, since I'd prefer to do the check if we're able, rather than relying on an assumption in some way (even if it will always be true).

PR #167, opened 2020-01-03
Mac - properly check for broken proc pointers

atcuno:


volatility/framework/plugins/mac/tasks.py
```diff
@@ -37,6 +37,8 @@ def list_tasks(cls,
         mac.MacUtilities.aslr_mask_symbol_table(context, darwin_symbols, layer_name)
 
         kernel = contexts.Module(context, darwin_symbols, layer_name, 0)
+        
+        kernel_as = context.layers[layer_name]
```
ikelos: I'll accept it, but you're still thinking in vol2 terms!  `kernel_layer` would be better these days...  5:S

PR #169, opened 2020-01-03
update sysctl v2

atcuno:


volatility/framework/symbols/mac/extensions/__init__.py
```diff
@@ -465,3 +465,30 @@ def get_address(self):
 
         return ip
 
+class sysctl_oid(objects.StructType):
+    def get_perms(self):
```
ikelos: Sorry, I know this has gone through multiple revisions, but could you please include docstrings and type information (it looks like `-> str` in both cases)?

volatility/framework/symbols/mac/extensions/__init__.py
```diff
@@ -465,3 +465,30 @@ def get_address(self):
...
+
+    def get_ctltype(self):
+        types = {1: 'CTLTYPE_NODE', 2: 'CTLTYPE_INT', 3: 'CTLTYPE_STRING', 4: 'CTLTYPE_QUAD', 5: 'CTLTYPE_OPAQUE'}
+
+        ctltype = self.oid_kind & 0xf
+
+        if 0 < ctltype < 6:
+            ret = types[ctltype]
+        else:
+            ret = ""
```
ikelos: I'm still kind of keen that this alerts a caller a bit more that something's gone very wrong, it would be better as a `renderers.UnparsableValue` or an exception of some kind, but I won't push so I'll commit it as long as the docstrings are in place...

volatility/framework/symbols/mac/extensions/__init__.py
atcuno: I will get the doc strings added and do a new commit

volatility/framework/symbols/mac/extensions/__init__.py
atcuno: There isn't anything wrong with the value. Its a bit mask of 0xf but only through 6 is defined. The code there is based on this code in the mac kernel:

http://xr.anadoxin.org/source/xref/macos-10.11.6-elcapitan/xnu-3248.60.10/bsd/kern/kern_newsysctl.c#516

In the switch, you can see see for outside the parsed types, it has a default case that just prints a newline

ikelos: Yep, ok, my bad, sorry.  I'm just eager to ensure not to let these things slip in.  5;)

PR #170, opened 2020-01-04
Mac - convert socket gathering to class method and clean up associate…

atcuno:
…d code

### DIFF
--- a/volatility/framework/automagic/mac.py
+++ b/volatility/framework/automagic/mac.py
@@ -219,6 +219,21 @@ class MacUtilities(object):
                                       context: interfaces.context.ContextInterface,
                                       task: interfaces.objects.ObjectInterface):
 
+        """Creates a generator for the file descriptors of a process
+
+        Args:
+            symbol_table_name: The name of the symbol table associated with the process
+            context:
+            task: The process structure to enumerate file descriptors from
+
+        Return:
+            A 3 element tuple is yielded for each file descriptor:
+            1) The file's object
+            2) The path referenced by the descriptor. 
+                The path is either empty, the full path of the file in the file system, or the formatted name for sockets, pipes, etc.
+            3) The file descriptor number
+        """
+
         try:
             num_fds = task.p_fd.fd_lastfile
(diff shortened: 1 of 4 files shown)


volatility/framework/automagic/mac.py
```diff
@@ -214,7 +214,8 @@ def virtual_to_physical_address(cls, addr: int) -> int:
         return addr - 0xffffff8000000000
 
     @classmethod
-    def files_descriptors_for_process(cls, config: interfaces.configuration.HierarchicalDict,
+    def files_descriptors_for_process(cls, 
+                                      symbol_table_name : str,
```
ikelos: Typically we pass the context as the first argument, but it's not a show stopper...

volatility/framework/plugins/mac/netstat.py
```diff
@@ -28,58 +30,74 @@ def get_requirements(cls):
...
         ]
 
-    def _generator(self, tasks):
-        for task in tasks:
+    @classmethod
+    def list_sockets(cls,
+                     context: interfaces.context.ContextInterface,
+                     layer_name: str,
+                     darwin_symbols: str,
+                     filter_func: Callable[[int], bool] = lambda _: False) -> \
```
ikelos: It'd be good to document this since it's an expose method for others to use.

volatility/framework/automagic/mac.py
ikelos: Also, good to have docstrings for a classmethod that's exposed as an API by this plugin.

ikelos: Any thoughts on putting `context` first?  I'd quite like people to consistently be able to intuit which arguments go in which order, and so far context has always come first if it's needed.  If there's a good reason not to, I'm open to it, but otherwise it'd make a nicer more consistent API to have it come first please...

ikelos:
Ok, it looked good, so I've rebased it and I'll apply it now (but will also keep an eye on the buildbot to make sure the rebase hasn't caused any issues)...

PR #171, opened 2020-01-04
ResourceAccessor: fix unverified retrieval for ssl.SSLError

iMHLv2:
On my Mac, I'm running Python 3.6.3 from python.org. The commit 2d3cceda only solved the problem halfway for me. This PR fixes the rest. In particular, the ssl shipped with my Python version doesn't have `ssl.SSLCertVerificationError`. Instead, it raises `ssl.SSLError` with a reason of `CERTIFICATE_VERIFY_FAILED`.

### DIFF
--- a/volatility/framework/layers/resources.py
+++ b/volatility/framework/layers/resources.py
@@ -70,7 +70,11 @@ class ResourceAccessor(object):
             fp = urllib.request.urlopen(url, context = self._context)
         except error.URLError as excp:
             if excp.args:
-                if isinstance(excp.args[0], ssl.SSLCertVerificationError):
+                unverified_retrieval = (hasattr(ssl, "SSLCertVerificationError") and isinstance(excp.args[0],
+                                                                                                ssl.SSLCertVerificationError)) or (
+                                               isinstance(excp.args[0], ssl.SSLError) and excp.args[
+                                           0].reason == "CERTIFICATE_VERIFY_FAILED")
+                if unverified_retrieval:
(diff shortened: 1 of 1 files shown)


ikelos:
Hmmm, could you check whether the exception you're encountering would be caught by `ssl.CertificateError`?  That's an alias for `SSLCertVerificationError` in python-3.7+ so should work for both.  If so, could you please update this and include a comment along the lines of `# TODO: Only necessary for < python 3.7`?  If it doens't work and you really need the basic `SSLError` then just record it here and we'll merge this.  Thanks!  5:)

iMHLv2:
I do have `ssl.CertificateError` but it is not a descendant or parent of `ssl.SSLError`. 

```
class CertificateError(exceptions.ValueError)
 |  Method resolution order:
 |      CertificateError
 |      exceptions.ValueError
 |      exceptions.StandardError
 |      exceptions.Exception
 |      exceptions.BaseException
 |      __builtin__.object

class SSLError(socket.error)
 |  An error occurred in the SSL implementation.
 |  
 |  Method resolution order:
 |      SSLError
 |      socket.error
 |      exceptions.IOError
 |      exceptions.EnvironmentError
 |      exceptions.StandardError
 |      exceptions.Exception
 |      exceptions.BaseException
 |      __builtin__.object
```

ikelos:
Yeah, ok.  The python folks have the exception, but openssl apparently still raises it as SSLError.  So much for appropriate exceptioning...  5:S  Alright, I'll put it in, but it looks really cumbersome (it's good code and it's good that it narrows down the checks, it just feels... ugly).  5:S

PR #172, opened 2020-01-04
catch more exceptions traversing windows vads

iMHLv2:
Catch more exceptions when traversing Windows VADs. In particular, catch `InvalidAddressException` on the LeftChild and RightChild separately, so an exception on one doesn't prevent traversing the other. Log exceptions in either case. I observed this issue when trying to parse VADs for a terminated process.

### DIFF
--- a/volatility/framework/symbols/windows/extensions/__init__.py
+++ b/volatility/framework/symbols/windows/extensions/__init__.py
@@ -101,11 +101,17 @@ class MMVAD_SHORT(objects.StructType):
             vad_object = self.cast(target)
             yield vad_object
 
-        for vad_node in self.get_left_child().dereference().traverse(visited, depth + 1):
-            yield vad_node
+        try:
+            for vad_node in self.get_left_child().dereference().traverse(visited, depth + 1):
+                yield vad_node
+        except exceptions.InvalidAddressException as excp:
+            vollog.log(constants.LOGLEVEL_VVV, "Invalid address on LeftChild: {0:#x}".format(excp.invalid_address))
 
-        for vad_node in self.get_right_child().dereference().traverse(visited, depth + 1):
-            yield vad_node
+        try:
+            for vad_node in self.get_right_child().dereference().traverse(visited, depth + 1):
+                yield vad_node
+        except exceptions.InvalidAddressException as excp:
+            vollog.log(constants.LOGLEVEL_VVV, "Invalid address on RightChild: {0:#x}".format(excp.invalid_address))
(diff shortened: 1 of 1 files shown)


iMHLv2:
I don't feel like we need `ignore_errors: bool = True` but happy to add it if you feel strongly. There are a lot of other instances in the framework that catch `InvalidAddressException` and either just continue, output a debug or log message, or use an alternate value. That might sound like a vote to add `ignore_errors: bool = True` everywhere, but on the contrary, I don't think it would be a heavily used option.

ikelos:
Fair enough, could you point out a few examples so I can see if I've let this kind of thing through before please?  I'm just really against catching errors that hide whether something wasn't right and I thought I'd been pretty strict on not allowing that kind of thing into the core code (in plugins, I have no problem with it)...

iMHLv2:
@ikelos Here are a few,  but there are others. Hopefully I'm not mis-reading what you were looking for. 

https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/extensions/services.py#L113

https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/extensions/services.py#L124

https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/extensions/registry.py#L87

https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/extensions/registry.py#L156

https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/extensions/registry.py#L203

https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/extensions/pool.py#L125

https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/extensions/pool.py#L141

https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/extensions/__init__.py#L58

https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/extensions/__init__.py#L282

https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/extensions/__init__.py#L447

ikelos:
Ok, that's a pretty conclusive list, looks like I have before, thanks!  5:)  Cool, I'll let it through this time then too.  It might be worth adding to the documentation the behaviour this uses, but that can happen in another commit.  I'm happy to merge this as is...  5:)

PR #174, opened 2020-01-12
expose process_dump to other plugins

doomedraven:
@ikelos can you check it please? and any suggestion on process dump proc description as we had in vad_dump?

### DIFF
--- a/volatility/framework/plugins/windows/procdump.py
+++ b/volatility/framework/plugins/windows/procdump.py
@@ -18,6 +18,8 @@ vollog = logging.getLogger(__name__)
 class ProcDump(interfaces.plugins.PluginInterface):
     """Dumps process executable images."""
 
+    _version = (1,1,0)
+
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
         # Since we're calling the plugin, make sure we have the plugin's requirements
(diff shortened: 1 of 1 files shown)


volatility/framework/plugins/windows/procdump.py
```diff
@@ -32,54 +34,61 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
             requirements.PluginRequirement(name = 'pslist', plugin = pslist.PsList, version = (1, 0, 0)),
         ]
 
-    def _generator(self, procs):
-
+    @classmethod
+    def process_dump(cls, proc)
```
ikelos: Ideally provide documentation for what this function does in the docstring.  Also, this needs properly parameterizing (anywhere you're using `self` won't work, so things from the config pass in as something like `symbol_table` or similar.  You'll also need to pass in `context` rather than using `self._context` or `self.context`.

It would also be nice to type annotate this, but not to worry if that's problematic.  Lastly, there needs to be a colon on the end of this line...  5;)

doomedraven: ah true a very nice point about `self`, but another plugin already should all that defined no? i will need to rework that part, the rest is fixed but not pushed yet

ikelos: It will, but we try not to use inheritance, and instead call class methods (hence the first parameter is cls).  This means that functionality inside a plugin doesn't suddenly change without warning, but is explicitly stored in the plugin that defined it and calls out to the plugin it needs...

volatility/framework/plugins/windows/procdump.py
```diff
@@ -35,60 +35,53 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
         ]
 
     @classmethod
-    def process_dump(cls, proc)
-        pe_table_name = intermed.IntermediateSymbolTable.create(self.context,
-                                                                self.config_path,
+    def process_dump(cls, proc, context: interfaces.context.ContextInterface, config_path: str, config: dict):
```
ikelos: Sorry I didn't mention this earlier, but we try to request the context first, then things like `symbol_table_name` and then finally the `proc`.  As for the `config_path`, it's only used for creating the PE table (which only needs to happen once), so I'd suggest pulling that out of this method, and adding a parameter `pe_table_name`.

So I'd expect the final method signature to be something like:
```
def process_dump(cls, context: interfaces.context.ContextInterface, kernel_table_name: str, pe_table_name: str, proc: interfaces.objects.ObjectInterface) -> interfaces.plugins.FileInterface:
   """Extracts the complete data for a process as a FileInterface

   Args:
     context: the context to operate upon
     kernel_table_name: the name for the symbol table containing the kernel's symbols
     pe_table_name: the name for the symbol table containing the PE format symbols
     proc: the process object whose memory should be output

   Returns:
     A FileInterface object containing the complete data for the process
"""
```

Sorry to keep giving more specific requests, but we still haven't gotten a style guide in place, so sadly it's just in my head so I end up forgetting to mention bits that we'd prefer to be done a particular way.  Sorry!  5:S

volatility/framework/plugins/windows/procdump.py
```diff
@@ -35,60 +35,53 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
         for proc in procs:
-            process_name = utility.array_to_string(proc.ImageFileName)
-            filedata, result_text = self.process_dump(proc)
-            self.produce_file(filedata)
-            result_text = "Stored {}".format(filedata.preferred_filename)
+            try:
+                process_name = utility.array_to_string(proc.ImageFileName)
+                filedata = self.process_dump(proc, self.context, self.config_path, self.config)
+                self.produce_file(filedata)
+                result_text = "Stored {}".format(filedata.preferred_filename)
```
ikelos: Looks much better, thanks!  5:)

doomedraven:
@ikelos  im more than happy to do all changes :P 

here is the question, is this correct? passing `self.config["nt_symbols"]` as `kernel_table_name`?
```
filedata = self.process_dump(self.context, self.config["nt_symbols"], pe_table_name, proc)

def process_dump(cls, context: interfaces.context.ContextInterface, kernel_table_name: str, pe_table_name: str, proc: interfaces.objects.ObjectInterface) -> interfaces.plugins.FileInterface:
      
```

doomedraven:
great, now i need to test it hard again and check why there so huge diff between vol2 and vol3 dumped processes

ikelos:
Hmm, so taking a quick look into, the `reconstruct` method says that it's the equivalent of using `--memory` in volatility 2 (and sure enough the output with `--memory` seems identical) so it must just be the fix-ups to get the process back into executable format.

doomedraven:
i will check, bcz i had problems with one plugin, once i will have more details i will provide them in slack

PR #175, opened 2020-01-13
Issue mac tasks

ikelos:
As of commit `d7f5a13c467878e7927fe55c8ba7b28a3fc56e77`, buildbot broke on the mac tasks test.  Looking at the change, I thought this was because `not filter_func(proc)` could be checked if `proc` were invalid (hence adding `fda64dde958c6e236ca099b52fcce88884ad13b0`).

It turns out that due to the fix, it no longer stops as soon as it hits an error (meaning the next item in the queue is requested, which is where the break actually happened).  There was exception handling around the first item read of `next` or `prev`, but not around the later attempts to access the next item.  I've extended the exception handling around each of the loops, and then optimized it so that the loop code isn't duplicated but parameterized instead.

Please could you check that this behaves as you'd expect, and if so either directly merge it or let me know and I'll merge it.  I think it's right, so I'll merge it in 7 days if you'd don't but there might be some subtlety that I haven't spotted so better if you could review it before then please...  5:)

### DIFF
--- a/volatility/framework/symbols/mac/extensions/__init__.py
+++ b/volatility/framework/symbols/mac/extensions/__init__.py
@@ -390,27 +390,17 @@ class queue_entry(objects.StructType):
                   max_size: int = 4096) -> Iterable[interfaces.objects.ObjectInterface]:
         yielded = 0
 
-        try:
-            n = self.next.dereference().cast(type_name)
-            while n is not None and n.vol.offset != list_head:
-                yield n
-                yielded = yielded + 1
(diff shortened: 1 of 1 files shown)


ikelos:
Scratch that, I'll leave it until after our next meeting (which is only 2 days after that).  5:)

ikelos:
Ok, having discussed this with @atcuno who's agreed to review it (but because it is breaking buildbot) I'm resetting the merge window for this to the 27th if there's been no activity...

PR #176, opened 2020-01-14
Procdump: Don't loop through twice without good reason.

ikelos:
Just taking a look at the procdump code (with a view towards rebuild EXEs rather than just memory) and it seemed as though we went through the list twice, without overly good reason?  `VirtualSize` seems like a reasonable way of determining how big the section should be, rather than doing a weird dance of figuring out the difference between the previous section and the next section?

As such, I've hopefully condensed this down into a single loop through, but I'm really not sure why we do it this way, so I'm nervous about doing the update.

Technically the `VirtualSize` could end up being huge (so mebbe we should have a max limit)?  Also, rounding up could make it stretch outside of its bounds, but since they're layered in order, new sections will start at their correct offset, so there should be no issue with that...

Anyway, this is auxiliary rather than necessary, so no time limit on the review, but it'd be nice to consider it before we put in reconstructing EXEs.  5:)

### DIFF
--- a/volatility/framework/symbols/windows/extensions/pe.py
+++ b/volatility/framework/symbols/windows/extensions/pe.py
@@ -110,8 +110,10 @@ class IMAGE_DOS_HEADER(objects.StructType):
         fixed_data = self.fix_image_base(raw_data, nt_header)
         yield 0, fixed_data
 
-        prevsect = None
-        sect_sizes = []
+        start_addr = nt_header.FileHeader.SizeOfOptionalHeader + \
+                     (nt_header.OptionalHeader.vol.offset - self.vol.offset)
+
+        counter = 0
         for sect in nt_header.get_sections():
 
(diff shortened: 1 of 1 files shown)


iMHLv2:
@ikelos I reviewed this and the plugin seems to function the same as before, at least in terms of the output it produces. No problem merging this PR in my opinion. 

When testing, I compared volatility2 output with volatility3 master and volatility3 `improve-procdump-doubleloop`. Regarding the latter, I created a local branch with recent changes from master, otherwise it was complaining about `ObjectTemplate object has no attribute size`. 

Lastly, this also affects moddump and dlldump, so I tested those plugins as well. To complete the dlldump testing, I needed to cherry-pick the commit from here: 

https://github.com/volatilityfoundation/volatility3/pull/172

So in summary, we're good to merge this, and I'll go take a closer look at 172 so we can merge that as well.

PR #185, opened 2020-01-30
Non linear support

ikelos:
This adds support for NonLinearlyMapped layers (ones where the domain vector doesn't translate directly to a vector of the same length in the range of the function).

It makes some reasonably deep changes (notably, the _scan_iterator that returns chunks of space to scan need to know the domain size, not just the range size of the chunks it gets, so segments must maintain, for linear ones the last two fields of a segment are always identical, but for non-linear they may not be).

This also adds decode/encode functions to all translation layers where data read at the range's vector can be decoded to fill the domain vector provided.

I'd very much like this reviewed (a fair amount) because it's pretty deep in the core.  Things to check for are:

* Efficiency (has this slowed things down noticably?)
* Accuracy (do things map the way we expect)
* Ease of use (are the API and changes to existing code sensible/easy to use)

### DIFF
--- a/volatility/framework/automagic/pdbscan.py
+++ b/volatility/framework/automagic/pdbscan.py
@@ -323,7 +323,7 @@ class KernelPDBScanner(interfaces.automagic.AutomagicInterface):
             try:
                 kvp = vlayer.mapping(kvo, 0)
                 if (any([(p == kernel['mz_offset'] and layer_name == physical_layer_name)
-                         for (_, p, _, layer_name) in kvp])):
+                         for (_, _, p, _, layer_name) in kvp])):
                     valid_kernels[virtual_layer_name] = (kvo, kernel)
                     # Sit the virtual offset under the TranslationLayer it applies to
                     context.config[kvo_path] = kvo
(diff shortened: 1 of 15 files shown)


ikelos:
Something particularly to look at @awalters and @npetroni is whether the decode/encode API makes sense and takes enough parameters.  My hope is that the data passed in is enough to locate the outer block that's been "encoded" (read compressed, encrypted, both, etc), decrypt it and return the length from the right offset in the domain (the top of the layer) for the right length (from the top of the layer).

We could change the method to handle in the complete 4-tuple (offset, length, mapped_offset, mapped_length) but I didn't know if the mapped_length would get used much (particularly as it *should* be derivable from the length of the data passed in).  Another option would be to do the reads within the decode method (which might allow more flexibility in the encoding/decoding, although there's technically nothing to stop additional reads/writes at that point).

Anyway, please lemme know what you think.  I don't have a merge limit on this, but since it's quite a deep change, I'm keen to get it in sooner than have people start building low-level stuff and then be hit by the change...

Issue #157, opened 2019-12-11
Demangle VC++ names from PDB


tklengyel:
The Windows kernel PDBs contain a lot of constants that are currently saved in the json in their mangled format, such as:

```
"??_C@_1BM@FMDCFHKI@?$AAM?$AAe?$AAm?$AAo?$AAr?$AAy?$AAM?$AAa?$AAn?$AAa?$AAg?$AAe?$AAr@NNGAKEGL@": {
      "address": 7877184
    },
```

In the Rekall version of the JSON we can see the unmangled version of such strings:

```
  "str:MemoryManager": 7877184, 
```

Implementing something similar to [Rekall's Demangler](https://github.com/google/rekall/blob/8db49aa92439d4787023c1ea37561b6903cabf9b/rekall-core/rekall/plugins/overlays/windows/pe_vtypes.py#L1057) would improve the usefulness of the Volatility IST jsons.

ikelos:
Thanks for the suggestion.  Do you have a specific scenario in mind where you'd make use of those names?  There are means to do proper demangling (which we've investigated) including shelling out to llvm to do it, but they've all been suboptimal for one reason or another.

I'd be more keen on creating a sanitzer that would allow the symbols to be more easily referenced within volatility, but I'm not sure including special symbols (such as the colon from your example) will make them easier to access from inside the code?

I'm not much of a C/C++ person myself though, so I'll pass this over to our symbols guy to see if he has any thoughts on the matter...  5:)

tklengyel:
We are using these jsons outside Volatility (with [LibVMI](https://github.com/libvmi/libvmi)), so it would be preferred if the json itself has the demangled strings. There is no particular scenario (other then general reverse engineering stuff), but we would like to see feature parity betweet the Volatility jsons and the Rekall jsons. Since Rekall became unmaintained it has been a pain to get it working with newer versions of Python so we'd like to drop it as soon as possible.

PR #189, opened 2020-02-13
pdbconv: save _imp functions in JSON

### LINKED ISSUE #184: pdbconv: include _imp_ functions in the JSON
Some internal Windows functions are currently missing from the JSON generated by pdbconv. For example, some syscall handlers in the SSDT on Windows 10 1903 point to functions whose address is not currently present in the JSON. These functions are effectively just stub functions that load the real syscall address into r10.

For example:
```
    "NtCommitTransaction": {
      "address": 5682416,
      "linkage_name": "__imp_NtCommitTransaction"
    },
```

The SSDT for this syscall doesn't point to this address.  This address is where the actual `implementation` of the function is located. Comparing the JSON to one generated by Rekall we see the following:
```
  "_imp_NtCommitTransaction": 5682416,
  "NtCommitTransaction": 1676048, 
```

The SSDT in fact points to stub function at kernel_base + 1676048, which then loads the address of kernel_base + 5682416 to r10. So pdbconv should include the address for both functions in the JSON and switch their naming so that it matches up with the SSDT.

tklengyel:
Don't filter out functions which begin with `__imp` since these functions are separate from their "full" version. The SSDT often points to the `__imp` functions in Windows 10.

Fixes #184

### DIFF
--- a/volatility/framework/symbols/windows/pdbconv.py
+++ b/volatility/framework/symbols/windows/pdbconv.py
@@ -520,11 +520,7 @@ class PdbReader:
         """Strips unnecessary components from the start of a symbol name."""
         new_name = name
 
-        if new_name[:7] in ["__imp__", "__imp_@"]:
-            new_name = new_name[7:]
-        elif new_name[:6] in ["__imp_"]:
-            new_name = new_name[6:]
-        elif new_name[:1] in ["_", "@", "\u007F"]:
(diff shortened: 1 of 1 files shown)


ikelos:
Cool, when we get #157 sorted, this may all move to the `linkage_name` field and we'll use the demangled names for the primary lookup, but I'll commit it for now regardless.  Thanks!  5:)

PR #190, opened 2020-02-13
Mac kernel handlers

atcuno:


volatility/framework/plugins/mac/check_syscall.py
```diff
@@ -23,13 +24,16 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
             requirements.TranslationLayerRequirement(name = 'primary',
                                                      description = 'Memory layer for the kernel',
                                                      architectures = ["Intel32", "Intel64"]),
-            requirements.SymbolTableRequirement(name = "darwin", description = "Mac kernel symbols")
+            requirements.SymbolTableRequirement(name = "darwin", description = "Mac kernel symbols"),
+            requirements.PluginRequirement(name = 'lsmod', plugin = lsmod.Lsmod, version = (1, 0, 0))
         ]
 
-    def _generator(self):
+    def _generator(self, mods: Iterator[Any]):
```
ikelos: Since this is not a class method, it's ok to use `self.config` to look up things like `primary` and `darwin`.  Also, because it's private and can only be used from this class, it seems a bit strange to have some functionality in the run method, and the rest in the generator?  I'd move the mods creating code into the top of the generator and then everything's easy to read in one go (unless you're intending to use _generator from within this class more than once?).  Not a show stopper, but seems a bit weird given the method can't ever be called by anyone else...

volatility/framework/plugins/mac/check_sysctl.py
```diff
@@ -109,30 +111,28 @@ def _process_sysctl_list(self, kernel, sysctl_list, recursive = 0):
             except exceptions.InvalidAddressException:
                 break
 
-    def _generator(self):
+    def _generator(self, mods: Iterator[Any]):
```
ikelos: Again, odd to parameterize the private `_generator` method if it doesn't get called more than once.

volatility/framework/automagic/mac.py
```diff
@@ -118,6 +118,66 @@ def stack(cls,
 class MacUtilities(object):
     """Class with multiple useful mac functions."""
 
+    @classmethod
+    def mask_mods_list(cls,
```
ikelos: Are `Module` objects what we're doing here, or is it a specific type?  If so, would it be useful code for all `Module` objects?  It feels like it's related to [the `mask_symbol_table` function](https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/__init__.py#L248) in some way?

volatility/framework/plugins/mac/check_trap_table.py
```diff
@@ -24,35 +25,40 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
             requirements.TranslationLayerRequirement(name = 'primary',
                                                      description = 'Memory layer for the kernel',
                                                      architectures = ["Intel32", "Intel64"]),
-            requirements.SymbolTableRequirement(name = "darwin", description = "Mac kernel symbols")
+            requirements.SymbolTableRequirement(name = "darwin", description = "Mac kernel symbols"),
+            requirements.PluginRequirement(name = 'lsmod', plugin = lsmod.Lsmod, version = (1, 0, 0))
         ]
 
-    def _generator(self):
+    def _generator(self, mods: Iterator[Any]):
```
ikelos: Same again

volatility/framework/plugins/mac/timers.py
```diff
@@ -25,13 +26,16 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
             requirements.TranslationLayerRequirement(name = 'primary',
                                                      description = 'Memory layer for the kernel',
                                                      architectures = ["Intel32", "Intel64"]),
-            requirements.SymbolTableRequirement(name = "darwin", description = "Mac kernel symbols")
+            requirements.SymbolTableRequirement(name = "darwin", description = "Mac kernel symbols"),
+            requirements.PluginRequirement(name = 'lsmod', plugin = lsmod.Lsmod, version = (1, 0, 0))
         ]
 
-    def _generator(self):
+    def _generator(self,  mods: Iterator[Any]):
```
ikelos: And one more  5;)

volatility/framework/plugins/mac/check_sysctl.py
atcuno: I moved the module gathering for all plugins into the generator with my latest commit

volatility/framework/automagic/mac.py
atcuno: It is a list of linux kernel modules gathered by walking the list. To do a proper symbol lookup we have to mask the addresses to mask how Volatility treats them.

ikelos: Ok, cool.  I'm happy to commit it you are?

PR #199, opened 2020-03-05
Issue151 flatten anonymous

### LINKED ISSUE #151: Modern linux types can sometimes just be levels of indirection
Just so we've got it recorded somewhere...

This has been noticed specifically on linux (in the 5.3.0 kernel, at least), but certain types (such as `mm_struct`) can contain `unnamed_field`s which just act as levels of indirection (ie, they just contain another struct, unnamed, without much purpose or reason to need to access it).

```
    "mm_struct": {
      "size": 1024,
      "fields": {
        "cpu_bitmap": {
          "type": {
            "count": 0,
            "kind": "array",
            "subtype": {
              "kind": "base",
              "name": "long unsigned int"
            }
          },
          "offset": 1024
        },
        "unnamed_field_0": {
          "type": {
            "kind": "struct",
            "name": "unnamed_b06fa817540c10e0"
          },
          "offset": 0
        }
      },
      "kind": "struct"
    },
```

This makes accessing members of `mmstruct` difficult without knowing the precise sub-structure they're within, and we might want to contemplate a way that we can reasonably remove these if they're literally just unnamed struct members (unions and other types might need more thinking about)...

ikelos:
Add in support for flatten anonymous, and whilst adding the new schema solidify what information is and isn't allowed in the linux/mac metadata areas.

Resolves #151

### DIFF
--- a/volatility/framework/symbols/intermed.py
+++ b/volatility/framework/symbols/intermed.py
@@ -593,3 +593,49 @@ class Version6Format(Version5Format):
 class Version7Format(Version6Format):
     """Class for storing intermediate debugging data as objects and classes."""
     version = (6, 1, 0)
+
+
+class Version8Format(Version7Format):
+    """Class for storing intermediate debugging data as objects and classes."""
+    version = (6, 2, 0)
+
+    def _reduce_indirect_members(self, members: Dict[str, Dict[str, Any]]) -> Dict[str, Dict[str, Any]]:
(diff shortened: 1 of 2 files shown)


ikelos:
Ok, cool.  We'll sit on it until you guys are happy with the ELF/macho fields you want, thanks for giving it a look over!  5:)

PR #207, opened 2020-04-08
Store and re-store the original get_symbol method masking

ikelos:
Stores the original get_symbol method on the symbol table that's being adapted, so that if re-applied the original results can be masked, rather than masking twice.

Fixes issue 206.


ikelos:
@atcuno So this slightly changes the semnatics of the aslr_masking, so that it returns a new table with the offsets applied (so the symbol tables within the context can be treated as essentially immutable).  This better allows people to figure out if they're double masking, or anything else if they're reusing a symbol table for any reason.

One issue I ran into is that for some reason multiple masked symbol tables were constructed (not overly an issue), but then there were parts of the linux file path code that checked whether multiple symbols were returned, and failed if there were.  As it turns out, they were the same symbol from copied tables, so I narrowed down the symbol search to just the table involved, but this kind of thing may crop up elsewhere, which is why I'm keen that you review it.

atcuno:
There is a issue with this branch when multiple code paths adjust the table, which is a common pattern it causes.

As an example that triggered it, running mac.lsmod from this branch works the same as master, but mac.check_syscall causes a backtrace on this branch but works as expected in master.

 If you look at the code for this plugin, it relies on list_modules, which uses the new masked_symbol pattern:

https://github.com/volatilityfoundation/volatility3/blob/issue206_replace-mask-for-mask_symbol_table/volatility/framework/plugins/mac/lsmod.py#L29

This works fine for lsmod as its only called once. It breaks check_syscall though in this function call:

https://github.com/volatilityfoundation/volatility3/blob/issue206_replace-mask-for-mask_symbol_table/volatility/framework/plugins/mac/check_syscall.py#L39

which leads to here:

https://github.com/volatilityfoundation/volatility3/blob/c5c2423b4886060ff7ed6eff35fd5db905e03036/volatility/framework/automagic/mac.py#L155

and when the module list is walked for masking, this causes a call to the same list_modules function as lsmod, but the symbol table being passed in is already masked by code earlier in check_syscall. This causes the ASLR shift to get applied twice, which points the kmod symbol to an invalid address and causes a paging exception.

ikelos:
Hmmm,

That's not supposed to happen due to the check we do for an existing mask:

https://github.com/volatilityfoundation/volatility3/pull/207/commits/488eaf9376490631edde00c627be26f36167c069#diff-fb5f2987971f0a1c9567674efa3bb39eR257

So if a symbol table's been masked this way, the original symbol table should be used instead.  I'll try and recreate what you're seeing, but the intended operation of this is that applying the change overwrites any other masks that may have been applied (which I think is the correct thing to do), I just don't know why it doesn't do it...  5:S

ikelos:
I've just had a successful run of linux.check_syscall both on this branch and master and they return the same results?  Is it a specific image that would cause the call to happen twice, or were you reviewing an older version of this branch?

ikelos:
@atcuno I've reworked this branch, so now the ASLR masking is part of the automagic, and the table you're left with is the modified one.  It's possible to turn it off in the automagic if necessary, and the rest of the table machinery should work as expected.  There will be an original table still, and a masked one present, but the config variable should be filled with the masked table's name (ie, you don't need to worry about it as long as you haven't hardcoded the table name anywhere).  Essentially, whatever the OS, the symbol table you're given by default is masked, but if you want to construct your own, you'll need to go through the steps of masking it yourself.  I've also now removed the "if we're given a value of 0, assume we need to find_the_aslr" parameter overloading.  If someone passes in 0, they meant 0, if they don't pass anything in, assume they meant 0.  Also, at the moment, it's not cumulative, but we could make it cumulative if you wanted, as long as we specify which we're doing?

I think this'll be more efficient, because it shouldn't double mask unless explicitly requested, and keeps pretty much everyone happy (or should do).  Can you think of anything I've missed?  It should now apply to both linux and mac, and I've tried mac.pslist, linux.pslist and mac.check_syscall, and they all seem to work fine?

ikelos:
It also just occurred to me that moving this to the automagic may require parameterizing the symbol tables in some way (in order to recreate them correctly from a config file, if the automagic hasn't been used to find the values).  I'll need to think over this a little bit more, so don't worry about the review for now...

ikelos:
Ok, this has been completely reworked, and now IntermediateSymbolTables have a configuration option to set the shift.  This is automatically set in linux/mac automagic, and no additional symbol table is stored.  The reason for this conversion is to allow the shift to be recorded in the configuration so it can be recreated without the automagic (and therefore also altered).  @atcuno Please could we you review this version now it's ready...

It should also properly handle the configuration being partially populated (the old code had issues around SymbolTableRequirements).

atcuno:
It looks good except that the address masking is no longer being applied, so then when plugins use [lookup_module_address](https://github.com/volatilityfoundation/volatility3/blob/0c43beb42d3f1d7e3c7ac6ccb97c8f95c438042f/volatility/framework/automagic/mac.py#L159) to map an address to a symbol, it fails because [get_symbols_by_location](https://github.com/volatilityfoundation/volatility3/blob/0c43beb42d3f1d7e3c7ac6ccb97c8f95c438042f/volatility/framework/automagic/mac.py#L168) misses it as a full 64 bit address from the symbol table is being compared to a truncated one.

This is currently handled in master by the call here: https://github.com/volatilityfoundation/volatility3/blob/0c43beb42d3f1d7e3c7ac6ccb97c8f95c438042f/volatility/framework/automagic/mac.py#L194

ikelos:
Hmmm, that's probably a bug in `lookup_module_address` which I guess should be masking it based on the size of the layer's address mask...  Ugh.  Nasty.  Ok, I'll take a look into it and see if I can find a solution.  Thanks for reviewing it!  5:)

ikelos:
Ok, I think I've now catered by that, by adding another parameter.  This means that when a symbol table is constructed automagically, it's designed specifically for one layer (because it masks to that layer and ASLR shifted by the layer).  It doesn't prevent one manually creating another layer without the masking and shifting (which can be changed in the configuration before construction of the table) and it can be used by windows symbol tables too if needed.

I may need to think about altering it, so that the values are cached at creation time, otherwise if the config changes and new (never before seen) symbols are requested, they made end up shifted differently, which I think would be more confusing than expecting the values to be set at the time of table creation...

Issue #186, opened 2020-02-12
hivescan/hivelist issues


superponible:
**Describe the bug**

There are a few issues with hivelist and hivescan.  

1. hivelist only walks the list forward now

There was a change I tracked back to this commit https://github.com/volatilityfoundation/volatility3/commit/2cad54ea6bc7a6a016f1624671a1e000a0743e85 that prevented the list walking from raising an exception, so the walk backward is no longer performed.

2. there are cases where walking the list forward and backward can still miss hives

Sometimes invalid addresses (paged, compressed) are encountered when walking the list.  If this happens only once in the list, then walking backward from CmpHiveListHead is sufficient.  However, if multiple links are broken, it can leave the middle of the list unreachable.  E.g., if the list has 10 items, and the 3rd and 7th links are invalid when walking the list, the middle items can't be reached by walking forward or backward.  In this case, reverting to scanning may be necessary to find the remaining hives.

3. Looks like some changes to the pool scanner have broken hivescan

4. hivescan doesn't work on Win8.1 and Win10 x64 because it's doesn't handle big pool pages.

superponible:
Started this branch with initial attempts at fixing 1, 2, and 3: https://github.com/volatilityfoundation/volatility3/tree/hivelist_hivescan_fixes

@ikelos I haven't gone over the pool scan changes in detail, but it looked like the type check was based on the _POOL_HEADER which _CMHIVE doesn't have, so that's why I just had it skip, not sure if there's a better fix.

ikelos:
Cool, looks good.  Is it ready to be turned into a pull request or are you still working on it?  It all looks great, my spidey sense is tingling about the `is_valid` call, but I think I just had an issue with people relying on them too much/them not being implemented everywhere in vol2.  I don't think I've got a legitimate reason for not liking it (and the code inside it seems right)...  5:)

Yeah, if the pool header doesn't exist, then skipping the test should be fine...

What needs to happen for big pool page support?  Is it involved or just a little extra code to get going?  Happy to take a look at it when I get some cycles if it's not a major overhaul?  5:)

PR #213, opened 2020-05-05
issue #186 - Hivelist hivescan fixes

superponible:


volatility/framework/plugins/windows/bigpools.py
```diff
@@ -0,0 +1,138 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
```
ikelos, line 1: We should check with @awalters whether we need to update this header to 2020 for new files?

volatility/framework/plugins/windows/bigpools.py
```diff
@@ -0,0 +1,133 @@
...
+            symbol_table: The name of the table containing the kernel symbols
+            tags: An optional list of pool tags to filter big page pool tags by
+
+        Yields:
+            A big page pool object
+        """
+        kvo = context.layers[layer_name].config['kernel_virtual_offset']
+        ntkrnlmp = context.module(symbol_table, layer_name = layer_name, offset = kvo)
+
+        big_page_table_offset = ntkrnlmp.get_symbol("PoolBigPageTable").address
```
ikelos, line 65: Can we rely on this symbol always existing?  Not something to worry about here particularly, but I can't tell if we need error handling around it...

volatility/framework/plugins/windows/registry/hivelist.py
```diff
@@ -8,10 +8,28 @@
...
 vollog = logging.getLogger(__name__)
 
 
+class HiveGenerator():
+    """Walks the registry HiveList linked list in a given direction and stores an invalid offset
+    if it's unable to fully walk the list"""
+
+    def __init__(self, cmhive, forward = True):
+        self.cmhive = cmhive
+        self.forward = forward
```
ikelos, line 22: What happens if these are changed after construction?  I know it's a little long winded, but I'd prefer they were made into properties so that they're read only.  So essentially, `self._forward = forward` and then a forward property that simply returns `self._forward` for each of these variables.

volatility/framework/plugins/windows/registry/hivescan.py
```diff
@@ -4,16 +4,21 @@
...
+from volatility.plugins.windows import poolscanner, bigpools
 
 
 class HiveScan(interfaces.plugins.PluginInterface):
     """Scans for registry hives present in a particular windows memory
     image."""
 
+    _version = (1, 0, 0)
+
+    is_windows_8_1_or_later = poolscanner.os_distinguisher(version_check = lambda x: x >= (6, 3),
```
ikelos, line 19: We probably want to keep these all in one place and pull them when needed?  It's fine keeping them here for now, but as some point we should pull all the distinguishers out into a central place...

volatility/framework/plugins/windows/registry/hivelist.py
superponible: I started making properties for those, but really `invalid` is the only one that needs to be read publicly, so I made the other private and just a property for invalid.  I think that's what you were getting at, but let me know if I misunderstood.

ikelos:
Cool, a couple things we'll need to remember to contemplate, but for now it's all good for inclusion.  Thanks!  5:)

PR #216, opened 2020-05-17
Refactor stacker automagic

ikelos:
Ok,

So it occurred to me that the LayerStacker automagic was a little non-generic and not really reusable.  Previous it look a single_location parameter that was a URI and attempted to stack as high as possible on top of, and then attach it to the configuration at a TranslationRequirement that wasn't yet satisified.

It seemed as though the "construct layers" code could be applied to any starting layer, rather than taking an explicit URI and always starting from a physical layer.  It also occurs to me that people may want to apply it without a requirement/config tree beside it.  That's surprisingly hard to arrange with the existing system, since it all gets built in a temporary context which is thrown away if it's not attached to a requirement in the original context, so I think I'm gonna leave the output bit as it (otherwise it may well clutter a config with random layers that might've stacked). 

This has been reworked, the first commit was the most basic implementation but wasn't backwards compatible.  The second commit changes names to allow for backwards compatibility of old clients (for whatever reason).

@npetroni I've asked for a review because I'm interested in your perspective on whether the change is worthwhile, and whether we being cautious about backwards compatibility (a good thing) is worth the cost of the extra code to get it going (and the use of the longer class name to not stomp on the old one)?

### DIFF
--- a/volatility/cli/__init__.py
+++ b/volatility/cli/__init__.py
@@ -256,7 +256,7 @@ class CommandLine(interfaces.plugins.FileConsumerInterface):
                 # This must be specific to get us started, setup the config and run
                 ctx.config[interfaces.configuration.path_join(current_config_path, "location")] = single_location
                 physical_layer = physical.FileLayer(ctx, current_config_path, current_layer_name)
-                ctx.config['automagic.TranslationLayerStacker.initial_layer'] = physical_layer.name
+                ctx.config['automagic.LayerStacker.single_location'] = physical_layer.name
                 ctx.add_layer(physical_layer)
 
         # UI fills in the config, here we load it from the config file and do it before we process the CL parameters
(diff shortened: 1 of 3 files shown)


npetroni:
These changes look good, the first commit in particular. I don't think we need to do anything backwards compatible, since I doubt there are any programs in the real world that we would break. I can live with the second commit though. If we keep it, I think we need to update the [comment](https://github.com/volatilityfoundation/volatility3/blob/c252b7a360f897dad4575688c002f2e5274bec4e/volatility/cli/__init__.py#L242) in cli/__init__.py that refers to LayerStacker.

ikelos:
Awesome, I think the first commit will be cleaner easier to look after, so I'm happy with the design decision to potentially break rare cases (particularly pre-initial release)...  I'll revert the second commit and then merge it.  Thanks for the review!  5:)

ikelos:
Hmmm, I'm actually not too keen on this.  The automagic does take a parameter, but it's no longer one that needs exposing in the CLI automatically, and the name `single_location` isn't accurate any more, it really should be `initial_layer`.  5:S

So the question is, should the automagic requirements be assumed to be for a UI to configure and that's it, or not?  I think the answer is, we leave the automagic as it was, but split the core out into a class method that can be called.  5:S  I think I'll revert the entire branch and come back with a second attempt...

PR #220, opened 2020-05-25
Renderers: Support partial results during population

ikelos:


ikelos:
This seems like it should be reasonable, but it opens the possibility to user interfaces just squashing any errors that the framework tries to raise, but I guess that is up tot he interface to decide.  If there's no concerns raised and it solves the linked problem, then I'll push it.

garanews:
Thanks!

PR #221, opened 2020-05-27
fix off by one and loop counters

superponible:


ikelos:
Yep, looks good, thanks!  5:)

PR #223, opened 2020-06-03
Registry stringlike data

ikelos:
Here's a partial resolution to ensuring the registry API returns all data, but hints that it's likely to be a string (and leaves it to the renderer to determine how to render it).  Any renderer that does not support the StrLike hint will display the bytes instead.


ikelos:
Cool, very happy with this, all the components feel like they're in the right place.  Since it now handles ints, strings and binary data I've renamed it to MultiTypeData.  I'm happy to rename it to RegistryData give it does special splitting on null characters, but equally I'm slightly against having a windows specific data type for rendering.  Please check that the rebasing I did worked correctly and everything's still in the right place and works as intended and then I'll merge it...  5:)

ikelos:
Awesome, thanks for getting it sorted!  5:)

PR #226, opened 2020-06-04
Netscan

japhlange:
Adds netscan to vol3, currently only for Win10x64_15063.

Open questions:
- is this way of adding object structures (netscan.json) OK? How best to handle different versions? With several JSONs like with https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/services-win10-16299-x64.json etc.?
- PoolType conditional check in PoolHeaderScanner.\_\_call\_\_()  checks the tag to be odd. Doesn't work for objects with Pooltag 2 (NONPAGED) as 2%2 == 0.
                        `elif (constraint.page_type & PoolType.NONPAGED) and header.PoolType % 2 == 1:`

I've also noticed two issues which appear to be independent of this plugin as they also appear e.g. in PSScan:
- the output layout appears to be broken 
- the performance (calculation time) is worse than it was in vol2

### DIFF
--- a/volatility/framework/plugins/windows/netscan.py
+++ b/volatility/framework/plugins/windows/netscan.py
@@ -245,16 +245,13 @@ class NetScan(interfaces.plugins.PluginInterface):
                                                                 self.config["nt_symbols"],
                                                                 self.config_path)
 
-        tcp_states = self.context.symbol_space.get_enumeration(netscan_symbol_table + constants.BANG + "TCPStateEnum")
-
         for netw_obj in self.scan(self.context, self.config['primary'], self.config['nt_symbols'],
                                   netscan_symbol_table):
 
             vollog.debug("Found netw obj @ 0x{:2x} of assumed type {}".format(netw_obj.vol.offset, type(netw_obj)))
             # objects passed pool header constraints. check for additional constraints if strict flag is set.
-            if not show_corrupt_results:
-                if not netw_obj.is_valid():
-                    continue
+            if not show_corrupt_results and not netw_obj.is_valid():
+                continue
 
             if isinstance(netw_obj, network._UDP_ENDPOINT):
                 vollog.debug("Found UDP_ENDPOINT @ 0x{:2x}".format(netw_obj.vol.offset))
@@ -280,7 +277,7 @@ class NetScan(interfaces.plugins.PluginInterface):
                     proto = "TCPv?"
 
                 try:
-                    state = tcp_states.lookup(netw_obj.State)
+                    state = netw_obj.State.description
                 except ValueError:
                     state = renderers.UnreadableValue()
 
--- a/volatility/framework/plugins/windows/netscan.py
+++ b/volatility/framework/plugins/windows/netscan.py
@@ -4,10 +4,9 @@
 
 import logging
 import datetime
-from typing import Iterable, List
+from typing import Iterable, List, Optional
 
 from volatility.framework import constants, exceptions, interfaces, renderers, symbols, layers
-from volatility.framework.automagic.pdbscan import scan
 from volatility.framework.configuration import requirements
 from volatility.framework.renderers import format_hints
 from volatility.framework.symbols import intermed
@@ -39,7 +38,7 @@ class NetScan(interfaces.plugins.PluginInterface):
         ]
 
     @staticmethod
-    def create_netscan_constraints(context, symbol_table: str) -> List[poolscanner.PoolConstraint]:
+    def create_netscan_constraints(context: interfaces.context.ContextInterface, symbol_table: str) -> List[poolscanner.PoolConstraint]:
         """Creates a list of Pool Tag Constraints for network objects.
 
         Args:
@@ -218,7 +217,7 @@ class NetScan(interfaces.plugins.PluginInterface):
             _constraint, mem_object, _header = result
             yield mem_object
 
-    def _generator(self, show_corrupt_results = None):
+    def _generator(self, show_corrupt_results: Optional[bool] = None):
         """ Generates the network objects for use in rendering. """
 
         netscan_symbol_table = self.create_netscan_symbol_table()
--- a/volatility/framework/plugins/windows/netscan.py
+++ b/volatility/framework/plugins/windows/netscan.py
@@ -20,7 +20,6 @@ class NetScan(interfaces.plugins.PluginInterface):
     """Scans for network objects present in a particular windows memory image."""
 
     _version = (1, 0, 0)
-    CORRUPT_DEFAULT = False
 
     @classmethod
     def get_requirements(cls):
@@ -32,7 +31,7 @@ class NetScan(interfaces.plugins.PluginInterface):
             requirements.PluginRequirement(name = 'poolscanner', plugin = poolscanner.PoolScanner, version = (1, 0, 0)),
             requirements.BooleanRequirement(name = 'include-corrupt',
                 description = "Radically eases result validation. This will show partially overwritten data. WARNING: the results are likely to include garbage and/or corrupt data. Be cautious!",
-                default = cls.CORRUPT_DEFAULT,
+                default = False,
                 optional = True
             ),
         ]
@@ -73,10 +72,17 @@ class NetScan(interfaces.plugins.PluginInterface):
                                        page_type = poolscanner.PoolType.NONPAGED | poolscanner.PoolType.FREE)
         ]
 
-    def determine_tcpip_version(self) -> str:
+    @classmethod
+    def determine_tcpip_version(cls,
+             context: interfaces.context.ContextInterface,
+             layer_name: str,
+             nt_symbol_table: str) -> str:
         """Tries to determine which symbol filename to use for the image's tcpip driver. The logic is partially taken from the info plugin.
 
         Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            layer_name: The name of the layer on which to operate
+            nt_symbol_table: The name of the table containing the kernel symbols
 
         Returns:
             The filename of the symbol table to use.
@@ -88,7 +94,7 @@ class NetScan(interfaces.plugins.PluginInterface):
         # therefore we determine the version based on the kernel version as testing
         # with several windows versions has showed this to work out correctly.
 
-        is_64bit = symbols.symbol_table_is_64bit(self.context, self.config['nt_symbols'])
+        is_64bit = symbols.symbol_table_is_64bit(context, nt_symbol_table)
 
         if is_64bit:
             arch = "x64"
@@ -97,19 +103,18 @@ class NetScan(interfaces.plugins.PluginInterface):
 
         # the following code is taken from the windows.info plugin.
 
-        virtual_layer_name = self.config["primary"]
-        virtual_layer = self.context.layers[virtual_layer_name]
+        virtual_layer = context.layers[layer_name]
         if not isinstance(virtual_layer, layers.intel.Intel):
             raise TypeError("Virtual Layer is not an intel layer")
 
         kvo = virtual_layer.config["kernel_virtual_offset"]
 
-        ntkrnlmp = self.context.module(self.config["nt_symbols"], layer_name = virtual_layer_name, offset = kvo)
+        ntkrnlmp = context.module(nt_symbol_table, layer_name = layer_name, offset = kvo)
 
         vers_offset = ntkrnlmp.get_symbol("KdVersionBlock").address
 
         vers = ntkrnlmp.object(object_type = "_DBGKD_GET_VERSION64",
-                               layer_name = virtual_layer_name,
+                               layer_name = layer_name,
                                offset = vers_offset)
 
         vollog.debug("Determined OS Major/Minor Version: {}.{}".format(vers.MajorVersion, vers.MinorVersion))
@@ -123,7 +128,7 @@ class NetScan(interfaces.plugins.PluginInterface):
             kuser_addr = 0xFFFFF78000000000
 
         kuser = ntkrnlmp.object(object_type = "_KUSER_SHARED_DATA",
-                                layer_name = virtual_layer_name,
+                                layer_name = layer_name,
                                 offset = kuser_addr,
                                 absolute = True)
 
@@ -174,18 +179,33 @@ class NetScan(interfaces.plugins.PluginInterface):
 
         return filename, class_types
 
-    def create_netscan_symbol_table(self) -> str:
+    @classmethod
+    def create_netscan_symbol_table(cls,
+             context: interfaces.context.ContextInterface,
+             layer_name: str,
+             nt_symbol_table: str,
+             config_path: str) -> str:
         """Creates a symbol table for TCP Listeners and TCP/UDP Endpoints.
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            layer_name: The name of the layer on which to operate
+            nt_symbol_table: The name of the table containing the kernel symbols
+            config_path: The config path where to find symbol files
 
         Returns:
             The name of the constructed symbol table
         """
-        table_mapping = {"nt_symbols": self.config["nt_symbols"]}
+        table_mapping = {"nt_symbols": nt_symbol_table}
 
-        symbol_filename, class_types = self.determine_tcpip_version()
+        symbol_filename, class_types = cls.determine_tcpip_version(
+            context,
+            layer_name,
+            nt_symbol_table,
+        )
 
-        return intermed.IntermediateSymbolTable.create(self.context,
-                                                       self.config_path,
+        return intermed.IntermediateSymbolTable.create(context,
+                                                       config_path,
                                                        "windows",
                                                        symbol_filename,
                                                        class_types = class_types,
@@ -220,7 +240,10 @@ class NetScan(interfaces.plugins.PluginInterface):
     def _generator(self, show_corrupt_results: Optional[bool] = None):
         """ Generates the network objects for use in rendering. """
 
-        netscan_symbol_table = self.create_netscan_symbol_table()
+        netscan_symbol_table = self.create_netscan_symbol_table(self.context,
+                                                                self.config["primary"],
+                                                                self.config["nt_symbols"],
+                                                                self.config_path)
 
         for netw_obj in self.scan(self.context, self.config['primary'], self.config['nt_symbols'],
                                   netscan_symbol_table):
--- a/volatility/framework/plugins/windows/netscan.py
+++ b/volatility/framework/plugins/windows/netscan.py
@@ -12,7 +12,7 @@ from volatility.framework.renderers import format_hints
 from volatility.framework.symbols import intermed
 from volatility.framework.symbols.windows.extensions import network
 from volatility.plugins import timeliner
-from volatility.plugins.windows import poolscanner
+from volatility.plugins.windows import info, poolscanner
 
 vollog = logging.getLogger(__name__)
 
@@ -29,6 +29,7 @@ class NetScan(interfaces.plugins.PluginInterface):
                                                      architectures = ["Intel32", "Intel64"]),
             requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
             requirements.PluginRequirement(name = 'poolscanner', plugin = poolscanner.PoolScanner, version = (1, 0, 0)),
+            requirements.PluginRequirement(name = 'info', plugin = info.Info, version= (1, 0, 0)),
             requirements.BooleanRequirement(name = 'include-corrupt',
                 description = "Radically eases result validation. This will show partially overwritten data. WARNING: the results are likely to include garbage and/or corrupt data. Be cautious!",
                 default = False,
@@ -101,36 +102,13 @@ class NetScan(interfaces.plugins.PluginInterface):
         else:
             arch = "x86"
 
-        # the following code is taken from the windows.info plugin.
+        vers = info.Info.get_version_structure(context, layer_name, nt_symbol_table)
 
-        virtual_layer = context.layers[layer_name]
-        if not isinstance(virtual_layer, layers.intel.Intel):
-            raise TypeError("Virtual Layer is not an intel layer")
-
-        kvo = virtual_layer.config["kernel_virtual_offset"]
-
-        ntkrnlmp = context.module(nt_symbol_table, layer_name = layer_name, offset = kvo)
-
-        vers_offset = ntkrnlmp.get_symbol("KdVersionBlock").address
-
-        vers = ntkrnlmp.object(object_type = "_DBGKD_GET_VERSION64",
-                               layer_name = layer_name,
-                               offset = vers_offset)
+        kuser = info.Info.get_kuser_structure(context, layer_name, nt_symbol_table)
 
         vollog.debug("Determined OS Major/Minor Version: {}.{}".format(vers.MajorVersion, vers.MinorVersion))
 
         vers_minor_version = int(vers.MinorVersion)
-
-        # this is a hard-coded address in the Windows OS
-        if virtual_layer.bits_per_register == 32:
-            kuser_addr = 0xFFDF0000
-        else:
-            kuser_addr = 0xFFFFF78000000000
-
-        kuser = ntkrnlmp.object(object_type = "_KUSER_SHARED_DATA",
-                                layer_name = layer_name,
-                                offset = kuser_addr,
-                                absolute = True)
 
         nt_major_version = str(kuser.NtMajorVersion)
         nt_minor_version = str(kuser.NtMinorVersion)
--- a/volatility/framework/plugins/windows/netscan.py
+++ b/volatility/framework/plugins/windows/netscan.py
@@ -1,0 +1,306 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import logging
+import datetime
+from typing import Iterable, List
+
+from volatility.framework import constants, exceptions, interfaces, renderers, symbols, layers
+from volatility.framework.automagic.pdbscan import scan
+from volatility.framework.configuration import requirements
+from volatility.framework.renderers import format_hints
+from volatility.framework.symbols import intermed
+from volatility.framework.symbols.windows.extensions import network
+from volatility.plugins import timeliner
+from volatility.plugins.windows import poolscanner
+
+vollog = logging.getLogger(__name__)
+
+class NetScan(interfaces.plugins.PluginInterface):
+    """Scans for network objects present in a particular windows memory image."""
+
+    _version = (1, 0, 0)
+    CORRUPT_DEFAULT = False
+
+    @classmethod
+    def get_requirements(cls):
+        return [
+            requirements.TranslationLayerRequirement(name = 'primary',
+                                                     description = 'Memory layer for the kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
+            requirements.PluginRequirement(name = 'poolscanner', plugin = poolscanner.PoolScanner, version = (1, 0, 0)),
+            requirements.BooleanRequirement(name = 'include-corrupt',
+                description = "Radically eases result validation. This will show partially overwritten data. WARNING: the results are likely to include garbage and/or corrupt data. Be cautious!",
+                default = cls.CORRUPT_DEFAULT,
+                optional = True
+            ),
+        ]
+
+    @staticmethod
+    def create_netscan_constraints(context, symbol_table: str) -> List[poolscanner.PoolConstraint]:
+        """Creates a list of Pool Tag Constraints for network objects.
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            symbol_table: The name of an existing symbol table containing the symbols / types
+
+        Returns:
+            The list containing the built constraints.
+        """
+
+        tcpl_size = context.symbol_space.get_type(symbol_table + constants.BANG + "_TCP_LISTENER").size
+        tcpe_size = context.symbol_space.get_type(symbol_table + constants.BANG + "_TCP_ENDPOINT").size
+        udpa_size = context.symbol_space.get_type(symbol_table + constants.BANG + "_UDP_ENDPOINT").size
+
+        # ~ vollog.debug("Using pool size constraints: TcpL {}, TcpE {}, UdpA {}".format(tcpl_size, tcpe_size, udpa_size))
+
+        return [
+            # TCP listener
+            poolscanner.PoolConstraint(b'TcpL',
+                                       type_name = symbol_table + constants.BANG + "_TCP_LISTENER",
+                                       size = (tcpl_size, None),
+                                       page_type = poolscanner.PoolType.NONPAGED | poolscanner.PoolType.FREE),
+            # TCP Endpoint
+            poolscanner.PoolConstraint(b'TcpE',
+                                       type_name = symbol_table + constants.BANG + "_TCP_ENDPOINT",
+                                       size = (tcpe_size, None),
+                                       page_type = poolscanner.PoolType.NONPAGED | poolscanner.PoolType.FREE),
+            # UDP Endpoint
+            poolscanner.PoolConstraint(b'UdpA',
+                                       type_name = symbol_table + constants.BANG + "_UDP_ENDPOINT",
+                                       size = (udpa_size, None),
+                                       page_type = poolscanner.PoolType.NONPAGED | poolscanner.PoolType.FREE)
+        ]
+
+    def determine_tcpip_version(self) -> str:
+        """Tries to determine which symbol filename to use for the image's tcpip driver. The logic is partially taken from the info plugin.
+
+        Args:
+
+        Returns:
+            The filename of the symbol table to use.
+        """
+
+        # while the failsafe way to determine the version of tcpip.sys would be to
+        # extract the driver and parse its PE header containing the versionstring,
+        # unfortunately that header is not guaranteed to persist within memory.
+        # therefore we determine the version based on the kernel version as testing
+        # with several windows versions has showed this to work out correctly.
+
+        is_64bit = symbols.symbol_table_is_64bit(self.context, self.config['nt_symbols'])
+
+        if is_64bit:
+            arch = "x64"
+        else:
+            arch = "x86"
+
+        # the following code is taken from the windows.info plugin.
+
+        virtual_layer_name = self.config["primary"]
+        virtual_layer = self.context.layers[virtual_layer_name]
+        if not isinstance(virtual_layer, layers.intel.Intel):
+            raise TypeError("Virtual Layer is not an intel layer")
+
+        kvo = virtual_layer.config["kernel_virtual_offset"]
+
+        ntkrnlmp = self.context.module(self.config["nt_symbols"], layer_name = virtual_layer_name, offset = kvo)
+
+        vers_offset = ntkrnlmp.get_symbol("KdVersionBlock").address
+
+        vers = ntkrnlmp.object(object_type = "_DBGKD_GET_VERSION64",
+                               layer_name = virtual_layer_name,
+                               offset = vers_offset)
+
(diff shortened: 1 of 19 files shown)


ikelos:
First off, thanks, that's awesome!  5:)

For working with multiple versions of the code, ideally we'd write a function that identifies the library (finding the MZ header, pulling out the PDB GUID and age) and then pulls it down from the internet on first use.  I'll try to mock up a function for being able to do that at some point.

I'll take a look through the output formatting and see if I can help offer some guidance there (it may take me a little while).

As to the conditional check, I'll leave that one to @iMHLv2 to talk about.

Lastly, the performance is likely due to scanning windows 10 virtual memory.  Due to a change in Windows, they now mark some pages as valid that lead to a dummy page.  We've opted for a more robust but slightly slower algorithm for coping with this (where every page table with identical pages is ignored, whereas vol2 used to ignore every other instance of consecutive identical pages, which could lead to identical pages being ignored).  We've tried hard to optimize the speed that it operates at and at the moment given the general speed enhancements of vol3, we're ok with the speed we currently get, but we're happy to consider alternative algorithms or other optimizations that may help (for instance, we've got a branch in place that can quickly scan for allocated memory regions and then only scan those, which is the technique I believe rekall uses to get quicker speeds at the cost of potentially missing artefacts in unallocated memory).

volatility/framework/symbols/windows/extensions/pool.py
```diff
@@ -16,6 +20,7 @@ class POOL_HEADER(objects.StructType):
     def get_object(self,
                    type_name: str,
                    use_top_down: bool,
+                   kernel_symbol_table: str,
```
ikelos, line 23: This is technically a change to the API, so we'd need to make sure that all core plugins which rely on it are changed, and we'd need to bump a version number somewhere, so this bit probably requires a little more consideration.  I'm all for the change I just want to figure out how best to make it...  5:)

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,158 @@
...
+                        pid = 0
+                        proc_name = "-"
+                    # For UdpA, the state is always blank and the remote end is asterisks
+                    for ver, laddr, _ in netw_obj.dual_stack_sockets():
+                        yield (0, (format_hints.Hex(netw_obj.vol.offset), "UDP" + ver, "{}:{}".format(
+                            laddr, netw_obj.Port), "*:*", "", pid, proc_name, netw_obj.get_create_time()))
+
+            elif isinstance(netw_obj, network._TCP_ENDPOINT):
+                if netw_obj.is_valid():
+                    vollog.debug(format_hints.Hex(netw_obj.vol.offset))
```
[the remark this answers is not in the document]
japhlange: done

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,158 @@
...
+            requirements.TranslationLayerRequirement(name = 'primary',
+                                                     description = 'Memory layer for the kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
+        ]
+
+    @staticmethod
+    def create_netstat_symbol_table(context: interfaces.context.ContextInterface, symbol_table: str,
+                                    config_path: str) -> str:
+        """Creates a symbol table for TCP Listeners and TCP/UDP Endpoints.
```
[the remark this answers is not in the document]
japhlange: Thanks, but I merely copied svcscan's style :)

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,158 @@
...
+                    elif netw_obj.get_address_family() == network.AF_INET6:
+                        proto = "TCPv6"
+
+                    if netw_obj.get_owner() != None:
+                        pid = int(netw_obj.get_owner().UniqueProcessId)
+                        proc_name = netw_obj.get_owner().ImageFileName.cast(
+                            "string", max_length = netw_obj.get_owner().ImageFileName.vol.count, errors = "replace")
+                    else:
+                        pid = 0
+                        proc_name = "-"
```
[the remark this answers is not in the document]
japhlange: I used renderers.UnreadableValue(), doesn't appear to be different from NotApplicableValue etc.

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,158 @@
...
+            elif isinstance(netw_obj, network._TCP_ENDPOINT):
+                if netw_obj.is_valid():
+                    vollog.debug(format_hints.Hex(netw_obj.vol.offset))
+                    if netw_obj.get_address_family() == network.AF_INET:
+                        proto = "TCPv4"
+                    elif netw_obj.get_address_family() == network.AF_INET6:
+                        proto = "TCPv6"
+
+                    if netw_obj.get_owner() != None:
+                        pid = int(netw_obj.get_owner().UniqueProcessId)
```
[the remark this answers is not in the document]
japhlange: done and appears to work. However it might be useful to make it more stable by catching InvalidAddressExceptions when trying to dereference invalid pointers. On the other hand UniqueProcessId isn't a pointer and thus always contains *some kind* of value, which might be invalid but should not crash... thoughts?

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,158 @@
...
+                        state = ""
+
+                    yield (0, (format_hints.Hex(netw_obj.vol.offset), proto,
+                               "{}:{}".format(netw_obj.get_local_address(), netw_obj.LocalPort),
+                               "{}:{}".format(netw_obj.get_remote_address(),
+                                              netw_obj.RemotePort), state, pid, proc_name, netw_obj.get_create_time()))
+
+            elif isinstance(netw_obj, network._TCP_LISTENER):
+                if netw_obj.is_valid():
+                    vollog.debug(format_hints.Hex(netw_obj.vol.offset))
```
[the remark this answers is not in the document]
japhlange: done, same as above

volatility/framework/plugins/windows/poolscanner.py
```diff
@@ -317,6 +322,21 @@ def builtin_constraints(symbol_table: str, tags_filter: List[bytes] = None) -> L
                            type_name = symbol_table + constants.BANG + "_CMHIVE",
                            size = (800, None),
                            page_type = PoolType.PAGED | PoolType.NONPAGED | PoolType.FREE),
+            # TCP listener
```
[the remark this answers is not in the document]
japhlange: pulled it back to `netscan`. Should be able to be used by other plugins by calling `create_netscan_constraints()`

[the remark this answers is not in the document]
ikelos: Looks good  5:)

volatility/framework/plugins/windows/netscan.py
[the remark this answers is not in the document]
ikelos: It's tricky, I think the compromise we've reached within the team is the addition of a has_valid_member method, which allows you to check whether the member is actually accessible or not.  It's not yet landed, but it's in the `has_valid_member` branch for testing.  We try to avoid catching exceptions excessively, particularly for code that will be used by other code rather than just the user, because they may need to know what's gone wrong and why.  Catching an exception says that you're taking responsibility for it in all situations.  In certain cases, that makes a lot of sense, but in vol2 we just indiscriminately said "that's fine, don't worry about it" which for a forensics tool seems a little too easy going...  5:P

I'm not sure we can vet all data in a memory dump, there is a degree where we simply report what's present so in the cases of UniqueProcessId, that's what it said, but if you start looking around it turns out that can't be right.  If we start deciding what we think is right, we'll end up building filters that malware may find ways to abuse (oh, volatility always ignores processes with a vastly higher pid, so we just do that and then we vanish).  As I say, it's a tricky one, but my own view is to show what you've been given and allow the code to figure out if something weird's gone on in the interval unless you're certain it's supposed to have happened...

volatility/framework/plugins/windows/netscan.py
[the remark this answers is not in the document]
ikelos: They're all the same underneath, but it's how they're interpreted.  So NotApplicable would be when a particular field just doesn't make sense in the output (if you're showing handles, and you want to list the registry key, for other types of handle that value will be NotApplicable).  UnreadableValue means "we tried to read it, but couldn't for some reason" usually a pointer failure or something similar.  UnparsableValue is when you've got some data back but it doesn't make any sense or is outside acceptable value ranges...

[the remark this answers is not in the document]
japhlange: I think that UnreadableValue makes sense then - we've tried to read Owner data that *should* exist, but it doesn't make sense.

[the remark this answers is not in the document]
ikelos: Yep, sounds reasonable.  5:)

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,158 @@
...
+            config_path: The configuration path within the context of the symbol table to create
+
+        Returns:
+            The name of the constructed callback table
+        """
+        table_mapping = {"nt_symbols": symbol_table}
+
+        symbol_filename = "netscan"
+
+        return intermed.IntermediateSymbolTable.create(context,
```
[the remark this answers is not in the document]
iMHLv2: AFAIK there aren't any symbols for this, the networking structs are undocumented.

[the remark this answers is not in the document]
ikelos: Awwww, it's just one thing after another!  5:\  Well, looks like it'll be manual route and lots and lots of JSON files then...  5:(

japhlange:
I've added the structure definitions based on https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/overlays/windows/tcpip_vtypes.py, however the implementation of plugin-side selection of the correct file is still pending.

volatility/framework/plugins/windows/netscan.py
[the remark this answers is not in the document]
ikelos: `has_valid_member` has now landed in master, so should be usable.  5:)

japhlange:
reverted premature commit, some structures still need fixing up.

ikelos:
Just so people aren't confused, we know this includes unnecessary commits at the moment.  We're working through it to get to a point where the branch is in order and it's just the necessary patches...  5:)

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,306 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import logging
+import datetime
+from typing import Iterable, List
+
+from volatility.framework import constants, exceptions, interfaces, renderers, symbols, layers
+from volatility.framework.automagic.pdbscan import scan
```
ikelos, line 10: To avoid namespace pollution we try not to import method, but only import modules.  So this would end up being `from volatility.framework.automagic import pdbscan` and then later on `scan` would become `pdbscan.scan`.

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,306 @@
...
+    def get_requirements(cls):
+        return [
+            requirements.TranslationLayerRequirement(name = 'primary',
+                                                     description = 'Memory layer for the kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
+            requirements.PluginRequirement(name = 'poolscanner', plugin = poolscanner.PoolScanner, version = (1, 0, 0)),
+            requirements.BooleanRequirement(name = 'include-corrupt',
+                description = "Radically eases result validation. This will show partially overwritten data. WARNING: the results are likely to include garbage and/or corrupt data. Be cautious!",
+                default = cls.CORRUPT_DEFAULT,
```
ikelos, line 36: This should probably always be False, since otherwise there would likely be no way to turn if off?  (If it defaults to true and providing the flag to true, it's sort of fixed on).

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,306 @@
...
+                optional = True
+            ),
+        ]
+
+    @staticmethod
+    def create_netscan_constraints(context, symbol_table: str) -> List[poolscanner.PoolConstraint]:
```
ikelos, line 42: I realize it's a little long winded, but we try to provide types for all parameters, even context (which would be `interfaces.context.ContextInterface`).  It will then also turn up in the documentation and give people a quick link to the interface when they read up on this method.

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,306 @@
...
+
+    @classmethod
+    def get_requirements(cls):
+        return [
+            requirements.TranslationLayerRequirement(name = 'primary',
+                                                     description = 'Memory layer for the kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
+            requirements.PluginRequirement(name = 'poolscanner', plugin = poolscanner.PoolScanner, version = (1, 0, 0)),
+            requirements.BooleanRequirement(name = 'include-corrupt',
```
japhlange, line 34: For now I've included a flag which shows objects which fulfill the Poolheader requirement but little more. This is because vol2 sometimes found alleged endpoints which only fulfilled the _structural_ requirements without actually having valid _content_. This could actually lead to things like showing what looks like a valid IP address, but which in reality was a pointer being parsed as an IP address.
`--include-corrupt` simulates this behaviour and right now literally only exists because it is possible. I am *not* sure if I should keep this option because I severely doubt it will be useful, but it might be in ony tiny edge-case scenario...

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,306 @@
...
+        """Creates a symbol table for TCP Listeners and TCP/UDP Endpoints.
+
+        Returns:
+            The name of the constructed symbol table
+        """
+        table_mapping = {"nt_symbols": self.config["nt_symbols"]}
+
+        symbol_filename, class_types = self.determine_tcpip_version()
+
+        return intermed.IntermediateSymbolTable.create(self.context,
```
ikelos, line 188: Awesome, this using the table_mapping facility makes me happy!  5:D

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,306 @@
...
+                                       size = (tcpe_size, None),
+                                       page_type = poolscanner.PoolType.NONPAGED | poolscanner.PoolType.FREE),
+            # UDP Endpoint
+            poolscanner.PoolConstraint(b'UdpA',
+                                       type_name = symbol_table + constants.BANG + "_UDP_ENDPOINT",
+                                       size = (udpa_size, None),
+                                       page_type = poolscanner.PoolType.NONPAGED | poolscanner.PoolType.FREE)
+        ]
+
+    def determine_tcpip_version(self) -> str:
```
japhlange, line 77: This method relies on `windows.info` being correct. All other approaches, including
- checking tcpip.sys's PDB GUID
- checking tcpip.sys in memory for its file version PE header
- dumping the tcpip.sys file from memory for the version PE header
have not worked reliably as those artifacts are either insufficient or got corrupted along the way.

volatility/framework/plugins/windows/poolscanner.py
```diff
@@ -85,12 +85,22 @@ def __call__(self, data: bytes, data_offset: int):
                 if constraint.page_type is not None:
                     checks_pass = False
 
+                    # we only care about these types even though Windows supports a lot more pool types.
+                    # see also https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ne-wdm-_pool_type
+                    # and Kernel Pool Exploitation on Windows 7 by Tarjei Mandt (2011))
+                    if header.PoolType > PoolType.FREE | PoolType.NONPAGED | PoolType.PAGED:
```
japhlange, line 91: needs verification because I think this might be wrong. @iMHLv2 @ikelos

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,306 @@
...
+        """ Generates the network objects for use in rendering. """
+
+        netscan_symbol_table = self.create_netscan_symbol_table()
+
+        for netw_obj in self.scan(self.context, self.config['primary'], self.config['nt_symbols'],
+                                  netscan_symbol_table):
+
+            vollog.debug("Found netw obj @ 0x{:2x} of assumed type {}".format(netw_obj.vol.offset, type(netw_obj)))
+            # objects passed pool header constraints. check for additional constraints if strict flag is set.
+            if not show_corrupt_results:
```
ikelos, line 231: It feels like this could be an and statement, but I'm not too worried about it.  5:)

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,306 @@
...
+                filename = "netscan-win81-{arch}".format(arch=arch)
+        else:
+            # default to a fallback, but this *should* not happen.
+            filename = "netscan-win{vers}-{arch}".format(vers=major_version, arch=arch)
+
+        vollog.debug("Determined symbol filename: {}".format(filename))
+
+        return filename, class_types
+
+    def create_netscan_symbol_table(self) -> str:
```
ikelos, line 178: I'd somewhat prefer to see this and `determine_tcpip_version` as `@classmethod`, so that other plugins can call them  It would require parameterizing them anywhere that self is used (so I'd expect a `context`, a `config_path` and a `symbol_table_name` parameter for this one, looks like `determine_tcpip_version` would just need `context`, `layer_name`, `symbol_table_name` parameters, that also keeps all parameters except the context as simple strings).

japhlange:
I've tested the current state of `netscan` with memory dumps of the following systems:

```
vistaSP1x64
vistaSP2x86
win7x86
win7x64
win8SP0x64
win8SP1x86
win8SP1x64
win10x86_10240
win10x64_10586
win10x86_14393
win10x64_14393
win10x86_15063
win10x64_15063
win10x64_16299
win10x64_19041
```
Haven't been able to get my hands on more images, e.g. Win8SP0x86 and Vista without Service Packs.

volatility/framework/plugins/windows/poolscanner.py
```diff
@@ -370,6 +380,7 @@ def generate_pool_scan(cls,
 
             mem_object = header.get_object(type_name = constraint.type_name,
                                            use_top_down = is_windows_8_or_later,
+                                           kernel_symbol_table = symbol_table,
```
ikelos, line 383: Again, a bit of an API change, I'll let @iMHLv2 comment on the efficacy of this.  If he says it's good, it's good.  5:)

volatility/framework/symbols/windows/extensions/network.py
```diff
@@ -0,0 +1,297 @@
...
+from volatility.framework import exceptions
+from volatility.framework.objects import Array
+from volatility.framework.renderers import conversion
+from volatility.framework.symbols.wrappers import Flags
+from volatility.framework import renderers
+from typing import Dict, Tuple
+
+vollog = logging.getLogger(__name__)
+
+def inet_ntop(address_family: int, packed_ip: Array) -> str:
```
ikelos, line 20: So far we've just been using ObjectInterface rather than Array, but Array is more accurate, so I'd say leave it that way, and if it throws up too many errors in the type checker then we'll have to figure something out about it...

volatility/framework/symbols/windows/extensions/network.py
```diff
@@ -0,0 +1,208 @@
...
+# need Microsoft's since that's what is found in memory.
+AF_INET = 2
+AF_INET6 = 0x17
+
+# String representations of INADDR_ANY and INADDR6_ANY
+inaddr_any = inet_ntop(socket.AF_INET, [0] * 4)
+inaddr6_any = inet_ntop(socket.AF_INET6, [0] * 16)
+
+# copied to here as symbol space does not appear to support custom Enums
+TCP_STATE_ENUM = {
```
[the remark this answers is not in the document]
ikelos: Yep, I'd still like it these could be in the JSON files, rather than as code.  There's a whole section for Enums in there, and then they can be accessed based on name, etc.  Send me a message if you need more help details around this (and I might write some documentation about it at the same time, if we're lucky).  5:)

volatility/framework/symbols/windows/extensions/network.py
```diff
@@ -0,0 +1,297 @@
...
+        except exceptions.InvalidAddressException:
+            return None
+
+    def get_create_time(self):
+        dt_obj = conversion.wintime_to_datetime(self.CreateTime.QuadPart)
+
+        if isinstance(dt_obj, interfaces.renderers.BaseAbsentValue):
+            return dt_obj
+
+        # return None if the timestamp seems invalid
```
ikelos, line 165: Hmmmm, probably fine, since futuristic connections shouldn't exist, but limits like this make me nervous.

volatility/framework/symbols/windows/extensions/network.py
```diff
@@ -0,0 +1,297 @@
...
+
+            return self._ipv4_or_ipv6(inaddr)
+
+        except exceptions.InvalidAddressException:
+            return None
+
+    def is_valid(self):
+
+        if self.State not in TCP_STATE_ENUM:
+            vollog.debug("invalid due to invalid tcp state {}".format(self.State))
```
ikelos, line 251: Glad there are debug statements in there!  Good work!  5:)

volatility/framework/plugins/windows/netscan.py
japhlange: that import is unused anyway and was left over from a previous attempt to determine driver version. I removed it ;)

volatility/framework/plugins/windows/netscan.py
japhlange: default is False (view `cls.CORRUPT_DEFAULT`) but it's set to optional=True - I took this as meaning that this flag is optional.

volatility/framework/symbols/windows/extensions/network.py
japhlange: Me too, but when testing some (invalid) objects returned stuff like the year 1600 and I don't want data like that to be recognized as valid.

volatility/framework/plugins/windows/netscan.py
ikelos: It is, I just meant that defining CORRUPT_DEFAULT is a little bit redundant, since setting it to true would essentially force the flag on.

volatility/framework/symbols/windows/extensions/network.py
ikelos: Yeah, that's fair enough.  Shouldn't really be possible to have tcp connections before 1950, and hopefully there'll be a newer version of volatility that people use in 150 years, so happy to mark this as resolved.  5:)  I'd also prefer this were pulled out clearly as a class variable (like `CORRUPT_DEFAULT`) though, just so it's very clear this happens and can be changed/fixed easily if it becomes necessary.  Mebbe I'm being too cautious.  Happy for you to resolve this however you'd like.  5:)

volatility/framework/symbols/windows/extensions/network.py
japhlange: What do you mean by type checker?

volatility/framework/plugins/windows/netscan.py
japhlange: Oh, you meant removing the class variable altogether? Makes sense, I'll do that.

volatility/framework/symbols/windows/extensions/network.py
ikelos: So we try to use mypy against the codebase, to ensure all the type information makes sense (and throw up any bugs of weird types).  We haven't got the type model perfectly specific (mostly because most functions return `ObjectInterface` rather than specific types, even if we *assume* they'll always return an Array, or Struct, or whatever), which is why having such a specific type there might throw up issues, but as I say, we'll figure it out when that happens.  Happy for you to mark this as resolved if you want.  It's more my own thinking...  5:)

japhlange: Ok! I'll resolve this for now, but I'm open to updating the definition if need arises :)

volatility/framework/plugins/windows/netscan.py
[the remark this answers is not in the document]
japhlange: Resolved by using `windows.info` for now.

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,305 @@
...
+        # with several windows versions has showed this to work out correctly.
+
+        is_64bit = symbols.symbol_table_is_64bit(self.context, self.config['nt_symbols'])
+
+        if is_64bit:
+            arch = "x64"
+        else:
+            arch = "x86"
+
+        # the following code is taken from the windows.info plugin.
```
ikelos, line 98: If this is identical code from `windows.info`, it might be worth us pulling that into a classmethod, and then you putting in a `PluginRequirement` and just calling that?

volatility/framework/plugins/windows/netscan.py
japhlange: done!

volatility/framework/plugins/windows/netscan.py
japhlange: yes, this would be great. Currently it's buried inside `windows.info`'s `_generator()`.

volatility/framework/symbols/windows/extensions/network.py
```diff
@@ -0,0 +1,297 @@
...
+            return None
+
+    def get_owner(self):
+        try:
+            return self.member('Owner').dereference()
+
+        except exceptions.InvalidAddressException:
+            return None
+
+    def get_owner_pid(self):
```
ikelos, line 138: Hmmm, undecided on how I feel about bubbling up child attributes in order to put the error checking in a single place.  5:S  It's probably the best way of doing it, but I'm almost wondering if it's worth an adaptor you can sit on top of an object that checks validity and returns None for an attribute if it would otherwise fail?

Something like:
`safe_owner = SafePointerAccess(self.get_owner())`
and then `safe_owner` could do the is_valid check on construction, and also check `has_valid_member`, so that all the other calls could become `safe_owner.pid`, etc?  Any thoughts?  @japhlange @atcuno

volatility/framework/symbols/windows/extensions/network.py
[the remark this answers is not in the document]
japhlange: fixed

ikelos:
Ok, so we're likely to use #269 for splitting out the windows.info methods.  Hopefully it'll be pretty easy to get just the information you need without duplicating code.  I'm happy merging this and then making the change afterwards, so it hasn't become a dependency of this, just something we need to remember to do...  5:)

volatility/framework/symbols/windows/extensions/network.py
japhlange: Putting this into a class variable seems like a good idea... on the off chance that someone has set their system to 1650 (for whatver reason) this would make it easier for the investigator to change that value. But I don't deem it important enough to put it into a flag.

ikelos: Yep, I'd whole heartedly agree.  No need to expose it to the user, but good to pull it out and make it more visible for anyone that goes looking...

volatility/framework/plugins/windows/netscan.py
ikelos: The flag doesn't overly hurt, but I'm also happy for it to be dropped if it really doesn't serve a purpose.

japhlange: Unless someone has a reasonable argument against it I would like to keep it for now. Removing it is easily done if need arises.

ikelos: Yep, that'll be fine.

volatility/framework/symbols/windows/extensions/network.py
japhlange: That would require all attributes wrapped within a `SafePointerAccess` to possess an `is_valid()` method, right? (Or at least to wrap their validation within a function of that name, if existent.)
In the end I think this is a design question. While I agree that a `try/catch` per attribute per object clutters the code and makes it longer it _could_ make it a bit easier to understand on first read. Otherwise you'd need to look into another abstraction layer.
On the other hand this abstraction might be worthwhile if you would need to use several ways to validate the object, e.g. `is_valid()`, `is_valid_member()` and maybe some additional semantic checks as well. If those got wrapped within one adaptor it could actually help readability. To abstract or not to abstract? :)
In this case we would change from
```python
def get_owner_pid(self):
    try:
        if self.get_owner().is_valid():
            return self.get_owner().UniqueProcessId
        else:
            return None
        return None
```
via
```python
def get_owner_pid(self):
    if self.get_owner().is_valid():
        if self.get_owner().has_valid_member("UniqueProcessId"):
            return self.get_owner().UniqueProcessId
    return None
```
to this?
```python
def get_owner_pid(self):
    safe_owner = SafePointerAccess(self.get_owner())
    return safe_owner.UniqueProcessId # returns either int or None
```

japhlange: But then this still begs the question what to do with the returned `None` - at some point in the interface it has to be swapped out with a `renderers.UnreadableValue()`.

ikelos: So I was thinking more along the lines of:
```
    safe_owner = SafePointerAccess(self, on_fail_returns = renderers.UnreadableValue()).get_owner()
```
But you're right, it does need to end up as the right value by the end.  I'm still not convinced None is the right value at the moment.  Recall, it wouldn't need to be in its own separate method, it could just be used where it's needed, so it should reduce a lot of proxied child calls, but I admit, it may hard readability...

volatility/framework/symbols/windows/extensions/network.py
japhlange: updated. please check and resolve if ok :)

volatility/framework/plugins/windows/poolscanner.py
japhlange: resolved because outdated. While I am still unsure about this, I removed it for now because after more testing it doesn't appear necessary.

volatility/framework/plugins/windows/netscan.py
japhlange: updated.

volatility/framework/symbols/windows/extensions/network.py
japhlange: If the adaptor returns `UnreadableValue()` this will also be returned when called by a different plugin, so that plugin would need to catch that value.

volatility/framework/symbols/windows/extensions/pool.py
ikelos: Ok, so a better way of doing this would be make it optional (as `None`) and then check if it's None and if so, make it symbol_table_name.  Also, I think the variable should be either `symbol_table` (or `symbol_table_name`) to be in keeping with other methods.  If it's symbol_table_name, you'll need to watch out for it being reset by the line above.

After that, we can just make an additive bump to the framework version.  In `constants/__init__.py`, bump the `VERSION_MINOR` variable, and then in any plugin that passes in a value to this method, it will need the class variable set as follows: `_required_framework_verison = (1, <whatever-it-was-bumped-to>, 0)`.

volatility/framework/plugins/windows/poolscanner.py
ikelos: See the later comment, but in this plugin it needs the `_required_framework_verison` setting appropriately after that change is made.

volatility/framework/plugins/windows/netscan.py
ikelos: The `windows.info` merge to master has happened, so hopefully you can make the updates now.  5:)

volatility/framework/symbols/windows/extensions/network.py
ikelos: Still thinking about this, but it's not a blocker.

volatility/framework/plugins/windows/poolscanner.py
```diff
@@ -91,6 +91,10 @@ def __call__(self, data: bytes, data_offset: int):
                         checks_pass = True
                     elif (constraint.page_type & PoolType.NONPAGED) and header.PoolType % 2 == 1:
                         checks_pass = True
+                    elif (constraint.page_type
```
ikelos, line 94: @iMHLv2, can you please give this a quick look over?  It's a change to the core so I want to make sure you're happy with it...

ikelos: (Or @superponible, just someone windows-y to ok this as a core change).  5:)

volatility/framework/symbols/windows/extensions/pool.py
japhlange: Okay, please take a look at https://github.com/volatilityfoundation/volatility3/pull/226/commits/7c584a5955e15a6ce6e5d65bf146b140e72ec717 if this fits. 

Also, I think only `poolscanner` uses this method so I just added the variable to that plugin.

ikelos: Yep, looks good, thanks!  5:)

japhlange:
> Also, just to verify, do all these JSON files pass their respective schemas? (That needs the `jsonschema` package installing, and then the tool in `development/schema_validate.py` used with the appropriate schema from `volatility/schemas` to make sure everything works properly. If they all pass, then it's good, if not, they'll need to be fixed up to pass the validation...

Running `schema_validate.py volatility/framework/symbols/windows/netscan*` yields no failures, all files pass :)

volatility/framework/plugins/windows/netscan.py
japhlange: done: https://github.com/volatilityfoundation/volatility3/pull/226/commits/cb5c234b06bc4aec1f5e63ed8fd586323cc98554

volatility/framework/plugins/windows/poolscanner.py
superponible: @ikelos @japhlange @iMHLv2 This doesn't look necessary but there's actually a different problem.  All the builtin constraints check paged, nonpaged, and free, so it wasn't evident. The page type checks in vol2's poolscan.py use the FreePool, NonPagedPool, and PagedPool properties defined in _POOL_HEADER here: https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/overlays/windows/windows.py#L1153 

It looks like vol3's PoolHeaderScanner took those checks directly into the page type checks here: https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/plugins/windows/poolscanner.py#L88

However, vol2 redefines those checks in https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/overlays/windows/vista.py#L42, basically swapping PagedPool and NonPagedPool.

So the vol3 page type checks will need to be modified to use a different check based on the OS version.

ikelos: Hmmm, so I think @superponible is gonna look into this and we may make some core changes that will simplify (but require changes to) this merge request.  Nearly there!  5:)

japhlange: Give me a shout if anything is needed or if I can help in any way.

volatility/framework/plugins/windows/netscan.py
```diff
@@ -0,0 +1,305 @@
...
+                filename = "netscan-win7-{arch}".format(arch=arch)
+            elif nt_minor_version == "2":
+                # 8
+                filename = "netscan-win8-{arch}".format(arch=arch)
+            elif nt_minor_version == "3":
+                # 8.1
+                filename = "netscan-win81-{arch}".format(arch=arch)
+        else:
+            # default to a fallback, but this *should* not happen.
+            filename = "netscan-win{vers}-{arch}".format(vers=major_version, arch=arch)
```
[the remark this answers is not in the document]
japhlange: I pushed a new explicit but redundant method of determining the correct filename which also omits the use of format strings.

ikelos (approved):
Looks good, let's get it in the tree.  5:)

PR #228, opened 2020-06-09
Removal of UnicodeDecodeError from psaux.py

Khanejo:
During testing my sample, I came across UnicodeDecodeError and the processing of next PID'S in my sample halted(as error handling was not present here), thus I propose a small patch that'll allow processing of next PID's even if UnicodeDecodeError is seen. This will ensure complete output to the end user.

### DIFF
--- a/volatility/framework/plugins/mac/psaux.py
+++ b/volatility/framework/plugins/mac/psaux.py
@@ -80,8 +80,11 @@ class Psaux(plugins.PluginInterface):
                     args.append(arg)
 
                 argc = argc - 1
-
-            args_str = " ".join([s.decode("utf-8") for s in args])
+                
+            try:
+                args_str = " ".join([s.decode("utf-8") for s in args])
+            except:
+                pass
 
             yield (0, (task.p_pid, task_name, task.p_argc, args_str))
 


volatility/framework/plugins/mac/psaux.py
```diff
@@ -80,8 +80,11 @@ def _generator(self, tasks: Iterator[Any]) -> Generator[Tuple[int, Tuple[int, st
                     args.append(arg)
 
                 argc = argc - 1
-
-            args_str = " ".join([s.decode("utf-8") for s in args])
+                
+            try:
+                args_str = " ".join([s.decode("utf-8") for s in args])
+            except:
```
ikelos, line 86: This `except` is too general, and should really be `except UnicodeDecodeError` if that's the only exception you're trying to catch.

volatility/framework/plugins/mac/psaux.py
```diff
@@ -80,8 +80,11 @@ def _generator(self, tasks: Iterator[Any]) -> Generator[Tuple[int, Tuple[int, st
...
+                pass
```
ikelos, line 87: Also, just `pass`ing doesn't tell something calling the plugin why something's gone wrong (or even that it has).  At a minimum this should be a `vollog.log` line (fine to set it to the most verbose output), but probably a better solution would be use `s.decode('utf-8', errors='replace')`...

volatility/framework/plugins/mac/psaux.py
atcuno: I believe this needs to get refactored even further as it will be bail out on processing/returning all arguments, even if only one cannot be decoded.

atcuno:
@Khanejo is this something you are still willing to work on? It is definitely an issue that needs to be addressed, so if you are up for making the requested changes then we can give time for it. Otherwise, we can just fix it if you prefer.

volatility/framework/plugins/mac/psaux.py
Khanejo: Respected sir, Actually I tried doing that along with:

               _.decode("utf-8", "replace")_

               _.decode("utf-8", "backslashreplace")_

                _.decode("utf-8", "ignore")_

and stuff like broadening the decoding standard to 'iso-8859-1' ; 'cp1252' ; and some more..

but still, I lost the data on which the UnicodeDecodeError was being generated.(means something gibberish replaced it)

Only when I ran a simple **pass**, then only I recovered the error generating line along with the data after that. I completely agree that there should be an addition of **vollog.log** line.

The exact error is: **UnicodeDecodeError: 'utf-8' codec can't decode byte 0x90 in position 0: invalid start byte**

If you want the link to the memory dump, then I can provide you with that!

ikelos: @Khanejo Thanks for getting back to us!

Returning gibberish is what I'd expect, the data is not encoded in a way that makes sense, so `replace` offers the best option.  Pass has two issues with it:
* it loses data
* if the very first entry fails, then `args_str` won't have been defined and it will throw an exception, but if it has been defined (it's not the first row returned), then you'll be seeing the value for the previous entry (which may be why you think you're losing data without pass?)

My preferred option would still be to use `replace` since, whilst the data will look like gibberish, it will have been interpreted as accurately as possible whilst still displaying something that may be legible data...

Khanejo: Ok! got it, sir. Thanks for your clarification. I completely agree with you. Sir, you may make the desired changes, I too think then the .decode("utf-8", "replace") would be the best option.

ikelos:
I committed the change directly, but thanks for your contribution and pointing out the issue we had!  5:)

PR #229, opened 2020-06-10
Has valid member

ikelos:
Adds in the new `has_valid_member` capability, bumping the API for plugins that might use it.  All plugins should now have the `_required_framework_version` member, which should be set to a tuple of (1,1,0) for those that need to use this feature.

### DIFF
--- a/volatility/framework/interfaces/objects.py
+++ b/volatility/framework/interfaces/objects.py
@@ -176,6 +176,32 @@ class ObjectInterface(metaclass = ABCMeta):
         """
         return False
 
+    def has_valid_member(self, member_name: str) -> bool:
+        """Returns whether the dereferenced type has a valid member."""
+        if self.has_member(member_name):
+            try:
+                _ = getattr(self, member_name)
+                return True
+            finally:
+                return False
+        return False
+
+    def has_valid_members(self, member_names: List[str]) -> bool:
+        """Returns whether the object has all of the members listed in member_names
+
+        Args:
+            member_names: List of names to test as to members with those names validity
+        """
+        # Use only a single try/except for efficiency
+        try:
+            for member_name in member_names:
+                if not self.has_member(member_name):
+                    return False
+                _ = getattr(self, member_name)
+            return True
+        finally:
+            return False
+
     class VolTemplateProxy(metaclass = abc.ABCMeta):
         """A container for proxied methods that the ObjectTemplate of this
         object will call.  This is primarily to keep methods together for easy
--- a/volatility/framework/interfaces/objects.py
+++ b/volatility/framework/interfaces/objects.py
@@ -6,7 +6,6 @@ interpreted values of data from a layer."""
 import abc
 import collections
 import collections.abc
-import contextlib
 import logging
 from abc import ABCMeta, abstractmethod
 from typing import Any, Dict, List, Mapping, Optional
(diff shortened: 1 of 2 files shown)


volatility/framework/interfaces/objects.py
```diff
@@ -176,6 +176,32 @@ def has_member(self, member_name: str) -> bool:
         """
         return False
 
+    def has_valid_member(self, member_name: str) -> bool:
+        """Returns whether the dereferenced type has a valid member."""
+        if self.has_member(member_name):
+            try:
+                _ = getattr(self, member_name)
+                return True
+            finally:
```
atcuno, line 185: Putting a return False in a finally block means that this part of the function will always return False, even if the return True is hit in the try block. I tested and this is definitely the case, even for members that are in the structure. This link explains why: https://stackoverflow.com/a/19805746

volatility/framework/interfaces/objects.py
```diff
@@ -176,6 +176,32 @@ def has_member(self, member_name: str) -> bool:
...
+        """
+        # Use only a single try/except for efficiency
+        try:
+            for member_name in member_names:
+                if not self.has_member(member_name):
+                    return False
+                _ = getattr(self, member_name)
+            return True
+        finally:
+            return False
```
atcuno, line 203: same here about finally blocks

volatility/framework/interfaces/objects.py
```diff
@@ -176,6 +176,32 @@ def has_member(self, member_name: str) -> bool:
...
+    def has_valid_members(self, member_names: List[str]) -> bool:
+        """Returns whether the object has all of the members listed in member_names
+
+        Args:
+            member_names: List of names to test as to members with those names validity
+        """
+        # Use only a single try/except for efficiency
+        try:
+            for member_name in member_names:
+                if not self.has_member(member_name):
```
atcuno, line 198: can we just call has_valid_member for each element in the loop vs repeat code?

volatility/framework/interfaces/objects.py
atcuno: @ikelos  tagging you here as I made 3 comments on this and want to make sure you see them. I am not sure if GitHub sends emails if you aren't tagged.

ikelos: I'm the assignee, so I got the emails, but also glad to be tagged so I spot them.  5:)

Interesting, I guess it makes sense but I hadn't realized the python flow control worked that way.  I find a way to avoid `except Exception` using `contextlib.suppress`, so I'll add that to this branch.

volatility/framework/interfaces/objects.py
ikelos: As above

volatility/framework/interfaces/objects.py
ikelos: The reason for not doing it was to avoid constructing loads of try/except blocks, rather than just one.  It can actually stack up if this is going to be called often, which it will, so for efficiency reasons I'm going to leave it that way.  There's not a lot of code duplicated and the speed savings could end up being significant...

ikelos:
Otherwise all good?

volatility/framework/interfaces/objects.py
ikelos: Hmmm, having moved to a context manager, and the fact that these methods may be reimplemented by higher up classes, we'll give the looped call a go and see how it works out...  5:)

ikelos: Ok, may have to give that a rethink.  Setting up the try/catch is minimal, but catching an exception is 8x slower than just letting it through.  Turns out, the contextlib is always at least 25x slower, so I'm gonna go back to the exception route, "too broad" exceptions be damned...  5:P

ikelos: For reference, adding a try/except clause (that doesn't fire) adds about 6 nanoseconds on my i7-8656U system.  So millions of those probably still won't add up to a very significant slowdown.

Issue #200, opened 2020-03-05
MutantScan issues 


SuperbSonik:
**Describe the bug**
unclear information

**Context**
Volatility Version: 3 beta
Operating System: Win10x64
Python Version:  3.8
Suspected Operating System:  Win10x64
Command:  
![image](https://user-images.githubusercontent.com/61529488/75943170-39d93980-5ecf-11ea-94c2-f09b92debdc8.png)

**Expected behavior**
This is the result obtained from Volatility 2
![image](https://user-images.githubusercontent.com/61529488/75943198-4c537300-5ecf-11ea-9fae-077c894ed82f.png)

**Screenshots**
Currently, I am using the mutantscan command but it shows something like this:
![image](https://user-images.githubusercontent.com/61529488/75943327-a7856580-5ecf-11ea-8f2f-03e0b4c5ba3e.png)

![image](https://user-images.githubusercontent.com/61529488/75943302-9d636700-5ecf-11ea-91cd-8e7277b60f87.png)

Maybe it is my problem.. Feel free to leave any comment, thank you all.

ikelos:
Hiya, this has hopefully been fixed in the recent merge of #247.  Please verify this, and if you're still encoutering issues feel free to reopen this report.  5:)

PR #230, opened 2020-06-10
refs #200 raise AttributeError if an object header doesn't have a nam…

ikelos:
…e struct

if we stick with this approach, several other plugins that reference NameInfo.Name (directly or downstream) will need to start catching AttributeError as well


iMHLv2:
Closing this in favor of https://github.com/volatilityfoundation/volatility3/pull/247

PR #237, opened 2020-06-25
print correct db/.. offset: increment offset after printing, not before

dl9rdz:
Currently, the db(..) commands in volshell will print a wrong offset:

(primary) >>> db(0xffff81a000e0)
0xffff81a000f0    4c 69 6e 75 78 20 76 65 72 73 69 6f 6e 20 34 2e    Linux.version.4.
0xffff81a00100    34 2e 30 2d 31 32 34 2d 67 65 6e 65 72 69 63 20    4.0-124-generic.
(the data is actually from offset ...e0, but the offset shown in the left column is off by 0x10)

This pull request fixed this issue by incrementing offset only after it was printed, not before.


ikelos:
Yep, good spot, thanks for the fix!  5:)

PR #238, opened 2020-06-28
Add process listing through examination of the process group hash table

atcuno:


volatility/framework/plugins/mac/pslist_process_groups.py
```diff
@@ -0,0 +1,89 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
```
ikelos, line 1: I think we decided that when new files were added they need to have the right year in the header?  Existing files don't need to be updated though...  5:S

volatility/framework/plugins/mac/pslist_process_groups.py
```diff
@@ -0,0 +1,89 @@
...
+
+        kernel_as = context.layers[layer_name]
+
+        table_size = kernel.object_from_symbol(symbol_name = "pgrphash") 
+
+        pgrphashtbl = kernel.object_from_symbol(symbol_name = "pgrphashtbl")  
+
+        proc_array = kernel.object(object_type = "array",
+                                   offset = pgrphashtbl, 
+                                   count = table_size + 1,
```
ikelos, line 45: Why is the table_size off by one?  Is it a zero-index thing?  I don't doubt it, I just like to know how +1s make their way into the code...  5;)

volatility/framework/plugins/mac/pslist_process_groups.py
```diff
@@ -0,0 +1,89 @@
...
+        proc_array = kernel.object(object_type = "array",
+                                   offset = pgrphashtbl, 
+                                   count = table_size + 1,  
+                                   subtype = kernel.get_type("pgrphashhead"))
+
+        for proc_list in proc_array:
+            # test the validity of the current element
+            # it is expected that many won't be initialized
+            try:
+                pgrp = proc_list.lh_first
```
ikelos, line 52: This needs to be defined beforehand, or in the exception handler it needs to be set to a default. The next time it's accessed it may not exist due to the exception handler having fired.

volatility/framework/plugins/mac/pslist_process_groups.py
```diff
@@ -0,0 +1,89 @@
...
+                seen_pgrps.add(pgrp.vol.offset)
+            
+                # nothing can be done if this list pointer is invalid, so move on
+                try:
+                    p = pgrp.pg_members.lh_first
+                except exceptions.PagedInvalidAddressException:
+                    break
+
+                seen_pg = set()
+                while p and p.vol.offset not in seen_pg:
```
ikelos, line 69: This is about 3 levels of loops.  It looks like they're all necessary, but just watch for inefficiencies in the following code...

volatility/framework/plugins/mac/pslist_process_groups.py
```diff
@@ -0,0 +1,89 @@
...
+import logging
+from typing import Callable, Dict, Iterable
+
+from volatility.framework import constants, interfaces, objects, contexts, exceptions
+from volatility.plugins.mac import pslist
+
+vollog = logging.getLogger(__name__)
+
+class Pslist_Process_Groups(pslist.PsList):
+    """Lists the processes present in a particular mac memory image."""
```
ikelos, line 14: This is already used in two separate existing plugins, it has to be made more accurate (across all three ideally), otherwise there's no way of knowing how this differs from plain `pslist` differs from `tasks`.

volatility/framework/plugins/mac/pslist_process_groups.py
```diff
@@ -0,0 +1,89 @@
...
+                        p = p.p_pglist.le_next 
+                    except exceptions.PagedInvalidAddressException:
+                        break
+                try:
+                    pgrp = pgrp.pg_hash.le_next
+                except exceptions.PagedInvalidAddressException:
+                    break
+
+
+
```
ikelos, line 85: 5:S  I really don't want to see these (pslist, pslist_process_groups, tasks) mount up, and I'd almost prefer them as different classmethods on pslist, with parameters/requirements/flags to choose between them.  Inheritance doesn't easily allow for suitable dependency verification if a breaking, but not compilation breaking, change occurs...

Happy to hear reasons for keeping them separate plugins, but I can't see why flags wouldn't work out with a single plugin instead?

volatility/framework/plugins/mac/pslist_process_groups.py
atcuno: included in the new commit.

volatility/framework/plugins/mac/pslist_process_groups.py
atcuno: yes, + 1 because in the kernel its used as a bit mask to the hash table index so we would miss the last element entry otherwise

volatility/framework/plugins/mac/pslist_process_groups.py
atcuno: the except handler does a continue on the loop, so I don't see how pgrp could be accessed without ever being defined/valid?

volatility/framework/plugins/mac/pslist_process_groups.py
atcuno: yea, its a nested hash table of lists so we don't really have a choice :/

volatility/framework/plugins/mac/pslist_process_groups.py
atcuno: included in the new commit.

volatility/framework/plugins/mac/pslist_process_groups.py
ikelos: Ah, you're right.  It feels like bad practice, but I won't hold it up over this...  5:S

volatility/framework/plugins/mac/pslist_process_groups.py
atcuno: A single plugin would likely lead to people just assuming we only have support for the one method since in vol2 we have separate process enumeration plugins in Windows, Mac, and Linux. Also, the alternate methods would not appear when people used the help option to see all the plugins, instead all they would see is pslist even though it really would have numerous ways to enumerate process. The list that pslist walks can be broken by malware without negatively affecting the system whereas the other sources will have the processes due to instability if changed.

ikelos: So, `vol.py --help` would show the pslist plugin `vol.py pslist --help` would show the various options for the plugin, which would list the various means of enumerating the values.
The fact that vol2 did something a particular way doesn't imply that it's necessarily a good way of doing it and shouldn't bind us to do things that way forever after.  I'm not sure how great the user confusion would be if we document it well (the plugin docstring could mention that it contains different ways of finding these values) given we've already got other plugins that work differently to vol2, but I think developers would find it beneficial to have different methods of the same thing grouped in the same place?
Given that they all return the same data, that no one of them works properly, but one stands out as the most likely (but vulnerable to malware) means of getting the data, I'd suggest that be the default, and then a choice option to pick another method would work pretty well?  It also saves us coming up with weird names (why is `tasks` not `pslist_tasks` for example)...

None of those reasons feel solid enough to justify not having a single plugin.  I'm still open to the concept, but also still leaning towards a single plugin as the way forward?  A single plugin would mean that any single change would require the version number to be bumped, but I don't think that's unreasonable?  If we ever needed to split them out in the future, we could do so by bumping the plugin version number.  Given that all the others inherit from the single original plugin, they all (currently) share a version number, so there's no big change there...  We could even move the class methods to non-plugin classes with their own versions which pslist could check before using them, but that feels like it might complicate things unnecessarily.  Is there something else I'm not seeing?  If it's a strongly held conviction that they should be separate I'll consider that, but overall I'm not seeing the downside?

atcuno: ok, we can try to push them all into one and see how it plays out over time. Would you made merging this one into pslist.py as you described and then I can do the others?

ikelos: Sure, I'll try and mock it up tonight.  I'll push the results directly to this branch so we can follow what's going on...  5:)

ikelos:
Hmmm, turned out to be a little more involved, but I'm kinda happy with the results.  At the moment all the requirements and things work as expected, and we can easily swap the pslist method for any plugin that uses it (at the moment, we always default to tasks, which is what each of the plugins originally used).  The only downside would be that allowing users to choose their pslist method for any particular plugin *could* _technically_ break with an exception (since accessing the method list necessarily has to happen before the plugin requirement has been processed).  If we're willing to accept that cost, then all plugins except netstat can be parameterized with a pslist_method choice.  Netstat would need a plugin version bump because it would change the API for determining processes.  Shout if there's any questions or comments or anything, in fact, the changes are pretty big, so perhaps I'll turn this into a review by you, @atcuno, please!  5;)

volatility/framework/plugins/mac/psaux.py
```diff
@@ -86,11 +86,12 @@ def _generator(self, tasks: Iterator[Any]) -> Generator[Tuple[int, Tuple[int, st
             yield (0, (task.p_pid, task_name, task.p_argc, args_str))
 
     def run(self) -> renderers.TreeGrid:
-        filter_func = tasks.Tasks.create_pid_filter([self.config.get('pid', None)])
+        filter_func = pslist.PsList.create_pid_filter([self.config.get('pid', None)])
+        list_tasks = pslist.PsList.get_list_tasks(self.config.get('pslist_method', pslist.PsList.pslist_methods[0]))
```
atcuno, line 90: can pslist.PsList.pslist_methods[0] just be 'tasks' live above?

Also, why does this one do config.get('pslist_method', ...), but then the _generator does  config.get('method',...) ?

volatility/framework/plugins/mac/bash.py
```diff
@@ -84,24 +83,23 @@ def _generator(self, tasks):
...
-                                                             filter_func = filter_func)))
+                                      list_tasks(self.context,
+                                                 self.config['primary'],
+                                                 self.config['darwin'],
+                                                 filter_func = filter_func)))
 
     def generate_timeline(self):
-        filter_func = tasks.Tasks.create_pid_filter([self.config.get('pid', None)])
+        filter_func = pslist.PsList.create_pid_filter([self.config.get('pid', None)])
+        list_tasks = pslist.PsList.get_list_tasks(self.config.get('pslist_method', pslist.PsList.pslist_methods[0]))
```
atcuno, line 99: can pslist.PsList.pslist_methods[0] just be 'tasks' live above?

Also, why does this one do config.get('pslist_method', ...), but then the _generator does config.get('method',...) ?

volatility/framework/plugins/mac/pslist.py
```diff
@@ -85,13 +121,120 @@ def list_tasks(cls,
             else:
                 seen[proc.vol.offset] = 1
 
-            if not filter_func(proc) and kernel_as.is_valid(proc.vol.offset, proc.vol.size):
+            if not filter_func(proc) and kernel_layer.is_valid(proc.vol.offset, proc.vol.size):
                 yield proc
 
             try:
                 proc = proc.p_list.le_next.dereference()
             except exceptions.InvalidAddressException:
```
atcuno, line 129: should this be PagedInvalidAddressException or InvalidAddressException? I notice this method uses InvalidAddressException whereas all the other enumeration methods use PagedInvalidAddressException. I assume this is how the plugins were before and you just moved the code over, but it seems like now is a good time to make consistent.

volatility/framework/plugins/mac/pslist.py
```diff
@@ -85,13 +121,120 @@ def list_tasks(cls,
             else:
                 seen[proc.vol.offset] = 1
 
-            if not filter_func(proc) and kernel_as.is_valid(proc.vol.offset, proc.vol.size):
+            if not filter_func(proc) and kernel_layer.is_valid(proc.vol.offset, proc.vol.size):
```
atcuno, line 124: I also noticed some inconsistency in these checks. 

1) I think kernel_layer.is_valid call should switch to is_readable() like we discussed recently and is done in one method.

2) It also seems like is_readable() should be called before filter_func as filter_func accesses the pid member, which would cause an exception if the process is at a bad pointer (which would be caught by is_readable if called first)

atcuno:
I left a couple comments. Looking pretty nice so far though

volatility/framework/plugins/mac/pslist.py
ikelos: So, if we only catch PagedInvalidAddressException, we don't get faults that might happen in lower layers.  If we do InvalidAddressException we catch any time an address is invalid?  Depends what you're after, but I agree consistency would be good, so once you've chosen I can make the changes throughout.

volatility/framework/plugins/mac/bash.py
ikelos: Well, making it the [0] element allows us to pick later by just having it in one place.  If we hard code it we'd have to change the order (the first should be the default) *and* we'd need to change the places it was hard coded.  Seems better to use the first element?

Not entirely sure about the pslist_method vs just method.  I think it might've been the first one, and I later decided it should say which method it's talking about.  Feel free to correct it, or leave this unresolved and I'll fix it when I next get a chance to code...  5:)

volatility/framework/plugins/mac/pslist.py
ikelos: Sure, happy with both of those, again feel free to change it or leave it unresolved and I'll fix it when I get a chance.  5:)

ikelos: What we have at the moment is `is_valid` on layers, and `is_readable` *only* on pointers.  The `is_readable` method essentially operates on the the target of the object.  We currently don't have an `is_valid` method on objects, because simple objects read all their data on instantiation, but structs do not.  We could add a check that all their data is present before allowing a Struct object to be built, but we don't intentionally, so that if it crosses a memory boundary but we don't need the part of it that isn't accessible, we can still get to the data that's there.  

The general guidance is that we don't check the object wholesale and only use it if it's all there, we check what we need (by catching exceptions).  It essentially ensures we support and promote the ability to process partial structs.  In this instance, I'm ok checking it here because the consumer expects the methods to yield valid procs.  If we wanted to check it, we'd need to update the filter function to pass over those with invalid `pids`, but we'd also still have to check on member accesses later on.  That would become very inconvenient for future consumers of the methods, so I'm ok checking the entire thing as they're being listed.  However, I don't want to make checking of the entirety of an object easier, so I would not add an is_valid method to normal objects.

I have changed the order of the checks though.  5;)

volatility/framework/plugins/mac/pslist.py
ikelos: Yep, unless there's an explicit reason to know which page or that it happened in the virtual layer, rather than a lower layer, we should be catching InvalidAddressException.  I've fixed that up (across the codebase) in here.

ikelos:
Ok, looks like all comments have been addressed, so I'll merge it.  Any issues/changes should be made as a new pull request (or directly to master if they're urgent/critical/etc)...  5:)

PR #245, opened 2020-07-01
Configurable stacker list

ikelos:
This will significantly speed up mac and linux plugins by not running the Wintel stacker uselessly against such plugins.  It also allows CLI users to choose which stackers to use, should they so desire, and allows generic plugins to still try all the stacking options.

This is built on top of #244, which should be reviewed/get committed first...

### DIFF
--- a/volatility/cli/__init__.py
+++ b/volatility/cli/__init__.py
@@ -271,7 +271,7 @@ class CommandLine(interfaces.plugins.FileConsumerInterface):
 
         # It should be up to the UI to determine which automagics to run, so this is before BACK TO THE FRAMEWORK
         automagics = automagic.choose_automagic(automagics, plugin)
-        ctx.config['automagic.LayerStacker.stackers'] = stacker.choose_stackers(plugin)
+        ctx.config['automagic.LayerStacker.stackers'] = stacker.choose_os_stackers(plugin)
         self.output_dir = args.output_dir
 
         ###
--- a/volatility/cli/__init__.py
+++ b/volatility/cli/__init__.py
@@ -26,6 +26,7 @@ import volatility.symbols
 from volatility import framework
 from volatility.cli import text_renderer, volargparse
 from volatility.framework import automagic, constants, contexts, exceptions, interfaces, plugins, configuration
+from volatility.framework.automagic import stacker
 from volatility.framework.configuration import requirements
 
 # Make sure we log everything
@@ -270,6 +271,7 @@ class CommandLine(interfaces.plugins.FileConsumerInterface):
 
         # It should be up to the UI to determine which automagics to run, so this is before BACK TO THE FRAMEWORK
         automagics = automagic.choose_automagic(automagics, plugin)
+        ctx.config['automagic.LayerStacker.stackers'] = stacker.choose_stackers(plugin)
         self.output_dir = args.output_dir
 
         ###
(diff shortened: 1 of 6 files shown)


ikelos:
Ok, this now uses an exclusion list for stackers to say they're unnecessary for certain oses (technically first-level hierarchies of plugins).  All other plugins will be included by default (ie, reverts to original behaviour).

npetroni:
The overall structure seems fine here. I'm not sure if it makes sense to do exclude versus include, since new OSes will require more places to be updated, but I can live with it. It's better than the previous approach, in my opinion.

This branch fails for me when running `windows.pslist.PsList` on `xp-laptop-2005-06-25.dmp`. It works fine on master.

ikelos:
Awww, I didn't expect to have broken something, I'll check it out.  The reason for exclusion rather than inclusion is that in the case of new OSes (or new categories of plugin) it would require adding to all file format stackers (likely to be the biggest proportion ofstackers), whereas with exclusion, it's just os specific plugins that need changing and there's only ever likely to be around up to 5 or so of those.

Thanks for looking it over!  I'll investigate the breakage before merging.  5:)

ikelos:
Hmmm, I've just tried it (using both windows.zip and generating the JSON from the internet) and `windows.pslist.PsList` works in both on my `xp-laptop-2005-06-25.dmp` (md5sum: f24e72d09406371d202a7e780385d03d).  Could you double check the issue and if it's still there send me the `-vvvv` output please?

npetroni:
I did a fresh checkout. Works on master, but not on the branch for the XP and win10 images that I have.

```
./vol.py  -vvvv -f win10-x86-2016-07-08.lime windows.pslist.PsList > out.txt 2>&1
```

[out.txt](https://github.com/volatilityfoundation/volatility3/files/4865472/out.txt)

ikelos:
So mine finds a DTB at 0x39000 where yours goes into the "no DTB found in standard locations".  Even without pefile that happens, so the missing dep isn't the issue.  5:\  I'm thoroughly stumped?  Here's the working output on my side from a fresh checkout, change branch, venv using setup.py for requirements and then run?  I need to look into it further because it seems to still be trying to stack on Linux/Mac later on, but I've really no idea what causes it to go wrong.  Would you be able to git bisect it and narrow it down to the first bad commit please (should only be a couple of runs since there's only 3 commits in this?)

[out-working.txt](https://github.com/volatilityfoundation/volatility3/files/4865620/out-working.txt)

ikelos:
Ok, I think I got it.  My guess is your image isn't a raw image, but is compressed or similar.  There were two bugs, one was that we were selecting the wrong part of the plugin hierarchy (so `volatility` rather than `windows` from `volatility.plugins.windows.pslist.PsList`) and also, the layer plugins hadn't been loaded by the point we chose the stackers.  Since several of the stackers are file formats, and thus live in the layers directory, they weren't being included.  Mine succeeded because Intel was the only required layer, but if yours was lime or gz or anything, it would have failed.

This should have been fixed in `e89de9a`.  Please let me know if that resolves it, and either way, thanks for helping spot a nasty bug!  5:D

ikelos:
The first bug highlights that fail open where no stackers are excluded is also a better option (although harder to spot when things go wrong I guess?).  5;)

npetroni:
Yes, works for both images now. Thanks!

ikelos:
No problem, thanks for the speedy reviews!  5:)

PR #248, opened 2020-07-03
Yarascan: Move most of yarascanning into a versionable plugin

ikelos:
This refactors common yara tasks, so we can use the plugin versioning to
keep track of changes to the YaraScanner class.

Should be straightforward, just putting it up for review.  If there's no quibbles I'll merge it in a few days, only review if you feel like it.  5:)

### DIFF
--- a/volatility/framework/configuration/requirements.py
+++ b/volatility/framework/configuration/requirements.py
@@ -380,19 +380,19 @@ class SymbolTableRequirement(interfaces.configuration.ConstructableRequirementIn
         return context.symbol_space[value].build_configuration()
 
 
-class PluginRequirement(interfaces.configuration.RequirementInterface):
+class VersionRequirement(interfaces.configuration.RequirementInterface):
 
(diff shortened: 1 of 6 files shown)


ikelos:
Ok, I've refactored this and created a `VersionRequirement` which can be used to require any `VersionableInterface` inherited object.  Both plugins and scanners inherit from `VersionableInterface`, but any other classtype can be added as we see fit and it *should* be introducable later if we discover we need it on a particular type of class (I'm looking at you LayerInterface classes).  5:P

For now, there's a backwards-compatible `PluginRequirement`, but happy to swap that over in all the core plugins (potentially leave it in place with a deprecation warning for any external plugins already using it?).  Sorry it's all mushed in with the YaraScan changes, but hopefully they're all quite small and straight-forward.  Hopefully this is new review-worthy, @npetroni.   think it is a better way of doing things, thanks for letting me talk myself out of the other method...  5:)

PR #249, opened 2020-07-04
Automagic: Heavily refactor pdbscan

ikelos:
This separates out a new PDBUtility class that can take a (contiguous) layer and MZ offset, and produce identify, download if necessary and load the appropriate symbol table (if any).

This is designed to allow plugin authors to locate their module/dll/exe in memory, and then very quickly load the required symbols for it.  The core kernel-finding code has been adapted to use this also.

The mz-to-pdb-guid method requires pefile (which was already a requirement of volatility, although apparently no longer used othre than in one of the core plugins, verinfo).  This has specifically been added as an optional dependency (ie, the core doesn't require it, but plugins might), but if we think it would easy/possible/not too cumbersome to use internal code rather than pefile, I'm all for it.  I'd prefer not to just scan for the RSDS record once we know roughly where it is, just for precision, hence using the PE info.  The PDBUtility class could be moved out of the automagic hierarchy if we think it would live better somewhere else?

I would strongly like eyes on this, and I'd like to catch all the little "but what abouts", no matter how nit-picky they may be, so please critique away.  This is in the critical path and likely to be a major component of future plugins (hopefully) so I wanna make sure it's absolutely rock solid (and I'm fairly sure this is a bit below release quality as is).  This won't get merged until it's had at least two out of three reviews, no time limit...


volatility/framework/automagic/pdbscan.py
```diff
@@ -75,54 +76,232 @@ def __call__(self, data: bytes, data_offset: int) -> Generator[Tuple[str, Any, b
...
+        filter_string = os.path.join(pdb_name.strip('\x00'), guid.upper() + "-" + str(age))
+
+        isf_path = False
+        # Take the first result of search for the intermediate file
+        for value in intermed.IntermediateSymbolTable.file_symbol_url("windows", filter_string):
+            isf_path = value
+            break
+        else:
+            # If none are found, attempt to download the pdb, convert it and try again
+            PDBUtility.download_pdb_isf(context, guid.upper(), age, pdb_name, progress_callback)
```
npetroni, line 130: Is there a reason this is called via the absolute name `PDBUtility` rather than the `cls` variable?

ikelos: There is not, good spot!  5;)

ikelos:
Hmmm, this seems like it wasn't moved to where I was expecting to have put it.  I think the PDBUtility class should likely live under `framework/symbols/windows/pdb.py` or something?

npetroni:
> Hmmm, this seems like it wasn't moved to where I was expecting to have put it. I think the PDBUtility class should likely live under `framework/symbols/windows/pdb.py` or something?

Yeah, I was going to suggest moving it. That seems like a better place.

volatility/framework/symbols/windows/pdb.py
```diff
@@ -0,0 +1,288 @@
...
+            layer_name: The name of the (contiguous) layer within the context that contains the MZ file
+            offset: The offset in the layer at which the MZ file begins
+            symbol_table_class: The type of symbol class to construct the SymbolTable
+            config_path: New path for the produced symbol table configuration with the config tree
+            progress_callback: Callable called to update ongoing progress
+
+        Returns:
+            None if no pdb information can be determined, else returned the name of the loaded symbols for the MZ
+        """
+        guid, age, pdb_name = cls.get_guid_from_mz(context, layer_name, offset)
```
npetroni, line 47: `get_guid_from_mz` can return `None`, so I think this is not safe. It could cause an exception in framework code.

volatility/framework/symbols/windows/pdb.py
```diff
@@ -0,0 +1,288 @@
...
+        if config_path is None:
+            config_path = interfaces.configuration.path_join('pdbutility', pdb_name.strip('\x00').replace('.', '_'))
+
+        return cls.load_windows_symbol_table(context, guid, age, pdb_name, symbol_table_class, config_path,
```
npetroni, line 51: It's a little awkward that clients of `get_guid_from_mz` need to strip off the trailing `\x00` from the name. Should this be pushed down into the function?

volatility/framework/symbols/windows/pdb.py
ikelos: Good call, should now be fixed.  5:)

volatility/framework/symbols/windows/pdb.py
ikelos: Yep, I've pushed in down into the function.

ikelos:
So this is still pending at least one of a couple of reviews, and could do with a trivial example of pulling down some symbols for a userland program.

volatility/framework/symbols/windows/pdb.py
```diff
@@ -0,0 +1,291 @@
...
+                if not ctx.layers[layer_name].is_valid(i * page_size, 2):
+                    break
+
+                data = ctx.layers[layer_name].read(i * page_size, 2)
+                if data == b'MZ':
+                    mz_offset = i * page_size
+                    break
+            min_pfn = sig_pfn
+
+            yield {
```
atcuno, line 245: @ikelos should this check if mz_offset == None before yielding? Otherwise it seems like it would yield with a None value for it.

volatility/framework/symbols/windows/pdb.py
```diff
@@ -0,0 +1,291 @@
...
+
+            for i in range(sig_pfn, min_pfn, -1):
+                if not ctx.layers[layer_name].is_valid(i * page_size, 2):
+                    break
+
+                data = ctx.layers[layer_name].read(i * page_size, 2)
+                if data == b'MZ':
+                    mz_offset = i * page_size
+                    break
+            min_pfn = sig_pfn
```
atcuno, line 243: @ikelos it seems like the way min_pfn and sig_pfn are handled would prevent multiple signatures on the same page from being processed/checked. This would occur as line 243 would set min_pfn to sig_pfn and then if the next loop iteration generated the same sig_pfn (if another signature was on the same page), then the loop on 235 would never iterate as sig_pfn and min_pfn would be the same. Was the intention to prevent multiple signatures on the sam page? If so, what led to that decision?

volatility/framework/symbols/windows/pdb.py
atcuno: Nevermind, I see where a None mz_offset is specifically checked in method_fixed_mapping

volatility/framework/automagic/pdbscan.py
```diff
@@ -310,10 +146,12 @@ def method_fixed_mapping(self,
         physical_layer_name = self.get_physical_layer_name(context, vlayer)
         kvo_path = interfaces.configuration.path_join(vlayer.config_path, 'kernel_virtual_offset')
 
-        kernels = scan(ctx = context,
-                       layer_name = physical_layer_name,
-                       page_size = vlayer.page_size,
-                       progress_callback = progress_callback)
+        kernel_pdb_names = [bytes(name + ".pdb", "utf-8") for name in constants.windows.KERNEL_MODULE_NAMES]
```
atcuno, line 149: The three discovery methods seems they could use a little cleanup to be more understandable and less repeated code:

1) method_module_offset & method_kdbg_offset seem to be copy/paste of each other except for the string searched and how much is subtracted from the address of the string found to the containing data structure. An internal function that took those values as parameters would collapse the the current methods to one line each

2) method_fixed_mapping:
 - The check of "or not isinstance(kernel['mz_offset'], int)" should be removed as mz_offset can only be None or the result of multiplying two integers.

- this method sets kernel_virtual_offset itself whereas the other two let set_kernel_virtual_offset set it. From reading the code, it seems like the same value will be eventually set, so the duplication be should be removed.

volatility/framework/automagic/pdbscan.py
```diff
@@ -410,17 +248,19 @@ def check_kernel_offset(self,
         # Scan a few megs of the virtual space at the location to see if they're potential kernels
 
         valid_kernels = {}  # type: ValidKernelsType
```
atcuno, line 250: @ikelos the name "valid_kernels" throughout the method_ function names and variable is confusing as each method only will return one kernel and breaks their loops when a valid kernel is found. If we can change the variable names and function names to singular that would make initial reading much easier to grasp.

volatility/framework/automagic/pdbscan.py
ikelos: 1. Cool, I'll give it a look
2. 
* The check may be there to help mypy (which otherwise complains the value may not be an int because it's not clever enough to figure that out), but I'll check the git blame to see why it's there and add a comment if it's mypy related, or remove it if not.
* Sure, I can look into that code and improve it.

volatility/framework/symbols/windows/pdb.py
ikelos: This is literally a copy/paste from what we had previously (which had been in the code base for four years, see commit https://github.com/volatilityfoundation/volatility3/commit/d6039fb899c3e2447a5902ebcaa52569523ef313), so I can look into these, but reviewing old code somewhat makes it difficult to tell whether the changes were acceptable or not.  5:S  Looks like the code originally came from @npetroni 's pdbscanner, so I'll let him answer that question, if that's ok?

Given a valid MZ header is unlikely to have multiple signatures in a page (since loaded MZ headers are always mapped to the start of a page), I'm not sure only returning a single signature is problematic?  If we're looking for MZs that have just been arbitrarily loaded into memory then this function may not be suitable, we can add something to the docstring if that will help?

atcuno: Re: Old code.. I wasn't too familar with this file before looking at this PR so I had to go beyond the code changed to really understand what affect(s) might be caused by it. That is where those comments came from.

Re: MZ headers.. that makes sense! Thanks!

volatility/framework/automagic/pdbscan.py
ikelos: Sure, although I think originally it was supposed to return multiple valid kernels.  It feels like this review was significantly over existing code?  I take it that the changes made by the patch were generally acceptable?

You're right, it's very strange.  I think the original code looked for multiple potential offsets, but given this only allows one per layer name, it doesn't seem all that useful.  I'm wondering whether the methods should return more, perhaps by a generator?  Given the dictionary only ever gets one entry, I've converted it a tuple.  We'd need to rework if if we did ever want the automagic to return multiple options (and we'd have to figure out how to order the methods, whether that be one attempt per method, or all options from each method)...

volatility/framework/automagic/pdbscan.py
ikelos: Yeah, 2a was a typing thing, so I'm going to leave it in.  The static type checker can only figure out types like that with something like an explicit `isinstance`.  Eventually, when it gets smarter, we can take the redundant check out.

volatility/framework/automagic/pdbscan.py
atcuno: Yes, sorry again about reviewing code not part of the diff, but before today I had never reviewed this file so I needed to read/understand all the functions called to understand what the diff was doing. All the other code in the two pdb* files looks really nice

ikelos:
Ok, we've had two reviews (with changes that have been made), so I'll merge this tomorrow unless someone speaks up...

Issue #250, opened 2020-07-11
Documentation: How to pass configuration when running plugins.


NiklasBeierl:
**Is your feature request related to a problem? Please describe.**
I am trying to implement a custom plugin and now I want to pass some configuration to it, preferably through the cli.
I added: `requirements.IntRequirement(name="myparamname", description="some description")` to the requirements. 
I read:  https://volatility3.readthedocs.io/en/latest/simple-plugin.html but its not very exhaustive and has no information on how to "run" ones custom plugin. I was also unable to pass `-p <some-pid>` to linux.pslist.PsList as in previous versions of volatility.

**Describe the solution you'd like**
Some documentation as to how configuration can be passed to a custom plugin. (via the cli) Adding it to the article mentioned above might be a good Idea, if I get an answer, I might be able to submit a PR some time next week.

**Describe alternatives you've considered**
- Google wasn't helping me out.
- I looked at `volatility/cli/__init__.py` it seems like in line `215`, an argsparser is added for plugin configuration. Haven't investigted further,  tho. 
-  `--myparamname` was not recognized by the cli.

ikelos:
Hiya, we've got a man page/documentation for the volatility CLI in the works, but I'm afraid it's not finished yet.  The configuration can be extended by providing a configuration JSON (using global option `-c` or `--config`), you can set a single configuration option using the extend option (global option `-e` or `--extend`) or using the CLI which supports passing plugin specific parameters after the plugin name.

You should find that `vol.py --help` will give you all the global options you can use (and list the available plugins) and that `vol.py <pluginname> --help` will show you the options that you can use for that specific plugin (the order is important, you must list the plugin name before the plugin options).  

`--pid` wasn't listed as a requirement for `linux.pslist`, which was an oversight and is now corrected in commit `07ddbb6`.  Please note that plugin options are not automatically converted into single character options due to the difficulty of support this with dynamic options from plugins, so each parameter name must be explicitly listed (such as `--pid` rather than `-p`).

I'm not sure why `--myparamname` wasn't supported, you may be able to find more information in the debug output (by running `vol.py -vvv`).  I'll leave this bug open until we've added the man page/CLI documentation, but if you'd like to debug the issue further feel free to add more details here and we'll try and help...  5:)

NiklasBeierl:
Thanks for the speedy reply!
Getting help on a plugin this way is a very good hint: #251 
[...]

PR #251, opened 2020-07-11
Added instructions for help on a plugin

NiklasBeierl:
See #250.


ikelos:
Thanks for the request.  I'll add it so you get the credit, but probably reformat and reword it once it's in place...

PR #253, opened 2020-07-13
yarascan: Added Yara rule name to return value

shu-tom:
The return value of yarascan will contain the hit strings, but the Yara rule name is also important.

Volatility2 returned the Yara rule name.
https://github.com/botherder/volatility/blob/5eb15741a095c5f0dec9bd771f6c4a63b06137c1/volatility/plugins/malware/malfind.py#L114

Please add Yara rule name to yarascan return to maintain compatibility with Volatility2.

### DIFF
--- a/volatility/framework/plugins/yarascan.py
+++ b/volatility/framework/plugins/yarascan.py
@@ -27,10 +27,11 @@ class YaraScanner(interfaces.layers.ScannerInterface):
         super().__init__()
         self._rules = rules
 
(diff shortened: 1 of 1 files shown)


ikelos:
Thanks very much for the patch!  I'm afraid it changes the output of the YaraScanner's scans, which will break both yarascan plugins without additional changes.  However, I've made the necessary changes to YaraScan (including letting other plugins know that it's output has changed) so this should now be in master.  I hope that's ok?

shu-tom:
OK, thanks!

PR #255, opened 2020-07-14
Created Linux check modules

mtressler:
Added extension for kobject required for check_modules to run


atcuno:
@ikelos The commit looks a little messy due to the requirement for line-end swtiching, but if you look here:

https://github.com/volatilityfoundation/volatility3/pull/255/files

You can see the condensed/final view. I have looked this over already and seen the output, so I am good with it.

ikelos:
Cool, thanks, I'll give it a look and assuming it's all good we probably merge it by Thursday...  5:)

volatility/framework/plugins/linux/check_modules.py
```diff
@@ -0,0 +1,87 @@
+# Volatility
+
+# This file is Copyright 2020 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+"""
+@author:       Matt Tressler
```
ikelos, line 8: Hiya, we'd prefer not to have individual contact information in the files, it can lead to people being territorial over the code.  Hopefully the git log will be sufficient, but please let us know if that's not ok?

volatility/framework/plugins/linux/check_modules.py
```diff
@@ -0,0 +1,87 @@
+# Volatility
```
ikelos, line 1: We don't typically include these two lines in core, we just launch straight into the copyright, but it's not a show stopper.  5:)

volatility/framework/plugins/linux/check_modules.py
```diff
@@ -0,0 +1,87 @@
...
+                                                     description='Memory layer for the kernel',
+                                                     architectures=["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(
+                name="vmlinux", description="Linux kernel symbols"),
+
+            requirements.PluginRequirement(
+                name='lsmod', plugin=lsmod.Lsmod, version=(1, 0, 0))
+        ]
+
+    def get_kset_modules(self, vmlinux):
```
ikelos, line 43: If this would be useful for other plugins, it may be worth exposing it as a `classmethod`?  It will likely need to accept a context, and then individual values for things like `symbol_table` rather than `self.config['vmlinux']` and then hand those in through the generator.

volatility/framework/plugins/linux/check_modules.py
```diff
@@ -0,0 +1,87 @@
...
+
+        try:
+            module_kset = vmlinux.object_from_symbol("module_kset")
+        except exceptions.SymbolError:
+            vollog.error("this command is not supported by this profile")
```
ikelos, line 48: Given the check directly after this, it and the return seem a little redundant?  I'd probably just set `module_kset = None` and then let the check below fire (and perhaps raise an Exception, mebbe a RunTimeError?).

volatility/framework/plugins/linux/check_modules.py
```diff
@@ -0,0 +1,87 @@
...
+
+        ret = {}
+
+        kobj_off = self.context.symbol_space.get_type(
+            self.config['vmlinux'] + constants.BANG + 'module_kobject').relative_child_offset('kobj')
+
+        for kobj in module_kset.list.to_list(vmlinux.name + constants.BANG + "kobject", "entry"):
+
+            mod_kobj = vmlinux.object(
+                object_type="module_kobject", offset=kobj.vol.offset - kobj_off)
```
ikelos, line 62: Looks like yapf hasn't been run against this, which is fine, but be aware the formatting may change if someone does in the future...

volatility/framework/symbols/linux/extensions/__init__.py
```diff
@@ -388,3 +388,15 @@ def get_mnt_mountpoint(self):
 
     def get_mnt_root(self):
         return self.mnt_root
+
+
+class kobject(objects.StructType):
```
ikelos, line 393: A few more documentation strings would be handy, just to describe the object and the attached method?  If this whole object is only really usable for that plugin, it might be worth adding the class override to the symbol_table (as an example to others as to how to do it).  I'm in two minds, because this will end up being a core plugin, so again, whichever you think is better.  Not a showstopper...  5:)

volatility/framework/plugins/linux/check_modules.py
atcuno: This won't be used by other plugins unless Linux adds some new subsystem for tracking kernel modules, which is very unlikely seeing that they haven't in 10+ years. If they ever do, we could then break it out as that would make for a nice linux_lsmodxview type plugin, but I am not hopeful for that.

volatility/framework/symbols/linux/extensions/__init__.py
atcuno: What do you mean by class override to the symbol table? Is there an example of that elsewhere ?

volatility/framework/plugins/linux/check_modules.py
ikelos: Cool, no problem then.

volatility/framework/symbols/linux/extensions/__init__.py
ikelos: Sure, so here we construct a new table with a `class_types` argument, that contains the class types that are to be overridden: https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/plugins/windows/procdump.py#L70

The idea would be that if it's not useful for other plugins but critical for this one, this one adds the override when it needs it (by constructing its own table).  Again, it's a design decision, and either will work, just wanted to make sure it'd been considered...

atcuno: The existing table comes from the kernel itself and depending on how future development goes, it may be needed by other plugins. I would prefer to keep it as-is.

ikelos: Ok, if it's gonna be reused, that makes sense.  5:)

PR #256, opened 2020-07-14
Created linux_check_creds plugin

mtressler:


atcuno:
I am good with this one too @ikelos

volatility/framework/plugins/linux/check_creds.py
```diff
@@ -0,0 +1,72 @@
+# Volatility
+
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+"""
+@author:       Matt Tressler
```
ikelos, line 8: See similar comments in #255

volatility/framework/plugins/linux/check_creds.py
```diff
@@ -0,0 +1,72 @@
...
+        ]
+
+    def _generator(self):
+        vmlinux = contexts.Module(
+            self.context, self.config['vmlinux'], self.config['primary'], 0)
+
+        type_task = self.context.symbol_space.get_type(self.config['vmlinux'] + constants.BANG +"task_struct")
+
+        if not type_task.has_member("cred"):
+            vollog.error("this command is not supported by this profile")
```
ikelos, line 47: Not sure whether error stop execution or not, I might be tempted to raise here, but up to you.  5:)

volatility/framework/plugins/linux/check_creds.py
```diff
@@ -0,0 +1,72 @@
...
+        for (_, pids) in creds.items():
+            if len(pids) > 1:
+                pid_str = ""
+                for pid in pids:
+                    pid_str = pid_str + "{0:d}, ".format(pid)
+                pid_str = pid_str[:-2]
+                yield(0, [str(pid_str)])
+
+    def run(self):
+        return renderers.TreeGrid([("PIDs", str)], self._generator())
```
ikelos, line 72: Feels like we could put more useful output here?  Just a list of PIDs doesn't relate much to anything.  Is there any information from the creds structure we could output along side these (even if just an offset?)  Not sure if that's useful if this is just to list pids that share creds, but possibly something that could help people figure out which pids share which creds?

atcuno: We could list the creds address if that makes it more consistent, but it is not really something I would expect people to pivot into volsell for. The idea of listing the pids is so people can then use pslist and other process inspection plugins to dig into the suspicious/malicious one. The legitimate kernel will never share cred structures between processes, but rootkits often will get a reference to the cred structure for pid 1 / init, and then set the cred pointer of the process they want to elevate privileges for to the cred struct of init.

If you search linux_check_creds on this blogpost then you can see how it fits into investigations:

https://volatility-labs.blogspot.com/2012/09/movp-14-average-coder-rootkit-bash.html

volatility/framework/plugins/linux/check_creds.py
atcuno: It does seem like error doesn't exit... What would be the exception that should be thrown here? And should we change profile to symbol table? or OS version? or is leaving 'profile' okay?

ikelos: Errr, I'd prefer it were "symbol table` (and the sentence capitalized), but I didn't want to nitpick!  5:P  5:)  The exception I'd throw is that the structure's not valid so probably a TypeError (since we were expecting an object that had a "cred" member, and this doesn't)...

volatility/framework/plugins/linux/check_creds.py
ikelos: Ok, so essentially the only things listed in here are malicious?  Just seems like we probably know something more useful than just the pid at this point, but if not then so be it.  Happy for you to mark this resolved if you're happy that it is...  5:)

atcuno: yea, really the only point of this plugin is to say 'these pids are sharing a cred structure'... which is a system state that the legitimate kernel would never trigger, and is a favorite trick of kernel rootkits to elevate the privileges of an attacker's shell process to root.

volatility/framework/plugins/linux/check_creds.py
atcuno: ok cool, will get this fixed and standardized. This won't be the only plugin to need this code path / exception thrown, so wanted it settled.

mtressler: Do we want the type error to specifically say what the issue is? Like explicitly say it's the cred member we are missing or just a standard "This command is not supported by this symbol table"

volatility/framework/plugins/linux/check_creds.py
ikelos: So the bit that concerns me with this is that you say `sharing *a* cred structure` rather than `sharing *the* cred structure`.  If there's more than one, wouldn't it be useful to know which pids share which cred structure?  If there's only one cred structure to be shared, this makes perfect sense to me, otherwise I'd think it would make sense to differentiate them?  Happy if not, but wanted to make sure it was said...  5:)

volatility/framework/plugins/linux/check_creds.py
ikelos: So grepping through the code for `TypeError` should give other examples.  `This command is not ...` doesn't help all that much, so something that tells someone what's gone wrong.  Something like "This version of the operating systems doesn't have the necessary structures for this plugin`, or even `Task requires a cred member` would both be fine.  Ideally errors are unique so we can easily trace back where they're thrown if they're reported to us...

mtressler: Makes sense. Just wanted to clarify before I made the changes.

ikelos:
Cool, looks good, thanks for making the changes!  5:)

Issue #202, opened 2020-03-10
Says PageInvalidAddressException


mbhatt1:
**BUG**
Invalid Page Address

**Context**
Volatility Version:  3.0.0 beta 1
Operating System:  10.15.3
Python Version:  3.6/3.7.5
Suspected Operating System:  10.15.2
Command:  

**To Reproduce**
>python3.6 vol.py -vvv -f ~/memory/data.lime mac.netstat.Netstat
```
DEBUG    root        : Traceback (most recent call last):
  File "/Users/mbhatt/volatility3/volatility/cli/__init__.py", line 281, in run
    renderers[args.renderer]().render(constructed.run())
  File "/Users/mbhatt/volatility3/volatility/cli/text_renderer.py", line 163, in render
    grid.populate(visitor, outfd)
  File "/Users/mbhatt/volatility3/volatility/framework/renderers/__init__.py", line 196, in populate
    for (level, item) in self._generator:
  File "/Users/mbhatt/volatility3/volatility/framework/plugins/mac/netstat.py", line 80, in _generator
    filter_func = filter_func):
  File "/Users/mbhatt/volatility3/volatility/framework/plugins/mac/netstat.py", line 58, in list_sockets
    for filp, _, _ in mac.MacUtilities.files_descriptors_for_process(darwin_symbols, context, task):
  File "/Users/mbhatt/volatility3/volatility/framework/automagic/mac.py", line 271, in files_descriptors_for_process
    path = vnode.full_path()
  File "/Users/mbhatt/volatility3/volatility/framework/symbols/mac/extensions/__init__.py", line 137, in full_path
    self._do_calc_path(elements, self, self.v_name)
  File "/Users/mbhatt/volatility3/volatility/framework/symbols/mac/extensions/__init__.py", line 128, in _do_calc_path
    self._do_calc_path(ret, vnodeobj.v_parent, vnodeobj.v_parent.v_name)
  File "/Users/mbhatt/volatility3/volatility/framework/objects/__init__.py", line 313, in __getattr__
    return getattr(self.dereference(), attr)
  File "/Users/mbhatt/volatility3/volatility/framework/objects/__init__.py", line 681, in __getattr__
    native_layer_name = self.vol.native_layer_name))
  File "/Users/mbhatt/volatility3/volatility/framework/objects/templates.py", line 72, in __call__
    return self.vol.object_class(context = context, object_info = object_info, **arguments)
[...]

PR #258, opened 2020-07-15
Issue #202 - Check for invalid vnode pointers before processing path

atcuno:
This addresses issues #202


atcuno:
@mbhatt1 we are planning to merge this soon unless you report that the fix does not address the issue in your sample

PR #261, opened 2020-07-16
CLI: Support hiding data that may be inaccurate

ikelos:
Ok, this is the concept that we can hide/filter data that we know to be unparsable/unavailable (turned off by default).  This is very much an example and I don't expect this to be merged as is.

The lines currently aren't hidden (because it would require more complex code, and a decision to deal with a tree node's children).  It also doesn't operate on JSON or CSV renderers.  This is more a proof-of-concept to show how many rows would be removed (and whether the plugins are appropriately tagging fields, and if it is, whether hiding those lines is usefu).  Not likely to be merged anytime soon, but here for people to play with and see what they think...

### DIFF
--- a/volatility/cli/__init__.py
+++ b/volatility/cli/__init__.py
@@ -157,6 +157,10 @@ class CommandLine(interfaces.plugins.FileConsumerInterface):
                             help = "Clears out all short-term cached items",
                             default = False,
                             action = 'store_true')
+        parser.add_argument("--safe",
+                            help = "Does not display data that is unreadable or unparsable",
+                            default = False,
+                            action = 'store_true')
 
         # We have to filter out help, otherwise parse_known_args will trigger the help message before having
         # processed the plugin choice or had the plugin subparser added.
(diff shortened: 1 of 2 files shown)


japhlange:
This plays nicely when checking if at least one return value is Unreadable or Unparsable. However plugins might define their own sanity checks specific to the data they are handling, e.g. when handling partially corrupt data. A plugin returning several values of which one is Unreadable() but still deem the object as a whole as valid and would want to return it. 

How could this situation be handled best? Should plugins not do this or is there a clean way to tweak the --safe flag?

ikelos:
So plugins should return all the data, and leave decisions on display up to the UI.  If the `--safe` option doesn't hide the right things then we shouldn't include it.  Happy to figure different ways to allow plugins to note priorities (probably through format hints as a mechanism) if we think that's necessary, but definitely don't want plugins doing their own filtering...  5:P

ikelos:
If a plugin has a sanity check that says data's definitely wrong, then we can add either a format type hint, or another BaseAbsentValue type.  There's a way of achieving the goal, it's just putting in the right machinery to allow it to happen the right way...  5:)

PR #263, opened 2020-07-19
Reduce strictness of Linux Symbol Packs

deeso:
Added some logic in the Automagic Linux Class to search for "Linux version" strings.  If no matches for specific Linux Symbol packs are found, then the Automagics will look at the Major/Minor Linux Kernel Version (e.g. Linux version 5.4.0) instead of the entire banner.  There is no accounting for the Distro, but this could be added later on.

### DIFF
--- a/volatility/framework/automagic/linux.py
+++ b/volatility/framework/automagic/linux.py
@@ -38,13 +38,23 @@ class LinuxIntelStacker(interfaces.automagic.StackerLayerInterface):
         if not linux_banners:
             vollog.info("No Linux banners found - if this is a linux plugin, please check your symbol files location")
             return None
-
-        mss = scanners.MultiStringScanner([x for x in linux_banners if x is not None])
+        
+        # search for specific banners and the Linux version of the banners
+        search_banners = [x for x in linux_banners if x is not None]
+        search_banners = search_banners + list(set([x.split(b'-')[0] for x in search_banners if x.find(b'Linux version') == 0]))
+        
+        mss = scanners.MultiStringScanner([x for x in search_banners if x is not None])
+        # capture all banners for second pass if no matches found
+        found_banners = set()
         for _, banner in layer.scan(context = context, scanner = mss, progress_callback = progress_callback):
             dtb = None
             vollog.debug("Identified banner: {}".format(repr(banner)))
-
+            found_banners.add(banner)
             symbol_files = linux_banners.get(banner, None)
+            if symbol_files is None:
+                # banners may be versions, but we want to check all specific banners first
+                continue
+
             if symbol_files:
                 isf_path = symbol_files[0]
                 table_name = context.symbol_space.free_table_name('LintelStacker')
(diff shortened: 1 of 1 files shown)


ikelos:
Thanks for the patch, we've definitely had people wanting to be able to use linux analysis without having the exact JSON.

I'm quite concerned that allowing "fuzzy" symbols tables will lead to a huge number of "this doesn't work" issues being raised, and plugin authors no longer being able to rely on symbols being accurate.  As such, it needs a big warning (ie, shows up even without any `-v` flags) at an absolute minimum that this is now best guest and that issues arising from it are probably caused by the mismatch.  As much so that we can identify that the wrong table was used when we get given broken output, as the users.

I've also asked @npetroni and @atcuno to weigh in on their thoughts on how well/accurate this will be (and how subtle the bugs that it might introduce) would be.  Since structures change based the config options that are passed in, even two identical kernel versions may have different structures with the same name, the question is how often changes like that affect structures that people are likely to use (not just us, but also anyone else wanting to write any other plugin)...

The patch also appears to duplicate a bunch of code which should be refactored out, but that's easily doable.  5:)

deeso:
@ikelos I agree with the specificity vs. generality problem here, and its hard to strike that delicate balance between "stock" kernels and "custom" kernels user's compile for their own needs.  In this case, I am trying to bias towards distribution stock kernels where the symbols may not change significantly (if at all) between Linux versions.  That in and of itself might be an interesting study for another time (e.g. how often do Linux kernel symbols change from version to version (distro to distro) and whats the impact on memory analysis), but I digress.  

I think adding a strong warning when this is feature is used is a good idea, and an even better idea might be adding a config variable/parameter to turn this "expert" feature on or off.  I call it expert in this case, because the analyst will really need to understand the implications of the analysis without the exact symbols, and in some cases, they might need to explain this decision and weigh the accuracy of the result.

ikelos:
Yeah, I could go for an expert flag combined with the warning.  I'd also like to start building a library of stock kernels, so that when an update happens, we can grab the latest necessary files and generate the JSON information for everybody.  To that end, we've kind of made a start with a number of stock kernels (which I don't think is yet in out linux.zip), so we should be able to greatly reduce the chance of hitting a kernel you can't get the correct JSON for...

atcuno:
In order to properly analyze a memory sample, Volatility needs two pieces of information:

1) The type definitions. 

2) The address and name of a large number of kernel symbols - generally of global variables that reference artifacts like the head of the process list, head of the kernel module list, etc. These are stored in System.map upon compile and are also encoded in the debug (DWARF) information of the produced vmlinux file.

While you can have shared types between kernel versions, or in some cases the subset of types that Volatility uses are the same or  'close enough' that analysis still works... The symbol addresses will change on every kernel recompile, making sharing symbol information impossible, and requiring getting the specific symbol addresses for every kernel that you want to analyze.

This is why having a Volatility 2 profile or Volatility 3 symbol table json from a kernel that is extremely close, but not the same as the one being analyzed, will fail. 

Volatility 2 lacks the ability to auto detect the correct Linux version and then validate that a given --profile matches it. This has led to literally hundreds of GitHub tickets, emails, tweets, etc. with users who think they have the right profile, but do not, and given Volatility's lack of auto detection, it has no way to report the discrepancy. This is a very poor user experience. 

With Volatility 3, we made a conscience effort to ensure that Linux samples would only be processed if the given profile (symbol table in Vol 3 terms) matched the analyzed kernel exactly. If we were to loosen this restriction then we would be recreating the Volatility 2 problems all over again.

With all that said, you can do something like you are proposing wit the existing dwarf2json tool that we provide to create the symbol table json files (https://github.com/volatilityfoundation/dwarf2json/). In particular, you would want to gather the System.map that matches the analyzed kernel *exactly* - not one character can be different in the version string or analysis will break - and then you can take the types from a kernel version that is very close to the one being analyzed, and hope they are close enough to still work. By taking the System.map from the kernel being analyzed you ensure you that the symbol addresses will definitely be the same, and when taking the types, you hopefully are from a build (ideally from the same distro), that uses a very similar .config so that they data structures won't significantly change.

deeso:
@atcuno - what you described (decisions/trade-offs and likelihood of failed
analysis) makes sense.  With dwarf2json, how would I merge the symbols from
a "similar kernel" with the System Map of the specific kernel.
Conceptually, the task makes sense, but I am struggling understanding how
the two JSON files might need to be merged.  For example, I have not been
able to determine where in the symbol table the kernel banner is located.
If this task is something that needs to be scripted/documented, I don't
mind taking the time to work through it.    I just need a little guidance
to help speed the process along.

deeso:
@atcuno and @ikelos : I created a separate project that accomplishes the fuzzy linux logic: https://github.com/deeso/vol3-fuzzy-linux-layer.  I understand the concerns and why the project is avoiding this approach, so I figure that this side project might be more appropriate instead of introducing potential future issues into the main project.

pagabuc:
Hey @deeso just a quick heads up since I worked a bit on these problems. 
To extract the System.map from a memory dump you have a few options that should work:
1) https://github.com/pagabuc/kallsyms-extractor 
2) https://github.com/marin-m/vmlinux-to-elf

But as @atcuno and @ikelos suggested, the configuration used to build the kernel really changes the layout of several structures used for memory forensics (for example [task_struct](https://elixir.bootlin.com/linux/latest/source/include/linux/sched.h#L632) contains around 80 ifdefs) !
So definitely give it a shot if you want to try, but I wouldn't be too surprised if a profile for a specific kernel only partially works for a "similar" one.

deeso:
@pagabuc thanks for the pointers.  I'll check these out.  I appreciate the
guidance.

PR #266, opened 2020-07-23
Refactor nix utilities

ikelos:
Refactor the Utilities classes as discussed at the most recent call.Refactor the Utilities classes as discussed at the most recent call.

### DIFF
--- a/volatility/framework/automagic/linux.py
+++ b/volatility/framework/automagic/linux.py
@@ -3,13 +3,13 @@
 #
 
 import logging
-from typing import List, Optional, Tuple, Type
+from typing import Optional, Tuple, Type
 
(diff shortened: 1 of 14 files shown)


ikelos:
@atcuno Are you happy with this?  I don't want to hold up Matt's plugin any longer than necessary, but I would like this to go in first?  It's pretty much just a move of where the code lives at we discussed, so shouldn't be a big review, just want to make sure you're happy with what got moved (and to where) and what didn't...

atcuno:
This looks good. I used the branch and ran lsof which calls into the utility called and it worked as expected.

ikelos:
Cool, thanks for checking it over.  5:)

PR #267, opened 2020-07-23
Objects: Add a convenience function for validating enum values

ikelos:
Very quick one, just wanting views on the name for it (and whether it should be a property or a method, whose name would then start with a verb)?

I think `is_valid` is too generic and is already used in other contexts for readability of data, so I went with legal.  Also considered `valid_choice` as a property, but again valid without context isn't ideal.  5:S

Opinions and views welcomed, otherwise this'll go in in 7 days...  5:)

### DIFF
--- a/volatility/framework/objects/__init__.py
+++ b/volatility/framework/objects/__init__.py
@@ -466,6 +466,11 @@ class Enumeration(interfaces.objects.ObjectInterface, int):
     def choices(self) -> Dict[str, int]:
         return self._vol['choices']
 
+    @property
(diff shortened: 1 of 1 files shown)


npetroni:
I'd vote for `is_choice` or `is_valid_choice` to be parallel to `choices`. I don't think `legal` is the best name. We use the `is_` prefix for many predicates in the library, so I think we should stick with that pattern.

ikelos:
Awesome, thanks for the advice!  5:)

PR #268, opened 2020-07-23
added hashdump plugin

jxwenger:


volatility/framework/plugins/windows/hashdump.py
```diff
@@ -0,0 +1,346 @@
...
+    def get_requirements(cls):
+        return [requirements.TranslationLayerRequirement(name = 'primary',
+                                                     description = 'Memory layer for the kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols",
+                                                description = "Windows kernel symbols"),
+            requirements.PluginRequirement(name = 'hivelist', plugin = hivelist.HiveList, version = (1, 0, 0))
+            ]
+
+    odd_parity = [
```
ikelos, line 25: The entries in this table feel like they should be calculable?  Do you know how were they generated?  Not a show stopper but big chunks of hard coded data don't sit so well in the middle code...  5:S

volatility/framework/plugins/windows/hashdump.py
```diff
@@ -0,0 +1,346 @@
...
+    anum = b"0123456789012345678901234567890123456789\0"
+    antpassword = b"NTPASSWORD\0"
+    almpassword = b"LMPASSWORD\0"
+    lmkey = b"KGS!@#$%"
+
+    empty_lm = "aad3b435b51404eeaad3b435b51404ee"
+    empty_nt = "31d6cfe0d16ae931b73c59d7e0c089c0"
+
+    #empty treegrid object to return in case of error
+    none_obj = (0,('','','',''))
```
ikelos, line 59: These should be values derived from `BaseEmptyValue`, see https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/renderers/__init__.py#L21 for the four defined possibilities.  This is important so that the UI can determine how best to display the information being returned.

volatility/framework/plugins/windows/hashdump.py
```diff
@@ -0,0 +1,346 @@
...
+        root = samhive.root_cell_offset
+        if not root:
+            return None
+
+        sam_account_key = self.open_key(samhive.get_node(root), sam_account_path)
+        if not sam_account_key:
+            return None
+        
+
+        F = None
```
ikelos, line 133: We try to use descriptive variable names, otherwise such things get extremely difficult to follow down the line, even if this is literally called `F_value` something a bit more than just `F` that helps people know what it is would be useful...

volatility/framework/plugins/windows/hashdump.py
```diff
@@ -0,0 +1,346 @@
...
+            enc_nt_hash = V[nt_offset+4:nt_offset+20] if nt_exists else ""
+            nthash = self.decrypt_single_hash(rid, hbootkey, enc_nt_hash, self.antpassword)
+        elif nt_revision == b'\x02':
+            nt_exists = True if nt_len == 56 else False
+            nt_salt = V[nt_offset+8:nt_offset+24] if nt_exists else ""
+            enc_nt_hash = V[nt_offset+24:nt_offset+56] if nt_exists else ""
+            nthash = self.decrypt_single_salted_hash(rid, hbootkey, enc_nt_hash, self.antpassword, nt_salt)
+        return lmhash, nthash
+
+    def sid_to_key(self, sid):
```
ikelos, line 219: This definitely needs a docstring describing what it returns, and again `s1` and `s2` give no useful information about the contents what-so-ever, so these need renaming to be more useful.

volatility/framework/plugins/windows/hashdump.py
```diff
@@ -0,0 +1,346 @@
...
+        s1 += chr((sid >> 16) & 0xFF)
+        s1 += chr((sid >> 24) & 0xFF)
+        s1 += s1[0]
+        s1 += s1[1]
+        s1 += s1[2]
+        s2 = s1[3] + s1[0] + s1[1] + s1[2]
+        s2 += s2[0] + s2[1] + s2[2]
+        return self.str_to_key(s1), self.str_to_key(s2)
+
+    def str_to_key(self, s):
```
ikelos, line 233: Again, documentation to help figure out what all this is supposed to be do (is it part of an encryption algorithm, is it just rotating bytes?  Since our code is often referred to by people to find out how to do things themselves, it's quite important to us that the code is clear and well documentation so that those reading it can understand it.

volatility/framework/plugins/windows/hashdump.py
```diff
@@ -0,0 +1,346 @@
...
+                        ntout=nthash
+                    else:
+                        ntout = binascii.hexlify(nthash).decode('latin1')
+                    yield (0,   (name.decode('latin1'), 
+                                str(int(str(user.get_name()), 16)),
+                                lmout,
+                                ntout))
+        else:
+            print("Hbootkey is not valid")
+            yield self.none_obj
```
ikelos, line 326: This will end the generator, it seems better to throw an (informative) exception or error here.

Also, please use the logging framework rather than directly printing anything.  This is designed to be a library and the console output may never get seen if it's wired into a webpage or similar...

volatility/framework/plugins/windows/hashdump.py
```diff
@@ -0,0 +1,346 @@
...
+                        lmhash = self.empty_lm
+                    if not nthash:
+                        nthash = self.empty_nt
+                        ## temporary fix to prevent UnicodeDecodeError backtraces 
+                        ## however this can cause truncated user names as a result
+                    name = self.get_user_name(user, samhive)
+                    if name is not None:
+                        name = name.encode('ascii', 'ignore')
+                    else:
+                        name = "(unavailable)"
```
ikelos, line 308: This should be a `BaseAbsentValue` derivative so that other plugins which may make use of the output can easily determine that the value couldn't be found, and why it couldn't be returned.

volatility/framework/plugins/windows/hashdump.py
```diff
@@ -0,0 +1,346 @@
...
+    lmkey = b"KGS!@#$%"
+
+    empty_lm = "aad3b435b51404eeaad3b435b51404ee"
+    empty_nt = "31d6cfe0d16ae931b73c59d7e0c089c0"
+
+    #empty treegrid object to return in case of error
+    none_obj = (0,('','','',''))
+
+    #helper method to find a particular key in a hive
+    def open_key(self, root, key):
```
ikelos, line 62: If feels as though this operates similarly to `hive.get_key()`?  If this does something different, and it's a common task, then we can include it in the core registry code, but I think the only difference is that this looks to be case insensitive?

volatility/framework/plugins/windows/hashdump.py
```diff
@@ -0,0 +1,343 @@
...
+        str1 += chr((sid >> 16) & 0xFF)
+        str1 += chr((sid >> 24) & 0xFF)
+        str1 += str1[0]
+        str1 += str1[1]
+        str1 += str1[2]
+        str2 = str1[3] + str1[0] + str1[1] + str1[2]
+        str2 += str2[0] + str2[1] + str2[2]
+        return cls.str_to_key(str1), cls.str_to_key(str2)
+
+    #Build final DES key from the strings generated in sid_to_key
```
ikelos, line 227: This would be better as a docstring, since that'll then get automatically added to the documentation we generate (https://volatility3.readthedocs.io/en/latest/volatility.plugins.windows.html#submodules)

So in this case it would be:
```
@classmethod
def str_to_key(cls, s):
    """Builds final DES key from the strings generated in sid_to_ley"""
```

PR #273, opened 2020-07-29
New file interface

ikelos:
Ok, so this is a fairly deep-rooted change, but hopefully all works smoothly...

The concept is that when a plugin wants to create a file, it'll construct a `self.FileHandler` rather than a `interfaces.plugins.FileInterface`.  If it's a previously written plugin, it'll still use the old BytesIO interface.

The UI can then specify what a FileHandler should be (this is on a per-plugin basis, rather than a per-context basis, because plugins that run sub-plugins may find it very useful to choose to act as the file handler and/or consumer themselves).

The CLI has been updated to provide a file-backed handler (and all existing core plugins have been updated, the non-core plugins haven't, but that can easily be done).  Since plugins should be inheritting from the interface, they'll get the new `set_file_handler` method for free, and thus don't depend on a later version of the framework.  The CLI does make calls to this method though, and so requires that the framework be the increased version.

So @iMHLv2 could you please check this works ok and solves your memory issues and @npetroni could you please check it's nt too ugly and/or I haven't done something stupid?

### DIFF
--- a/doc/source/complex-plugin.rst
+++ b/doc/source/complex-plugin.rst
@@ -29,7 +29,7 @@ available plugins that feature a Timeliner interface).  This can be achieved wit
 
     automagics = automagic.choose_automagic(automagic.available(self._context), plugin_class)
     plugin = plugins.construct_plugin(self.context, automagics, plugin_class, self.config_path,
-                                self._progress_callback, self._file_consumer)
+                                self._progress_callback, self._file_consumer, self.FileHandler)
 
 This code will first generate suitable automagics for running against the context.  Unfortunately this must be re-run for
 each plugin in order to populate the context's configuration correctly based on the plugin's requirements (which may vary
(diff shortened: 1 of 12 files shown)


ikelos:
So, some thoughts on this.  FileHandler feels a big ugly, since they really represents Files, kinda.  5:S  Happy for someone to suggest a better name.  I've also added some documentation which makes me think we maybe need to define whether to allow path separators in the filename, or do some kind of sanitization?  I think the best for general UIs is to only allow filenames without paths.  On the other hands, plugins may want to output in a hierarchical form (and the WebUI could filter the path names when consuming the files?).  For the moment, I've not touched any sanitzation code, but it might be worth considering at some point...

doc/source/complex-plugin.rst
```diff
@@ -41,11 +41,30 @@ between plugins).  Once the automagics have been constructed, the plugin can be
...
+Writing out files
+-----------------
+The :py:class:`~volatility.framework.interfaces.plugins.PluginInterface` contains a FileHandler attribute, that should be
+used as a means of writing out files by passing them to the `consume_file` method.  When producing a file, construct an object
+handler using the :py:obj:`~volatility.framework.interfaces.plugins.PluginInterface.FileHandler` class.
+
+..code-block:: python
+
+    output = self.FileHandler(filename = 'newfile.out')
+    output.data.write(data_for_file)
```
npetroni, line 59: Why is this extra `data` layer here? I think this should look more like `open(); write(); close()`

volatility/cli/__init__.py
```diff
@@ -66,6 +68,23 @@ def __call__(self, progress: Union[int, float], description: str = None):
...
+        self._file_pointer = tempfile.NamedTemporaryFile(prefix = prefix, suffix = suffix, delete = False)
+
+    @property
+    def data(self):
+        return self._file_pointer
+
+    def finalize_file(self, output_filename: str):
+        """Moves the temporary file into its final destination"""
+        self._file_pointer.close()
+        shutil.move(self._file_pointer.name, output_filename)
```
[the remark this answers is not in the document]
ikelos: It should (and I haven't catered for if that fails due to space issues), but I feel using the appropriate method for creating a temporary file is the right thing to do?  I could see if I can generate a temp filename in the same directory/folder?

doc/source/complex-plugin.rst
ikelos: That's from our `FileInterface`, which has a prefered name, and data.  I could override a filelike object with the preferred name, but I went with having a data attribute which contained the file like object.  This is absolutely our opportunity to change it though.  We'd need to bump some numbers because it would be a framework change if we did though (although I think we do that for this change anyway?)

ikelos:
So, some of the design for this stems from the fact that you're not actually writing a file, you're stashing data that a UI can then choose to put into a file.  A WebUI may put it in a database, or feed it to an S3 bucket, or one of any number of different things.  The reason for the FileBackedHandler is because there were complaints that volatility was simply keeping all the data in memory until it was done, and then writing it all at the end.  So not knowing exactly where/how the data was going to be used is what led us to where we are now.

Having said all that, I'm happy to completely rewrite it (and to do it now, rather than after the main release) once we're happy with it.  The initial version may well have been poorly designed, I think the produce step is the equivalent of the close step, and I'm happy to change that (but closing gives the concept of being openable again, whereas unproducing something is harder to achieve).  5;)

The additions I made this in pull request were to allow CLIs to specify the type of backend (rather than forcing a `bytesIO` which is how we did it previously).  I'm entirely up for design suggestions, which is why I wanted this reviewed.  Now all we have to do is plan out what needs improving, and how we would expect the flow to go.  5:)

ikelos:
#319 is a newer implementation of this, so closing this.

PR #276, opened 2020-08-03
created linux_check_idt

mtressler:
not ready for merge. plugin currently is not finding the module names for each entry in idt table. Just wanted to get this opened so any other issues that may need to be fixed can be addressed now.

### DIFF
--- a/volatility/framework/plugins/linux/check_idt.py
+++ b/volatility/framework/plugins/linux/check_idt.py
@@ -1,0 +1,94 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import logging
+from typing import List
+
+from volatility.framework import interfaces, renderers, constants, contexts, exceptions, symbols
+from volatility.framework.configuration import requirements
+from volatility.framework.renderers import format_hints
+from volatility.framework.symbols import linux
+from volatility.plugins.linux import lsmod
+
+vollog = logging.getLogger(__name__)
+
+
+class Check_idt(interfaces.plugins.PluginInterface):
+    """ Checks if the IDT has been altered """
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.TranslationLayerRequirement(name='primary',
+                                                     description='Memory layer for the kernel',
+                                                     architectures=["Intel32", "Intel64"]),
+
+            requirements.SymbolTableRequirement(name="vmlinux", description="Linux kernel symbols"),
+            
+            requirements.PluginRequirement(name='lsmod', plugin=lsmod.Lsmod, version=(1, 0, 0))
+        ]
+
+    def _generator(self):
+        vmlinux = contexts.Module(self.context, self.config['vmlinux'], self.config['primary'], 0)
+
+        modules = lsmod.Lsmod.list_modules(self.context, self.config['primary'], self.config['vmlinux'])
+
+        handlers = linux.LinuxUtilities.generate_kernel_handler_info(self.context, self.config['primary'], self.config['vmlinux'], modules)
+
+
+        is_32bit = not symbols.symbol_table_is_64bit(self.context, self.config["vmlinux"])
+
+        tblsz = 256
+
+        # hw handlers + system call
+        check_idxs = list(range(0, 20)) + [128]
+
+        if is_32bit:
+            if vmlinux.has_type("gate_struct"):
+                idt_type = "gate_struct"
+            else:
+                idt_type = "desc_struct"
+        else:
+            if vmlinux.has_type("gate_struct64"):
+                idt_type = "gate_struct64"
+            elif vmlinux.has_type("gate_struct"):
+                idt_type = "gate_struct"
+            else:
+                idt_type = "idt_desc"
+
+        # this is written as a list b/c there are supposdly kernels with per-CPU IDTs
+        # but I haven't found one yet... 
+        # ^ This is from vol2. Not sure if this should stay a list or if object_from_symbol
+        # would even give u a multi-element list
+        addrs = [vmlinux.object_from_symbol("idt_table")]
+
+        for tableaddr in addrs:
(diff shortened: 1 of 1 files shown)


volatility/framework/plugins/linux/check_idt.py
```diff
@@ -0,0 +1,94 @@
...
+                idt_type = "desc_struct"
+        else:
+            if vmlinux.has_type("gate_struct64"):
+                idt_type = "gate_struct64"
+            elif vmlinux.has_type("gate_struct"):
+                idt_type = "gate_struct"
+            else:
+                idt_type = "idt_desc"
+
+        # this is written as a list b/c there are supposdly kernels with per-CPU IDTs
```
mtressler, line 60: Let me know what you want me to do for this

volatility/framework/plugins/linux/check_idt.py
```diff
@@ -0,0 +1,94 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
```
ikelos: The year should be the year it's added to the repo, hopefully shortly after it's submitted 5;)...

volatility/framework/plugins/linux/check_idt.py
```diff
@@ -0,0 +1,94 @@
...
+vollog = logging.getLogger(__name__)
+
+
+class Check_idt(interfaces.plugins.PluginInterface):
+    """ Checks if the IDT has been altered """
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.TranslationLayerRequirement(name='primary',
```
ikelos, line 23: Our yapf configuration puts spaces between `=` in parameters.  Happy to consider changing the yapf style file, but it'd be good to get things formatted using it, given it should do it all for us.  5:)

volatility/framework/plugins/linux/check_idt.py
```diff
@@ -0,0 +1,94 @@
...
+        vmlinux = contexts.Module(self.context, self.config['vmlinux'], self.config['primary'], 0)
+
+        modules = lsmod.Lsmod.list_modules(self.context, self.config['primary'], self.config['vmlinux'])
+
+        handlers = linux.LinuxUtilities.generate_kernel_handler_info(self.context, self.config['primary'], self.config['vmlinux'], modules)
+
+
+        is_32bit = not symbols.symbol_table_is_64bit(self.context, self.config["vmlinux"])
+
+        tblsz = 256
```
ikelos, line 42: As long as this variable name is clear/obvious in the world of linux, this is ok.  If it's just shorthand, then long hand is more descriptive...  5;)

volatility/framework/plugins/linux/check_idt.py
```diff
@@ -0,0 +1,94 @@
...
+                    middle = ent.offset_middle
+
+                    if hasattr(ent, "offset_high"):
+                        high = ent.offset_high
+                    else:
+                        high = 0
+
+                    idt_addr = (high << 32) | (middle << 16) | low
+
+                module_name, symbol_name = linux.LinuxUtilities.lookup_module_address(self.context, handlers, idt_addr)
```
ikelos, line 88: `lookup_module_address` doesn't seem well named (not anything to do with this patch, just feels easier making a comment here than filing a whole bug about it).  @atcuno, it seems like this describes the inputs without any mention of the outputs?  Perhaps `symbol_from_address` or something that involves both would be better?  From the current title, I'd expect to give it something and have it return a module address of some kind?  Even `lookup_module_from_address` would be better...

volatility/framework/plugins/linux/check_idt.py
ikelos: The bit above looks fine, if there's a few different cases where it can differ, then we need to differentiate them.  I dunno if it would be more confusing to check whether `gate_struct` is there first (since it's in both 32 and 64 bit ones), but it looks like `gate_struct64` may need to take precedence so best keep it this way.

If it's the bit below, then I'd need to go check the vol2 code, but even if there were multiple tables, you'll only get a single object back from `object_from_symbol`, so even if it's an array, you'd still be cycling through a list of one each time...

ikelos: From the vol2 code, it looks like this is priming the code to end up dealing with `idt_table` were it ever to be a list, but I think the structure name would change in that case?  It's all internal (ie, not an exposed API), so changing it later shouldn't have any effect, so I'd say drop the list here and if the structure does ever change for some reason, we can put the code back in then.

mtressler: sounds good

volatility/framework/plugins/linux/check_idt.py
mtressler: i fixed this in the most recent commit

volatility/framework/plugins/linux/check_idt.py
mtressler: I feel spaces make the code look cleaner so I'll fix that up

volatility/framework/plugins/linux/check_idt.py
mtressler: this is directly copied from vol2 so Im not sure how clear and obvious it would be. Happy to change it if you feel it would make it more readable

ikelos: I'm happy for @atcuno to weigh in, but the fact that's how vol2 did it isn't enough reason to keep it that way if there's reason to change it (clarity and readability is a goal of vol3).  As I say, if it immediately tells linux people what they expect, that's fine, if it's *extreme* effort saving by removing vowels, then we should probably expand on it a bit...  5;)

mtressler: I would agree. Based on a quick google their are 256 entries in a idt table so it would seem like it refers to that. Changing it to something like idt_table_size seems like a better name

ikelos: YES!  That is *so* much more descriptive than tblsz (although, looking at it now I can assume it was a shortening of tablesize, but before that it could've been anything!!!  Sleeping tables, tomatoes blowsz, someone sneezing as they typed the variable name).  So yes, `idt_table_size` would be a vast improvement!!!  5:D

mtressler:
Made those changes, once the plugin is properly tracking kernel modules and ready for merge I will let you know

mtressler:
should be all good now

PR #277, opened 2020-08-03
Stephanie's memdump & memmap plugin

skeene1:
Memmap: tested with grrcon-0a7030d.img and win10-x64-2016-05-16-10586.lime
--output with grrcon 100% matches vol2's output
--with the win10 file, it prints out more valid addresses than vol2 at the end, possibly because vol3 is reporting what's valid based on PTE entries 
--found an edge case for 64-bit .lime files, need to use patch when comparing data with vol2

Memdump: tested with grrcon-0a7030d.img and win10-x64-2016-05-16-10586.lime
--output with grrcon 100% matches vol2's output (compared with md5 hash)
--same discrepancies with the .lime file
--needs further testing with 64-bit file after patch is applied

### DIFF
--- a/volatility/framework/plugins/windows/memdump.py
+++ b/volatility/framework/plugins/windows/memdump.py
@@ -7,10 +7,11 @@ from typing import List
 from volatility.framework import exceptions, renderers, interfaces
 from volatility.framework.configuration import requirements
 from volatility.framework.renderers import format_hints
+from volatility.framework.objects import utility
 from volatility.plugins.windows import pslist, dlllist, ssdt, vadinfo
 
 
-class Memmap(interfaces.plugins.PluginInterface):
+class Memdump(interfaces.plugins.PluginInterface):
     """Prints the memory map"""
 
     @classmethod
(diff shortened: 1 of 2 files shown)


ikelos:
Awesome, thanks for the submission!  5:)  I've asked @iMHLv2 to take a look from a windows perspective and I'll try to give it a review from the code side but from first glance, the coding looks spot on.  5:)

volatility/framework/plugins/windows/memdump.py
```diff
@@ -0,0 +1,76 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+from typing import List
+import sys
+from volatility.framework import exceptions, renderers, interfaces
+from volatility.framework.configuration import requirements
+from volatility.framework.renderers import format_hints
+from volatility.framework.objects import utility
```
iMHLv2, line 10: This `utility` import can be dropped

volatility/framework/plugins/windows/memdump.py
```diff
@@ -0,0 +1,76 @@
...
+from volatility.plugins.windows import pslist, dlllist, ssdt, vadinfo
```
iMHLv2, line 11: The dlllist, ssdt, and vadinfo imports can be dropped

volatility/framework/plugins/windows/memdump.py
```diff
@@ -0,0 +1,76 @@
...
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        # Since we're calling the plugin, make sure we have the plugin's requirements
+        return [
+            requirements.TranslationLayerRequirement(name = 'primary',
+                                                     description = 'Memory layer for the kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
+            requirements.PluginRequirement(name = 'pslist', plugin = pslist.PsList, version = (1, 0, 0)),
+            requirements.PluginRequirement(name = 'vadinfo', plugin = vadinfo.VadInfo, version = (1, 0, 0)),
```
iMHLv2, line 26: I don't think the vadinfo plugin is required for operation of this plugin, so this can be removed.

volatility/framework/plugins/windows/memdump.py
```diff
@@ -0,0 +1,76 @@
...
+            try:
+                pid = proc.UniqueProcessId
+                proc_layer_name = proc.add_process_layer()
+                proc_layer = self.context.layers[proc_layer_name]
+            except exceptions.InvalidAddressException as excp:
+                vollog.debug("Process {}: invalid address {} in layer {}".format(pid, excp.invalid_address, excp.layer_name))
+                continue
+
+            #Create file for writing
+            filedata = interfaces.plugins.FileInterface("{}.dmp".format(proc.UniqueProcessId))
```
iMHLv2, line 48: To avoid issues with pid re-use (this isn't common, but has been known to happen)...we should use another factor in addition to pid as the file names. For example, we could add `proc.vol.offset` which will be unique throughout the memory sample. In that case, if a process terminates but stays in the process list, and the another process starts with the same pid as the old one, we won't need to worry about file name conflicts.

volatility/framework/plugins/windows/memmap.py
```diff
@@ -0,0 +1,65 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+from typing import List
+
+from volatility.framework import exceptions, renderers, interfaces
+from volatility.framework.configuration import requirements
+from volatility.framework.renderers import format_hints
+from volatility.plugins.windows import pslist, dlllist, ssdt, vadinfo
```
iMHLv2, line 10: dlllist, ssdt, vadinfo aren't needed

volatility/framework/plugins/windows/memmap.py
```diff
@@ -0,0 +1,65 @@
...
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        # Since we're calling the plugin, make sure we have the plugin's requirements
+        return [
+            requirements.TranslationLayerRequirement(name = 'primary',
+                                                     description = 'Memory layer for the kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
+            requirements.PluginRequirement(name = 'pslist', plugin = pslist.PsList, version = (1, 0, 0)),
+            requirements.PluginRequirement(name = 'vadinfo', plugin = vadinfo.VadInfo, version = (1, 0, 0)),
```
iMHLv2, line 25: The vadinfo requirement can go away

volatility/framework/plugins/windows/memmap.py
```diff
@@ -0,0 +1,65 @@
...
+            requirements.IntRequirement(name = 'pid',
+                                        description = "Process ID to include (all other processes are excluded)",
+                                        optional = True)
+        ]
+
+    def _generator(self, procs):
+        for proc in procs:
+            reverse_map = dict()
```
iMHLv2, line 33: This isn't needed, so it can be removed

skeene1:
submitting an updated pull request

PR #282, opened 2020-08-06
Lsadump and cachedump

jxwenger:
Made minor changes to hashdump.py and added cachedump and lsadump.

### DIFF
--- a/volatility/framework/plugins/windows/cachedump.py
+++ b/volatility/framework/plugins/windows/cachedump.py
@@ -15,6 +15,9 @@ from struct import unpack
 
 class Cachedump(interfaces.plugins.PluginInterface):
     """Dumps lsa secrets from memory"""
+
+    _version = (1, 0, 0)
+    
     @classmethod
     def get_requirements(cls):
         return [requirements.TranslationLayerRequirement(name = 'primary',
@@ -22,7 +25,8 @@ class Cachedump(interfaces.plugins.PluginInterface):
                                                      architectures = ["Intel32", "Intel64"]),
             requirements.SymbolTableRequirement(name = "nt_symbols",
                                                 description = "Windows kernel symbols"),
-            requirements.PluginRequirement(name = 'hivelist', plugin = hivelist.HiveList, version = (1, 0, 0))
+            requirements.PluginRequirement(name = 'hivelist', plugin = hivelist.HiveList, version = (1, 0, 0)),
+            requirements.PluginRequirement(name = 'lsadump', plugin = lsadump.Lsadump, version = (1, 0, 0))
             ]
 
     def get_nlkm(self, sechive, lsakey, is_vista_or_later):
@@ -55,9 +59,10 @@ class Cachedump(interfaces.plugins.PluginInterface):
         enc_data = cache_data[96:]
         return (uname_len, domain_len, domain_name_len, enc_data, ch)
 
-    #Get the data from the cache and separate it into the username, domain name, and hash data
+    
     def parse_decrypted_cache(self, dec_data, uname_len,
             domain_len, domain_name_len):
+        """Get the data from the cache and separate it into the username, domain name, and hash data"""
         uname_offset= 72
         pad = 2 * ((uname_len / 2) % 2)
         domain_offset= int(uname_offset+ uname_len + pad)
--- a/volatility/framework/plugins/windows/cachedump.py
+++ b/volatility/framework/plugins/windows/cachedump.py
@@ -1,3 +1,6 @@
+# This file is Copyright 2020 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
 
 from volatility.framework import interfaces, renderers
 from volatility.framework.configuration import requirements
@@ -21,12 +24,12 @@ class Cachedump(interfaces.plugins.PluginInterface):
                                                 description = "Windows kernel symbols"),
             requirements.PluginRequirement(name = 'hivelist', plugin = hivelist.HiveList, version = (1, 0, 0))
             ]
-    @classmethod
-    def get_nlkm(cls, sechive, lsakey, is_vista_or_later):
+
+    def get_nlkm(self, sechive, lsakey, is_vista_or_later):
         return lsadump.Lsadump.get_secret_by_name(sechive, 'NL$KM', lsakey, is_vista_or_later)
     
-    @classmethod
-    def decrypt_hash(cls, edata, nlkm, ch, xp):
+
+    def decrypt_hash(self, edata, nlkm, ch, xp):
         if xp:
             hmac_md5 = HMAC.new(nlkm.encode('latin1'), ch)
             rc4key = hmac_md5.digest()
@@ -43,8 +46,7 @@ class Cachedump(interfaces.plugins.PluginInterface):
                 data += aes.decrypt(buf)
         return data
     
-    @classmethod
-    def parse_cache_entry(cls, cache_data):
+    def parse_cache_entry(self, cache_data):
         (uname_len, domain_len) = unpack("<HH", cache_data[:4])
         if len(cache_data[60:62]) == 0:
             return (uname_len, domain_len, 0, '', '')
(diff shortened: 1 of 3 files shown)


volatility/framework/plugins/windows/cachedump.py
```diff
@@ -0,0 +1,131 @@
...
+        domain = domain.decode('utf-16-le', 'replace')
+        domain_name = dec_data[domain_name_off:domain_name_off + domain_name_len]
+        domain_name = domain_name.decode('utf-16-le', 'replace')
+
+        return (username, domain, domain_name, hashh)
+
+    def _generator(self, syshive, sechive):
+        bootkey = hashdump.Hashdump.get_bootkey(syshive)
+        if not bootkey:
+            return []
```
ikelos, line 77: Since this returns a generator (rather than a list), these should probably just be `return` rather than `return []`

volatility/framework/plugins/windows/cachedump.py
```diff
@@ -0,0 +1,131 @@
...
+        (uname_len, domain_len) = unpack("<HH", cache_data[:4])
+        if len(cache_data[60:62]) == 0:
+            return (uname_len, domain_len, 0, '', '')
+        (domain_name_len,) = unpack("<H", cache_data[60:62])
+        ch = cache_data[64:80]
+        enc_data = cache_data[96:]
+        return (uname_len, domain_len, domain_name_len, enc_data, ch)
+
+    @classmethod
+    def parse_decrypted_cache(cls, dec_data, uname_len,
```
ikelos, line 57: Some docstrings providing a little explanation would be handy.

If these aren't ever intended to be called by other plugins, they don't need to be class methods and they should be marked as private (as private as python allows) by starting the method name with `_`.

volatility/framework/plugins/windows/cachedump.py
```diff
@@ -0,0 +1,131 @@
...
+    @classmethod
+    def get_requirements(cls):
+        return [requirements.TranslationLayerRequirement(name = 'primary',
+                                                     description = 'Memory layer for the kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols",
+                                                description = "Windows kernel symbols"),
+            requirements.PluginRequirement(name = 'hivelist', plugin = hivelist.HiveList, version = (1, 0, 0))
+            ]
+    @classmethod
```
ikelos, line 24: The formatting suggests yapf hasn't been run over this.  Ideally, this would have yapf run on it so that it fits the standard (but it's not a show stopper, just a nice to have).  5:)

volatility/framework/plugins/windows/cachedump.py
ikelos: If this is intended to be called by other plugins, it should also define `_version = (1, 0, 0)` to allow other plugins to version it if/when the API changes.

volatility/framework/plugins/windows/cachedump.py
```diff
@@ -0,0 +1,131 @@
...
+        ch = cache_data[64:80]
+        enc_data = cache_data[96:]
+        return (uname_len, domain_len, domain_name_len, enc_data, ch)
+
+    @classmethod
+    def parse_decrypted_cache(cls, dec_data, uname_len,
+            domain_len, domain_name_len):
+        uname_off = 72
+        pad = 2 * ((uname_len / 2) % 2)
+        domain_off = int(uname_off + uname_len + pad)
```
ikelos, line 61: Stylistically, I'd prefer `offset` than `off`, but in this case it is clear enough that this doesn't need clarification.

In general though, we prefer clear variable names, over something that a developer has to type a couple of times and then never touch again.  Particularly when autocomplete on IDEs makes it much easier to use longer variable names.

volatility/framework/plugins/windows/cachedump.py
```diff
@@ -0,0 +1,131 @@
...
+        if not bootkey:
+            return []
+        
+        is_vista_or_later = poolscanner.os_distinguisher(version_check = lambda x: x >= (6, 0),
+                                                     fallback_checks = [("KdCopyDataBlock", None, True)])
+        vista_or_later = is_vista_or_later(context = self.context, symbol_table = self.config['nt_symbols'])
+
+        lsakey = lsadump.Lsadump.get_lsa_key(sechive, bootkey, vista_or_later)
+        if not lsakey:
+            return []
```
ikelos, line 85: It might also be better to raise an exception, or use `vollog.error` to indicate that the plugin went wrong somehow.  Otherwise the user just gets no results, and has no idea why...

volatility/framework/plugins/windows/cachedump.py
```diff
@@ -0,0 +1,131 @@
...
+        nlkm = self.get_nlkm(sechive, lsakey, vista_or_later)
+        if not nlkm:
+            return []
+
+        cache = sechive.get_key("Cache")
+        if not cache:
+            return []
+        
+
+        for v in cache.get_values():
```
ikelos, line 96: Again, really not keen on `v` as a variable name.  `value` wouldn't be much better, but would be slightly better.  `cache_item` would be ideal, or something that actually describes what it is...

volatility/framework/plugins/windows/hashdump.py
```diff
@@ -0,0 +1,345 @@
+# This file is Copyright 2020 Volatility Foundation and licensed under the Volatility Software License 1.0
```
ikelos, line 1: Hash dump's already been merged, suggesting these changes need to be rebased.  This patch won't get accepted until this hashdump is either removed, or a diff from the one in master.

The previous Hashdump required a bit of work afterwards, because the plugin wasn't actually working straight off the bat, at least not with python 3.7, and it had a number of python2 sting/byte issues which needed resolving.

volatility/framework/plugins/windows/lsadump.py
```diff
@@ -0,0 +1,202 @@
...
+class Lsadump(interfaces.plugins.PluginInterface):
+    """Dumps lsa secrets from memory"""
+    @classmethod
+    def get_requirements(cls):
+        return [requirements.TranslationLayerRequirement(name = 'primary',
+                                                     description = 'Memory layer for the kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols",
+                                                description = "Windows kernel symbols"),
+            requirements.PluginRequirement(name = 'hivelist', plugin = hivelist.HiveList, version = (1, 0, 0))#,
```
ikelos, line 27: The hash at the end of this line is a bit weird.  Could it be cleared up please?

volatility/framework/plugins/windows/lsadump.py
```diff
@@ -0,0 +1,202 @@
...
+        if not enc_reg_value:
+            return None
+        
+        obf_lsa_key = sechive.read(enc_reg_value.Data+4, enc_reg_value.DataLength)
+
+        if not obf_lsa_key:
+            return None
+        if not vista_or_later:
+            md5 = MD5.new()
+            md5.update(bootkey.encode('latin1'))
```
ikelos, line 82: The bootkey should presumably be in bytes, and again, this calls into hashdump, which was reworked so that bytes were used uniformly as possible.  Please update this to work with the hashdump in the master branch.

volatility/framework/plugins/windows/lsadump.py
```diff
@@ -0,0 +1,202 @@
...
+            lsa_key = lsa_key[68:100]
+        return lsa_key
+    
+    @classmethod
+    def get_secret_by_name(cls, sechive, name, lsakey, is_vista_or_later):
+        try:
+            enc_secret_key = sechive.get_key("Policy\\Secrets\\" + name + "\\CurrVal")
+        except KeyError:
+            vollog.debug("Unable to read cache from memory")
+            exit()
```
ikelos, line 101: Please don't call exit.  This may be being called from inside a web UI or some other kind of interface.  Terminating the program is not the right way to handle this, it should probably raise an exception instead.

volatility/framework/plugins/windows/cachedump.py
```diff
@@ -0,0 +1,131 @@
+
```
ikelos, line 1: These don't include the volatility copyright header.  We'll be happy to include them in the community plugins area, but we can't include them in the core framework unless you append that copyright header to the file I'm afraid.  Please ensure that all the code you're submitting is in agreement with the terms of the license.

PR #283, opened 2020-08-07
Cachedump

jxwenger:


ikelos:
This looks like a duplicate of #282?  Since they both require rebasing to apply cleanly, I'd recommend we stick with #282 unless there's reason to use this one instead?

ikelos:
I'm happy to keep working on this, but at the moment it contains the three plugins, rather than just cachedump.  Could you please rebase it, or start an entirely new branch based off master, so that I can review just the changes you're after please?

ikelos:
It looks like #284 supersedes this?  Should this be closed?

ikelos:
Since we've #284 is being worked on, and contains cachedump, I'll close this one off.

PR #286, opened 2020-08-10
Stephanie's memdump & memmap plugin

skeene1:
Changes suggested from first pull request.

Memdump outputs to pid.offset.dmp rather than pid.dmp


ikelos:
Referencing #277 to link back to the history of this.

PR #287, opened 2020-08-12
Improved strings efficiency

AsafEitani:
Strings takes forever and there is no indication of progress.
1. Replaced readlines() with readline() to prevent loading the entire strings file to memory all at once.
2. Added process callback to indicate the progression.
3. Converted the re.compile of strings parse_line to global in order to prevent compilation for each line.
4. Changed the regex to not include the trailing newline (\n)

### DIFF
--- a/volatility/framework/plugins/windows/strings.py
+++ b/volatility/framework/plugins/windows/strings.py
@@ -5,7 +5,7 @@
 import logging
 import re
 from typing import Dict, Generator, List, Set, Tuple
-from os.path import getsize
+from os import path
 
 from volatility.framework import interfaces, renderers, exceptions
 from volatility.framework.configuration import requirements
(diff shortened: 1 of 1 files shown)


volatility/framework/plugins/windows/strings.py
```diff
@@ -5,6 +5,7 @@
 import logging
 import re
 from typing import Dict, Generator, List, Set, Tuple
+from os.path import getsize
```
ikelos, line 8: Typically (save for the typing module which is used literally everywhere), we tend to do `from os import path` and then do `path.getsize`.  This helps ensure that no one does `from volatility.plugins.strings import getsize` because it's easy and then it breaks later on.

AsafEitani: Great tip! thanks 👍 
I updated the code so you are more than welcomed to merge 🥇 !

ikelos:
Awesome, thanks for that!  5:)

PR #288, opened 2020-08-12
Maldump: Initial implementation

### LINKED ISSUE #270: malfind output directory
malfind plugin doesn't save files 

on old vol2:  volatility -f [memory_dump_file] –profile=xxxxx malfind -D [dir_path]
on vol3: have output dir for malfind

ikelos:
There were two options here: either use --dump and make malfind depend
on vaddump (which is incosistent because vaddump should be vadinfo
--dump in that case) or maintain consistency and not burden malfind with
additional requirements.

I went the second route, but I'll be happy to move things to a --dump
flag if that's the prefered route, as long as we apply it everywhere we
support dumping files.

Closes #270

### DIFF
--- a/volatility/framework/plugins/windows/maldump.py
+++ b/volatility/framework/plugins/windows/maldump.py
@@ -1,0 +1,64 @@
+# This file is Copyright 2020 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import logging
+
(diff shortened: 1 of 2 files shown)


ikelos:
Still happy to change this to `--dump` if we prefer that in general?  I don't mind what we do, I'd just like to make sure is the same everywhere.  I'll merge this in 7 days unless anyone shouts...

ikelos:
Right, I think we've decided to go for `--dump`, so I'll close this off.

ikelos:
(See #290 for those who are interested).  5:)

PR #289, opened 2020-08-12
Issue 197 windows dumpfiles

iMHLv2:
Add the windows.dumpfiles plugin. This can be used to dump a single FILE_OBJECT by virtual or physical offset, or it can dump all files for a process (found via VADs and the handle table).

### DIFF
--- a/volatility/framework/plugins/windows/dumpfiles.py
+++ b/volatility/framework/plugins/windows/dumpfiles.py
@@ -1,0 +1,233 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import logging
+import ntpath
+from volatility.framework import interfaces, renderers, exceptions, constants
+from volatility.plugins.windows import handles
+from volatility.plugins.windows import pslist
+from volatility.framework.configuration import requirements
+from volatility.framework.renderers import format_hints
+from volatility.framework.objects import utility
+from typing import List, Tuple
+vollog = logging.getLogger(__name__)
+
+FILE_DEVICE_DISK = 0x7
+FILE_DEVICE_NETWORK_FILE_SYSTEM = 0x14
+EXTENSION_CACHE_MAP = {
+    "dat": "DataSectionObject",
+    "img": "ImageSectionObject",
+    "vacb": "SharedCacheMap",
+}
+
+class DumpFiles(interfaces.plugins.PluginInterface):
+    """Dumps cached file contents from Windows memory samples."""
+
+    _version = (1, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        # Since we're calling the plugin, make sure we have the plugin's requirements
+        return [
+            requirements.TranslationLayerRequirement(name='primary',
+                                                     description='Memory layer for the kernel',
+                                                     architectures=["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name="nt_symbols", description="Windows kernel symbols"),
+            requirements.IntRequirement(name='pid',
+                                        description="Process ID to include (all other processes are excluded)",
+                                        optional=True),
+            requirements.IntRequirement(name='fileoffset',
+                                        description="Dump a single _FILE_OBJECT at this offset",
+                                        optional=True),
+            requirements.PluginRequirement(name='pslist', plugin=pslist.PsList, version=(1, 0, 0)),
+            requirements.PluginRequirement(name='handles', plugin=handles.Handles, version=(1, 0, 0))
+        ]
+
+    def dump_file_producer(self, file_object: interfaces.objects.ObjectInterface,
+                           memory_object: interfaces.objects.ObjectInterface,
+                           layer: interfaces.layers.DataLayerInterface,
+                           desired_file_name: str) -> str:
+        """Produce a file from the memory object's get_available_pages() interface.
+
+        :param file_object: the parent _FILE_OBJECT
+        :param memory_object: the _CONTROL_AREA or _SHARED_CACHE_MAP
+        :param layer: the memory layer to read from
+        :param desired_file_name: name of the output file
+        :return: result status
+        """
+        filedata = interfaces.plugins.FileInterface(desired_file_name)
+        try:
+            # Description of these variables:
+            #   memoffset: offset in the specified layer where the page begins
+            #   fileoffset: write to this offset in the destination file
+            #   datasize: size of the page
+            for memoffset, fileoffset, datasize in memory_object.get_available_pages():
+                data = layer.read(memoffset, datasize, pad = True)
+                filedata.data.seek(fileoffset)
+                filedata.data.write(data)
+
+            # Avoid writing files to disk if they are going to be empty or all zeros.
+            cached_length = len(filedata.data.getvalue())
+            if cached_length == 0 or filedata.data.getvalue().count(0) == cached_length:
+                result_text = "No data is cached for the file at {0:#x}".format(file_object.vol.offset)
+            else:
+                self.produce_file(filedata)
+                result_text = "Stored {}".format(filedata.preferred_filename)
+        except exceptions.InvalidAddressException:
+            result_text = "Unable to dump file at {0:#x}".format(file_object.vol.offset)
+
+        return result_text
+
+    def process_file_object(self, file_obj: interfaces.objects.ObjectInterface) -> Tuple:
+        """Given a FILE_OBJECT, dump data to separate files for each of the three file caches.
+
+        :param file_object: the FILE_OBJECT
+        """
+
+        # Filtering by these types of devices prevents us from processing other types of devices that
+        # use the "File" object type, such as \Device\Tcp and \Device\NamedPipe.
+        if file_obj.DeviceObject.DeviceType not in [FILE_DEVICE_DISK, FILE_DEVICE_NETWORK_FILE_SYSTEM]:
+            vollog.log(constants.LOGLEVEL_VVV,
+                       "The file object at {0:#x} is not a file on disk".format(file_obj.vol.offset))
+            return
+
+        # Depending on the type of object (DataSection, ImageSection, SharedCacheMap) we may need to
+        # read from the memory layer or the primary layer.
+        memory_layer = self.context.layers["memory_layer"]
+        primary_layer = self.context.layers[self.config["primary"]]
+
+        obj_name = file_obj.file_name_with_device()
+
+        # This stores a list of tuples, describing what to dump and how to dump it.
+        # Ex: (
+        #     memory_object with get_available_pages() API (either CONTROL_AREA or SHARED_CACHE_MAP),
+        #     layer to read from,
+        #     file extension to apply,
+        #     )
+        dump_parameters = []
+
+        # The DataSectionObject and ImageSectionObject caches are handled in basically the same way.
+        # We carve these "pages" from the memory_layer.
+        for member_name, extension in [("DataSectionObject", "dat"), ("ImageSectionObject", "img")]:
+            try:
+                section_obj = getattr(file_obj.SectionObjectPointer, member_name)
+                control_area = section_obj.dereference().cast("_CONTROL_AREA")
+                if control_area.is_valid():
+                    dump_parameters.append((control_area, memory_layer, extension))
+            except exceptions.InvalidAddressException:
+                vollog.log(constants.LOGLEVEL_VVV,
+                           "{0} is unavailable for file {1:#x}".format(member_name, file_obj.vol.offset))
+
+        # The SharedCacheMap is handled differently than the caches above.
+        # We carve these "pages" from the primary_layer.
+        try:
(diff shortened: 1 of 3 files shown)


volatility/framework/plugins/windows/dumpfiles.py
```diff
@@ -0,0 +1,245 @@
...
+            requirements.IntRequirement(name='pid',
+                                        description="Process ID to include (all other processes are excluded)",
+                                        optional=True),
+            requirements.IntRequirement(name='virtaddr',
+                                        description="Dump a single _FILE_OBJECT at this virtual address",
+                                        optional=True),
+            requirements.IntRequirement(name='physaddr',
+                                        description="Dump a single _FILE_OBJECT at this physical address",
+                                        optional=True),
+            requirements.PluginRequirement(name='pslist', plugin=pslist.PsList, version=(1, 0, 0)),
```
ikelos, line 46: I'm encouraging new plugins to use `VersionRequirement(..., component=<whateverobject>...)` over `PluginRequirement(..., plugin=<whateverobject>,...)`, they're functionally equivalent, but VersionRequirement is more generic and at some point I'd like to get rid of PluginRequirement unless we think this particularly aides readability?

volatility/framework/plugins/windows/dumpfiles.py
```diff
@@ -0,0 +1,245 @@
...
+            requirements.PluginRequirement(name='handles', plugin=handles.Handles, version=(1, 0, 0))
+        ]
+
+    def dump_file_producer(self, file_object: interfaces.objects.ObjectInterface,
```
ikelos, line 50: It would be better if this returned a `FileInterface` object, so it doesn't need `self` and can become a `classmethod`?  I'd return `None` in place of error text, but definitely include useful information in a `vollog.debug` message.

ikelos: Either that, or we could actually raise an Exception?  It'll need catching in either circumstance?  Even though I'm usually a fan of exceptions, they do tend to require a lot of code for catching, so as long as it's caught and logged in here, I'm ok returning None.

volatility/framework/plugins/windows/dumpfiles.py
```diff
@@ -0,0 +1,245 @@
...
+                result_text = "No data is cached for the file at {0:#x}".format(file_object.vol.offset)
+            else:
+                self.produce_file(filedata)
+                result_text = "Stored {}".format(filedata.preferred_filename)
+        except exceptions.InvalidAddressException:
+            result_text = "Unable to dump file at {0:#x}".format(file_object.vol.offset)
+
+        return result_text
+
+    def process_file_object(self, file_obj: interfaces.objects.ObjectInterface) -> Tuple:
```
ikelos, line 85: This definitely feels like a candidate for becoming a `classmethod`.  I think it would need `context, layer_name`.  I feel we could test that the layer is an Intel layer, and then we'd know we could get the physical layer using `contexts.layers[layer_name].config['base_layer']`.  It's slightly dangerous but far less dangerous than hard coding the layer name to pull.  For the safest option, we'd ask for two layer names to the classmethod...

volatility/framework/plugins/windows/dumpfiles.py
```diff
@@ -0,0 +1,245 @@
...
+        # Filtering by these types of devices prevents us from processing other types of devices that
+        # use the "File" object type, such as \Device\Tcp and \Device\NamedPipe.
+        if file_obj.DeviceObject.DeviceType not in [FILE_DEVICE_DISK, FILE_DEVICE_NETWORK_FILE_SYSTEM]:
+            vollog.log(constants.LOGLEVEL_VVV,
+                       "The file object at {0:#x} is not a file on disk".format(file_obj.vol.offset))
+            return
+
+        # Depending on the type of object (DataSection, ImageSection, SharedCacheMap) we may need to
+        # read from the memory layer or the primary layer.
+        memory_layer = self.context.layers["memory_layer"]
```
ikelos, line 100: This is assumes that every plugin will always create a primary layer that manages to construct a memory layer, and that it's called memory layer.  This breaks if there's more than one intel layer floating around.  This either needs to check that primary is an intel layer, and then can ask it for its config option `base_layer`, or you can try to traverse `layer.dependencies` but that will list every layer it sits on (including any swap layers).

volatility/framework/plugins/windows/dumpfiles.py
ikelos: Consistency would suggest adding a `context` parameter, and using `layer_name` rather than `layer` (simple variables over complex ones carrying their own context, but in this circumstance, because the context would only be used to lookup the layer, I'll let the layer be passed in whole.  Equally happy for people to consider it inconsistent and us to add the context in as a parameter too, so takes your choice.  5;P

volatility/framework/plugins/windows/dumpfiles.py
```diff
@@ -0,0 +1,245 @@
...
+
+        for memory_object, layer, extension in dump_parameters:
+            cache_name = EXTENSION_CACHE_MAP[extension]
+            desired_file_name = "file.{0:#x}.{1:#x}.{2}.{3}.{4}".format(file_obj.vol.offset,
+                                                                        memory_object.vol.offset,
+                                                                        cache_name,
+                                                                        ntpath.basename(obj_name),
+                                                                        extension)
+
+            result_text = self.dump_file_producer(file_obj, memory_object, layer, desired_file_name)
```
ikelos, line 144: As mentioned, I'd return a `FileInterface` object here, and if you still want specific text (even though it's effectively a boolean flag) then I'd return a tuple here.  (Personally, I'd prefer the boolean based on whether FileInterface is None, but I can live with either).  5:)

volatility/framework/plugins/windows/dumpfiles.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def _generator(self, procs: List, offsets: List):
+
+        if procs:
+            # The handles plugin doesn't expose any staticmethod/classmethod, and it also requires stashing
+            # private variables, so we need an instance (for now, anyway). We _could_ call Handles._generator()
+            # to do some of the other work that is duplicated here, but then we'd need to parse the TreeGrid
+            # results instead of just dealing with them as direct objects here.
+            handles_plugin = handles.Handles(context=self.context, config_path=self._config_path)
+            type_map = handles_plugin.get_type_map(context=self.context,
```
ikelos, line 158: I'm surprised these aren't classmethods?  Definitely looks like one to fix up...

volatility/framework/plugins/windows/dumpfiles.py
```diff
@@ -0,0 +1,245 @@
...
+        # Since we're calling the plugin, make sure we have the plugin's requirements
+        return [
+            requirements.TranslationLayerRequirement(name='primary',
+                                                     description='Memory layer for the kernel',
+                                                     architectures=["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name="nt_symbols", description="Windows kernel symbols"),
+            requirements.IntRequirement(name='pid',
+                                        description="Process ID to include (all other processes are excluded)",
+                                        optional=True),
+            requirements.IntRequirement(name='virtaddr',
```
ikelos, line 40: We might want to make these ListRequirements instead?

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -361,7 +362,9 @@ def is_valid(self) -> bool:
     def file_name_with_device(self) -> Union[str, interfaces.renderers.BaseAbsentValue]:
         name = renderers.UnreadableValue()  # type: Union[str, interfaces.renderers.BaseAbsentValue]
 
-        if self._context.layers[self.vol.layer_name].is_valid(self.DeviceObject):
+        # this pointer needs to be checked against native_layer_name because the object may
+        # be instantiated from a primary (virtual) layer or a memory (physical) layer.
+        if self._context.layers[self.vol.native_layer_name].is_valid(self.DeviceObject):
```
ikelos, line 367: It's a small nitpick, but I'd prefer we use the property self.context, rather than the underlying dataholder.  It gives us more flexibility in the future and prevents untrained developers from thinking they can access private members publicly...

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -667,3 +670,305 @@ def to_list(self,
...
+                             offset=offset,
+                             native_layer_name=self.vol.native_layer_name)
+
+    def get_available_pages(self) -> Iterable[Tuple[int, int, int]]:
+        """Get the available pages that correspond to a cached file.
+
+        The tuples generated are (physical_offset, file_offset, page_size).
+        """
+        symbol_table_name = self.get_symbol_table_name()
+        mmpte_type = self._context.symbol_space.get_type(symbol_table_name + constants.BANG + "_MMPTE")
```
ikelos, line 723: And again, self.context should do the same thing

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -667,3 +670,305 @@ def to_list(self,
...
+                # Go to the next PTE entry
+                ptecount += 1
+
+            # Go to the next Subsection in the single-linked list
+            subsection = subsection.NextSubsection
+
+class VACB(objects.StructType):
+    """A class for _VACB structures"""
+
+    FILEOFFSET_MASK = 0xFFFFFFFFFFFF0000
```
ikelos, line 781: This feels nasty.  Presumably it needs changing if memory spaces ever grow beyond 64-bits?  I'd prefer not to have bounded unbounded masks if avoidable?  Feel like something that `val ^ (val & 0xffff)` would achieve?

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -667,3 +670,305 @@ def to_list(self,
...
+    VACB_ARRAY = 0x80
+
+    def is_valid(self) -> bool:
+        """Determine if the object is valid."""
+
+        if self.FileSize.QuadPart <= 0 or self.ValidDataLength.QuadPart <= 0:
+            return False
+
+        if self.SectionSize.QuadPart < 0 or ((self.FileSize.QuadPart < self.ValidDataLength.QuadPart) and (
+                   self.ValidDataLength.QuadPart != 0x7fffffffffffffff)):
```
ikelos, line 805: In this instance it feels like a specific value, but will this always be the right value for this check and type combination?

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -667,3 +670,305 @@ def to_list(self,
...
+        :return: Collected VACBs
+        """
+        if vacb_list is None:
+            vacb_list = []
+
+        if level > limit:
+            return []
+
+        symbol_table_name = self.get_symbol_table_name()
+        pointer_type = self._context.symbol_space.get_type(symbol_table_name + constants.BANG + "pointer")
```
ikelos, line 827: More `_context` and two lines down too

mtressler:
Hey ikelos, I'll be taking over finishing this plugin. I don't have write access to the vol3 repo so I can't add to this PR so I'll be opening up a new one but I'm trying to have a new commit for each of the issues you raised in this PR so it can be easily referenced.

volatility/framework/plugins/windows/dumpfiles.py
mtressler: Running into some issues with this. I was attempting to go with the two layer name option where the caller passes in the primary and base layer but MHL and I were having some issues with figuring out how the caller should determine what the base layer is. config["base_layer"] is giving a key error. If you have any ideas on how the caller can determine the base layer that would be much appreciated!

ikelos: Sorry, my bad, it's not `base_layer` for intel layers, it's `memory_layer` (from https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/layers/intel.py#L226 ).  So `memory_layer` is a configuration option of `intel` layers specifically.  Hopefully that's enough to get what you needed, but you'll need to get it from the intel layer's config, not `self.config`.

volatility/framework/symbols/windows/extensions/__init__.py
iMHLv2: @ikelos We are confused by this comment. There is no `self.context` property - attempting to do that results in `AttributeError: StructType has no attribute: nt_symbols1!_FILE_OBJECT.context`. There are over 20 references to `self._context` among other extensions in the same `volatility/framework/symbols/windows/extensions/__init__.py ` file, but 0 references to `self.context`. Can you clarify please?

ikelos: Yep, sorry, `self.context` doesn't exist on objects, it's mostly for layers.  I could move it to `self.vol.context` except it's kinda different.  I'm not sure you'd ever need to externally ask an object for its context?  Given it's *self*, it's allowed to access protected variables, so I'm probably being overly sensitive.  Yep, completely ignore this, sorry!  5:S

volatility/framework/plugins/windows/dumpfiles.py
ikelos: Not `base_layer`, `memory_layer`, sorry.  5:S

ikelos:
This is essentially resolved through #396 so I'm going to mark it as closed.  Feel free to reopen if I missed something...  5:)

PR #290, opened 2020-08-13
Plugins: Change most *dump plugins to --dump

ikelos:
Having had discussions around this, we decided we wanted some consistency around whether dumping files should be separate functions or functionality of an appropriate plugin.  This demonstrates the changes and a testable branch where dumping is done by `--dump`.  No time limit, since this a fairly big shift.  Hopefully everything's good, but lemme know what you think...


volatility/framework/plugins/windows/dlllist.py
```diff
@@ -25,13 +32,65 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
             requirements.ListRequirement(name = 'pid',
                                          element_type = int,
                                          description = "Process IDs to include (all other processes are excluded)",
-                                         optional = True)
+                                         optional = True),
+            requirements.BooleanRequirement(name = 'dump',
+                                            description = "Extract listed processes",
```
iMHLv2, line 37: This should be Extract listed DLLs

volatility/framework/plugins/windows/memmap.py
```diff
@@ -43,22 +47,40 @@ def _generator(self, procs):
...
                 yield (0, (
                     format_hints.Hex(offset),
                     format_hints.Hex(mapped_offset),
                     format_hints.Hex(mapped_size),
-                    format_hints.Hex(offset)))
+                    format_hints.Hex(offset),
+                    dumped))
                 offset += mapped_size
 
+            import pdb
```
iMHLv2, line 74: Snuck in by accident?

volatility/framework/plugins/windows/memmap.py
```diff
@@ -43,22 +47,40 @@ def _generator(self, procs):
                     pid, excp.invalid_address, excp.layer_name))
                 continue
 
+            filename = str(pid)
```
iMHLv2, line 50: Why does this need str()?

volatility/framework/plugins/windows/registry/hivelist.py
```diff
@@ -51,15 +52,52 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                                            optional = True,
                                            default = None),
             requirements.PluginRequirement(name = 'hivescan', plugin = hivescan.HiveScan, version = (1, 0, 0)),
+            requirements.BooleanRequirement(name = 'dump',
+                                            description = "Extract listed processes",
```
iMHLv2, line 56: Extract listed hives?

volatility/framework/plugins/windows/vadinfo.py
```diff
@@ -57,6 +57,10 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                                              element_type = int,
                                              optional = True),
                 requirements.PluginRequirement(name = 'pslist', plugin = pslist.PsList, version = (1, 0, 0)),
+                requirements.BooleanRequirement(name = 'dump',
+                                                description = "Extract listed processes",
```
iMHLv2, line 61: Extract listed VADs (or Extract listed memory ranges)

volatility/framework/plugins/windows/dlllist.py
```diff
@@ -25,13 +32,65 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+                                            default = False,
+                                            optional = True)
         ]
 
+    @classmethod
+    def dump_dll(cls,
```
iMHLv2, line 43: We could potentially rename this to dump_pe since it is also used by modules/modscan and those aren't technically DLLs, but its not a big deal.

volatility/framework/plugins/windows/memmap.py
ikelos: Pid's an int, not a string, and we're going to cat a string on the end of it, so it needs to be cast...

ikelos:
Thanks, awesome spots (so disappointed when an `import pdb` sneaks in!).  All changes made.  If you're happy I'll push this since I'm pretty happy (although I know we'll have to answer a billion questions about where the dump plugins went)...  5:)

volatility/framework/plugins/windows/memmap.py
iMHLv2: I see...I was confused because I think format() will convert it to a string anyway, but explicitly doing it before passing to format() works too.

iMHLv2:
> Any news on running those plugins @iMHLv2 ? This seems good and ready to go in unless you find an issue?

I ran without issues, so it seems good to go....unless you want to hang onto it for just another couple hours and make sure no one else on the call sees an issue with moving all the *dump plugins to --dump.

volatility/framework/plugins/windows/memmap.py
ikelos: Oh, no, you're right.  I guess I was pulling it out for debugging, we might as well pass it straight to the format thing instead of putting it in an holding variable for one line...  5:)

PR #292, opened 2020-08-13
Strings efficiency - bugfix

AsafEitani:
Sorry about that - due to the fact that accessor returns a urllib file pointer and not a regular filepointer, there is a need to specify the .file attribute


ikelos:
Thanks, I've made the change manually because it was a small fix and this wasn't made off current master, but off the previous branch which had already been merged.  The fix should have been applied in commit 595a9430, but please let me know if there's anything I missed...  5:)

PR #293, opened 2020-08-13
Add dumping option to malfind

AsafEitani:
1. Added optional requirement to output_dir,
2. Added dumping the memory region to file.
3. Changed list_injections from classmethod to normal method (in order to use self.product_file like in dlldump)
4. Changed is_vad_empty to staticmethod
5. Added the result output, can be "Displayed" if no output_dir is provided, "Stored" if successfully dumped the memory region and "Unable to dump" if the operation was unsuccessful.

### DIFF
--- a/volatility/framework/plugins/windows/malfind.py
+++ b/volatility/framework/plugins/windows/malfind.py
@@ -21,20 +21,23 @@ class Malfind(interfaces.plugins.PluginInterface):
     def get_requirements(cls):
         # Since we're calling the plugin, make sure we have the plugin's requirements
         return [
-            requirements.TranslationLayerRequirement(name = 'primary',
-                                                     description = 'Memory layer for the kernel',
-                                                     architectures = ["Intel32", "Intel64"]),
-            requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
-            requirements.ListRequirement(name = 'pid',
-                                         element_type = int,
-                                         description = "Process IDs to include (all other processes are excluded)",
-                                         optional = True),
-            requirements.PluginRequirement(name = 'pslist', plugin = pslist.PsList, version = (1, 0, 0)),
-            requirements.PluginRequirement(name = 'vadinfo', plugin = vadinfo.VadInfo, version = (1, 0, 0)),
+            requirements.TranslationLayerRequirement(name='primary',
+                                                     description='Memory layer for the kernel',
+                                                     architectures=["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name="nt_symbols", description="Windows kernel symbols"),
+            requirements.ListRequirement(name='pid',
+                                         element_type=int,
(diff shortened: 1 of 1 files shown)


volatility/framework/plugins/windows/malfind.py
```diff
@@ -21,20 +21,23 @@ class Malfind(interfaces.plugins.PluginInterface):
...
-            requirements.PluginRequirement(name = 'vadinfo', plugin = vadinfo.VadInfo, version = (1, 0, 0)),
+            requirements.TranslationLayerRequirement(name='primary',
+                                                     description='Memory layer for the kernel',
+                                                     architectures=["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name="nt_symbols", description="Windows kernel symbols"),
+            requirements.ListRequirement(name='pid',
+                                         element_type=int,
+                                         description="Process IDs to include (all other processes are excluded)",
+                                         optional=True),
+            requirements.StringRequirement(name='output_dir',
```
ikelos, line 32: Thanks, but we don't ask for directories within plugins (because we can't know if this will be access through a webUI or some other form as a library).  That's why we use the `FileInterface` and only provide it a filename, that then allows UIs to render the files however they feel best.  In #290 we've been using a boolean flag of `--dump` and for consistency we'd prefer this is how all dump options work?

volatility/framework/plugins/windows/malfind.py
```diff
@@ -101,11 +105,24 @@ def list_injections(
...
+                        a = proc_layer.read(vad.get_start(), vad.get_end() - vad.get_start(), pad=True)
+                        filedata.data.write(a)
+
+                        self.produce_file(filedata)
+                        result_text = "Stored {}".format(filedata.preferred_filename)
+                    except Exception as ex:
+                        print(ex)
+                        result_text = "Unable to dump memory region at {0:#x}".format(vad.get_start())
+
+                yield vad, data, result_text
```
ikelos, line 125: I had previously been providing `result_text`, but I realized there's not a great deal of information conveyed and we should probably do this using `vollog` (again, examples in #290).  If this is just dumping a vad, you may be able to use the classmethod from vaddump (or as of #290, from vadinfo)...

volatility/framework/plugins/windows/malfind.py
```diff
@@ -88,8 +90,10 @@ def list_injections(
             return
 
         proc_layer = context.layers[proc_layer_name]
+        output_dir = self.config.get('output_dir', None)
```
ikelos, line 93: This has changed `list_injections` from being usable by other plugins, to being only usable by this plugin.  This seems like quite a restrictive change (and may impact other plugins, and would require an increase of the plugin's major version number since it's a removal of functionality).  This all seems in aid of getting `self.config` rather than accepting a `dump` field.  Since that's unlikely to be something other plugins care about, I'd sooner it were in the `_generator` method, which gets to see each vad as well, and so can do the dumping and see the config option directly.  #290 should show examples of this...

PR #295, opened 2020-08-14
Add malfind --dump functionality as in #290

AsafEitani:
rePR according to #290


volatility/framework/plugins/windows/malfind.py
```diff
@@ -62,6 +66,43 @@ def is_vad_empty(cls, proc_layer, vad):
 
         return True
 
+    @classmethod
+    def vad_dump(cls, context: interfaces.context.ContextInterface, proc: interfaces.objects.ObjectInterface, vad)\
```
ikelos, line 70: So, in #290 there's already a method called `vad_dump` which seems like it does the same thing (although this appears to cleaner).  In either circumstance, I feel like the functionality should live in the vadinfo plugin (because there's nothing inherently malware related in the `vad_dump` method).  So I'd like to move this code into the vad plugin, but rather than messing with #290, it might be better to wait until it lands, then swap this method in for `vad_dump` in the `vadinfo` plugin, and just add a VersionRequirement on `vadinfo` and call the `vad_dump` method from there...  Hopefully #290 will land quite soon, it seems like the review's gone well.

volatility/framework/plugins/windows/malfind.py
```diff
@@ -21,16 +21,20 @@ class Malfind(interfaces.plugins.PluginInterface):
...
+            requirements.SymbolTableRequirement(name="nt_symbols", description="Windows kernel symbols"),
+            requirements.ListRequirement(name='pid',
+                                         element_type=int,
+                                         description="Process IDs to include (all other processes are excluded)",
+                                         optional=True),
+            requirements.BooleanRequirement(name='dump',
+                                            description="Extract injected VADs",
+                                            default=False,
+                                            optional=True),
+            requirements.PluginRequirement(name='pslist', plugin=pslist.PsList, version=(1, 0, 0)),
```
ikelos, line 36: So `PluginRequirement` is now technically the old way of doing things, and this will work, but to save us having to change it all later, I'd prefer to use `requirements.VersionRequirement(name='pslist', component=pslist.PsList, version=(1,0,0))`.  It's functionally identical, but we'll be using VersionRequirement in the future because it's more generic (and allows us to version anything we want rather than just plugins).

ikelos: Happy to be overridden on this one too, if people feel it'd be better to have specific names here...  5:)

AsafEitani:
Cool, I made the changes. Waiting for #290 to merge and I'll commit the final changes.

ikelos:
So #290 has been merged, so feel free to push your changes and we can get this in as well.  5:)

dadokkio:
I'm testing the new dump flag and it's working, the only difference with old releases is that there was a Result column with the name of stored dumped files. Now I can only see the Dumped True/False column and I'm not able anymore to link each row with his dumped file. Can I found that information in different way? Or can that be added to --dump?

ikelos:
So it's up to the plugins to decide what to output, and there's strictly no link between the number of files dumped and the number of rows if the plugin doesn't want it.  Having said that, we're reworking the file interface, and as part of that I've changed the column to include the filename or an error rather than just a boolean flag.

AsafEitani:
Updated the vadinfo and malfind to support the --dump flag.
Tested them both - works great 5:)

dadokkio:
> So it's up to the plugins to decide what to output, and there's strictly no link between the number of files dumped and the number of rows if the plugin doesn't want it.  Having said that, we're reworking the file interface, and as part of that I've changed the column to include the filename or an error rather than just a boolean flag.  You can see what it will probably look like in the (untested) `new-file-interface2` branch.

I'll test thant on Tuesday.. just one other little info.. if I don't set dump option true or false it return a generic 'dump' warning/error.. it's that ok? I need to explicitly set it or there will be a default?

ikelos:
Yeah, it should be a default (as specified in the requirement)?  If I've missed it on one of them lemme know and I'll fix it...

dadokkio:
For the moment I'm guessing the name of the dumped process with some glob help :smiley: 
For now it's ok, I'll be interesting to find out what happens when 2 files are generated.

Regarding the default I think it was related with issue 324, but now I can omit dump=False and it's working properly.

PR #299, opened 2020-08-15
Documentation: Add an initial glossary

ikelos:
This is an attempt to add some definitions and terms, so that everybody uses the same nomenclature and knows how to refer to particular elements/aspects within the volatility world.  Happy for any changes (hence the review), but ideally this will be the place we define things that people can refer to for clarification of names.

@npetroni You seem best placed to look this over, please let me know if I've been inaccurate (almost certain) or even whaaaay off (entirely possible) with some of my definitions and/or if there's any important ones I've missed.

The concept is not necessarily to link every time we use one of these words in the rest of the documentation, but this is a possibility should we desire it...

### DIFF
--- a/doc/source/glossary.rst
+++ b/doc/source/glossary.rst
@@ -162,27 +162,31 @@ Template
 .. _Translation Layer:
 
 Translation Layer
-    This is a specific type of :ref:`data layer<Data Layer>`, a non-contiguous group of bytes that can be references by
-    a unique :ref:`offset<Offset>` within the layer.  In particular, translation layers translates (or :ref:`maps<Map>`)
-    requests made of it to a location within a lower layer.  This can be either linear (a one-to-one mapping between bytes)
-    or non-linear (a group of bytes :ref:`maps<Map>` to a larger or smaller group of bytes.
+    This is a type of data layer which allows accessing data from lower layers using addresses different to those
+    used by the lower layers themselves.  When accessing data in a translation layer, it translates (or :ref:`maps<Map>`)
+    addresses from its own :ref:`address space<Address Space>` to the address space of the lower layer and returns the
+    corresponding data from the lower layer.  Note that multiple addresses in the higher layer might refer to the same
+    address in the lower layer.  Conversely, some addresses in the higher layer might have no corresponding address in the
+    lower layer at all.  Translation layers most commonly handle the translation from virtual to physical addresses,
+    but can be used to translate data to and from a compressed form or translate data from a particular file format
+    into another format.
 
 .. _Type:
 
 Type
     This is a structure definition of multiple elements that expresses how data is laid out.  Basic types define how
-    the data should be interpretted in terms of a run of bits (or more commonly a collection of 8 bits at a time,
-    called bytes).  More complex types can be made up of other types combined together at specific locations known
-    as :ref:`structs<Struct>` or repeated, known as :ref:`array<Array>`.  They can even defined types at the same
-    location depending on the data itself, known as :ref:`Unions<Union>`.  Once a type has been linked to a specific
-    chunk of data, the result is referred to as an :ref:`object<object>`.
+    the data should be interpreted in terms of a run of bits (or more commonly a collection of 8 bits at a time,
+    called bytes).  New types can be constructed by combining other types at specific relative offsets, forming something
+    called a :ref:`struct<Struct>`, or by repeating the same type, known as an :ref:`array<Array>`.  They can even
+    contain other types at the same offset depending on the data itself, known as :ref:`Unions<Union>`.  Once a type
+    has been linked to a specific chunk of data, the result is referred to as an :ref:`object<object>`.
 
 U
 -
 .. _Union:
 
 Union
-    A union is a type that can have can hold multiple different subtypes, which specifically overlap.  A union is means
-    for holding two different types within the same size of data, meaning that not all types within the union will hold
-    valid data at the same time, more that depending on what the union is holding, a subset of the type will point to
-    accurate data (assumption no corruption).
+    A union is a type that can hold multiple different subtypes, whose relative offsets specifically overlap.
+    A union is a means for holding multiple different types within the same size of data, the relative offsets of the
+    types within the union specifically overlap.  This means that the data in a union object is interpreted differently
+    based on the types of the union used to access it.
--- a/doc/source/glossary.rst
+++ b/doc/source/glossary.rst
@@ -1,0 +1,179 @@
+Glossary
+========
+There are many terms when talking about memory forensics, this list hopes to define the common ones and
+provide some commonality on how to refer to particular ideas within the field.
+
+A
+-
+.. _Address:
+    An address is another name for an :ref:`offset<Offset>`, specifically an offset within memory.  Offsets can be
+    both relative or absolute, whereas addresses are almost always absolute.
+
+.. _Address Space:
+
+Address Space
+    This is the name in volatility 2 for what's referred to as a :ref:`Translation Layer<translation layer>`.  It
+    encompasses all values that can be addresses, usually in reference to addresses in memory.
+
+.. _Alignment:
+
+Alignment
+    This value is what all data :ref:`offsets<offset>` will typically be a multiple of within a :ref:`type<type>`.
+
+.. _Array:
+
+Array
+    This represents a list of items, which can be access by an index, which is zero-based (meaning the first
+    element has index 0).  Items in arrays are almost always the same size (it is not a generic list, as in python)
+    even if they are :ref:`pointers<pointer>` to different sized objects.
+
+D
+-
+.. _Data Layer:
+
+Data Layer
+    A group of bytes, where each byte can be addressed by a specific offset.  Data layers are usually contiguous
+    chunks of data.
+
+.. _Dereference:
+
+Dereference
+    The act of taking the value of a pointer, and using it as an offset to another object, as a reference.
+
+.. _Domain:
+
+Domain
+    This the grouping for input values for a mapping or mathematical function.
+
+M
+-
+.. _Map:
+
+Map, mapping
+    A mapping is a relationship between two values, where one value (the :ref:`Domain<domain>` maps to the :ref:`Range<range>` value).
+    Mappings can be seen as a mathematical function, and therefore volatility 3 attempts to use mathematical functional
+    notation where possible.
+
+.. _Member:
+
+Member
+    The name of subcomponents of a type, similar to attributes of objects in common programming parlance.  These
+    are usually recorded as :ref:`offset<offset>` and :ref:`type<type>` pairs within a :ref:`structure<struct>`.
+
+O
+-
+.. _Object:
+
+Object
+    This has a specific meaning within computer programming (as in Object Oriented Programming), but within the world
+    of Volatility it is used to refer to a type that has been associated with a chunk of data.  See all :ref:`Type<type>`.
+
+.. _Offset:
+
+Offset
+    A numeric value that identifies a distance within a group of bytes, to uniquely identify a single byte, or the
+    start of a run of bytes.  This is often relative (offset from another object/item) but can be absolute (offset from
+    the start of a region of data).
+
+P
+-
+.. _Packed:
+
+Packed
+    Structures are often :ref:`aligned<alignment>` meaning that the various members (subtypes) are always aligned at
+    particular values (usually multiples of 2, 4 or 8).  Thus if a particular value is an odd number of bytes, the
+    next chunk of data containing useful information would start at an even offset, and a single byte of
+    :ref:`padding<padding>` would be used to ensure appropriate :ref:`alignment<alignment>`.  In packed structures, no
+    padding is used, and offsets may be at odd offsets.
+
+.. _Padding:
+
+Padding
+    Data that (usually) contains no useful information.  The typical value used for padding is 0, so should a string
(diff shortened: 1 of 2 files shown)


ikelos:
And yes, I was bored, sorry...  5:P

NiklasBeierl:
Hello,
I am currently diving through volatilitys docs and code as part of my bachelor thesis.
Would you mind if I add my thoughts as a review to this PR? I didn't dare to just go ahead and do it, since I am not in any way a maintainer. :innocent:

ikelos:
@NiklasBeierl, that's ok, this a community project, we'd be happy to get more eyes on the work we're doing and help make it better, so please feel free!  5:)

doc/source/glossary.rst
```diff
@@ -0,0 +1,179 @@
...
+
+Domain
+    This the grouping for input values for a mapping or mathematical function.
+
+M
+-
+.. _Map:
+
+Map, mapping
+    A mapping is a relationship between two values, where one value (the :ref:`Domain<domain>` maps to the :ref:`Range<range>` value).
```
NiklasBeierl: Suggestion: 
> ...is a relationship between two sets. Where elements of the Domain map to the elements of the Range.
Compare: https://en.wikipedia.org/wiki/Function_(mathematics) (Found under: https://en.wikipedia.org/wiki/Mapping)

doc/source/glossary.rst
```diff
@@ -0,0 +1,179 @@
...
+    The name of subcomponents of a type, similar to attributes of objects in common programming parlance.  These
+    are usually recorded as :ref:`offset<offset>` and :ref:`type<type>` pairs within a :ref:`structure<struct>`.
+
+O
+-
+.. _Object:
+
+Object
+    This has a specific meaning within computer programming (as in Object Oriented Programming), but within the world
+    of Volatility it is used to refer to a type that has been associated with a chunk of data.  See all :ref:`Type<type>`.
```
NiklasBeierl: Can I think of it like an "Instance" of a type? If so, that might be a helpfull addition.

doc/source/glossary.rst
```diff
@@ -0,0 +1,179 @@
...
+
+.. _Offset:
+
+Offset
+    A numeric value that identifies a distance within a group of bytes, to uniquely identify a single byte, or the
+    start of a run of bytes.  This is often relative (offset from another object/item) but can be absolute (offset from
```
NiklasBeierl: Subjective: "This" seems confusing here.
I'd go with: "**An offset** is often relative (to another object / item)"

doc/source/glossary.rst
```diff
@@ -0,0 +1,179 @@
...
+    the start of a region of data).
+
+P
+-
+.. _Packed:
+
+Packed
+    Structures are often :ref:`aligned<alignment>` meaning that the various members (subtypes) are always aligned at
+    particular values (usually multiples of 2, 4 or 8).  Thus if a particular value is an odd number of bytes, the
+    next chunk of data containing useful information would start at an even offset, and a single byte of
```
NiklasBeierl: Its not so much about the value itself but rather about the length of the value, no?

>  Structures are often :ref:`aligned<alignment>` meaning that the various members (subtypes) have an offset that is a multiple of some number (usually multiples of 2, 4 or 8). Thus, if a particular member has a length which is not a multiple of the chosen number, there will be padding between it and the next member. In packed structs, no padding is used and the offset of the next member depends on the length of the previous one.

doc/source/glossary.rst
```diff
@@ -0,0 +1,179 @@
...
+    :ref:`padding<padding>` would be used to ensure appropriate :ref:`alignment<alignment>`.  In packed structures, no
+    padding is used, and offsets may be at odd offsets.
+
+.. _Padding:
+
+Padding
+    Data that (usually) contains no useful information.  The typical value used for padding is 0, so should a string
```
NiklasBeierl: The second sentence has a bit of an abrupt transition into an example. I think the reader could use something like: 

"A typical example is a string object....."

doc/source/glossary.rst
```diff
@@ -0,0 +1,179 @@
...
+    and page tables can point to pages that are in fact other page tables.  Using tables that point to tables, it's
+    possible to use them as a way to map a particular address within a (potentially larger, but sparsely populated)
+    virtual space to a concrete (and usually contiguous) physical space, through the process of :ref:`mapping<map>`.
+
+.. _Pointer:
+
+Pointer
+    A value within memory that points to a different area of memory.  This allows objects to contain references to
+    other objects without containing all the data of the other object.  Following a pointer is known as :ref:`dereferencing<dereference>`
+    a pointer.  Pointers are usually as large as the size of the
```
NiklasBeierl: sentence is not completed.

doc/source/glossary.rst
```diff
@@ -0,0 +1,179 @@
...
+
+R
+-
+.. _Range:
+
+Range
+    This is the grouping the output values for a mapping or mathematical function.
```
NiklasBeierl: grouping vs set is probably a matter of tast. Whatever we choose, it should be consistent with the phrases used in "Map".

doc/source/glossary.rst
```diff
@@ -0,0 +1,179 @@
...
+
+S
+-
+.. _Struct:
+
+Struct, Structure
+    A means of containing multiple different :ref:`type<types>` associated together.  A struct typically contains
+    other :ref:`type<types>`, one directly after another (unless :ref:`packing<packed>` is involved).  In this way
```
NiklasBeierl: Isn't `directly after another` exactly what packing does?

doc/source/glossary.rst
```diff
@@ -0,0 +1,179 @@
...
+    the :ref:`members<member>` of a type can be accessed by finding the data at the relative :ref:`offset<offset>` to
+    the start of the structure.
+
+.. _Symbol:
+
+Symbol
+    This is used in many different contexts, as short term for many things.  A symbol is a construct that usually
```
NiklasBeierl: Perhaps: 
> as **a** short term for many things. 

> **In the Context of Volatility**, a symbol is usually a (con)struct encompassing ...

Might deserve a further reading linl? https://en.wikipedia.org/wiki/Symbol_table

NiklasBeierl (changes requested):
First round. I still need to go over T and U.
Overall this would have been very helpful when I first go started on the documentation. :+1:

doc/source/glossary.rst
ikelos: Yeah, I tend to see it more in terms of an input and an output, but mathematically you're right it's a relationship between two sets.  Updated to reflect this.  5:)

doc/source/glossary.rst
ikelos: Yep, added to provide this as an option for people who prefer the concept.  5:)

doc/source/glossary.rst
ikelos: I agree, updated.  5:)

doc/source/glossary.rst
ikelos: Much more nicely worded, thanks!  5:D

doc/source/glossary.rst
ikelos: Yep, separated the typical value bit from the example a little more.

doc/source/glossary.rst
ikelos: Hehehe, good catch, thanks!  5;)

doc/source/glossary.rst
ikelos: Yeah, I'm ok with set given that's what we use earlier.

doc/source/glossary.rst
ikelos: It is, I've reworded it a little.  I can't tell whether structs are typically packed or not, but hopefully my correction clarifies that alignment and packing are somewhat opposite ends of the spectrum.

doc/source/glossary.rst
ikelos: Yep, I think those changes are enough, I didn't intend to describe formal definitions for everything, just give people an idea of how it's used in volatility (from my own understanding, which may be wrong, but this was at least a start).  5;)

ikelos:
Are people generally happy with this?  There's no rush, but we can also do it in two batches, so just lemme know if there's reason to keep it out of develop, otherwise I'll probably push it in a week or so...  5:)

doc/source/glossary.rst
```diff
@@ -0,0 +1,188 @@
...
+
+Template
+    Within volatility 3, the term template applies to a :ref:`type<Type>` that has not yet been instantiated or linked
+    to any data or a specific location within memory.  Once a type has been tied to a particular chunk of data, it is
+    called an :ref:`object<Object>`.
+
+.. _Translation Layer:
+
+Translation Layer
+    This is a specific type of :ref:`data layer<Data Layer>`, a non-contiguous group of bytes that can be references by
```
NiklasBeierl: I am not sure about the term "linear" here. I guess a one-to-one mapping would be a bijective mapping. And the other one non-bijective. When I hear "linear" I think of "constant offset", like the linux kernel does for kernel / userspace separation, but even that gets me weird looks from my friends in the math dept. :laughing:

Suggestion:
This is a type of data layer which allows accessing data from lower layers using addresses other than the addresses used by the lower layers themselves. When accessing data in a translation layer, it translates (or :ref:`maps<Map>`) addresses from its own :ref`address space<Address Space>` to the address space of the lower layer and returns the corresponding data from the lower layer. Note that multiple addresses in the higher layer might refer to the same data in the lower layer. Inversely, some addresses in the higher layer might have no corresponding address in the lower layer at all. Most prominently, translation layers handles the translation from virtual to physical addresses. (For more information on that, see: `https://lwn.net/Articles/253361/`)

We might also cosnider adding that link to section "address" instead.

doc/source/glossary.rst
```diff
@@ -0,0 +1,188 @@
...
+    a unique :ref:`offset<Offset>` within the layer.  In particular, translation layers translates (or :ref:`maps<Map>`)
+    requests made of it to a location within a lower layer.  This can be either linear (a one-to-one mapping between bytes)
+    or non-linear (a group of bytes :ref:`maps<Map>` to a larger or smaller group of bytes.
+
+.. _Type:
+
+Type
+    This is a structure definition of multiple elements that expresses how data is laid out.  Basic types define how
+    the data should be interpretted in terms of a run of bits (or more commonly a collection of 8 bits at a time,
+    called bytes).  More complex types can be made up of other types combined together at specific locations known
```
NiklasBeierl: Consider: `specific offsets` instead of `specifc locations`. 
Sentence structure made me trip, suggesting:

New types can be made up by combining other types at specific locations, known as structs, or by repeating the same type, known as arrays. 

Next sentence is missing some words:
They can even be defined as multiple types at the same location, depending on .....

doc/source/glossary.rst
```diff
@@ -0,0 +1,188 @@
...
+    as :ref:`structs<Struct>` or repeated, known as :ref:`array<Array>`.  They can even defined types at the same
+    location depending on the data itself, known as :ref:`Unions<Union>`.  Once a type has been linked to a specific
+    chunk of data, the result is referred to as an :ref:`object<object>`.
+
+U
+-
+.. _Union:
+
+Union
+    A union is a type that can have can hold multiple different subtypes, which specifically overlap.  A union is means
```
NiklasBeierl: Suggesting: 
In essence, a union is an alias for multiple types. Just like a struct, it has multiple members, but the bytes assigned to a union will only ever hold the value of one member, the one that was last written to. Depending on what data a union is holding, only a subset of its members will point to accurate data. (Assuming the data wasn't corrupted in the first place) To illustrate: Picture a box that can only ever have one thing inside. The box is suitable for a set of items (the union members): Apple, melon and cherry. The box will be big enough to hold a melon, but even though multiple apples and a lot of cherries could fit in the spot of one melon, our union will only ever hold a single apple, cherry or melon.

NiklasBeierl (changes requested):
Added suggestions for T and U.

NiklasBeierl:
> Are people generally happy with this? There's no rush, but we can also do it in two batches, so just lemme know if there's reason to keep it out of develop, otherwise I'll probably push it in a week or so... 5:)

Sorry for letting this sit so long, holidays got in the way. :man_shrugging:

doc/source/glossary.rst
ikelos: Yeah, I like that, the reason it's linear is because in general `<size of function input> == <size of function output>` which isn't true for compression (for example).  With compression a chunk of 1024 bytes might map to a region that's say 500 bytes, and that's what I class as non-linear.  If there's a better term for it, I'm happy to go for it.  Generally `translate(input) + c = translate(input + c)` assuming `c` stays within the contiguous block that `input` is a part of (for example a page). I'm not sure an Associative Layer would give the right impression on hearing it?  5:S

doc/source/glossary.rst
ikelos: Mmmm, I'm not keen on the illustration, here.  It doesn't feel as though it clarifies matters and I also don't want to potentially talk down to people.  I've added a rewording that hopefully gets across what we want to communicate better without having to resort to metaphors or similes which don't quite fit...

ikelos:
No problem, I just appreciate the help!  5:)

ikelos:
Ok, so I may merge the changes we've got into develop, and then we can always tweak it with further changes down the line...  5:)  Please don't feel as though this is the end of improving the documentation, it's just the beginning!  5:D

PR #303, opened 2020-08-18
Lsadump

jxwenger:


ikelos:
Thanks for this, I'm afraid it still won't apply cleanly.  Did you try the link I posted in the previous pull request, about rebasing?  Please get in touch to discuss how to get this into the tree, I'd sooner not keep getting submissions with the same problem without actually addressing it?  I'm not going to review this until it would apply cleanly to master I'm afraid...  5:S

jxwenger:
I did the rebase and it says there are no merge conflicts so I'm not sure what else to do.

ikelos:
Ok, no problem, so just to check the steps, I'll run through them here and we can see if that matches up with what you've done:

First up, we're going to ensure we're using the original volatility3 repository for our rebase:
```
git remote add vol3 https://github.com/volatilityfoundation/volatility3
```
This now lets you refer directly to the volatility3 repository, meaning any previous changes you've made in your local master won't come into play.  Now we'll swap over to your lsadump branch.
```
git checkout lsadump
```
Finally you can rebase against the vol3 repository master by doing
```
git rebase vol3/master
```
This should then try to replay your changes on top of the old history.  Since you're mostly editing a new file, it shouldn't have any trouble, but it may stop and tell you there was a conflict.  If that happens, you'll need to fix the conflict (using something like `git mergetool`) and after that `git rebase --continue`.  Once the rebase is complete, you can push the new state of your branch up to github by going `git push -f`.  It needs the `-f` to force it, because rebasing rewrites history.

If all of that goes ahead without problems, then we can check and see if the branch can be merged.  If you run into problems, come back here (or the '#vol3-dev` channel) and we can try sort anything that crops up out.  5:)  We'll get there, don't worry, git's always a bit finicky, but I'm sure we can tame it...  5:)

jxwenger:
When I try to run git rebase vol3/master I get the error: fatal: invalid upstream 'vol3/master' even though it says remote vol3 already exists

ikelos:
Thanks for giving that a go!  It turns out that was my bad, I missed a step, sorry.  5:S
To get all the information about the vol3 repo, you'll need to do:
```
git fetch vol3
```
before the rebase step.  Then you should be able to do the rebase after that.

ikelos:
Yep, looks much better, thanks  5:)

Issue #304, opened 2020-08-19
MMVAD cannot find children


slyles1001:
In windows 10, there is another layer between MMVAD and children. When `volatility.framework.symbols.windows.extensions.MMVAD` inherits from `MMVAD_SHORT`, the `find_child` functions cannot find a member of the nested objects.

**Context**
Volatility Version:  commit 73f3c217bd0b752ac43d576605bb1931d8ec8a5d
Operating System:  WSL
Python Version:  3.6.9
Suspected Operating System:  Windows 10

**To Reproduce**
call `get_left_child()` on MMVAD object

**Expected behavior**
returns left child address

slyles1001:
I'm happy to work on this, i think the fix will be quick. The `get_parent` function in `MMVAD_SHORT` already has the `VadNode` or `Core` checks.

ikelos:
Ok, cool, please feel free to and submit it as a pull request.  5:)  We'll get @iMHLv2 to look over it, as our resident MMVAD expert...  5;)

iMHLv2:
@slyles1001 We believe this might be an issue with the volshell tool. Going back to commit https://github.com/volatilityfoundation/volatility3/commit/73f3c217bd0b752ac43d576605bb1931d8ec8a5d, are you able to run the vadinfo plugin against your Windows 10 memory sample? In particular, your screen shots from https://github.com/volatilityfoundation/volatility3/pull/305 show `ps()[-1]` so perhaps you could print the process ID of that process and then use `vadinfo --pid <PID>` to see how it goes? Also, do you have any more specific version information for the Windows 10 memory sample? For example, the build number (i.e., 10.0.17763.x) or the GUID that volatility3 identifies (can be seen in the output of vol.py -vvvvvvv)?

slyles1001:
Hey @iMHLv2, I'm using Windows 10 (10.0.10586 Build 10586) in virtualbox for the dump. I was having trouble in a plugin i'm writing and was able to reproduce in the volshell.

The plugin has this as content:
```python
            for vad in [v for v in proc.get_vad_root().traverse()]:
                typ = "None"
                fil = "None"
                if vad.get_start() in heaps:
                    typ = "heap"
                elif vad.get_start() in stacks:
                    typ = "stack"
                elif vad.get_start() in mods:
                    typ = "DLL"
                    fil = mods[vad.get_start()]
                else:
                    fil = vad.get_file_name()
                    if type(fil) != renderers.NotApplicableValue:
[...]

PR #305, opened 2020-08-19
Bugfix/304

slyles1001:
#304 
Before:
image

After:
image

### DIFF
--- a/volatility/framework/symbols/windows/extensions/__init__.py
+++ b/volatility/framework/symbols/windows/extensions/__init__.py
@@ -122,14 +122,15 @@ class MMVAD_SHORT(objects.StructType):
         elif self.has_member("Right"):
             return self.Right
 
+        # this is for windows 8 and 10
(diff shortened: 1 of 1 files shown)


ikelos:
We finally figured out why this was happening.  The reason is because the `traverse` method casts the type `MMVAD_SHORT` based on the tags that the vad has.  What this means is that `RTL_BALANCED_NODE`s becomes `MMVAD_SHORT`.  Whilst this has the same class override, the structure/offsets are different, and the various child attributes aren't present in `MMVAD_SHORT`, even though they are present in `RTL_BALANCED_NODE`.  We think it's good (given that it's possible) to make the `MMVAD_SHORT` more capable and be able to return the children even without the direct attributes, so this patch helps achieve that.  Hopefullyt his explains what's going on, and is somewhat a long-winded way of saying, yep, we'll merge this, thanks!  5:D

Issue #306, opened 2020-08-20
poolscanner page_type checks for vista and later


superponible:
**Describe the bug**
The current page_type checks for the poolscanner work on pre-vista.  Vista and later change the value of _POOL_HEADER.PoolType for paged and nonpaged.

Not currently evident in results from a plugin like psscan because all the builtins are currently checking all 3 page types.  Vol2, for example, only checked nonpaged and free for EPROCESS.  The netscan PR #226 was submitted only checking nonpaged and free as well which is where the issue was encountered.

PR #307, opened 2020-08-20
issue #306 - _POOL_HEADER class based on OS

superponible:
@ikelos Let me know how this looks to fix the issue netscan #226 is waiting on.  It set the type class for _POOL_HEADER based on OS version.  Then defines some methods in the two pre and post Vista classes and uses those methods in the page_type check.

### DIFF
--- a/volatility/framework/plugins/windows/poolscanner.py
+++ b/volatility/framework/plugins/windows/poolscanner.py
@@ -85,11 +85,11 @@ class PoolHeaderScanner(interfaces.layers.ScannerInterface):
                 if constraint.page_type is not None:
                     checks_pass = False
 
-                    if (constraint.page_type & PoolType.FREE) and header.PoolType == 0:
+                    if (constraint.page_type & PoolType.FREE) and header.is_free_pool():
                         checks_pass = True
-                    elif (constraint.page_type & PoolType.PAGED) and header.PoolType % 2 == 0 and header.PoolType > 0:
+                    elif (constraint.page_type & PoolType.NONPAGED) and header.is_nonpaged_pool():
                         checks_pass = True
-                    elif (constraint.page_type & PoolType.NONPAGED) and header.PoolType % 2 == 1:
+                    elif (constraint.page_type & PoolType.PAGED) and header.is_paged_pool:
                         checks_pass = True
 
                     if not checks_pass:
@@ -201,6 +201,9 @@ class PoolScanner(plugins.PluginInterface):
     is_windows_7 = os_distinguisher(version_check = lambda x: x == (6, 1),
                                     fallback_checks = [("_OBJECT_HEADER", "TypeIndex", True),
                                                        ("_HANDLE_TABLE", "HandleCount", True)])
+
+    is_vista_or_later = os_distinguisher(version_check = lambda x: x >= (6, 0),
+                                         fallback_checks = [("KdCopyDataBlock", None, True)])
 
     def _generator(self):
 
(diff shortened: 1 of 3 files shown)


volatility/framework/symbols/windows/extensions/pool.py
```diff
@@ -165,6 +165,29 @@ def _calculate_optional_header_lengths(cls, context: interfaces.context.ContextI
...
+
+    Exists at the base of the allocation and provides a tag that we can
+    scan for.
+    """
+
+    def is_paged_pool(self):
+        return self.PoolType % 2 == 1
+
+    def is_nonpaged_pool(self):
+        return self.PoolType % 2 == 0 and self.PoolType > 0
```
japhlange, line 189: Asking to understand: Why the check for `> 0`? According to Kernel Pool Exploitation on Windows 7 by Tarjei Mandt (https://www.exploit-db.com/docs/english/16032-kernel-pool-exploitation-on-windows-7.pdf, page 6) the type for `NonPagedPool` appears to be 0.
```C
typedef enum _POOL_TYPE{
    NonPagedPool = 0 /*0x0*/,
    <...>
}
```
Is it because of the line earlier that only unused/free chunks have a type of 0? Is it not possible for data to reside in a chunk with a type of 0 (maybe as residue?)?

ikelos: So a question I'd have is are "NonPagedPool"s also "FreePool"s, since `is_free_pool` would return True is `PoolType` were 0?  This may be why they inverted them (so they could coalesce the NonPaged and Free pool types)?

volatility/framework/plugins/windows/poolscanner.py
```diff
@@ -445,13 +448,19 @@ def _get_pool_header_module(cls, context, layer_name, symbol_table):
...
+                class_type = extensions.pool.POOL_HEADER
+
             new_table_name = intermed.IntermediateSymbolTable.create(
                 context = context,
                 config_path = configuration.path_join(context.symbol_space[symbol_table].config_path, "poolheader"),
                 sub_path = "windows",
                 filename = pool_header_json_filename,
                 table_mapping = {'nt_symbols': symbol_table},
-                class_types = {'_POOL_HEADER': extensions.pool.POOL_HEADER})
+                class_types = {'_POOL_HEADER': class_type})
```
ikelos, line 463: This seems like it only picks up one very specific place that this is used (poolscanner) and the logic wouldn't be applied to normal windows symbol tables.  I think the logic should also be applied at https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/__init__.py#L39 which means mebbe moving the decision to a function that lives somewhere under `framework/extensions/windows` (mebbe `extensions/pool.py`?) and then calling it from the poolscanner...

ikelos (changes requested):
All seems fine, but I'd prefer the decision logic weren't solely in poolscanner, but were applied to all windows tables (taking note that a very specific number don't have that structure at all).

volatility/framework/plugins/windows/poolscanner.py
ikelos: Either that or writing the class to do the decision when it runs, but I think that'll cause more issues in the longer term...

volatility/framework/symbols/windows/extensions/pool.py
japhlange: Well, the constraints defined by netscan2 in https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/netscan.py#L47 (migrated to netscan3 in https://github.com/japhlange/volatility3/blob/netscan/volatility/framework/plugins/windows/netscan.py#L60) suggest that a `PoolType` of Free or NonPaged is valid. In that case as long as the possible network object's `Type` is even this constraint check passes.
Every object I've checked the type for seems to have a Type of 2, but of course I have not specifically searched for that value, so I only have sparse evidence. But it might be reasonable to assume this to be the case, maybe. :shrug:

superponible: The PoolType value in the _POOL_HEADER is dual purpose, saying if it's in use or not and if it's paged or nonpaged.

It does say in that PDF in the section right before that enum is listed, bottom of the previous page:

> If a chunk is free, PoolType is set to zero

Then explains the bitmask that's applied to the enum value when the pool is allocated, which explains part of the difference in values pre and post Vista. But I believe on xp/2k3, the PoolType is also incremented by 1, which is why the check here changes across versions.

As an example, if you print the PoolType value for results from psscan and compare XP with later results, the XP results will have a lot with a value of 5 (which is 0 for NonPaged or'd with 4 + 1), whereas on Vista or later, you'll see values of 2 (just 0 for NonPaged or'd with 2).

Also see page 10 and 11 https://www.imf-conference.org/imf2006/23_Schuster-PoolAllocations.pdf mentioning the PoolType being increased by 1 and the even/odd values for PoolType.

volatility/framework/plugins/windows/poolscanner.py
superponible: Yeah, that makes sense.  This only happens in the backup case when the symbols aren't found in the kernel pdb.  Not sure if we can determine the OS version at the point in the code? Will look later.

volatility/framework/symbols/windows/extensions/pool.py
japhlange: Thank you for the explanation and the source, I've been looking for something like that but didn't find it :)

PR #311, opened 2020-08-22
Move os distinguisher

ikelos:
As a follow up to a previous bug, this shifts all (windows) os distinguishers (since they currently check the PE metadata, which is windows specific) to a winver module.  This will allow us to create a descendant in the future should we need to (since the version_check callable should be generic) should we wish to create similar checks for linux/mac in the future...  It also ensures we use the same checks and that they're reliable, rather than having checks defined per-plugin unnecessarily...

This is fairly low contention, but I'll give it 7 days so people have a chance to review it if they want...


ikelos:
Refactored to versions as suggested by @npetroni , still going to merge in 3 days unless someone shouts.

PR #314, opened 2020-08-25
CmdLine windows plugin: call on a single process added

cecio:
Added method "get_cmdline" to be able to call the plugin on a single process.

The general plugin behavior is exactly the same, just moved the "query" part in a
single method, callable from outside

### DIFF
--- a/volatility/framework/plugins/windows/cmdline.py
+++ b/volatility/framework/plugins/windows/cmdline.py
@@ -15,6 +15,8 @@ vollog = logging.getLogger(__name__)
 class CmdLine(interfaces.plugins.PluginInterface):
     """Lists process command line arguments."""
 
+    _version = (1, 0, 0)
+
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
         # Since we're calling the plugin, make sure we have the plugin's requirements
@@ -59,9 +61,10 @@ class CmdLine(interfaces.plugins.PluginInterface):
 
         for proc in procs:
             process_name = utility.array_to_string(proc.ImageFileName)
-            proc_id = proc.UniqueProcessId
+            proc_id = "Unknown"
 
             try:
+                proc_id = proc.UniqueProcessId
                 result_text = self.get_cmdline(self.context, self.config["nt_symbols"], proc)
 
             except exceptions.SwappedInvalidAddressException as exp:
--- a/volatility/framework/plugins/windows/cmdline.py
+++ b/volatility/framework/plugins/windows/cmdline.py
@@ -31,7 +31,7 @@ class CmdLine(interfaces.plugins.PluginInterface):
         ]
 
     @classmethod
-    def get_cmdline(self, context: interfaces.context.ContextInterface, 
+    def get_cmdline(cls, context: interfaces.context.ContextInterface, 
                     kernel_table_name: str, proc):
         """Extracts the cmdline from PEB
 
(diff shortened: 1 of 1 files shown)


volatility/framework/plugins/windows/cmdline.py
```diff
@@ -30,35 +30,54 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                                          optional = True)
         ]
 
+    @classmethod
+    def get_cmdline(self, context: interfaces.context.ContextInterface,
```
ikelos, line 34: Self should be cls for classmethods.

volatility/framework/plugins/windows/cmdline.py
```diff
@@ -30,35 +30,54 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+            A string with the command line
+        """
+        try:
+            proc_id = proc.UniqueProcessId
+            proc_layer_name = proc.add_process_layer()
+
+        except exceptions.InvalidAddressException as excp:
+            vollog.debug("Process {}: invalid address {} in layer {}".format(proc_id, excp.invalid_address,
+                                                                             excp.layer_name))
+            return "Invalid address for PID"
```
ikelos, line 53: It's difficult to distinguish this from having had this typed in the commandline.  Ideally the classmethod would throw the exception and the generator (whatever called it) would do the catching and provide the appropriate response.  In the previous plugin, the generator was what was interpreting the exceptions.

ikelos (changes requested):
Thanks, this looks good, but we'd prefer that classmethods (essentially APIs) don't handle errors themselves, but allow the caller to handle them, so I'd prefer the exception handling be pushed back to the generator please.  The method may end up looking pretty sparse, but that's fine.

volatility/framework/plugins/windows/cmdline.py
cecio: ok, thanks for the suggestions. I did another commit with the modification. For this specific part I followed how it has been done in procdump.py

volatility/framework/plugins/windows/cmdline.py
```diff
@@ -30,25 +30,39 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
             process_name = utility.array_to_string(proc.ImageFileName)
-            proc_id = "Unknown"
-            try:
-                proc_id = proc.UniqueProcessId
-                proc_layer_name = proc.add_process_layer()
-            except exceptions.InvalidAddressException as excp:
-                vollog.debug("Process {}: invalid address {} in layer {}".format(proc_id, excp.invalid_address,
-                                                                                 excp.layer_name))
-                continue
+            proc_id = proc.UniqueProcessId
```
ikelos, line 62: This may end up throwing the exception we want to use for the exception message!  5:P  Better to set it to "Unknown" and then try to set it in the try/except block.

volatility/framework/plugins/windows/cmdline.py
```diff
@@ -30,25 +30,39 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                                          optional = True)
         ]
 
+    @classmethod
```
ikelos, line 33: Also, I realized I forgot one thing.  Now that this is a classmethod and so exposes an API, the class should define a:
```
_version = (1, 0, 0)
```
So that people can add a plugin requirement against it...

ikelos (commented):
Much better, and very nearly there, thanks!  Something I forgot (the version) and getting the proc_id may cause the exception it's there to inform about, so best to shield it in a try/except and set a default value before that.  Then it should be good to go in.  5:)

volatility/framework/plugins/windows/cmdline.py
cecio: ok, pushed mods. Sorry for the back and forth: next time I'll try to be more effective  :-)

ikelos (approved):
Yep, looks good, thanks!  5:D

PR #323, opened 2020-08-31
BUGFIX: C strings can be zero length. 

gcmoreira:
We should be able to read zero length strings without raising an exception. We can still check if the string is empty as we do for instance [here](https://github.com/volatilityfoundation/volatility3/blob/fe2b2ccaeb075d2c83d7f1bc8ae9fa0bb870ed68/volatility/framework/symbols/linux/__init__.py#L53)

### DIFF
--- a/volatility/framework/objects/utility.py
+++ b/volatility/framework/objects/utility.py
@@ -24,8 +24,8 @@ def pointer_to_string(pointer: 'objects.Pointer', count: int, errors: str = 'rep
     """Takes a volatility Pointer to characters and returns a string."""
     if not isinstance(pointer, objects.Pointer):
         raise TypeError("pointer_to_string takes a Pointer")
-    if count < 1:
-        raise ValueError("pointer_to_string requires a positive count")
+    if count < 0:
+        raise ValueError("pointer_to_string requires a positive or zero count")
     char = pointer.dereference()
     return char.cast("string", max_length = count, errors = errors)
 


ikelos:
`count` is intentionally optional, it's to put a cap on runaway strings, not to specify the length of the return string.

0-length C-Strings should still be null terminated (and therefore have a byte count of 1).  Getting a string from a pointer where you're saying its `max_length`/`count` must be 0, guarantees you'll get an empty string every time, and I don't really see the point of that (or testing for whether it's an empty string or not)?

I'm having difficulty seeing when passing 0 in a function like that would ever be useful?  It doesn't prohibit you from testing for the empty string on response, but the value passed to that function must be at least a byte.  Otherwise the semantics go weird, if you ask for a 0 byte string at an offset, does it matter if the offset's invalid, since the string would still be ""?  Whatever the problem is, this is not the solution, so I won't be merging it I'm afraid...  5:(

gcmoreira:
I'm not calling it with count=0, the other code is doing that. Check for instance [here](https://github.com/volatilityfoundation/volatility3/blob/fe2b2ccaeb075d2c83d7f1bc8ae9fa0bb870ed68/volatility/framework/symbols/linux/extensions/__init__.py#L366)
dentry calls qstr.name_as_str() which does the self.len which is zero, so
objects.utility.pointer_to_string(self.name, 0) ends up in a exception.

ikelos:
Ok, that sounds like a bug in our code then.  I don't think the solution is to allow 0 value checks, but instead fix the calling code.  Sounds like one for @atcuno if it's mostly in the linux/mac code.  Might be worth filing an issue, as I say, this PR is not the right solution to the problem...

atcuno:
@gcmoreira can you paste the full command line input/output from what this looks like on a plugin run? I am not understanding what the issue is exactly.

ikelos:
So that one example you chose is the only part in the whole codebase where `pointer_to_string` is called with a variable count value.  I think the most straightforward fix is to catch `ValueError` as well as `InvalidAddressException`.  Essentially, we construct a string from a pointer, and if a structure had a `len` value, we use that as the count/maxlen.  If that value is 0, then we trying to make a string without using any bytes, so I think we should either catch the situation where len is 0 in that call (either explicitly, or just catching the `ValueError` at https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/linux/extensions/__init__.py#L372

gcmoreira:
> @gcmoreira can you paste the full command line input/output from what this looks like on a plugin run? I am not understanding what the issue is exactly.

```python
>> d
<volatility.framework.symbols.linux.extensions.dentry object at 0x7f8a97c597c0>
special variables
function variables
class variables
d_child:<volatility.framework.symbols.linux.extensions.list_head object at 0x7f8a97559970>
d_flags:4194456
d_fsdata:281473172132864
d_hash:<volatility.framework.objects.StructType object at 0x7f8a97e06130>
d_iname:<volatility.framework.objects.Array object at 0x7f8a97fd7610>
d_inode:162923353590752
d_lockref:<volatility.framework.objects.StructType object at 0x7f8a979ba280>
d_lru:<volatility.framework.symbols.linux.extensions.list_head object at 0x7f8a972250a0>
d_name:<volatility.framework.symbols.linux.extensions.qstr object at 0x7f8a97fda430>
d_op:281473171323072
d_parent:162923382502912
d_sb:162924182161408
d_seq:<volatility.framework.objects.StructType object at 0x7f8a96f103d0>
d_subdirs:<volatility.framework.symbols.linux.extensions.list_head object at 0x7f8a96f10220>
d_time:0
d_u:<volatility.framework.objects.UnionType object at 0x7f8a96f10b20>
d_wait:162923382503040
vol:<volatility.framework.interfaces.objects.ReadOnlyMapping object at 0x7f8a97770eb0>
_abc_impl:<_abc_data object at 0x7f8aca084fc0>
_check_members:<bound method AggregateType._check_members of <class 'volatility.framework.symbols.linux.extensions.dentry'>>
_concrete_members:{'d_child': <volatility.framewor...a97559970>, 'd_flags': 4194456, 'd_fsdata': 281473172132864, 'd_hash': <volatility.framewor...a97e06130>, 'd_iname': <volatility.framewor...a97fd7610>, 'd_inode': 162923353590752, 'd_lockref': <volatility.framewor...a979ba280>, 'd_lru': <volatility.framewor...a972250a0>, 'd_name': <volatility.framewor...a97fda430>, 'd_op': 281473171323072, 'd_parent': 162923382502912, 'd_sb': 162924182161408, 'd_seq': <volatility.framewor...a96f103d0>, 'd_subdirs': <volatility.framewor...a96f10220>, ...}
_context:<volatility.framework.contexts.Context object at 0x7f8ac4004f70>
_vol:ChainMap({}, <volatility.framework.interfaces.objects.ObjectInformation object at 0x7f8a97c59d30>, {'type_name': 'vmlinux1!dentry', 'offset': 162923382502912}, {'size': 192, 'members': {'d_child': (144, <volatility.framework.objects.templates.ReferenceTemplate object at 0x7f8a980d5e80>), 'd_flags': (0, <volatility.framework.objects.templates.ObjectTemplate object at 0x7f8a980d5fd0>), 'd_fsdata': (120, <volatility.framework.objects.templates.ObjectTemplate object at 0x7f8a980d9190>), 'd_hash': (8, <volatility.framework.objects.templates.ReferenceTemplate object at 0x7f8a980d9310>), 'd_iname': (56, <volatility.framework.objects.templates.ObjectTemplate object at 0x7f8a980d9490>), 'd_inode': (48, <volatility.framework.objects.templates.ObjectTemplate object at 0x7f8a980d96d0>), 'd_lockref': (88, <volatility.framework.objects.templates.ReferenceTemplate object at 0x7f8a980d9850>), 'd_name': (32, <volatility.framework.objects.templates.ReferenceTemplate object at 0x7f8a980d9910>), 'd_op': (96, <volatility.framewor...

>> d.d_name
<volatility.framework.symbols.linux.extensions.qstr object at 0x7f8a97fda430>

>> d.d_name.len
0
```

atcuno:
@gcmoreira Thank you for the paste.

So it seems the len is 0 inside the structure, which from my reading of the code would return an empty Python string. Is it not? Or if it is returning an empty string is there something else you would expect?

ikelos:
So len is 0, but `pointer_to_string` won't accept a string capped at length 0, because it doesn't make much sense, so the `pointer_to_string` needs to either set max length to 1 (so it can contain the null byte, and thus truely by a 0 length string) or catch the `ValueError` exception, which will return "" anyway.  Either way is fine, but the solution is not to allow `pointer_to_string` to accept max string lengths of 0, because that's a none action, if you know the length is 0, then you know the string is ""...

gcmoreira:
Exactly, so the fix I think should be to catch the those zero length strings in qstr. It could be before or after calling pointer_to_string. I think before is better as exceptions, in terms of performance, are usually expensive in any language. Either way that won't happen often but it's still good practice.

ikelos:
Ok, this should have been resolved by `6c3176b`, but if it's not, this needs to be filed as a ticket rather than a closed pull request.  Exceptions are not particularly expensive in modern python, particularly when not run in the millions (which this won't be).  I've added one to avoid 0 length strings not including the null terminator (this could allow one extra character if the string is corrupt) and I've caught the exception if it's ever negative for some reason...  5;)

Issue #324, opened 2020-08-31
Config defaults aren't used for optional values within the framework


ikelos:
When a requirement is defined as `optional = True, default = <thing>` the configuration should contain `<thing>` rather than being empty.

ikelos:
It looks like these are currently populated by the UI, rather than by the framework (see https://github.com/volatilityfoundation/volatility3/blob/master/volatility/cli/__init__.py#L414).

Options for having this happen as part of the framework include:
* Changing every plugin to access `requirement.config_value` or some method instead (since the configuration and the requirements aren't technically tied together, this seems a bad move)
* Making sure that when `unsatisified` is called, it actually populates the configuration with the default if it's not present. This may require additional subpopulation (since I'm not sure we descend into optional requirements), and if the UI doesn't call `unsatisified`, then it's probably not doing its job, but it might cover most places this could be a problem.
* The third option is just pulling `populate_config` inside the framework in some way.

AsafEitani:
@ikelos getting an error on the fix:
```
Traceback (most recent call last):
  File "/home/dev/volatility3/vol.py", line 10, in <module>
    volatility.cli.main()
  File "/home/dev/volatility3/volatility/cli/__init__.py", line 515, in main
    CommandLine().run()
  File "/home/dev/volatility3/volatility/cli/__init__.py", line 299, in run
    constructed = plugins.construct_plugin(ctx, automagics, plugin, base_config_path, progress_callback, self)
  File "/home/dev/volatility3/volatility/framework/plugins/__init__.py", line 51, in construct_plugin
    constructed = plugin(context, plugin_config_path, progress_callback = progress_callback)
  File "/home/dev/volatility3/volatility/framework/interfaces/plugins.py", line 106, in __init__
    self.config[requirement.name] = requirement.default
  File "/home/dev/volatility3/volatility/framework/interfaces/configuration.py", line 150, in __setitem__
    self._setitem(key, value)
  File "/home/dev/volatility3/volatility/framework/interfaces/configuration.py", line 160, in _setitem
    self._data[key] = self._sanitize_value(value)
  File "/home/dev/volatility3/volatility/framework/interfaces/configuration.py", line 188, in _sanitize_value
    raise TypeError("Invalid type stored in configuration")
TypeError: Invalid type stored in configuration
```

ikelos:
[...]

PR #325, opened 2020-08-31
Added maxsize option to vadinfo as optional parameter, as in Vol2.

cecio:
Right now the default is set to 0 (no limit) as opposed to Vol2 where it is set to 1GB.

### DIFF
--- a/volatility/framework/plugins/windows/vadinfo.py
+++ b/volatility/framework/plugins/windows/vadinfo.py
@@ -34,6 +34,7 @@ class VadInfo(interfaces.plugins.PluginInterface):
     """Lists process memory ranges."""
 
     _version = (1, 1, 0)
+    MAXSIZE_DEFAULT = 0
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -60,7 +61,12 @@ class VadInfo(interfaces.plugins.PluginInterface):
                 requirements.BooleanRequirement(name = 'dump',
                                                 description = "Extract listed memory ranges",
                                                 default = False,
-                                                optional = True)
+                                                optional = True),
+                requirements.IntRequirement(name = 'maxsize',
+                                            description = "Maximum size for dumped VAD sections " \
+                                                          "(all the bigger sections will be ignored)",
+                                            default = cls.MAXSIZE_DEFAULT,
+                                            optional = True),
                 ]
 
     @classmethod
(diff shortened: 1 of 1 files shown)


volatility/framework/plugins/windows/vadinfo.py
```diff
@@ -102,17 +108,31 @@ def list_vads(cls, proc: interfaces.objects.ObjectInterface,
...
+            maxsize: Max size of VAD section (default MAXSIZE_DEFAULT)
 
         Returns:
             A FileInterface object containing the complete data for the process or None in the case of failure
         """
+
+        try:
+            vad_start = vad.get_start()
+            vad_end = vad.get_end()
+        except Exception as excp:
```
ikelos, line 128: Ideally is a very generic exception, and it would be much better to catch just the expected exceptions (so we don't catch things like KeyboardInterrupt, etc).

volatility/framework/plugins/windows/vadinfo.py
```diff
@@ -102,17 +108,31 @@ def list_vads(cls, proc: interfaces.objects.ObjectInterface,
...
+            vollog.debug("Unable to get VAD information")
+            return
+
+        if maxsize != 0 and (vad_end - vad_start) > maxsize:
```
ikelos, line 132: Just to be safe, I'd prefer we check maxsize is positive in case someone decides it might be fun to break things with negative numbers.

volatility/framework/plugins/windows/vadinfo.py
```diff
@@ -102,17 +108,31 @@ def list_vads(cls, proc: interfaces.objects.ObjectInterface,
 
     @classmethod
     def vad_dump(cls, context: interfaces.context.ContextInterface, proc: interfaces.objects.ObjectInterface,
-                 vad: interfaces.objects.ObjectInterface) -> interfaces.plugins.FileInterface:
+                 vad: interfaces.objects.ObjectInterface, 
+                 maxsize = MAXSIZE_DEFAULT) -> interfaces.plugins.FileInterface:
```
ikelos, line 112: I realize this wasn't in the original but should have been, however if we're fixing things up, the signature for this should be `-> Optional[interfaces.plugins.FileInterface]`...

ikelos (commented):
Thanks, looks good, just a couple minor things that could be slightly nicer...

volatility/framework/plugins/windows/vadinfo.py
cecio: I was applying the mods you suggested, and running the tests with vadinfo plugin, I had the same error reported in issue #324.
If you need, I also dumped the values you requested in the original issue, just before the exception:

```
key: address
value: None
```

ikelos: Thanks, could you please report that in bug #324, and make sure that you're testing it with commit `eb02ee1` or later.

cecio: just done. Tx

cecio: I'll wait resolution for #324  before committing the mods, otherwise I can not test them.

cecio: I did a commit including your suggestions.

ikelos (approved):
Looks good from my side, thanks!  5:)

ikelos:
Thanks, this pull request seems to have picked up the changes from the other pull request you added, I think they're based on the same branch?  Would you please be able to put those changes on a different branch?  It muddies this one (which had been reviewed) and the new one likely won't be able to be applied to master with this one applied, if they're both based off the same branch...

ikelos:
Oh, looks like this is on master, and the other is on a branch?  I wonder why commit 
`64d0c41` is listed in this PR?

cecio:
Yes, the other one is on a branch. This one on master....may be I did something wrong here. Should I re-do this one on a branch? In this case should I close this one?
Thanks,

ikelos:
It should be ok, this just needs the last commit (`64d0c41`) reverting and then it should be fine, and the other branch should still have the commit.  That one may need rebasing once this one's applied, but we can cross that bridge when we come to it...  5;)

ikelos:
So this branch will now need rebasing against volatility:master, I'm afraid.

cecio:
OK, just did the rebase...

ikelos (approved):

PR #326, opened 2020-09-01
Memoryforensics1 windows exstantions

memoryforensics1:


memoryforensics1:
add functionality to eprocess (environmet variables), file_object (access_string)
and a new structs: kthread, token.

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -666,3 +700,83 @@ def to_list(self,
...
+                    25: 'Spare6', 26: 'WrKernel', 27: 'WrResource', 28: 'WrPushLock', 29: 'WrMutex',
+                    30: 'WrQuantumEnd', 31: 'WrDispatchInt', 32: 'WrPreempted',
+                    33: 'WrYieldExecution', 34: 'WrFastMutex', 35: 'WrGuardedMutex',
+                    36: 'WrRundown', 37: 'MaximumWaitReason'}
+        waitReason = int(self.WaitReason)
+        stringWaitReason = ''
+        for flag in dictWaitReason:
+            if waitReason == flag:
+                stringWaitReason += '{} '.format(dictWaitReason[flag])
+
```
atcuno, line 781: @memoryforensics1 I am not sure I understand this loop over the hash table.

Since its a hash table, each integer key can only be present once, and it seems like your keys in 'dictWaitReason' are all you unique. Because of that, I am not sure why you loop over all of the element and attempt to allow for building of multiple values when only one could match.

I looked up the related Volatility 2 code and it seems to only just pull the one value:

https://gitlab.research.volexity.com/Volcano/volatility_volc/-/blob/master/volatility/plugins/malware/threads.py#L509

The code in your patch for get_state() does a similar pattern that I do not understand.

atcuno:
Thank you for this work @memoryforensics1 . I left one comment as I did understand the purpose of looping over the hash tables, but otherwise looks very clean!

volatility/framework/symbols/windows/extensions/__init__.py
memoryforensics1: My bad
I will close this and open another pull request

atcuno:
@memoryforensics1 you can commit changes to this one if easier. You don't have to make a new one for each change :)

ikelos (commented):
I agree, looks really good!  5:)  Since this is technically an addition to the framework, it should bump the `MINOR` version constant in `framework/constants/__init__.py`, so that plugins can known to depend upon it, but since we're still pre-release it doesn't technically matter.

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -608,6 +616,32 @@ def get_vad_root(self):
...
+        unicode strings. Each string cannot be more than 0x7FFF chars. 
+        End of the list is a quad-null. 
+        """
+        context = self._context
+        process_space = self.add_process_layer()
+
+        try:
+            block = self.get_peb().ProcessParameters.Environment
+            block_size = self.get_peb().ProcessParameters.EnvironmentSize
+            envars = context.layers[process_space].read(block, block_size).decode().split('\x00\x00\x00')[:-1]
```
ikelos, line 632: Could you explain why we're spliting on nulls, and why there's a decode applied to the bytes without an encoding (so `utf-8` decoding?  My guess would be that this should be `.decode('utf-16-le')` or `utf-16-be` or something, with an errors parameter (perhaps `replace`?), and then we can just split on the null character as normal?

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -608,6 +616,32 @@ def get_vad_root(self):
...
+        except exceptions.InvalidAddressException:
+            return renderers.UnreadableValue()
+
+        for envar in envars:
+            split_index = envar.find('=')
+            env = envar[:split_index].replace('\x00', '')
```
ikelos, line 638: This feels like a poor means to do unicode conversion, please can we not just strip all nulls, that's not how to read unicode (particularly as for character sets that aren't all english this will leave a mess of half-length characters).

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -608,6 +616,32 @@ def get_vad_root(self):
...
+            var = envar[split_index+1:].replace('\x00', '')
+
+            # Exlude parse problem with some types of env
+            if env!= '' and var != '':
```
ikelos, line 642: These can be written as `if env and var`, which is neater and copes with things like `None`.

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -666,3 +700,81 @@ def to_list(self,
...
+
+        userAndGroupCount = int(self.UserAndGroupCount)
+        if userAndGroupCount < 0xFFFF:
+            layer_name = self.vol.layer_name
+            kvo = self._context.layers[layer_name].config["kernel_virtual_offset"]        
+            symbol_table = self.get_symbol_table_name()
+            ntkrnlmp = self._context.module(symbol_table,
+                                      layer_name = layer_name,
+                                      offset = kvo)
+            UserAndGroups = ntkrnlmp.object(object_type="array",
```
ikelos, line 719: The spacing isn't consistent here, and suggests `yapf` has been run against this.  Please format your submissions with yapf, it saves us a lot of time having to reformat it ourselves...

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -666,3 +700,81 @@ def to_list(self,
...
+                                       offset=self.UserAndGroups.dereference().vol.get("offset") - kvo,
```
ikelos, line 720: `vol.get` is a bit of a waste, there's no default (meaning if it doesn't exist, you'll get back a `None` which won't subtract, and it cleaner just to write `.vol.offset`.

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -666,3 +700,81 @@ def to_list(self,
...
+                                       subtype = ntkrnlmp.get_type("_SID_AND_ATTRIBUTES"),
+                                       count=userAndGroupCount)
+            for saa in UserAndGroups:
```
ikelos, line 723: We don't need to save that much space, saa isn't as clear a variable name as `sid_and_attributes`.

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -666,3 +700,81 @@ def to_list(self,
...
+                sid = saa.Sid.dereference().cast("_SID")
+                 # catch invalid pointers (UserAndGroupCount is too high)
+                if sid == None:
```
ikelos, line 726: `None` equality test should is `if sid is None`, not `if sid == None`, since None is a singleton.

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -666,3 +700,81 @@ def to_list(self,
...
+                    raise StopIteration
+                # this mimics the windows API IsValidSid
+                if sid.Revision & 0xF != 1 or sid.SubAuthorityCount > 15:
+                    raise StopIteration
```
ikelos, line 730: In modern python StopIteration is deprecated, and should simply be a return within a Generator.

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -666,3 +700,81 @@ def to_list(self,
...
+                                               count= int(sid.SubAuthorityCount))
+                yield "S-" + "-".join(str(i) for i in (sid.Revision, id_auth) +
+                                      tuple(SubAuthority))
+            
+
+    def privileges(self):
+        "Return a list of privileges for the current token object."
+        for priv_index in range(64):
+            yield (priv_index,
+                   self.Privileges.Present & (2**priv_index) !=0,
```
ikelos, line 746: I don't think the `!=0` is necessary, at worst this should simply be wrapped in `bool()`

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -666,3 +700,81 @@ def to_list(self,
...
+
+class KTHREAD(objects.StructType):
+    """A class for thread control block objects."""
+        
+    def get_state(self) -> str:
+        dictState = {0:'Initialized', 1: 'Ready', 2: 'Running', 3: 'Standby', 4: 'Terminated',
+                    5: 'Waiting', 6: 'Transition', 7: 'DeferredReady', 8: 'GateWait'}
+        state = int(self.State)
+
+        if state in dictState:
```
ikelos, line 759: These three lines could probably be reduced to:
`return dictState.get(state, renderers.NotApplicableValue()`

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -666,3 +700,81 @@ def to_list(self,
...
+                   self.Privileges.EnabledByDefault & (2**priv_index) !=0)
+
+
+class KTHREAD(objects.StructType):
+    """A class for thread control block objects."""
+        
+    def get_state(self) -> str:
+        dictState = {0:'Initialized', 1: 'Ready', 2: 'Running', 3: 'Standby', 4: 'Terminated',
+                    5: 'Waiting', 6: 'Transition', 7: 'DeferredReady', 8: 'GateWait'}
+        state = int(self.State)
```
ikelos, line 757: `self.State` should already be an `int`, so I'm keen to know if this was necessary or just expected because of how volatility 2 worked?

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -666,3 +700,81 @@ def to_list(self,
...
+                    4: 'DelayExecution', 5: 'Suspended', 6: 'UserRequest', 7: 'WrExecutive',
+                    8: 'WrFreePage', 9: 'WrPageIn', 10: 'WrPoolAllocation', 11: 'WrDelayExecution',
+                    12: 'WrSuspended', 13: 'WrUserRequest', 14: 'WrEventPair', 15: 'WrQueue',
+                    16: 'WrLpcReceive', 17: 'WrLpcReply', 18: 'WrVirtualMemory', 19: 'WrPageOut',
+                    20: 'WrRendezvous', 21: 'Spare2', 22: 'Spare3', 23: 'Spare4', 24: 'Spare5',
+                    25: 'Spare6', 26: 'WrKernel', 27: 'WrResource', 28: 'WrPushLock', 29: 'WrMutex',
+                    30: 'WrQuantumEnd', 31: 'WrDispatchInt', 32: 'WrPreempted',
+                    33: 'WrYieldExecution', 34: 'WrFastMutex', 35: 'WrGuardedMutex',
+                    36: 'WrRundown', 37: 'MaximumWaitReason'}
+        waitReason = int(self.WaitReason)
```
ikelos, line 775: And again with casting to an `int`

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -666,3 +700,81 @@ def to_list(self,
...
+
+        if waitReason in dictWaitReason:
```
ikelos, line 777: The can probably be shortened the same way...

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -666,3 +700,81 @@ def to_list(self,
...
         return self.to_list(self.vol.parent.vol.type_name, self.vol.member_name)
+
+
+class TOKEN(objects.StructType):
+    """A class for process etoken object."""
+
+    def get_sids(self) -> Iterable[str]:
+        """Yield a sid for the current token object."""
+
+        userAndGroupCount = int(self.UserAndGroupCount)
```
ikelos, line 711: I don't think this needs casting as an int?  Is there a specific reason this was necessary?

ikelos (commented):
Looks like it'll be handy but a number of optimizations, some stylistic points and minor tidy ups to look at first if possible.

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -402,6 +400,25 @@ def owning_process(self, kernel_layer: str = None) -> interfaces.objects.ObjectI
...
+                                'PS_CROSS_THREAD_FLAGS_SYSTEM': 4,
+                                'PS_CROSS_THREAD_FLAGS_HARD_ERRORS_DISABLED': 5,
+                                'PS_CROSS_THREAD_FLAGS_BREAK_ON_TERMINATION': 6,
+                                'PS_CROSS_THREAD_FLAGS_SKIP_CREATION_MSG': 7,
+                                'PS_CROSS_THREAD_FLAGS_SKIP_TERMINATION_MSG': 8}
+        
+        flags = self.CrossThreadFlags
+        stringCrossThreadFlags = ''
+        for flag in dictCrossThreadFlags:
+            if flags & 2**dictCrossThreadFlags[flag]:
```
ikelos, line 417: We'd typically do `(1 << dictCrossThreadFlags[flag])`, but it's definitely not a blocker

ikelos:
Yep, looks good thanks for making the fixes!  5:)

PR #335, opened 2020-09-10
Change bad copy/descriptions for Mac that say Linux

atcuno:
A quick fix to strings that didn't make sense.


ikelos (approved):
Looks good

PR #336, opened 2020-09-11
Add new method for safe walking of generic list heads. Convert pslist…

atcuno:
… to use it.

### DIFF
--- a/volatility/framework/plugins/mac/pslist.py
+++ b/volatility/framework/plugins/mac/pslist.py
@@ -8,6 +8,7 @@ from typing import Callable, Iterable, List, Dict
 from volatility.framework import renderers, interfaces, contexts, exceptions
 from volatility.framework.configuration import requirements
 from volatility.framework.objects import utility
+from volatility.framework.symbols import mac
 
 vollog = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


volatility/framework/plugins/mac/pslist.py
```diff
@@ -258,41 +244,11 @@ def list_tasks_process_group(cls,
...
-                except exceptions.InvalidAddressException:
-                    break
-
-                seen_pg = set()
-                while p and p.vol.offset not in seen_pg:
-                    seen_pg.add(p.vol.offset)
-
-                    if p.is_readable() and not filter_func(p):
+            for pgrp in mac.MacUtilities.walk_list_head(proc_list, "pg_hash"):
+                for p in mac.MacUtilities.walk_list_head(pgrp.pg_members, "p_pglist"):
```
ikelos, line 248: Not entirely happy about just p as a variable name, but it's used immediately after, so I'm not gonna quibble about it too much, just point it out and mebbe look over the top of my glasses at you...  5;P

ikelos (approved):
Looks really good, I like it.  5:)

volatility/framework/symbols/mac/__init__.py
```diff
@@ -169,9 +171,30 @@ def walk_tailq(cls,
...
+    def walk_tailq(cls,
+                   queue: interfaces.objects.ObjectInterface,
+                   next_member: str,
+                   max_elements: int = 4096) -> Iterable[interfaces.objects.ObjectInterface]:
+ 
+        for element in cls._walk_iterable(queue, "tqh_first", "tqe_next", next_member, max_elements):
+            yield element
+
+    @classmethod
+    def walk_list_head(cls,
```
ikelos, line 192: Actually, I forgot this is an additive change and MacUtilities is versioned, so you'll need to bump the minor version to `(1, 1, 0)` on line 34, and the plugin that uses the `walk_list_head` (ie, the PsList plugin) needs to have a dependency on MacUtilities added (that needs `(1, 1, 0)` too...

ikelos (changes requested):
Sorry, just one little change to keep everything in order (but no problems with the code, just a forgotten thing).

volatility/framework/plugins/mac/pslist.py
frispete: @ikelos <off-topic/> I really like your style. This is exactly what distinguishes good projects from awesome projects.

atcuno: good call. Changed these

atcuno:
committed to address your feedback. Thanks for looking.

ikelos (approved):
Awesome, looks good, thanks!  5:)

PR #343, opened 2020-09-22
getsids getservicesids

memoryforensics1:
getsids, getservicesids, json file


ikelos (approved):
Cool, looks good, thanks very much!  5:D

PR #344, opened 2020-09-22
fix get_sids function

memoryforensics1:
fix get_sids function in broken images

### DIFF
--- a/volatility/framework/symbols/windows/extensions/__init__.py
+++ b/volatility/framework/symbols/windows/extensions/__init__.py
@@ -768,23 +768,26 @@ class TOKEN(objects.StructType):
                                             subtype = ntkrnlmp.get_type("_SID_AND_ATTRIBUTES"),
                                             count=self.UserAndGroupCount)
             for sid_and_attr in UserAndGroups:
-                sid = sid_and_attr.Sid.dereference().cast("_SID")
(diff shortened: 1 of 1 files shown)


volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -768,23 +768,26 @@ def get_sids(self) -> Iterable[str]:
...
+                    for i in sid.IdentifierAuthority.Value:
+                        id_auth = i
+                    SubAuthority = ntkrnlmp.object(object_type="array",
+                                                   offset=sid.SubAuthority.vol.offset - kvo,
+                                                   subtype = ntkrnlmp.get_type("unsigned long"),
+                                                   count= int(sid.SubAuthorityCount))
+                    yield "S-" + "-".join(str(i) for i in (sid.Revision, id_auth) +
+                                          tuple(SubAuthority))
+                except exceptions.InvalidAddressException:
+                    pass
```
ikelos, line 789: Ideally this would include a `vollog.debug()` message to explain that a SID wasn't found or that something went wrong...

ikelos (commented):
A trivial change, but good if we can give someone that needs to check information that it didn't go according to plan...

ikelos (approved):
Perfect, thanks!  5:D

PR #345, opened 2020-09-22
add envars plugin

memoryforensics1:
parse _EPROCESS environment variables

### DIFF
--- a/volatility/framework/plugins/windows/envars.py
+++ b/volatility/framework/plugins/windows/envars.py
@@ -1,0 +1,155 @@
+# This file is Copyright 2020 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+
+from typing import Callable, List, Generator, Iterable, Dict
+from volatility.framework import renderers, interfaces, objects, exceptions
+from volatility.framework.configuration import requirements
+from volatility.framework.objects import utility
+from volatility.plugins.windows import pslist
+from volatility.plugins.windows.registry import hivelist
+
+class Envars(interfaces.plugins.PluginInterface):
+    "Display process environment variables"
+
+    _version = (1, 0, 0)
+
+    @classmethod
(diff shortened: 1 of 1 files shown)


volatility/framework/plugins/windows/envars.py
```diff
@@ -0,0 +1,155 @@
...
+        # Since we're calling the plugin, make sure we have the plugin's requirements
+        return [requirements.TranslationLayerRequirement(name = 'primary',
+                                                         description = 'Memory layer for the kernel',
+                                                         architectures = ["Intel32", "Intel64"]),
+                requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
+                requirements.ListRequirement(name = 'pid',
+                                             description = 'Filter on specific process IDs',
+                                             element_type = int,
+                                             optional = True),
+                requirements.BooleanRequirement(name='SILENT',
```
ikelos, line 27: Looks good, I'm not sure SILENT needs to be all capitalized and might affect some UIs in the future so I'd recommend putting it to lowercase.  At some point we'll probably run `yapf` over this which will reformat it to include spaces around the equals signs in parameters, but that'll happen of its own accord.  5:)

volatility/framework/plugins/windows/envars.py
```diff
@@ -0,0 +1,155 @@
...
+                    pass
+            if sys:
+                try:
+                    for node in key.get_values():
+                        try:
+                            value_node_name = node.get_name()
+                            if value_node_name:
+                                values.append(value_node_name)
+                        except (exceptions.InvalidAddressException, RegistryFormatException) as excp:
+                            continue
```
ikelos, line 73: Again, when things happen we don't expect, it'd be good to give the user a way to find out about it (even if it's hidden in debug logs)  5:)

ikelos (commented):
Looks good, a couple tweaks if possible, but not major ones.

ikelos:
This was fixed by #349

PR #349, opened 2020-09-22
envars plugin

memoryforensics1:
parse environment variables


ikelos (approved):
Awesome, thanks for sticking with it!  5:)

Issue #333, opened 2020-09-09
windows.dlllist PagedInvalidAddressException Error


AsafEitani:
**Describe the bug**
Running `windows.dlllist` on dump causes an error due to a paged invalid address dll entry.
The error is popped up on the first occasion of reading data from the entry (getting the dll load time) but will also pop up on the final `yield` command (reading again from the invalid address).

**Context**
Volatility Version: Volatility 3 Framework 1.2.0-beta.1
Operating System: CentOS Linux 7 (Core)
Python Version: Python 3.8.3 (default, May 19 2020, 18:47:26)
Suspected Operating System: Windows Server 2016
Command: python /home/dev/volatility3/vol.py -vvv -r json -f dump.dd windows.dlllist

**To Reproduce**
Run dlllist on image - probably has one or more dlls in an invalid address.

**Expected behavior**
If the plugin can't access the dll information at least continue on reading the rest of the dlls and notify the user.

**Additional information**
Full -vvv output:
```
Volatility 3 Framework 1.2.0-beta.1
INFO     root        : Volatility plugins path: ['/home/dev/volatility3/volatility/plugins', '/home/dev/volatility3/volatility/framework/plugins']
INFO     root        : Volatility symbols path: ['/home/dev/volatility3/volatility/symbols', '/home/dev/volatility3/volatility/framework/symbols']
INFO     volatility.plugins.yarascan: Python Yara module not found, plugin (and dependent plugins) not available
DEBUG    volatility.framework: /opt/conda/lib/python3.8/site-packages/yara.cpython-38-x86_64-linux-gnu.so: undefined symbol: sk_free
DEBUG    volatility.framework: Failed to import module volatility.plugins.yarascan based on file: yarascan
DEBUG    volatility.framework: No module named 'Crypto'
DEBUG    volatility.framework: Failed to import module volatility.plugins.windows.cachedump based on file: windows/cachedump
INFO     volatility.plugins.yarascan: Python Yara module not found, plugin (and dependent plugins) not available
DEBUG    volatility.framework: /opt/conda/lib/python3.8/site-packages/yara.cpython-38-x86_64-linux-gnu.so: undefined symbol: sk_free
DEBUG    volatility.framework: Failed to import module volatility.plugins.windows.callbacks based on file: windows/callbacks
DEBUG    volatility.framework: No module named 'Crypto'
DEBUG    volatility.framework: Failed to import module volatility.plugins.windows.hashdump based on file: windows/hashdump
[...]

PR #350, opened 2020-09-23
refs #333 catch InvalidAddressException instead of AttributeError when accessing DLL timestamps

iMHLv2:
This should address #333. In particular, there's no need to catch `AttributeError` if we're already using `dll_load_time_field` which limits accesses to only the Windows versions that have the member.


ikelos (approved):
Yep, I like it.  It differentiates DllLoad based on why it's not available.  5:)

PR #352, opened 2020-09-29
fix windows xp bug

memoryforensics1:
fix windows xp:
environment_variables
privileges


volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -790,12 +793,29 @@ def get_sids(self) -> Iterable[str]:
...
+            for priv_index in range(64):
+                yield (priv_index,
+                       bool(self.Privileges.Present & (2**priv_index)),
+                       bool(self.Privileges.Enabled & (2**priv_index)),
+                       bool(self.Privileges.EnabledByDefault & (2**priv_index)))
+        except AttributeError: # Windows XP
+            layer_name = self.vol.layer_name
+            kvo = self._context.layers[layer_name].config["kernel_virtual_offset"]        
+            symbol_table = self.get_symbol_table_name()
+            ntkrnlmp = self._context.module(symbol_table,
```
ikelos, line 808: I don't think you need to make a module, which also means you don't need to load the kvo.  I think you should just be able to ask for `self.context.symbol_space[self.get_symbol_table_name()].get_type("_LUID_AND_ATTRIBUTES")` to get the subtype.  That should remove a lot of these extra lines. and hopefully speed up the call...

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -790,12 +793,29 @@ def get_sids(self) -> Iterable[str]:
...
+                                            layer_name = layer_name,
+                                            offset = kvo)
+            if self.PrivilegeCount < 1024:
```
ikelos, line 811: What happens if this isn't the case?  It looks like we'd return an empty list.  If we'd gone to the effort of differentiating this, it seems like we should log when it goes wrong?

volatility/framework/symbols/windows/extensions/__init__.py
```diff
@@ -790,12 +793,29 @@ def get_sids(self) -> Iterable[str]:
...
+        
+        try:
+            for priv_index in range(64):
+                yield (priv_index,
+                       bool(self.Privileges.Present & (2**priv_index)),
+                       bool(self.Privileges.Enabled & (2**priv_index)),
+                       bool(self.Privileges.EnabledByDefault & (2**priv_index)))
+        except AttributeError: # Windows XP
+            layer_name = self.vol.layer_name
+            kvo = self._context.layers[layer_name].config["kernel_virtual_offset"]
```
ikelos, line 806: We've let a few of these slip in, but if possible please use the public `self.context` to get the context, rather than referring to the private internal value `self._context`.  It's just in case in the future we want to calculate/change the context, and the context property would let us do that easily, whereas the private variable references would make it trickier for use to do (all highly unlikely, so definitely not a blocker, just a good practice thing).  5:)

ikelos (commented):
Looks good, just a minor optimization and then we should be good to go (sorry it took me so long to review)...

volatility/framework/symbols/windows/extensions/__init__.py
memoryforensics1: only the _context exists

ikelos: Ah, you're quite right, my mistake.  I thought we exposed it as a property.  Ignore this then!  5:)

ikelos:
Looks good, thanks!  5:)

PR #359, opened 2020-10-19
Mac kevents

atcuno:


volatility/framework/plugins/mac/kevents.py
```diff
@@ -0,0 +1,185 @@
...
+
+        filters = self.all_filters[filter_index]
+        for flag, index in filters:
+            if filter_flags & index == index:
+                context.append(flag)
+
+        return ",".join(context)
+
+    @classmethod
+    def _walk_klist_array(cls, kernel, fdp, array_pointer_member, array_size_member):
```
ikelos, line 85: Looks good, just want to check that these aren't ever intended to be used separately (since they're private)?

volatility/framework/plugins/mac/kevents.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def _generator(self):
+        filter_func = pslist.PsList.create_pid_filter(self.config.get('pid', None))
+
+        for task_name, pid, kn in self.list_kernel_events(self.context,
+                                                self.config['primary'],
+                                                self.config['darwin'],
+                                                filter_func = filter_func):
+
+            filter_index = kn.kn_kevent.filter * -1
```
ikelos, line 166: This seems a bit odd, is it just flipping the sign bit (should the type really be unsigned) or is something else going on?

ikelos (approved):
Looks really good, I'm a massive fan of the version history, and thanks for adding in the typing for the API method (feel free to also type other methods too, it all helps).  5:)  A couple of questions, but very happy to approve this as is...  5:)

volatility/framework/plugins/mac/kevents.py
atcuno: it is actually stored as a negative number in the kernel.

Open this:

https://opensource.apple.com/source/xnu/xnu-517.9.5/bsd/sys/event.h.auto.html

then search:

EVFILT_READ

and you will see they are all negative numbers, it is strange.

PR #363, opened 2020-10-23
Add plugins for enumerate of mac kauth listeners and scopes

atcuno:


ikelos (approved):
All looks good, you're making it harder and harder to find things to comment on!  5:P

PR #364, opened 2020-10-23
implement a smear proof mac lsmod

atcuno:


volatility/framework/plugins/mac/lsmod.py
```diff
@@ -38,12 +38,29 @@ def list_modules(cls, context: interfaces.context.ContextInterface, layer_name:
...
+
+        kmod = kmod.next
+
+        seen = set()
+
+        while kmod != 0 and \
+              kmod not in seen and \
+              len(seen) < 1024:
+
+            if not kernel_layer.is_valid(kmod.dereference().vol.offset, kmod.dereference().vol.size):
```
ikelos: I think that dereferencing the pointer and takes its offset would be the same as just `kmod`?  It may make more sense so that it matches the size, so not a big deal

volatility/framework/plugins/mac/lsmod.py
```diff
@@ -38,12 +38,29 @@ def list_modules(cls, context: interfaces.context.ContextInterface, layer_name:
...
 
+        kernel_layer = context.layers[layer_name]
+
         kmod_ptr = kernel.object_from_symbol(symbol_name = "kmod")
 
-        # TODO - use smear-proof list walking API after dev release
         kmod = kmod_ptr.dereference().cast("kmod_info")
-        while kmod != 0:
+
+        yield kmod
```
ikelos, line 50: Hmmmm, feels like the first one could presumably be invalid too?  Might be worth a check and then bombing out with an empty list if it is?

ikelos (approved):
Just a couple minor bits, but pretty happy with this too...  5:)

volatility/framework/plugins/mac/lsmod.py
atcuno: I had issues with this. I believe we talked about this in the past, but then I don't remember the resolution. Basically, when you have a pointer, if you do variable.vol.{offset, size} you get the values for the actual pointer, meaning the value of the address its pointing to and then 4 or 8 for the size depending on 32 or 64 bit layer. When you do .dereference() you get the actual address and size of whatever object the pointer is referencing.

volatility/framework/plugins/mac/lsmod.py
atcuno: True, will fix this pending your reply to my first comment on dereference(). I would like feedback on that first to make sure we are on the same page and/or if I missed something.

volatility/framework/plugins/mac/lsmod.py
ikelos: So, you're entirely right, I think I wasn't clear in what I was say:
```
kmod (pointer value)
kmod.vol.offset (offset of pointer)
kmod.dereference() (object at pointer value)
kmod.dereference().vol.offset (offset of object at pointer value = pointer value)
```
So essentially `kmod.vol.offset` is the wrong value to use, but `kmod` should be identical to `kmod.dereference().vol.offset` only without having to construct the dereferenced object to get it.  Since you shortly after use `kmod.deferefence().vol.size` which we can't use that trick for, the new object will get created regardless.  I might also suggest putting `kmod.deference()` into a variable, so we don't have to construct the object twice.  It's a small optimization, but if this is run many times over it might add up?  As I say, not a blocker and absolutely happy for you to merge it as is, but figured I'd mention it so that we're aware of the trade-offs...  5:)

ikelos:
Cool, hopefully I'm not holding this up, but if there's anything else you need me to do on this, lemme know...  5:)

ikelos:
We're going to try and get this reviewed and sorted before the code-freeze tomorrow night, but otherwise it'll go on the develop branch...

ikelos:
Looks good, how critical a bugfix is this (as in will people get bad results if it doesn't go in the release branch or can it live in develop for a few months before we do another point release)?

atcuno:
> Looks good, how critical a bugfix is this (as in will people get bad results if it doesn't go in the release branch or can it live in develop for a few months before we do another point release)?

develop is fine :)

PR #370, opened 2020-11-01
Renderers: Add in very rudimentary renderer options

ikelos:
This allows the CLI to expose options that the renderers can offer.
These options are limited (string, int, bool and bytes), so no lists
or choices, etc.

Each renderer's options are also global options, meaning they're always
present (although grouped).  There's two ways of handling this:

a) Prefix each option with the renderer's name, duplicate options are
unique
b) Deduplicate options with duplicate names and prefix 'renderer'

The current choice is a) because b) might confuse people that an option
which only applied to one renderer would be available for all, but
comes with the inconvenience that users must adapt the option name for
those that are duplicated with the same meaning.

This is at least functional and extendable in the future but still a bit
clunky.  We could have brought the entire configuration mechanism into
play, but each renderer would require a context and a config_path and
so on, meaning massive overkill and a non-addition-only API change.

This is not an urgent issue, nor one I would rush through, so I'm happy for this just to sit here and gather comments until everyone's happy with it...

### DIFF
--- a/volatility/cli/__init__.py
+++ b/volatility/cli/__init__.py
@@ -284,6 +284,7 @@ class CommandLine:
         self.output_dir = args.output_dir
 
         self.populate_config(ctx, chosen_configurables_list, args, plugin_config_path)
+        options = self.populate_renderer_option_values(renderer, args)
 
         if args.extend:
             for extension in args.extend:
@@ -314,7 +315,6 @@ class CommandLine:
 
         try:
             # Construct and run the plugin
-            options = []
             if constructed:
                 renderer(options).render(constructed.run())
         except (exceptions.VolatilityException) as excp:
@@ -589,6 +589,7 @@ class CommandLine:
                 extra_options = {
                     'help': option.description,
                     'type': option.option_type,
+                    'default': None,
                     'dest': config_name.replace('-', '_')
                 }
                 if option.option_type == bool:
@@ -596,6 +597,18 @@ class CommandLine:
                     extra_options['action'] = 'store_true'
                 renderer_parser.add_argument("--" + config_name,
                                              **extra_options)
+
+    def populate_renderer_option_values(self, renderer: Type[text_renderer.CLIRenderer], cli_options):
+        """Populates an options dictionary with the appropriate values"""
+        options = []
+        for option in renderer.get_render_options():
+            config_name = '-'.join([renderer.name, option.name]).replace('-', '_')
+            if hasattr(cli_options, config_name):
+                if getattr(cli_options, config_name) is not None:
+                    option.value = getattr(cli_options, config_name)
+                    options.append(option)
+        return options
+
 
 
 def main():
--- a/volatility/cli/__init__.py
+++ b/volatility/cli/__init__.py
@@ -167,7 +167,8 @@ class CommandLine:
         partial_args, _ = parser.parse_known_args(known_args)
 
         banner_output = sys.stdout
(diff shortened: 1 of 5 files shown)


ikelos:
@npetroni  I don't want to rush this, but there is now actual call for it being present in the code.  Technically the renderers only affect the CLI, even though the interfaces are part of the framework.  We can move the interfaces to the CLI code and then not have to bump the version number, but I left them in there because I wanted to give people an idea of how they'd need to handle the data output, and just giving them the treegrid seemed mean...  5:P  If you'd be able to review this before Wednesday then we can decided whether it makes it in before the code freeze or not...  5:)

volatility/cli/__init__.py
```diff
@@ -311,8 +314,9 @@ def run(self):
 
         try:
             # Construct and run the plugin
+            options = []
             if constructed:
-                renderers[args.renderer]().render(constructed.run())
+                renderer(options).render(constructed.run())
```
N1neSun, line 319: I took a look at the implementation of this renderer option and would like to ask although the renderer options has been added globally, the value of the option cannot be passed to the corresponding renderer. Do I need to pass in parameters for the options here?

ikelos: No, you shouldn't have to change the CLI code, you need to define your options within the renderer, as an example you can see:
https://github.com/volatilityfoundation/volatility3/pull/370/files#diff-7360151e7175c6a777de1f856eb59bb2f4937d47cbd6c0198f0a675457165543R150

That should then show up as an `--<renderer_name>-<option_name>`

N1neSun: Yes, but something is wrong. I used --<renderer_name>-<option_name>, it didn't work (e.x,:--pretty-separator), and I didn't find the option value being passed to the value in the RenderOption class, did I do something wrong?

ikelos: Hmmm, I'll check it tomorrow and make sure it works...  Happy New Year, by the way!  5:)

N1neSun: Thanks a lot, The same to you,Happy New Year!

ikelos: You were quite right, I completely neglected to connect the standard options together with their render options.  I've added that in now, so please let me know if that helps and we can mark this comment as resolved.  5;)

N1neSun: I think the problem has been solved, thanks again！

N1neSun:
hi, I have a question to ask whether all global args parameters should be passed to renderers so that I can get the plugin name, export directory, etc.Is this more reasonable?

ikelos:
Errr, tying those things together would make it more difficult to change either side of the equation later on, for any reason.  It would be better to identify specifically what an output renderer may need, and then change the API so that that's passed through to the renderer.  The renderers are designed to be simplistic though, and my hope was that people would come along and improve on the CLI or write web UIs that would take up the task, but that may be a while away yet...

N1neSun:
thanks,I also know this may not be a simple thing, because there may be many things to consider.I will take a closer look.And trying to improve this.

atcuno:
@ikelos this is from 2021. I assume this is handled elsewhere? From reading the initial description it sounds like what we already support.

ikelos:
No, we still haven't got option renderers in place, although we got slightly we closer recently, since they'll be needed to allow for setting the context output for the CLI (so the data can have bytes before and bytes after output).  This is probably dead code though, so I'll mark it as closed.

See #1622

PR #378, opened 2020-11-18
OS check for physical to virtual

cecio:
fix: the bounce from physical to virtual process is required only for OS < Win10.

Added the check in the "dump" option.

### DIFF
--- a/volatility/framework/plugins/windows/psscan.py
+++ b/volatility/framework/plugins/windows/psscan.py
@@ -91,40 +91,38 @@ class PsScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
         """
 
         version = cls.get_osversion(context, layer_name, symbol_table)
-        if version < (10, 0, 0):
-            # If it's WinXP->8.1 we have now a physical process address.
-            # We'll use the first thread to bounce back to the virtual process
-            kvo = context.layers[layer_name].config['kernel_virtual_offset']
-            ntkrnlmp = context.module(symbol_table, layer_name = layer_name, offset = kvo)
-    
-            tleoffset = ntkrnlmp.get_type("_ETHREAD").relative_child_offset("ThreadListEntry")
-            # Start out with the member offset
-            offsets = [tleoffset]
-    
-            # If (and only if) we're dealing with 64-bit Windows 7 SP1
-            # then add the other commonly seen member offset to the list
-            bits = context.layers[layer_name].bits_per_register
-            if version == (6, 1, 7601) and bits == 64:
-                offsets.append(tleoffset + 8)
-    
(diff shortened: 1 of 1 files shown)


iMHLv2:
@cecio Thanks for reporting this bug and proposing a fix. Your example looks fine, but I wanted to share another option that is a little more generic. For example, instead of treating Windows 10 in a special way, we should be able to just check which layer the process object was instantiated from, and try to bounce if needed. What do you think of this? @ikelos is this an appropriate way to check if an object is in the virtual or physical layer?

```
% git diff                                            
diff --git a/volatility/framework/plugins/windows/psscan.py b/volatility/framework/plugins/windows/psscan.py
index 769b3e87..8424851f 100644
--- a/volatility/framework/plugins/windows/psscan.py
+++ b/volatility/framework/plugins/windows/psscan.py
@@ -139,8 +139,12 @@ class PsScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
 
             file_output = "Disabled"
             if self.config['dump']:
-                vproc = self.virtual_process_from_physical(self.context, self.config['primary'],
-                                                           self.config['nt_symbols'], proc)
+                # windows 10 objects (maybe others in the future) are already in virtual memory
+                if proc.vol.layer_name == self.config['primary']:
+                    vproc = proc
+                else:
+                    vproc = self.virtual_process_from_physical(self.context, self.config['primary'],
+                                                               self.config['nt_symbols'], proc)
 
                 file_handle = pslist.PsList.process_dump(self.context, self.config['nt_symbols'], pe_table_name,
                                                          vproc, self.open)
```

ikelos:
@iMHLv2 Yeah, that's pretty nice.  I'm just trying to figure out if there'd be a case where the layer might not be 'primary' (if we ever allowed searching in process address spaces somehow), but I can't think of one, given that's the layer we're targetting in the first place.  You could also check `native_layer_name == layer_name`, but that doesn't guarantee it's in primary, just that it was constructed outside of its native layer, so not as good a test.  Yeah, pretty happy with that...  5:)

cecio:
Hello. Thanks @iMHLv2  and @ikelos.
I re-worked the fix as suggested and pushed a new commit.
Let me know if you need any additional mod.

iMHLv2:
Cool, thanks! This looks good now. @ikelos feel free to merge. 

@cecio Just FYI there's another way to detect the OS of a memory sample, see https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/versions.py#L82. Although that method exists, I'm happy keeping your implementation of `cls.get_osversion()` since it provides a different perspective of the version details, and may come in handy in the future if other methods don't work.

PR #380, opened 2020-11-18
Fix 'maxsize variable doing nothing' bug

AsafEitani:
In one of the earlier commits the varible was erased from being passed to vaddump - that corrects it.


ikelos:
Ahh, good spot, looks like it went missing in commit `56266f97d`, thanks!  5:)

PR #388, opened 2020-12-04
support sqlite output

N1neSun:
Convert csv to sqlite output through pandas module

### DIFF
--- a/volatility/cli/__init__.py
+++ b/volatility/cli/__init__.py
@@ -160,6 +160,10 @@ class CommandLine:
                             help = "Clears out all short-term cached items",
                             default = False,
                             action = 'store_true')
+        parser.add_argument("--output-sqlitepath",
+                            help = "Path in which to output sqlite db",
+                            default = os.getcwd(),
+                            type = str)
 
         # We have to filter out help, otherwise parse_known_args will trigger the help message before having
         # processed the plugin choice or had the plugin subparser added.
@@ -314,7 +318,10 @@ class CommandLine:
         try:
             # Construct and run the plugin
             if constructed:
-                renderers[args.renderer]().render(constructed.run())
+                if args.renderer != "sqlite":
+                    renderers[args.renderer]().render(constructed.run())
+                else:
+                    renderers[args.renderer]().render(constructed.run(), [args.plugin, args.output_sqlitepath])
         except (exceptions.VolatilityException) as excp:
             self.process_exceptions(excp)
 
--- a/volatility/cli/text_renderer.py
+++ b/volatility/cli/text_renderer.py
@@ -6,10 +6,9 @@ import json
 import logging
 import random
 import string
-import sys,os
+import sys
 import sqlite3
 import tempfile
-import pandas
 from functools import wraps
 from typing import Any, List, Tuple, Dict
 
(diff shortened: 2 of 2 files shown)


ikelos:
Thanks for this!  There's a few points that I'd suggest doing a different way:

* The temporary file to hold the csv looks well done and suitably considered, but having a temporary file shouldn't really be necessary, the data could just be written directly to a sqlite database with appropriate sqlite commands?
* Pandas is an external dependency which would then need to be added to the core requirements just for this one output mechanism, and it's only necessary for the conversion
* It modifies the core command line interface, meaning it can't just be dropped into an existing installation.  I realize this is because it needs to write a file and we don't yet really handle giving options to plugins (although it is on the roadmap, see #370).

Areas to consider include using the already provided `output-dir` option, rather than adding yet another option for a very specific output method.  I'd also contemplate working the file interface handler (for plugins to write out files) into the plugin options, but it needs to support more than just this plugin, we're aiming to only add things to the CLI that are modular rather than specific to a particular output mechanism where possible.

As such, I'm likely to hold off on accepting this as is, I'm afraid...

N1neSun:
Thank you for your suggestion. I was just to implement the function at the time and did not consider the volatility framework. I will rewrite this place to adapt to this framework better.

N1neSun:
I realize that you have implemented the renderer option, could I know when it will merge?

ikelos:
Unfortunately the renderer options issue still needs a little more consideration, so I'm afraid there isn't a specific deadline for it getting added, but the issue will be updated when there's any news...

N1neSun:
Thanks a lot, I saw your question about the renderer options, I will continue to pay attention to this #370 and discuss with you.

N1neSun:
I changed the sqlite output with reference to volatility2. I don’t know if this is feasible?

ikelos:
That much better, thanks!  Unfortunately we're still waiting on getting the renderer options into the tree (due to the filename argument needing to be used), but having this here should be useful for people looking to get sqlite output....  5:)

N1neSun:
It’s okay, I know there are still some problems that need to be solved, and I also hope I can help with something.

JonathanOuellet1275:
Hello, do we have an estimate as to when this functionality will be made available?

ikelos:
@JonathanOuellet1275 We don't has a timeline for this pull request.  It alters the CLI in a way that's very specific to sql support, which won't extend nicely if we gets lots of requests for lots of formats.  As such we don't think it's the best way of supporting this need.

At the moment there's a straightforward workaround which is to output as csv (`-r csv`) and then start a new sqlite file, and import the csv data into it (`.import FILE TABLE`).  Hopefully this will help you get the data into sqlite before we're able to come up with a workable solution for adding support directly...

JonathanOuellet1275:
@ikelos Thank you for the tip

atcuno:
@ikelos is the CSV output to then import to sqlite still the way we will support? If so, should we close this one?

ikelos:
Yeah, I think adding sqlite direct support cause a whole lot of problems and would be more easily handled by a wrapper script to pull the data out via CSV.  Eventually it should be possible to provide renderers option values (such as the output filename), but a whole extra global option for the CLI is too much for such a niche requirement.

PR #389, opened 2020-12-05
Doc fix: Explanation of surjective Intel layer.

NiklasBeierl:
Terms physical and virtual address were mixed up.

### DIFF
--- a/doc/source/complex-plugin.rst
+++ b/doc/source/complex-plugin.rst
@@ -182,8 +182,9 @@ of data.  Each chunk contains the following information, in order:
 
 An example (and the most common layer encountered in memory forensics) would be an Intel layer, which models the intel
 page mapping system.  Based on a series of tables stored within the layer itself, an intel layer can convert a virtual
-address to a physical address.  It should be noted that intel layers are surjective in that a single virtual address can
-map to multiple physical addresses, but a single virtual address can only ever map to a single physical address.
+address to a physical address.  It should be noted that intel layers are surjective. Meaning a single virtual address 
+can only ever map to a single physical address, but a single physical address can be mapped to multiple virtual 
+addresses.
 
 As a simple example, in a virtual layer which looks like `abracadabra` but maps to a physical layer that looks
 like `abcdr`, requesting `mapping(5, 4)` would return:


ikelos (changes requested):
Thanks, I'm not sure it's suitable to start a sentence with Meaning (since it refers to the contents of the previous sentence), but I'm happy to accept the previous sentence is run on.  Would you be ok with "This means that" instead of "Meaning"?

doc/source/complex-plugin.rst
```diff
@@ -182,8 +182,9 @@ of data.  Each chunk contains the following information, in order:
 
 An example (and the most common layer encountered in memory forensics) would be an Intel layer, which models the intel
 page mapping system.  Based on a series of tables stored within the layer itself, an intel layer can convert a virtual
-address to a physical address.  It should be noted that intel layers are surjective in that a single virtual address can
-map to multiple physical addresses, but a single virtual address can only ever map to a single physical address.
+address to a physical address.  It should be noted that intel layers are surjective. Meaning a single virtual address 
+can only ever map to a single physical address, but a single physical address can be mapped to multiple virtual
```
ikelos, line 186: I think this correction is the wrong way around.  The mapping goes from Virtual -> Physical, so a surjective mapping would allow multiple virtual addresses to map to a single physical address (which is what the text said originally).  I'm pretty sure I've got that the right way around, and you can definitely have two virtual addresses that point to the same physical address, but there isn't really a way for a virtual address to map to two different physical addresses (and that would be an injective function, if it were).

https://en.wikipedia.org/wiki/Surjective_function

ikelos (changes requested):
Sorry, having reread this I'm willing to accept that it's a run on sentence, but this correction inverts the meaning.  I'm happy to clarify what's meant, but I'm not sure the correction here is accuate?

doc/source/complex-plugin.rst
NiklasBeierl: You are making my head smoke. :exploding_head: :laughing: 
Thinking about it further, I think we are both wrong. :smile: 

Original:
> It should be noted that intel layers are surjective in that   
a single virtual address can map to multiple physical addresses, but 
a single virtual address can only ever map to a single physical address.

I hope you can agree with me that this is contradictory.

I refreshed my memory on the definition of surjective. (Thanks for the wiki link) Now I am not sure whether we are getting this part right. A surjective function is one where EVERY element in the codomain has a pre-image in the domain. In our terms that would mean that EVERY physical address (without exception) has a virtual address which is resolved to that physical address by the intel layer. I honestly can't answer that question, because I don't know enough about this technology.
To make this a little easier to think about, look at the four images on this article:  https://en.wikipedia.org/wiki/Injective_function
Is our situation described by the 3rd or 4th picture? Are there physical addresses without a virtual address mapped to them?
The mapping is certainly not injective, but whether it is surjective or not is a different story. Injective and surjective are not opposites. 

Now for my Correction:
>Meaning 
a single virtual address can only ever map to a single physical address, 
but a single physical address can be mapped to multiple virtual 

What I wanted to express here is that our mapping is **not injective**, without touching the question of surjective.
The key here is that a virtual address **maps** (active, arrow goes away from virtual addr.), 
and a physical address **is mapped to** (passive, arrows points at physical addr.).
Now in my circles this is a clear distinction: We are looking at the **image** of the virtual address
and at the **pre-image** of the physical address.  But this phrasing might not work for everyone. I will try to formulate it more concisely and push an update.

ikelos: Yep, so I didn't notice the typo in the original, but I agree, it essentially says `A something B` and then `A not something B`, which it entirely wrong.

You've also corrected me, quite accurately, in that surjection requires that every element in the codomain is the output for some entry in the domain, which is not the case (you can have slack space in physical memory which has not been mapped).

Unfortunately, that's a shame, because what I was trying to get across was that you can have `V -> P` and `V' -> P` but you can't ever have `V -> P` and `V -> P'`.  I'd though surjection would describe that, but my maths is a little rusty...  5;P

I think the idea "can be mapped to" made me think the relation was going in the other direction.  Perhaps attempting to use precise mathematical terminology isn't the best way to get the point across (since either the maths or me have managed to confuse us both  5:P  )?

I'm happy to go with a suitable correction, the best I've got to offer would be:
```
It should be noted that intel layers allow multiple virtual addresses to map to the same physical address (but a single virtual address cannot ever map to more than one physical address).
```
That avoids the surjection bit, and keeps the mapping always going in the same direction to avoid confusing people?

ikelos: Did we reach a conclusion on this?  We'd really like to get this in the release coming up soon, and we're very close to entering a hard code freeze.  Since this is inaccurate documentation we can fix it during the time, but we need to do it as soon as possible...

ikelos:
Ok, I've updated this for the release because I don't want to ship with incorrect documentation.  I just removed the word surjective completely, and I'll close this off, but feel free to reopen it (or submit a new one) if you think there's a better way of describing it...  5:)

PR #392, opened 2020-12-07
Netscan fixes

japhlange:
Fixes a few things for `netscan`.

- the symbol table is chosen based on the newest one available for the given Windows version not just for Win10, but for older versions as well (>= Vista)
- fixes bug which can occur when trying to dereference a nullpointer actually works because the image has a valid virtual memory address of `0`, which happened to me in exactly one memory image of 30. Honestly no idea how often this will happen...
- adds a bunch of new symbol tables for different Windows versions, including the change for Win10x64_18363 found by @mikevsunny in https://github.com/volatilityfoundation/volatility/issues/763. This obviously does not fix the issue that 18363 is not properly detected, but one step after the other :)

Note: https://github.com/volatilityfoundation/volatility3/pull/392/files#diff-4f922f46a5f105772c8f988caf36079c512c87a01949d0639a991ff080024187 contains a bunch of symbols which are needed for parsing the "official" kernel network address lists. I have only documented those for Win10x64_16299 for now, but they will (most likely) be added for the other Windows versions later on. I hope it's okay that they are already contained here, but if necessary I can remove them for this PR.

### DIFF
--- a/volatility/framework/plugins/windows/netscan.py
+++ b/volatility/framework/plugins/windows/netscan.py
@@ -153,10 +153,10 @@ class NetScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                 (10, 0, 14393): "netscan-win10-14393-x86",
                 (10, 0, 15063): "netscan-win10-15063-x86",
                 (10, 0, 16299): "netscan-win10-15063-x86",
-                (10, 0, 17134): "netscan-win10-15063-x86",
-                (10, 0, 17763): "netscan-win10-15063-x86",
-                (10, 0, 18362): "netscan-win10-15063-x86",
-                (10, 0, 18363): "netscan-win10-15063-x86"
+                (10, 0, 17134): "netscan-win10-17134-x86",
+                (10, 0, 17763): "netscan-win10-17134-x86",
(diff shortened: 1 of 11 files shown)


ikelos (approved):
Yep, pretty happy with all of that, thanks!  5:)  No problem adding the other bits if they'll be useful in the future...

ikelos:
Also, we're headed towards a code-freeze (as in, on Thursday), so please let me know if this can standalone and should go in before the code freeze, or be held until after it?

japhlange:
> Also, we're headed towards a code-freeze (as in, on Thursday), so please let me know if this can standalone and should go in before the code freeze, or be held until after it?

This PR is standalone, so I see no particular reason to hold it. How long is the code-freeze going to be?

ikelos:
The code-freeze will actually be more a branching, so we can still add code, but it won't make it into the release.  The expectation is that the branch will only get fixes for 2-4 weeks (documentation updates, minor alterations are fine, but no additions), but then 2 weeks of no changes (only critical fixes) during which we'll check packaging and make sure everything's ready to roll.  We'll still be accepting new changes on master, but not on the branch.

japhlange:
I added a workaround to distinguish between 18362 and 18363. This should fix the issue mentioned above.

PR #397, opened 2020-12-19
Netstat for x64 Windows

japhlange:
This plugin parses the "official" tcpip.sys kernel structures to find network connections, similar to what `netstat` does. 
~It is named `netlist` because it is to `netscan` what `pslist` is to `psscan`.~ Renamed to `netstat` for consistency :)

It finds...
- TCP Endpoints by parsing the TCP Partition Table (see _The Art of Memory Forensics_, chapter 11)
- UDP Endpoints by parsing the UDP port pool
- TCP Listeners by parsing the TCP port pool

It depends on some requirements:
- the given image must have its `tcpip.sys` PE image headers accessible in memory to fetch the PDB GUID (this has worked for every image I have tested this with, so I consider this given)
- the offsets must be correct, but that of course is required for every plugin with custom objects ;)

This plugin is of course way faster than `netscan`, but might not find every object the poolscanner will find (especially older, not-yet-deleted ones). Also, kernel-mode malware might be able to hide entries from the mentioned structures, but to counter that you can easily compare the outputs of `netscan` and `netstat`. (Also, if I was a rootkit, I'd "just" hook the enumeration APIs and not bother touching weird, undocumented structures deep inside drivers...)

For now it only supports 64-bit Windows as I haven't gotten around adding x86 support yet and I would prefer feedback earlier than later :)

List of supported builds (more will follow at some point):
- Win7x64
- Win10x64_16299
- Win10x64_17134
- Win10x64_17763
- Win10x64_18362
- Win10x64_18363

There is probably a lot of room for optimization in the code, but I wanted to get this one out for discussion :D

And of course, testing is welcome!!

### DIFF
--- a/volatility/framework/plugins/windows/netlist.py
+++ b/volatility/framework/plugins/windows/netlist.py
@@ -42,7 +42,9 @@ class NetList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
 
     @classmethod
     def _decode_pointer(self, value):
-        """Windows encodes pointers to objects and decodes them on the fly
+        """Copied from `windows.handles`.
+
+        Windows encodes pointers to objects and decodes them on the fly
         before using them.
 
         This function mimics the decoding routine so we can generate the
@@ -59,6 +61,17 @@ class NetList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                      layer_name: str,
                      offset: int,
                      length: int) -> int:
+        """Reads a pointer at a given offset and returns the address it points to.
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            layer_name: The name of the layer on which to operate
+            offset: Offset of pointer
+            length: Pointer length
+
+        Returns:
+            The value the pointer points to.
+        """
 
         return int.from_bytes(context.layers[layer_name].read(offset, length), "little")
 
@@ -68,6 +81,17 @@ class NetList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                      layer_name: str,
                      bitmap_offset: int,
                      bitmap_size_in_byte: int) -> list:
+        """Parses a given bitmap and looks for each occurence of a 1.
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            layer_name: The name of the layer on which to operate
+            bitmap_offset: Start address of bitmap
+            bitmap_size_in_byte: Bitmap size in Byte, not in bit.
+
+        Returns:
+            The list of indices at which a 1 was found.
+        """
         ret = []
         for idx in range(bitmap_size_in_byte-1):
             current_byte = context.layers[layer_name].read(bitmap_offset + idx, 1)[0]
@@ -96,8 +120,22 @@ class NetList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                        layer_name: str,
                        net_symbol_table: str,
                        port: int,
-                       ppobj,
-                       proto="tcp"):
+                       port_pool: interfaces.objects.ObjectInterface,
+                       proto="tcp") -> \
+            Iterable[interfaces.objects.ObjectInterface]:
+        """Lists all UDP Endpoints and TCP Listeners by parsing UdpPortPool and TcpPortPool.
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            layer_name: The name of the layer on which to operate
+            net_symbol_table: The name of the table containing the tcpip types
+            port: Current port as integer to lookup the associated object.
+            port_pool: Port pool object
+            proto: Either "tcp" or "udp" to decide which types to use.
+
+        Returns:
+            The list of network objects from this image's TCP and UDP `PortPools`
+        """
         if proto == "tcp":
             obj_name = net_symbol_table + constants.BANG + "_TCP_LISTENER"
             ptr_offset = context.symbol_space.get_type(obj_name).relative_child_offset("Next")
@@ -105,36 +143,168 @@ class NetList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
             obj_name = net_symbol_table + constants.BANG + "_UDP_ENDPOINT"
             ptr_offset = context.symbol_space.get_type(obj_name).relative_child_offset("Next")
         else:
+            # invalid argument.
             yield
+
+        # the given port serves as a shifted index into the port pool lists
         list_index = port >> 8
         truncated_port = port & 0xff
-        inpa = ppobj.PortAssignments[list_index].dereference()
+
+        # first, grab the given port's PortAssignment (`_PORT_ASSIGNMENT`)
+        inpa = port_pool.PortAssignments[list_index].dereference()
+
+        # then parse the port assignment list (`_PORT_ASSIGNMENT_LIST`) and grab the correct entry
         assignment = inpa.InPaBigPoolBase.dereference().Assignments[truncated_port]
+
         if not assignment:
             yield
+
+        # the value within assignment.Entry is a) masked and b) points inside of the network object
+        # first decode the pointer
         netw_inside = cls._decode_pointer(assignment.Entry)
+
         if netw_inside:
+            # if the value is valid, calculate the actual object address by subtracting the offset
             curr_obj = context.object(obj_name, layer_name = layer_name, offset = netw_inside - ptr_offset)
-            vollog.debug("Found object @ 0x{:2x}, yielding...".format(curr_obj.vol.offset))
+            vollog.debug("Found {} object @ 0x{:2x}, yielding...".format(proto, curr_obj.vol.offset))
             yield curr_obj
 
-            vollog.debug("PrevPointer val: {}".format(curr_obj.Next))
+            # if the same port is used on different interfaces multiple objects are created
+            # those can be found by following the pointer within the object's `Next` field until it is empty
             while curr_obj.Next:
                 curr_obj = context.object(obj_name, layer_name = layer_name, offset = cls._decode_pointer(curr_obj.Next) - ptr_offset)
                 yield curr_obj
-                vollog.debug("Checking if PrevPointer is valid (val: {})".format(curr_obj.Next))
 
     @classmethod
-    def get_tcpip_module(cls, context, layer_name, nt_symbols):
+    def get_tcpip_module(cls,
+                         context: interfaces.context.ContextInterface,
+                         layer_name: str,
+                         nt_symbols: str) -> interfaces.objects.ObjectInterface:
+        """Uses `windows.modules` to find tcpip.sys in memory.
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            layer_name: The name of the layer on which to operate
+            nt_symbols: The name of the table containing the kernel symbols
+
+        Returns:
+            The constructed tcpip.sys module object.
+        """
         for mod in modules.Modules.list_modules(context, layer_name, nt_symbols):
-            # ~ print(mod.BaseDllName.get_string())
             if mod.BaseDllName.get_string() == "tcpip.sys":
-                vollog.debug("Found tcpip.sys offset @ 0x{:x}".format(mod.DllBase))
+                vollog.debug("Found tcpip.sys image base @ 0x{:x}".format(mod.DllBase))
                 return mod
 
     @classmethod
-    def get_tcpip_guid(cls, context, layer_name, tcpip_module):
-        return list(
+    def parse_hashtable(cls,
+                        context: interfaces.context.ContextInterface,
+                        layer_name: str,
+                        ht_offset: int,
+                        ht_length: int,
+                        alignment: int,
+                        pointer_length: int) -> list:
+        """Parses a hashtable quick and dirty.
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            layer_name: The name of the layer on which to operate
+            ht_offset: Beginning of the hash table
+            ht_length: Length of the hash table
+            pointer_length: Length of this architecture's pointers
+
+        Returns:
+            The hash table entries which are _not_ empty
+        """
+        for index in range(ht_length):
+            # mask pointer so we do not get confused with abbreviated virtual offsets
+            # this currently only works for 64-bit.
+            # TODO: add x86 support.
+            current_qword = (0xffff000000000000 | cls.read_pointer(context, layer_name, ht_offset + index * alignment, pointer_length))
+            if current_qword == (0xffff000000000000 | (ht_offset + index * alignment)):
+                continue
+            yield current_qword
+
+    @classmethod
+    def parse_partitions(cls,
+                       context: interfaces.context.ContextInterface,
+                       layer_name: str,
+                       net_symbol_table: str,
+                       tcpip_symbol_table: str,
+                       tcpip_module_offset: int,
+                       pointer_length: int) -> Iterable[interfaces.objects.ObjectInterface]:
+        """Parses tcpip.sys's PartitionTable containing established TCP connections.
+        The amount of Partition depends on the value of the symbol `PartitionCount` and correlates with
+        the maximum processor count (refer to Art of Memory Forensics, chapter 11).
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            layer_name: The name of the layer on which to operate
+            nt_symbols: The name of the table containing the kernel symbols
+            net_symbol_table: The name of the table containing the tcpip types
+            tcpip_module: The created vol Windows module object of the given memory image
+            tcpip_symbol_table: The name of the table containing the tcpip driver symbols
+
+        Returns:
+            The list of TCP endpoint objects from the `layer_name` layer's `PartitionTable`
+        """
+        if symbols.symbol_table_is_64bit(context, net_symbol_table):
+            alignment = 0x10
+        else:
+            alignment = 8
+
+        obj_name = net_symbol_table + constants.BANG + "_TCP_ENDPOINT"
+
+        # part_table_symbol is the offset within tcpip.sys which contains the address of the partition table itself
+        part_table_symbol = context.symbol_space.get_symbol(tcpip_symbol_table + constants.BANG + "PartitionTable").address
+        part_count_symbol = context.symbol_space.get_symbol(tcpip_symbol_table + constants.BANG + "PartitionCount").address
+
+        # part_table is the actual partition table offset
+        part_table = cls.read_pointer(context, layer_name, tcpip_module_offset + part_table_symbol, pointer_length)
+        part_count = int.from_bytes(context.layers[layer_name].read(tcpip_module_offset + part_count_symbol, 1), "little")
+
+        part_table_size = context.symbol_space.get_type(net_symbol_table + constants.BANG + "_PARTITION").size
+
+        partitions = []
+
+        # create partition objects for each partition and append to list
+        for part_idx in range(part_count):
+            current_partition = context.object(net_symbol_table + constants.BANG + "_PARTITION",
+                                               layer_name = layer_name,
+                                               offset = part_table + part_table_size * part_idx)
+
+            partitions.append(current_partition)
+
+        for partition in partitions:
+            if partition.Endpoints.NumEntries > 0:
+                for endpoint_entry in cls.parse_hashtable(context,
+                                                          layer_name,
+                                                          partition.Endpoints.Directory,
+                                                          part_table_size,
+                                                          alignment,
+                                                          pointer_length):
+
+                    entry_offset = context.symbol_space.get_type(obj_name).relative_child_offset("ListEntry")
+                    endpoint = context.object(obj_name, layer_name = layer_name, offset = endpoint_entry - entry_offset)
+                    yield endpoint
+
+    @classmethod
+    def create_tcpip_symbol_table(cls,
+                                    context: interfaces.context.ContextInterface,
+                                    config_path: str,
+                                    layer_name: str,
+                                    tcpip_module: interfaces.objects.ObjectInterface) -> str:
+        """Creates
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            config_path: The config path where to find symbol files
+            layer_name: The name of the layer on which to operate
+            tcpip_module: The created vol Windows module object of the given memory image
+
+        Returns:
+            The name of the constructed and loaded symbol table
+        """
+        guids = list(
             PDBUtility.pdbname_scan(
                 context,
                 layer_name,
@@ -145,49 +315,8 @@ class NetList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
             )
         )
 
-    @classmethod
-    def parse_hashtable(cls, context, layer_name, ht_offset, ht_length, pointer_length) -> list:
-        # ~ ret = []
-        for idx in range(ht_length):
-            current_qword = (0xffff000000000000 | cls.read_pointer(context, layer_name, ht_offset + idx * 16, pointer_length))
-            if current_qword == (0xffff000000000000 | (ht_offset + idx * 16)):
-                continue
-            yield current_qword
-
-    @classmethod
-    def parse_partitions(cls, context, layer_name, net_symbol_table, tcpip_symbol_table, tcpip_module_offset, pointer_length):
-        # ~ endpoints = []
-        obj_name = net_symbol_table + constants.BANG + "_TCP_ENDPOINT"
-        pto = context.symbol_space.get_symbol(tcpip_symbol_table + constants.BANG + "PartitionTable").address
-        pco = context.symbol_space.get_symbol(tcpip_symbol_table + constants.BANG + "PartitionCount").address
-        part_table = cls.read_pointer(context, layer_name, tcpip_module_offset + pto, pointer_length)
-        part_count = int.from_bytes(context.layers[layer_name].read(tcpip_module_offset + pco, 1), "little")
-        partitions = []
-        for part_idx in range(part_count):
-            current_partition = context.object(net_symbol_table + "!_PARTITION", layer_name = layer_name, offset = part_table + 128 * part_idx)
-            partitions.append(current_partition)
-        for partition in partitions:
-            if partition.Endpoints.NumEntries > 0:
-                for endpoint_entry in cls.parse_hashtable(context, layer_name, partition.Endpoints.Directory, 128, pointer_length):
-                    # ~ yield endpoint
-                    entry_offset = context.symbol_space.get_type(obj_name).relative_child_offset("HashTableEntry")
-                    endpoint = context.object(obj_name, layer_name = layer_name, offset = endpoint_entry - entry_offset)
-                    yield endpoint
-                # ~ endpoints.extend(parse_hashtable(partition.Endpoints.Directory, 128))
-        # ~ return endpoints
-
-    @classmethod
-    def create_tcpip_symbol_table(cls,
-                                    context: interfaces.context.ContextInterface,
-                                    config_path: str,
-                                    layer_name: str,
-                                    tcpip_module):
-
-        guids = cls.get_tcpip_guid(context, layer_name, tcpip_module)
-
         if not guids:
-            print("no pdb found!")
-            raise
+            raise exceptions.VolatilityException("Did not find GUID of tcpip.pdb in tcpip.sys module @ 0x{:x}!".format(tcpip_module.DllBase))
 
         guid = guids[0]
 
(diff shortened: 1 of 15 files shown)


volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
```
ikelos: Sorry, I just always tag this, it should probably be 2020, but not a show stopper.  5:P

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+from volatility.framework.symbols import intermed
+from volatility.framework.symbols.windows.extensions import network
+from volatility.framework.symbols.windows.pdbutil import PDBUtility
+from volatility.plugins import timeliner
+from volatility.plugins.windows import info, poolscanner, netscan, modules
+
+vollog = logging.getLogger(__name__)
+
+
+class NetList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
```
ikelos: Nice, thanks for adding timeliner support!  5:D

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+from typing import Iterable, List, Optional, Callable
+
+from volatility.framework import constants, exceptions, interfaces, renderers, symbols, layers
+from volatility.framework.configuration import requirements
+from volatility.framework.renderers import format_hints
+from volatility.framework.symbols import intermed
+from volatility.framework.symbols.windows.extensions import network
+from volatility.framework.symbols.windows.pdbutil import PDBUtility
+from volatility.plugins import timeliner
+from volatility.plugins.windows import info, poolscanner, netscan, modules
```
ikelos: Looks like some of these imports may be unnecessary.  I couldn't see info or poolscanner used anywhere (but modules will need a `VersionRequirement`)...  5:)

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+            pointer_length: Length of this architecture's pointers
+
+        Returns:
+            The hash table entries which are _not_ empty
+        """
+        for index in range(ht_length):
+            # mask pointer so we do not get confused with abbreviated virtual offsets
+            # this currently only works for 64-bit.
+            # TODO: add x86 support.
+            current_qword = (0xffff000000000000 | cls.read_pointer(context, layer_name, ht_offset + index * alignment, pointer_length))
```
ikelos: Seems like a lot of machinery to essentially process an array of pointers?  Pointer objects are integers themselves (and can therefore be ored), and rather than doing all the index calculations manually, this could just be iterating through an array?  There's no issue doing things this way, particularly, but I'm eager for people to use the parsing functionality of volatility rather than doing it themselves to avoid duplications and to ensure that fixes get applied everywhere (rather than each plugin having their own way of parsing data).  I guess the check against the `ht_offset + index * alignment` afterwards might not be any easier using pointer objects, but just a thing to consider...

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+            context: The context to retrieve required elements (layers, symbol tables) from
+            layer_name: The name of the layer on which to operate
+            offset: Offset of pointer
+            length: Pointer length
+
+        Returns:
+            The value the pointer points to.
+        """
+
+        return int.from_bytes(context.layers[layer_name].read(offset, length), "little")
```
ikelos: I've given more details below, but it seems like a pointer object should do this for you already?

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+            bitmap_size_in_byte: Bitmap size in Byte, not in bit.
+
+        Returns:
+            The list of indices at which a 1 was found.
+        """
+        ret = []
+        for idx in range(bitmap_size_in_byte-1):
+            current_byte = context.layers[layer_name].read(bitmap_offset + idx, 1)[0]
+            current_offs = idx*8
+            if current_byte&1 != 0:
```
ikelos: Definitely feels like these if statements could be more easily expressed as a loop?

```
for bit in range(7):
  if current_byte & (1 << bit) != 0:
    ret.append(bit + current_offsets)
```

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+                ret.append(7 + current_offs)
+        return ret
+
+    @classmethod
+    def enumerate_structures_by_port(cls,
+                       context: interfaces.context.ContextInterface,
+                       layer_name: str,
+                       net_symbol_table: str,
+                       port: int,
+                       port_pool: interfaces.objects.ObjectInterface,
```
ikelos: So, I'm quite a fan of having classmethod functions that don't take what I call "complex" arguments, like objects.  If another plugin needs to make use of this, they've got to have already constructed the TCP or UDP port pool, but since we're already handing in the `net_symbol_table` and `proto`, it feels as though this could instantiate the port pool itself, by taking an offset (simple int) rather than a pre-made port_pool.  Looking at where this is called in the plugin, there doesn't seem to be a case where you'd need a differently constructed port-pool than one that the plugin would construct?  I guess if we scanned for port pools, and then wanted the ports instantiated in the virtual layer, but I'd still probably opt for a `native_layer_name` parameter in that instance.  Have I missed a circumstance where you may want to control the creation of the port pool?  I realize it may get used again by the calling function, but I think I'd still be happier having that done twice, over requiring the calling plugin to know how to construct it itself...

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+        else:
+            # invalid argument.
+            yield
+
+        # the given port serves as a shifted index into the port pool lists
+        list_index = port >> 8
+        truncated_port = port & 0xff
+
+        # first, grab the given port's PortAssignment (`_PORT_ASSIGNMENT`)
+        inpa = port_pool.PortAssignments[list_index].dereference()
```
ikelos: Explicit dereferencing here shouldn't be necessary.  All the members of the dereferenced inpa should be available on the pointer, but I don't overly mind if it you think it's clearer, it's just quite a lot of text.  5:P

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+
+        # then parse the port assignment list (`_PORT_ASSIGNMENT_LIST`) and grab the correct entry
+        assignment = inpa.InPaBigPoolBase.dereference().Assignments[truncated_port]
```
ikelos: Again, you *ought* to be able to just call `BigPoolBase.Assignments[truncated_port]`?  If that's not working let me know, but we didn't want it to be this troublesome to dive through layers and layers of pointers.  5:P

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+    _version = (1, 0, 0)
+
+    @classmethod
+    def get_requirements(cls):
+        return [
+            requirements.TranslationLayerRequirement(name = 'primary',
+                                                     description = 'Memory layer for the kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
+            requirements.VersionRequirement(name = 'netscan', component = netscan.NetScan, version = (1, 0, 0)),
```
ikelos: Looks like there should be a `requirements.VersionRequirement` on `modules.Modules`

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+        else:
+            alignment = 8
+
+        obj_name = net_symbol_table + constants.BANG + "_TCP_ENDPOINT"
+
+        # part_table_symbol is the offset within tcpip.sys which contains the address of the partition table itself
+        part_table_symbol = context.symbol_space.get_symbol(tcpip_symbol_table + constants.BANG + "PartitionTable").address
+        part_count_symbol = context.symbol_space.get_symbol(tcpip_symbol_table + constants.BANG + "PartitionCount").address
+
+        # part_table is the actual partition table offset
```
ikelos: This feels like it should be possible to be a structure in one of the various tables we've loaded?  I think I'd find that a bit cleaner if you feel there's good reason to keep it this way, so not a showstopper, but if you're indifferent that's how I'd probably do it.  5:)

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+        for partition in partitions:
+            if partition.Endpoints.NumEntries > 0:
+                for endpoint_entry in cls.parse_hashtable(context,
+                                                          layer_name,
+                                                          partition.Endpoints.Directory,
+                                                          part_table_size,
+                                                          alignment,
+                                                          pointer_length):
+
+                    entry_offset = context.symbol_space.get_type(obj_name).relative_child_offset("ListEntry")
```
ikelos: Does this offset change?  If not, might be better to get it once outside the loop (or possibly both loops) in some way, just for efficiency.  If it can be different between structures, then no problem.  5:P

ikelos: This also feels really reminiscent of a `LIST_ENTRY` object (https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/extensions/__init__.py#L715), but I dunno if there's enough similarity there to make it worth trying to shoehorn into one of those.  Almost certainly not, is my guess.  5;)

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+                    endpoint = context.object(obj_name, layer_name = layer_name, offset = endpoint_entry - entry_offset)
+                    yield endpoint
+
+    @classmethod
+    def create_tcpip_symbol_table(cls,
+                                    context: interfaces.context.ContextInterface,
+                                    config_path: str,
+                                    layer_name: str,
+                                    tcpip_module: interfaces.objects.ObjectInterface) -> str:
+        """Creates
```
ikelos: Profound, but could possibly do with a bit more detail.  5;P

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+import logging
+import datetime
+from typing import Iterable, List, Optional, Callable
+
+from volatility.framework import constants, exceptions, interfaces, renderers, symbols, layers
+from volatility.framework.configuration import requirements
+from volatility.framework.renderers import format_hints
+from volatility.framework.symbols import intermed
+from volatility.framework.symbols.windows.extensions import network
+from volatility.framework.symbols.windows.pdbutil import PDBUtility
```
ikelos: In general, the style we try to go for would be `from volatility.framework.symbols.windows import pdbutil`, this just helps us avoid name space pollution and people importing `PDBUtility` from the wrong place...

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+                       context: interfaces.context.ContextInterface,
+                       layer_name: str,
+                       net_symbol_table: str,
+                       tcpip_symbol_table: str,
+                       tcpip_module_offset: int,
+                       pointer_length: int) -> (int, int):
+        """Finds the given image's port pools. Older Windows versions (presumably < Win10 build 14251) use driver
+        symbols called `UdpPortPool` and `TcpPortPool` which point towards the pools.
+        Newer Windows versions use `UdpCompartmentSet` and `TcpCompartmentSet`, which we first have to translate into
+        the port pool address. See also: http://redplait.blogspot.com/2016/06/tcpip-port-pools-in-fresh-windows-10.html
```
ikelos: Awesome!  References are ideal, thanks for including them!!!  5:D

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+        elif "UdpCompartmentSet" in context.symbol_space[tcpip_symbol_table].symbols:
+            # newer Windows versions since 10.14xxx
+            ucs = context.symbol_space.get_symbol(tcpip_symbol_table + constants.BANG + "UdpCompartmentSet").address
+            tcs = context.symbol_space.get_symbol(tcpip_symbol_table + constants.BANG + "TcpCompartmentSet").address
+
+            ucs_offset = cls.read_pointer(context, layer_name, tcpip_module_offset + ucs, pointer_length)
+            tcs_offset = cls.read_pointer(context, layer_name, tcpip_module_offset + tcs, pointer_length)
+
+            ucs_obj = context.object(net_symbol_table + constants.BANG + "_INET_COMPARTMENT_SET", layer_name = layer_name, offset = ucs_offset)
+            upp_addr = ucs_obj.InetCompartment.dereference().ProtocolCompartment.dereference().PortPool
```
ikelos: Yep, the more I see them, the more ugly they look.  Should definitely be possible without calling `dereference` explicitly.  I shall make it my mission, should it not already be possible, to ensure that you don't need to call it ever again (except it very limited circumstances, and even then it might be avoidable).  5:P

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+
+            tcs_obj = context.object(net_symbol_table + constants.BANG + "_INET_COMPARTMENT_SET", layer_name = layer_name, offset = tcs_offset)
+            tpp_addr = tcs_obj.InetCompartment.dereference().ProtocolCompartment.dereference().PortPool
+
+        else:
+            # this branch should not be reached.
```
ikelos: It should not, and yet I deeply approve of handling errors the right way, good work!  5:)

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+        vollog.debug("Found PortPools @ 0x{:x} (TCP) && 0x{:x} (UDP)".format(upp_addr, tpp_addr))
+        return upp_addr, tpp_addr
+
+    @classmethod
+    def list_sockets(cls,
+                       context: interfaces.context.ContextInterface,
+                       layer_name: str,
+                       nt_symbols: str,
+                       net_symbol_table: str,
+                       tcpip_module: interfaces.objects.ObjectInterface,
```
ikelos: Same reasoning as above (much, *much* further above).  5:)

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+        # create port pool objects at the detected address and parse the port bitmap
+        upp_obj = context.object(net_symbol_table + constants.BANG + "_INET_PORT_POOL", layer_name = layer_name, offset = upp_addr)
+        udpa_ports = cls.parse_bitmap(context, layer_name, upp_obj.PortBitMap.Buffer, upp_obj.PortBitMap.SizeOfBitMap // 8)
+
+        tpp_obj = context.object(net_symbol_table + constants.BANG + "_INET_PORT_POOL", layer_name = layer_name, offset = tpp_addr)
+        tcpl_ports = cls.parse_bitmap(context, layer_name, tpp_obj.PortBitMap.Buffer, tpp_obj.PortBitMap.SizeOfBitMap // 8)
+
+        # given the list of TCP / UDP ports, calculate the address of their respective objects and yield them.
+        for port in tcpl_ports:
+            if port == 0:
```
ikelos: The "pythonic" way would apparently be to say `if not port`, but I don't care that much, whichever you think is clearest.  5:)

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+                continue
+            for obj in cls.enumerate_structures_by_port(context, layer_name, net_symbol_table, port, upp_obj, "udp"):
+                yield obj
+
+    def _generator(self, show_corrupt_results: Optional[bool] = None):
+        """ Generates the network objects for use in rendering. """
+
+        # can this be checked via a PluginRequirement?
+        if not symbols.symbol_table_is_64bit(self.context, self.config['nt_symbols']):
+            raise exceptions.LayerException("This plugin currently only supports 64-bit memory images.")
```
ikelos: Very nice detail!  5:D

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+                               or renderers.UnreadableValue()))
+
+            elif isinstance(netw_obj, network._TCP_ENDPOINT):
+                vollog.debug("Found _TCP_ENDPOINT @ 0x{:2x}".format(netw_obj.vol.offset))
+                if netw_obj.get_address_family() == network.AF_INET:
+                    proto = "TCPv4"
+                elif netw_obj.get_address_family() == network.AF_INET6:
+                    proto = "TCPv6"
+                else:
+                    proto = "TCPv?"
```
ikelos: May also be worth a vollog saying unknown address family found?  That could be overkill, I guess it's only likely to happen with corrupt data...

volatility/framework/plugins/windows/netlist.py
```diff
@@ -0,0 +1,544 @@
...
+                               or renderers.UnreadableValue()))
+            else:
+                # this should not happen therefore we log it.
+                vollog.debug("Found network object unsure of its type: {} of type {}".format(netw_obj, type(netw_obj)))
+
+    def generate_timeline(self):
+        for row in self._generator():
+            _depth, row_data = row
+            # Skip network connections without creation time
+            if not isinstance(row_data[9], datetime.datetime):
```
ikelos: Hmmmm, I'd prefer being able to link these indices to the column names so they're more readable and stay in sync, but I'm not sure I've added a way of pulling the link between them in the TreeGrid interfaces.  Lemme know if you find a way, otherwise I may consider adding something to make it easier (although, that may then involve a framework version bump, so maybe not...).  5:S

ikelos (approved):
Well, frankly, this is one of the most well crafted plugins I've seen (and that includes the plugins I've written myself!)  5:)  You hit all the boxes (docstrings, error handling, the lot).  Several of my comments are just marvelling that you've included something most forget.  5:)  I think the only points are about avoiding manually parsing when the entire framework is designed to do just that, but it's really not a big point.  I'll approve this but give you time to make any changes based on the comments, in you'd like to.  We'll also probably hold of a little bit on merging this just because we've branched for our first release and we need to discuss whether we adopt gitflow as our chosen branching strategy (so whether this would get merged to master or a new develop branch), but we should get that figured out soon...  5:)

Thanks again, this is really an awesome plugin!  5:D

japhlange:
> I'll approve this but give you time to make any changes based on the comments, in you'd like to. 

Sure, give me a few days, I'll certainly address a few things soon :)

> Thanks again, this is really an awesome plugin! 5:D

No, thank you for the awesome framework which (if I do say so myself) really shows its power when pulling together different interfaces and interactions like I've did for this plugin and for all the help you've provided in Slack and beyond!

volatility/framework/plugins/windows/netlist.py
japhlange: Yeah, makes sense. There are some things in the current state which are not perfect, to say the least ;)

Which way would be best to overlay a pointer? Like this?

```python
part_table = context.object(net_symbol_table + constants.BANG + "pointer", 
                            layer_name = layer_name, 
                            offset = tcpip_module_offset + part_table_symbol)
```

volatility/framework/plugins/windows/netlist.py
japhlange: Didn't know that, great! Looks way better now.

volatility/framework/plugins/windows/netlist.py
japhlange: removed unnecessary calls ;)

volatility/framework/plugins/windows/netlist.py
japhlange: added it

volatility/framework/plugins/windows/netlist.py
japhlange: Yeah good point, but... the partition table itself is of dynamic length dependent on the amount of physical processor cores (cf. Art of Memory Forensics ch.11) and consists out of `number_of_cores` amount of `_PARTITION` structures which are defined in the symbol tables. So I've figured to do it like this, unless we have a clean way to support structs of dynamic length?

japhlange: I've added a comment to clear things up a bit, but I'm still open to suggestions to make this cleaner.

volatility/framework/plugins/windows/netlist.py
japhlange: You're absolutely correct, which is why I properly cross-referenced this in the Win7 symbol table but forgot in the later ones: https://github.com/volatilityfoundation/volatility3/pull/397/files#diff-d7569df42e8ea72b3c08ed120bdb06b59f7ad6541db3b3d06d26f3a7e39d0dd2L317

I've added it to the other ones as well. However I'm not entirely sure I can leverage anything out of that mapping in the code.

volatility/framework/plugins/windows/netlist.py
japhlange: true... :D

volatility/framework/plugins/windows/netlist.py
japhlange: Done!

volatility/framework/plugins/windows/netlist.py
ikelos: Yeah, although that defaults to being a pointer to void.  It might be worth defining a structure in the JSON which is then `pointer to actual structure` to make it easier to get around?  Otherwise I think you can do something like provide a `subtype`  parameter (which I think has to be a template, so use `get_type` to get it).

volatility/framework/plugins/windows/netlist.py
japhlange: Yes! I wish I could include more references, but almost all structures I use in this plugin are not documented anyplace I could find unfortunately.

volatility/framework/plugins/windows/netlist.py
japhlange: Done!

volatility/framework/plugins/windows/netlist.py
japhlange: Thanks! Was not sure which exception to use as `SymbolError` only takes one symbol in its main argument but I think it works out ok here.

volatility/framework/plugins/windows/netlist.py
ikelos: So we can change the length of arrays and unions, but if the structure changes too significantly then we can't really cope with it.  We also can't shuffle things down, so you still have to do size calculations for things that would come after it if they're packed.  If you construct the array, you can change its `count` attribute, and that'll update all the necessary bits and pieces: https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/objects/__init__.py#L547

volatility/framework/plugins/windows/netlist.py
ikelos: Sure, but just providing any references at all is awesome!  This was one of those complimentary comments I try to include (otherwise reviews can come across quite negative).  5:D

volatility/framework/plugins/windows/netlist.py
japhlange: I actually think I'd prefer the verbose way, because I found it often that the bit at index 0 (which would correlate with port 0) is set to 1. I dislike about the pythonic way that it mushes together 0 and None, because in this scenario a value of 0 is actually expected and I'd like to explicitly skip that. We do not expect a value of None, so no special handling for that.

However, I've changed it and added comments to clarify - I think this is better in the long run ;)

volatility/framework/plugins/windows/netlist.py
japhlange: Thanks! Although I hope to support x86 at some point when I (or someone else) gets to it.

volatility/framework/plugins/windows/netlist.py
japhlange: Yeah, but logging does not hurt. Added the debugging statement.

volatility/framework/plugins/windows/netlist.py
japhlange: I've added a dictionary if that helps? Still dependent on the order but it's at least indexed by a title now.

```python
    def generate_timeline(self):
        for row in self._generator():
            _depth, row_data = row
            row_dict = {}
            row_dict["Offset"], row_dict["Proto"], row_dict["LocalAddr"], row_dict["LocalPort"], \
            row_dict["ForeignAddr"], row_dict["ForeignPort"], row_dict["State"], \
            row_dict["PID"], row_dict["Owner"], row_dict["Created"] = row_data

            # Skip network connections without creation time
            if not isinstance(row_dict["Created"], datetime.datetime):
                continue
            row_data = [
                "N/A" if isinstance(i, renderers.UnreadableValue) or isinstance(i, renderers.UnparsableValue) else i
                for i in row_data
            ]
            description = "Network connection: Process {} {} Local Address {}:{} " \
                          "Remote Address {}:{} State {} Protocol {} ".format(row_dict["PID"], row_dict["Owner"],
                                                                              row_dict["LocalAddr"], row_dict["LocalPort"],
                                                                              row_dict["ForeignAddr"], row_dict["ForeignPort"],
                                                                              row_dict["State"], row_dict["Proto"])

            yield (description, timeliner.TimeLinerType.CREATED, row_dict["Created"])
```

japhlange: And another thing @ikelos about timeliner. I think this is great plugin, but when running it with `-vvvv` I see that it scans the layer very often to look for `nt_symbols`:
```
Level 9  volatility.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Timeliner.NetScan.nt_symbols
Level 9  volatility.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Timeliner.NetScan.nt_symbols
Level 9  volatility.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Timeliner.NetScan.nt_symbols
DEBUG    volatility.framework.automagic.pdbscan: Kernel base determination - searching layer module list structure
Level 9  volatility.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Timeliner.DllList.nt_symbols
Level 9  volatility.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Timeliner.DllList.nt_symbols
Level 9  volatility.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Timeliner.DllList.nt_symbols
DEBUG    volatility.framework.automagic.pdbscan: Kernel base determination - searching layer module list structure
```
and so on. I do not completely understand how Timeliner works yet but can this be optimized somehow? Maybe with some kind of sharing of nt_symbols?

volatility/framework/plugins/windows/netlist.py
japhlange: But wouldn't that require a pointer subtype for each actual structure (e.g. `pointer to _INET_PORT_POOL`, `pointer to _PARTITION` etc.)? I'm tending to say that there are about 5 occurrences in `netlist` right now and I think we can work around that for now with void pointers. Of course if there is any big advantage to using proper subtypes we can change that at any point (or if something changes later on or it gets more complex) but right now I'm not really feeling the effort, you know? :S

japhlange: Oh nevermind, I understood what you meant. However, I've got an error. When updating the partition count of the partition table with this it does not really get updated...
```python
# part_table is the actual partition table offset and consists out of a dynamic amount of _PARTITION objects
# ~ part_table = context.object(net_symbol_table + constants.BANG + "_PARTITION_TABLE", 
                            # ~ layer_name = layer_name, 
                            # ~ offset = part_table_addr)
part_table = context.object(net_symbol_table + constants.BANG + "pointer",
                            layer_name = layer_name,
                            subtype = context.symbol_space.get_type(net_symbol_table + constants.BANG + "_PARTITION_TABLE"),
                            offset = tcpip_module_offset + part_table_symbol)
part_count = int.from_bytes(context.layers[layer_name].read(tcpip_module_offset + part_count_symbol, 1), "little")
part_table.Partitions.count = part_count
# part_count == 4
# part_table.Partitions.count is still 1 (default value)?
```

volatility/framework/plugins/windows/netlist.py
ikelos: So I don't want to police stylistic choices where it doesn't really have an impact!  5:S  Happy for you to end up doing it whichever way you think is best.  5:)

volatility/framework/plugins/windows/netlist.py
ikelos: Hmmm, ok, that really should work.  Would you mind filing me an issue so I don't forget to investigate it please?  5:S

volatility/framework/plugins/windows/netlist.py
ikelos: It's only if someone comes to traverse a structure, the mapping will help it look up elsewhere.  You set the mapping when the symbol table is constructed so when you load it you need to map `nt_symbols` to `config['nt_symbols']` to make sure everything ties together.  I'm not sure if I entirely got what you wanted to leverage, but hopefully that answers general questions around it?  5:S

volatility/framework/plugins/windows/netlist.py
ikelos: Yeah, I think that's the right choice, and just putting one of the symbols would be the equivalent of raising twice with one of them first, so that's fine too.  5:)

volatility/framework/plugins/windows/netlist.py
ikelos: Yeah, that's much nicer!  5:)

So unfortunately timeliner runs multiple different plugins.  At the moment, we don't know what the automagic plugins might fill in (so far, it's usually just primary and nt_symbols, which are pretty common, but we can't guarantee those requirements will even be called that, so we need to go through the automagic every time unfortunately.  I'm not keen on locking down the requirements that timerliner plugins must have, because that'll essentially force people to work a particular way all for the sake of a single plugin.  We *could* theoretically hand the same context through to each plugin (so with layers already made, and transfer the config across *assuming* that they share names, the automagic would fill in the blanks, so should be quick if there's nothing new needed), but I'd definitely put that under a flag to explain what was going on, and it would technically allow a malicious plugin to muddy the context before the next plugin gets given it...  I dunno, might need some thinking about, but happy to have it added as potential idea to the issue tracker?

volatility/framework/plugins/windows/netlist.py
japhlange: Well, let's approach it from this point. The use case is to parse a hash table and find its values. Entries of the hash table which are valid can be detected by them being not pointers to themselves, but different ones. This is why I hacked together this function here.

The reason behind the mask is that e.g. the address 0xc88d3b427b70 is read as 0xffffc88d3b427b70 and I needed a quick workaround :)

Just based upon this requirement, maybe we can find a better way to parse the table without hardcoded masks? :D

volatility/framework/plugins/windows/netlist.py
japhlange: addressed in #400

volatility/framework/plugins/windows/netlist.py
japhlange: Well, luckily I've already done that in `netscan` which we depend on here with `net_symbol_table`: https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/plugins/windows/netscan.py#L237

So I think this should be fine for now - while we do not need to access `LIST_ENTRY` attributes right now it's still smoother to have this properly mapped.

volatility/framework/plugins/windows/netlist.py
japhlange: I dove into this again. What we have here is a structure of type `_RTL_DYNAMIC_HASH_TABLE`'s `Directory` entries. The struct is defined as follows: http://librecrops.github.io/lost-sdk/files/RTL_DYNAMIC_HASH_TABLE.h.html

When we follow the `Directory` pointer we see the following in memory:
```
(primary) >>> dq(rtl_dyn_hash_table.Directory)
0xc88d39c81800    ffffc88d39c81800 ffffc88d39c81800    ....9... ....9...
0xc88d39c81810    ffffc88d39c81810 ffffc88d39c81810    ....9... ....9...
0xc88d39c81820    ffffc88d39c81820 ffffc88d39c81820    ....9... ....9...
0xc88d39c81830    ffffc88d39c81830 ffffc88d39c81830    ....9..0 ....9..0
0xc88d39c81840    ffffc88d39c81840 ffffc88d39c81840    ....9..@ ....9..@
0xc88d39c81850    ffffc88d39c81850 ffffc88d39c81850    ....9..P ....9..P
0xc88d39c81860    ffffc88d39c81860 ffffc88d39c81860    ....9.. ....9..
<..>
0xc88d39c81950    ffffc88d39c81950 ffffc88d39c81950    ....9..P ....9..P
0xc88d39c81960    ffffc88d3d649228 ffffc88d3d649228    ....=d.( ....=d.(
0xc88d39c81970    ffffc88d39c81970 ffffc88d39c81970    ....9..p ....9..p
```
and we are looking for values like at `0xc88d39c81960` which are not their own address. Unfortunately if we look at the definition above there is no defined data structure for those entries. While every instance of them looks like a `_LIST_ENTRY` there is no NT-symbols-defined struct for the shown list.  ( There are some nt structs which look like they would fit, for instance `_RTL_DYNAMIC_HASH_TABLE_ENTRY` (size of 24 byte, does not fit) or `_SINGLE_LIST_ENTRY` (size of 8 bytes, would fit twice) but I don't think they help much given that we would require a array-like struct for easier parsing...)

I see two options: either parse that list manually like I do it right now or define a custom "list of `_LIST_ENTRY`" in my netscan symbol tables so we can benefit from... a little bit more readability for a lot more effort?

I tend towards option 1. Or maybe I'm overlooking something :D

ikelos: Parsing it manually is fine, but given it's an address, you can cast it as a pointer and it should mask it to the appropriate size of the address space.  It may not be worthwhile, but hardcoded things will inevitably need changing, so if we can just cast those offsets as pointers (possibly to a known structure, but even if not), then the framework should handle the bulk of it for you?  That doesn't require hand coding a new full structure, but should give us all the benefits of letting the framework look after all the details.  Does that sound workable?  5:)

volatility/framework/plugins/windows/netlist.py
ikelos: Cool, that looks good then.  5:)

volatility/framework/plugins/windows/netlist.py
japhlange: I've reworked it into comparing `pointer == pointer.vol.offset` which works. Is this stable enough or is there a prettier way?
```python
        for index in range(ht_length):
            current_addr = ht_offset + index * alignment
            current_pointer = context.object(net_symbol_table + constants.BANG + "pointer",
                                    layer_name = layer_name,
                                    offset = current_addr)
            # check if addr of pointer is equal to the value pointed to
            if current_pointer.vol.offset == current_pointer:
                continue
            yield current_pointer
```

volatility/framework/plugins/windows/netlist.py
japhlange: done

japhlange:
I've removed the 64-bit check as I've added support for Windows 7 x86. Unfortunately I have not been able to make Windows 10 32-bit work (tried it for builds 10240, 15063 and 16299) as following the normal algorithm yields way less results (if any) than it should. In order to find out why I'd need to resort to debugging the system which in turn requires me to setup VMs again.

For documentation's sake: enumerating the ports from the bitmap and jumping to the correct `_PORT_ASSIGNMENT` object do work. However, checking the `_PORT_ASSIGNMENT_LIST` array shows it's empty even though we expect values there. No idea if there is a further translation step for 32-bit.

I'll probably get to it at some point but honestly, I'm not sure how important 32-bit is today anymore except of course for backwards compatibility...

volatility/framework/plugins/windows/netlist.py
japhlange: Yeah not sure, don't know how much effort we'd save by transferring over the context. Might be a thing to keep in mind and test at some point.

Regarding potentially malicious plugins: I think we should consider the threat model here - if a malicious plugin is installed, what use case would it have to manipulate the context? It needs user context to install itself into vol's directory at which point it could manipulate vol's core source code too. Furthermore it would be able to manipulate the context of other plugin's only if called via `Timeliner` and not if the other plugins are called directly (except of course the core code is altered too...). Not so sure about this attack vector, but maybe I'm just not creative enough :D

And maybe the base context or base symbol tables or whatever could be set to read-only once they've been created for the first time. Of course, that read-only lock would only be set via user-run code, so a user-run malicious plugin could circumvent that too probably...

volatility/framework/plugins/windows/netlist.py
japhlange: Good points! I removed all complex / object arguments and changed them for the respective offsets. It'd be great if you could look at the current state ;)

volatility/framework/plugins/windows/netlist.py
ikelos: Yep, that looks fine to me.  5:)

volatility/framework/plugins/windows/netlist.py
ikelos: Well, ok, malicious is probably the wrong word.  I've seen people write _misguided_ plugins that think the easiest way of fixing things they want to have happen would be to just tinker with the context for their own purposes (I'll just make it *look* like something or other).  So not necessarily badly intentioned, just not a thing we want to make easy for people to code lazily.

I think there'd be a lot of machinery involved in letting people do what they want to the context without messing with it.  What we could probably do is run the automagic over the context and then pass that context on, and re-run the automagic (which would already have a lot of the requirements pre-populated, so shouldn't need to fulfill them again).  I think there's the ability to clone a config, so it would reduce the risk to a malicious (or misguided) automagic (which as you point out, is unlikely), so that might be a nice compromise...  That could save a bunch of time.  I'll look into mocking that up at some point...  5:)

ikelos:
> I'll probably get to it at some point but honestly, I'm not sure how important 32-bit is today anymore except of course for backwards compatibility...

Yep, no problem, I don't think it should hold up the plugin, but we're looking at changing the git workflow we use, so I probably won't merge it until that's done.

volatility/framework/plugins/windows/netlist.py
ikelos: Yep, very happy with that, thanks!  5:)

japhlange:
@ikelos I've been thinking about the plugin's name - in vol2, there is `netscan` and `netstat` for Linux and `netscan` for Windows. Vol3 currently has `netscan` for Windows and nothing yet for Linux.

I initially chose `netlist` because the symmetry with `pslist` and `psscan`, but if the Linux plugin gets ported at some point it would make sense from a compatibility-to-vol2 POV to name it `netstat` as well. Then we'd have `netlist` for Windows and `netstat` for Linux, serving the same purpose but being inconsistently named.

I know that this is not a complex issue, but maybe it would be better to proactively think about this before changing anything breaks stuff. One option would be to rename this plugin to `netstat` or vice versa with the Linux one. Any opinions?

ikelos:
I don't really use the plugins often enough to know which should mimic the other, but I agree there should probably be parity across the operating systems.  I can bring it up at our next meeting, but given netstat's already in use in vol2, and that's also the name of the command in both operating systems, it probably makes sense to go with that one...

japhlange:
I've renamed the file and plugin and it should work fine now. Seeing as Windows 2004 (19041) is supported as well - haven't gotten to 20H2 yet - I'd say I'm happy with this, for now :)

ikelos:
Looks good, thanks for the amazing contribution!  5:D

PR #401, opened 2020-12-22
Objects: Avoid reconstructing pointed objects

### LINKED ISSUE #400: Changing the attribute of an object via a pointer does not work
**Describe the bug**
As mentioned in https://github.com/volatilityfoundation/volatility3/pull/397#discussion_r547307215.

When trying to access an object constructed via a pointer like `part_table` in the example below I can access the object's attributes like `Partitions`, but changing them does not appear to change the actual value, as the value of `count` like in the last line is still 1.
```python
# part_table is the actual partition table offset and consists out of a dynamic amount of _PARTITION objects
# ~ part_table = context.object(net_symbol_table + constants.BANG + "_PARTITION_TABLE", 
                            # ~ layer_name = layer_name, 
                            # ~ offset = part_table_addr)
part_table = context.object(net_symbol_table + constants.BANG + "pointer",
                            layer_name = layer_name,
                            subtype = context.symbol_space.get_type(net_symbol_table + constants.BANG + "_PARTITION_TABLE"),
                            offset = tcpip_module_offset + part_table_symbol)
part_count = int.from_bytes(context.layers[layer_name].read(tcpip_module_offset + part_count_symbol, 1), "little")
part_table.Partitions.count = part_count
# part_count == 4
# part_table.Partitions.count is still 1 (default value)?
```
Here is the definition:
```json
"_PARTITION_TABLE": {
    "fields": {
        "Partitions": {
            "offset": 0,
            "type": {
                "count": 1,
                "kind": "array",
                "subtype": {
                    "kind": "struct",
                    "name": "_PARTITION"
                }
            }
        }
    },
    "kind": "struct",
    "size": 128
}
```
When changing the value of `count` when the part_table is constructed via the commented-out section in the example above it works.

**Context**
Volatility Version:  2.0.0
Operating System:  Manjaro Linux
Python Version:  3.8
Suspected Operating System:  Win10x64_16299
Command:  see code example above.

**To Reproduce**

1. Construct a struct in memory via a pointer and its subtype, then change its value.

ikelos:
Fixes #400

### DIFF
--- a/volatility3/framework/objects/__init__.py
+++ b/volatility3/framework/objects/__init__.py
@@ -3,6 +3,7 @@
 #
 
 import collections
+import functools
 import logging
 import struct
 from typing import Any, ClassVar, Dict, List, Iterable, Optional, Tuple, Type, Union as TUnion, overload
@@ -307,6 +308,7 @@ class Pointer(Integer):
         value = int.from_bytes(data, byteorder = endian, signed = signed)
         return value & mask
 
+    @functools.lru_cache(3)
     def dereference(self, layer_name: Optional[str] = None) -> interfaces.objects.ObjectInterface:
         """Dereferences the pointer.
 


ikelos:
This has a couple of attached questions.  Firstly, does the lru_cache hang onto the data long enough to never give weird results (ie, looking like a fixed object that then gets forgotten and recreated randomly)?  I've set the limit at 3 because we're unlikely to instantiate a pointer at the same offset against 3 different layer names.  Could we do better by doing it ourselves, or should I just set the limit higher at the possible expense of memory?  We also need to consider the implications for adding live memory support in the future, do we need to add cache invalidation support to the pointer now?  So not quite ready for a merge without a review...

ikelos:
It'd be nice to get this in the release, but it feels like it might be too much of a rush, and it's not a critical issue (because it can be worked around and we've only had one instance where it's caused issue), but it is a reasonably large change to the way volatility does things, so getting it in the first release would be handy.  Just needs an extra pair of eyes and those questions confirming/answering...  5:)

PR #406, opened 2020-12-22
Plugin zips

ikelos:
This should add support to pick up plugin bundles, so that users can just drop a zip file into their plugins directory, rather than needing to unpack files and put them in the right places (particularly helpful for plugins that come with their own JSON files).  It could do with a review/kicking the tyres, etc. though...


ikelos:
No complaints, so I guess we'll put it in and see what buildbot thinks...

PR #410, opened 2020-12-30
Feature/disable caching

ikelos:
This feature adds support for (very early) deciding as to whether caching of layer reads should be enabled throughout the codebase.  It's not ideal because the main toggle *doesn't* live in volatility.frameworks.constants like it should, because volatility.framework imports other parts of the framework that already need to know whether the caching is enabled or not.  To avoid this issue, the toggle lives just in `volatility`.

### DIFF
--- a/volatility/cli/__init__.py
+++ b/volatility/cli/__init__.py
@@ -22,6 +22,11 @@ import traceback
 from typing import Dict, Type, Union, Any
 from urllib import parse, request
 
+import volatility
+
+if "--live" in sys.argv:
+    volatility.CACHING = False
+
 import volatility.plugins
 import volatility.symbols
 from volatility import framework
@@ -134,6 +139,7 @@ class CommandLine:
                             help = "Log output to a file as well as the console",
                             default = None,
                             type = str)
+        parser.add_argument("--live", help = "Disable caching of layer reads", action = 'store_true')
         parser.add_argument("-o",
                             "--output-dir",
                             help = "Directory in which to output any generated files",
--- a/volatility/cli/__init__.py
+++ b/volatility/cli/__init__.py
@@ -107,7 +107,7 @@ class CommandLine:
                             default = None,
                             type = str)
         parser.add_argument("--parallelism",
-                            help = "Enables parallelism (defaults to processes if no argument given)",
+                            help = "Enables parallelism (defaults to off if no argument given)",
                             nargs = '?',
                             choices = ['processes', 'threads', 'off'],
                             const = 'processes',
--- a/volatility/framework/caching.py
+++ b/volatility/framework/caching.py
@@ -1,0 +1,9 @@
+import functools
+import volatility
+
+
+def lru_cache(*args, **kwargs):
+    if not volatility.CACHING:
+        return lambda x: x
+    raise Exception
+    return functools.lru_cache(*args, **kwargs)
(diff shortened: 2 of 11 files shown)


volatility/framework/caching.py
```diff
@@ -0,0 +1,9 @@
+import functools
+import volatility
+
+
+def lru_cache(*args, **kwargs):
+    if not volatility.CACHING:
+        return lambda x: x
+    raise Exception
```
npetroni: I must be missing something... why is this exception here?

ikelos: That's entirely my bad, it's so I could determine when the wrapper was being applied and make sure the caching value had been set correctly (setting it to late would cause some things to have been wrapped by it early).  Good spot, thanks!  5:D

npetroni: Works now, thanks.

volatility/cli/__init__.py
```diff
@@ -22,6 +22,11 @@
 from typing import Dict, Type, Union, Any
 from urllib import parse, request
 
+import volatility
+
+if "--live" in sys.argv:
```
npetroni, line 27: Hmm, this doesn't seem right. Why isn't this handled by the argparse logic below the --live option?

npetroni:
What if we tried something more localized to caching, like this?

https://github.com/volatilityfoundation/volatility3/tree/feature/disable-caching-try2

volatility/cli/__init__.py
ikelos: So I concur, it's thoroughly ugly, but unfortunately the chain of imports (of `volatility.framework` upwards) would import and therefore wrap all the calls before we've set the flag.  So we've got two options:
* for every read check whether we're allowed to use the cache, and then use it or not
* wrap everything early with a wrapper that chooses whether to cache or not

From an efficiency perspective, the second option is better (I haven't timed to see how much, and chances are it's a couple lookups and an if, so should be pretty fast, but given our runs regularly do hundreds of thousands of reads, I went with the slight ugliness in the CLI to set the flag early enough.  I might look into a way so that change the variable late will throw an exception (so that UI developers know they need to set it *really* early, I'm just not sure how I'd go about making a variable read-only after a certain point...

Anyway, that's why it needs to be read out of sys.argv so early.  Once the parser's run, we've already imported everything to be able to enumerate it properly.

ikelos:
That won't trip in time (and you can add an exception in to show that changing the flag would have no effect).  The wrapping and therefore that if statement are evaluated as the files are imported.  By the time you've reached there, they're all wrapped already.  You'd need to write your own `lru_cache` method which would check whether to use the cache or not each time if you wanted to be changable dynamically.

ikelos:
Another option may be to do through and explicitly unwrap all wrapped objects (or maybe set their cache size to 0?), but that would be monkeying deep inside of python and probably make it difficult to distinguish if there were some values we wanted cached and others we didn't...

npetroni:
We can discuss it today. The exception trips for me, so I am missing something.

ikelos:
Yeah, no problem.  Adding the exception in after the if statement in `disable-caching-try2`, I get the exception whether I'm using `--live` or not, meaning the flag isn't getting interpreted before the wrapper's being called (and the caching is always being applied), but I'm happy to go through it at the meeting.  5:)

ikelos:
Following a discussion, we've considered that performance (particularly for the common case) is important, but this approach is a bit complex/ugly so we're going to defer making this change for a bit.

atcuno:
Do you want to keep this open still @ikelos ?

ikelos:
We can close the PR, but we should leave the branch/code in place to remind us what works been done on it already for if we come back to revisit it...

Issue #423, opened 2021-01-13
windows.netscan doesn't generate any data for win10-x86 samples


mtressler:
**Describe the bug**
When the plugin windows.netscan is run on an x86 sample it runs without error but no data is output.

**Context**
Volatility Version:  release/v2.0.0
Operating System:  Windows/WSL
Python Version:  3.8.2
Suspected Operating System:  win10-x86
Command:  python3 vol.py -f samples/win10-x86-2016-07-08.lime windows.netscan

**To Reproduce**
Run netscan plugin on x86 sample

**Expected behavior**
Should output all network objects in the sample

ikelos:
Sounds like one for @japhlange to look at when he gets time (I understand he's busy at the moment)...

japhlange:
Do you know the Windows build? It'd be great if you could either run `winver` on the live machine or the `windows.info` vol3 plugin :)

mtressler:
Do you want the build of the sample or my machine?

japhlange:
Of the memory sample.

mtressler:
This is what I got from windows.info. I think this is the version 
Major/Minor     15.10586

ikelos:
I can confirm on what appears to be the same image that no results are returned.  If it helps, I think the symbols it's using are: `ntkrpamp.pdb/44B89EEAB81F466591D6570331648068-1`.

I get a bunch of errors in the verbose log along the lines of:
```
DEBUG    volatility.plugins.windows.netscan: Found netw obj @ 0x826f538f of assumed type <class 'volatility.framework.symbols.windows.extensions.network._TCP_LISTENER'>
DEBUG    volatility.framework.symbols.windows.extensions.network: netw obj 0x826f538f invalid due to invalid address_family None
DEBUG    volatility.plugins.windows.netscan: Found netw obj @ 0x82713462 of assumed type <class 'volatility.framework.symbols.windows.extensions.network._TCP_ENDPOINT'>
DEBUG    volatility.framework.symbols.windows.extensions.network: <class 'volatility.framework.symbols.windows.extensions.network._TCP_ENDPOINT'> 0x82713462 invalid due to invalid tcp state 2189234800
DEBUG    volatility.plugins.windows.netscan: Found netw obj @ 0x85254350 of assumed type <class 'volatility.framework.symbols.windows.extensions.network._TCP_LISTENER'>
[...]

PR #428, opened 2021-01-14
Adds netscan symbols for Win10 10586 x86

japhlange:
As per #423 the symbols for Win10 10586 x86 were missing.

No results were generated because all three object's validations failed (TCP endpoints, TCP listeners and UDP listeners). This is because the validation function checks for address family (protocol version, IPv4/v6) among other things and the pointers to the structures containing the protocol version could not be followed.

Note: this does not fix `netstat` for that Windows version. But `netscan` should work again.

@ikelos @mtressler it'd be great if you could try out the symbols with the memory samples of that Windows version you have available. Just run `netscan` against it and see if it produces results which seem correct.

We also don't have to merge this immediately. I just wanted to publish the changed offsets as soon as possible so we can check and change them if any of them are still wrong.


mtressler:
Just tested this on my x86 sample and it appears to working correctly

japhlange:
If you are unsure and you have access to the original system you could try to open a few known network connections, list them with `netstat.exe` and see if they appear in the volatility `netscan` results.

mtressler:
I don't have access to the machine. This sample is used for regression testing so I was just given this sample to use when creating the tests. It looks similar to what I've seen from other samples in that set.

ikelos (approved):
Looks fine to make it into develop, thanks!  5:)

PR #430, opened 2021-01-20
Rename the base namespace as volatility3

ikelos:
This is a big shift, but would allow us to reduce the version number to 1.0.0, and keep the pypi package name, github repo and import values the same across the board.  I'm kinda warming to it, but still only kinda, so it's entirely up for comment at this point.

It may also take some work to merge it back into develop, but happy to deal with that if/when we come to it...


npetroni:
I think I really like this approach. I think the main benefits are (1) consistency, (2) version reset, and (3) makes it impossible to confuse code snippets from Volatility 2 with Volatility 3.

ikelos:
Ok, I'm happy to go for it, so I'm merge it into release, then we probably need to broadcast the change on the channel somewhere and then merge it into develop...  5:S

PR #438, opened 2021-01-28
add GitHub Action to build PyPI and source artifacts

npetroni:


ikelos:
Cool, looks good, just a question of quantity.  Do we want pypi packages for every push/pull request to develop?  Seems like it might create a lot of packages, but I'm fine with that if it'll be useful and we've got unlimited storage.  I just don't know if we want pypi to expose packages for all of those?  Possibly just master and release branches (and PRs to them)?

I would like tests and so on to run for develop and all PRs, I just dunno if pypi packages are necessary?

npetroni:
> Cool, looks good, just a question of quantity. Do we want pypi packages for every push/pull request to develop? Seems like it might create a lot of packages, but I'm fine with that if it'll be useful and we've got unlimited storage. I just don't know if we want pypi to expose packages for all of those? Possibly just master and release branches (and PRs to them)?

It's important to distinguish artifacts that are created and stored in github from what is published upstream to PyPI. Nothing in this workflow publishes to production -- and we plan to do that manually to start. Even if we automated publishing, it would only be from master. pypi would not be "exposing" packages for all of these builds.

Regarding storage limits, my understanding is we have enough space and the builds roll off every 90 days. Each build is only about 1.5M.

> I would like tests and so on to run for develop and all PRs, I just dunno if pypi packages are necessary?

I have no problem removing package builds for develop, but I don't think the concerns you raised are likely barriers. If we ever used this framework for testing on develop, we'd need to turn them back on so that we can test the actual package. For now, testing lives outside of github, so the packages are not strictly necessary.

ikelos:
Ok, the 90 days is the bit I didn't know about (and you're right, I was seeing pypi in the title and assuming that they might all end up there somehow).  Yep, all sounds good then, thanks!  5:)

Issue #440, opened 2021-01-29
Windows.pslist


lprat:
**Describe the bug**
```
$python3 /opt/volatility3/vol.py -vvvvvvv -f test.ram windows.pslist
Volatility 3 Framework 1.0.0
INFO     root        : Volatility plugins path: ['/opt/volatility3/volatility3/plugins', '/opt/volatility3/volatility3/framework/plugins']
INFO     root        : Volatility symbols path: ['/opt/volatility3/volatility3/symbols', '/opt/volatility3/volatility3/framework/symbols']
Level 6  volatility3.framework: Importing from the following paths: /opt/volatility3/volatility3/plugins, /opt/volatility3/volatility3/framework/plugins
Level 6  volatility3.framework: Importing from the following paths: /opt/volatility3/volatility3/framework/automagic
Level 7  root        : Cache directory used: /root/.cache/volatility3
INFO     volatility3.framework.automagic: Detected a windows category plugin
Level 6  volatility3.framework: Importing from the following paths: /opt/volatility3/volatility3/framework/layers
INFO     volatility3.framework.automagic: Running automagic: ConstructionMagic
Level 6  volatility3.framework: Importing from the following paths: /opt/volatility3/volatility3/framework/layers
Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.primary
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.nt_symbols
Level 6  volatility3.framework: Importing from the following paths: /opt/volatility3/volatility3/framework/layers
Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.primary
Level 6  volatility3.framework: Importing from the following paths: /opt/volatility3/volatility3/framework/layers
Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.primary
Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.primary
Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList
Level 6  volatility3.framework: Importing from the following paths: /opt/volatility3/volatility3/framework/layers
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.nt_symbols
Level 6  volatility3.framework: Importing from the following paths: /opt/volatility3/volatility3/framework/layers
[...]

PR #442, opened 2021-01-29
Solution to issue #440

lprat:
Hi,

I propose solution to my problem, by add option "--no-fail".

Thank for your help.
Sincerly,
Lionel


lprat:
remove for #465

PR #444, opened 2021-01-31
rename master branch to stable in github workflows

npetroni:


ikelos (approved):
Sorry for the delay, just wanted to check that we intended to drop PRs from develop from things that get artifacts built?  I don't know how much use they'd be so I'm happy either way.  5:)

ikelos:
So happy that I just merged it.  5:P

PR #448, opened 2021-02-01
Windows: Memmap include file offset

### LINKED ISSUE #447: Memmap doesn't output the file offset appropriately
The windows.memmap plugin doesn't let users know how a virtual offset maps to the file offset, but instead provides two copies of the virtual offset.

ikelos:
Fixes #447


ikelos:
Shouldn't be contentious, but I'll wait a few days before pushing it to develop.  5:)

PR #452, opened 2021-02-03
fixes for crash.py

jxwenger:


volatility/framework/layers/crash.py
```diff
@@ -42,7 +37,11 @@ def __init__(self, context: interfaces.context.ContextInterface, config_path: st
         self._context = context
         self._config_path = config_path
         self._page_size = 0x1000
-        self._base_layer = self.config["base_layer"]
+        try: 
+            self._base_layer = self.config["base_layer"]
+        except KeyError:
+            self._base_layer = 'base_layer'
```
ikelos, line 48: This doesn't seem like the right solution to this?  Better to set a default in the requirement, but then what would the default be?  I'm fine with logging the issue and providing a nicer error, but I don't think this will fix the problem?

volatility/framework/layers/crash.py
```diff
@@ -1,7 +1,3 @@
-# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
-# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
-#
```
ikelos, line 3: Hmmm, any reason to remove the license from this file?

volatility/framework/layers/crash.py
```diff
@@ -71,30 +70,33 @@ def __init__(self, context: interfaces.context.ContextInterface, config_path: st
...
+                                     layer_name = self._base_layer)
+            offset = self.headerpages                      
+            header.PhysicalMemoryBlockBuffer.Run.count = header.PhysicalMemoryBlockBuffer.NumberOfRuns
+            for x in header.PhysicalMemoryBlockBuffer.Run:
+                segments.append((x.BasePage * 0x1000, offset * 0x1000, x.PageCount * 0x1000, x.PageCount * 0x1000))
+                # print("Segments {:x} {:x} {:x}".format(x.BasePage * 0x1000,
+                #                  offset * 0x1000,
+                #                  x.PageCount * 0x1000))
+                offset += x.PageCount
+
```
ikelos, line 94: Should there be an else here?

volatility/framework/layers/crash.py
```diff
@@ -146,7 +146,7 @@ def _load_segments(self) -> None:
                 offset += x.PageCount
 
         elif self.dump_type == 0x05:
-            summary_header.BufferLong.count = (summary_header.BitmapSize + 31) // 32
+            summary_header.BufferLong.count = (summary_header.BitmapSize + 31) // 32 + 0x2000
```
ikelos: Could you document why you're making this change please?  Why `0x2000` in particular?  Is this the offset for the first two pages?

volatility/framework/layers/crash.py
```diff
@@ -166,11 +167,15 @@ def _load_segments(self) -> None:
                         segments.append((start_position * 0x1000, mapped_offset, length, length))
                         mapped_offset += length
 
+
                 # Finish it off
-                if bit_position == (len(summary_header.BufferLong) * 32) - 1 and current_bit == 1:
+                if (bit_position == (len(summary_header.BufferLong) * 32) - 1 or bit_position == (len(summary_header.BufferLong) * 32) - 1 -32*0x2000) and current_bit == 1:
```
ikelos: This is now a very complex expression, so I'd prefer it were cut down to make it more easily readable and again documented to explain what the changes are for.

volatility/framework/plugins/windows/crashinfo.py
```diff
@@ -0,0 +1,42 @@
...
+                                                     architectures = ["Intel32", "Intel64"]), 
+            ]
+        
+    def _generator(self, segments):
+        for seg in segments:
+            yield(0,(seg[0],seg[1],seg[2]))
+
+    def run(self):
+
+        if self.config["primary.memory_layer.class"] == "volatility.framework.layers.crash.WindowsCrashDump32Layer":
```
ikelos: This would be better to access the `primary` layer and then traverse down its dependencies and verifying their instances.

volatility/framework/plugins/windows/crashinfo.py
```diff
@@ -0,0 +1,42 @@
...
+            crashdump = crash.WindowsCrashDump32Layer(self.context, self.config_path, self.config['primary'])
```
ikelos: Again, better to access the existing layer rather than creating a new one?

volatility/framework/plugins/windows/crashinfo.py
```diff
@@ -0,0 +1,42 @@
+  
+# This file is Copyright 2021 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import logging
+from volatility.framework import interfaces, renderers
```
ikelos, line 7: The namespace was very recently changed from `volatility` to `volatility3`.  We took the opportunity to reset the framwork verison as well, so `_required_framework_version` should be set to `(1, 0, 0)`.

volatility/framework/plugins/windows/crashinfo.py
```diff
@@ -0,0 +1,42 @@
...
+
+        elif self.config["primary.memory_layer.class"] == "volatility.framework.layers.crash.WindowsCrashDump64Layer":
+            crashdump = crash.WindowsCrashDump64Layer(self.context, self.config_path, self.config['primary'])
+        
+        else:
+            vollog.log(constants.LOGLEVEL_VVVV, "Error: Windows crashdump file needed")
+            return
+
+
+        return renderers.TreeGrid([("StartAddress", int),("FileOffset", int),("Length", int)],self._generator(crashdump._segments))
```
ikelos: This strongly ties crashinfo to the implementation of crashdump.  Accessing a protected member (starting with `_`) is generally a bad idea and will break if the implementation decides to change any of its internals.  It would be much better to use the `mapping` method which is part of the public API...

ikelos (changes requested):
Thanks very much for your additions, but they could do with a little more description (ideally in comments in the code, but even just in the pull request).  I'd also prefer to keep fixes and new code separate so that they're easier to review, but since this plugin was small it's ok.  5:)

There's a couple of areas where the plugin could be improved by using more generic calls and the public API rather than directly accessing various components that rely on things specifically working a particular way.  If you could get these changed up, and the explanations make sense, then we can get this merged.  5:)

volatility/framework/layers/crash.py
```diff
@@ -168,8 +176,8 @@ def _load_segments(self) -> None:
                         mapped_offset += length
 
 
-                # Finish it off
-                if (bit_position == (len(summary_header.BufferLong) * 32) - 1 or bit_position == (len(summary_header.BufferLong) * 32) - 1 -32*0x2000) and current_bit == 1:
+                # Find the last segment in a file which will be at the end or two pages from the end
+                if (bit_position == bitmap_len - 1 or bit_position == bitmap_len - 1 -32*0x2000) and current_bit == 1:
```
ikelos, line 180: So I think the -32*0x2000 is still a little difficult to figure?  Any chance of a bit more clarification around what's going on here please?

volatility/framework/plugins/windows/crashinfo.py
```diff
@@ -22,21 +22,12 @@ def get_requirements(cls):
                                                      architectures = ["Intel32", "Intel64"]), 
             ]
         
-    def _generator(self, segments):
-        for seg in segments:
-            yield(0,(seg[0],seg[1],seg[2]))
+    def _generator(self, layer):
+        for seg in layer.mapping(0x0, layer.maximum_address, ignore_errors = True):
```
ikelos: Ideally, for readability this would be `for offset, length, mapped_offset, _, _ in ...`, referring to seg indices doesn't let people understand what's going on.

volatility/framework/layers/crash.py
```diff
@@ -146,13 +152,15 @@ def _load_segments(self) -> None:
                 offset += x.PageCount
 
         elif self.dump_type == 0x05:
-            summary_header.BufferLong.count = (summary_header.BitmapSize + 31) // 32 + 0x2000
+            #Add 0x2000 as some bitmaps are too short by one offset
+            summary_header.BufferLong.count = (summary_header.BitmapSize + 31) // 32 + 0x2000
```
ikelos, line 156: I'm still a little confused why the BufferLong calculation isn't right and more how adding 0x2000 is the right solution?  If they're short by a single offset, then shouldn't it be something like `(summary_header.BitmapSize + 32) // 32` but then the question is why is the wrong value included in the calculation?  This should essentially be calculating `math.ceil(BitmapSize / 32)`, so is that value inaccurate, or is something else going on?

ikelos (changes requested):
Much better, thanks!  5:)  Still some bits that are difficult to understand though, any chance you could help with those?

volatility/framework/layers/crash.py
jxwenger: Yeah, I'm not entirely sure why the calculation is wrong either, but it is definitely off by that amount and the issue doesn't seem to be anywhere in crash.py. I'm not sure where this calculation is actually done be but if you know I could take a look

iMHLv2:
This MR was superseded by https://github.com/volatilityfoundation/volatility3/pull/498

PR #453, opened 2021-02-04
Replace \xe2\x80\x94 with normal "-" in README.md

leohearts:
A strange problem (?


ikelos (approved):

ikelos:
Yep, nice spot.  I guess the editor that created it decided that hypens were actually sentence breaks and inserted unicode characters (thete's also a unicode apostrophe in there too).  Thanks for the spot, happy to merge this in  5:)

PR #454, opened 2021-02-05
Layers: Don't map intel out-of-bounds addresses

ikelos:
Currently the intel space could be asked to map a virtual address greater than 48-bits, which it would do, causing potential errors.  This adds a check before translation, and optimizes requests to static values so they're not recalculated each time (not sure how much difference it will make, but hopefully a slight speed up given how often they're called.


ikelos:
I'll probably merge this in a few days.

PR #458, opened 2021-02-23
Fix bytes to str concat - resulting in error

AsafEitani:
fix bytes to str concat


ikelos (approved):
Yep, looks good, we really need to get type signatures into the code to spot things like that.  Thanks very much for finding it!  5:)

PR #459, opened 2021-02-23
changing PID field names

AsafEitani:
Changing the pid name field of the result to match the PID of all other plugins


ikelos (approved):
Yep, looks fine, thanks for catching it...  5;)

PR #465, opened 2021-03-04
Add no-fail flag to the CLI

lprat:
Hi,

I propose solution to my problem, by add option "--no-fail".

Thank for your help.
Sincerly,
Lionel

### DIFF
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -218,7 +218,7 @@ class CommandLine:
 
         if partial_args.clear_cache:
             framework.clear_cache()
-            
+
         if partial_args.no_fail:
             constants.FAIL_ON_ERROR = False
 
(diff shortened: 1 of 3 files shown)


ikelos:
Thanks, I'm not overly keen on using constants as flags, but I'll have a think over this and figure out how best to work it...

lprat:
Thanks for your help.

atcuno:
I am confused what this PR was trying to accomplish. Is this something we are still interested in @ikelos ?

ikelos:
I think the idea was to catch exceptions during the rendering process so that layer results may be returned.  I don't think it would work properly (since the core generator would have bailed by then).  We did add a `fail_on_errors` parameter that could be passed in I believe but never exposed it in the UI because it would likely cause problems (and goes against our forensics perspective of clearly highlighting when an issues has occurred and that future results would be known to be invalid).  These days our generators tend to be more forgiving if they can, but the functionality is still present in case other UIs want to make use of it...

PR #466, opened 2021-03-04
CLI: Add support for any URL being passed to the file parameter

ikelos:
This changes the handling of the file parameter into one that
accepts any URL, not just an existing file.


ikelos:
Similar to #462

PR #472, opened 2021-03-07
Plugins: Initial prototype crashdump writer

ikelos:
This is just a placeholder but has most of the code for creating a valid crashdump file.  The things it's currently missing are:

* Proper version numbers as extracted from the structure preceding the `KD_DEBUGGER_DATA` structure and also anything to do with the KPCR (setting dummy values, decoding it, etc).

@iMHLv2 This isn't done yet, but I figure this gives us visibility of the branch.  I know that writing seems less simple than in volatility 2 (`.write()` rather than `=`) but it turns out there's time want to change the programmatic structure without writing it back to the file (for example, some tricks we play generating structures in pdbconv).  It's also not clear whether assignment should fail if the file is unwritable for some reason (whereas `write` failed is to be expected).  I think that's ok, but if you can think of something simpler that still allows us to differentiate between in-memory assignment and physical rewriting then I'm all for it...  5:)

### DIFF
--- a/volatility3/framework/plugins/windows/writecrashdump.py
+++ b/volatility3/framework/plugins/windows/writecrashdump.py
@@ -1,0 +1,121 @@
+import math
+from typing import List, Type
+
+from volatility3.framework import interfaces, renderers, constants, exceptions
+from volatility3.framework.configuration import requirements
+from volatility3.framework.interfaces import plugins, configuration
+from volatility3.framework.layers import physical, intel
+from volatility3.framework.symbols import intermed
+from volatility3.plugins.windows import info
+
+
+class WriteCrashDump(plugins.PluginInterface):
+    """Runs the automagics and writes the output to a crashdump format file"""
+    default_block_size = 0x500000
+
+    _required_framework_version = (1, 0, 0)
+    _version = (2, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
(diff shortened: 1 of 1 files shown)


ikelos:
I think this is very rudimentary/may not do everything it needs to, but it feels like it's starting to get stale.  @iMHLv2 or @awalters could either of you take a look and make sure that it's doing what it's supposed?  If it is I'll get it merged so it gets updated if/when anything else does...

volatility3/framework/plugins/windows/writecrashdump.py
```diff
@@ -0,0 +1,121 @@
...
+        if is_64_bit:
+            crashdump_json = 'crash64'
+            dump_header_name = '_DUMP_HEADER64'
+            valid_dump_suffix = [ord('6'), ord('4')]
+        else:
+            crashdump_json = 'crash'
+            dump_header_name = '_DUMP_HEADER'
+            valid_dump_suffix = [ord('M'), ord('P')]
+
+        config_path = 'whatever'
```
ikelos, line 48: Bad past @ikelos , very, very bad.  This sets a really bad example, and this needs changing before this gets committed.

atcuno:
Is this something you still plan to complete @ikelos ?

ikelos:
I don't, I thought AAron was taking on the crashdump stuff?  I'm ok leaving PRs to rot, if it helps remind of us what we need to work on.  Again, it's a question of whether people will search to find anything that's been done before or just start from scratch?  Most of our old PRs won't apply these days, but that's not to say the work that was done is useless, just that it'll take a lot to update it.

PR #475, opened 2021-03-11
Temporary workaround for changes in transition PTE

f-block:
When a PTE is in Transition state, Windows 10 since at least 1909 seems to use some of the higher bits of the PageFrameNumber field as flags, leading to a wrong physical offset during the address translation process.

Here the WinDBG output for such a case. The page for the virtual address 246af810000 is in transition:

```
1: kd> !pte 246af810000
                                           VA 00000246af810000
PXE at FFFFE3F1F8FC7020    PPE at FFFFE3F1F8E048D0    PDE at FFFFE3F1C091ABE0    PTE at FFFFE3812357C080
contains 8A00000059B75867  contains 0A0000005EAE3867  contains 0A00000061030867  contains 0000200060533860
pfn 59b75     ---DA--UW-V  pfn 5eae3     ---DA--UWEV  pfn 61030     ---DA--UWEV  not valid
                                                                                  Transition: 60533
                                                                                  Protect: 3 - ExecuteRead


1: kd> dt _MMPTE_TRANSITION FFFFE3812357C080
nt!_MMPTE_TRANSITION
   +0x000 Valid            : 0y0
   +0x000 Write            : 0y0
   +0x000 Spare            : 0y0
   +0x000 IoTracker        : 0y0
   +0x000 SwizzleBit       : 0y0
   +0x000 Protection       : 0y00011 (0x3)
   +0x000 Prototype        : 0y0
   +0x000 Transition       : 0y1
   +0x000 PageFrameNumber  : 0y001000000000000001100000010100110011 (0x200060533)
   +0x000 Unused           : 0y0000000000000000 (0)

!pte reports a PFN of 0x60533 while the field states 0x200060533.

```

This issue also reflects in other areas.
dumpfiles plugin output without fix:

```
ImageSectionObject      0xd806a4c65e70  InputService.dll        Error dumping file
ImageSectionObject      0xd806a4c66320  TextInputFramework.dll  Error dumping file
ImageSectionObject      0xd806a4c62c70  msutb.dll       Error dumping file
ImageSectionObject      0xd806a2e5a460  winsta.dll      Error dumping file
ImageSectionObject      0xd806a2f09d70  CoreMessaging.dll       Error dumping file
ImageSectionObject      0xd806a2faa210  CoreUIComponents.dll    Error dumping file
ImageSectionObject      0xd806a2faa850  WinTypes.dll    Error dumping file
ImageSectionObject      0xd806a2ff96c0  policymanager.dll       Error dumping file
ImageSectionObject      0xd806a5baed50  WordBreakers.dll        Error dumping file
ImageSectionObject      0xd806a2f082e0  propsys.dll     Error dumping file
ImageSectionObject      0xd806a2f0a6d0  msvcp110_win.dll        Error dumping file
ImageSectionObject      0xd806a2f05ef0  uxtheme.dll     Error dumping file
ImageSectionObject      0xd806a270b860  ntmarta.dll     Error dumping file
ImageSectionObject      0xd806a0e94a20  bcryptprimitives.dll    Error dumping file
ImageSectionObject      0xd806a25c66a0  kernel.appcore.dll      Error dumping file
ImageSectionObject      0xd806a0e94700  gdi32full.dll   Error dumping file
ImageSectionObject      0xd806a0e94bb0  KernelBase.dll  Error dumping file
ImageSectionObject      0xd806a0e94570  msvcp_win.dll   Error dumping file
ImageSectionObject      0xd806a0e95830  ucrtbase.dll    Error dumping file
ImageSectionObject      0xd806a0e95e70  win32u.dll      Error dumping file
ImageSectionObject      0xd806a23510c0  kernel32.dll    Error dumping file
ImageSectionObject      0xd806a23526a0  msctf.dll       Error dumping file
ImageSectionObject      0xd806a0e95b50  ole32.dll       Error dumping file
ImageSectionObject      0xd806a0e95510  gdi32.dll       Error dumping file
ImageSectionObject      0xd806a2351250  imm32.dll       Error dumping file
ImageSectionObject      0xd806a2351890  clbcatq.dll     Error dumping file
ImageSectionObject      0xd806a2352ce0  sechost.dll     Error dumping file
ImageSectionObject      0xd806a2351bb0  advapi32.dll    Error dumping file
ImageSectionObject      0xd806a23529c0  SHCore.dll      Error dumping file
ImageSectionObject      0xd806a2351d40  combase.dll     Error dumping file
ImageSectionObject      0xd806a0539890  rpcrt4.dll      Error dumping file
ImageSectionObject      0xd806a06896a0  msvcrt.dll      Error dumping file
ImageSectionObject      0xd806a0539d40  user32.dll      Error dumping file
ImageSectionObject      0xd806a0426210  ntdll.dll       Error dumping file
```

After the transition fix in volatility3/framework/symbols/windows/extensions/__init__.py:

```
ImageSectionObject      0xd806a4c65e70  InputService.dll        file.0xd806a4c65e70.0xd806a4c52010.ImageSectionObject.InputService.dll.img
ImageSectionObject      0xd806a4c66320  TextInputFramework.dll  file.0xd806a4c66320.0xd806a494ac40.ImageSectionObject.TextInputFramework.dll.img
ImageSectionObject      0xd806a4c62c70  msutb.dll       file.0xd806a4c62c70.0xd806a242d2b0.ImageSectionObject.msutb.dll.img
ImageSectionObject      0xd806a2e5a460  winsta.dll      file.0xd806a2e5a460.0xd806a2639d50.ImageSectionObject.winsta.dll.img
ImageSectionObject      0xd806a2f09d70  CoreMessaging.dll       file.0xd806a2f09d70.0xd806a0b9ccb0.ImageSectionObject.CoreMessaging.dll.img
ImageSectionObject      0xd806a2faa210  CoreUIComponents.dll    file.0xd806a2faa210.0xd806a08edcb0.ImageSectionObject.CoreUIComponents.dll.img
ImageSectionObject      0xd806a2faa850  WinTypes.dll    file.0xd806a2faa850.0xd806a090acb0.ImageSectionObject.WinTypes.dll.img
ImageSectionObject      0xd806a2ff96c0  policymanager.dll       file.0xd806a2ff96c0.0xd806a096ecb0.ImageSectionObject.policymanager.dll.img
ImageSectionObject      0xd806a5baed50  WordBreakers.dll        file.0xd806a5baed50.0xd806a4e8fbf0.ImageSectionObject.WordBreakers.dll.img
ImageSectionObject      0xd806a2f082e0  propsys.dll     file.0xd806a2f082e0.0xd806a0b93cb0.ImageSectionObject.propsys.dll.img
ImageSectionObject      0xd806a2f0a6d0  msvcp110_win.dll        file.0xd806a2f0a6d0.0xd806a0bb6cb0.ImageSectionObject.msvcp110_win.dll.img
ImageSectionObject      0xd806a2f05ef0  uxtheme.dll     file.0xd806a2f05ef0.0xd806a0addcb0.ImageSectionObject.uxtheme.dll.img
ImageSectionObject      0xd806a270b860  ntmarta.dll     file.0xd806a270b860.0xd806a0ad1cb0.ImageSectionObject.ntmarta.dll.img
ImageSectionObject      0xd806a0e94a20  bcryptprimitives.dll    file.0xd806a0e94a20.0xd806a0854cb0.ImageSectionObject.bcryptprimitives.dll.img
ImageSectionObject      0xd806a25c66a0  kernel.appcore.dll      file.0xd806a25c66a0.0xd806a08a0cb0.ImageSectionObject.kernel.appcore.dll.img
ImageSectionObject      0xd806a0e94700  gdi32full.dll   file.0xd806a0e94700.0xd806a089ecb0.ImageSectionObject.gdi32full.dll.img
ImageSectionObject      0xd806a0e94bb0  KernelBase.dll  file.0xd806a0e94bb0.0xd806a089acb0.ImageSectionObject.KernelBase.dll.img
ImageSectionObject      0xd806a0e94570  msvcp_win.dll   file.0xd806a0e94570.0xd806a08a9cb0.ImageSectionObject.msvcp_win.dll.img
ImageSectionObject      0xd806a0e95830  ucrtbase.dll    file.0xd806a0e95830.0xd806a0857cb0.ImageSectionObject.ucrtbase.dll.img
ImageSectionObject      0xd806a0e95e70  win32u.dll      file.0xd806a0e95e70.0xd806a0856cb0.ImageSectionObject.win32u.dll.img
ImageSectionObject      0xd806a23510c0  kernel32.dll    file.0xd806a23510c0.0xd806a085bcb0.ImageSectionObject.kernel32.dll.img
ImageSectionObject      0xd806a23526a0  msctf.dll       file.0xd806a23526a0.0xd806a0888cb0.ImageSectionObject.msctf.dll.img
ImageSectionObject      0xd806a0e95b50  ole32.dll       file.0xd806a0e95b50.0xd806a088ecb0.ImageSectionObject.ole32.dll.img
ImageSectionObject      0xd806a0e95510  gdi32.dll       file.0xd806a0e95510.0xd806a088fcb0.ImageSectionObject.gdi32.dll.img
ImageSectionObject      0xd806a2351250  imm32.dll       file.0xd806a2351250.0xd806a0889cb0.ImageSectionObject.imm32.dll.img
ImageSectionObject      0xd806a2351890  clbcatq.dll     file.0xd806a2351890.0xd806a088ccb0.ImageSectionObject.clbcatq.dll.img
ImageSectionObject      0xd806a2352ce0  sechost.dll     file.0xd806a2352ce0.0xd806a087fcb0.ImageSectionObject.sechost.dll.img
ImageSectionObject      0xd806a2351bb0  advapi32.dll    file.0xd806a2351bb0.0xd806a0883cb0.ImageSectionObject.advapi32.dll.img
ImageSectionObject      0xd806a23529c0  SHCore.dll      file.0xd806a23529c0.0xd806a0881cb0.ImageSectionObject.SHCore.dll.img
ImageSectionObject      0xd806a2351d40  combase.dll     file.0xd806a2351d40.0xd806a087ccb0.ImageSectionObject.combase.dll.img
ImageSectionObject      0xd806a0539890  rpcrt4.dll      file.0xd806a0539890.0xd806a0878cb0.ImageSectionObject.rpcrt4.dll.img
ImageSectionObject      0xd806a06896a0  msvcrt.dll      file.0xd806a06896a0.0xd806a085ccb0.ImageSectionObject.msvcrt.dll.img
ImageSectionObject      0xd806a0539d40  user32.dll      file.0xd806a0539d40.0xd806a0879cb0.ImageSectionObject.user32.dll.img
ImageSectionObject      0xd806a0426210  ntdll.dll       file.0xd806a0426210.0xd806a02bbcf0.ImageSectionObject.ntdll.dll.img
```

### DIFF
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -265,7 +265,7 @@ class Intel32e(Intel):
     _direct_metadata = collections.ChainMap({'architecture': 'Intel64'}, Intel._direct_metadata)
     _entry_format = "<Q"
     _bits_per_register = 64
-    _maxphyaddr = 44
+    _maxphyaddr = 52
     _maxvirtaddr = 48
     _structure = [('page map layer 4', 9, False), ('page directory pointer', 9, True), ('page directory', 9, True),
                   ('page table', 9, True)]
@@ -329,5 +329,7 @@ class WindowsIntelPAE(WindowsMixin, IntelPAE):
 
 class WindowsIntel32e(WindowsMixin, Intel32e):
 
+    _maxphyaddr = 45
+    
     def _translate(self, offset: int) -> Tuple[int, int, str]:
         return self._translate_swap(self, offset, self._bits_per_register // 2)
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -329,6 +329,10 @@ class WindowsIntelPAE(WindowsMixin, IntelPAE):
 
 class WindowsIntel32e(WindowsMixin, Intel32e):
 
+    # TODO: Fix appropriately in a future release.
+    # Currently just a temprorary workaround to deal with custom bit flag
+    # in the PFN field for pages in transition state.
+    # See https://github.com/volatilityfoundation/volatility3/pull/475
     _maxphyaddr = 45
     
     def _translate(self, offset: int) -> Tuple[int, int, str]:
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -265,7 +265,7 @@ class Intel32e(Intel):
     _direct_metadata = collections.ChainMap({'architecture': 'Intel64'}, Intel._direct_metadata)
     _entry_format = "<Q"
     _bits_per_register = 64
-    _maxphyaddr = 52
+    _maxphyaddr = 44
     _maxvirtaddr = 48
     _structure = [('page map layer 4', 9, False), ('page directory pointer', 9, True), ('page directory', 9, True),
                   ('page table', 9, True)]
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -972,7 +972,8 @@ class CONTROL_AREA(objects.StructType):
 
                 # If the entry is not a valid physical address then see if it is in transition.
                 elif mmpte.u.Trans.Transition == 1:
-                    physoffset = (mmpte.u.Trans.PageFrameNumber &~ (0b1111 << 32)) << 12
+                    # Strips the bit flag in 'PageFrameNumber' for pages in transition state 
+                    physoffset = (mmpte.u.Trans.PageFrameNumber & (( 1 << 33 ) - 1 ) ) << 12
                     yield physoffset, file_offset, self.PAGE_SIZE
 
                 # Go to the next PTE entry
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -972,7 +972,7 @@ class CONTROL_AREA(objects.StructType):
 
                 # If the entry is not a valid physical address then see if it is in transition.
                 elif mmpte.u.Trans.Transition == 1:
-                    physoffset = mmpte.u.Trans.PageFrameNumber << 12
+                    physoffset = (mmpte.u.Trans.PageFrameNumber &~ (0b1111 << 32)) << 12
                     yield physoffset, file_offset, self.PAGE_SIZE
 
                 # Go to the next PTE entry
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -972,8 +972,12 @@ class CONTROL_AREA(objects.StructType):
 
                 # If the entry is not a valid physical address then see if it is in transition.
                 elif mmpte.u.Trans.Transition == 1:
-                    # Strips the bit flag in 'PageFrameNumber' for pages in transition state 
+                    # TODO: Fix appropriately in a future release.
+                    # Currently just a temprorary workaround to deal with custom bit flag
+                    # in the PFN field for pages in transition state.
+                    # See https://github.com/volatilityfoundation/volatility3/pull/475
                     physoffset = (mmpte.u.Trans.PageFrameNumber & (( 1 << 33 ) - 1 ) ) << 12
+                    
                     yield physoffset, file_offset, self.PAGE_SIZE
 
                 # Go to the next PTE entry


superponible:
Good catch, this was actually something we've seen a while back in some testing I was doing, but I guess a change was never brought over to vol3.

Based on some notes I have from the testing I did at the time across a few samples, I saw the 64th and 45th bit occasionally being set.  The 64th bit was already masked off, but not the 45th bit, as you've shown here.

As mentioned and with the commit here so far, the 44 bit mask would allow up to 16TB.  There are a couple versions of Windows that allow a physical limit up to 24TB (https://docs.microsoft.com/en-us/windows/win32/memory/memory-limits-for-windows-releases).  Not sure what would happen on those systems, but it's probably more of a theoretical problem.  Not sure how many people will be acquiring over 16 TB of memory.

So I'm not sure if the `_maxphyaddr` value should be 44 or 45, but a change to one of those is definitely needed.

volatility3/framework/layers/intel.py
```diff
@@ -265,7 +265,7 @@ class Intel32e(Intel):
     _direct_metadata = collections.ChainMap({'architecture': 'Intel64'}, Intel._direct_metadata)
     _entry_format = "<Q"
     _bits_per_register = 64
-    _maxphyaddr = 52
+    _maxphyaddr = 44
```
ikelos: This ideally wouldn't be applied to the whole Intel32e space, but instead just to the `WindowsIntel32e` class, since it might otherwise unnecessarily (and inaccurately) limit the architecture when it's more a windows limit.  I also wonder whether this should be handled in a way similar to the swap (so we allow the page fault to occur, but then catch and return the right value for it)?  I'm not sure how windows actually does it, and I'm not sure how they'd get around artificially limiting the architecture so high up (if there's a need for flags, why use bit 44 or 45 for the flag?).

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -972,7 +972,7 @@ def get_available_pages(self) -> Iterable[Tuple[int, int, int]]:
 
                 # If the entry is not a valid physical address then see if it is in transition.
                 elif mmpte.u.Trans.Transition == 1:
-                    physoffset = mmpte.u.Trans.PageFrameNumber << 12
+                    physoffset = (mmpte.u.Trans.PageFrameNumber &~ (0b1111 << 32)) << 12
```
ikelos: I'm not a big fan of using `~` for things (since python data types aren't bounded, so it ends up being represented by a negative number).  Is there a way to make this positive (such as `(mmpte.u.Trans.PageFrameNumber ^ (mmpte.u.Trans.PageFrameNumber & 0xf00000000)) << 12` or something)?

ikelos (commented):
Looks good, definitely needs to be applied just to windows systems, and I'd like to avoid `~` if possible, but it's not a hard blocker...  5:)

f-block:
@superponible there is yet no requirement for setting it to 44, as I also only observed bit 45. I just wanted to avoid any problems if there is also a 4th bitflag, and I thought supporting 16TB Ram should be sufficient for most cases (at least at the moment ; ) )

@ikelos how about:
`physoffset = (mmpte.u.Trans.PageFrameNumber & (( 1 << 33 ) - 1 ) )`

or 32, depending on how many bits should be stripped

ikelos:
Yep, I know it may appear less readable but I'd prefer that over what was there.  I might not have thought it through fully, but the new proposed version has the potential to cut off more bits than the previous version (ie, bits above bit 35/36).  Will that affect anything?  If not, then perhaps casting the value to an array of two 32 bit numbers and using the second one might be easier?

f-block:
Not sure if I understand you correctly, but in the context of volatility3/framework/symbols/windows/extensions/__init__.py, the modification only affects the `PageFrameNumber` field, which currently only consists of 36 bits, so `& (( 1 << 33 ) - 1 ) )` should only cut out the 3 highest bits, respectively using `32` only the 4 most significant bits of the PFN value, nothing more.

ikelos:
Ok, I missed that it was a bitfield and only 36-bits.  I'm even *more* happy masking it that way than the `~` way now.  5;)

One question though, if this is a value being read from a structure, rather than values that live within page maps themselves, why aren't we just doing the local masking, and why are we also shrinking the maxphysaddr for the intelspace?  The lookups presumably only come from these structures, so changing them before use would mean the space itself wouldn't need the cap lowering?

f-block:
Because you are not using structures/fields in volatility3/framework/layers/intel.py to access the PFN, but the `_mask` method with bit offsets [https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/layers/intel.py#L85](url)

f-block:
I just didn't want to change your logic there too much

ikelos:
Thanks, I appreciate not tinkering with the core layers too much, but I'm a little confused.  As I see it:
```
PFN value <-- A
  gives value for
Memory address <-- B
  which looks up
Page tables <-- C
  which eventually yields
A page in memory
```
So what I'm wondering is, why are we restricting at B (the addresses that the layer can accept) if the bads one are always coming from A?  Surely adding the limit at A will fix the problem, or are there entries in C that also need to be masked?

Sorry if I'm being a bit slow on the uptake, I don't understand exactly how windows uses the PFN or anything like that, but it seems odd to be changing the architecture's limits based on a specific operating system?  So I guess my question is, are bad entries even actually in the system's page tables, or just in Microsoft's shadow tables?  When we're reading these values from the structures they live in, why can't we just crop them there, rather than in the middle of the page table translation process?

Hopefully that makes sense, and thanks for talking me through it...

f-block:
Except for _MMPTE_HARDWARE, which has the Valid flag set, every other PTE state (including _MMPTE_TRANSITION) is in the eyes of the CPU (or rather the MMU) not valid, and trying to access the page leads to a page fault, which the Windows kernel handles. The kernel knows in which state the PTE/corresponding page is (based on the flags/fields of the PTE) and can handle each case appropriately. In the case of a paged out page, it would read the page from the pagefile and load it in memory. In the case of a page in transition state, the kernel normally would (almost) only have to set the valid flag and the CPU could from now on access the page. This mechanism is the reason why Windows, or every other OS, can use the PTEs (except for _MMPTE_HARDWARE) just as they like, because they typically get a chance to "fix" everything. So in the concrete scenario, access to a page in transition state will lead to a page fault, the kernel takes over and at least sets the valid flag and unsets the bit 45.

Not sure if I get the beginning of your message right, but I'm trying to give an answer: The bits above the PFN, from 48 till the end (counting from 0) are not relevant for translating the virtual to a physical page, at least not in the versions of windows that I so far had a look at, where the fields in `_MMPTE_HARDWARE` above the PFN are defined as follows:

```
  0x0    PageFrameNumber                 [BitField(12-48):PageFrameNumber]: 0x00000000
  0x0    ReservedForHardware             [BitField(48-52):ReservedForHardware]: 0x00000000
  0x0    ReservedForSoftware             [BitField(52-56):ReservedForSoftware]: 0x00000000
  0x0    WsleAge                         [BitField(56-60):WsleAge]: 0x00000000
  0x0    WsleProtection                  [BitField(60-63):WsleProtection]: 0x00000000

```

Moreover, those bits MUST be ignored, because otherwise the current algorithm would read bits up until 52, and if  `ReservedForHardware` contains something other than just zeroes, will lead to a wrong PFN. So the parsing of the PFN should be context aware, which is probably what you were saying?! The reason to use the `_maxphyaddr` value was probably to support the maximum possible physical address supported by the Intel architecture, but if Microsoft is not supporting such high values, it can use the bits otherwise, as long as it doesn't bother the CPU/MMU, and hence potentially leading to translation problems.

The limits mentioned by @superponible suggest that Microsoft is currently not using more than 33 out of the 36 bits for the PFN, which explains why they can use bit 45 of the Transition PTE (bit 33 of the PFN, when counting from 0) for some sort of flag. I hence would recommend to lower the value of `_maxphyaddr` from 52 to 45 (which strips only the 3 most significant bits of the PFN). As long as `_maxphyaddr` is only used for getting the PFN, this currently shouldn't be a problem, but must be taken into account with newer Windows versions, potentially supporting more memory. But those versions will then probably not anymore be able to use the bit 45 as a flag, so it would require some more logic in volatility that differentiates these cases.

ikelos:
Ok, thanks for the explanation!  5:)  So it is actual PTEs in the real page table (rather than a shadow structure that windows maintains and uses).  That was the bit I was trying to get to.  We actually already [handle swap faults](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/layers/intel.py#L289), so it might be worth implementing it the way that windows does.  It feels as though otherwise we're kind of masking over something we don't fully understand and just saving up the issue for when we do need to figure out what's going on (when 16Tb of memory is common).  We could also tidy up the swap handler a little so that it's more of a generic fault handler...  If that would take significant effort then we can just put this fix in, but it would be good to have a `# TODO: ` comment in there, so if we go looking we know that it needs implement properly at some point...

f-block:
Yes, at least I'm not aware of shadow PTEs. The closest thing to that are probably Prototype PTEs, but those are used for another purpose. 

With swap faults you mean the pagefile support? Thanks for that  btw.  : )

So are you thinking about reimplementation, or do you want to/should I just add the two TODOs to the code for now?

ikelos:
Add the two TODOs for now.  It lowers the need to get the code properly implemented, but it also buys us sometime.  I'll try to take a look at how much work re-implementing it as a fault handler would be, and then we can decide whether to push this in or go down the fault handler route...  Thanks!  5:D

f-block:
done

ikelos (approved):
Awesome, as I say I'll see if I can get time to look at moving this to a fault handler, but I'll mark it as approved so we know if doesn't need any further work if I don't get round to the fault handler stuff soon...  Thanks again for all your help!  5:)

f-block:
Sure thing, thank you for all the great work in vol3 : )

PR #476, opened 2021-03-11
adds netstat symbols to Win8x64 ISF files

### LINKED ISSUE #474: Enumeration not found in netscan-win81-x641 table: _PARTITION_TABLE
**Describe the bug**
I am having trouble running windows.netstat on any of the AWS Windows images (Server 2012R2, 2016, and 2019). Each time, I receive an error that _PARTITION_TABLE is not defined in the symbol tables. Netscan runs fine on the same image

**Context**
Volatility Version: Volatility 3 Framework 1.0.1 (just pulled)
Operating System: Ubuntu 20.04.2 LTS (AWS AMI)
Python Version: 3.8.5
Suspected Operating System: AWS Images for Windows Server 2012 R2, Server 2016, Server 2019 (all 64-bit)
Memory Capture: FTK Imager Lite
Command: ./vol.py -vvv -f memdump.mem windows.netstat

**To Reproduce**
1. Create new AWS EC2 instance of any of those Windows versions
2. Capture image (I used FTK Imager Lite from a second volume and saved memory to a third volume)
3. Move capture volume to forensic instance and mount drive
4. Run command 

**Example of netstat on 2012 R2 (6.3.9600)**

> Volatility 3 Framework 1.0.1
> INFO     volatility3.cli: Volatility plugins path: ['/home/ubuntu/Downloads/volatility3/volatility3/plugins', '/home/ubuntu/Downloads/volatility3/volatility3/framework/plugins']
> INFO     volatility3.cli: Volatility symbols path: ['/home/ubuntu/Downloads/volatility3/volatility3/symbols', '/home/ubuntu/Downloads/volatility3/volatility3/framework/symbols']
> DEBUG    volatility3.framework: No module named 'Crypto'
> DEBUG    volatility3.framework: Failed to import module volatility3.plugins.windows.cachedump based on file: windows/cachedump
> DEBUG    volatility3.framework: No module named 'Crypto'
> DEBUG    volatility3.framework: Failed to import module volatility3.plugins.windows.hashdump based on file: windows/hashdump
> DEBUG    volatility3.framework: No module named 'Crypto'
> DEBUG    volatility3.framework: Failed to import module volatility3.plugins.windows.lsadump based on file: windows/lsadump
> INFO     volatility3.cli: The following plugins could not be loaded (use -vv to see why): volatility3.plugins.windows.cachedump, volatility3.plugins.windows.hashdump, volatility3.plugins.windows.lsadump
> INFO     volatility3.framework.automagic: Detected a windows category plugin
> INFO     volatility3.framework.automagic: Running automagic: ConstructionMagic
> Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
> Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
> Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
> Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.NetStat.primary
> Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
> Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.NetStat
> Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
> Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.NetStat.nt_symbols
> Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
> Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.NetStat
> INFO     volatility3.framework.automagic: Running automagic: LayerStacker
> Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
> Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
> DEBUG    volatility3.framework.automagic.windows: DTB was found at: 0x1aa000
> Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
> Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
> Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
> Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
> Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary.memory_layer
> Level 9  volatility3.framework.interfaces.configuration: TypeError - kernel_virtual_offset requirements only accept int type: None
> Level 9  volatility3.framework.interfaces.configuration: TypeError - kernel_virtual_offset requirements only accept int type: None
> Level 9  volatility3.framework.interfaces.configuration: TypeError - kernel_banner requirements only accept str type: None
> Level 9  volatility3.framework.interfaces.configuration: TypeError - kernel_banner requirements only accept str type: None
> Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
> Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.NetStat.nt_symbols
> Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
> Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.NetStat
> DEBUG    volatility3.framework.automagic.stacker: Stacked layers: ['IntelLayer', 'FileLayer']
> INFO     volatility3.framework.automagic: Running automagic: WinSwapLayers
> INFO     volatility3.framework.automagic: Running automagic: WintelHelper
> INFO     volatility3.framework.automagic: Running automagic: KernelPDBScanner
> Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
> Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
> Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
> DEBUG    volatility3.framework.automagic.pdbscan: Kernel base determination - searching layer module list structure
> DEBUG    volatility3.framework.automagic.pdbscan: Kernel base determination - searching layer module list structure
> DEBUG    volatility3.framework.symbols.windows.pdbutil: Using symbol library: ntkrnlmp.pdb/287E489F93AA4C6D94B9CD1469B7F9DE-1
> DEBUG    volatility3.framework.automagic.pdbscan: Setting kernel_virtual_offset to 0xf80165a7f000
> Progress:  100.00		PDB scanning finished                     
> Offset	Proto	LocalAddr	LocalPort	ForeignAddr	ForeignPort	State	PID	Owner	Created
> DEBUG    volatility3.plugins.windows.netscan: Determined OS Version: 6.3 15.9600
> DEBUG    volatility3.plugins.windows.netscan: Determined symbol filename: netscan-win81-x64
> DEBUG    volatility3.framework.symbols: Unresolved reference: nt_symbols1!_ACTIVATION_CONTEXT
> DEBUG    volatility3.framework.symbols: Unresolved reference: nt_symbols1!_LDRP_DLL_SNAP_CONTEXT
> DEBUG    volatility3.plugins.windows.netstat: Found tcpip.sys image base @ 0xf800c28b6000
> DEBUG    volatility3.plugins.windows.netstat: Found tcpip.pdb: EF5FEB3F24CD434F84253EC4DBCDC3CC-2
> DEBUG    volatility3.framework.symbols.windows.pdbutil: Using symbol library: tcpip.pdb/EF5FEB3F24CD434F84253EC4DBCDC3CC-2
> 
> 
> DEBUG    volatility3.cli: Traceback (most recent call last):
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/symbols/__init__.py", line 139, in _weak_resolve
>     return getattr(self._dict[table_name], get_function)(component_name)
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/symbols/intermed.py", line 54, in _delegate_function
>     return getattr(self._delegate, name)(*args, **kwargs)
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/symbols/intermed.py", line 674, in get_type
>     return self.natives.get_type(self.name + constants.BANG + type_name)
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/symbols/native.py", line 74, in get_type
>     _native_type, native_format = self._native_dictionary[type_name]
> KeyError: '_PARTITION_TABLE'
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/contexts/__init__.py", line 104, in object
>     object_template = self._symbol_space.get_type(object_type)
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/symbols/__init__.py", line 181, in get_type
>     self._resolved[type_name] = self._weak_resolve(SymbolType.TYPE, type_name)  # type: ignore
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/symbols/__init__.py", line 141, in _weak_resolve
>     raise exceptions.SymbolError(component_name, table_name,
> volatility3.framework.exceptions.SymbolError: Type netscan-win81-x641!_PARTITION_TABLE references missing Type/Symbol/Enum: '_PARTITION_TABLE'
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/home/ubuntu/Downloads/volatility3/volatility3/cli/__init__.py", line 327, in run
>     renderers[args.renderer]().render(constructed.run())
>   File "/home/ubuntu/Downloads/volatility3/volatility3/cli/text_renderer.py", line 178, in render
>     grid.populate(visitor, outfd)
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/renderers/__init__.py", line 211, in populate
>     for (level, item) in self._generator:
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/plugins/windows/netstat.py", line 443, in _generator
>     for netw_obj in self.list_sockets(self.context, self.config['primary'], self.config['nt_symbols'],
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/plugins/windows/netstat.py", line 393, in list_sockets
>     for endpoint in cls.parse_partitions(context, layer_name, net_symbol_table, tcpip_symbol_table,
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/plugins/windows/netstat.py", line 245, in parse_partitions
>     part_table = context.object(net_symbol_table + constants.BANG + "_PARTITION_TABLE",
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/contexts/__init__.py", line 106, in object
>     object_template = self._symbol_space.get_enumeration(object_type)
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/symbols/__init__.py", line 217, in get_enumeration
>     retval = self._weak_resolve(SymbolType.ENUM, enum_name)
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/symbols/__init__.py", line 139, in _weak_resolve
>     return getattr(self._dict[table_name], get_function)(component_name)
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/symbols/intermed.py", line 54, in _delegate_function
>     return getattr(self._delegate, name)(*args, **kwargs)
>   File "/home/ubuntu/Downloads/volatility3/volatility3/framework/symbols/intermed.py", line 438, in get_enumeration
>     raise exceptions.SymbolError(enum_name, self.name,
> volatility3.framework.exceptions.SymbolError: Enumeration not found in netscan-win81-x641 table: _PARTITION_TABLE
> 
> Volatility experienced a symbol-related issue:
> netscan-win81-x641!_PARTITION_TABLE: Enumeration not found in netscan-win81-x641 table: _PARTITION_TABLE
> 
> 	* An invalid symbol table
> 	* A plugin requesting a bad symbol
> 	* A plugin requesting a symbol from the wrong table
> 
> No further results will be produced

**Additional information**
1. Looking through the symbols files, I don't see _PARTITION_TABLE in many of them, including netscan-win81-x64.json (referenced in the error above)
2. After doing some really hacky things (just copying the _PARTITION and _PARTITION_TABLE from other symbols), netstat started showing connections, but errored again on other missing symbols
3. A fuller list of missing symbols:
    * _PORT_ASSIGNMENT_ENTRY
    * _PORT_ASSIGNMENT_LIST
    * _PORT_ASSIGNMENT
    * _INET_PORT_POOL
    * _PARTITION
    * _PARTITION_TABLE
4. Although adding these symbols resolved the errors, netstat didn't finish (it just sat there) and it doesn't show the PIDs. I don't know if the missing PIDs are because the symbol values are wrong or if that is a separate issue (I have actually never seen netstat PIDs using volatility 3).

Thanks in advance for looking into this and for all of the work you all do.

japhlange:
Related to #474. I could not find the issue with missing PIDs when testing locally, so @michaelmaygar please check if these symbols resolve your issue :)


ikelos (approved):
Some fairly sizable shifts in offsets for CreateTime, but if you're happy they're ok, I'm happy to commit it.  5:)

japhlange:
Good catch, I made a mistake there. Unfortunately when sifting through the TCP listener's memory ranges I don't find anything looking like a proper timestamp, so no idea what's going on there. I'll correct it with the offset of the next (chronological) offset which is from Win10x64...

japhlange:
Would be great still to wait until Michael checks it with his samples.

ikelos:
Yep, no problem, thanks for fixing up the CreateTime offsets!  5:)

japhlange:
Still a bit annoying that there appears to be no CreateTime value in any TCP Listener/Endpoint object but oh well. Maybe my local machine is a bit broken.

michaelmagyar:
You are awesome! Works on my test sample from the most recent AWS Windows 2012 R2 AMI. Unfortunately, still no PIDs =(

image

Here is the -vvv output:

> Volatility 3 Framework 1.0.1
INFO     root        : Volatility plugins path: ['/home/ubuntu/Downloads/volatility3/volatility3/plugins', '/home/ubuntu/Downloads/volatility3/volatility3/framework/plugins']
INFO     root        : Volatility symbols path: ['/home/ubuntu/Downloads/volatility3/volatility3/symbols', '/home/ubuntu/Downloads/volatility3/volatility3/framework/symbols']
DEBUG    volatility3.framework: No module named 'Crypto'
DEBUG    volatility3.framework: Failed to import module volatility3.plugins.windows.cachedump based on file: windows/cachedump
DEBUG    volatility3.framework: No module named 'Crypto'
DEBUG    volatility3.framework: Failed to import module volatility3.plugins.windows.hashdump based on file: windows/hashdump
DEBUG    volatility3.framework: No module named 'Crypto'
DEBUG    volatility3.framework: Failed to import module volatility3.plugins.windows.lsadump based on file: windows/lsadump
INFO     root        : The following plugins could not be loaded (use -vv to see why): volatility3.plugins.windows.cachedump, volatility3.plugins.windows.hashdump, volatility3.plugins.windows.lsadump
INFO     volatility3.framework.automagic: Detected a windows category plugin
Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.NetStat.primary
Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.NetStat
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.NetStat.nt_symbols
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.NetStat
Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
DEBUG    volatility3.framework.automagic.windows: DTB was found at: 0x1aa000
Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary
Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.NetStat.primary.memory_layer
Level 9  volatility3.framework.interfaces.configuration: TypeError - kernel_virtual_offset requirements only accept int type: None
Level 9  volatility3.framework.interfaces.configuration: TypeError - kernel_virtual_offset requirements only accept int type: None
Level 9  volatility3.framework.interfaces.configuration: TypeError - kernel_banner requirements only accept str type: None
Level 9  volatility3.framework.interfaces.configuration: TypeError - kernel_banner requirements only accept str type: None
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.NetStat.nt_symbols
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.NetStat
DEBUG    volatility3.framework.automagic.stacker: Stacked layers: ['IntelLayer', 'FileLayer']
INFO     volatility3.framework.automagic: Running automagic: WinSwapLayers
INFO     volatility3.framework.automagic: Running automagic: WintelHelper
INFO     volatility3.framework.automagic: Running automagic: KernelPDBScanner
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.NetStat.nt_symbols
DEBUG    volatility3.framework.automagic.pdbscan: Kernel base determination - searching layer module list structure
DEBUG    volatility3.framework.automagic.pdbscan: Kernel base determination - searching layer module list structure
DEBUG    volatility3.framework.symbols.windows.pdbutil: Using symbol library: ntkrnlmp.pdb/287E489F93AA4C6D94B9CD1469B7F9DE-1
DEBUG    volatility3.framework.automagic.pdbscan: Setting kernel_virtual_offset to 0xf80165a7f000
Offset	Proto	LocalAddr	LocalPort	ForeignAddr	ForeignPort	State	PID	Owner	Created
DEBUG    volatility3.plugins.windows.netscan: Determined OS Version: 6.3 15.9600
DEBUG    volatility3.plugins.windows.netscan: Determined symbol filename: netscan-win81-x64
DEBUG    volatility3.framework.symbols: Unresolved reference: nt_symbols1!_ACTIVATION_CONTEXT
DEBUG    volatility3.framework.symbols: Unresolved reference: nt_symbols1!_LDRP_DLL_SNAP_CONTEXT
DEBUG    volatility3.plugins.windows.netstat: Found tcpip.sys image base @ 0xf800c28b6000
DEBUG    volatility3.plugins.windows.netstat: Found tcpip.pdb: EF5FEB3F24CD434F84253EC4DBCDC3CC-2
DEBUG    volatility3.framework.symbols.windows.pdbutil: Using symbol library: tcpip.pdb/EF5FEB3F24CD434F84253EC4DBCDC3CC-2
DEBUG    volatility3.plugins.windows.netstat: Found TCP connection PartitionTable @ 0xe0001425b330 (partition count: 1)
DEBUG    volatility3.framework.symbols: Unresolved reference: netscan-win81-x641!__unnamed_2
DEBUG    volatility3.framework.symbols: Unresolved reference: nt_symbols1!_EPROCESS_QUOTA_BLOCK
DEBUG    volatility3.framework.symbols: Unresolved reference: nt_symbols1!_PAGEFAULT_HISTORY
DEBUG    volatility3.framework.symbols: Unresolved reference: nt_symbols1!_JOB_ACCESS_STATE
DEBUG    volatility3.framework.symbols: Unresolved reference: nt_symbols1!_JOB_CPU_RATE_CONTROL
DEBUG    volatility3.framework.symbols: Unresolved reference: nt_symbols1!_JOB_NOTIFICATION_INFORMATION
DEBUG    volatility3.framework.symbols: Unresolved reference: nt_symbols1!_ACTIVATION_CONTEXT_DATA
DEBUG    volatility3.framework.symbols: Unresolved reference: nt_symbols1!_FLS_CALLBACK_INFO
DEBUG    volatility3.framework.symbols: Unresolved reference: nt_symbols1!_ASSEMBLY_STORAGE_MAP
DEBUG    volatility3.framework.symbols: Unresolved reference: nt_symbols1!_SCSI_REQUEST_BLOCK
DEBUG    volatility3.plugins.windows.netstat: Parsing partition 0
DEBUG    volatility3.plugins.windows.netstat: Found _TCP_ENDPOINT @ 0xe00016e59010

0xe00016e59010	TCPv4	172.x.x.x	49184	3.x.x.x	443	ESTABLISHED	-	-	N/ADEBUG    volatility3.plugins.windows.netstat: Found _TCP_ENDPOINT @ 0xe00016e34840

0xe00016e34840	TCPv4	172.x.x.x	49185	99.x.x.x	80	ESTABLISHED	-	-	N/ADEBUG    volatility3.plugins.windows.netstat: Found _TCP_ENDPOINT @ 0xe00017ad14f0

0xe00017ad14f0	TCPv4	172.x.x.x	49186	169.x.x.x	80	CLOSE_WAIT	-	-	N/ADEBUG    volatility3.plugins.windows.netstat: Found _TCP_ENDPOINT @ 0xe00017044d00

0xe00017044d00	TCPv4	172.x.x.x	3389	107.x.x.x	28384	ESTABLISHED	-	-	N/ADEBUG    volatility3.plugins.windows.netstat: Found PortPools @ 0xe000141c9000 (UDP) && 0xe000141cc000 (TCP)
DEBUG    volatility3.plugins.windows.netstat: Found TCP Ports: []
DEBUG    volatility3.plugins.windows.netstat: Found UDP Ports: []

japhlange:
Glad to see that! I assume you cannot share your memory image? 

Do you know for certain for any network connection which process it belongs to (no matter which one, as long as you know the expected PID)? If so, could you do the following:
1. Note down the virtual offset of the connection object you know the process of.
2. Run pslist/psscan and note down the virtual offset of the process you know the network connection of.
3. Open volshell and please run the following: `dq(<connection_object_address>, 608)`. With our current offsets, the pointer to the EPROCESS object (which is the process offset as printed by pslist) should be located at offset 600 in a _TCP_ENDPOINT object. This is wrong, apparently, so maybe we can find the pointer in the dumped memory range.
4. Share the results here :)

If that doesn't work I'm unsure how much we can do remotely, sadly. But we might find a way...

Another less complex idea:
1. Note down the offset of a network connection object lacking process info.
2. run the following in volshell:
```python
(primary) >>> read_qword = lambda x: int.from_bytes(self.context.layers[self.current_layer].read(x, 8), "little")
(primary) >>> for i in range(0, 616, 8):
    if context.object(self.config["nt_symbols"]+"!_EPROCESS", self.current_layer, read_qword(t_o + i)).is_valid():
        print(hex(t_o+i), ", ", i)
```
3. See if produces a result, which should print out the offset if a correct EPROCESS pointer was found...

michaelmagyar:
> 1. Note down the virtual offset of the connection object you know the process of.
> 2. Run pslist/psscan and note down the virtual offset of the process you know the network connection of.
> 3. Open volshell and please run the following: `dq(<connection_object_address>, 608)`. With our current offsets, the pointer to the EPROCESS object (which is the process offset as printed by pslist) should be located at offset 600 in a _TCP_ENDPOINT object. This is wrong, apparently, so maybe we can find the pointer in the dumped memory range.
> 4. Share the results here :)

This worked! Thanks for the instructions.

image

The correct offset for me was **608** in netscan-win81-x64.json on line 321 of your PR

image

This was the raw AWS x64 Windows Server 2012 R2 AMI (deployed on Tuesday).

EDIT:
The other symbol tables had an additional 16 offset for the Created time, so I tried that here and it worked as well, but only for one of the connections:

image

I tried digging around with volshell in the other connections, but couldn't find a similar memory location to the time for the connection that worked (01d714f9e5336d34)

The offset that produced that screenshot was **624** in netscan-win81-x64.json on line 332 of your PR

japhlange:
That _is_ weird. What is also annoying is all those images have the same build ID of  6.3.9600 which means if we want do distinguish between them (read: support different offsets) we need to find another deciding factor so we know when to pick offset 600 and 608 respectively... What works sometimes is checking the members of the NT kernel symbols (which we can always download automatically thanks to MS providing the kernel symbol debug table ID) and check for differences - some symbols are added in newer Windows versions, so we might have a Windows 8.1 version where the offset is 600 (like on my local machine, freshly installed via ISO) and another 8.1 where it's 608 like in your case... 

@ikelos maybe you have another idea how we might distinguish this within one build ID? :S

ikelos:
When we say build ID, are we talking windows kernel, or tcpip.sys version information itself?  If the tcpip.sys version information can help us distinguish, then I'll knock up some code for going from a loaded MZ address to a version number so that we can differentiate better.  If they give the same version numbers then I've no idea how Microsoft keeps anything straight, so I guess we wing it...  5:S

japhlange:
tcpip.sys's version info is often (not sure if always, but often enough to be significant) stripped from the PE in memory - at least when trying to parse the PE headers. They might still be in there somewhere but displaced, so we might be able to find it when scanning tcpip.sys's PE-loaded-in-memory range for something looking like a Version header. 

Because of this issue I opted for using ntoskrnl.exe's version (as reported by vol3's `windows.version` by parsing the KUSER block) as it fit "close enough" for the most part, except for issues like this... The build ID as reported by the PE / kernel in its original format is a bit more verbose than anything I got from memory. The original header ID is something like "10.0.19041.804" (navigate to a Windows PE binary, right-click, properties, details, version value) while all I got from memory was only "10.0.19041". So maybe we find a way to reliably parse the missing minor version too, then we could use that to uniquely identify the netscan ISF tables :)

ikelos:
Well, I've been digging about a bit.  I figured since we had the pdb file, there might be version information in there about the file it was generated from, but there is not.  There's version info for the toolchain and library used to build the pdb, but nothing I think we could use.  As to the version info, it does seem to be around the area, of the library, but I really can't figure out why it would bother being in memory but not where it's supposed to be?  5:S  We probably can write some code to track it down and help us getter accuracy on the library, but it's feeling uglier and uglier to achieve...  5:S

ikelos:
Ok, so as I think @japhlange mentioned, it turns out that the version information structure typically isn't loaded into memory, even when the rest of the DLL is.  Very annoying.  However, sometimes remnants are still in memory, and I've found this with tcpip.sys quite often.  As such, there's a new branch ( #480 ), which will allow searching the physical layer for the tcpip.sys file version information, to try and help distinguish which version it should be using.  I haven't hooked into into netscan or anything yet, but it's a proof of concept and if it works I think I'd be ok adding it in.  If that value isn't enough to distinguish then I'm completely out of ideas...

ikelos:
I'd make any changes for that in a separate branch though.  Just lemme know if/when you're happy for me to merge @japhlange...  5:)

japhlange:
@michaelmagyar Could you check for which tcpip.sys version your other offsets (https://github.com/volatilityfoundation/volatility3/pull/476#issuecomment-797866996) will work?
- start the VM
- navigate to C:\Windows\System32\drivers\
- right-click tcpip.sys, select Properties -> Details
- could show the results like this? (this shows ntoskrnl but it looks the same for tcpip.sys):
ntoskrnl-properties

I need to know the FileVersion value for the exact machine / image version your proposed changes work, that way I can build some kind of distinguisher based on #480 into netstat and we can distinguish _within_ Windows 8.1 :)

michaelmagyar:
> I need to know the FileVersion value for the exact machine / image version your proposed changes work, that way I can build some kind of distinguisher based on #480 into netstat and we can distinguish _within_ Windows 8.1 :)

Looks like it is 6.3.9600.19935

image

japhlange:
Alrighty. I pushed a tcpip.sys version distinguisher based on #480. @michaelmagyar if you could test this it'd be greatly appreciated. You need to use #480's code for verinfo though (which I realized by taking https://github.com/volatilityfoundation/volatility3/blob/4e50402f1123519f71d7d8878c9344341d7557b6/volatility3/framework/plugins/windows/verinfo.py and pasting it over my local verinfo.py instance ;))

@ikelos  please take a look at the version distinguisher, I'm looking forward to your opinion :)


It is able to find all TCP Endpoints as well as both TCP and UDP listeners, just not with all CreateTime entries. I took the 624 offset mentioned above even though it doesn't work for me, maybe it works für Michael...

volatility3/framework/plugins/windows/netscan.py
```diff
@@ -13,7 +13,7 @@
 from volatility3.framework.symbols.windows import versions
 from volatility3.framework.symbols.windows.extensions import network
 from volatility3.plugins import timeliner
-from volatility3.plugins.windows import info, poolscanner
+from volatility3.plugins.windows import info, poolscanner, verinfo
```
ikelos, line 17: This needs a corresponding `VersionRequirement` in the plugin.

volatility3/framework/plugins/windows/netscan.py
```diff
@@ -142,76 +142,94 @@ def determine_tcpip_version(cls, context: interfaces.context.ContextInterface, l
...
+        # Win10_18363 is not recognized by windows.info as 18363
         # because all kernel file headers and debug structures report 18363 as
         # "10.0.18362.1198" with the last part being incremented. However, we can use
         # os_distinguisher to differentiate between 18362 and 18363
         if vers_minor_version == 18362 and is_18363_or_later:
             vollog.debug("Detected 18363 data structures: working with 18363 symbol table.")
             vers_minor_version = 18363
 
+        # Win 8 SP 1 also may have different structures based on specific tcpip.sys version
+        if (nt_major_version, nt_minor_version, vers_minor_version) == (6, 3, 9600):
```
ikelos: Seems a little specific?  Is there way of checking if there tcpip.sys options?  Something like:

```
if [ a,b,c,d for x in version_dict if (a,b,c) == (nt_major_version, nt_minor_version, vers_minor_version) and d != 0]:
```
Ideally I'd prefer longer variable names since but it depends how clean that would look (and for this comment, short is fine).

volatility3/framework/plugins/windows/netscan.py
```diff
@@ -142,76 +142,94 @@ def determine_tcpip_version(cls, context: interfaces.context.ContextInterface, l
...
+            vollog.debug("Requiring further version inspection due to OS version by checking tcpip.sys's FileVersion header")
+            physical_layer_name = context.layers[layer_name].config.get('memory_layer', None)
```
ikelos, line 207: This is intel specific.  We haven't yet got a good way of stepping just below the virtual layer, but adding in a comment mentioning that it will need adapting if other arches come along might be handy.

volatility3/framework/plugins/windows/netscan.py
```diff
@@ -142,76 +142,94 @@ def determine_tcpip_version(cls, context: interfaces.context.ContextInterface, l
...
+            ver = verinfo.VerInfo.find_version_info(context, physical_layer_name, "tcpip.sys")
```
ikelos: Not sure what happens if this is handed a `None` for the layer name?

volatility3/framework/plugins/windows/netscan.py
```diff
@@ -142,76 +142,94 @@ def determine_tcpip_version(cls, context: interfaces.context.ContextInterface, l
...
+            if ver:
+                tcpip_mod_version = ver[3]
```
ikelos: Is the very last part of the version number enough to work given the OS number?

ikelos (commented):
Cool, just a couple comments (and a presumption that JSON values are all correct)...  5:)

volatility3/framework/plugins/windows/netscan.py
ikelos: Yeah, this'll break badly if `physical_layer_name` is None, best to check it before doing the call please...

volatility3/framework/plugins/windows/netscan.py
japhlange: Sure, for which version? As far as I can see verinfo does not define a version tuple yet. (https://github.com/volatilityfoundation/volatility3/blob/4e50402f1123519f71d7d8878c9344341d7557b6/volatility3/framework/plugins/windows/verinfo.py)

volatility3/framework/plugins/windows/netscan.py
japhlange: Yes, thanks for spotting this :)

volatility3/framework/plugins/windows/netscan.py
japhlange: Yeah, I guess so. In my experience the first elements of the `FileVersion` header of `tcpip.sys` are the same values as the Windows kernel version values (which are equal to ntoskrnl.exe's FileVersion). I see thus no reason to use the first three `FileVersion` values of tcpip.sys. If we come across a situation where `tcpip.sys`'s `FileVersion` is different from the kernel's version number we would need to deal with that accordingly. But this case is not covered yet - as it always should be the same, checking it would be redundant.

 I'd rather rely on the kernel version number because it should always be detectable as opposed to this header. Or what do you think?

volatility3/framework/plugins/windows/netscan.py
japhlange: IMO we only need the tcpip_mod_version number for certain Windows versions (such as 6.3.9600 / 8.1). I don't think we need to check the FileVersion header in all other cases where we don't have ambiguity. 

I'd suggest adding more options to this condition if we discover more cases where we need to consider this header. Something like this, maybe.
```python
if (nt_major_version, nt_minor_version, vers_minor_version) in [(6, 3, 9600), (10, 0, 19042), ...]:
    ...
```

volatility3/framework/plugins/windows/netscan.py
japhlange: I've added an info comment.

volatility3/framework/plugins/windows/netscan.py
ikelos: Yeah, I'm happy with that, just wanted to make sure the question had been asked...  5;)

volatility3/framework/plugins/windows/netscan.py
ikelos: Errrr, ok, but it feels like adding a new fix requires changing two places, I'd prefer people only need to add to one place and the rest of the code just does the right thing?   Also that list could one day get long enough to make the code very ugly, and I'd prefer to code for the future than for the right now.  I'm not gonna push it if you really object, but I'm gonna need one more "I really think this is better" before I fold...  5;P

volatility3/framework/plugins/windows/netscan.py
ikelos: Ah, good spot, I bumped the wrong value, so you helped me fix an error too!  5:D  Should be resolved in the branch now...  5:)

ikelos: Yep, so (1, 0, 0) would be the version to depend on please.  5:)

michaelmagyar:
> Alrighty. I pushed a tcpip.sys version distinguisher based on #480. @michaelmagyar if you could test this it'd be greatly appreciated. You need to use #480's code for verinfo though (which I realized by taking https://github.com/volatilityfoundation/volatility3/blob/4e50402f1123519f71d7d8878c9344341d7557b6/volatility3/framework/plugins/windows/verinfo.py and pasting it over my local verinfo.py instance ;))
> 
> @ikelos please take a look at the version distinguisher, I'm looking forward to your opinion :)
> 
> It is able to find all TCP Endpoints as well as both TCP and UDP listeners, just not with all CreateTime entries. I took the 624 offset mentioned above even though it doesn't work for me, maybe it works für Michael...

This is working! The first attempt is with the original pull request 476, while the second includes the updates and the verinfo.py you linked:

image

By the way, I always get that error with this image (current raw AWS AMI for Windows 2012 R2). Volatility finds "4" partitions but only parses "0" before erroring out on "2". I'm not sure if that is expected or worth creating another issue.

~~Note: The time offset did work for one of four connections on a previous memory capture. I don't think I have it anymore (these are just samples for a forensics course module that needs to use AWS for various reasons). I can spin another up and saturate it with connections if you would like for me to validate that.~~ Edit: looks like the created time offset didn't match when I created a bunch of new connections - I'll look into it when I have some time (can always add a new pull request)

ikelos:
If you do want to investigate it further, it would be much better in its own issue please, otherwise requests like this get bogged down with rabbit holes, and get delayed from going into the main branch...  5:S  I'm glad it's (mostly) working for you though.  5:)

volatility3/framework/plugins/windows/netscan.py
japhlange: done!

japhlange:
So folks, I ran into another problem. A test image I've created didn't show TCP and UDP listeners because it was not possible for `verinfo` to read tcpip.sys's FileVersion, which is why not the 8.1-19935 symbols where chosen but rather the normal 8.1 symbols (which were wrong).

Honestly, I have no idea how to mitigate this. I guess we can assume the FileVersion PE header to be readable _often_ but obviously not always. In this case it lead to the distinguisher choosing the wrong ISF table and thus being unable to parse listeners (in `netstat` - `netscan` still works because only `netstat`-specific symbols are different). Neither `windows.info` nor `windows.verinfo` work in this case. But I was able to make out differences in the symbol sets of two different snapshots with different object layouts. This could allow us to use the os_distinguisher here but I'm unsure how best to progress. (So we could fall back to the os_distinguisher when this very Windows build is detected...)

On another note, I've come across a case in this Windows version where two snapshots have _literally_ the same tcpip.pdb GUID (EF5FEB3F24CD434F84253EC4DBCDC3CC) but different object layouts. I hope hope hope that this because of a bit of inconsistency during performing a system update (one snapshot was made during a upgrade, the second one after the finished upgrade and a reboot) to pull it to the newest version but we'll need to keep an eye open for that.

ikelos:
Ugh, I really can't see how Microsoft themselves could produce two identically PDBed DLLs that are different, so for my sanity, I'm gonna assume that didn't happen.  5:P

As to figuring out the version, we could try pulling other information from the MZ (compilation time, or some other field?) but I'd sooner just say best efforts to support what's going on rather than encoding the bytes of every tcpip DLL to differentiate properly.  5:S  We should (and may already) have a `vollog.debug` saying which version numbers were check and what JSON was then loaded, but we should also be able to tell if we're in a situation where the tcpip.sys was needed but couldn't be found, and then we should `vollog.warning` to tell the user "you might not be seeing everything"...

japhlange:
We're already doing this but of course we could increase verbosity:
```
DEBUG    volatility3.plugins.windows.netscan: Determined OS Version: 6.3 15.9600
DEBUG    volatility3.plugins.windows.netscan: Requiring further version inspection due to OS version by checking tcpip.sys's FileVersion header
DEBUG    volatility3.plugins.windows.netscan: Determined tcpip.sys's FileVersion: 18264
DEBUG    volatility3.plugins.windows.netscan: Unable to find exact matching symbol file, going with best fit: netscan-win81-x64
DEBUG    volatility3.plugins.windows.netscan: Determined symbol filename: netscan-win81-x64
```

I think it might be time to add an option to manually override the automatic decision. Reasoning:
1. We have a scenario with ambiguity which ISF to choose (e.g. Win8Sp1x64)
2. We cannot completely rely on checking the FileVersion tcpip version value because it's not always in memory
3. We know that in this case choosing the wrong ISF may lead to most results disappearing from `netstat`
4. Therefore I propose adding a way to make the above steps transparent (e.g. by emphasizing that the ISF decision could not be made with confidence as values are missing/unreadable) and then allowing the analyst to override the algorithmic decision and say something like "Current symbol table doesn't work, let's try the other Win8Sp1x64 table(s)".

Not bringing back the old vol2 profiles completely, but something like it maybe?

volatility3/framework/plugins/windows/netscan.py
japhlange: You're right. I've added it like this - https://github.com/volatilityfoundation/volatility3/pull/476/commits/10c1808e22f68921f7538659f34f3428d84bc389 - I hope the short var names are OK for now. I think at this point it appears necessary to create some more extensive docs about netscan & netstat and their peculiarities anyway...

ikelos: Yeah, in this instance they're ok.  I believe we now cover the majority of situations, which is good, so we're narrowing the bad results down further and further.  5:)

ikelos:
Cool, I like the most recent addition.  I'm a little wary of exposing things like the symbol table filenames to load to the user, but I think limiting control to our own choice would also be problematic.  I'm ok with an override switch as long as it's optional, but I'm not sure exactly how you'd want the user to choose which JSON file to use (and would you allow them to use an arbitrary one, probably not I'm guessing, just due to complexity of the required config option).  Sooooo, yep, in general ok with the idea, but still interested to know the decision mechanism you choose...  5;)

michaelmagyar:
What if there was a plugin option for "try all symbols tables" or something?

Get Outlook for Android<https://aka.ms/AAb9ysg>

________________________________

ikelos:
That seems guaranteed to give known bad results?  I know it would be easy to give a true/false flag but what we're trying to do is give people more control (otherwise if we can detect a bad situation, which it seems we can, why don't we just enable that flag automatically).  I was thinking more along the lines of exposing the JSON file to load, but it just feels like there's dangers that way too.  Definitely not an easy one to cope with...

ikelos:
Sorry, the various JSON files were getting out of hand, so I've moved them into subdirectories, this will just need the changes that were made to the old JSON file moving to where the new one is...

Also, #480 landed, so the changes in that need to not conflict with this (which apparently they do at the moment)...

japhlange:
I'm still thinking about this. No idea how we can proceed.

1. Try to guess (status quo): Apparently does not always work, some kind of override would be useful.
2. Override by allowing the user to specify the symbol table filename: unsure if that could be exploited somehow, but appears reasonable to me?
3. Override by allowing the user to specify the buildstring of the kernel/`tcpip.sys`: how to best make this easy to use? Supplying only the specific build ID (such as 19041 or 19041.1234) requires either "fully-qualified" strings (10.0.19041.12345) or only the build itself (which in turn would be too narrowly focused on Win10 specifically). Also, this requires the user to find out the specific driver version (or trial-and-error their way through all possible options, not unsimilar to vol2's profiles).

@ikelos ideas? 😅

ikelos:
Hiya, I'm not too big a fan of number 3.  Number 1 is acceptable (we can't work miracles, just the best we can, so I'm fine with having corner cases that don't work).  We could do Number 2, but I'd do it as a separate pull request, and let's get the good work that's been done here into the mainline (which means rebasing it after I moved all the JSON and tweaked bits, sorry!).

For implement the second option, we've still essentially got those two specific questions:
* Do we offer a fixed choice of the files available
* If not, do we allow any file on the system, or just those in a particular directory

That's really tricky to consider.  I think offering a fixed set (which will be easier now they're in specific subdirectories) might be the way to go, what I'd need to consider is whether to allow the user to dump their own JSON somewhere, and if so, where.  For now though, that's my preference.  The list of choices would turn up in `--help` if it were made a `ChoiceRequirement` and we can build the list from a quick walk of the directory holding the JSON, so it should all hang together?  As I say, definitely an optional extra and not one I feel should hold up getting everything you've done into the tree...

ikelos:
Just wanted to check in on progress around this.  This pull request will need rebasing if it's to be applied, and I've honestly forgotten where we got to with it.  Is it still relevant?  What needs to happen to this to get the changes in the develop branch?

japhlange:
Still suffering from the limbo that `windows.verinfo` is not really reliable enough to distinguish tcpip.sys's build ID. Have there been any other changes making it easier to detect PE versions in memory?

ikelos:
Not yet, if the pefile doesn't contain the appropriate version information (or it's not present in memory), then we're somewhat stuck...  5:S

I think getting the work that's been done here in is worthwhile in its own right, I guess we just opt for guessing as well as we can if we can't be certain.  So when the version stuff works, use it, when the version stuff doesn't, then guess until we can find a better way of being certain?

ikelos:
@japhlange If we can get this rebased, then I'll add it in as is and that way we can build off what we've got.  I think the version API in this is suitable and usable?

japhlange:
Yeah, sounds good. I'll take a look at it soon and contact you if I need assistance :)

ikelos:
Any progress on this?  I'm just concerned that the longer we leave it the more the code changes and the harder it'll be to integrate properly.  We're also looking to get v2.0.0 published, but at this rate this code won't be included unless we restart the code-freeze clock...

japhlange:
Yeah, although by now I'm a bit rusty on the merge process.. What exactly do you need, a rebase onto develop? :)

ikelos:
Yep, just a rebase onto develop so that t5he merge will apply cleanly please.  Shout if you get into problems, we can figure it out together...  5:)

japhlange:
@ikelos I rebased this PR on top of develop in my repo. Could you take a look if it works out now?

volatility3/framework/plugins/windows/verinfo.py
```diff
@@ -1,204 +0,0 @@
-# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
```
ikelos: Mmmmm, not too keen on this bit...  5;P

ikelos (changes requested):
Generally looks good, but I'm afraid it completely removed one of the plugins it depended upon...  5:S  Once that goes back in I'm happy to merge it...  5:)

volatility3/framework/plugins/windows/verinfo.py
japhlange: whoops, seems I wanted to drop the change made to this file by my branch and accidentally deleted the entire file :D

ikelos (approved):
Looks good, thanks!  5:D

PR #477, opened 2021-03-12
enable support for compiled rules in Yara

dadokkio:


ikelos (commented):
Awesome, thanks for that!  5:)  I wondered if there's a way to distinguish compiled file from normal yara files?  Having an extra option gives us precedence issues (if I do `--yara-file` and `--yara-compiled-file` which one works and which doesn't?), so rather than adding yet another option in, it'd be good if we could just accept it as part of yara file and figure out for ourselves whether we need to run it as compiled or not.  Do you know if that's a possibility?

volatility3/framework/plugins/yarascan.py
```diff
@@ -78,6 +79,8 @@ def process_yara_options(cls, config: Dict[str, Any]):
             rules = yara.compile(sources = {'n': 'rule r1 {{strings: $a = {} condition: $a}}'.format(rule)})
         elif config.get('yara_file', None) is not None:
             rules = yara.compile(file = resources.ResourceAccessor().open(config['yara_file'], "rb"))
+        elif config.get('yara_compiled_file', None) is not None:
+            rules = yara.load(file = resources.ResourceAccessor().open(config['yara_file'], "rb"))
```
ikelos: Also, this should say `config['yara_compiled_file']` presumably?

dadokkio:
I'm not aware.. My idea was to avoid a try except so I've added another value. 
If I try to compile a binary rules it generates error so I don't know exactly how to manage this properly.

ikelos:
Ok, I might look into it a bit before merging this, but thanks for putting something up here so people can use it until we get full support in the tool.  5:)

ikelos:
Hmmm, so it looks like upstream are concerned about people potentially executing malicious code and therefore have separated out the compiled yara files from normal yara files.
```
This is a security measure to prevent users from inadvertenly using compiled rules coming from a third-party. Using compiled rules from untrusted sources can lead to the execution of malicious code in your computer.
```
We should probably do the same, even if it's a bit ugly.  5:S  I might add a comment to it pointing out that the distinction was explicit from the yara makers, but otherwise it looks good to go.  Thanks for the contribution!  5:)

PR #478, opened 2021-03-13
Add PID support for Windows Server 2019 netstat

michaelmagyar:
The offset was incorrect for the AWS x64 Windows Server 2019 (17763) AMI

If PID is working for other samples, I don't know if somehow this is a different version with different offsets (AWS is using datacenter?). That worries me because that may require another symbol version.

Regardless, here is the existing output:

image

And the fix:

image

And the updated output:

image

### DIFF
--- a/volatility3/framework/symbols/windows/netscan-win10-17763-x64.json
+++ b/volatility3/framework/symbols/windows/netscan-win10-17763-x64.json
@@ -192,7 +192,7 @@
     "_TCP_ENDPOINT": {
         "fields": {
             "Owner": {
-                "offset": 656,
+                "offset": 712,
                 "type": {
                     "kind": "pointer",
                     "subtype": {
--- a/volatility3/framework/symbols/windows/netscan-win10-17763-x64.json
+++ b/volatility3/framework/symbols/windows/netscan-win10-17763-x64.json
@@ -202,7 +202,7 @@
                 }
             },
             "CreateTime": {
-                "offset": 672,
+                "offset": 728,
                 "type": {
                     "kind": "union",
                     "name": "_LARGE_INTEGER"


michaelmagyar:
The original offset was off by 56, so I added 56 to the "Created" offset and that seems to work now as well:

image

ikelos:
@japhlange I'll have to defer to your knowledge about whether this change is suitable or not.  I dunno whether they values were ever correct, or if we need new JSON for this specific version in some way?

japhlange:
Yeah we've had the same issue with Win10 18363 where the same build ID has different `tcpip.sys` implementations with different offsets.

1. we need to make sure if the proposed new offsets work for _all_ Windows images with the same version (== build ID as reported by vol3's `windows.version` or checking the PE headers). 
2. If not, we need to pinpoint the change and map it to e.g. a change in a symbol table set so we could use vol3's `os_distinguisher` (for want of a better method, I'm all ears...)

Otherwise we run the risk of only supporting the newer versions then and dropping support for older versions of the same build ID (before the `tcpip.sys` change) :(

michaelmagyar:
I love clean solutions, especially in DF, so obviously I think targeting specific versions is the way to go. That being said, would it be possible to add multiple offsets and check each one? I did that as a quick fix for an autopsy BAM registry plugin now that that key location has changed (try one; if exception, handle and try other). It's not pretty, but it works, and updating future symbols tables for future changes would simply require adding the alternate offset locations.

This would require some plugin logic changes, and I haven't looked into how easy/feasible that would be, but this might be able to be implemented as either a single alternate value or as an array of values to check. This could even be implemented by plugin by checking to see if the type of "offset" is a string or an array and handle accordingly.

japhlange:
In my opinion it's hard for many attributes to try and guess automatically if the value at a given offset is _correct_ (so we know when to move on to the next alternative). At the end it's all bytes we try to parse as a specific data type, so we could easily parse a pointer as an IP address, for instance (at least 4 bytes of it). CreateTime datetime values we could sanity-check by looking at the date, but take the local IP address for instance (https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/symbols/windows/netscan-win10-19041-x64.json#L136). 

It is a pointer to a `_LOCAL_ADDRESS` struct (https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/symbols/windows/netscan-win10-19041-x64.json#L250) which in turn is a pointer to a pointer to a `_IN_ADDR` struct which contains the IP address directly as 4 (or 16) bytes. This means that even if we guess the offset of the first pointer in this pointer chain correctly, we wouldn't know if the IP address value is a correct IP address because any 4 byte value can be (correctly) parsed as an IP address. This value of course can also be empty. So IMO it's hard to know if we have arrived at a correct IP address or if by chance we jumped _somewhere_ completely off...

There are some ways around this dependent on object: if the object has a corresponding implementation in vol3 and has an `is_valid()` method (just as EPROCESS for instance) we could overlay that and see if it appears valid. If the object we are looking for has a pool tag (which some but not all kernel mode object allocations have) we could check for that too. But I think most network connection's attributes lack both of those :(

I'm a bit hesitant though for the plugin to try and guess too much for fear of returning false positives and swaying the results too much. Maybe I'm too paranoid and overthinking this too :D

michaelmagyar:
That makes a lot of sense. Memory forensics is definitely not my area, so I am unfamiliar with how easy/hard it would be to validate a proper value.

japhlange:
Don't wanna discourage these discussions though, there probably are some feasible ways I (or others, maybe) haven't seen yet 😅

ikelos:
Just to add to the conversation, I've added in #480 which *if* the version info is in the image (slack space, etc) it should give you the right four tuple.  It's a bit rough and ready, and I'd kinda prefer it if it used volatility for its parsing rather than struct, but it didn't seem complex enough to go through the effort of making a JSON just for it.  If we start needing to pull back the strings from the following structure, then I'll definitely convert it to a JSON file.  Anyway, this should allow someone to see whether magic could be done...  5:)

ikelos:
Sorry, the various JSON files were getting out of hand, so I've moved them into subdirectories, this will just need the changes that were made to the old JSON file moving to where the new one is...

ikelos:
Any news on this?  Did any of the recent additions help with targetting the specific netstat version?  Do the changes to the JSON file still need making?  I'm happy for you to rebase this and make the change if you'd like to keep attribution, or I can just make the changes manually, but I'd like to get this resolved one way or another...  5:)

japhlange:
I did not get around to find out if it's easier to distinguish tcpip versions now. We can include the different offsets in a new JSON in order to have it documented for now. If at any point later we find a convenient distinguisher then we will have that base covered already.

My proposal would be to move the current 17763 symbols into a new file (`netscan-win10-17763-18264-x64.json`) and put Michael's changes into the "canon" `netscan-win10-17763-x64.json` file. There would then be no code path to reach the older symbols, but at least we would still store the offsets for later...

I'd appreciate it if you could add the changes manually to save me from the rebase. Otherwise I could also create a new PR for the file and offset changes.

ikelos:
The changes will be in git, so we can get back to them that way, rather than having a full copy of the file sat around next to it.  5:)  Manually applied, so closing this request...

PR #480, opened 2021-03-14
Windows: Add additional version info finding method

ikelos:
Adds in the ability to scan a layer (recommended to be a physical layer) for a particular file's version information.  This will extract the file version numbers from it if successful or return None if not.  Since it's an addition it bumps the minor number of the verinfo plugin.

We only apply this to modules, since the hope is that most other dlls are properly loaded, otherwise it could take an extremely long time (and we might want to investigate caching for reloaded DLLs, etc)

### DIFF
--- a/volatility3/framework/plugins/windows/verinfo.py
+++ b/volatility3/framework/plugins/windows/verinfo.py
@@ -27,7 +27,8 @@ except ImportError:
 class VerInfo(interfaces.plugins.PluginInterface):
     """Lists version information from PE files."""
 
-    _required_framework_version = (1, 1, 0)
+    _version = (1, 0, 0)
+    _required_framework_version = (1, 0, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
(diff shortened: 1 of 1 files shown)


japhlange:
Testing this out, but does this really perform a new full physical layer scan for each module? Takes a long time for me... :D

Edit to add: Yes, it did find tcpip.sys's verinfo! That's brilliant, gotta test it more, but it looks really good! 

Would it be possible to improve speed by for instance passing boundaries for the scan to limit the search range? Is this feasible?

ikelos:
The speed is because it's rescanning the entire physical layer for each and every module.  I found that the version information wasn't present in the module's boundaries, and they actually turned up in slack space (not even kernel owned), which is why it checks the entire physical space.  We could adapt it to do it within the boundaries, but I suspect we'd get far fewer results.

One way to speed it up for the verinfo case would be to generate the list of module names and then do a multistring scan, which builds a regex trie which should be much faster and only have to pass over the physical layer once (and then a little bit of tinkering to figure out which hit was found).  However, since there could be bad versions lying around it may well return more than one and we should technically handle that situation.

So in short, yep, there's definite improvements we could make, but none of those should really alter the API which was the bit I wanted to get down first.  Also, keying off OriginalFilename helps, but it may be that certain files don't list the correct filename (or have case issues) which could cause it to fail, so it's very much a semi-tidy proof of concept, but mostly I wanted to check if it would be adaptable for your purposes...

PR #481, opened 2021-03-14
Windows: Generalize symbol_table_from_pdb

ikelos:
This moves the tcpip specific module loading capability across to a separate method stored in pdbutil, which will pull down and load a PDB file after scanning for the debug information from a module.  It also allows the method to be versioned by turning pdbutility into a `VersionableInterface` which should become a requirement of plugins that use it...

### DIFF
--- a/volatility3/framework/symbols/windows/pdbutil.py
+++ b/volatility3/framework/symbols/windows/pdbutil.py
@@ -309,7 +309,7 @@ class PDBUtility(interfaces.configuration.VersionableInterface):
 
         if not guids:
             raise exceptions.VolatilityException(
-                "Did not find GUID of tcpip.pdb in tcpip.sys module @ 0x{:x}!".format(module_offset))
+                "Did not find GUID of {} in module @ 0x{:x}!".format(pdb_name, module_offset))
 
         guid = guids[0]
 
--- a/volatility3/framework/symbols/windows/pdbutil.py
+++ b/volatility3/framework/symbols/windows/pdbutil.py
@@ -12,7 +12,7 @@ from typing import Any, Dict, Generator, List, Optional, Tuple, Union
 from urllib import request, parse
 
 from volatility3 import symbols
-from volatility3.framework import constants, interfaces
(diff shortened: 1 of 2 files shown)


ikelos:
Will commit in 7 days unless there's any complaints/comments (tagging @japhlange since it might be something he's got an opinion on)...  5:)

volatility3/framework/symbols/windows/pdbutil.py
```diff
@@ -279,6 +281,47 @@ def pdbname_scan(cls,
...
+        guids = list(
+            cls.pdbname_scan(context,
+                             layer_name,
+                             context.layers[layer_name].page_size, [bytes(pdb_name, 'latin-1')],
+                             start = module_offset,
+                             end = module_offset + module_size))
+
+        if not guids:
+            raise exceptions.VolatilityException(
+                "Did not find GUID of tcpip.pdb in tcpip.sys module @ 0x{:x}!".format(module_offset))
```
japhlange: Wrong pdb name :)

japhlange (commented):
Looks good I think. Have you tested it against a few images (if netstat still works)?

volatility3/framework/symbols/windows/pdbutil.py
ikelos: Good catch, thanks!  5:)

ikelos:
Yep, they all got their symbols fine, but a win10 from 2016 gave me an error about `_PARTITION_TABLE` not found (`Enumeration not found in netscan-win10-x641`) which I assume is not related to pulling down the symbols...

atcuno:
the create_tcpip_symbol_table function should be removed from netstat.py. Otherwise, this looks very nice :)

ikelos:
Removing it would require a version bump, so I'm going to deprecate it for a release or two, and then remove it...

atcuno:
> Removing it would require a version bump, so I'm going to deprecate it for a release or two, and then remove it...

ohhh nice. I like this approach

PR #483, opened 2021-03-19
Prevent pslist from backtracing on invalid process.  Include a warnin…

atcuno:
…g with the offset.

### DIFF
--- a/volatility3/framework/plugins/windows/pslist.py
+++ b/volatility3/framework/plugins/windows/pslist.py
@@ -202,7 +202,7 @@ class PsList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                    proc.get_is_wow64(), proc.get_create_time(), proc.get_exit_time(), file_output))
 
             except exceptions.InvalidAddressException:
-                vollog.debug("Invalid process found at address: {:x}. Skipping".format(proc.vol.offset))  
+                vollog.info("Invalid process found at address: {:x}. Skipping".format(proc.vol.offset))  
 
     def generate_timeline(self):
         for row in self._generator():
--- a/volatility3/framework/plugins/windows/pslist.py
+++ b/volatility3/framework/plugins/windows/pslist.py
@@ -6,7 +6,7 @@ import datetime
 import logging
 from typing import Callable, Iterable, List, Type
 
-from volatility3.framework import renderers, interfaces, layers, constants
+from volatility3.framework import renderers, interfaces, layers, exceptions, constants
 from volatility3.framework.configuration import requirements
 from volatility3.framework.objects import utility
 from volatility3.framework.renderers import format_hints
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/pslist.py
```diff
@@ -187,17 +187,22 @@ def _generator(self):
...
+                        file_handle.close()
+                        file_output = str(file_handle.preferred_filename)
+
+                yield (0, (proc.UniqueProcessId, proc.InheritedFromUniqueProcessId,
+                   proc.ImageFileName.cast("string", max_length = proc.ImageFileName.vol.count, errors = 'replace'),
+                   format_hints.Hex(offset), proc.ActiveThreads, proc.get_handle_count(), proc.get_session_id(),
+                   proc.get_is_wow64(), proc.get_create_time(), proc.get_exit_time(), file_output))
+
+            except exceptions.InvalidAddressException:
+                vollog.debug("Invalid process found at address: {:x}. Skipping".format(proc.vol.offset))
```
ikelos: I'm wondering if we should up this from debug to warning?  I don't think most people will turn on the debug flags by default which is the reason for throwing this exception in the first place.

ikelos (commented):
Seems reasonable, but I just need to think a little about shifting the error quashing from the consumer to the producer.  This really prevents a consumer from knowing that anything's gone wrong, but letting it continue means the producer won't generate as much information as it could.  We could contemplate a rust style response that collects errors and returns them alongside the results, but I feel that might make everything a bit too ugly to code properly...  5:S

Please keep in mind also things like #465 .  We've had a strong philosophy of throwing errors when errors happen but I can also see the benefit of this, and how the error isn't strictly an error, except this isn't in something like psscan, it's in pslist and it's also exposed as part of the plugin's API, which means that people calling it have no way of knowing what's going on (they can't easily tap the debug messages)from their results.  That makes me more nervous...  Definitely one for discussion before we can make a call on it.  We could also consider conditional exception catching (so a flag that volatility runs with, by default?, which says every exception is fatal)?  I'll add it to my list for discussion next meeting...

volatility3/framework/plugins/windows/pslist.py
atcuno: warning sounds good.. or even info.. some samples smear early so you only get a few processes listed

atcuno: I chose to try/except in the output yield as its the old issue of list_processes returning invalid process objects. In the past we discussed this making it possible for structures that overlap pages to be partially valid and leaving that as an option to developers, but forcing try/except when accessing structure members as the object may be invalid and referencing invalid members throws the exception. As long as list_processes returns invalid objects then the calling code will need to try/except to avoid backtracing on object access.

ikelos:
If you get this fixed up with `.info` or `.warning` I'm happy to commit.  5:)

atcuno:
> If you get this fixed up with `.info` or `.warning` I'm happy to commit. 5:)

done!

PR #484, opened 2021-03-19
Skeleton key check plugin

atcuno:
This plugin attempts to detect skeleton-key like hooks through structured analysis of the CSystems array inside of cryptdll. A complete blogpost on this plugin's operation and R&D background will follow soon after the merge of this PR.

### DIFF
--- a/volatility3/framework/plugins/windows/skeleton_key_check.py
+++ b/volatility3/framework/plugins/windows/skeleton_key_check.py
@@ -13,7 +13,7 @@
 
 import logging, io
 
-from typing import Iterable, Tuple
+from typing import Iterable, Tuple, List
 
 from volatility3.framework.symbols.windows import pdbutil
 from volatility3.framework import interfaces, symbols, exceptions
@@ -26,6 +26,8 @@ from volatility3.framework.renderers import format_hints
 from volatility3.plugins.windows import pslist, vadinfo
 
 from volatility3.framework.symbols.windows.extensions import pe
+
+import pefile
 
 try:
     import capstone
@@ -54,6 +56,17 @@ class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
         ]
 
     def _get_pefile_obj(self, pe_table_name: str, layer_name: str, base_address: int) -> pefile.PE:
+        """
+        Attempts to pefile object from the bytes of the PE file
+
+        Args:
+            pe_table_name: name of the pe types table
+            layer_name: name of the lsass.exe process layer
+            base_address: base address of cryptdll.dll in lsass.exe
+
+        Returns:
+            the constructed pefile object
+        """
         pe_data = io.BytesIO()
 
         try:
@@ -108,6 +121,33 @@ class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
         """ 
         return csystem.Initialize != rc4HmacInitialize or csystem.Decrypt != rc4HmacDecrypt
 
+    def _construct_ecrypt_array(self, array_start: int, count: int, \
+                                cryptdll_types: interfaces.context.ModuleInterface) -> interfaces.context.ModuleInterface:
+        """
+        Attempts to construct an array of _KERB_ECRYPT structures
+
+        Args:
+            array_start: starting virtual address of the array
+            count: how many elements are in the array
+            cryptdll_types: the reverse engineered types
+
+        Returns:
+            The instantiated array
+        """
+
+        try:
+            array = cryptdll_types.object(object_type = "array",
+                                      offset = array_start,
+                                      subtype = cryptdll_types.get_type("_KERB_ECRYPT"),
+                                      count = count,
+                                      absolute = True)
+
+        except exceptions.InvalidAddressException:
+            vollog.debug("Unable to construct cSystems array at given offset: {:x}".format(array_start))
+            array = None
+            
+        return array
+
     def _find_array_with_pdb_symbols(self, cryptdll_symbols: str, 
                                            cryptdll_types: interfaces.context.ModuleInterface, 
                                            proc_layer_name: str, 
@@ -124,12 +164,15 @@ class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
 
         Returns:
             Tuple of:
-            array_start: Where CSystems begins
-            count: Number of array elements
+            array: The cSystems array
             rc4HmacInitialize: The runtime address of the expected initialization function
             rc4HmacDecrypt: The runtime address of the expected decryption function
         """
         cryptdll_module = self.context.module(cryptdll_symbols, layer_name = proc_layer_name, offset = cryptdll_base)
+
+        rc4HmacInitialize = cryptdll_module.get_symbol("rc4HmacInitialize").address + cryptdll_base
+
+        rc4HmacDecrypt = cryptdll_module.get_symbol("rc4HmacDecrypt").address + cryptdll_base
 
         count_address = cryptdll_module.get_symbol("cCSystems").address
 
@@ -142,11 +185,12 @@ class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
 
         array_start = cryptdll_module.get_symbol("CSystems").address + cryptdll_base
 
-        rc4HmacInitialize = cryptdll_module.get_symbol("rc4HmacInitialize").address + cryptdll_base
+        array = self._construct_ecrypt_array(array_start, count, cryptdll_types)
+        
+        if array is None:
+            vollog.debug("The CSystem array is not present in memory. Stopping PDB based analysis.")
 
-        rc4HmacDecrypt = cryptdll_module.get_symbol("rc4HmacDecrypt").address + cryptdll_base
-
-        return array_start, count, rc4HmacInitialize, rc4HmacDecrypt
+        return array, rc4HmacInitialize, rc4HmacDecrypt
 
     def _get_cryptdll_types(self, context: interfaces.context.ContextInterface, 
                                   config, 
@@ -173,25 +217,19 @@ class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
 
         return context.module(cryptdll_symbol_table, proc_layer_name, offset = cryptdll_base)
 
-    def _find_and_parse_cryptdll(self, proc_list: Iterable) -> \
-                                Tuple[interfaces.context.ContextInterface, str, int, int]:  
+    def _find_lsass_proc(self, proc_list: Iterable) -> \
+                               Tuple[interfaces.context.ContextInterface, str]:
         """
-        Finds the base address of cryptdll.dll insode of lsass.exe
+        Walks the process list and returns the first valid lsass instances.
+        There should be only one lsass process, but malware will often use the
+        process name to try and blend in.
 
         Args:
-            proc_list: the process list filtered to just lsass.exe instances
+            proc_list: The process list generator
 
-        Returns:
-            A tuple of:
-            lsass_proc: the process object for lsass.exe
-            proc_layer_name: the name of the lsass.exe process layer
-            cryptdll_base: the base address of cryptdll.dll
-            crytpdll_size: the size of the VAD for cryptdll.dll
+        Return:
+            The process object for lsass
         """
-        lsass_proc    = None
-        proc_layer_name = None
-        cryptdll_base = None
-        cryptdll_size  = None
 
         for proc in proc_list:
             try:
@@ -202,21 +240,31 @@ class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
                                                                                  excp.layer_name))
                 continue
 
-            proc_layer = self.context.layers[proc_layer_name]
+            return proc, proc_layer_name
+            
+        return None, None
 
-            for vad in proc.get_vad_root().traverse():
-                filename = vad.get_file_name()
-                
-                if isinstance(filename, str) and filename.lower().endswith("cryptdll.dll"):
-                    cryptdll_base = vad.get_start()
-                    cryptdll_size = vad.get_end() - cryptdll_base
+    def _find_cryptdll(self, lsass_proc: interfaces.context.ContextInterface) -> \
+                                Tuple[int, int]:  
+        """
+        Finds the base address of cryptdll.dll inside of lsass.exe
 
-                    break
+        Args:
+            lsass_proc: the process object for lsass.exe
 
-            lsass_proc = proc
-            break
+        Returns:
+            A tuple of:
+            cryptdll_base: the base address of cryptdll.dll
+            crytpdll_size: the size of the VAD for cryptdll.dll
+        """
+        for vad in lsass_proc.get_vad_root().traverse():
+            filename = vad.get_file_name()
+            
+            if isinstance(filename, str) and filename.lower().endswith("cryptdll.dll"):
+                base = vad.get_start()
+                return base, vad.get_end() - base
 
-        return lsass_proc, proc_layer_name, cryptdll_base, cryptdll_size
+        return None, None
 
     def _find_csystems_with_symbols(self, proc_layer_name: str, 
                                           cryptdll_types: interfaces.context.ModuleInterface, 
@@ -247,21 +295,14 @@ class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
                                                                     cryptdll_base,
                                                                     cryptdll_size)
         except exceptions.VolatilityException:
+            vollog.debug("Unable to use the cryptdll PDB. Stopping PDB symbols based analysis.")
             return None, None, None
 
-        array_start, count, rc4HmacInitialize, rc4HmacDecrypt = \
+        array, rc4HmacInitialize, rc4HmacDecrypt = \
             self._find_array_with_pdb_symbols(cryptdll_symbols, cryptdll_types, proc_layer_name, cryptdll_base) 
        
-        try: 
-            array = cryptdll_types.object(object_type = "array",
-                                    offset = array_start,
-                                    subtype = cryptdll_types.get_type("_KERB_ECRYPT"),
-                                    count = count,
-                                    absolute = True)
-
-        except exceptions.InvalidAddressException:
+        if array is None:
             vollog.debug("The CSystem array is not present in memory. Stopping PDB symbols based analysis.")
-            return None, None, None
 
         return array, rc4HmacInitialize, rc4HmacDecrypt
 
@@ -292,7 +333,8 @@ class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
         return inst.address + inst.size + opnd.mem.disp
 
     def _analyze_cdlocatecsystem(self, function_bytes: bytes,
-                                       function_start: int, 
+                                       function_start: int,
+                                       cryptdll_types: interfaces.context.ModuleInterface,
                                        proc_layer_name: str) -> Tuple[int, int]:
         """
         Performs static analysis on CDLocateCSystem to find the instructions that
@@ -304,9 +346,7 @@ class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
             proc_layer_name: the name of the lsass.exe process layer
 
         Return:
-            Tuple of:
-            array_start: address of CSystem
-            count: the count from cCsystems or 16
+            The cSystems array of ecrypt instances
         """
         found_count = False
         array_start = None
@@ -344,12 +384,17 @@ class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
                 # we find the count before, so we can terminate the static analysis here
                 break
 
-        return array_start, count
+        if array_start and count:
+            array = self._construct_ecrypt_array(array_start, count, cryptdll_types)
+        else:
+            array = None
+
+        return array
 
     def _find_csystems_with_export(self, proc_layer_name: str, 
                                          cryptdll_types: interfaces.context.ModuleInterface, 
                                          cryptdll_base: int, 
-                                         _) -> Tuple[int, None, None]:
+                                         _) -> interfaces.context.ModuleInterface:
         """
         Uses export table analysis to locate CDLocateCsystem
         This function references CSystems and cCsystems
@@ -360,14 +405,12 @@ class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
             cryptdll_base: Base address of cryptdll.dll inside of lsass.exe
             _: unused in this source
         Returns:
-            Tuple of:
-            array_start: Where CSystems begins
-            None: this method cannot find the expected initialization address
-            None: this method cannot find the expected decryption address
+            The cSystems array
         """
+
         if not has_capstone:
             vollog.debug("capstone is not installed so cannot fall back to export table analysis.")
-            return None, None, None
+            return None
 
         vollog.debug("Unable to perform analysis using PDB symbols, falling back to export table analysis.")
 
@@ -377,15 +420,14 @@ class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
                                                                 "pe",
                                                                 class_types = pe.class_types)
   
-  
+ 
         cryptdll = self._get_pefile_obj(pe_table_name, proc_layer_name, cryptdll_base)
-        if not cryptdll or not hasattr(cryptdll, 'DIRECTORY_ENTRY_EXPORT'):
-            return None, None, None
-        
-        cryptdll.parse_data_directories(directories = [pefile.DIRECTORY_ENTRY["IMAGE_DIRECTORY_ENTRY_EXPORT"]])
+        if not cryptdll:
+            return None
 
-        array_start = None
-        count = None
+        cryptdll.parse_data_directories(directories = [pefile.DIRECTORY_ENTRY["IMAGE_DIRECTORY_ENTRY_EXPORT"]])
+        if not hasattr(cryptdll, 'DIRECTORY_ENTRY_EXPORT'):
+            return None
 
         # find the location of CDLocateCSystem and then perform static analysis
         for export in cryptdll.DIRECTORY_ENTRY_EXPORT.symbols:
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,573 @@
...
+from volatility3.framework.symbols.windows.extensions import pe
+
+try:
+    import capstone
+    has_capstone = True
+except ImportError:
+    has_capstone = False
+
+try:
+    import pefile
```
ikelos: I think this is a hard requirement on volatility (https://github.com/volatilityfoundation/volatility3/blob/develop/setup.py#L40) so you should be safe to take this out of the exception handler, but you're welcome to leave it in if you prefer...  5:)

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,573 @@
...
+        ]
+
+    # @ikelos 
+    # these lines are copy/paste from inside of verinfo->get_version_information
+    # not sure if this is worthy of making it an API or not though
+    # basically it taskes in a pe symbol table, layer name, and base address
+    # and then kicks back a pefile instance
+    # we can either make it a common API or we can just delete this comment
+
+    # @ikelos I don't know how to specify the return value as a pefile object...
```
ikelos: I'd probably specify it as an `io.IOBase` or `io.RawIOBase` depending on which fits the calls you make to it/expect others to make to it.

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+    has_pefile = True
+except ImportError:
+    has_pefile = False
+
+vollog = logging.getLogger(__name__)
+
+class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
+    """Lists process memory ranges that potentially contain injected code."""
+
+    _required_framework_version = (1, 0, 0)
```
ikelos, line 43: I'm surprised there's no `_version` here, but it looks like no classmethods are exposed.  I don't know if any of the methods are worth exposing, but it feels like a lot of good code that could be exposed?

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
+            requirements.VersionRequirement(name = 'pslist', component = pslist.PsList, version = (2, 0, 0)),
+            requirements.VersionRequirement(name = 'vadinfo', component = vadinfo.VadInfo, version = (2, 0, 0)),
+            requirements.VersionRequirement(name = 'pdbutil', component = pdbutil.PDBUtility, version = (1, 0, 0)),
+        ]
+
+    # @ikelos 
+    # these lines are copy/paste from inside of verinfo->get_version_information
+    # not sure if this is worthy of making it an API or not though
```
ikelos: It's only about 7 lines, so I'm ok having it repeated, it just depends how often as to whether we make a full API...

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+            rc4HmacDecrypt: The runtime address of the expected decryption function
+        """
+        cryptdll_module = self.context.module(cryptdll_symbols, layer_name = proc_layer_name, offset = cryptdll_base)
+
+        count_address = cryptdll_module.get_symbol("cCSystems").address
+
+        try:
+            count = cryptdll_types.object(object_type = "unsigned long", offset = count_address)
+        except exceptions.InvalidAddressException:
+            count = 16
```
ikelos, line 184: Is this the maximum, or just a chosen likely number?  A quick comment here would help

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+                                                                    interfaces.configuration.path_join(self.config_path, 'cryptdll'),
+                                                                    proc_layer_name,
+                                                                    "cryptdll.pdb",
+                                                                    cryptdll_base,
+                                                                    cryptdll_size)
+        except exceptions.VolatilityException:
+            return None, None, None
+
+        array_start, count, rc4HmacInitialize, rc4HmacDecrypt = \
+            self._find_array_with_pdb_symbols(cryptdll_symbols, cryptdll_types, proc_layer_name, cryptdll_base)
```
ikelos, line 303: So, this is a little weird, in `_find_array_with_pdb_symbols`we create the module with the offset, but then don't use `object_from_symbol` to construct the array, instead doing it out there and having to add the base onto the offsets?  I'd be tempted to roll these two methods together?

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+            csystems, rc4HmacInitialize, rc4HmacDecrypt = \
+                    source(proc_layer_name, 
+                           cryptdll_types,
+                           cryptdll_base,
+                           cryptdll_size)        
+
+            if csystems is not None:
+                break
+
+        if csystems == None:
```
ikelos: Better to use `is None` rather than `== None`

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+
+        array_start = cryptdll_module.get_symbol("CSystems").address + cryptdll_base
+
+        rc4HmacInitialize = cryptdll_module.get_symbol("rc4HmacInitialize").address + cryptdll_base
+
+        rc4HmacDecrypt = cryptdll_module.get_symbol("rc4HmacDecrypt").address + cryptdll_base
+
+        return array_start, count, rc4HmacInitialize, rc4HmacDecrypt
+
+    def _get_cryptdll_types(self, context: interfaces.context.ContextInterface,
```
ikelos, line 195: Doesn't seem to save much time over calling the create table method?  I'm not against it and I guess it handles the table mapping, so if it's for educational purposes that's fine, but doesn't seem necessary.  5:)

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+            proc_list: the process list filtered to just lsass.exe instances
+
+        Returns:
+            A tuple of:
+            lsass_proc: the process object for lsass.exe
+            proc_layer_name: the name of the lsass.exe process layer
+            cryptdll_base: the base address of cryptdll.dll
+            crytpdll_size: the size of the VAD for cryptdll.dll
+        """
+        lsass_proc    = None
```
ikelos: Looks like yapf hasn't been run over this, might be worthwhile before merge...

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+            except exceptions.InvalidAddressException as excp:
+                vollog.debug("Process {}: invalid address {} in layer {}".format(proc_id, excp.invalid_address,
+                                                                                 excp.layer_name))
+                continue
+
+            proc_layer = self.context.layers[proc_layer_name]
+
+            for vad in proc.get_vad_root().traverse():
+                filename = vad.get_file_name()
+                if type(filename) == renderers.NotApplicableValue or not filename.lower().endswith("cryptdll.dll"):
```
ikelos: Might be worth reversing this into an `isinstance(filename, str)` but not sure it really matters?  I'm just not convinced the continue *and* break are both necessary, feels like the flow is the wrong way around having both?

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+                    continue
+       
+                cryptdll_base = vad.get_start()
+                cryptdll_size = vad.get_end() - cryptdll_base
+
+                break
+
+            lsass_proc = proc
+            break
```
ikelos: Again, feels a bit weird waiting until here to set the `lsass_proc` value and again, the continue/break situation seems odd, I'd tend to if then break since we're only searching for that one situation and the continuing happens for free...

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+try:
+    import pefile
+    has_pefile = True
+except ImportError:
+    has_pefile = False
+
+vollog = logging.getLogger(__name__)
+
+class Skeleton_Key_Check(interfaces.plugins.PluginInterface):
+    """Lists process memory ranges that potentially contain injected code."""
```
ikelos: I don't remember if we use plugin docstrings as description lines anywhere, but even so we can fix that code but it would be nice to have more of a description possibly an introduction/quick glossary.  I've no idea what a csystem, or ccsystem is (or how they differ) and this is probably where I'd go looking for a bit of a description.

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+                function_bytes = self.context.layers[proc_layer_name].read(function_start, 0x50)
+            except exceptions.InvalidAddressException:
+                vollog.debug("The CDLocateCSystem function is not present in the lsass address space. Stopping export based analysis.")
+                break
+
+            array_start, count = self._analyze_cdlocatecsystem(function_bytes, function_start, proc_layer_name)
+        
+            break
+
+        if array_start:
```
ikelos: This could be moved into the `_analyze_cdlocatecsystem` method presumably?

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+        Performs static analysis on CDLocateCSystem to find the instructions that
+        reference CSystems as well as cCsystems
+
+        Args:
+            function_bytes: the instruction bytes of CDLocateCSystem
+            function_start: the address of CDLocateCSystem
+            proc_layer_name: the name of the lsass.exe process layer
+
+        Return:
+            Tuple of:
```
ikelos: As mentioned later, it's not clear why the output isn't just the constructed array?  It's good to use simple types for inputs to functions but there's no as good a reason for doing so in the outputs...

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+            array_start: address of CSystem
+            count: the count from cCsystems or 16
+        """
+        found_count = False
+        array_start = None
+        count = None
+
+        ## we only support 64bit disassembly analysis
```
ikelos, line 356: Do we need to do any kind of checking of the DLL on this in some way?

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+        md.detail = True
+
+        for inst in md.disasm(function_bytes, function_start):
+            # we should not reach debug traps
+            if inst.mnemonic == "int3":
+                break
+
+            # cCsystems is referenced by a mov instruction
+            elif inst.mnemonic == "mov":
+                if found_count == False:
```
ikelos: `not found_count` would be more pythonic

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+                    target_address = self._get_rip_relative_target(inst)
+
+                    # we do not want to fail just because the count is not memory
```
ikelos: Not to nitpick, but since you wanted to fix typos, I think this should be `not in memory` rather than `not memory`  5:)

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,578 @@
...
+                    # 16 was the size on samples I tested, so I chose it as the default
```
ikelos, line 371: Good comment, I approve!  5:D

ikelos (commented):
Yep, looks really nice.  A couple bits to tidy but predominantly suggestions.  Happy for you to resolve each comment how you will and then I'll commit it once you're happy...  5:)

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: took it out and the associated check in the plugin

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: it is the actual pefile instance though that comes back from pefile.PE(). I read through here and I don't see how that fits:

https://docs.python.org/3/library/io.html

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: I took a look through and the only function that I see being re-usable (not skeleton key specific) is __get_rip_relative_target_. This is actually usable by plugins for any OS analyzing x86_64 targets. Would you like me to version based on that one? Or should that function go in API somewhere instead?

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: sounds good. deleted the comment.

volatility3/framework/plugins/windows/skeleton_key_check.py
ikelos: Oh, then mark it as returning the classname I guess (so `pefile.PE`)?  I think that should work.

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: I copy/pasted the same comment on the other place I set to 16 so its clear in both instances.

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: I updated the doc here to be correct. The blog post will replace the marker in the beginning comments once the code and blog post are live / MR accepted. It will have a ton of documentation as the data structures were all reverse engineered.

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: oh cool, I did not know you could specify return values besides Python's built-in ones.

volatility3/framework/plugins/windows/skeleton_key_check.py
ikelos: Hmmm, it's a pretty straight forward calculation, but there's enough that could go wrong in there we probably do want it somewhere common.  I don't think people would think to look in the `skeleton_key_check` file when they wanted things related to disassembly, but I'd like to avoid adding libraries of useful functions to core (since it doesn't show external developers how they can do something similar), so it might be worthwhile starting a `disassembly.py` that contains a `VersionableInterface` derived object (doesn't have to be a plugin, can just be a collection of class/static methods) which people can then pull in using the requirements mechanism.  In that case, the skeleton key plugin doesn't need versioning if it doesn't expose anything...

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: yea, I wanted a standalone method to build out the types to show as an example, and it made the calling function spaghetti(ish) code otherwise.

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: perfect. I changed the flow to remove the continue and break also.

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: lsass_proc is set to show that we for sure found an lsass.exe processes. Python does let loop variables bleed through after the loop, but I never liked doing that. 

Also, I want to break after the first lsass.exe as windows keeps the process list ordered by when the processes started (new ones get added to the end of list), so in the case of fake lsass.exe, they would be after the real one in the list, unless you had something like EFI/MBR type malware, which is a whole different issue.

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: We check that the sample is 64bit. The real lsass won't be 32bit on a 64 bit system and the 64bit lsass will load the 64bit cryptdll, so I think we are okay.

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: I made it "is not in memory". THanks!

volatility3/framework/plugins/windows/skeleton_key_check.py
ikelos: It was more the multiple breaks and the flow.  I'd have set `lsass_proc` when we found the right value and rather than `break` from one loop and `break` from the second loop, just do a return as soon as we've found what we're looking for?  It might already be different due to the comment above, but it just felt a little inefficient to be breaking out of loops to get a return statement.  No real problem with it if you want to leave it in there...

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: done

atcuno:
> Yep, looks really nice. A couple bits to tidy but predominantly suggestions. Happy for you to resolve each comment how you will and then I'll commit it once you're happy... 5:)

Thanks! I addressed everything in my local file but the way the array info is passed around and the arrays being constructed. I agree those are messy, so I am going to fix those up and re-test before committing again.

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: done

volatility3/framework/plugins/windows/skeleton_key_check.py
ikelos: I guess my question is, where do we ensure the sample is 64-bit?  The `primary` requirement seems to allow for `intel32` architecture, so that was why I was asking if a check was needed.  I might've missed it, but I couldn't see anything from the generator down that checked it was explicitly a 64-bit image?

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: i like the disassembly.py idea. Which folder would that go in since its usable across all OSes?

ikelos: Hmmm, I'd have said `symbols` and then either directly in there, or we can start a `generic` directory for all of them (similar to volshell's terminology), but that then prevents people from overriding it, so it might be better if it lives directly under plugins?  I might ask for opinions at the next meeting (remind me if I forget)...  5:)

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: the check and message is here: https://github.com/volatilityfoundation/volatility3/blob/68a6fd252f686f386d60565281cf93655c2c98c6/volatility3/framework/plugins/windows/skeleton_key_check.py#L500

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: I am condensing the array building code across functions, but I am not sure how I would use object_from_symbol for the array. We have _cryptdll_symbols_ which is from the PDB and only has global variable addresses (no type info), and then we have _cryptdll_types_ which is the hand-written JSON from my reversing. I am not sure how to combine those two in a way to then be able to replace manually constructing the array of _KERB_ECRYPT_ instances. If there is a way to do this then I would definitely like to learn how.

volatility3/framework/plugins/windows/skeleton_key_check.py
ikelos: Oh, ok, I was just being blind then, cool!  5:D

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -0,0 +1,561 @@
...
+                continue
+
+            proc_layer = self.context.layers[proc_layer_name]
+
+            for vad in proc.get_vad_root().traverse():
+                filename = vad.get_file_name()
+                
+                if isinstance(filename, str) and filename.lower().endswith("cryptdll.dll"):
+                    cryptdll_base = vad.get_start()
+                    cryptdll_size = vad.get_end() - cryptdll_base
```
ikelos: Sooooo, if we know we're done here, why not just return at this point?  Going through the breaks seems a little pointless, the `lsass_proc = proc` could be moved up to here, and then just the `return lsass_proc, proc_layer_name, cryptdll_base, cryptdll_size` used?  If it's for style (ie, not returning mid-call) I'd argue that breaks are worse for code comprehension.  You may need a `return None, None, None` at the end anyway, but I think overall it's a better way of letting the user figure out what's going on (than iterating over a proc_list that looks like you immediately break out of after a single iteration)?

volatility3/framework/plugins/windows/skeleton_key_check.py
ikelos: Ok, I see.  Yeah, if the symbols are from the PDB and they don't include a type, then it's not going to end up saving much time to build it any other way.  Still seems a bit odd to be returning the `array_start` and using that here, rather than constructing the array in the `_find_array_with_pdb_symbols` call.  It feels like the function call was designed just for this one use, I'd have thought it would be more useful to just create the array in the call and return it here?  I can't see an occasion where you'd want just the offset, rather than the whole array...

I figured out what's going on, and I think it might cause confusion ugliness down the road.  5:S  I think you were expecting to be using the `symbol_shift` which you can apply when loading the symbol table, which means that all symbols loaded come out with that offset (that's how linux handles ASLR).  I think we might've built up the ability for confusion, because if you've shifted the symbols and provide the right offset for the module, when you construct an object that way, it'll add the module offset on *again*.  So you've got to use one or the other, but not both...  5:S  I knew we were headed into dark waters when I added that feature...  Needs some thinking about.  For now, any solution you decide to go for (three additions, or building the symbol table with a `symbol_shift`) will be fine, you just need to be careful about building a module and handing in an offset to it...

atcuno:
Ok @ikelos , check the latest please. I made two commits today. The second addresses the array instantiation, loop breaks, etc.

There are still two outstanding items though:

1) Can you remind me where the final destination of disassembly.py will be? I will create it then with a version tag and move over the rip-relative code.

2) It seems like we should wait on https://github.com/volatilityfoundation/volatility3/issues/488 before pushing this into develop. The blog post that will accompany the plugin is meant to serve as a fully-documented and long-term plugin development guide, so I don't want it to go stale right after publication.

ikelos:
Hmmmm, so probably I'd go with `volatility3/symbols/disassembly.py` or `volatility3/symbols/generic/disassembly.py`.  I am somewhat using symbols as a generic place to dump utility files, but I don't think we've got a good name for generic framework code, and disassembly is tricky to categorize (it's kinda architecture specific, but that would put it under layers and I think that'd be more confusing).

Sure, I'm happy to wait for #488 to be resolved before pushing out a teaching piece.  5:)

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -202,21 +240,31 @@ def _find_and_parse_cryptdll(self, proc_list: Iterable) -> \
                                                                                  excp.layer_name))
                 continue
 
-            proc_layer = self.context.layers[proc_layer_name]
+            return proc, proc_layer_name
```
ikelos: I'd put this inside the try and lose the `continue`?  If the exception happens, the return won't, so it's the same outcome but fewer lines of code/flow to follow?  I'm ok if you want to keep it this way, but it just feels more confusing to me...

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -292,7 +333,8 @@ def _get_rip_relative_target(self, inst) -> int:
         return inst.address + inst.size + opnd.mem.disp
 
     def _analyze_cdlocatecsystem(self, function_bytes: bytes,
-                                       function_start: int, 
+                                       function_start: int,
+                                       cryptdll_types: interfaces.context.ModuleInterface,
                                        proc_layer_name: str) -> Tuple[int, int]:
```
ikelos: Looks like the return type needs updating to something like `Optional[interfaces.objects.ObjectInterface]`?

volatility3/framework/plugins/windows/skeleton_key_check.py
```diff
@@ -344,12 +384,17 @@ def _analyze_cdlocatecsystem(self, function_bytes: bytes,
...
+        else:
+            array = None
+
+        return array
 
     def _find_csystems_with_export(self, proc_layer_name: str, 
                                          cryptdll_types: interfaces.context.ModuleInterface, 
                                          cryptdll_base: int, 
-                                         _) -> Tuple[int, None, None]:
+                                         _) -> interfaces.context.ModuleInterface:
```
ikelos: If this can return `None`, it should be `Optional[interfaces.context.ModuleInterface]` although, I'm not sure it does return a `ModuleInterface`?

ikelos (commented):
Looking much better, I think there's still one ugly `continue` but otherwise it's just return types that could do with some tweaking.  I won't commit until we resolve #488 and possibly also get the disassembly interface ironed out if you want to get that sorted first...

volatility3/framework/plugins/windows/skeleton_key_check.py
atcuno: changed.

atcuno:
@ikelos I believe this is done now, all comments should be addressed. I also switched to the new absolute symbol address function from your features2 branch, which makes the code much nicer.

It also means this should not be merged until that features2 branch is, but after that then we are good to go.

ikelos:
This contains a bunch of changes that I think have already been added.  Could you please rebase and force push this, so I can check through just the actual changes that need to go in?

atcuno:
> This contains a bunch of changes that I think have already been added. Could you please rebase and force push this, so I can check through just the actual changes that need to go in?

I did a rebase before pushing in my changes from today so that I could make sure it worked with the latest code. Is that not what I was supposed to do?

ikelos:
No, that is right, I'm just surprised because it's got a bunch of changes about f-strings and similar, that I thought were already in the develop branch?  5:S

ikelos:
Did you force push the rebase back to this branch?  There should be an `atcuno force pushed this branch` line in the github page?

atcuno:
> Did you force push the rebase back to this branch? There should be an `atcuno force pushed this branch` line in the github page?

I did not force push :/ Is this something I can fix? Git is not my strong point.

ikelos:
Yep, no problem.  Just go to your local branch, make sure it's what you'd like to see up on github instead of what's here, and then do `git push -f`.  That'll blat over what's here (because rebasing makes new commits and kinda rewrites history) and say that instead the commits on your machine are the real ones for this branch.  My hope is that will remove all the commits doing things like f-string changes and similar...

atcuno:
> Yep, no problem. Just go to your local branch, make sure it's what you'd like to see up on github instead of what's here, and then do `git push -f`. That'll blat over what's here (because rebasing makes new commits and kinda rewrites history) and say that instead the commits on your machine are the real ones for this branch. My hope is that will remove all the commits doing things like f-string changes and similar...

I think I managed to fix it! All those old commits from develop went away and now I have only the changes in from the plugin itself... Let me know if that is what you were looking for.

ikelos:
Yep, it went from 167 changed files to 2!  5:D  Much easier to review, I'll give it a quick once over, and then if you're happy, I'll push it?

ikelos (approved):
Happy to merge this, the `disassembly.py` stuff we can do later.  `_find_csystems_with_symbols` seems a little simplistic to be its own function, but happy to leave it in there if you think it's clearer?  Ready to merge, just say the word...  5:)

atcuno:
> Happy to merge this, the `disassembly.py` stuff we can do later. `_find_csystems_with_symbols` seems a little simplistic to be its own function, but happy to leave it in there if you think it's clearer? Ready to merge, just say the word... 5:)

Yes I like it separate to keep generator as clean as possible

PR #491, opened 2021-04-13
Fix reading QEMU-images without JSON configuration

cstation:
Sometimes I come across QEMU-images that do not have a JSON-configuration at the end of the file.
Currently, the code throws an exception in such cases, causing the image-file not to be read as a QEMU-image.

This PR fixes this issue, whilst still thowing an exception when invalid JSON is encountered.
The info from JSON does not seem to make a big difference anyhow, the only place where it is used is around line 191.

### DIFF
--- a/volatility3/framework/layers/qemu.py
+++ b/volatility3/framework/layers/qemu.py
@@ -58,12 +58,15 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
         data = b''
         for i in range(base_layer.maximum_address, base_layer.minimum_address, -chunk_size):
             if i != base_layer.maximum_address:
-                data = base_layer.read(i, chunk_size) + data
+                data = base_layer.read(i, chunk_size).rstrip(b'\x00') + data
                 if b'\x00' in data:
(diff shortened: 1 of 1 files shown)


volatility3/framework/layers/qemu.py
```diff
@@ -62,8 +62,10 @@ def _read_configuration(self, base_layer: interfaces.layers.DataLayerInterface,
                 if b'\x00' in data:
                     start = data.rfind(b'\x00')
                     data = data[data.find(b'{', start):]
-                    return json.loads(data)
-        raise exceptions.LayerException(name, "Could not load JSON configuration from the end of the file")
+                    if len(data) > 1:
```
ikelos: I'd kinda prefer the check were whether the find was `>= 0`, since that requires less comprehension to figure out that data[-1:] will still have a size of 1.  Otherwise, looks good.  5:)

ikelos (commented):
I'd prefer the if were on the value of the find (even if it's slightly more lines) just because it's easier to figure out what's going on.  Happy to hear compelling arguments otherwise though...  5:)

PR #492, opened 2021-04-14
Fix reading JSON configration of QEMU-images

cstation:
Continuation of #491. I renamed my branch, causing the PR to be closed. My apologies for that!

I changed the code as suggested. Moreover, I added `rstrip(b'\x00')`, since the JSON sometimes has trailing zero-bytes. When this is the case, `last_null_byte` would otherwise point to the last byte.

### DIFF
--- a/volatility3/framework/layers/qemu.py
+++ b/volatility3/framework/layers/qemu.py
@@ -58,12 +58,15 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
         data = b''
         for i in range(base_layer.maximum_address, base_layer.minimum_address, -chunk_size):
             if i != base_layer.maximum_address:
-                data = base_layer.read(i, chunk_size) + data
+                data = (base_layer.read(i, chunk_size) + data).rstrip(b'\x00')
                 if b'\x00' in data:
-                    start = data.rfind(b'\x00')
-                    data = data[data.find(b'{', start):]
-                    return json.loads(data)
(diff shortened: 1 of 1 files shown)


volatility3/framework/layers/qemu.py
```diff
@@ -58,12 +58,15 @@ def _read_configuration(self, base_layer: interfaces.layers.DataLayerInterface,
         data = b''
         for i in range(base_layer.maximum_address, base_layer.minimum_address, -chunk_size):
             if i != base_layer.maximum_address:
-                data = base_layer.read(i, chunk_size) + data
+                data = base_layer.read(i, chunk_size).rstrip(b'\x00') + data
```
ikelos: Whilst I like taking the trailing null bytes into account, this could have unintended consequences (given the stripping may happen simply on a block that we tack more data to.  Since it's liable to be JSON it's not likely to ever exhibit itself, but the unintended consequences just make me slight nervous.  Better to do the `rstrip` at the very end, so it's definitely taking off the end of the JSON not somewhere in the middle.

ikelos (commented):
Generally looks good but, in true Columbo style, "just one more thing" please...  5;)

volatility3/framework/layers/qemu.py
cstation: You're right! 👍 I moved it to the end of the line.

ikelos (approved):
Yep, looks good.  A little wasteful rerunning the `rstrip`, but I don't see a better way of doing it...  5;)

PR #498, opened 2021-05-12
complete windows crash dump support

iMHLv2:
This MR supersedes https://github.com/volatilityfoundation/volatility3/pull/452. In particular, it completes the support for Windows crash dumps, including both type 0x1 (list) and 0x5 (bitmap).

### DIFF
--- a/volatility3/framework/layers/crash.py
+++ b/volatility3/framework/layers/crash.py
@@ -97,9 +97,9 @@ class WindowsCrashDump32Layer(segmented.SegmentedLayer):
 
             offset = self.headerpages
             header.PhysicalMemoryBlockBuffer.Run.count = header.PhysicalMemoryBlockBuffer.NumberOfRuns
-            for x in header.PhysicalMemoryBlockBuffer.Run:
-                segments.append((x.BasePage * 0x1000, offset * 0x1000, x.PageCount * 0x1000, x.PageCount * 0x1000))
-                offset += x.PageCount
+            for run in header.PhysicalMemoryBlockBuffer.Run:
+                segments.append((run.BasePage * 0x1000, offset * 0x1000, run.PageCount * 0x1000, run.PageCount * 0x1000))
+                offset += run.PageCount
 
         elif self.dump_type == 0x05:
             summary_header = self.get_summary_header()
--- a/volatility3/framework/layers/crash.py
+++ b/volatility3/framework/layers/crash.py
@@ -8,6 +8,7 @@ from typing import Tuple, Optional
 from volatility3.framework import constants, exceptions, interfaces
 from volatility3.framework.layers import segmented
 from volatility3.framework.symbols import intermed
+from volatility3.framework.symbols.windows.extensions import crash
 
 vollog = logging.getLogger(__name__)
 
@@ -50,7 +51,8 @@ class WindowsCrashDump32Layer(segmented.SegmentedLayer):
         # the _SUMMARY_DUMP is shared between 32- and 64-bit
         self._crash_common_table_name = intermed.IntermediateSymbolTable.create(context, self._config_path,
                                                                                 'windows',
-                                                                                'crash_common')
+                                                                                'crash_common',
+                                                                                class_types=crash.class_types)
 
         # Check Header
         hdr_layer = self._context.layers[self._base_layer]
@@ -88,9 +90,6 @@ class WindowsCrashDump32Layer(segmented.SegmentedLayer):
 
         segments = []
 
-        # instead of hard coding 0x2000, use 0x1000 * self.headerpages so this works for
-        # both 32- and 64-bit dumps
-        summary_header = self.get_summary_header()
         if self.dump_type == 0x1:
             header = self.context.object(self._crash_table_name + constants.BANG + self.dump_header_name,
                                          offset=0,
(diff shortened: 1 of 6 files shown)


volatility3/framework/layers/crash.py
```diff
@@ -67,34 +72,96 @@ def __init__(self, context: interfaces.context.ContextInterface, config_path: st
...
-            #                  x.PageCount * 0x1000))
-            offset += x.PageCount
+        if self.dump_type == 0x1:
+            header = self.context.object(self._crash_table_name + constants.BANG + self.dump_header_name,
+                                         offset=0,
+                                         layer_name=self._base_layer)
+
+            offset = self.headerpages
+            header.PhysicalMemoryBlockBuffer.Run.count = header.PhysicalMemoryBlockBuffer.NumberOfRuns
+            for x in header.PhysicalMemoryBlockBuffer.Run:
```
ikelos: Not a deal breaker by any means, but may be nice to have a more descriptive variable name than just `x`?  I know its further use it only two lines aware, but even just `run` would be better?

volatility3/framework/layers/crash.py
```diff
@@ -67,34 +72,96 @@ def __init__(self, context: interfaces.context.ContextInterface, config_path: st
...
+            buffer_long = summary_header.get_buffer_long()
+
+            for outer_index in range(0, ((summary_header.BitmapSize + 31) // 32)):
+                if buffer_long[outer_index] == 0:
+                    if first_bit is not None:
+                        last_bit = ((outer_index - 1) * 32) + 31
+                        segment_length = (last_bit - first_bit + 1) * 0x1000
+                        segments.append((first_bit * 0x1000, first_offset, segment_length, segment_length))
+                        first_bit = None
+                elif buffer_long[outer_index] == 0xFFFFFFFF:
```
ikelos, line 122: Just a note, this looks like a constant based off the data type.  If it's unlikely to change type, that's ok, but if it's just defined as -1 then we probably ought to calculate what -1 would be...

volatility3/framework/layers/crash.py
```diff
@@ -67,34 +72,96 @@ def __init__(self, context: interfaces.context.ContextInterface, config_path: st
...
+                            if first_bit is None:
+                                first_offset = offset
+                                first_bit = bit_addr
+                            offset = offset + 0x1000
+                        else:
+                            if first_bit is not None:
+                                segment_length = ((bit_addr - 1) - first_bit + 1) * 0x1000
+                                segments.append((first_bit * 0x1000, first_offset, segment_length, segment_length))
+                                first_bit = None
+                last_bit_seen = (outer_index * 32) + 31
```
ikelos, line 140: Feels like we might want some byte alignment methods just so everything's in one please?  If they're useful we could pull them up into some shared utility code somewhere, just feels like a lot of +/- 32s and 31s going on whereas it would be nicer to developers to say "bit_aligned_to_word` or something?

volatility3/framework/layers/crash.py
```diff
@@ -67,34 +72,96 @@ def __init__(self, context: interfaces.context.ContextInterface, config_path: st
...
+                                                                                         mapped_offset,
+                                                                                         length))
 
         self._segments = segments
 
     @classmethod
     def check_header(cls, base_layer: interfaces.layers.DataLayerInterface, offset: int = 0) -> Tuple[int, int]:
         # Verify the Window's crash dump file magic
-
+
```
ikelos: Not ideal, but I'm sure it'll get quashed in a future commit.  It suggests yapf hasn't been run over the code though...

volatility3/framework/plugins/windows/crashinfo.py
```diff
@@ -0,0 +1,83 @@
...
+                  str(uptime),
+                  utility.array_to_string(header.Comment),
+                  conversion.wintime_to_datetime(header.SystemTime),
+                  bitmap_header_size,
+                  bitmap_size,
+                  bitmap_pages,
+                  ))
+
+    def run(self):
+        layer = self._context.layers[self.config['primary.memory_layer']]
```
ikelos: So this plugin won't run against a crashdump layer unless it also managed to succesfully stack an intel layer on top of it.  With this particular format, that's exceptionally likely, but it would be better as example code, if it took whatever layer it was given and drilled down until it found a CrashDump layer.  This also avoids issues with calls to crashdump-specific methods (like the new `get_header`).  Calling those on an image that formed an intel layer, but whose base isn't a crashdump will throw some very ugly exceptions, which we should be coping with...

volatility3/framework/plugins/windows/crashinfo.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @classmethod
+    def get_requirements(cls):
+        return [
+            requirements.TranslationLayerRequirement(name = 'primary',
+                                                     description = 'Memory layer for the kernel',
+                                                     architectures = ["Intel32", "Intel64"]), 
+            ]
+        
+    def _generator(self, layer):
```
ikelos: Typing information would help draw out the implicit assumption here.  Layer needs to be a CrashDump layer here, and that needs to have been checked by whatever's handing in layer.

volatility3/framework/symbols/windows/crash64.json
```diff
@@ -65,25 +65,25 @@
           "offset": 40,
           "type": {
             "kind": "base",
-            "name": "unsigned long"
+            "name": "unsigned long long"
           }
         },
         "MachineImageType": {
-          "offset": 44,
+          "offset": 48,
```
ikelos, line 72: Ah, this was likely me taking from the 32-bit version to make a 64-bit version.  5:S  Sorry!

volatility3/framework/symbols/windows/extensions/crash.py
```diff
@@ -0,0 +1,27 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
```
ikelos: Interesting that crash.py's year got updated, but this one got a 2019 year instead...  5;P  I think once they're in there it's no so bad, but they should probably go in with the right year at least...  5;)

volatility3/framework/plugins/windows/crashinfo.py
```diff
@@ -0,0 +1,83 @@
...
+import datetime
+from volatility3.framework import interfaces, renderers
+from volatility3.framework.configuration import requirements
+from volatility3.framework.renderers import format_hints, conversion
+from volatility3.framework.objects import utility
+
+vollog = logging.getLogger(__name__)
+
+class Crashinfo(interfaces.plugins.PluginInterface):
+    _required_framework_version = (1, 0, 0)
```
ikelos: Technically we should be bumping the middle version triplet on core and this plugin should be depending on that version.  Don't know if you want to go that far, but since it needs the changes to the crashdump layer that you made, this strictly won't work with earlier versions.  I'm keen that we get comfortable and familiar doing that (and therefore making our changes additions rather than changes), but I'm happy to do it gently and therefore let this one slide whilst people are getting used to the idea...

ikelos (commented):
Cool, generally very good.  The plugin itself needs to do a little more checking that the layer it's given is actually a crashdump, but otherwise good (and the maths is easier to understand, although I think we could still do a bit for clarity if there's the inclination)...  5:)

volatility3/framework/symbols/windows/extensions/crash.py
iMHLv2: Fixed by 7b73f8d

volatility3/framework/plugins/windows/crashinfo.py
iMHLv2: That was fixed in d2db22c

volatility3/framework/layers/crash.py
iMHLv2: Fixed in 7d9c66c

volatility3/framework/plugins/windows/crashinfo.py
iMHLv2: Fixed in b15e69e

volatility3/framework/layers/crash.py
iMHLv2: Fixed in a0adbde

volatility3/framework/plugins/windows/crashinfo.py
iMHLv2: Let's leave it - this is on a branch so there shouldn't be any plugins depending on it yet.

volatility3/framework/layers/crash.py
iMHLv2: It shouldn't change type (unsigned long).

volatility3/framework/plugins/windows/crashinfo.py
ikelos: Sorry, I'm not sure I was clear.  Nothing depends on `crashinfo` yet, because it's not yet been released, however, `crashinfo` itself depends on the changes we're making to the crash layer.  As such, the `constants.py` MINOR version should be bumped (checking that it's only an additive change, and doesn't affect the framework for anyone that might've written something else) and then the _required framework version_ for this plugin would need to say it needs at least that version to work.  Sorry, there's a lot of version numbers flying around, but essentially, there's the following chain:

```
Core: includes layers and anything else that's unversioned
Plugins: and other versioned components
Anything dependent on a versioned component
```
It all ends up making a big dependency graph so that we have some granularity over stuff, but sadly core currently still contains layers and other bits of the core API (the main interfaces).  We could consider versioning the layers, but then each plugin will have to say which versions of which layers it expects.

Hopefully that makes more sense now?  I've unresolved this, but if you'd prefer to sneak this in and handler potential breakages we can do that.  It will be important to get used to the version numbers and how to handle them down the road though, and since this is an additive change, it should be pretty painless...

ikelos:
Cool, all the other changes look good.  Just holding on the versioning thing before merging...  5:)

volatility3/framework/plugins/windows/crashinfo.py
iMHLv2: This should be resolved in f1079e7 if I understood correctly. I think the `constants.py` reference was meant to be `constants/__init__.py` but let me know if that is not right.

ikelos: Yep, you're quite right, and the change looks good, thanks!  5:)

PR #499, opened 2021-05-13
Linux kernel ring buffer reader plugin

gcmoreira:
### Usage
$ vol.py -f [mem_dump] linux.kmsg.Kmsg 

### Features:
- It works with a wide range, if not all, kernel versions until now (Tested from 3.13 to 5.10)
- It can read the dictionary of properties on each log record  (see example 1).
- In recent kernels, it extracts and shows the caller information (task id or CPU id) from each log record (see example 2).
- It mimics the kernel when calculating the timestamp figure, so it will match up to the last decimal digit with other userland tools.
- It makes it easier to generate timelines crossing the time/task id with other time-based information from other plugins or other forensic sources.

### Example 1:
```
kern    info    1.200118                usb 1-1: new high-speed USB device number 2 using ehci-pci
kern    info    1.200118                 SUBSYSTEM=usb
kern    info    1.200118                 DEVICE=+usb:1-1
kern    info    1.360829                usb 1-1: New USB device found, idVendor=0627, idProduct=0001, bcdDevice= 0.00
kern    info    1.360829                 SUBSYSTEM=usb
kern    info    1.360829                 DEVICE=c189:1
...
```
### Example 2:
```
daemon  info    1.519061        Task(1) systemd[1]: Started Journal Service.
kern    info    1.821892        Task(36)        input: ImExPS/2 Generic Explorer Mouse as /devices/platform/i8042/serio1/input/input3
kern    info    1.912604        Task(310)       cryptd: max_cpu_qlen set to 1000
kern    info    1.918795        Task(302)       AVX2 version of gcm_enc/dec engaged.
kern    info    1.918798        Task(302)       AES CTR mode by8 optimization enabled
kern    info    2.040504        Task(105)       pcieport 0000:00:02.5: pciehp: Slot(0-5): No device found
kern    info    2.040504        Task(105)        SUBSYSTEM=pci
...
```
### Testing
Tested in kernels 3.13, 4.4, 4.10, 4.18 and 5.10.

### Full output example:
[kmsg_output.txt](https://github.com/volatilityfoundation/volatility3/files/6471694/kmsg_output.txt)

### DIFF
--- a/volatility3/framework/plugins/linux/kmsg.py
+++ b/volatility3/framework/plugins/linux/kmsg.py
@@ -216,8 +216,9 @@ class KmsgLegacy(ABCKmsg):
 
         log_first_idx = int(self.vmlinux.object_from_symbol(symbol_name='log_first_idx'))
         cur_idx = log_first_idx
-        end_idx = log_first_idx  # We don't need log_next_idx here. See below msg.len == 0
-        while True:
+        end_idx = None  # We don't need log_next_idx here. See below msg.len == 0
+        while cur_idx != end_idx:
+            end_idx = log_first_idx
             msg_offset = log_buf_ptr + cur_idx  # type: ignore
             msg = self.vmlinux.object(object_type='printk_log', offset=msg_offset)
             if msg.len == 0:
@@ -236,9 +237,6 @@ class KmsgLegacy(ABCKmsg):
                     yield facility_txt, level_txt, timestamp, caller, line
 
                 cur_idx += msg.len
-
-            if cur_idx == end_idx:
-                break
 
 
 class KmsgFiveTen(ABCKmsg):
@@ -343,8 +341,9 @@ class KmsgFiveTen(ABCKmsg):
         desc_id_mask = ~desc_flags_mask
 
         cur_id = desc_ring.tail_id.counter
-        end_id = desc_ring.head_id.counter
-        while True:
+        end_id = None
+        while cur_id != end_id:
+            end_id = desc_ring.head_id.counter
             desc = desc_arr[cur_id % desc_count]  # type: ignore
             info = info_arr[cur_id % desc_count]  # type: ignore
             desc_state = DescStateEnum((desc.state_var.counter >> desc_flags_shift) & 3)
@@ -360,8 +359,6 @@ class KmsgFiveTen(ABCKmsg):
 
             cur_id += 1
             cur_id &= desc_id_mask
-            if cur_id == end_id:
-                break
 
 
 class Kmsg(plugins.PluginInterface):
--- a/volatility3/framework/plugins/linux/kmsg.py
+++ b/volatility3/framework/plugins/linux/kmsg.py
@@ -1,0 +1,386 @@
+# This file is Copyright 2021 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import logging
+from typing import List, Iterator, Tuple, Generator
+
+from abc import ABC, abstractmethod
+from enum import Enum
+
+from volatility3.framework import renderers, interfaces, constants, contexts, class_subclasses
+from volatility3.framework.configuration import requirements
+from volatility3.framework.interfaces import plugins
+from volatility3.framework.objects import utility
+
+
+vollog = logging.getLogger(__name__)
+
+
+class DescStateEnum(Enum):
+    desc_miss = -1          # ID mismatch (pseudo state)
+    desc_reserved = 0x0     # reserved, in use by writer
+    desc_committed = 0x1    # committed by writer, could get reopened
+    desc_finalized = 0x2    # committed, no further modification allowed
+    desc_reusable = 0x3     # free, not yet used by any writer
+
+
+class ABCKmsg(ABC):
+    """Kernel log buffer reader"""
+    LEVELS = (
+        "emerg",     # system is unusable
+        "alert",     # action must be taken immediately
+        "crit",      # critical conditions
+        "err",       # error conditions
+        "warn",      # warning conditions
+        "notice",    # normal but significant condition
+        "info",      # informational
+        "debug",     # debug-level messages
+    )
+
+    FACILITIES = (
+        "kern",      # kernel messages
+        "user",      # random user-level messages
+        "mail",      # mail system
+        "daemon",    # system daemons
+        "auth",      # security/authorization messages
+        "syslog",    # messages generated internally by syslogd
+        "lpr",       # line printer subsystem
+        "news",      # network news subsystem
+        "uucp",      # UUCP subsystem
+        "cron",      # clock daemon
+        "authpriv",  # security/authorization messages (private)
+        "ftp"        # FTP daemon
+    )
+
+    def __init__(
+        self,
+        context: interfaces.context.ContextInterface,
+        config: interfaces.configuration.HierarchicalDict
+    ):
+        self._context = context
+        self._config = config
+        self.layer_name = self._config['primary']  # type: ignore
+        symbol_table_name = self._config['vmlinux']  # type: ignore
+        self.vmlinux = contexts.Module(context, symbol_table_name, self.layer_name, 0)  # type: ignore
+
+    @classmethod
+    def run_all(
+            cls,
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/linux/kmsg.py
```diff
@@ -0,0 +1,386 @@
...
+        if begin & 1:
+            return ""
+
+        # This means a wrap-around to the beginning of the buffer
+        if begin > end:
+            begin = 0
+
+        # Each element in the ringbuffer is "ID + data".
+        # See prb_data_ring struct
+        desc_id_size = 8  # sizeof(long)
```
atcuno: if this is truly a long value then sizeof() will be 4 on 32bit systems.

You can do something like:

module.get_type("long").size

To grab it dynamically.

I see its used in two places, so setting the size to an instance variable in __init__ with the other variables will make it accessible everywhere.

atcuno:
@ikelos I reviewed and left one comment that can be quickly resolved.

I also had a question for you.. there is usage of "f" like this:

```
yield f" DEVICE={dict_text}"
```

and I remember at some point we had to support a minimum Python 3 version that did not support the use of the "f"... is that still the case or are we passed that now?

volatility3/framework/plugins/linux/kmsg.py
gcmoreira: Yeah, that's true and needs to be fixed. I use module.get_type(TYPE).size in other parts of the code, but the reason I ended up hardcoing the value is the following:

```
>>> vmlinux.get_type('printk_log').size
16
>>> vmlinux.get_type('int').size
4
>>> vmlinux.get_type('long').size
  File "/home/user/volexity/volatility3/volatility3/framework/symbols/__init__.py", line 139, in _weak_resolve
    return getattr(self._dict[table_name], get_function)(component_name)
  File "/home/user/volexity/volatility3/volatility3/framework/symbols/intermed.py", line 54, in _delegate_function
    return getattr(self._delegate, name)(*args, **kwargs)
  File "/home/user/volexity/volatility3/volatility3/framework/symbols/intermed.py", line 674, in get_type
    return self.natives.get_type(self.name + constants.BANG + type_name)
  File "/home/user/volexity/volatility3/volatility3/framework/symbols/native.py", line 74, in get_type
    _native_type, native_format = self._native_dictionary[type_name]
KeyError: 'long'

During handling of the above exception, another exception occurred:

  File "<string>", line 1, in <module>
  File "/home/user/volexity/volatility3/volatility3/framework/contexts/__init__.py", line 159, in wrapper
    return getattr(self._context.symbol_space, method)(name)
  File "/home/user/volexity/volatility3/volatility3/framework/symbols/__init__.py", line 181, in get_type
    self._resolved[type_name] = self._weak_resolve(SymbolType.TYPE, type_name)  # type: ignore
  File "/home/user/volexity/volatility3/volatility3/framework/symbols/__init__.py", line 141, in _weak_resolve
    raise exceptions.SymbolError(component_name, table_name,
volatility3.framework.exceptions.SymbolError: Type vmlinux1!long references missing Type/Symbol/Enum: 'long'
```

gcmoreira: oh ok, checking the json ISF I figured out why that's happening. We need to use "long int" instead of "long":
```
>>> vmlinux.get_type('long int').size
8
```

gcmoreira: cool, fixed. It was actually needed the long unsigned int type there.

gcmoreira: thanks @atcuno for the review ;)

volatility3/framework/plugins/linux/kmsg.py
```diff
@@ -0,0 +1,387 @@
...
+                facility_txt = self.get_facility_text(facility)
+
+                for line in self.get_log_lines(text_data_ring, desc, info):
+                    yield facility_txt, level_txt, timestamp, caller, line
+                for line in self.get_dict_lines(info):
+                    yield facility_txt, level_txt, timestamp, caller, line
+
+            cur_id += 1
+            cur_id &= desc_id_mask
+            if cur_id == end_id:
```
ikelos: Since this comes at the end, I'd much prefer it if the `while True:` became `which cur_id != end_id:`.  I don't like the possibility of run away loops so `while True1 just always gives me the creeps...  5:)

volatility3/framework/plugins/linux/kmsg.py
```diff
@@ -0,0 +1,387 @@
...
+            # More than one class could be executed for an specific kernel
+            # version i.e. Netfilter Ingress hooks
+            # We expect just one implementation to be executed for an specific kernel
+            yield from kmsg_inst.run()
+            break
+
+        if kmsg_inst is None:
+            vollog.error("Unsupported Netfilter kernel implementation")
+
+    @abstractmethod
```
ikelos, line 106: I'm a little confused by the need for the abstractmethod stuff?  Seems like a bunch of ABC stuff coming in here for a single method (when there are other shared methods, like `symbtab_checks`, even though they're classmethods)?  It seems they're class methods just so that a few variables can be stashed in the instance rather than passing them as parameters?  I don't particularly have a problem with it, I just wonder which is easier to read?  Happy with this if you prefer it this way...

volatility3/framework/plugins/linux/kmsg.py
```diff
@@ -0,0 +1,387 @@
...
+                facility_txt = self.get_facility_text(facility)
+
+                for line in self.get_log_lines(msg):
+                    yield facility_txt, level_txt, timestamp, caller, line
+                for line in self.get_dict_lines(msg):
+                    yield facility_txt, level_txt, timestamp, caller, line
+
+                cur_idx += msg.len
+
+            if cur_idx == end_idx:
```
ikelos: Again, feels like a C way of breaking a loop rather than a python way.  Could we move the guard statement to become part of the while statement please?

ikelos (commented):
Really nice, thanks for all the docstrings (it'd be great to have them on every method, but I'm not gonna push my luck!).  5;)  Difficult to tell if splitting the code into classes is worth it, but no strong reason against it.  A couple of loop breaks in `while True` loops that I'd prefer as just straight guarded while loops, bug generally pretty awesome.  5:)

ikelos:
> and I remember at some point we had to support a minimum Python 3 version that did not support the use of the "f"... is that still the case or are we passed that now?

You're right, good catch.  We technically still haven't gotten rid of it yet, so those would need turning into normal `"thing".format()` strings, but I'm hoping we'll move to 3.6 as our minimum sometime soon, and then format strings come into play!  5:)

gcmoreira:
Thanks for the comments. 
Yeah, I saw [here](https://github.com/volatilityfoundation/volatility3/commit/0214fa35a11697f03e7af33a6f90644cde0325c5) that 3.6 is already the minimum required version, so Python f'strings are in already.

volatility3/framework/plugins/linux/kmsg.py
gcmoreira: It's that Python needs a do/while ;)  lol ... hope now it looks better for you.

volatility3/framework/plugins/linux/kmsg.py
gcmoreira: re abstract methods: I make sure the subclasses will implement _run_ and _symtab_checks_.  
Note that I forgot to make _symtab_checks_ an _abstractmethod_ but it was just fixed. I also added a docstring to clarify the _symtab_checks_ usage.

volatility3/framework/plugins/linux/kmsg.py
gcmoreira: done, see the other comment

ikelos:
Yep, you prompted me to push for it, so f-strings are now fine to use.  5;)

volatility3/framework/plugins/linux/kmsg.py
ikelos: So, as you discovered, the names of base types aren't necessarily fixed.  This will probably work fine, but if it's an address you're after, you should always be able to look up `pointer` and that should have the right size for the architecture...

ikelos:
Sorry this took me so long to re-review.  I don't see anything outstanding, so let's get it merged...  5:)

PR #504, opened 2021-05-18
ability to reverse filtering see long description

doomedraven:
Hello, this is very useful to optimize some scans, like in case of sandboxing, imagine:
1 round it scans only all processes that was captured by sandbox aka pid_list
2. round it scans all the rest processes ignoring pid_list from round 1

if you have a better idea how to improve/implement this, let me know, we use our custom function, but i think it might be useful for the rest

### DIFF
--- a/volatility3/framework/plugins/windows/pslist.py
+++ b/volatility3/framework/plugins/windows/pslist.py
@@ -83,12 +83,13 @@ class PsList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
         return file_handle
 
     @classmethod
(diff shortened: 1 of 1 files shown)


ikelos (commented):
Looks reasonable, but since it's technically an addition to the API, the third value of the version number for pslist plugin needs to be increased and any plugin that expects to be able to use exclude and then depend on it...

doomedraven:
increased

doomedraven:
do i need to fix any current plugin in vol to fix the version increase?

ikelos:
Nope, as long as the only part of the (x, y, z) version number that was increased was the `z` then all existing plugins should work with it fine...

doomedraven:
ah cool

ikelos (approved):

PR #513, opened 2021-06-04
Fixed CM_KEY_VALUE get_decode method

bonafont:
Commit 0214fa35a11697f03e7af33a6f90644cde0325c5 made the decode_data() method in registry.py not work anymore because of the removed get() method.
Changed to use the new __missing__ method added in the aforementioned commit.

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/registry.py
+++ b/volatility3/framework/symbols/windows/extensions/registry.py
@@ -260,7 +260,7 @@ class CM_KEY_VALUE(objects.StructType):
             # but the length at the start could be negative so just adding 4 to jump past it
             data = layer.read(self.Data + 4, datalen)
 
-        self_type = RegValueTypes.get(self.Type)
+        self_type = RegValueTypes(self.Type)
         if self_type == RegValueTypes.REG_DWORD:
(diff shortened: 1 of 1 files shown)


ikelos:
Thanks very much.  At the moment we need to support python 3.5.3 and I believe the missing function wasn't added until python 3.6.  If you could explain a little more the error you get we might be able to find a solution that works (although there is already a move to shift the minimum python version up to 3.6 which would then include the `__missing__` method.  You can see if it's necessary in branch `feature/better-module-support`.  I'll leave this open for now until it lands in the branch and eventually develop...  5:)

bonafont:
I've just pulled the latest commit in the develop branch today and I've started having issues with a plugin that I developed. The decode_data() method in the CM_KEY_VALUE class that I'm using no longer works because of the removed get() method in the aforementioned commit. So I've supposed that the support for python 3.5 has been dropped.

So I guess in order to still support python 3.5, the get() method should be reinstated.

PS : Love all the stuff done in Volatility 3 !

ikelos:
Blargh!  You're quite right, that already got merged (sorry, in my head it was sat in a branch waiting to be merged).  Thanks for the fix!  5:)

PR #523, opened 2021-06-30
Fix issue #522: private attribute names mangle

### LINKED ISSUE #522: AttributeError trying to get a type member prefixed by double underscore from other class methods
When calling a type member from a class method and the type member's name is prefixed by a double underscore (a private keyword for Python) the framework won't be able to find that attribute.

That behavior can be reproduced with the following code:
```python
#!/usr/bin/python3 
class A:
    def __getattr__(self, name):
        print(name)

class B:
    @classmethod
    def blah(cls):
        A().foo
        A().__bar

A().foo
A().__bar

B.blah()
```

The above doesn't return:
```
foo
__bar
foo
__bar
```

But the following:
```
foo
__bar
foo
_B__bar
```

Python is mangling the attribute's name and calling _\_\_getattribute\_\__ and _\_\_getattr\_\__ with it prefixed by the class name. So, the attribute **__attrname** will be  **_classname__attrname** in those methods.

In vol3, the issue is in [AgregateType](https://github.com/volatilityfoundation/volatility3/blob/af090bf29e6bb26a5961e0a6c25b5d1ec6e82498/volatility3/framework/objects/__init__.py#L733) where under these circumstances and trying to access a type member like [this](https://github.com/torvalds/linux/blob/62fb9874f5da54fdb243003b386128037319b219/include/net/sock.h#L354), the framework won't be able to find that attribute.

Below is my attempt to fix this issue. Basically, it strips the _\_classname_ prefix from the attribute's name ONLY when this is called from a class method. It works in my tests but would like to hear your thoughts about it before submitting a PR.

The fix requires using the Python inspect module to get the call stack.

```diff
diff --git a/volatility3/framework/objects/__init__.py b/volatility3/framework/objects/__init__.py
index 2f282965..7f840598 100644
--- a/volatility3/framework/objects/__init__.py
+++ b/volatility3/framework/objects/__init__.py
@@ -6,6 +6,7 @@ import collections
 import functools
 import logging
 import struct
+import inspect
 from typing import Any, ClassVar, Dict, List, Iterable, Optional, Tuple, Type, Union as TUnion, overload
 
 from volatility3.framework import interfaces, constants
@@ -732,6 +733,21 @@ class AggregateType(interfaces.objects.ObjectInterface):
 
     def __getattr__(self, attr: str) -> Any:
         """Method for accessing members of the type."""
+
+        frame_locals = inspect.stack()[1][0].f_locals
+        if 'cls' in frame_locals:
+            # When this is called from a class method using an attribute name with private name syntax i.e. __attrname,
+            # Python replaces the attribute's name '__attrname' for '_CLASSNAME__attrname'.
+            # If a type has a member with this syntax, we won't be able to find the attribute. A concrete example:
+            # https://github.com/torvalds/linux/blob/62fb9874f5da54fdb243003b386128037319b219/include/net/sock.h#L354
+            # The code below strips the '_CLASSNAME' prefix from the attribute's name only when this is called from
+            # a class method.
+            calling_cls = frame_locals['cls'].__name__
+            class_prefix = "_{class_name}".format(class_name=calling_cls)
+            attr_prefix = "{class_prefix}__".format(class_prefix=class_prefix)
+            if attr.startswith(attr_prefix):
+                attr = attr[len(class_prefix):]
+
         if attr in ['_concrete_members', 'vol']:
             raise AttributeError("Object has not been properly initialized")
         if attr in self._concrete_members:
```

gcmoreira:
As per https://docs.python.org/3/tutorial/classes.html#private-variables
Python will mangle private attribute names from `__attrname` to `_classname__attrname` to avoid name clashes of names with names defined by subclasses.
This will happen even if subclasses are not involved i.e.: calling `type_member.__foo` from a plugin classmethod.
Note that `__foo` is not meant to be a Python private attribute, but the actual name of the type member.
Like sock.__sk_common here: https://github.com/torvalds/linux/blob/62fb9874f5da54fdb243003b386128037319b219/include/net/sock.h#L354

We need to strip the '_classname' prefix from the attribute's name before continuing with the member attribute lookup.

### DIFF
--- a/volatility3/framework/objects/__init__.py
+++ b/volatility3/framework/objects/__init__.py
@@ -733,14 +733,13 @@ class AggregateType(interfaces.objects.ObjectInterface):
     def __getattr__(self, attr: str) -> Any:
         """Method for accessing members of the type."""
 
-        if  attr.startswith("_") and not attr.startswith("__") and "__" in attr:
-            attr = attr[attr.find("_", 1):]  # See issue #522
-
         if attr in ['_concrete_members', 'vol']:
             raise AttributeError("Object has not been properly initialized")
         if attr in self._concrete_members:
             return self._concrete_members[attr]
-        elif attr in self.vol.members:
+        if  attr.startswith("_") and not attr.startswith("__") and "__" in attr:
+            attr = attr[attr.find("_", 1):]  # See issue #522
+        if attr in self.vol.members:
             mask = self._context.layers[self.vol.layer_name].address_mask
             relative_offset, template = self.vol.members[attr]
             if isinstance(template, templates.ReferenceTemplate):
--- a/volatility3/framework/objects/__init__.py
+++ b/volatility3/framework/objects/__init__.py
@@ -732,6 +732,10 @@ class AggregateType(interfaces.objects.ObjectInterface):
 
     def __getattr__(self, attr: str) -> Any:
         """Method for accessing members of the type."""
+
+        if  attr.startswith("_") and not attr.startswith("__") and "__" in attr:
+            attr = attr[attr.find("_", 1):]  # See issue #522
+
         if attr in ['_concrete_members', 'vol']:
             raise AttributeError("Object has not been properly initialized")
         if attr in self._concrete_members:
--- a/volatility3/framework/objects/__init__.py
+++ b/volatility3/framework/objects/__init__.py
@@ -738,7 +738,7 @@ class AggregateType(interfaces.objects.ObjectInterface):
         if attr in self._concrete_members:
             return self._concrete_members[attr]
         if  attr.startswith("_") and not attr.startswith("__") and "__" in attr:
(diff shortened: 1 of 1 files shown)


ikelos:
Thanks for this.  I think we should do the stripping after we've looked for the attr name in 
```
if attr in ['_concrete_members', 'vol']:
            ...
if attr in self._concrete_members:
            ...
```
So the following `elif` should be turned into its own if statement after the name mangling has happened.  This is because we want calls to lookup `.vol` (which are guaranteed to happen after any member statement) to not go through any extra work before they're handled.  This also allows `_concrete_members` to specify any special values they may want (and will make it easier to move the check to a separate method, which we'll probably want to do soon).

gcmoreira:
Done, thanks @ikelos

ikelos (approved):

PR #527, opened 2021-07-12
Fix offsets in LF_UDT_SRC_LINE

gohar94:
Offsets for `LF_UDT_SRC_LINE` are all starting at `0` which seems incorrect - they need to be fixed to be 4 bytes apart.

### DIFF
--- a/volatility3/framework/symbols/windows/pdb.json
+++ b/volatility3/framework/symbols/windows/pdb.json
@@ -1305,14 +1305,14 @@
           }
         },
         "source_file": {
-          "offset": 0,
+          "offset": 4,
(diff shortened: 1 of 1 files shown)


ikelos:
Thanks very much, it's not a structure we saw/used often, but it looks like there was a copy/paste mistake which has been fixed.  I've also increased the size of LF_UDT_MOD_SRC_LINE which was 4 bytes short.  I don't know if that'll have a significant impact, but if you run into any issues please let me know...

ikelos:
Sorry, I missed that this was a pull request, it looked good, sorry for not using it directly....  5:S

PR #531, opened 2021-07-12
[API] fix mutantscan (and possibly others)

### LINKED ISSUE #530: windows.mutantscan.MutantScan bug
**Describe the bug**
`windows.mutantscan.MutantScan` does not work for the image python3 win10-x86-2016-07-08.lime

**Context**
Volatility Version:  Volatility 3 Framework 1.1.1
Operating System:  win10-x86
Python Version:  3.9.1
Suspected Operating System:  
Command:  python3 /Users/fgomulka/vfoundation/volatility3/vol.py -r json --single-location file:////Users/fgomulka/Downloads/win/win10-x86-2016-07-08.lime -q windows.mutantscan.MutantScan

**To Reproduce**

1. Use command found above.
2. See error

**Expected behavior**
There should be output from the mutantscan plugin. 

**Screenshots**
N/A

**Additional information**
N/A

iMHLv2:
Refs Issue #530 

N.B. This PR requires a major version change which is not currently part of the PR. (Also see #537) 

When `get_object()` uses the "top down" approach, it guesses offsets and accepts the first object that passes the `is_valid()` check. The caller (`generate_pool_scan()`) then checks the object's type and discards the finding if it doesn't match what it expects. In some cases, the first object that passes `is_valid()` does not pass the object type check, so the mutant in that pool allocation is not reported. 

There are two ways to solve this problem. The one I implemented is to turn `get_object()` into a generator. It yields any objects that pass `is_valid()` and lets the caller continue to do the type check. It is possible, but very unlikely, that this could result in more than one mutant being reported, if there happens to be multiple squished into the same pool. The other option is to pass in the variables necessary for validating the object type (`type_map` and `cookie`) to `get_object()` so it only returns objects that pass `is_valid()` AND the type check. 

Ultimately, both checks need to happen. The question is...do we want the caller doing one check and the callee doing the other; or do we want the callee to do both. 

Also, in researching this problem, I stumbled upon an incorrect size for 32-bit `_POOL_HEADER` structures. The JSON said 16, but these are really 8 bytes.

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/pool.py
+++ b/volatility3/framework/symbols/windows/extensions/pool.py
@@ -59,7 +59,7 @@ class POOL_HEADER(objects.StructType):
                                               layer_name = self.vol.layer_name,
                                               offset = self.vol.offset + pool_header_size,
                                               native_layer_name = native_layer_name)
-            return mem_object
+            yield mem_object
 
         # otherwise we have an executive object in the pool
         else:
@@ -145,7 +145,7 @@ class POOL_HEADER(objects.StructType):
                                                           native_layer_name = native_layer_name)
 
                         if mem_object.is_valid():
-                            return mem_object
+                            yield mem_object
 
                     except (TypeError, exceptions.InvalidAddressException):
                         pass
@@ -168,8 +168,7 @@ class POOL_HEADER(objects.StructType):
                     if mem_object.is_valid():
                         return mem_object
                 except (TypeError, exceptions.InvalidAddressException):
-                    return None
-        return None
+                    pass
 
     @classmethod
     @functools.lru_cache()
--- a/volatility3/framework/symbols/windows/extensions/pool.py
+++ b/volatility3/framework/symbols/windows/extensions/pool.py
@@ -166,7 +166,7 @@ class POOL_HEADER(objects.StructType):
 
                 try:
                     if mem_object.is_valid():
-                        return mem_object
+                        yield mem_object
                 except (TypeError, exceptions.InvalidAddressException):
                     pass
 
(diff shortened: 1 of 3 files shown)


ikelos:
If more than one mutant could exist in a pool, and it would be accepted/work within windows then we should support that possibility.  I think I'm ok with the caller having to do an additional check as long as a) it's documented and b) we figure out how much it deviates from the previous API and either bump the API number accordingly, or add it as a second method (keeping the old one intact, but adding a deprecation warning and then removing it in a few versions).

iMHLv2:
@ikelos I actually don't think a single pool will contain multiple objects, outside of severe smearing. At least I've never seen it before, and we've never supported it before. 

Also, the caller is already doing the additional check, so if we accept the proposed patch, then that doesn't change. The method being called changed from a regular function to a generator, though, so callers would need to be aware of that so they can iterate over the results. 

The `get_object()` lives in `framework/symbols/windows/extensions/pool.py` so a change there would require bumping the framework's minor version? And then: 

```
class PoolScanner(plugins.PluginInterface):
    """A generic pool scanner plugin."""

    _version = (1, 0, 0) <---- bump to (2, 0, 0) or (1, 1, 0)?
    _required_framework_version = (1, 0, 0)  <---- change to (1, 1, 0)?
```

ikelos:
Unfortunately, the minor version is only for additive changes.  We'd need to do a larger bump for a non-backwards compatible change (such as returning a list rather than a single element).  So either we add an additional method that returns a list, or we do a major version bump.  5:S  Passing in the variables to carry out the check would theoretically be additive, but might be ugly, and I don't want the version numbers to define how we code...

If necessary we could try to find a way to version those objects, but given they may be used by some framework code, I dunno how we'd check that before use...  5:S

ikelos:
Could we please get this rebased in preparation for merging?

PR #536, opened 2021-07-15
Core: Change pointer caching

### LINKED ISSUE #532: windows.netscan.NetScan not working for Win10-x86
**Describe the bug**
There is an image of Windows10 which returns an error

**Context**
Operating System:  Win10-x86
Python Version:  3.9.1
Suspected Operating System:  win10-x86
Command:  python3 vol.py --single-location win10-x86-2016-07-08.lime -q windows.netscan.NetScan

**To Reproduce**

1. Download the windows image win10-x86-2016-07-08.lime from https://buildbot.volatilityfoundation.org/volatility3/images/download/
1. Use command found above.
1. See error

**Expected behavior**
The plugin should not return this error 

**Screenshots**
Screen Shot 2021-07-12 at 12 14 06 PM

**Additional information**
N/A

ikelos:
Fixes #532 by caching pointer dereferences explicitly, since we found an instance where a pointer was reused with a different vol subtype dict and so was returning the cached object from a new pointer.  Now, rather than the class hanging onto the cache and indexing it on self and layer_name, the instance itself keep a copy of its own dereference keyed by layer_name.  A new parameter `new_instance` was added to the dereference call to allow new instances to be created if wanted.  This now means that once a pointer is referenced it will be maintained by the object's calling stack, so may create the memory usage in certain conditions.

I don't think this will have a knock on effect because multiple context references, etc are being stored, but this is all deep low level stuff, which is why I wanted a review of it before commiting it directly...

@japhlange if you could check it over too, that'd be much appreciated.  The more eyes the better on something like this...  5;)

### DIFF
--- a/API_CHANGES
+++ b/API_CHANGES
@@ -1,0 +1,4 @@
+Lists of changes made to API in each version
+============================================
+
+1.2.1  Added new_instance flag to pointer.dereference call
--- a/volatility3/framework/constants/__init__.py
+++ b/volatility3/framework/constants/__init__.py
@@ -39,7 +39,7 @@ BANG = "!"
 
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 1  # Number of releases of the library with a breaking change
-VERSION_MINOR = 1  # Number of changes that only add to the interface
+VERSION_MINOR = 2  # Number of changes that only add to the interface
 VERSION_PATCH = 1  # Number of changes that do not change the interface
 VERSION_SUFFIX = ""
 
--- a/volatility3/framework/objects/__init__.py
+++ b/volatility3/framework/objects/__init__.py
@@ -3,7 +3,6 @@
 #
 
 import collections
-import functools
 import logging
 import struct
 from typing import Any, ClassVar, Dict, List, Iterable, Optional, Tuple, Type, Union as TUnion, overload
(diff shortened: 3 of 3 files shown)


japhlange:
Looks good to me, without having tested it. What I don't understand though is how this prevents the error - wouldn't I (or someone else along the call chain) need to pass `new_instance` as `True` if the pointer is a null pointer so the type/attribute mixup does not happen?

ikelos:
No, that was just me guessing at what the problem could be.  I'm still not entirely certain what the issue with the mru_cache was, but it seems as though just taking the object's memory location isn't enough to determine whether it's changed or not, whereas actually stashing the value into `self` means that the cache isn't shared across the class and differentiated by `self` (which caused the problems), but is instead stored as part of `self`.  So I've tested this on the example, and this succeeds whereas the caching fails (and removing the cache also succeeds, but comes with the associated problem of the value being rebuilt each time).

The `new_instance` value is just there should anyone ever need the value recalculating (otherwise they'd need to do a cast on the value to itself and that would go oddly).

ikelos:
Ping @atcuno @npetroni @iMHLv2 , please could one of you give this a once over and make sure it looks sane/kick the tyres?  5:)

ikelos:
This was resolved with some urgency in commit 260f87933406e4646f2cd12db5b76b88214174ae.  The new commit doesn't change the API (but also doesn't allow forcibly clearing the cache).  It also only caches once per object, as opposed to once per object per layer, which is a change I may add.

PR #538, opened 2021-07-22
Feature/better modules support

### LINKED ISSUE #488: Potential confusion in the API regarding symbol offsets
This is to keep track of how we deal with a potential confusion that got introduced to the API.  There was a direct request to allow all symbols in a particular symbol table to be shifted by a fixed amount, there was also a separate request for a `module` which would construct an object based on a symbol with the module at the offset *within the module*.  We implemented both mechanisms.  We allowed the module to be created with a 0 offset if necessary, in which case no symbol addresses change.  What that gives us is two ways that the symbol addresses may get modified for a module, and this could lead to the change being added twice, causing confusion.

Mechanism A: 
Load symbol table with `symbol_shift` by `module_offset` amount 
* All symbol addresses come out as `symbol_address + module_offset`

Mechanism B:
Construct module with `module_offset` as parameter
* symbols and their addresses come out as they were in the symbol table
* objects created using `module.create` have the `module_offset` added to them automatically

Trouble condition:
Module constructed with `offset` parameter, then symbol table created with `symbol_shift` parameter, then object constructed using module
* Outcome: object constructed at `symbol_address + module_offset + module_offset`

This feels like we messed up.  The module mechanism was supposed to be handy, but we wanted the data only to be changed as needed (on use) and for the symbol addresses to be exactly what's been read from the data.  The `symbol_shift` was strongly requested and is now deeply embedded in the linux system, due to ASLR (rather than having to add the ASLR value to every symbol).

I'm not sure how best to resolve this, the big unknown is whether symbol addresses for a module will ever be used *without* adding the module offset to them?  If not, I suggest we deprecate the `offset` value of the module interface (checking there's no other use cases for what it provides) and move to `symbol_shift` as the preferred mechanism.  If there are cases where you might need the original symbol address value, then we need to figure out what we're gonna do and how we're gonna avoid it all getting thoroughly confusing for people.

@npetroni I'm after your critical logic first, but @iMHLv2 and @atcuno I wanted you both to be in the loop in case you've got a preference over how this is resolved...

ikelos:
This one's pretty big and will need some carefully looking over because it touches so much and what it looks to deprecate is fairly deeply woven in.  Essentially the `symbol_shift` feature (and to some extent the `symbol_mask` feature) change the semantics of a symbol table, meaning each symbol table needs to be loaded per module, and the original data of the symbols is difficult to get to.

This patchset introduces a new means of handling things, a ModuleRequirement which will return a module (that is carried around by and looked up from the context).  These modules are an offset to a location in memory and a symbol table.  This allows for the same library to make use of the same symbol table, even if it is loaded twice at different offsets.  It also doesn't grant the kernel some form of increased privilege, or confuse people that load symbol tables because the kernel ones have been offset by the automagic, but normal symbols aren't.

This predominantly impacts the LinuxUtilties which now needs to be given a module name in some places, rather than a flat symbol table.  There's one chunk of code in particular that's only passed a symbol table and we try to figure out the module from that.  All of this is new and will need reviewing with a careful eye.  It's been tested with a single ASLR linux image and some non-ASLR linux images, and a single Mac image.  These all seemed to work, but the more eyes the better.

Windows plugins haven't been converted except for pslist because a) there isn't really a need since they never used the symbol shift mechanism and b) pslist proves that it can be done and work for windows plugins.  We can convert the rest, but this felt like enough of a change in itself.

Since it's such a big change, I'd like all hands on deck please.  I've managed to avoid significantly changing any of the main API, but the LinuxUtilities class underwent a number of changes and so has been separately versioned.  This will likely mean that all linux plugins will need their versions bumping at a minimum (and a check to make sure they work with the new API), but this also means that this commit itself isn't the right one to roll API breaks into.  There will be another one later on that does carry out the break and removes the symbol_shift functionality, but that's further down the line so we deprecate it and then give people time to remove their reliance on it, before it goes.  We should also consider whether we roll a release before this, after this or after the main break.  We should be coming up on another official release, so we just want to figure out how much potential churn to put developers through...

Any questions, just shout...  5:)

### DIFF
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -1,0 +1,19 @@
+API Changes
+===========
+
+When an addition to the existing API is made, the minor version is bumped.
+When an API feature or function is removed or changed, the major version is bumped.
+
+
+1.2.0
+=====
+* Added support for module collections
+* Added context.modules
+* Added ModuleRequirement
+* Added get\_symbols\_by\_absolute\_location
+
+* Remove support for symbol\_shift and symbol\_mask from symbol tables
+  Symbols should be the data values from the JSON, and if they need modifying,
+  a module wrappr, or similar, should be used
+
+
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -12,8 +12,4 @@ When an API feature or function is removed or changed, the major version is bump
 * Added ModuleRequirement
 * Added get\_symbols\_by\_absolute\_location
 
-* Remove support for symbol\_shift and symbol\_mask from symbol tables
-  Symbols should be the data values from the JSON, and if they need modifying,
-  a module wrappr, or similar, should be used
-
 
--- a/volatility3/framework/automagic/__init__.py
+++ b/volatility3/framework/automagic/__init__.py
@@ -21,11 +21,13 @@ from volatility3.framework.configuration import requirements
 
 vollog = logging.getLogger(__name__)
 
-windows_automagic = ['ConstructionMagic', 'LayerStacker', 'WintelHelper', 'KernelPDBScanner', 'WinSwapLayers']
+windows_automagic = [
+    'ConstructionMagic', 'LayerStacker', 'WintelHelper', 'KernelPDBScanner', 'WinSwapLayers', 'KernelModule'
+]
 
-linux_automagic = ['ConstructionMagic', 'LayerStacker', 'LinuxBannerCache', 'LinuxSymbolFinder']
+linux_automagic = ['ConstructionMagic', 'LayerStacker', 'LinuxBannerCache', 'LinuxSymbolFinder', 'KernelModule']
 
-mac_automagic = ['ConstructionMagic', 'LayerStacker', 'MacBannerCache', 'MacSymbolFinder']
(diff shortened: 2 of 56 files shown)


atcuno:
These changes look good to me and the resulting code in the plugins all looks much smoother. The only thing I noticed was this instantiation of the vmlinux variable looked unnecessary unless something is going on in the background with it:

https://github.com/volatilityfoundation/volatility3/pull/538/files#diff-7a4e59076cb947ec77ba506b4386867f759b7573154d91888cc15f90d3de82ffR37

I ran through several Linux and Mac samples with all the plugins use this branch.

ikelos:
Sorry, not sure what you mean by the instantiation of the vmlinux variable?  The link seems to be a diff of the entire lot, so I'm not sure which bit you're referring to?

ikelos:
Also, are we happy with this only being linux and mac plugins, and doing the windows plugins later?

ikelos:
Ok, I'm gonna merge this as-in, and then add another request that rips out the `symbol_shift` and `symbol_mask` features and bumps the API to 2.0.0, the other two API changes alterations can get merged at the same time as that one...

PR #543, opened 2021-07-27
Skip paged out and/or smeared Mac process maps, add debug statement w…

atcuno:
…hen breaking iteration


atcuno:
@ikelos this is to fix a bug found in testing

PR #546, opened 2021-07-27
Fix several code paths that could infinite loop and were triggered ba…

atcuno:
…sed on @fgomulka testing

### DIFF
--- a/volatility3/framework/plugins/mac/list_files.py
+++ b/volatility3/framework/plugins/mac/list_files.py
@@ -44,20 +44,22 @@ class List_Files(plugins.PluginInterface):
         return v_name
 
     @classmethod
-    def _get_parent(cls, vnode):
-        parent = None
-
+    def _get_parent(cls, context, vnode):
         # root entries do not have parents
         # and parents of normal files can be smeared
         try:
-            parent = vnode.v_parent
+            parent = vnode.v_parent.dereference()
         except exceptions.InvalidAddressException:
-            pass
+            return None
+
+        if parent and not context.layers[vnode.vol.native_layer_name].is_valid(parent.vol.offset,
+                                                                               parent.vol.size):
+            return None
 
         return parent
 
     @classmethod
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/mac/list_files.py
```diff
@@ -44,28 +44,34 @@ def _vnode_name(cls, vnode: interfaces.objects.ObjectInterface) -> Optional[str]
...
+
+        if parent and not context.layers[vnode.vol.native_layer_name].is_valid(parent.vol.offset,
+                                                                               parent.vol.size):
+            return None
 
         return parent
 
     @classmethod
-    def _add_vnode(cls, vnode, loop_vnodes):
+    def _add_vnode(cls, context, vnode, loop_vnodes):
```
ikelos, line 62: This is technically an API change depending on whether `_add_vnode` is classed as a public method or a private one?  If vnode is an object, you could use vnode's context.  I don't see too much problem in doing that, because I can't consider a situation where someone would have made a vnode in one context and then want to pass in a different context?

ikelos (commented):
So, the one big thing with all of these is that they change the signature for published functions that could be seen as part of the API.  If you can ensure that these aren't called externally then it's ok (they do start with `_` meaning they should be considered internal functionality), but otherwise we need to get creative with how you get the context out...

So the options are:
a) The methods are private and it can stay as is
b) The methods are private, but we can use the context from the other objects anyway
c) The methods aren't private, but we can use the context from the other objects not to change the API
d) The methods aren't private, and we'll going to bump the major API version number

I'd strongly try to dissuade people from choosing d, but I think there's enough other options we can find something workable...  5;)

ikelos:
Any progress on this?  It should be easy enough to keep backwards compatibility?

ikelos:
It was decided this was a private method, and therefore did not require a major version bump.  Merging as is...

PR #555, opened 2021-08-30
Fixed an issue where the symbol file download from the Microsoft site failed

shu-tom:
I found an issue with the _get_guid_from_mz_ function where the guid value was not zero padded. This issue causes the PDB file download from the Microsoft website to fail.


ikelos:
Ah, really good spot, thanks very much!  5:D

ikelos (approved):
Yep, looks good...  5:)

PR #559, opened 2021-09-03
Feature/unify dtb selfref checks

### LINKED ISSUE #505: PML4 self reference scanning error in some Windows dumps
Hello!

There was good patch for supporting fresh Windows versions:

https://github.com/volatilityfoundation/volatility3/commit/a1c5f5e5e5ed2ba53711b34eeb63271fddf18ef2#diff-95515258a22a0ced4292d19bfd7dd395e507e8a7f9925efb6c4ec4207b00dd41R131

But some images have ptr_reference below, then 0x1E0, for example 0x10a:

0: kd> ? (poi(nt!MmPteBase) >> 0n39) & 0x1ff
Evaluate expression: 266 = 00000000`0000010a

0: kd> vertarget
Windows 10 Kernel Version 19042 MP (2 procs) Free x64
Product: WinNt, suite: TerminalServer SingleUserTS
Edition build lab: 19041.1.amd64fre.vb_release.191206-1406
Machine Name:
Kernel base = 0xfffff80679217000 PsLoadedModuleList = 0xfffff80679e41310

Is it possible to change bottom edge ptr_reference range in windows.py or add parameter for custom ptr_reference configuring in vol.py?

ikelos:
Ok,

So this is a *big* shift of how the windows automagic works, but there were a lot of false starts that needed fixing (particularly to support #505).  This does away with the duplicated DTB checkers.  It limits DTB checking to very specific ranges (which may need to be enlarged eventually), but should handle the more recent windows 10 range of self-referential pointers (randomized between entry 0x100 and 0x1ff) as well as the older specific self-referential parameters that might have happened in a larger chunk of starting memory.

I've tested this on all the images I've got access to and it seems to work on all previous images, but I think they may be one where we just have to bite the bullet and apply it and tweak it over time if we get outliers...

So questions are, should I put back the old code and make it into a switch so we can choose or will that just complicate things?


ikelos:
Ok, I've run this against every windows image I had, it finds all the DTBs the previous one did, plus one that previous code didn't, and it correctly identifies two DTBs that were previously mis-identified.  The code is much easier to understand and follow, the test regions and tests are neatly pulled out.  I'm not certain what speed impact it will have, but I think in the general case it will speed up identification too.

Happy for people to review this now (sorry for the early tag), but should be good to go now....  5:)

PR #563, opened 2021-09-08
Doc: list_head.to_list, pslist pid 0

### LINKED ISSUE #501: linux.pslist not listing init_task / listhead "skipping" start element.
**Describe the bug**
The pslist plugin does not list the init task. (pid 0) I do not know if that is intentional, to me it seems like a bug.

The underlying problem / way to fix it has two ways of looking at it. That is because of a certain detail in `volatility3.framework.symbols.linux.extensions.list_head.to_list`:  We initialize the `seen` set containing `self.vol.offset`.
That means that the `__iter__` function of the `list_head` will not `yield` the element from which the iteration was started. 
(In the case of PsList its the `init_task`)
This behaviour may or may not make sense depending on how you look at it. Either way, it should be specified and documented. 

Personally I think that the `seen` set should be initialized empty, so the initial element will be yielded in the last iteration if the linked list is "closed". 

The other solution would be to have `volatility3.framework.plugins.linux.pslist.PsList.list_tasks` `yield` the `init_task` explicitily. (If the filter function permits that.)

**Context**
Volatility Version:  1.0.1
Operating System:  Linux
Python Version:  3.9.4
Suspected Operating System:  
Command:  vol -f .. `linux.pslist.PsList`

**To Reproduce**

1. Use command ' vol -f /your/memdump `linux.pslist.PsList`
2. See that pid 0 is missing.

**Expected behavior**
Process with pid 0 should be listed.

NiklasBeierl:
Added Documentation for the params of list_head.to_list

Added a comment to linux.pslist explaining why the init task is not yielded.

Closing #501.

### DIFF
--- a/volatility3/framework/plugins/linux/pslist.py
+++ b/volatility3/framework/plugins/linux/pslist.py
@@ -78,6 +78,7 @@ class PsList(interfaces.plugins.PluginInterface):
 
         init_task = vmlinux.object_from_symbol(symbol_name = "init_task")
 
+        # Note that the init_task itself is not yielded, since "ps" also never shows it.
         for task in init_task.tasks:
             if not filter_func(task):
                 yield task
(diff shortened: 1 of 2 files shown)


ikelos (commented):
Looks good, thanks!  5:)

PR #575, opened 2021-10-06
refs #566 refactor dependencies

### LINKED ISSUE #566: Document setup.py requirements, or implement requirements.txt file
This is so we don't forget the discussion.  A lot of users don't know where to look for the requirements, or use the existing requirements in `setup.py` (I believe there's a way of tagging an install to say which dependencies are needed).

The goals we're looking for are:
* Easy to install all requirements
* Easy to install bare volatility with as few additional requirements as possible (the rationale for this is if an optional subdepencny fails to compile or install, we still want people to be able to use the bulk of volatility).
* Scalability in future, even if there are many additional optional requirements.

We can: 
* document things better to give people instructions as to how to grab the dependencies
* move to `requirements.txt` but since it doesn't do dependency handling well, we'd need a `requirements-alldeps.txt` as well
* Investigate an alternative mechanism (toml file, etc?)
* Some other solution (Makefile, etc?)

Happy with any of the options, but we should think it through and decide what to do...

iMHLv2:
# Bugs and fixes 

* README said pefile>=2017.8.1 but setup.py said "pefile"
* README said python >= 3.6.0 but setup.py said python >= 3.6.0
* The oldest pycryptodome available is 3, so >= 3 is the same as not mentioning a version at all 
* setup.py mentions capstone for Linux and capstone-windows for Windows, but no versions for either, and nothing for Mac. The README says capstone >= 3.0.0. If we bump to 3.0.5 then the "capstone" package can be used for all 3 platforms. 

# Refactors

* Moved sphinx to doc/requirements.txt (out of sight, this is mainly for us and other devs)
* Moved core reqs to requirements-minimal.txt, and that is read dynamically by setup.py 
* The requirements.txt contains all dependencies, with comments 
* README discusses how to install with minimal, partial, and complete dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,13 +18,24 @@ the Volatility Software License (VSL). See the [LICENSE](LICENSE.txt) file for m
 
 ## Requirements
 
-- Python 3.6.0 or later. <https://www.python.org>
-- Pefile 2017.8.1 or later. <https://pypi.org/project/pefile/>
+Volatility 3 requires Python 3.6.0 or later. To install the most minimal set of dependencies (some plugins will not work) use a command such as:
 
-## Optional Dependencies
+```shell
+pip3 install requirements-minimal.txt
+```
 
-- yara-python 3.8.0 or later. <https://github.com/VirusTotal/yara-python>
-- capstone 3.0.0 or later. <https://www.capstone-engine.org/download.html>
+Alternately, the minimal packages can be installed automatically when Volatility 3 is installed. However, as noted in the Quick Start section below, Volatility 3 does not *need* to be installed prior to using it.
+
+```shell
+python3 setup.py build 
+python3 setup.py install
+```
+
+To enable the full range of Volatility 3 functionality, use a command like the one below. For partial functionality, comment out any unnecessary packages in [requirements.txt](requirements.txt) prior to running the command.
+
+```shell
+pip3 install requirements.txt
+```
 
 ## Downloading Volatility
 
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ the Volatility Software License (VSL). See the [LICENSE](LICENSE.txt) file for m
 Volatility 3 requires Python 3.6.0 or later. To install the most minimal set of dependencies (some plugins will not work) use a command such as:
 
 ```shell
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -18,13 +18,24 @@ the Volatility Software License (VSL). See the [LICENSE](LICENSE.txt) file for m
...
-- Python 3.6.0 or later. <https://www.python.org>
-- Pefile 2017.8.1 or later. <https://pypi.org/project/pefile/>
+Volatility 3 requires Python 3.6.0 or later. To install the most minimal set of dependencies (some plugins will not work) use a command such as:
 
-## Optional Dependencies
+```shell
+pip3 install -r requirements-minimal.txt
+```
+
+Alternately, the minimal packages can be installed automatically when Volatility 3 is installed. However, as noted in the Quick Start section below, Volatility 3 does not *need* to be installed prior to using it.
```
ikelos: Add in something about `when using setup.py`, just so it's clear how volatility 3 needs to be installed, to get the dependencies automatically...

requirements.txt
```diff
@@ -0,0 +1,24 @@
...
+# It can also improve accuracy of Windows 8 and later memory samples.
+capstone>=3.0.5
+
+# This is required by plugins that decrypt passwords, password hashes, etc.
+pycryptodome
+
+# This can improve error messages regarding improperly configured ISF files.
+jsonschema>=2.3.0
+
+# This is required for memory acquisition via leech.
```
ikelos: I think it's either leechcore or pcileech.  Just leech on its own isn't quite enough to help people track it down...

requirements.txt
```diff
@@ -0,0 +1,24 @@
...
+leechcorepyc>=2.4.0
+
+# This is required for analyzing Linux samples acquired with AVML.
```
ikelos: Only for AVML compressed format files, but I dunno if that needs pointing out.  It's possible someone could acquire using AVML but save in LiME format?  Probably not worth bothering over...  5:P

ikelos (approved):
Cool, looks good.  Just a couple minor things, but no major blockers.  Happy to merge then fix, or have you fix and then merge, we'll see which happens first...  5;)

PR #579, opened 2021-10-13
Windows: Set a 100Gb VAD max size for vadinfo

### LINKED ISSUE #578: windows.vadinfo.VadInfo `--maxsize` has no default value
**Describe the bug**
When I try to run `windows.vadinfo.VadInfo` against a certain memory image, one of the allocations looks like the following:

```
  |  PID |        Process |         Offset |     Start VPN |       End VPN |  Tag |                       Protection |     CommitCharge | PrivateMemory |         Parent |                                                                                                                                                File | File output
* | 2516 |   iexplore.exe | 0xfa8003b0a810 |    0x7fff0000 | 0x7fffffeffff | Vadl |                    PAGE_READONLY | 2251799813685247 |             1 | 0xfa8004e0fdf0 |                                                                                                                                                 N/A |    Disabled
```
So that would account for 8793945538559 bytes which simply can't be true.

**Context**
Volatility Version:  2d39dc5a919f02722f41d83f9c289b6cc6cb07c3
Operating System:  Debian 11
Python Version:  3.9.2
Command:  `python3 vol.py -o results/vads/ -r pretty -f win7-64-Snapshot7.vmem windows.vadinfo.VadInfo --dump 1>./results/windows.vadinfo.VadInfo.txt`

**To Reproduce**
N/A. Might be related to this particular memory image.

**Expected behavior**
Volatility3 not filling my disk with hundreds of gigabytes of data.

**Screenshots**
```
-rw------- 1 xyz xyz  261G Oct 12 12:46 pid.2516.vad.0x7fff0000-0x7fffffeffff.dmp
```

ikelos:
Potential fix for #578

@iMHLv2 is 100Gb a reasonable limit (ie, one that we are unlikely to have to change in the near future, but won't lead people to run out of disk disk space so quickly as without a limit)?

### DIFF
--- a/volatility3/framework/plugins/windows/vadinfo.py
+++ b/volatility3/framework/plugins/windows/vadinfo.py
@@ -35,7 +35,7 @@ class VadInfo(interfaces.plugins.PluginInterface):
 
     _required_framework_version = (2, 0, 0)
     _version = (2, 0, 0)
-    MAXSIZE_DEFAULT = 100 * 1024 * 1024 * 1024  # 100 Gb
+    MAXSIZE_DEFAULT = 1024 * 1024 * 1024  # 1 Gb
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


ikelos:
We could also move this value into the volatility windows constants to make it easier to change in a single place and in the future potentially have them be set by configuration files?

iMHLv2:
My opinion would be to set the default much lower. For example, vol2 uses 1 GB as a maximum.

ikelos:
Sure, sounds good, thanks!  5:)

PR #582, opened 2021-10-18
fix some typo

garanews:


ikelos (approved):
Wow, I never realized we had so many typos (and most of them are probably mine!).  Thanks very much for going through and finding all of them, it's much appreciated!  5:)

PR #584, opened 2021-10-24
issue-449 - add --physical flag

### LINKED ISSUE #449: Psscan doesn't seem to provide a flag --physical
The offset printed in psscan is the virtual offset. However, the possibility of providing a flag --physical, which will change the output to the physical address, is missing. 
The plugin pslist already supports this feature. A similar solution for psscan is still missing.

CaptWake:


volatility3/framework/plugins/windows/psscan.py
```diff
@@ -39,6 +40,10 @@ def get_requirements(cls):
             requirements.BooleanRequirement(name = 'dump',
                                             description = "Extract listed processes",
                                             default = False,
+                                            optional = True),
+            requirements.BooleanRequirement(name = 'physical', 
+                                            description = "Display physical offset instead of virtual",
+                                            default = cls.PHYSICAL_DEFAULT,
```
ikelos: This should probably be a specific default, because if this is set to true, I don't think adding `--physical` would turn it off (and I'm not sure that would make sense even if it did).

So whilst I'm actually a big fan of having the default easily accessible up at the top, and appreciate you adding it into the PR, I'd probably fix it at being false and that means you can probably do away with the `PHYSICAL_DEFAULT` variable...

ikelos (approved):
Generally looks awesome, thanks very much!!!  5:D  There's one little bit about the `PHYSICAL_DEFAULT`, but I'm still happy to take it as is if you want.  Lemme know if you want/will have time to fix it, otherwise happy to merge it...

Thanks again, really glad this issue got picked up by someone, and really pleased there's such a nice PR to go with it!  5:)

volatility3/framework/plugins/windows/psscan.py
CaptWake: Understand, thanks for the suggestion, I committed the changes.

ikelos (approved):
Yep, looks good, thanks!  5:D

PR #586, opened 2021-11-08
Layers: Speed up qemu by not bisecting

### LINKED ISSUE #585: QEMU layers are extremely slow
I generate lots of snapshots for validating forensics heuristics. I do that by spinning up qemu-vms and "virsh dump"ing their memory. As volatility points out, working with qemu files is slow, conversion to "raw" files with the `layerwriter` pllugin is the solution. However, running the layerwriter plugin itself is also quite slow, especially if you do it for dozens of snapshots.

Using `-vvvv` on my vol command, I saw that the plugin spends quite some time with trying different stackers:
```
Level 8  volatility3.framework.automagic.stacker: Stacked QemuSuspendLayer using QemuStacker
Level 8  volatility3.framework.automagic.stacker: Attempting to stack using Elf64Stacker
Level 6  volatility3.framework.layers.elf: Exception: Bad magic 0xf000ff53 at file offset 0x0
Level 8  volatility3.framework.automagic.stacker: Attempting to stack using LimeStacker
Level 8  volatility3.framework.automagic.stacker: Attempting to stack using WindowsCrashDumpStacker
Level 8  volatility3.framework.automagic.stacker: Attempting to stack using VmwareStacker
Level 8  volatility3.framework.automagic.stacker: Attempting to stack using WindowsIntelStacker
DEBUG    volatility3.framework.automagic.windows: Self-referential pointer not in well-known location, moving to recent windows heuristic
Level 8  volatility3.framework.automagic.stacker: Attempting to stack using MacIntelStacker
INFO     volatility3.framework.automagic.mac: No Mac banners found - if this is a mac plugin, please check your symbol files location
Level 8  volatility3.framework.automagic.stacker: Attempting to stack using LinuxIntelStacker
DEBUG    volatility3.framework.automagic.linux: Identified banner: ...
```
As a first speedup I would like to tell the layerwriter plugin to go straight to the LinuxIntelStacker. That could perhaps be achieved with an additional cli option.

I could probably go figure into the code and change the order in which the Stackers are attempted, but I haven't gotten around to that yet. I'd highly appreciate hints.

ikelos:
So we previously were bisecting the address space to determine whether the data was compressed.  Instead we now mask the low 12 bits, because QEMU pages are a fix size of 0x1000 bytes.  This provides a significant speed-up, but testing to ensure that the results are identical to the old implementation would be useful...

Fixes #585

### DIFF
--- a/volatility3/framework/layers/qemu.py
+++ b/volatility3/framework/layers/qemu.py
@@ -1,7 +1,6 @@
 # This file is Copyright 2020 Volatility Foundation and licensed under the Volatility Software License 1.0
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
-import bisect
 import functools
 import json
 import math
@@ -218,9 +217,8 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
         of the starting data.  It is the responsibility of the layer to turn the provided data chunk into the right
         portion of data necessary.
         """
-        start_offset, _, start_mapped_offset, _ = self._segments[
-            bisect.bisect_right(self._segments, (offset, 0xffffffffffffff,)) - 1]
-        if start_mapped_offset in self._compressed:
+        start_offset = offset ^ (offset & 0xfff)
+        if start_offset in self._compressed:
             data = (data * 0x1000)
         result = data[offset - start_offset:output_length + offset - start_offset]
         return result


ikelos:
If this speed things up, we can also remove the "slow layer" warning.

ikelos:
No response back on issue #585, so hopefully this actually fixes the problem, but I guess we need to try breaking to find out...

PR #593, opened 2021-12-02
Mountinfo

gcmoreira:
The Linux **MountInfo** plugin provides information about mount points on every single mount namespaces.
It has two output formats and it could also be easily split in two different plugins if using module arguments is not desired.

1) MountInfo format: The default view includes a very detailed output showing pretty much the same information than the 
userspace **_/proc/[pid]/mountinfo_** counterpart. It distinguishes between 'mount' and 'superblock' options, shows propagation 
state, mount and parent identifiers and other information that is missing in the older /proc/[pid]/mounts file.
2) Mount format: A simplify view which mimics the output of _**/proc/[pid]/mounts**_ or **_mount_** tool.

The default behavior is to show the mount point information per namespace. 

Tasks can be filtered using the `--pids` argument. Example `... --pids 678 1200`. When this plugin argument is used, a new column is added to the output showing the respective PID.

The output can also be filtered by mount namespace id/s, using the `--mntns` agument. For instance, `... --mntns 4026531840 4026532841` 

An older version of this plugin was submitted by myself in the Volatility 3 plugin contest 2020. You can find the old code [here](https://github.com/volatilityfoundation/community3/tree/master/Moreira_Mountinfo). 

### Summary
* Allows two format outputs, _mountinfo_ and _mount_. This could also be split in two different plugins.
* Allows to filter the output by PID/s.
* Allows to filter the output by mount namespace id/s.
* Distinguishes between 'mount' and 'superblock' options.
* Shows propagation state.
* Mount and parent identifiers and other information that is missing in the older /proc/[pid]/mounts  file.
* Adds a new _do_get_path() implementation based on the Linux kernel source code. The current implementation in LinuxUtilities, which is a legacy of Volatility 2, is incorrect and leads to inaccurate information. Unfortunately, I was unable to fix that one because it was adapted with some hacks/workaround to provide same extra functionality to other parts of the framework which made hard to extend. 
* Extended many linux/extension types with helper functions to provide similar functionality than the
  Linux Kernel. i.e.: superblock, mount, dentry and added mnt_namespace.
* Replaced linux/extension dentry->path() method with a implementation which mimics the Linux kernel 
  __dentry_path() function. The old one didn't return the path but just the dentry name.

### Testing
The plugin was tested using the following Linux kernels and distros:
 * 3.10.0-862 / CentOS 7
 * 3.13.0-117 / Ubuntu 12.04
 * 4.4.0-142 / Ubuntu 14.04 
 * 4.10.0-42 / Ubuntu 16.04
 * 4.15.0-45 / Ubuntu 16.04
 * 4.18.0-10  / Ubuntu 18.10
 * 5.8.0-53 / Ubuntu 20.04
 * 5.10.0-14 / Ubuntu 20.04

### Limitations
* Unfortunately, the symbol types needed to parse specific filesystem flags and options are not available in the kernel debug 
symbols, so that information is missing in the output.
* It uses `linux.pslist` to obtain the list of tasks. It could take a while in systems where there are a large number of processes.


volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -0,0 +1,227 @@
...
+
+        sb_opts: List[str] = []
+        sb_opts.append(superblock.get_flags_access())
+        sb_opts.extend(superblock.get_flags_opts())
+
+        return MountInfoData(mnt_id, parent_id, st_dev, mnt_root_path, path_root, mnt_opts, fields, 
+                             mnt_type, devname, sb_opts)
+
+    def _get_mnt_namespace_mountpoints(self, mnt_namespace):
+        mnt_type = self._get_symbol_fullname("mount")
```
ikelos: Maintaining state to use the symbol table is a little bit... it kinda feels bad.  I'm trying to think how I'd do this.  I think originally symbols were supposed to lookup without a symbol table name for conveniences, but I think that fell by the wayside.  As such, I'd still suggest creating the module from `self.config['kernel']` just so that if the this could be called without `_get_tasks_mountpoints` having been executed.  Not a major point, and not a blocker, but if you are going to add state, I'd feel safer if you ensure the variable is defined in `__init__`.

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -0,0 +1,227 @@
...
+        if not self.context.symbol_space.has_type(mnt_type):
+            # Old kernels ~ 2.6
+            mnt_type = self._get_symbol_fullname("vfsmount")
+
+        for mount in mnt_namespace.list.to_list(mnt_type, "mnt_list"):
+            yield mount
+            
+    def _get_tasks_mountpoints(self, pids: Iterable[int]):
+        self._vmlinux = self.context.modules[self.config['kernel']]
```
ikelos: `__init__` should call super and also define any local state (like `_vmlinux`).

ikelos: I'd also be happy if the symbol_tablename were instead passed manually into `_get_mnt_namespace_mountpoints` and `_get_symbol_fullname`.

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -0,0 +1,227 @@
...
+            
+            for mount in self._get_mnt_namespace_mountpoints(mnt_namespace):
+                yield task, mount, mount_ns_id
+
+    def _generator(self):
+        pids = self.config.get('pids')
+
+        for task, mnt, mnt_ns_id in self._get_tasks_mountpoints(pids=pids):
+            mnt_info = self.get_mountinfo(mnt, task)
+            if mnt_info is None:
```
ikelos, line 167: I'm not a massive fan of continue unless there's extremely good cause.  It feels here as though the if statement could be inverted and all the following lines indented to achieve the same result.  Again, not a block, just a stylistic thing.  5:)

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -0,0 +1,227 @@
...
+            fields_values.extend(extra_fields_values)
+            
+            yield (0, fields_values)
+
+    def run(self):
+        if self.config.get('all-processes') and self.config.get('pids'):
+            raise ValueError("Unable to use --all-processes and specified a pid")
+        
+        # When no arguments are specified, it displays the mountpoints per namespace
+        self._show_mountpoints_per_namespace = not any([self.config.get('pids'), self.config.get('all-processes')])
```
ikelos: So it feels like there's 3 states here:
* pids set
* pids not set and all-processes
* pids not set and not all-processes

It might be helpful to consider if it's neccessary for all of them (I think basically, I'm not quite clear on what `show_mountpoints_per_namespace` would do and why it *wouldn't* output pids?

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -0,0 +1,227 @@
...
+            extra_columns = [("DEVNAME", str), ("PATH", str), ("FSTYPE", str), ("MNT_OPTS", str)]
+        else:
+            # /proc/[pid]/mountinfo output format
+            extra_columns = [("MOUNT ID", int), ("PARENT_ID", int), ("MAJOR:MINOR", str), ("ROOT", str),
+                             ("MOUNT_POINT", str), ("MOUNT_OPTIONS", str), ("FIELDS", str), ("FSTYPE", str),
+                             ("MOUNT_SRC", str), ("SB_OPTIONS", str)]
+            
+        columns.extend(extra_columns)
+
+        return renderers.TreeGrid(columns, self._generator())
```
ikelos: This is kind of icky.  The number of generator can produce different numbers of columns based on state (`_show_mountpoints_per_namespace`) rather than a parameter, and the columns are separately set by the same state.  I'd be much happier with either:
* A system that always included the pid (again, may just be me not understanding in what circumstances you *wouldn't* had an associated pid)
* An if statement that had the two column lists and then provided a parameter to the generator, so that you could clearly see based on the parameter that you either get one type of output or another.

I'm thinking of this for other plugins that might want to use the information and them needing to figure out which form the plugin returns.  I'd also be tempted to just fill the Pid column with `NotApplicable` so that the shape of the output data doesn't change between runs/options...

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -29,6 +29,7 @@ def __init__(self, *args, **kwargs) -> None:
         self.set_type_class('files_struct', extensions.files_struct)
         self.set_type_class('vfsmount', extensions.vfsmount)
         self.set_type_class('kobject', extensions.kobject)
+        self.set_type_class('mnt_namespace', extensions.mnt_namespace)
```
ikelos, line 32: Changes to core framework files are.. delicate.  We can do this, but it might be nice to give people an example of how to keep it all self contained.  This would basically mean calling this on the `symbol_table` of the `kernel` within the plugin itself.  This is just because not every plugin will be able to make core changes like this, so the more examples of how to do it without specialist help, the better.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -108,7 +108,7 @@ def get_symbols(self):
                                                                  "linux",
                                                                  "elf",
                                                                  native_types = None,
-                                                                 class_types = extensions.elf.class_types)
+                                                                 class_types = elf.class_types)
```
ikelos: Good catch, but since this is a critical and standalone fix, I've made it directly to the tree.  It just makes it decouples the pull request from a problem fix, in case we need to track it down at some point.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -373,7 +410,44 @@ def name_as_str(self) -> str:
 class dentry(objects.StructType):
 
     def path(self) -> str:
-        return self.d_name.name_as_str()
+        """ Based on __dentry_path Linux kernel function"""
```
ikelos: How do these changes affect existing linux plugins?  I'm going to ask @atcuno to review this because I don't know these functions well enough...

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -498,6 +593,66 @@ def get_mnt_parent(self):
     def get_mnt_mountpoint(self):
         return self.mnt_mountpoint
 
+    def get_flags_access(self) -> str:
```
ikelos, line 644: These seems useful, so I'm willing to accept them as a change to the core files as long as @atcuno is happy with them, since they'll be a change to the API.  This, as an additive change to the API, should also include a bump to the minor framework version in the `constants/__init__.py` file.  This is so that plugins that need it can say they need at least this version to operate.  Your plugin making use of these should also require the bumped version of the framework.  No other plugins should need to be changed though...

ikelos (commented):
Ok, very cool, thanks for this!  It looks really good, the style is good, docstrings are good.  I've added a number of comments, some of which are just stylistic, but towards the end some of the changes are to the core API and will need a little further review.  I'd really like to get @atcuno to give these changes the OK before I merge them, so that's what this is waiting on at the moment.  If you still haven't heard anything in a few weeks, please give me a prod...  5:)

volatility3/framework/plugins/linux/mountinfo.py
gcmoreira: Yeap, sorry. Absolutely, it could be inverted as you said, but I guess it somehow comes to the eternal debate of optimal line limits. In Python, I prefer to return/abort/skip loops/functions asap and avoid nested code contexts. IMO that way helps to keep the code simple, readable and within reasonable line limits. Otherwise, easily you exceed the optimal line length forcing you to massacre your code losing readability. Anyway, probably it's not exactly that case and we can avoid it here. Let me know if you want me change that or otherwise feel free to change that yourself

volatility3/framework/plugins/linux/mountinfo.py
gcmoreira: Yes, it could be a bit confusing. In all the cases the code walks the whole list of tasks. The difference is in the output.
The --pids argument is just a PID filter which is used to call the pslist's PID filter. If used, it only shows the information about these specific PIDs. Later, I found a `namespace filter` instead (or on top) of the pid filter could be also useful. Thoughts?

On the other hand, the default behaviour is to group the information by `mount namespace` and avoid showing PIDs. That's in essence what "mount" does showing only /proc/**self**/mount instead of each process mount namespace. Basically, this default behaviour is to avoid repeating rows, showing a less verbose output. 
In general many processes share a given namespace, so otherwise it will show a large number of rows with basically the same information, only changing the PID column values. Here's where **--all-processes** come into play. Probably not the best argument name but with the later, instead of showing the mount points per namespace, it shows the mount points per namespace and PID. I hope that was helpful

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Yeah, it would be better to double-check this change didn't break anything. However, `d_name` is not the path, is just the dentry name. This function mimics the kernel [__dentry_path() ](https://elixir.bootlin.com/linux/v5.8/source/fs/d_path.c#L325) and resolves the full dentry path.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: We can move the constants to the constants file. I thought maybe it could look more organised to have them encapsulated in their class given them even more context but not too fussy with it.

volatility3/framework/symbols/linux/__init__.py
gcmoreira: Ouch haha... just wait for the next pull request :). 
I really love the flexibility this feature gives to the framework. Being able to extend types allows to make the differences between kernel versions transparent to the plugins. That's fantastic and super powerful.
We can move these changes to the plugin, however I think moving this kind of changes from the core to the surface leads to have duplicate code in the future. I believe it is better these differences be managed by the framework and not by the plugins. Anyway, I understand you being reluctant to these sort of changes. If you still believe it's not appropriate just let me know or again feel free to move that code.

volatility3/framework/plugins/linux/mountinfo.py
gcmoreira: > The number of generator can produce different numbers of columns based on state (_show_mountpoints_per_namespace) rather than a parameter, and the columns are separately set by the same state. 

The `_show_mountpoints_per_namespace` is indeed the plugins parameters, It's just a simplification to avoid checking both arguments every single time.
`self._show_mountpoints_per_namespace = not any([self.config.get('pids'), self.config.get('all-processes')])`

The **--mount-format** completely changes the output view and make it looks like the userland _mount_ output or _/proc/pid/mounts_, which most Linux users are more familiar with. The _mountinfo_ output format (default), which is based on the _/proc/pid/mount_ output, has quite more details.
I made the layout of `run()` and the bottom of  `_generator()` as similar as possible to make it easier to modify in the future.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -468,6 +542,27 @@ def get_max_fds(self) -> interfaces.objects.ObjectInterface:
...
+    MNT_SHARED = 0x1000
+    MNT_UNBINDABLE = 0x2000
+
+    MNT_FLAGS = {
+        MNT_NOSUID:     "nosuid",
+        MNT_NODEV:      "nodev",
+        MNT_NOEXEC:     "noexec",
+        MNT_NOATIME:    "noatime",
+        MNT_NODIRATIME: "nodiratime",
+        MNT_RELATIME:   "relatime",
```
gcmoreira: Sorry, we need to remove the multiple spaces here.

volatility3/framework/plugins/linux/mountinfo.py
ikelos: Thanks, I guess I've never read continue statements that often, since I don't write them, so it's a bit jarring to come across them.  Line limits I'm more flexible on unless it goes to three lines or make the thing a mess to read.  I don't think I can easily change the pull request, so best if you do it please.  I'll try to be a little easier on continues where I can...  5;)

volatility3/framework/plugins/linux/mountinfo.py
ikelos: Sooooo... how does it know which pid to use if `pids` isn't set and `all-processes` isn't set?  Is that just the kernel namespace?  I think adding the namespace filter and clarifying when/how these filters act/what their defaults are, would be the best way out of this.  Something about `--all-processess` and the `--pids` flag doesn't sit right with me...  I don't think `--all-processes` is accurate enough for what it's doing.

volatility3/framework/plugins/linux/mountinfo.py
ikelos: If it's just for display then the plugin shouldn't be making that decision.  I know it's difficult and it feels like you're being unkind in the most common use of the plugin, but it's really not the plugin's job to decide what data it should and shouldn't return.

One of the main goals of volatility 3 was to allow it to be used as a library, and potentially allow the data to be processed automatically.  Given the UI might be a web UI, or a GUI, where you can toggle columns as you want, reorder them etc, it doesn't make sense for there to be a flag just to appease the CLI users.  I really don't want plugins trying to make decisions because of what they think will look nice for the CLI specifically.  It will limit people's thinking and set a bad precedent.

It *looks* as though removing the pid column is the only thing that changes, and in my opinion that's really not worthwhile.  For this to be used as a library, the plugin needs to return everything and then let whatever UI is in use make decisions about what's displayed and how.  We can try our best to make it easily human readable, but display stuff should really be dealt with by the UI rather than library or plugins.

volatility3/framework/symbols/linux/__init__.py
ikelos: Ok, that's fine.  I've seen the next pull request and I agree with you that it should probably live somewhere central, but that means it'll need review by our core linux guy, @atcuno, to make sure he's happy with it.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: That's ok, yapf would do it anyway.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Yep, that's fine, what I was saying about the constants file is that we need to bump the framework minor version (making it `(2, 1, 0)` since these are core changes that people might want to rely on in the future.  They need to be able to say their plugin will work with 2.1.0 but not 2.0.0, so just include a bump at https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/constants/__init__.py#L42 (`VERSION_MINOR`) please...

volatility3/framework/plugins/linux/mountinfo.py
gcmoreira: I ended up removing `--all-process` and adding a mount namespace filter `--mntns`. The PID column is only shown when a pid filter `--pids` is used. Otherwise, the output is per mount namespace, so the PID column doesn't make sense in the default behavior.

volatility3/framework/plugins/linux/mountinfo.py
gcmoreira: Fixed

volatility3/framework/symbols/linux/__init__.py
gcmoreira: Cool. I moved one more function here which made everything easier from the plugin side. That allowed me to remove all that dodgy stuff we were discussing above in the first comments. So, it feels like it was the place to be.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: oh ok. Does it mean I should also bump the MountInfo `_required_framework_version` version? After changing the VERSION_MINOR = 1 it still works even using _required_framework_version = (2, 0, 0)

volatility3/framework/plugins/linux/mountinfo.py
gcmoreira: Oh ok, I checked it. As per the [code](https://github.com/volatilityfoundation/volatility3/blob/3c11099e6621027777ac5a9eeb3f8fbdc5b39b4b/volatility3/framework/__init__.py#L50), the major only has to be the same. The minor module version requirement could be same or lower and the patch level is not checked at all.
hm, I think it would be good idea to use something like a standard python module [packaging.version](https://packaging.pypa.io/en/latest/version.html) for comparing versions instead of doing it manually.

Also, note that the exception message is not accurate. In the hypothetical case that the framework has a lower version than the module requirement, the message would be, for instance:
'Framework interface version 2 is an older revision than the required version 2.2'
instead of:
'Framework interface version 2.1 is an older revision than the required version 2.2'

```diff
--- /tmp/qRIePE___init__.py
+++ /home/gmoreira/vol3-latest/volatility3/framework/__init__.py
@@ -51,7 +51,7 @@
             if args[1] > interface_version()[1]:
                 raise RuntimeError(
                     "Framework interface version {} is an older revision than the required version {}".format(
-                        ".".join([str(x) for x in interface_version()[0:1]]), ".".join([str(x) for x in args[0:2]])))
+                        ".".join([str(x) for x in interface_version()[0:2]]), ".".join([str(x) for x in args[0:2]])))
 
 
 class NonInheritable(object):
```

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Oh ok, I checked it. As per the [code](https://github.com/volatilityfoundation/volatility3/blob/3c11099e6621027777ac5a9eeb3f8fbdc5b39b4b/volatility3/framework/__init__.py#L50), the major only has to be the same. The minor module version requirement could be same or lower and the patch level is not checked at all.

hm, I think it would be good idea to use something like a standard python module [packaging.version](https://packaging.pypa.io/en/latest/version.html) for comparing versions instead of doing it manually.

Also, note that the exception message is not accurate. In the hypothetical case that the framework has a lower version than the module requirement, the message would be, for instance:
'Framework interface version 2 is an older revision than the required version 2.2'
instead of:
'Framework interface version 2.1 is an older revision than the required version 2.2'

```diff
--- /tmp/qRIePE___init__.py
+++ /home/gmoreira/vol3-latest/volatility3/framework/__init__.py
@@ -51,7 +51,7 @@
             if args[1] > interface_version()[1]:
                 raise RuntimeError(
                     "Framework interface version {} is an older revision than the required version {}".format(
-                        ".".join([str(x) for x in interface_version()[0:1]]), ".".join([str(x) for x in args[0:2]])))
+                        ".".join([str(x) for x in interface_version()[0:2]]), ".".join([str(x) for x in args[0:2]])))
 
 
 class NonInheritable(object):
```

gcmoreira:
Attached the tests output, so you can have a better understanding of how it works.
[test_results.zip](https://github.com/volatilityfoundation/volatility3/files/7716541/test_results.zip)

In the `coverage` folder, there are two set of files. The first set `coverage_VERSION*.txt` is the mountinfo plugin output tested -using no arguments- on each kernel version mentioned on the main description above. BTW, I just added the 5.10.0 kernel test as well. The `coverage_mount_VERSION*.txt` is the plugin output using the mount format `--mount-format`.

On the other hand, in the `filters` folder there are two files, one for the output of testing the PID filtering `filter_pids_VERSION*.txt`, and another for the output of testing the mount namespace filter `filter_mntns_VERSION*.txt`

Every file has the used volatility3 command line on the top, so you will figure it out anyway.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: > Does it mean I should also bump the MountInfo `_required_framework_version` version? 

In the same commit I also bumped the module required framework version. Let me know if that's not necessary.

ikelos: Thanks, if `mountinfo` needs the features that you're adding (that wouldn't be present in `(2, 0, 0)`) then yes, you'd need to bump it.

As to using packaging.version, that would require a large external dependency for something we've already got code for.  It also has a lot more options than we need.  We also use semantic versioning with is mostly compatible with PEP 440, but again, why drag in a big dependency for something that's mostly compatible with what we already use.  I'm not against using a package for versioning, but I haven't heard a compelling reason to do so yet...

volatility3/framework/plugins/linux/mountinfo.py
ikelos: Still not certain I'm in favour of the plugin making the decision over what should be displayed.  I think I'd prefer that all the data was output and the UI (or program consuming the data) got to decide what to display, but if you really think it affect readers of the output far more than it would affect programs which may want consistent output, then I'm not gonna make it a showstopper...

ikelos:
Current state of this is still waiting on @atcuno for review.

ikelos:
Sorry, I realized I bumped the wrong value, the version is now (2, 1, 0), my bad!  5:S

gcmoreira:
@ikelos I fetched the latest upstream changes so that is ready to merge. There was a conflict with the constant VERSION_* as they have been modified since I made this pull request but it also needed to bump the VERSION_MINOR. 
Since the VERSION_MINOR was incremented, do I need to set the VERSION_PATCH to 0 ? it sounds like I should but not sure.

ikelos:
Yep, if you bump `VERSION_blah`, all lesser version numbers are set to 0 (so change `VERSION_MINOR` would reset `VERSION_PATCH` to 0).

gcmoreira:
> Yep, if you bump `VERSION_blah`, all lesser version numbers are set to 0 (so change `VERSION_MINOR` would reset `VERSION_PATCH` to 0).

Oh ok, yeah it was kind of obvious. Anyway, maybe a comment there clarifying that won't hurt

gcmoreira:
Ok done

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -0,0 +1,220 @@
...
+        superblock = mnt.get_mnt_sb()
+
+        mnt_id: int = mnt.mnt_id
+        parent_id: int = mnt.mnt_parent.mnt_id
+
+        st_dev = f"{superblock.major}:{superblock.minor}"
+
+        path_root = cls._do_get_path(mnt, task.fs.root)
+        if path_root is None:
+            return None
```
atcuno: would prefer this call to _do_get_path and the check of the return value to happen on line 91 so that the calls to path() and get_mnt_sb() can be avoided if the function is just going to bail out anyway. path() loops and can be slow

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -388,7 +425,44 @@ def name_as_str(self) -> str:
 class dentry(objects.StructType):
 
     def path(self) -> str:
-        return self.d_name.name_as_str()
+        """ Based on __dentry_path Linux kernel function"""
+        reversed_path = []
+        current_dentry = self
+        while not current_dentry.is_root():
```
atcuno: all loops like this need a 'seen' that tracks elements traversed already and breaks if duplicates are found. smear will definitely cause infinite loops eventually if not.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -388,7 +425,44 @@ def name_as_str(self) -> str:
...
+
+    def d_ancestor(self, ancestor_dentry):
+        """Search for an ancestor
+
+        Returns the ancestor dentry which is a child of "ancestor_dentry",
+        if "ancestor_dentry" is an ancestor of "child_dentry", else None.
+        """
+
+        current_dentry = self
+        while not current_dentry.is_root():
```
atcuno: all loops like this need a 'seen' that tracks elements traversed already and breaks if duplicates are found. smear will definitely cause infinite loops eventually if not.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -513,6 +608,66 @@ def get_mnt_parent(self):
...
+    def get_devname(self) -> str:
+        return utility.pointer_to_string(self.mnt_devname, count=255)
+
+    def has_parent(self) -> bool:
+        return self.vol.offset != self.mnt_parent
+
+    def get_dominating_id(self, root) -> int:
+        """Get ID of closest dominating peer group having a representative under the given root."""
+        current_mnt = self.mnt_master
+        while current_mnt and current_mnt.vol.offset != 0:
```
atcuno: all loops like this need a 'seen' that tracks elements traversed already and breaks if duplicates are found. smear will definitely cause infinite loops eventually if not.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -513,6 +608,66 @@ def get_mnt_parent(self):
...
+            peer = current_mnt.get_peer_under_root(self.mnt_ns, root)
+            if peer and peer.vol.offset != 0:
+                return peer.mnt_group_id
+
+            current_mnt = current_mnt.mnt_master
+        return 0
+
+    def get_peer_under_root(self, ns, root):
+        current = self
+        while True:
```
atcuno: same: re loops and seen

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -513,6 +608,66 @@ def get_mnt_parent(self):
...
+            if current.vol.offset == self.vol.offset:
+                break
+
+        return None
+
+    def is_path_reachable(self, current_dentry, root):
+        """Return true if path is reachable
+        """
+        current_mnt = self
+        while current_mnt.mnt.vol.offset != root.mnt and current_mnt.has_parent():
```
atcuno: same: re loops and seen

atcuno:
Besides re-arranging the order of two calls and then adding smear protection, this looks good. I commented all the places where those changes are needed.

volatility3/framework/plugins/linux/mountinfo.py
gcmoreira: done

volatility3/framework/symbols/linux/extensions/__init__.py

volatility3/framework/symbols/linux/extensions/__init__.py

volatility3/framework/symbols/linux/extensions/__init__.py

volatility3/framework/symbols/linux/extensions/__init__.py

volatility3/framework/symbols/linux/extensions/__init__.py

atcuno:
This is ready for merge @ikelos

volatility3/framework/constants/__init__.py
```diff
@@ -39,8 +39,8 @@
 
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 2  # Number of releases of the library with a breaking change
-VERSION_MINOR = 0  # Number of changes that only add to the interface
-VERSION_PATCH = 4  # Number of changes that do not change the interface
+VERSION_MINOR = 1  # Number of changes that only add to the interface
```
ikelos: The main tree is already up to 2.1.0, if this makes an additive change, please bump it to 2.2.0 (and update any plugins that rely on it).

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -0,0 +1,220 @@
...
+
+vollog = logging.getLogger(__name__)
+
+MountInfoData = namedtuple("MountInfoData", ("mnt_id", "parent_id", "st_dev", "mnt_root_path", "path_root",
+                                             "mnt_opts", "fields", "mnt_type", "devname", "sb_opts"))
+
+class MountInfo(plugins.PluginInterface):
+    """Lists mount points on processes mount namespaces"""
+
+    _required_framework_version = (2, 1, 0)
```
ikelos: This should get bumped when the above does.

volatility3/framework/constants/__init__.py
```diff
@@ -39,8 +39,8 @@
...
+VERSION_PATCH = 0  # Number of changes that do not change the interface
```
digitalisx: `2.1.0` version has already been achieved at this point, so it seems necessary to change it.

volatility3/framework/plugins/linux/mountinfo.py
digitalisx: It will depend on the version update mentioned above. You can leave it as it is if it is valid.

volatility3/framework/plugins/linux/mountinfo.py
ikelos: This is ok if you don't want to fix, it's entirely stylistic.  If it was just forgotten though then now's the time to fix it before I merge...  5;P

ikelos (commented):
Awesome, thanks for checking it @atcuno.  Just needs the version numbers tidying up and then I'll merge it...  5:)

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -0,0 +1,220 @@
...
+                fields.append(f"propagate_from:{dominating_id}")
+
+        if mnt.is_unbindable():
+            fields.append("unbindable")
+
+        mnt_type = superblock.get_type()
+
+        devname = mnt.get_devname()
+        if not devname:
+            devname = "none"
```
digitalisx, line 127: The `devname` variable is finally returned and appears to the user. If possible, I think we have a better direction than the `none` string. (like `UnreadableValue` or `UnparsableValue`)
```suggestion
        devname = mnt.get_devname()
        if not devname:
            devname = renderers.UnreadableValue()
```

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -0,0 +1,220 @@
...
+            requirements.BooleanRequirement(name="mount-format",
+                                            description="Shows a brief summary of the mount points information "
+                                            "with similar output format to the older /proc/[pid]/mounts or the "
+                                            "user-land command 'mount -l'.",
+                                            optional=True,
+                                            default=False),
+        ]
+
+    @classmethod
+    def _do_get_path(cls, mnt, fs_root) -> Union[None, str]:
```
digitalisx, line 52: It's very good. I think explicit typing (argument type, return type) of methods is not essential, but it will reduce the time of good code and contributors. If you have time, could you type the rest of the methods below for uniformity? This is pretty NIT, so you can just resolve it.

digitalisx (commented):
Hello @gcmoreira, Thank you for your wonderful work.
With the approval of experts and maintainers, this will soon be a great plugin.
Lastly, I added a comment about some NITs.
Please feel free to reflect!

volatility3/framework/constants/__init__.py
digitalisx: Oh... There was already a sincere review of the maintainer. I'll remove the duplicates.

digitalisx:
Oh... There was already a sincere review of the maintainer. If it overlaps, you can resolve it right away.

volatility3/framework/constants/__init__.py
gcmoreira: done

volatility3/framework/plugins/linux/mountinfo.py
gcmoreira: done

ikelos:
Thanks for making the changes  5:)

gcmoreira:
Thank you!

PR #596, opened 2021-12-10
Assorted fixes

gcmoreira:
* Undefined module imports
* Wrong comparison with None
* `not in` usage

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -651,7 +651,10 @@ class EPROCESS(generic.GenericIntelProcess, pool.ExecutiveObject):
         except AttributeError:
             return False
 
-        return not not value
+        if value:
+            return True
+
+        return False
 
     def get_vad_root(self):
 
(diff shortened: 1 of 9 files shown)


volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -651,7 +651,7 @@ def get_is_wow64(self):
         except AttributeError:
             return False
 
-        return value != 0 and value != None
+        return not not value
```
ikelos: This doesn't seem to make things clearer...  I agree it should be `is not None`, but I'd prefer `if value` to `if not not value`, if you really feels it needs condensing that much...

ikelos (commented):
Looks good, just one minor change (`not not`).  It may take me a while to get used to `x not in y` rather than `not x in y`, but I'll do what I can to stop it slipping in the codebase again.  As to the `!= None`, that should already be known, but not all developers followed it from the beginning, thanks for catching those!  5:)

I'm also really surprised that the missing modules didn't turn up in something like lgtm, but you're quite right, thanks for spotting/fixing them!  5:)  If you get that last bit changed, I'll happily merge it...  5:)

volatility3/framework/symbols/windows/extensions/__init__.py
gcmoreira: Done. Hope it works now

ikelos (approved):
Looks good, thanks for making the changes!

PR #597, opened 2021-12-10
Linux sockstats plugin

gcmoreira:
Added Sockstat linux plugin to enumerate every single process socket.

The output format is based on the `ss` tools. 

It supports:
* Unix socket
* Inet/Inet6 sockets (Based on Vol2 code).
* Netlink sockets
* VSock sockets (It requires vsock_sock type).
* Packet sockets. 
* XDP sockets (eBPF): eBPF name and ID
* Bluetooth sockets (It requires _bt_sock_ and _*_pinfo_ types)

It also shows cBPF/eBPF socket and reuseport filters infomation such as BPF program type, name and ID (if one of those filters are set in the socket).
If for some reason the types required are not present in the debugging symbols, it still shows basic information about the sockets such as: type, protocol and state.

Changes to the linux Lsof plugin were required to be able to reuse its filedescriptor listing capability.

Tested with the following linux kernel versions:
* [3.13.0-117.txt](https://github.com/volatilityfoundation/volatility3/files/7690649/sockstats_output.3.13.0-117.txt)
* [4.4.0-142.txt](https://github.com/volatilityfoundation/volatility3/files/7690650/sockstats_output.4.4.0-142.txt)
* [4.10.0-42.txt](https://github.com/volatilityfoundation/volatility3/files/7690651/sockstats_output.4.10.0-42.txt)
* [4.15.0-45.txt](https://github.com/volatilityfoundation/volatility3/files/7690652/sockstats_output.4.15.0-45.txt)
* [4.18.0-10.txt](https://github.com/volatilityfoundation/volatility3/files/7690656/sockstats_output.4.18.0-10.txt)
* [5.8.0-53.txt](https://github.com/volatilityfoundation/volatility3/files/7690657/sockstats_output.5.8.0-53.txt)
* [5.10.0-14.txt](https://github.com/volatilityfoundation/volatility3/files/7690658/sockstats_output.5.10.0-14.txt)


volatility3/framework/plugins/linux/lsof.py
```diff
@@ -21,6 +21,8 @@ class Lsof(plugins.PluginInterface):
 
     _required_framework_version = (2, 0, 0)
 
+    _version = (2, 0, 0)
```
ikelos: Technically, it only looks as though additive changes were made to the lsof API, meaning that any calls that could be make on the plugin previously can still be made on the plugin and work correctly.  As such, I think that can go up to (1, 1, 0) and all plugins that have an existing dependency will still work.

volatility3/framework/constants/linux/__init__.py
```diff
@@ -11,3 +11,219 @@
 # arch/x86/include/asm/page_types.h
 PAGE_SHIFT = 12
 """The value hard coded from the Linux Kernel (hence not extracted from the layer itself)"""
+
+# Standard well-defined IP protocols.
+# ref: include/uapi/linux/in.h
+IP_PROTOCOLS = {
```
ikelos, line 20: A lot of these look like general networking constants.  I'd leave them in here for now, but one day they might get moved to a constants/network.py file or something.  This is more musing to myself than anything else...  5:)

volatility3/framework/plugins/linux/lsof.py
```diff
@@ -34,26 +36,43 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+            fd_generator = linux.LinuxUtilities.files_descriptors_for_process(
+                context,
+                linuxutils_symbol_table,
+                task)
 
-    def run(self):
+            for fd_fields in fd_generator:
+                yield pid, task_comm, task, fd_fields
+
+    def _generator(self):
```
ikelos: I'd prefer the generator to be parametertized, as it had been.  It allows it to be reused if necessary and it just makes the inputs and outputs more obvious (the config is only really accessed within `run` and then everything flows through as a parameter).

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,374 @@
...
+from volatility3.framework.configuration import requirements
+from volatility3.framework.interfaces import plugins
+from volatility3.framework.objects import utility
+from volatility3.framework.symbols import linux
+from volatility3.plugins.linux import lsof
+
+
+vollog = logging.getLogger(__name__)
+
+class SockHandlers(object):
```
ikelos: Even though only the `__init__` and `process_sock` methods are public, it would still be best if it were versioned, so that plugins which make use of it can state that they need a particular version.  That can be done by implementing the ` interfaces.configuration.VersionableInterface` interface.  That means that should we need to make changes in the future, we won't go breaking everything...

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,374 @@
...
+            sock_reuseport_cb = sock.sk_reuseport_cb
+            self.__update_extra_socket_bpf(sock_reuseport_cb, extended)
+
+        saddr_tag = f"{dev_name}"
+        daddr_tag = ""
+        state = packet_sock.state
+        sock_stat = saddr_tag, daddr_tag, state
+        return packet_sock, sock_stat
+
+    def __update_extra_socket_bpf(self, sock_filter, extended):
```
ikelos: Double underscores have a special meaning in python.  This doesn't ever appear to be accessed dynamically, so please rename it to only use a single underscore.  We're not trying to match the linux platform exactly, but we *are* trying not to break python in weird ways...  5:)

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,374 @@
...
+        extended["bpf_filter_type"] = "cBPF"
+
+        if not sock_filter.has_member("prog"):
+            return
+
+        bpfprog = sock_filter.prog
+        if not bpfprog:
+            return
+
+        BPF_PROG_TYPE_UNSPEC = 0
```
ikelos: Seems a bit odd to stick this in a variable that looks like a constant?  I'd either move it to the constants file, or swap it out for the literal 0 and leave a comment in the code.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,374 @@
...
+            "AF_INET": self._inet_sock,
+            "AF_INET6": self._inet_sock,
+            "AF_NETLINK": self._netlink_sock,
+            "AF_VSOCK": self._vsock_sock,
+            "AF_PACKET": self._packet_sock,
+            "AF_XDP": self._xdp_sock,
+            "AF_BLUETOOTH": self._bluetooth_sock,
+        }
+
+    def _build_network_devices_map(self, netns_id):
```
ikelos: All of these would massively benefit from type annotations please?  So something like `netns_id: int` rather than just `netns_id`.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,374 @@
...
+        if bpfprog.type > BPF_PROG_TYPE_UNSPEC:
+            extended["bpf_filter_type"] = "eBPF"
+            bpfprog_aux = bpfprog.aux
+            if bpfprog_aux:
+                extended["bpf_filter_id"] = str(bpfprog_aux.id)
+                bpfprog_name = str(utility.array_to_string(bpfprog.aux.name))
+                if bpfprog_name:
+                    extended["bpf_filter_name"] = bpfprog_name
+
+    def _xdp_sock(self, sock, _extended):
```
ikelos: What's the point in the unnamed parameter `_extended`?  If it's just to match the other calls that's ok, but docstrings would go a long way to clarifying that...

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,374 @@
...
+        xdp_sock = sock.cast("xdp_sock")
+        device = xdp_sock.dev
+        if not device:
+            return
+
+        dev_name = utility.array_to_string(device.name)
+        saddr_tag = f"{dev_name}"
```
ikelos: This seems a little redundant?  It's not a show stopper, it's just making me scratch my head and wonder why it's there?

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,374 @@
...
+        state = bt_sock.state
+        sock_stat = saddr_tag, daddr_tag, state
+        return bt_sock, sock_stat
+
+class Sockstat(plugins.PluginInterface):
+    """Lists all network connections for all processes."""
+
+    _required_framework_version = (2, 0, 0)
+
+    _version = (2, 0, 0)
```
ikelos: This is a new plugin, the version should start at `(1, 0, 0)`?

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,374 @@
...
+            dentry = filp.get_dentry()
+            if not dentry:
+                continue
+
+            d_inode = dentry.d_inode
+            if not d_inode:
+                continue
+
+            socket_alloc = linux.LinuxUtilities.container_of(d_inode, "socket_alloc", "vfs_inode", vmlinux)
+            _socket = socket_alloc.socket
```
ikelos: Since `_socket` is used two lines down, this shouldn't start with an underscore.  Variables that are used in python should not start with an underscore, because python will treat them differently and it might make for weird bugs later on...

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,374 @@
...
+
+            child_sock = sock_fields[0]
+            protocol = child_sock.protocol if hasattr(child_sock, "protocol") else ""
+
+            net = task.nsproxy.net_ns
+            netns_id = net.proc_inum if net.has_member("proc_inum") else net.ns.inum
+            yield task, netns_id, fd_num, family, sock_type, protocol, sock_fields
+
+    def _generator(self):
+        pids = self.config.get('pids')
```
ikelos: Again, I'd prefer `self.config` access to be kept to `run` if possible.  Again, not a show stopper, just seems strange to create a function that isn't parameterized, but also isn't the top of the call chain...

There's a few more places further down, so if there are too many parameters, then it's ok, but ideally we'd try to avoid those (I don't know if there's other examples of generators calling the config, it's ok, but I'm trying to avoid it where possible, so not a rule, but guidelines where possible).  5:)

ikelos (commented):
Awesome, thanks very much for coding all this up, it looks really slick, good code style and so on!  I've got a couple of points dotted through out, and the additions at the end I'm really gonna need @atcuno to look over and make sure he's happy, but after the few tweaks, it looks on target for getting merged after @atcuno oks it...  5:)

volatility3/framework/plugins/linux/sockstat.py
gcmoreira: Sorry, yeah it's probably a bad habit. It's to avoid the IDE linter to raise a warning due to the unused variable.

volatility3/framework/plugins/linux/sockstat.py
gcmoreira: It makes sense

volatility3/framework/plugins/linux/sockstat.py
gcmoreira: Yeah, it's to make explicit it's a constant and avoid to hard code the 0, but at the same time it doesn't make sense to add it anywhere else as such because it's just used once in the code. Agree with changing it for a comment.

volatility3/framework/plugins/linux/sockstat.py
gcmoreira: Right. I thought array_to_string() returned an object other than str().
Is it the return type correct [here](https://github.com/volatilityfoundation/volatility3/blob/280c91212140ceb320ee59c647378e28f919592e/volatility3/framework/objects/utility.py#L12)? I think it was probably copied & pasted from _array_to_pointers_. Also, [pointer_to_string](https://github.com/volatilityfoundation/volatility3/blob/280c91212140ceb320ee59c647378e28f919592e/volatility3/framework/objects/utility.py#L23) needs a return type.

volatility3/framework/plugins/linux/sockstat.py
gcmoreira: Good catch! It was to avoid variable name collision with the socket module import which I was previously using, but then I moved that to extensions/\_\_init\_\_

gcmoreira:
All done. 
Also, fixed some minor stuff and changed the warning log lines to LOGLEVEL_V. Otherwise, those lines will be always shown in the output and that's not the intention.

volatility3/framework/plugins/linux/sockstat.py
ikelos: It returns a `String` volatility object (so the return type is correct), but volatility `Strings` should be inherited from `str` and therefore shouldn't need replacing.  I might be wrong, in which case this is no problem, but I believe it's unnecessary, and if so I'd prefer it be taken out please...  5:)

volatility3/framework/plugins/linux/sockstat.py
ikelos: If you still want to avoid name collisions, I'd give it a longer name, prepending it with _ has special meaning python and tells it that you're not going to use the value stored there...

ikelos:
Current state of this is still pending a review by @atcuno

volatility3/framework/plugins/linux/sockstat.py
gcmoreira: It was already changed same days ago. These changes are outdated.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -76,7 +99,51 @@ def process_sock(self, sock: objects.StructType):
 
         return sock, sock_stat, extended
 
-    def _unix_sock(self, sock: objects.StructType, extended: dict):
+    def _update_extended_socket_filters_info(self, sock: objects.Pointer, extended: dict) -> None:
```
ikelos: Hmmm, this seems to have pulled out the extended dict from the various `_sock` methods.  I'm ok with doing it that way, but I'd like to check with @atcuno as the main consumer of these methods that he's happy with that as a change of (internal) API...

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,505 @@
...
+        return sock, sock_stat, extended
+
+    def _update_extended_socket_filters_info(self, sock: objects.Pointer, extended: dict) -> None:
+        """Get infomation from the socket and reuseport filters
+
+        Args:
+            sock: The kernel sock (sk) struct
+            extended: Dictionary to store extended information
+        """
+        if sock.has_member("sk_filter") and sock.sk_filter:
```
atcuno, line 110: @ikelos sk_filter here on line 109 and sk_reuseport_cb on 114 are pointers. Are the current checks ok? "and sock.sk_reuseport_cb" or does that need to be a in try/catch looking for InvalidAdressException?

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,505 @@
...
+            sock_fields: A tuple with the *_sock object, the sock stats and the
+                         extended info dictionary
+        """
+        vmlinux = context.modules[symbol_table]
+
+        sfop_addr = vmlinux.object_from_symbol("socket_file_ops").vol.offset
+        dfop_addr = vmlinux.object_from_symbol("sockfs_dentry_operations").vol.offset
+
+        fd_generator = lsof.Lsof.list_fds(context, vmlinux.name, filter_func)
+        for _pid, _task_comm, task, fd_fields in fd_generator:
```
atcuno, line 422: @ikelos for '_pid' and '\_task_comm' that are not used, can we use the normal convention of just '\_' or do we need the prefixed name?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -539,3 +545,248 @@ def reference_count(self):
...
+    def get_inode(self):
+        if self.has_member("proc_inum"):
+            return self.proc_inum
+        elif self.ns.has_member("inum"):
+            return self.ns.inum
+        else:
+            raise AttributeError("Unable to find net_namespace inode")
+
+class sock(objects.StructType):
+    def __get_vol_kernel_module_name(self):
```
atcuno: why does this start with two underscores @gcmoreira?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -539,3 +545,248 @@ def reference_count(self):
...
+    @property
+    def type(self):
+        return SOCK_TYPES.get(self.sk_type, "")
+
+    @property
+    def inode(self):
+        if not self.sk_socket:
+            return 0
+
+        kernel_module_name = self.__get_vol_kernel_module_name()
```
atcuno: \_\_get\_vol\_kernel\_module\_name can raise ValueError but it is not caught here.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -539,3 +545,248 @@ def reference_count(self):
...
+            else:
+                saddr = sk_common.skc_rcv_saddr
+        elif family == socket.AF_INET6:
+            addr_size = 16
+            saddr = self.pinet6.saddr
+        else:
+            return
+
+        parent_layer = self._context.layers[self.vol.layer_name]
+        addr_bytes = parent_layer.read(saddr.vol.offset, addr_size)
```
atcuno: This needs a try/except for invalid addresses.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -539,3 +545,248 @@ def reference_count(self):
...
+            if hasattr(self.pinet6, "daddr"):
+                daddr = self.pinet6.daddr
+            else:
+                daddr = sk_common.skc_v6_daddr
+            addr_size = 16
+        else:
+            return
+
+        parent_layer = self._context.layers[self.vol.layer_name]
+        addr_bytes = parent_layer.read(daddr.vol.offset, addr_size)
```
atcuno: Also needs a try/except for invalid addresses.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: hm.. what should it be the expected action if an invalid address exception is catched? return None and the plugins handle it? Shouldn't the plugin take care of not calling it with an invalid address?

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: same question as above

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: fixed

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: That code was extracted from [LinuxUtilities::_get_new_sock_pipe_path](https://github.com/volatilityfoundation/volatility3/blob/9a0ba988abe8028a2a638bce4d9a8e43a11d7c4e/volatility3/framework/symbols/linux/__init__.py#L110) which raises that same exception [here](https://github.com/volatilityfoundation/volatility3/blob/9a0ba988abe8028a2a638bce4d9a8e43a11d7c4e/volatility3/framework/symbols/linux/__init__.py#L124). That function is only used by [LinuxUtilities::path_for_file](https://github.com/volatilityfoundation/volatility3/blob/9a0ba988abe8028a2a638bce4d9a8e43a11d7c4e/volatility3/framework/symbols/linux/__init__.py#L156) which doesn't care about the exception and neither its callers like [LinuxUtilities::files_descriptors_for_process](https://github.com/volatilityfoundation/volatility3/blob/9a0ba988abe8028a2a638bce4d9a8e43a11d7c4e/volatility3/framework/symbols/linux/__init__.py#L202) or [vm_area_struct::get_name](https://github.com/volatilityfoundation/volatility3/blob/9a0ba988abe8028a2a638bce4d9a8e43a11d7c4e/volatility3/framework/symbols/linux/extensions/__init__.py#L330). If so, we need to also consider those functions and their callers.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Actually, I realized when I wrote the comment [below](https://github.com/volatilityfoundation/volatility3/pull/597#discussion_r815225600) that it would probably better to move this function to somewhere it can be called by both `LinuxUtilities::_get_new_sock_pipe_path` in [these lines](https://github.com/volatilityfoundation/volatility3/blob/9a0ba988abe8028a2a638bce4d9a8e43a11d7c4e/volatility3/framework/symbols/linux/__init__.py#L115-LL124) and by `sock::inode` [here](https://github.com/volatilityfoundation/volatility3/blob/a7df7617dd68edd7e3bd3696b12a8bd6d3df3c51/volatility3/framework/symbols/linux/extensions/__init__.py#L595) . Thoughts?

volatility3/framework/plugins/linux/sockstat.py
atcuno: @ikelos bump on this comment

volatility3/framework/plugins/linux/sockstat.py
atcuno: @ikelos bump on this comment

volatility3/framework/symbols/linux/extensions/__init__.py
atcuno: I do think one place for it is best, but I am not sure where that would be. What you do think @ikelos ?

atcuno: in particular, all the extensions inherit from objects.StructType, so there is not a super class we can place it without changing them all.

volatility3/framework/symbols/linux/extensions/__init__.py
atcuno: ok, catch it here and we will update those other places in a different MR.

volatility3/framework/symbols/linux/extensions/__init__.py
atcuno: I am questioning the best way to handle this as its a bit more complicated than most other @property functions. The issue is that if a property function throws an exception, the the calling plugin will get a really strange error that the method does not exist (src_addr) in this case, when obviously it does. 

@ikelos given that this property function does more work than just retrieving a member, would this be better as a method like get_src_addr() that the plugins would call? Then it would be more normal to try/except the function call from the plugins.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,505 @@
+# This file is Copyright 2021 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+# Author: Gustavo Moreira
```
ikelos: We don't tend to include authors to avoid people getting territorial over changes to files they "own" because they originally wrote them.  We all tend to use git commits to show ownership (which is one of the reasons we use merge requests to keep the authorship information intact).  Would you be ok taking this out for consistency with the rest of the codebase please?

volatility3/framework/plugins/linux/sockstat.py
ikelos: Has member will return true as long as the structure is supposed to be present, I *think* you're ok with the pointer until you try to access any of its members, so this would do a test for 0 against the pointer...

volatility3/framework/plugins/linux/sockstat.py
ikelos: Err, you could use just `_`, but particularly when there are several being ignored, I find it better to just put the name behind the `_` so people referring to the code can tell what was hidden and what to expect if they need to access it.  I'd leave these as if you'd be ok with that?

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -266,4 +280,13 @@ def walk_internal_list(cls, vmlinux, struct_name, list_member, list_start):
         while list_start:
             list_struct = vmlinux.object(object_type = struct_name, offset = list_start.vol.offset)
             yield list_struct
-            list_start = getattr(list_struct, list_member)
\ No newline at end of file
+            list_start = getattr(list_struct, list_member)
+
+    @classmethod
+    def container_of(cls, addr, type_name, member_name, vmlinux):
```
ikelos: Typing information would be useful here, particularly that the return is `Optional` so people can tell that it may just return None rather than throwing an exception on failure.

ikelos: This feels as though maybe it should be a method on an object (something like `get_container` or similar?).  This could use a docstring too.  I think we do something similar for list structures (doing maths on the relative offset), if that's the case, it might be worth drawing this out as something more general.  @atcuno care to comment?

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Yeah, if there's a chance it may give a bogus result then it'll either need to return None (potentially a bit confusing and would require as many tests as wrapping the call in a try/except), or I think better would be @atcuno 's suggestion to go with `get_src_addr`.

ikelos (commented):
Just gave the code a brief skim and answered some outstanding comments, sorry for the delay!

volatility3/framework/symbols/linux/__init__.py
gcmoreira: Docstring and typing info added. 
Regarding the method name, I get your point but in this case I think it would be better to keep the method name as is. That's a quite common macro in the Linux kernel and I think it's more natural to read, write/translate or compare code using the same name. Anyway, no problem if you decide `get_container` is a better name.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Done. If it's unable to get the kernel module name, the inode property will return zero. Same as if it cannot the sk_socket. Let me know if you prefer to somehow flag it returning for instance -1 or log a warning message.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: I understand your point. if we convert the **src_addr** property to a method **get_src_addr** that will also require to change **dst_addr** and, for consistency, **src_port** and **dst_port**. And probably the others properties **family**, **protocol** and **state**. Is that all ok?

volatility3/framework/plugins/linux/sockstat.py
atcuno: sounds good to me. It was just a style question so the codebase would be consistent.

volatility3/framework/symbols/linux/extensions/__init__.py
atcuno: yes, I think any of them that have logic beyond determining which member exists in the symbol table should be converted.

volatility3/framework/symbols/linux/__init__.py
atcuno: container_of is the name of the corresponding macro from the Linux kernel so I would prefer to keep it this way

ikelos: Cool, that makes sense @atcuno.  Thanks @gcmoreira 5:)

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: We can slide one in between (so `class LinuxSocket(objects.StructType)`), in which case it would probably live under `symbols/linux/extensions` I guess?  It we just want generic code accessible by anyone, then probably the same place, but in a file that needs explicitly importing.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: done, properties changed to just getters

volatility3/framework/plugins/linux/sockstat.py
gcmoreira: @atcuno any comments on this?

gcmoreira:
Please check the changes in [9b0b254](https://github.com/volatilityfoundation/volatility3/pull/597/commits/9b0b2547d5a69b5ba78416fa3c858a41489e4dde). If an InvalidAddressException is raised reading the addresses, it returns a "?" value and a debug message is printed, but the plugin continues. Please let me know if you are fine with it or prefer to return a different value, raise the exception and abort the plugin or any other alternative

gcmoreira:
Also, in [5f70c70](https://github.com/volatilityfoundation/volatility3/pull/597/commits/5f70c7031c107645be93a71d392177c123c60307) I added more improvements in the way the socket family states are managed.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -754,7 +754,12 @@ def get_src_addr(self):
             return
 
         parent_layer = self._context.layers[self.vol.layer_name]
-        addr_bytes = parent_layer.read(saddr.vol.offset, addr_size)
+        try:
+            addr_bytes = parent_layer.read(saddr.vol.offset, addr_size)
+        except exceptions.InvalidAddressException:
+            vollog.debug(f"Unable to read socket src address from {saddr.vol.offset:#x}")
+            return "?"
```
digitalisx: The disadvantage of simply returning it as a question mark is that it is difficult for the user to track the cause of the value. Could you perhaps consider using the values this code? The example is `UnreadableValue`, but please use it based on the situation. Many plugins are using it.

https://github.com/volatilityfoundation/volatility3/blob/48c1533e894fe10ba151fc5ff1d8faa3b6d7f2b9/volatility3/framework/renderers/__init__.py#L21

digitalisx (commented):
@gcmoreira Thank you for your contribution, I left my thoughts in a short message. 🙂

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Hey @digitalisx, thanks for your suggestion, it looks good and programatically more elegant than just return a "?". Actually, I didn't know that existed. One of the many vol3 secrets ;)

I consider that for a "developer", particularly for tracking the cause of an error, the error/debug log message is a better option. With that, the developer have more detailed information and can search the exact line of code where it happens.

On the other hand, the normal "user" can't do much knowing the exact reason why one socket address is unavailable. I think that for a normal "user" the point is that there is a missing/unknown value and that's it. So, from UI perspective, I think the "?" (for the normal user) plus a detailed debug log message (for the developer) is not a bad idea either.

Last but not least, we may consider that it isn't something that we will see often. I've tested the plugin with several kernels versions from 3.13 to 5.10  (see above) and that code was never triggered.

Either way, to be considered, I leave here an example of how the output looks like in the hypothetical case it cannot read the address using your suggestion.

```
  |      NetNS |     Family |      Type |                  Proto |                                                                   Source Addr:Port |              Destination Addr:Port |       State |                                                                                                                                                                                   Tasks
* | 4026531992 |    AF_INET |     DGRAM |                    UDP |      <volatility3.framework.renderers.UnreadableValue object at 0x7f6d9c3e2630>:53 |                          0.0.0.0:0 | UNCONNECTED |                                                                                                                                                         (systemd-resolve,pid=674,fd=12)
* | 4026531992 |    AF_INET |    STREAM |                    TCP |      <volatility3.framework.renderers.UnreadableValue object at 0x7f6d9c1d74e0>:53 |                          0.0.0.0:0 |      LISTEN |                                                                                                                                                         (systemd-resolve,pid=674,fd=13)
...
```

digitalisx: @gcmoreira All right, I understand your intentions and concerns.
Attempting to provide good information in consideration of everyone is a very good thing and a difficult problem.
I suggested a good way to deal with those values on Volatility3, and it is up to the author and the maintainer to decide whether they are appropriate and reflected! 😊

However, the last example shown seems to be a little different when using the class presented above. In fact, the values that are finally rendered using class are as follows.

```python
yield (1, (
        format_hints.Hex(driver.vol.offset),
        "DEV",
        renderers.UnreadableValue(),
        device_name,
        renderers.NotApplicableValue(),
        device_type
    ))
```

```
renderers.UnreadableValue() -> "-"
renderers.UnparsableValue() -> "-"
renderers.NotApplicableValue() -> "N/A"

* 0x97065be786b0        DEV     -       00000096        N/A     FILE_DEVICE_UNKNOWN
```

I hope my comments were helpful and I am happy to continue to watch your good contribution! 🙌

gcmoreira: It's because in your example you are yielding the object directly. In this plugin the socket address and the port are concatenated. See [here](https://github.com/volatilityfoundation/volatility3/blob/db9f287cb146052d28ad75f1b9787b0131b46e8a/volatility3/framework/plugins/linux/sockstat.py#L174) and [here](https://github.com/volatilityfoundation/volatility3/blob/db9f287cb146052d28ad75f1b9787b0131b46e8a/volatility3/framework/plugins/linux/sockstat.py#L183)
The framework is checking the object type and replacing the output by a "-" or "N/A". See [here](https://github.com/volatilityfoundation/volatility3/blob/db9f287cb146052d28ad75f1b9787b0131b46e8a/volatility3/cli/text_renderer.py#L72)

ikelos: Just to weigh in on this, you're kind of both right.  In plugins we do definitely value using the `BaseAbsentValue`-derived objects, because it gives the UI more opportunity to understand what's going on and for plugin authors that might use the output to know what's going on, rather than just receiving a `?`.

However, those kinds of decisions should only be made just before display/rendering and not within a library function.  As such, the method should actually return None or throw an exception, and the caller will need to check that and decided what to do with it.  I know that it seems cumbersome and unfortunate, but it saves us from people writing plugins that end up checking for a return value of `?` to know whether a problem happened.  Inside of the volatility code we need as much information as possible.  Just before the point where it goes to the user is when we can start attempting to do interpretation.

Also, so that plugins can be used easily by another program, we strongly encourage plugins to output each piece of data into an individual column rather than concatenating it for human consumption.  Otherwise this makes the decision for the consumer of the data over exactly what they get and can do with it.  This is the same reason we prefer plugins to output `BaseAbsentValue`-derived objects.  It puts the most information into the output, and allows the display to be decided by the UI or some component higher in the stack.

I hope that makes sense?  This is all in an effort to make volatility a framework that can be used in many different instances, by many different user interfaces, rather than just a single CLI tool.

I'll go back and review the latest version in a bit...  5:)

volatility3/framework/constants/linux/__init__.py
```diff
@@ -14,3 +14,229 @@
 
 # include/linux/sched.h
 PF_KTHREAD = 0x00200000	 # I'm a kernel thread
+
+# Standard well-defined IP protocols.
+# ref: include/uapi/linux/in.h
+IP_PROTOCOLS = {
```
ikelos, line 20: These feel more generic, but I'm ok leaving them here for now.

volatility3/framework/plugins/linux/lsof.py
```diff
@@ -21,6 +21,8 @@ class Lsof(plugins.PluginInterface):
 
     _required_framework_version = (2, 0, 0)
 
+    _version = (1, 1, 0)
```
ikelos, line 24: This can start at (1, 0, 0), without an `_version` defined, I believe it defaults to (0, 0, 0) and is assumed not to contain any external methods that can be called.  It doesn't have to though, (1, 1, 0) is fine too.  5;)

volatility3/framework/plugins/linux/lsof.py
```diff
@@ -34,26 +36,45 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+                 symbol_table: str,
+                 filter_func: Callable[[int], bool] = lambda _: False):
+
+        linuxutils_symbol_table = None  # type: ignore
+        for task in pslist.PsList.list_tasks(context, symbol_table, filter_func):
+            if linuxutils_symbol_table is None:
                 if constants.BANG not in task.vol.type_name:
                     raise ValueError("Task is not part of a symbol table")
-                symbol_table = task.vol.type_name.split(constants.BANG)[0]
+                linuxutils_symbol_table = task.vol.type_name.split(constants.BANG)[0]
```
ikelos, line 59: Will there be another symbol table used?  Not sure why this is now `linuxutils_...`?  No problem with it, just wondering why.  5;)

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,506 @@
...
+vollog = logging.getLogger(__name__)
+
+class SockHandlers(interfaces.configuration.VersionableInterface):
+    """Handles several socket families extracting the sockets information."""
+
+    _required_framework_version = (2, 0, 0)
+
+    _version = (1, 0, 0)
+
+    def __init__(self, vmlinux, task):
```
ikelos, line 26: Typing information would be handy here, so we know whether `vmlinux` is supposed to be a module or a symbol table, or just a string.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,506 @@
...
+                            interface name
+        """
+        netdevices_map = {}
+        nethead = self._vmlinux.object_from_symbol(symbol_name="net_namespace_list")
+        net_symname = self._vmlinux.symbol_table_name + constants.BANG + "net"
+        for net in nethead.to_list(net_symname, "list"):
+            net_device_symname = self._vmlinux.symbol_table_name + constants.BANG + "net_device"
+            for net_dev in net.dev_base_head.to_list(net_device_symname, "dev_list"):
+                if net.get_inode() != netns_id:
+                    continue
```
ikelos, line 62: I'm not a huge fan of using continue, I find it makes following the flow more difficult.  It looks like this could easily be converted into `if net.get_inode() == netns_id:` and then the two statements and drop the need for the continue.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,506 @@
...
+            except exceptions.SymbolError as e:
+                # Cannot finds the *_sock type in the symbols
+                vollog.log(constants.LOGLEVEL_V, "Error processing socket family '%s': %s", family, e)
+        else:
+            vollog.log(constants.LOGLEVEL_V, "Unsupported family '%s'", family)
+
+        # Even if the sock family is not supported, or the required types
+        # are not present in the symbols, we can still show some general
+        # information about the socket that may be helpful.
+        saddr_tag = daddr_tag = "?"
```
ikelos: Hard coding "we don't know" values into functions is a bad idea.  This should either throw an exception or use an unambiguous value such as `None`, anything consuming the output of this will need to check for bad values.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,506 @@
...
+        if sock.has_member("sk_reuseport_cb") and sock.sk_reuseport_cb:
+            sock_reuseport_cb = sock.sk_reuseport_cb
+            extended["filter_type"] = "reuseport_filter"
+            self._extract_socket_filter_info(sock_reuseport_cb, extended)
+
+    def _extract_socket_filter_info(self, sock_filter: objects.Pointer, extended: dict) -> None:
+        extended["bpf_filter_type"] = "cBPF"
+
+        if not sock_filter.has_member("prog") or not sock_filter.prog:
+            return
```
ikelos, line 129: Similar to continue above, it looks like everything following could simply be under the inverse of this guard statement?  This is definitely a stylistic thing, so not a show stopper, but as I say I find it interrupts the flow when reading back the code.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,506 @@
...
+        unix_sock = sock.cast("unix_sock")
+        state = unix_sock.get_state()
+        saddr = unix_sock.get_name()
+        sinode = unix_sock.get_inode()
+        if unix_sock.peer != 0:
+            peer = unix_sock.peer.dereference().cast("unix_sock")
+            daddr = peer.get_name()
+            dinode = peer.get_inode()
+        else:
+            daddr = dinode = ""
```
ikelos: This feels like another one of those places where you may want to make it clear that you couldn't get valid data out?  I'd probably make `daddr_tag` to be `None` then ensure the caller checks the `sock_stat` value for errors before use.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,506 @@
...
+        sport = inet_sock.get_src_port()
+        daddr = inet_sock.get_dst_addr()
+        dport = inet_sock.get_dst_port()
+        state = inet_sock.get_state()
+
+        if inet_sock.get_family() == "AF_INET6":
+            saddr = f"[{saddr}]"
+
+        saddr_tag = f"{saddr}:{sport}"
+        daddr_tag = f"{daddr}:{dport}"
```
ikelos: Same as above, actually this one I think it's a little easier to understand with an unknown port number for example, still feels like something the code this deep shouldn't be deciding for everything above it (or at least document the behaviour in the docstring).

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,506 @@
...
+        saddr_list = []
+        src_portid = f"portid:{netlink_sock.portid}"
+        saddr_list.append(src_portid)
+        if netlink_sock.groups != 0:
+            groups_bitmap = netlink_sock.groups.dereference()
+            groups_str = f"groups:0x{groups_bitmap:08x}"
+            saddr_list.append(groups_str)
+
+        daddr_list = []
+        dst_portid = f"portid:{netlink_sock.dst_portid}"
```
ikelos: Prepending `portid` or `group` suggests that this dataa isn't standalone, but will be combined into a long string later on.  Since these may be processed by other programs rather than simply output for human consumption, we should think carefully about what it's returning and how it's returning that information.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,506 @@
...
+        dst_group = f"group:0x{netlink_sock.dst_group:08x}"
+        daddr_list.append(dst_group)
+        module = netlink_sock.module
+        if module and netlink_sock.module.name:
+            module_name_str = utility.array_to_string(netlink_sock.module.name)
+            module_name = f"lkm:{module_name_str}"
+            daddr_list.append(module_name)
+
+        saddr_tag = ",".join(saddr_list)
+        daddr_tag = ",".join(daddr_list)
```
ikelos: Really not so keen on this, maybe in the run/generate method, but not inside a helper function please, return a list, flatten it when the data is definitely not going to be used by anything else.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,506 @@
...
+            sock: Kernel generic `sock` object
+
+        Returns:
+            xdp_sock: The kernel's `xdp_sock` object
+            sock_stat: A tuple with the source, destination and state strings.
+        """
+        xdp_sock = sock.cast("xdp_sock")
+        device = xdp_sock.dev
+        if not device:
+            return
```
ikelos, line 278: As above please don't bail out of the code when it would be easy to invert the guard and indent so that it's clear what runs when.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,506 @@
...
+
+        saddr_tag = utility.array_to_string(device.name)
+
+        bpfprog = device.xdp_prog
+        if not bpfprog:
+            return
```
ikelos, line 285: Same as above.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -0,0 +1,506 @@
...
+        if bpfprog_aux:
+            bpfprog_id = bpfprog_aux.id
+            daddr_tag = f"ebpf_prog_id:{bpfprog_id}"
+            bpf_name = utility.array_to_string(bpfprog_aux.name)
+            if bpf_name:
+                daddr_tag += f",ebpf_prog_name:{bpf_name}"
+        else:
+            daddr_tag = ""
+
+        # Hallelujah, xdp_sock.state is an enum
```
ikelos: Hehehehe  5:P

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -183,6 +195,10 @@ def path_for_file(cls, context, task, filp) -> str:
     def files_descriptors_for_process(cls, context: interfaces.context.ContextInterface, symbol_table: str,
                                       task: interfaces.objects.ObjectInterface):
 
+        # task.files can be null
+        if not task.files:
+            return
```
ikelos, line 210: Same again

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -581,3 +587,262 @@ def reference_count(self):
...
+
+        kernel_module_name = module_names[0]
+        kernel = self._context.modules[kernel_module_name]
+        return kernel
+
+    def get_inode(self):
+        try:
+            kernel = self._get_vol_kernel()
+        except ValueError:
+            return 0
```
ikelos, line 871: This seems like you're making a decision that the caller might want to make?  Can an inode ever be 0, or is it always an error when it happens?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -581,3 +587,262 @@ def reference_count(self):
...
+        vfs_inode = socket_alloc.vfs_inode
+
+        return vfs_inode.i_ino
+
+    def get_state(self):
+        socket_state_idx = self.state
+        if 0 <= socket_state_idx < len(SOCKET_STATES):
+            return SOCKET_STATES[socket_state_idx]
+        else:
+            return "UNKNOWN"
```
ikelos: We do actually have examples of helper functions returning a `BaseAbsentValue`-derived class, so again, rather than hardcoding this string and making the decision for the caller, better to follow the examples such as https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/symbols/windows/extensions/pool.py#L262

ikelos (commented):
Thanks very much of this, most of it looks great (docstrings and so on).  It's a *really* substantial set of changes, so I'm not sure I'll have commented on every single instance of something that might need changing, but the two big things to come out of this are:

* Don't interpret the results before they're ready.  This is supposed to be a library that'll be used to goodness knows what.  At the moment, the main form of output is the CLI, but that is not its only use, and we don't want plugins or bits of core written as though it were.  As such, please try to avoid forming strings and almost papering over errors because humans don't want to read it.  For people trying to use this as a library, they'll want as much unfiltered information as they can get.  That means places where `?` or `UNKNOWN` are output should be changed, so that the caller can easily determine the difference between a literal value and an error.
* This one's stylistic, so I will led it slide if you feel strongly about it, but you've used it a lot.  You'll jump out of the middle of a function or some code rather than simply inverting the guard and letting the if statement prevent any further code from being run.  I find it very jarring to be ripped out of the middle of a method and try to reason about what has and hasn't happened when we return.  Where possible (and there's a lot of places where this would be possible) please could you let the code flow to the end with anything that needs to be behind a guard explicitly behind that guard?

Otherwise really good, really impressive, hopefully we'll get it in soon, but as I say, it's a lot of code going in all at once...  5:)

volatility3/framework/plugins/linux/sockstat.py
atcuno: yes, this is fine

gcmoreira:
Thanks for the review guys. I will finish and submit a new batch with the requested changes probably next week.

gcmoreira:
@ikelos @digitalisx 
Cool, here is the last batch of fixes and improvements.
Following your suggestions, I disaggregated pid, fd, address and port, and socket memory address to dedicated new columns. Previously, there were one record/line per kernel socket. This was because the kernel socket information is the same, and the task file descriptors using each socket were mentioned in a separate column. I found that awesome, summarised info and easy to spot all the 'users' of each socket. However, I also agree with you and It was hard to fit that info in just a field, and to be consumed by a library client, etc. So now that 'kernel socket -> tasks fds' association using the socket address has to be done by the end-user, or maybe can be performed by another plugin. 
Also, not every type of socket address has an `address` and a `port` to easily structure them in these two columns. That's why previously I decided to use a more generic `source` and `destination` addresses. I think that an unstructured data representation would be better in this case.
On the other hand, I replaced all the hardcoded strings. Instead, I used a BaseAbsentValue-derived class as you also suggested.

Find attached a detailed plugin output of my tests on each kernel version, including memory dumps for each type of socket (and its variants) supported by this plugin.
[output.log.gz](https://github.com/volatilityfoundation/volatility3/files/9898223/output.log.gz)

Last but not least, as the amount of source code in this PR is substantial, I also performed a code coverage of the same executions you find in `output.log` above, to make it sure that every (almost?) line of my code was executed in one way or another and nothing makes it crash. It looks pretty good to me. If you are curious, please have a look at the following html.

[coverage_html.zip](https://github.com/volatilityfoundation/volatility3/files/9898270/coverage_html.zip)

Hope you like it and we could merge it as soon as possible. Cheers

atcuno:
@ikelos the latest work from Gus looks good to me. Can you do a (final?) review on it? I don't see any comments left for me.

ikelos (commented):
Looks like my second set of comments around breaking code flow in the middle of functions are still present, but I'm ok accepting that as a stylistic issue.  Happy to merge it if you'd like it in @atcuno?

PR #600, opened 2021-12-17
Linux PSList and PSTree plugins improvements

gcmoreira:
**Linux PSList plugin**
The current Linux `PSList plugin` implementation only enumerates the thread group leaders tasks. The way it walks the task list is omitting the user threads, leading to a potential incomplete analysis.
In addition, there are seven other plugins using its functionality: _Bash_, _Check_creds_, _Elfs_, _Lsof_, _Malfind_, _Proc_ and _PsTree_, therefore, all of them could benefit from these changes.
To enable this feature, it's required to include the argument `--threads` in the plugin command line.
On the other hand, another argument `--decorate-comm` was added to be able to visually differentiate user-threads and kernel-threads from the other tasks in the output.
In the following example, we can see a thread group leader with PID & TID 1805 and two user threads with TIDS 1806 and 1807 which are
```shell
$ ./vol.py -r pretty -f ram_5.10.0-14.lime linux.pslist.PsList --threads --decorate-comm
  |  PID |  TID | PPID |              COMM
...
* | 1805 | 1805 | 1171 |    gvfsd-metadata
* | 1805 | 1806 | 1171 |           {gmain}
* | 1805 | 1807 | 1171 |           {gdbus}
* | 1831 | 1831 |    2 |     [kworker/0:4]
```

*Summary*
* A new column (TID) with the thread ID was added to differentiate between the thread group leader and the other tasks in the same group.
* Two new arguments were added:
    * **--threads**: Include user-space process threads to the output.
    * **--decorate-comm**: Decorate task (comm):
        * `user threads` in curly brackets
        * `kernel threads` in square brackets
* It is worth noting that these changes are transparent to the other plugins, except for PsTree, using the `pslist.PsList.list_tasks` class method. If any of them requires to include the user threads, they only need to add `include_threads=True` in the list_task function call.

**Linux PSTree plugin**
The main class of this plugin inherits from PSList, making it more sensitive to any change in the latter.
Because it also inherits the PSList plugin arguments, there was a bug where the plugin allowed to be called with a PID filter `--pid`, offering also this functionality in the PSTree plugin's help, but it never took that argument into consideration.

These changes allow to reuse any of the old and new PSList arguments: `--pid`, `--threads` and `--decorate-comm` and creates the task hierarchy tree including the user threads (if `--threads` is used). It follows the `pstree` idea of showing children processes and threads at the same level.

```shell
$ pstree -ntp
...
           |-colord(1597)-+-{gmain}(1599)
           |              `-{gdbus}(1601)
           |-systemd(1872)-+-(sd-pam)(1873)
```
```shell
$ ./vol.py -r pretty -f ram_5.10.0-14.lime linux.pstree.PsTree --threads --decorate-comm
          |   PID |   TID |  PPID |              COMM
...
**        |  1597 |  1597 |     1 |            colord
***       |  1597 |  1599 |     1 |           {gmain}
***       |  1597 |  1601 |     1 |           {gdbus}
```
*Summary*
* A new column (TID) with the thread ID was added to differentiate between the thread group leader and the other tasks in the same group.
* * Two new arguments were added:
    * **--threads**: Include user-space process threads to the output.
    * **--decorate-comm**: Decorate task (comm):
        * `user threads` in curly brackets
        * `kernel threads` in square brackets

**Tests**
The plugins were tested using the following Linux kernels and distros:

3.10.0-862 / CentOS 7
3.13.0-117 / Ubuntu 12.04
4.4.0-142 / Ubuntu 14.04
4.10.0-42 / Ubuntu 16.04
4.15.0-45 / Ubuntu 16.04
4.18.0-10 / Ubuntu 18.10
5.8.0-53 / Ubuntu 20.04
5.10.0-14 / Ubuntu 20.04

For a detailed output on each kernel version please check the following zip file.
[tests_output.zip](https://github.com/volatilityfoundation/volatility3/files/7732595/tests_output.zip)

**Merge note**
* For the sake of completeness, I included the changes in _volatility3/framework/symbols/linux/\_\_init\_\_.py_ , however, these changes were already submitted in pull request [SockStat plugin](https://github.com/volatilityfoundation/volatility3/pull/597).


volatility3/framework/plugins/linux/pslist.py
```diff
@@ -48,40 +57,118 @@ def filter_func(x):
         else:
             return lambda _: False
 
-    def _generator(self):
+    @staticmethod
+    def task_is_kernel_thread(task: interfaces.objects.ObjectInterface) -> bool:
```
ikelos: It almost feels like these should be methods on the task object, rather than in this plugin?

ikelos: Also, brief docstrings would be much appreciated...

volatility3/framework/plugins/linux/pslist.py
```diff
@@ -48,40 +57,118 @@ def filter_func(x):
...
+
+    def _get_task_fields(
+            self,
+            task: interfaces.objects.ObjectInterface,
+            decorate_comm: bool = False) -> Tuple[int, int, int, str]:
+        """Extract the fields needed for the final output
+
+        Args:
+            task: A task object from where to get the fields.
+            decorate_comm: If True, it decorates the comm string of
```
ikelos, line 68: Ideally this would tag them with data or another field (something like `thread_type`), and then allow the plugin/display mechanism to decide what to do about it.  I don't know that volatility is flexible enough to allow two fields to be combined into one during output, or whether we actually need a much more expressive mechanism per-plugin to decide how to display results.  I guess for now this is ok, but I'll keep its needs in mind if the situation arises again and see if there's a way we can keep data output as python types, and then display output as something the plugin can control...

volatility3/framework/plugins/linux/pstree.py
```diff
@@ -12,44 +11,74 @@ class PsTree(pslist.PsList):
...
+            else:
+                parent_pid = proc.tgid
 
-            child_list = self._children.get(ppid, set([]))
+            child_list = self._children.setdefault(parent_pid, set())
             child_list.add(proc.pid)
-            self._children[ppid] = child_list
-            proc = self._processes.get(ppid, None)
+
+            proc = self._tasks.get(parent_pid)
```
ikelos, line 36: Please provide a default (as we did previously) when calling get.  We can't be certain the `parent_pid` will definitely be in the list, so there should be a fallback that's handled.  If it's not this will throw a random `KeyError` to a user who won't know why...

`self._tasks`, unlike `self._children` has not has `setdefault` called on it, and if you *can* be guaranteed of the default value, then we don't need to use get we can just do index/key accesses.

ikelos (commented):
This, again, is one I'd like @atcuno to look over please.  Otherwise there's just a couple of points that could do with checking out...  5:)

volatility3/framework/plugins/linux/pstree.py
gcmoreira: > Please provide a default (as we did previously) when calling get. We can't be certain the `parent_pid` will definitely be in the list, so there should be a fallback that's handled. If it's not this will throw a random `KeyError` to a user who won't know why...

I understand, but that's exactly what get() does. Its default `default` is **None**, so `get(something, None)` looks redundant to me.
```python
>>> help({})
...
 |  get(self, key, default=None, /)
 |      Return the value for key if key is in the dictionary, else default.
```

On the other hand, `get()` doesn't throw _KeyError_ exceptions. That's actually where the default value comes to play with get().
```python
>>> {}.get("aaaa")
>>> 

>>> repr({}.get("aaaa"))
'None'

>>> repr({}.get("aaaa", 1))
'1'
```

volatility3/framework/plugins/linux/pslist.py
gcmoreira: Done

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -267,3 +267,12 @@ def walk_internal_list(cls, vmlinux, struct_name, list_member, list_start):
             list_struct = vmlinux.object(object_type = struct_name, offset = list_start.vol.offset)
             yield list_struct
             list_start = getattr(list_struct, list_member)
+
+    @classmethod
+    def container_of(cls, addr, type_name, member_name, vmlinux):
```
atcuno: this function seems useful, but I do not see it used anywhere in the rest of the commit. I just wanted to make sure this was meant to be part of this pull request.

atcuno:
Besides the container_of question, this looks good to me.

volatility3/framework/symbols/linux/__init__.py
gcmoreira: Good catch! It was used before the last commit changes #[192759b](https://github.com/volatilityfoundation/volatility3/pull/600/commits/192759b5d5c8c2622059847fbd915faf8af66e50). 
I just removed it as it is also included in the SockStats PR. I mentioned this in the "merge note" above.

PR #605, opened 2021-12-28
QEMU: Add 'dirty-bitmap' and 'pbs-state', handle page_size more consistently

cstation:
In rare cases, QEMU suspend-to-disk files contain sections named `dirty-bitmap` and `pbs-state`. These sections are defined at the beginning of the file, after the `ram`-section is defined. After the definitions of the sections (`QEVM_SECTION_START`s), the `QEVM_SECTION_PART`s belonging to the `ram`-section follow. This means that the sections do not occur once in total, but they occur in multiple parts accross the file. Therefore, the information about the sections needs to be stored and accessible throughout the parsing process.

Moreover, the actual offsets of the section parts are calculated in two ways, and the `page_size` variable is not used consistently. This PR fixes these issues.

### DIFF
--- a/volatility3/framework/layers/qemu.py
+++ b/volatility3/framework/layers/qemu.py
@@ -54,7 +54,7 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
 
     def _read_configuration(self, base_layer: interfaces.layers.DataLayerInterface, name: str) -> Any:
         """Reads the JSON configuration from the end of the file"""
-        chunk_size = 0x4096
+        chunk_size = 4096
         data = b''
         for i in range(base_layer.maximum_address, base_layer.minimum_address, -chunk_size):
             if i != base_layer.maximum_address:
@@ -65,6 +65,7 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
                     if start_of_json >= 0:
                         data = data[start_of_json:]
                         return json.loads(data)
+                    # No JSON configuration found at the end of the file, return empty dict
                     return dict()
         raise exceptions.LayerException(name, "Invalid JSON configuration at the end of the file")
 
@@ -79,9 +80,11 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
             addr = self.context.object(self._qemu_table_name + constants.BANG + 'unsigned long long',
                                        offset = index,
                                        layer_name = self._base_layer)
+            # Flags are stored in the n least significant bits, where n equals the bit-length of pagesize
             flags = addr & (page_size - 1)
-            page_size_bits = int(math.log(page_size, 2))
-            addr = (addr >> page_size_bits) << page_size_bits
+            # addr equals the highest multiple of pagesize <= offset
+            # (We assume that page_size is a power of 2)
+            addr = addr ^ (addr & (page_size - 1))
             index += 8
 
             if flags & self.SEGMENT_FLAG_MEM_SIZE:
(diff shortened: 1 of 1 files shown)


volatility3/framework/layers/qemu.py
```diff
@@ -217,9 +229,11 @@ def _decode_data(self, data: bytes, mapped_offset: int, offset: int, output_leng
         of the starting data.  It is the responsibility of the layer to turn the provided data chunk into the right
         portion of data necessary.
         """
-        start_offset = offset ^ (offset & 0xfff)
+        page_size = self._configuration.get('page_size', None) or 4096
```
ikelos: Is there ever a time this value could be 0?  I'm trying to figure out why (even in our original code), we set a default of None and then or it with 4096.  I'd probably just set the default of `4096`, but there's the very specific situation where `page_size` is 0 that this catches but a default of `4096` would not...

volatility3/framework/layers/qemu.py
```diff
@@ -79,9 +79,10 @@ def _get_ram_segments(self, index: int, page_size: int) -> Tuple[List[Tuple[int,
             addr = self.context.object(self._qemu_table_name + constants.BANG + 'unsigned long long',
                                        offset = index,
                                        layer_name = self._base_layer)
+            # Flags are stored in the n least significant bits, where n equals the bit-length of pagesize
             flags = addr & (page_size - 1)
-            page_size_bits = int(math.log(page_size, 2))
-            addr = (addr >> page_size_bits) << page_size_bits
+            # addr equals the highest multiple of pagesize <= offset
+            addr = addr ^ (addr & (page_size - 1))
```
ikelos, line 87: This makes an implicit assumption that the `page_size` will always be a power of 2.  I don't really have a problem with that, it's almost certainly always going to be true, but it might be nice just to include a comment that we're making that assumption.  Not a showstopper, just a nice to have...  5:)

ikelos (commented):
Awesome, thanks very much for catching those gotchas!  5:)  It all looks good, there's a couple of points I'm pondering but nothing that blocks the merge.  Let me know if you'd like to deal with them before I merge this, otherwise I'll merge it in a couple of days.  5:)

volatility3/framework/layers/qemu.py
cstation: I can't imagine a situation where `page_size` would be 0, since data is stored in chunks of `page_size` bytes. The only thing that could happen is that the value would be set to 0 due to errors while reading the configuration JSON, but that should (I guess) be handled in the function `_read_configuration`. Would you like me to change the `None` to `4096`?

volatility3/framework/layers/qemu.py
cstation: Didn't realize that :) I will add a comment!

volatility3/framework/layers/qemu.py
ikelos: Errr, yeah, I think it would be easier to read, but it occurs to me we could just force `_read_configuration` to populate the `page_size` entry whether there was one or not.  We could also deal with errors there too?  If you'd be happy doing that, then I think it might be the best way of dealing with it (if you'd prefer me to make the changes, just shout).  Thanks!  5:)

volatility3/framework/layers/qemu.py
ikelos: Perfect, thank you!  5:D

ikelos (approved):
Looks good, thanks very much!  5:D

volatility3/framework/layers/qemu.py
cstation: Issue is fixed now I think (after my force-pushed commit). An exception is raised if reading the configuration fails, and if no JSON is found at the end of the file (= no config is present), an emtpy dict is returned. I slightly prefer leaving the config as close to the original, rather than forcefully populating it. :relieved:

PR #606, opened 2021-12-28
Dump ELFs to file, more elaborate output of linux.PsList

cstation:
For Linux images, there hasn't been the possibility yet to dump processes or ELFs to a file.
This PR seeks to make this possible, in a way that is as consistent as possible with the behaviour and structure of Volatility 3's `windows.PsList` plugin and Volatility 2's `linux_procdump`.

The `--dump` flag on `linux.PsList` dumps the main executable of the process, whereas on `linux.Elfs`, it dumps all valid ELFs encountered in the memory space of the process. The MD5 hashes of the dumped main executables correspond to the MD5s of the results produced by Volatility 2.

Moreover, more details are printed at the `linux.PsList` output, in order to make it more than just a `ps aux`-output. This involves calculating the start time of a process, which relies on establishing the boottime from the kernel. I implemented the boottime only for kernels 3.19+. Older kernels could be supported as well, but would make the code probably a bit messier (probably not worth the pain, since kernels <3.19 hardly appear nowadays?).

### DIFF
--- a/volatility3/framework/plugins/linux/elfs.py
+++ b/volatility3/framework/plugins/linux/elfs.py
@@ -4,20 +4,28 @@
 """A module containing a collection of plugins that produce data typically
 found in Linux's /proc file system."""
 
-from typing import List
+import logging
+from typing import List, Type
 
-from volatility3.framework import renderers, interfaces
+from volatility3.framework import renderers, interfaces, constants
 from volatility3.framework.configuration import requirements
 from volatility3.framework.interfaces import plugins
 from volatility3.framework.objects import utility
 from volatility3.framework.renderers import format_hints
+from volatility3.framework.symbols import intermed
 from volatility3.plugins.linux import pslist
+
+from volatility3.framework.symbols.linux.extensions import elf
+
+vollog = logging.getLogger(__name__)
 
 
 class Elfs(plugins.PluginInterface):
     """Lists all memory mapped ELF files for all processes."""
 
     _required_framework_version = (2, 0, 0)
+
+    _version = (2, 0, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/linux/pslist.py
```diff
@@ -57,7 +78,37 @@ def _generator(self):
             if task.parent:
                 ppid = task.parent.pid
             name = utility.array_to_string(task.comm)
-            yield (0, (pid, ppid, name))
+            uid = task.real_cred.uid.val
+            gid = task.real_cred.gid.val
+
+            # task.(real_)start_time contains amount of nanoseconds since boottime
+            create_timespec = task.real_start_time if hasattr(task, "real_start_time") else task.start_time
+            create_timespec_sec = int(create_timespec / 1e9)
```
ikelos, line 86: This looks like a constant of some kind.  If it's used throughout the OS, then probably best to move it to constants.linux so that other plugin are consistent with this one...

volatility3/framework/plugins/linux/elfs.py
```diff
@@ -28,10 +36,94 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+                                            default = False,
+                                            optional = True)
         ]
 
+    @classmethod
+    def elf_dump(
+            cls, context: interfaces.context.ContextInterface,
+            layer_name: str,
+            elf_table_name: str,
+            vma: interfaces.objects.ObjectInterface,
```
ikelos, line 51: I generally prefer sending simple types in API function signatures, since volatility objects come with a lot of baggage so if we ever end up marshalling/unmarshalling calls to this function it might get sticky, but since we already pass `task` this way here (and in several places), it seems ok to allow it.  Just wanted to make sure it's more an exception than a rule...  5:)

volatility3/framework/plugins/linux/elfs.py
```diff
@@ -28,10 +36,94 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+            task: interfaces.objects.ObjectInterface,
+            open_method: Type[interfaces.plugins.FileHandlerInterface]) -> interfaces.plugins.FileHandlerInterface:
+        """Extracts an ELF as a FileHandlerInterface
+
+        Args:
+            context: the context to operate upon
```
ikelos, line 57: Awesome!  Really appreciate good docstrings, thanks!!!  5:D

ikelos (commented):
Awesome, thanks very much for this contribution!  It looks good to me, but unfortunately I'm not our resident linux guru, so I'm going to ask @atcuno to sign off on it before we merge it...

volatility3/framework/plugins/linux/elfs.py
cstation: Mmmm, you're right. I could replace the `vma` and `task`-variables with `vma.vm_start`, `path`, `task.comm` and `task.pid` if you think that's better :) The only (minor) disadvantage is that the logic for retrieving/calculating these variables has to be placed before every call to this function (as far as these variables aren't already present at those places), causing (very) little code duplication.

ikelos: No, as I say, @atcuno convinced me that it makes sense for certain methods to take a `task` object rather than listing each and every individual value we wanted to use from them.  I think best to leave it as is, I just wanted to make sure that my reasoning was documented somewhere so I (or someone else) could go "ah ha! but it says here..." at some point in the future...  5;P

cstation:
@atcuno: Could you please review this PR?

volatility3/framework/plugins/linux/pslist.py
```diff
@@ -57,7 +78,37 @@ def _generator(self):
             if task.parent:
                 ppid = task.parent.pid
             name = utility.array_to_string(task.comm)
-            yield (0, (pid, ppid, name))
+            uid = task.real_cred.uid.val
+            gid = task.real_cred.gid.val
```
atcuno, line 82: Two things:

1) Why is 'real_cred' being used here instead of 'cred'? Volatility 2 uses 'cred', but I am now questioning which one is correct based on some reading:

- https://lore.kernel.org/lkml/20080806153907.14351.98173.stgit@warthog.procyon.org.uk/
- https://www.kernel.org/doc/html/latest/security/credentials.html

I reached out to a colleague who is a Linux kernel expert to see what he thinks, but I am curious if you had a reason for this also.

2)

Accessing .val directly will not work across all versions as the format/type of uid and gid changed quite a bit over the years. You can see how Volatility 2 handled all the variations here:

https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/overlays/linux/linux.py#L1292

Regardless of if we use real_cred or cred, the handling of retrieving the value will need to match the flexibility of Volatility 2 in order to support all kernel versions needed.

volatility3/framework/plugins/linux/pslist.py
```diff
@@ -57,7 +78,37 @@ def _generator(self):
...
+                        vma = v
+                        break
+                if vma is not None:
+                    file_handle = elfs.Elfs.elf_dump(self.context, proc_layer_name, elf_table_name, vma, task,
+                                                     self.open)
+                file_output = "Error outputting file"
+                if file_handle:
+                    file_handle.close()
+                    file_output = str(file_handle.preferred_filename)
+
```
atcuno, line 110: The logic in lines 99-109 seems complicated, and I am fairly certain the plugin would backtrace if 'vma' is None going into line 103 as the file_handle variable would never be declared before 107 is hit.

It seems something like this would achieve the same goals (break after the right VMA is found, write out the file, report an error, etc.) with less complicated code flow.

```
file_output = "Error outputting file"

for v in task.mm.get_mmap_iter():
  if v.vm_start == task.mm.start_code:
       file_handle = elfs.Elfs.elf_dump(self.context, .....)
       if file_handle:
             file_output = str(file_handle.preferred_filename)
             file_handle.close()
             
       break
```

volatility3/framework/plugins/linux/pslist.py
```diff
@@ -49,6 +58,18 @@ def filter_func(x):
...
+        elf_table_name = intermed.IntermediateSymbolTable.create(self.context,
+                                                                 self.config_path,
+                                                                 "linux",
+                                                                 "elf",
+                                                                 class_types = elf.class_types)
+
+
+        vmlinux = self.context.modules[self.config['kernel']]
+        # These timekeeper variables are introduced in kernel 3.19
+        timekeeper = vmlinux.object_from_symbol(symbol_name = "tk_core").timekeeper
```
atcuno, line 70: This timekeeping code only supports a small range of kernels. For full support, all the variations would need to be added:

https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/linux/common.py#L253

It would also make sense to put this handling in its own function so its easier to maintain and _generator does not get huge. I can help with this if needed/preferred.

PR #613, opened 2022-01-05
QEMU: Take PCI-holes into account

### LINKED ISSUE #587: QEMU: Holes in physical address space not treated properly
**Describe the bug**

*Note*: If you are coming here because vol3 fails when analyzing a `virsh dump`, consider using `virsh dump --memory-only` instead.

I was working with QEMU dumps of various sizes: 0.5, 1, 2 and 4 GB. I converted them to "raw" files using the layerwriter plugin and then ran the pslist (and other) plugin against the raw files. I encountered this error, **but only with the 4GB dumps.**.
```
0x1002ff000 in layer memory_layer (Offset outside of the buffer boundaries)

        * The base memory file being incomplete (try re-acquiring if possible)

```
This also happened when running against the original qemu file. So I figured it has to be a problem with the qemu translation layer rather than layerwriter.

As you can see the phyiscal address in the warning is just above 4GB. A colleague of mine then did the following:
```
> strings 4gb-snapshot.raw | grep "BIOS-e820"
Oct 21 13:28:21 ubuntu kernel: [    0.000000] BIOS-e820: [mem 0x0000000000000000-0x000000000009fbff] usable
Oct 21 13:28:21 ubuntu kernel: [    0.000000] BIOS-e820: [mem 0x000000000009fc00-0x000000000009ffff] reserved
Oct 21 13:28:21 ubuntu kernel: [    0.000000] BIOS-e820: [mem 0x00000000000f0000-0x00000000000fffff] reserved
Oct 21 13:28:21 ubuntu kernel: [    0.000000] BIOS-e820: [mem 0x0000000000100000-0x00000000bffdffff] usable
Oct 21 13:28:21 ubuntu kernel: [    0.000000] BIOS-e820: [mem 0x00000000bffe0000-0x00000000bfffffff] reserved
Oct 21 13:28:21 ubuntu kernel: [    0.000000] BIOS-e820: [mem 0x00000000feffc000-0x00000000feffffff] reserved
Oct 21 13:28:21 ubuntu kernel: [    0.000000] BIOS-e820: [mem 0x00000000fffc0000-0x00000000ffffffff] reserved
Oct 21 13:28:21 ubuntu kernel: [    0.000000] BIOS-e820: [mem 0x0000000100000000-0x000000013fffffff] usable
``` 
Then: 
```
> cp 4gb-snapshot.raw test.raw
> dd bs=268435456 if=./4gb-snapshot.raw of=test.raw seek=16 skip=12
```
Effectively "moving up" the last GB of the raw file by one GB.
And boom, it worked:
```
vol -f test.raw linux.pslist
Volatility 3 Framework 1.0.1
Progress:  100.00 Stacking attempts finished                 
PID PPID COMM

1 0 systemd
2 0 kthreadd
3 2 rcu_gp
[...]
```
Now to my limited understanding, what happens is this:
The physical address space of the vm had some addresses at 3gig reserved for "something" (IOMem?) causing the 4th gig of memory to be mapped to the 4-5 GB range, and vol doesn't catch on to that. I don't know if its a shortcoming of vol or the qemu file format. 
On the smaller dumps this problem didn't show up because the entire physical address space fitted below 3gig.
When creating "ELF" dumps from the vms, the same "hole" is still there, but vol handles it correctly:
```
readelf -a ubuntu-18.04-server-cloudimg-amd64-4096-idle
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              CORE (Core file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x0
  Start of program headers:          64 (bytes into file)
  Start of section headers:          0 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         4
  Size of section headers:           0 (bytes)
  Number of section headers:         0
  Section header string table index: 0

There are no sections in this file.

There are no sections to group in this file.

Program Headers:
  Type           Offset                               VirtAddr                        PhysAddr
                      FileSiz                            MemSiz                        Flags  Align
  NOTE        0x0000000000000120  0x0000000000000000   0x0000000000000000
                    0x0000000000000330  0x0000000000000330   0x0
  LOAD        0x0000000000000450  0x0000000000000000   0x0000000000000000
                    0x00000000c0000000   0x00000000c0000000   0x0
  LOAD        0x00000000c0000450   0x00000000fffc0000      0x00000000fffc0000
                    0x0000000000040000   0x0000000000040000  0x0
  LOAD        0x00000000c0040450   0x0000000100000000  0x0000000100000000
                   0x0000000040000000   0x0000000040000000  0x0

There is no dynamic section in this file.

There are no relocations in this file.

The decoding of unwind sections for machine type Advanced Micro Devices X86-64 is not currently supported.

Dynamic symbol information is not available for displaying symbols.

No version information found in this file.
[...]
```  
**Context**
Operating System:  Linux
Python Version:  Python 3.9.2
Suspected Operating System:  Linux
Command:  vol -f /snapshot/qemu/or/raw some_plugin

**To Reproduce**

Good question, I don't know how to get a snapshot with the same holes in the physical address space. I think I could make a transfer happen if desired.

**Expected behavior**
Correctly translating the qemu file.

cstation:
Fixes #587

What [seemed like](https://github.com/juergh/lqs2mem.py/blob/master/lqs2mem.py#L218) like a simple additional offset to take into account, ended up in implementing two methods:

- 1 for determining the current target architecture
- 1 for determining where the PCI-hole is located for that architecture (+ RAM size).

The difficulty lies in the fact that the location of PCI-hole cannot be easily retrieved directly from a memory dump (or at least I was not able to find an elegant solution to do so). Linux does have quite a bunch of informing `dmesg`-messages available in memory, but for Windows, this is not the case.

This is a best-effort approach. The default offsets are known, but users could use the `max-ram-below-4g`-parameter, which influences the size and location of the PCI-hole, Once again, this cannot be easily retrieved directly from a memory dump.

The MicroVM and Xen architectures haven't been implemented yet. Luckily, PC i440FX and Q35 are by far the most used QEMU target architectures.

### DIFF
--- a/volatility3/framework/layers/qemu.py
+++ b/volatility3/framework/layers/qemu.py
@@ -5,6 +5,7 @@ import functools
 import json
 import logging
 import re
+import struct
 from typing import Optional, Dict, Any, Tuple, List, Set
 
 from volatility3.framework import interfaces, exceptions, constants
@@ -192,9 +193,8 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
         base_layer = self.context.layers[self._base_layer]
 
         while not done:
-            addr = self.context.object(self._qemu_table_name + constants.BANG + 'unsigned long long',
-                                       offset = index,
-                                       layer_name = self._base_layer)
+            # Use struct.unpack here for performance improvements
+            addr = struct.unpack('>Q', base_layer.read(index, 8))[0]
             # Flags are stored in the n least significant bits, where n equals the bit-length of pagesize
             flags = addr & (page_size - 1)
             # addr equals the highest multiple of pagesize <= offset
--- a/volatility3/framework/layers/qemu.py
+++ b/volatility3/framework/layers/qemu.py
@@ -3,13 +3,15 @@
 #
 import functools
 import json
-import math
+import logging
+import re
 from typing import Optional, Dict, Any, Tuple, List, Set
 
 from volatility3.framework import interfaces, exceptions, constants
-from volatility3.framework.layers import segmented
+from volatility3.framework.layers import segmented, scanners
 from volatility3.framework.symbols import intermed
 
+vollog = logging.getLogger(__name__)
 
 class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
     """A Qemu suspend-to-disk translation layer."""
@@ -40,6 +42,8 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
                  metadata: Optional[Dict[str, Any]] = None) -> None:
         self._qemu_table_name = intermed.IntermediateSymbolTable.create(context, config_path, 'generic', 'qemu')
         self._configuration = None
+        self._architecture = None
+        self._PCI_hole = None
         self._compressed: Set[int] = set()
         self._current_segment_name = b''
         super().__init__(context = context, config_path = config_path, name = name, metadata = metadata)
@@ -69,6 +73,117 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
                     return dict()
         raise exceptions.LayerException(name, "Invalid JSON configuration at the end of the file")
 
+    @staticmethod
+    def _get_pci_hole(architecture: Optional[Tuple[str, Optional[str]]], ram_size: int) -> \
+            Optional[Tuple[int, int]]:
+        """Returns the location and size of the PCI-hole for the given architecture and RAM size.
+
+        Args:
+            architecture: (Optionally) A tuple of strings containing the name (in lowercase) and (optionally) the
+                version of the used QEMU target architecture
+            ram_size: Size of the RAM in bytes
+
+        Returns:
+            (Optionally) A tuple containing the start and size of the PCI hole
+        """
+
+        # See https://qemu.readthedocs.io/en/latest/devel/memory.html for more info
+        #
+        # At least the following values could occur for devices using > 3-4 GB RAM:
+        # +--------------------------------+--------------------------------+------------+-------------+
+        # | Architecture                   | Reference Code                 | Hole Start | Hole End    |
+        # +--------------------------------+--------------------------------+------------+-------------+
+        # | PC i440FX + PIIX "New Default" | qemu/hw/i386/pc_piix.c:98      | 0xc0000000 | 0x100000000 |
+        # | PC i440FX + PIIX "Old Default" | qemu/hw/i386/pc_piix.c:98      | 0xe0000000 | 0x100000000 |
+        # | PC Q35 + ICH9                  | qemu/hw/i386/pc_q35.c:141      | 0x80000000 | 0x100000000 |
+        # | MicroVM                        | qemu/hw/i386/microvm.c:291     | 0xc0000000 | 0x100000000 |
+        # | Xen                            | qemu/hw/i386/xen/xen-hvm.c:248 | 0xf0000000 | 0x100000000 |
+        # +--------------------------------+--------------------------------+------------+-------------+
+        #
+        # MicroVM and Xen haven't been implemented yet
+        #
+        # For now, we assume that the parameter max-ram-below-4g is not set, since this parameter influences the size
+        # and location of the memory gap. Deviating hole sizes could eventually be detected for Linux by e.g. scanning
+        # for dmesg entries with a regex like rb'\[mem (0x[0-9a-f]{4,10})-0x[0-9a-f]{4,10}\] available for PCI devices'
+
+        end = 0x100000000
+
+        if architecture:
+            if architecture[0] == 'i440fx':
+                if architecture[1] and architecture[1] < '2.0':
+                    start = 0xe0000000
+                else:
+                    start = 0xc0000000
+            elif architecture[0] == 'q35':
+                if ram_size >= 0xb0000000:
+                    start = 0x80000000
+                else:
+                    vollog.log(constants.LOGLEVEL_VV, f"No PCI-hole present in this memory layout "
+                                                      f"(Q35 + RAM size < 0xb0000000)")
+                    return None
+            else:
+                vollog.warning(f"Default location of PCI-hole not known for target architecture {architecture[0]}")
+                return None
+
+            vollog.log(constants.LOGLEVEL_VV, f"PCI-hole for detected target architecture is usually located at "
+                                              f"0x{start:x}, using this location")
+            return start, end - start
+        else:
+            vollog.warning("Could not determine target architecture, the possible presence of a PCI-hole is not taken "
+                           "into account")
+            return None
+
+    def _determine_architecture(self, base_layer: interfaces.layers.DataLayerInterface) -> \
+            Optional[Tuple[str, Optional[str]]]:
+        """Tries to determine the QEMU target architecture being used
+
+        Args:
+            base_layer: Layer to operate on
+
+        Returns:
+            (Optionally) A tuple of strings containing the name (in lowercase) and (optionally) the version of the used
+                QEMU target architecture
+        """
+        # MicroVM and Xen haven't been implemented yet
+
+        # First try to find a string containing both architecture and version
+        architecture_pattern = rb'pc-(i440fx|q35)-([0-9]{1,2}.[0-9]{1,2}(?:.[0-9]{1,2})?)'
+        res = scanners.RegExScanner(architecture_pattern)
+        for offset in base_layer.scan(context = self.context, scanner = res):
+            line = base_layer.read(offset, 64)
+            regex_results = re.search(architecture_pattern, line)
+            architecture = regex_results.groups()[0].decode()
+            version = regex_results.groups()[1].decode()
+            vollog.log(constants.LOGLEVEL_VV, f"Detected QEMU target architecture to be {architecture}-{version}")
+            return architecture, version
+
+        # If that does not work, look in configuration JSON for devices specific to a certain architecture
+        architecture = None
+        for device in self._configuration.get('devices', []):
+            device_name = device.get('vmsd_name', '').lower()
+            if 'i440fx' in device_name or 'piix' in device_name:
+                architecture = 'i440fx'
+                break
+            elif 'ich9' in device_name:
+                architecture = 'q35'
+                break
+        if architecture:
+            vollog.log(constants.LOGLEVEL_VV, f"Detected QEMU target architecture to be {architecture}")
+            return architecture, None
+
+        # Still haven't found architecture, switch to fallback-method
+        architecture_pattern = rb'Standard PC \((i440FX|Q35)'
+        res = scanners.RegExScanner(architecture_pattern)
+        for offset in base_layer.scan(context = self.context, scanner = res):
+            line = base_layer.read(offset, 64)
+            regex_results = re.search(architecture_pattern, line)
+            architecture = regex_results.groups()[0].decode().lower()
+            vollog.log(constants.LOGLEVEL_VV, f"Detected QEMU target architecture to be {architecture}")
+            return architecture, None
+
+        vollog.warning("Could not determine QEMU target architecture!")
+        return None
+
     def _get_ram_segments(self, index: int, page_size: int) -> Tuple[List[Tuple[int, int, int, int]], int]:
         """Recovers the new index and any sections of memory from a ram section"""
         done = None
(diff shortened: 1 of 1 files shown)


ikelos:
Thanks very much, I'll take a review of this, but I have to say I'm still somewhat sceptical that we're "solving" a problem we don't understand.  I'd be much happier if I could clearly see in the qemu code why chunks of data, that have their own address, are written at known incorrect offsets.  The [page on memory](https://qemu.readthedocs.io/en/latest/devel/memory.html) only says the segments can overlap (and in their example himem and pci-hole occupy the same space), so that doesn't explain the shifting?

I'm happy to accept that something strange may be going on (and I'll still review the merge request), but I'm really uneasy committing such a complicated fix for a problem until I can clearly see what's causing it, or we have cast iron experiments to show that what we think is happening is actually what's happening...

volatility3/framework/layers/qemu.py
```diff
@@ -69,6 +73,117 @@ def _read_configuration(self, base_layer: interfaces.layers.DataLayerInterface,
...
+        #
+        # MicroVM and Xen haven't been implemented yet
+        #
+        # For now, we assume that the parameter max-ram-below-4g is not set, since this parameter influences the size
+        # and location of the memory gap. Deviating hole sizes could eventually be detected for Linux by e.g. scanning
+        # for dmesg entries with a regex like rb'\[mem (0x[0-9a-f]{4,10})-0x[0-9a-f]{4,10}\] available for PCI devices'
+
+        end = 0x100000000
+
+        if architecture:
```
ikelos, line 112: This feels like it's data, and there's not a great reason to have this stuff calculated, rather than in a big look-up table.  I think it would be easier to see what's going on just to have a `Dict[Tuple[...], Tuple[int, int]]`

volatility3/framework/layers/qemu.py
```diff
@@ -69,6 +73,117 @@ def _read_configuration(self, base_layer: interfaces.layers.DataLayerInterface,
...
+            base_layer: Layer to operate on
+
+        Returns:
+            (Optionally) A tuple of strings containing the name (in lowercase) and (optionally) the version of the used
+                QEMU target architecture
+        """
+        # MicroVM and Xen haven't been implemented yet
+
+        # First try to find a string containing both architecture and version
+        architecture_pattern = rb'pc-(i440fx|q35)-([0-9]{1,2}.[0-9]{1,2}(?:.[0-9]{1,2})?)'
```
ikelos, line 157: This information *has* to be in the qevm file (probably in the configuration), otherwise how would the memory be restored properly by qemu?

volatility3/framework/layers/qemu.py
```diff
@@ -87,15 +202,21 @@ def _get_ram_segments(self, index: int, page_size: int) -> Tuple[List[Tuple[int,
...
                 namelen = self._context.object(self._qemu_table_name + constants.BANG + 'unsigned char',
                                                offset = index,
                                                layer_name = self._base_layer)
                 while namelen != 0:
-                    # if base_layer.read(index + 1, namelen) == b'pc.ram':
-                    #     total_size = self._context.object(self._qemu_table_name + constants.BANG + 'unsigned long long',
-                    #                                       offset = index + 1 + namelen,
-                    #                                       layer_name = self._base_layer)
+                    name = base_layer.read(index + 1, namelen)
+                    total_size = self._context.object(self._qemu_table_name + constants.BANG + 'unsigned long long',
```
ikelos, line 221: It's a shame it needs this value, because constructing this object every time is unfortunately very slow.  I converted line 80 over to a struct statement to save a significant amount of time when parsing qemu files, so I'd avoid adding this in if possible (or use `struct`, I'd love to use volatility for all our parsing, but it's got quite a lot of overhead that isn't really necessary here).

ikelos (commented):
Technically, the code is good, the coding style matches ours perfectly, there's a couple of points including one for optimization, but I'm going to hold this one up until we've a clearer indication of exactly why this is happening and how qemu itself deals with this, sorry.  5:S

cstation:
Thanks! I must agree with you that this does not seem a very ideal fix, and that somewhere inside QEMU's code, there should be code for causing _and_ solving this (but where?). It also seems logical to me that this info (about the gaps) must reside in the qevm-file, but I haven't found it and their [analyze-migration](https://github.com/qemu/qemu/blob/7d4ae4d4978079d564d3b6354c90a949130409fe/scripts/analyze-migration.py)-script unfortunately doesn't provide additional clues. I'm happy with postponing the merge untill we've got a better clue what's going on.

cstation:
> Thanks very much, I'll take a review of this, but I have to say I'm still somewhat sceptical that we're "solving" a problem we don't understand. I'd be much happier if I could clearly see in the qemu code why chunks of data, that have their own address, are written at known incorrect offsets. The [page on memory](https://qemu.readthedocs.io/en/latest/devel/memory.html) only says the segments can overlap (and in their example himem and pci-hole occupy the same space), so that doesn't explain the shifting?

As far as I can see, the `himem` and `pci-hole` do _not_ occupy the same space, but `himem` is shifted up in favour of `pci-hole`: 

```
system_memory: container@0-2^48-1
 |
 +---- lomem: alias@0-0xdfffffff ---> #ram (0-0xdfffffff)
 |
 +---- himem: alias@0x100000000-0x11fffffff ---> #ram (0xe0000000-0xffffffff)
 |
 +---- vga-window: alias@0xa0000-0xbffff ---> #pci (0xa0000-0xbffff)
 |      (prio 1)
 |
 +---- pci-hole: alias@0xe0000000-0xffffffff ---> #pci (0xe0000000-0xffffffff)
```
So, `pci-hole` 'interrupts' the ram-segments, and _maybe_ only the mapped RAM gets migrated in QEVM-files?
```
+----------+-------------+-------------+------------------+----------------+
| Segment  |       Start |         End | Mapped RAM Start | Mapped RAM End |
+----------+-------------+-------------+------------------+----------------+
| lomem    |         0x0 |  0xdfffffff |              0x0 |     0xdfffffff |
| pci-hole |  0xe0000000 |  0xffffffff |                - |              - |
| himem    | 0x100000000 | 0x11fffffff |       0xe0000000 |     0xffffffff |
+----------+-------------+-------------+------------------+----------------+
```

volatility3/framework/layers/qemu.py
cstation: Wow, converting these to a `struct.unpack('>Q', ...)` indeed saves a massive amount of time! I haven't seen this change on line 80 been commited yet . Is that on purpose and do you still need to commit that? The same change could perhaps be applied to more places in this file, speeding it up even more.

ikelos:
Hmmm, I was looking at the secondary bits:
```
+---- himem: #ram (0xe0000000-0xffffffff)
+---- pci-hole: #pci (0xe0000000-0xffffffff)
```
But perhaps that does explain it better, so they are written to the same location but *don't* overlap.  That's kinda crazy?  Why would you write in the *wrong* address when you're having to explicitly list the address?

It seems we need to dig into what an `alias` is in the code:
```
if the subregion is an alias, the search is continued at the alias target (after the address is adjusted by the subregion offset and alias offset)
```

Ok, I'm becoming more convinced they've done something that on the surface appears crazy, but I'm that hoping they had good reason for.  I'll keep digging and see what I can come up with...

volatility3/framework/layers/qemu.py
ikelos: It's part of the patch I was rolling for #587 so hasn't been committed yet.  I was trying not to put it in too many places, because otherwise people question what benefit they get from volatility parsing the data.  I need to look at ways of making object creation more efficient, but for now I only wanted to do as few changes as possible.  I think there's two object creations that could be replace with `struct` in the big loop.  The others will get call a couple times and make next to no difference.

ikelos: If you want to get it in sooner, I'm happy to review just that as a PR and we can probably merge that pretty quickly.  Since we're in code-freeze though it's unlikely to make it into the next release tarball.  I'm hoping the one after this will only be 6 months away rather than a year, but just so you know the way we expect things to go...

volatility3/framework/layers/qemu.py
```diff
@@ -69,6 +73,117 @@ def _read_configuration(self, base_layer: interfaces.layers.DataLayerInterface,
...
+        end = 0x100000000
+
+        if architecture:
+            if architecture[0] == 'i440fx':
+                if architecture[1] and architecture[1] < '2.0':
+                    start = 0xe0000000
+                else:
+                    start = 0xc0000000
+            elif architecture[0] == 'q35':
+                if ram_size >= 0xb0000000:
```
ikelos, line 125: There should probably be a similar check for i440fx presumably?

ikelos (commented):
I also discovered that `env.mtrr_var[0]` from the state JSON (which I'm not sure we currently retrieve) may point to the start value, but I'm not sure if that's coincidence, and I only have the i440fx images you provided earlier to check.

ikelos:
Ok, I've filed this upstream as https://gitlab.com/qemu-project/qemu/-/issues/940, so hopefully they can shed some light on how we're supposed to know where the hole is.

volatility3/framework/layers/qemu.py
cstation: You're right! I fixed it in https://github.com/volatilityfoundation/volatility3/pull/613/commits/85139439a360931be44d772d21763c04b4e3ac94.

volatility3/framework/layers/qemu.py
cstation: > It's part of the patch I was rolling for #587 so hasn't been committed yet. I was trying not to put it in too many places, because otherwise people question what benefit they get from volatility parsing the data. I need to look at ways of making object creation more efficient, but for now I only wanted to do as few changes as possible. I think there's two object creations that could be replace with `struct` in the big loop. The others will get call a couple times and make next to no difference.

I've added the improvement to line 80 (now line 202-203). What was the other location you intended to change?

cstation:
> I also discovered that `env.mtrr_var[0]` from the state JSON (which I'm not sure we currently retrieve) may point to the start value, but I'm not sure if that's coincidence, and I only have the i440fx images you provided earlier to check.

I just checked this for a Q35 image, the value of `env.mtrr_var[0]` is there `0xc0000000` as well. So I'm afraid it's coincidence. :disappointed:

volatility3/framework/layers/qemu.py
ikelos: Hmmm, probably in `_load_segments` I'd check `section_byte` since it's used for every loop.  The easiest way to check is to run `python -m cProfile -o out.prof vol.py <some vol command involving qemu>`.  That'll create an out.prof, then you can use something like [Snakeviz](https://jiffyclub.github.io/snakeviz/) to see where the time's being spent.  That'll help you look around for what's using the most time.  In this instance, it boils down to calls to `object` in the `_get_ram_segments` method rather than in the `section_byte` call, but it could be any of the calls within that method.  After converting the `addr` call, you can see how much of an effect it had, and then see whether it's worth doing to the two `namelen` object calls as well...

image

cstation: Thanks for the explanation @ikelos! I tried to use `struct.unpack` at different other locations, but the performance impact is quite low (<0.5 second faster when applied to multiple locations). Compared to the performance gain that is achieved by parsing the address using `struct.unpack`, I think we should leave the other locations as they are.

cstation:
> Ok, I've filed this upstream as https://gitlab.com/qemu-project/qemu/-/issues/940, so hopefully they can shed some light on how we're supposed to know where the hole is.

I just saw the reply that someone posted on this issue. Seems there are no good prospects for finding better ways to determine the size and location of the gaps. 😢

ikelos:
You're right, I'm quite surprised but it seems despite having a large JSON blob of data, they don't store the holes and we'll need to keep a table of which architecture has which holes where (which this pull request contains).  I've asked if there's a particular identifier to find the architecture we're working with, which I'd hope there was, but it's possible there isn't.  Once we get an answer back on that we can either use their identifier, or if there isn't one then this pull request might be fine as is.  I'll give it another once over in either case, but it is as least progress of a form.  5:)

ikelos:
We should add comments to the PR documenting the bug and the response to indicate that this is the best way of doing things, so I don't forget and question it in the future...  5;)

ikelos:
Sorry for the spam, marked the wrong PR as draft.  5:S

ikelos:
Superseded by #698 which has now been merged!  5:D

PR #614, opened 2022-01-06
Create MFTScanner plugin

### LINKED ISSUE #312: MFTparser
The need of MFT parsing. used to have that in volatility 2.

Add MFT plugin from volatility 2

Filescan + dump + external MFT parser.
But right now every dump I tested seems to return an empty filescan result: #278 

**Additional information**
I mainly used it with the timeliner plugin in order to create a more extensive timeline.
Maybe merge them together?

### LINKED ISSUE #573: Request for mft plugin
- windows.hashdump 
- windows.cachedump
- windows.lsadump
- accessing mft data (can't figure out any way to do it in v3)

I have noticed that in infosec most of the people still tend to use volatility 2 over 3, some compatibility of functionalities would be nice for anyone who wants to make a transition.

kevthehermit:
First pass at an MFTScanner plugin 

This plugin uses the existing Yara extension to search for MFT record Headers. 
For each header found, read 1024 bytes from the layer (the length of an MFT record) and try to parse it.


volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+#
+
+import logging
+
+from struct import unpack
+from typing import Dict
+
+from volatility3.framework import constants, renderers, interfaces
+from volatility3.framework.configuration import requirements
+from volatility3.framework.exceptions import PagedInvalidAddressException
```
ikelos: I'd prefer to import `exceptions` rather than `PagedInvalidAddressException` to avoid people accidentally deciding to import it from here instead of where it lives.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+from volatility3.framework.objects import utility
+from volatility3.framework.renderers import conversion, format_hints
+from volatility3.plugins import yarascan
+
+vollog = logging.getLogger(__name__)
+
+try:
```
ikelos: Since we import yarascan, this check for yara *should* be done for us.  I'm ok having this code duplicated if we also need to pull in the yara module, but it appears we only use it for yara.compile, which seems to be something exposed by the yarascan plugin in `process_yara_options`.  If that method isn't suitable, I'd prefer to extend it/fix it rather than pulling in the compilation to this plugin and requiring each plugin that uses yara to duplicate this check...

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+    0x1:"Win32", # Case insensitive, allows most Unicide except specials ('/', '\', ';', '>', '<', '?')
+    0x2:"DOS",   # Case insensitive, upper case, no special chars, name is 8 or fewer chars in name and 3 or less extension
+    0x3:"Win32 & DOS", # Used when original name fits in DOS namespace and 2 names are not needed
+}
+
+MFT_FLAGS = {
+    0x0: "Removed",
+    0x1: "File", # "In Use",
+    0x2: "Directory", # if flag & 0x0002 == 0 this is a regular file
+    0x3: "Directory"
```
ikelos: This is a little confusing, if this is supposed to be for bit masking?  I can kinda understand why thrown it in as an additional flag, but I think I'd sooner just define file and directory, and then just check them manually...

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+}
+
+INDEX_ENTRY_FLAGS = {
```
ikelos: This one doesn't have a 0 defined, so again, there's inconsistency between how we deal with these two sets of flags.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
+# This file is Copyright 2022 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import logging
+
+from struct import unpack
```
ikelos: Please import struct, for the same reasons as below for the exceptions module.  We try only to import modules so that people can easily see what's part of this module's namespace, what isn't and if it isn't what namespace it does belong to...

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+
+        Returns:
+            Dict: a Dictionary that contains the Parse MFT Record
+        """
+        # https://github.com/Invoke-IR/ForensicPosters
+
+        flags = self.unpack_data(mft_record, 22, 'unsigned short')
+        file_type = MFT_FLAGS.get(flags, 'Unknown')
+
+        mft_entry = {
```
ikelos: This is practically a volatility structure, why not define it as one instead of lots of manual `struct.unpack` calls?  We prefer using volatility as the parser wherever possible, and only in instances where it makes a substantial difference to efficiency do we go with using struct.  It would also make it more reusable as other plugins could just load the symbol and use it itself...

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+            "SI": {},
+            "FN": []
+        }
+        }
+
+        attr_offset = mft_entry['FirstAttrOffset']
+        # Check at most for 6 entries
+        for i in range(6):
+            # If we attempt to overread the entry continue out
+            if attr_offset > 1000:
```
ikelos: Should this be `0x1000`?  Hard coding values (even safety limits) like this isn't ideal.  If everything is supposed to fit in a single page of memory, then it might be worth asking the layer how big each page is (once, at the start, for efficiency) and then using that...

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+                continue
```
ikelos: Generally not a fan of continues, they feel like they break flow for me, I'd sooner just invert the if and indent everything.  Not a show stopper though, since it's just stylistic...  5:)

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+
+            # attr_header
+            attr_type = self.unpack_data(mft_record, attr_offset, 'int')
```
ikelos: If this is used repeatedly, then again I'd have it as a type that can be referenced again and again.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+            attr_len = self.unpack_data(mft_record, attr_offset+4, 'int')
+
+            # As we look for strucutres of header + 1K we can not unpack non resident structures
+            nr_flag = self.unpack_data(mft_record, attr_offset+8, 'unsigned char')
+
+            # Skip headers
+            attr_data = attr_offset+24 # Len of Common and Resident Headers
```
ikelos: Lack of spacing suggests yapf hasn't been run on this.  Also not a showstopper, but we do provide a yapf style file to ensure the code looks the same across the whole codebase.  It'll get caught up in a sweep at some point if you don't want to change it yourself...  5:)

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+
+            if attr_type in ATTRIBUTE_TYPE_ID:
+                vollog.debug(f'Found Attribute {ATTRIBUTE_TYPE_ID[attr_type]}')
+                
+                if ATTRIBUTE_TYPE_ID[attr_type] == 'STANDARD_INFORMATION':
+                    creation_time_win = self.unpack_data(mft_record, attr_data,  'unsigned long long')
```
ikelos: This also feels like it would be better as a reusable type that returns an object, rather than manually populated a structure in the code.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+
+                    flags = self.unpack_data(mft_record, attr_data+56, 'unsigned short')
+                    permissions = VERBOSE_STANDARD_INFO_FLAGS.get(flags, 'Unknown')
+
+                    mft_entry['attributes']['FN'].append(
+                        {
+                            "creation_time": self.human_date(creation_time_win),
+                            "modified_time": self.human_date(modified_time_win),
+                            "updated_time": self.human_date(altered_time_win),
+                            "accessed_time":  self.human_date(access_time_win),
```
ikelos: This should be kept/turned into a python datetime, the output renderers can deal with them and the interpretation and display of dates should be done by the UI (in case of timezone shifts, or post processing etc) rather than per-plugin.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+
+class MFTScan(interfaces.plugins.PluginInterface):
+    """Scans for MFT FILE objects present in a particular windows memory image."""
+
+    _required_framework_version = (2, 0, 0)
+
+    @classmethod
+    def get_requirements(cls):
+        return [
+            requirements.TranslationLayerRequirement(name = 'primary',
```
ikelos, line 28: Does this only need the layer, and not the symbol table of the kernel?  That's fine if so, but most windows specific plugins rely on a `ModuleRequirement` to ensure windows was found.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+    def run(self):
+        return renderers.TreeGrid([
+                ('Offset', format_hints.Hex),
+                ('Record Type', str),
+                ('Record Number', int),
+                ('Link Count', int),
+                ('Permissions', str),
+                ('Attribute Type', str),
+                ('Filename', str),
+                ('Created', str),
```
ikelos: These should be datetimes.  It would also be massively handy if the plugin could implement the `TimeLinerInterface`, so that timeliner will pick up and use the output automatically...

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+                    # As this is variable and may or may not exist
+                    # And could have 0-6 entries lets do it per row.  
+                    yield 0, (
+                        format_hints.Hex(offset),
+                        signature,
+                        record_number,
+                        link_count,
+                        permissions,
+                        'FileName',
+                        entry.get('file_name', ''),
```
ikelos: If these aren't available they should be an `AbsentValue` of some kind, either `NotApplicable` or similar...

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,331 @@
...
+                    'N/A',
+                    si_creation_time,
+                    si_modified_time,
+                    si_updated_time,
+                    si_accessed_time)
+
+                for entry in fn:
+                    # As this is variable and may or may not exist
+                    # And could have 0-6 entries lets do it per row.  
+                    yield 0, (
```
ikelos: Since this feels related to the row above, it might be worth using `yield 1` rather than `yield 0`, to make it a subrow.

ikelos (commented):
Looks very cool, thanks!  5:D

A couple of stylistic points, but mainly it feels as though a lot of it should be converted into an ISF file, rather than repeated struct calls with hardcoded offsets.  It looks like it should lend itself to that fairly easily, it would just be writing the JSON and storing it in the right place (either the same place as the plugin, or in the `framework/symbols/windows/extensions` directory, similar to the pool/networking structures)...

volatility3/framework/plugins/windows/mftscan.py
kevthehermit: The way the `process_yara_options` is written right now you can pass it one of 3 things:

- A single String or single regex
- Path to a yara text file
- path to a yara compiled binary file. 

It's not possible to pass it a full rule as a string. 

We could add another option to yarascan in order to accept a full rule as an argument and then update this to match? 

Happy to make that change and PR it.

kevthehermit: Alternative woudl be to write the yara rule for MFT records as a regex, but this will slow down scanning.

volatility3/framework/plugins/windows/mftscan.py
kevthehermit: These were taken directly from the old MFT parser plugin in the same format 

https://github.com/volatilityfoundation/volatility/blob/a438e768194a9e05eb4d9ee9338b881c0fa25937/volatility/plugins/mftparser.py#L60


According to the spec on all the sources I can find there is no possible value for 0. We can add one but as its not a valid option it doesn't make sense?

volatility3/framework/plugins/windows/mftscan.py
kevthehermit: It's just a lookup table, I copied it from the existing MFTParser plugin in V2 comparing it against the known values from https://github.com/Invoke-IR/ForensicPosters and "File System Forensic Analysis" pg 353+

The actual values are:
- Not in Use
- In Use
- Directory
- In Use Directory

I just simplified them for the output so it makes more visual sense

kevthehermit:
> 

My original plan was to use an ISF table. But the structure is dynamic in length after the first 56 bytes. This is why we have to dynamically calculate and update the offsets for each pass-through the available attributes. 

I am also only processing 2 of the possible 16 attribute types as the offset calculations get a lot messier.

There can be zero or many attributes of varying length, each attribute can also include a variable-length filename that then pushes all the other offsets off, 

If the ISF can dynamically update for these changes happy to look at it again 

Or we need to write multiple ISFs and then apply each one to the specific attribute type, but we still need to calculate the rolling offset and length fields to get the start of the next attribute.

volatility3/framework/plugins/windows/mftscan.py
kevthehermit: It's a windows plugin but we don't really care about windows, we are scanning for raw MFT records from the $MFT file rather than finding the full MFT and trying to walk it that way. 

This is why I just opted to use the same layer that the default Yara scanner works on

volatility3/framework/plugins/windows/mftscan.py
ikelos: The regexp route would reduce the dependency on yara, but if yara rules are a better way of doing it, then go that way.  I'm up for changing the process_yara_options method, we  just need to figure out if the change is additive (and only bumps the minor version number) or a significant change (in which case we should think through all the use cases and try to make sure it doesn't need change often again).  It somewhat feels as though the reading of the file could be done by the plugins, but I need to think about it a little to figure out the best API layout.  Always open for suggestions too...  5;)

volatility3/framework/plugins/windows/mftscan.py
ikelos: Ok, but from the looks of it, bit 1 is file and bit 2 is directory (and takes priority), so it's either a directory, or a file, or removed.  The fact it was done that way previously doesn't mean it was done the best possible way, so I'm not that bother by what Vol2 did.

I'd do something like `if flag & directory: <directory> elif flag & file: <file> else: removed`.  The structure could either be a BitField or something else.  I'm just not overly keen on have a table of all options

volatility3/framework/plugins/windows/mftscan.py
ikelos: Yep, again, happy that other people do it a particular way, more just suggesting we don't need to do it that way if there's something else that makes more sense.

Specifically, if 0 isn't a valid option, what happens if we encounter it (because we were fed bad data).  How do we want to handle that situation?

volatility3/framework/plugins/windows/mftscan.py
ikelos: Ok, cool, happy with that.  5:)

ikelos:
So since we're doing a lot of code to parse it anyway, I'd have code to apply the templates to anything that's fixed.  (Ie, a template that reads the first 56 bytes, then based on those constructs objects using template after that, and figuring out where the next template goes as much as possible).  I haven't dug into it, so things may get very dynamic very quickly after the header, but it definitely looked as though there wasn't *that* much that was dynamic once the list items had been figured out.  For an example, you might want to look at the poolscanner stuff, where the pool scanner finds a header, and then parses the variable length data based off what the header said that data was...

volatility3/framework/plugins/windows/mftscan.py
kevthehermit: OK updated this to remove that dependency, still using yara but now uses the `process_yara_options`

ikelos:
This will likely become dependent upon #615

kevthehermit:
OK, last commit was a significant set of changes. 

There is now a symbol file and an `MFTIntermedSymbols(intermed.IntermediateSymbolTable)` that will parse out all the elements instead of using hardcoded structs. 

Still using Yara but modified the rule to work with the existing `yara_rules`  option. When #615 goes through we could update it to use that and we should see some minor improvement in the scanning speed with a better rule. 

I still need to look at TimeLine Interface

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,151 @@
...
+from typing import Dict
+
+from volatility3.framework import constants, renderers, interfaces
+from volatility3.framework.configuration import requirements
+from volatility3.framework import exceptions
+from volatility3.framework.renderers import conversion, format_hints
+from volatility3.framework.symbols.windows.extensions.mft import AttributeTypes, NameSpace, PermissionFlags, MFTFlags
+from volatility3.framework.symbols.windows.mft import MFTIntermedSymbols
+
+from volatility3.plugins import timeliner, yarascan
```
ikelos, line 15: Tick for not importing from volatility3.framework, good work!  5:D

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,151 @@
...
+        # Scan the layer for Raw MFT records and parse the fields
+        for offset, rule_name, name, value in layer.scan(context = self.context, scanner = yarascan.YaraScanner(rules = rules)):
+            try:
+                mft_record = self.context.object(mft_object, offset=offset, layer_name=layer.name)
+                # We will update this on each pass in the next loop and use it as the new offset.
+                attr_base_offset = mft_record.FirstAttrOffset
+
+                # There is no field that has a count of Attributes
+                # Keep Attempting to read attributes until we get an invalid attr_header.AttrType
+                while True:
```
ikelos: Not a fan of `while true`, given we've got a guard statement further in of `not AttributeTypes(attr_header.AttrType).value`, I prefer to pre-start the loop and put that as the guard.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,151 @@
...
+                    attr_resident_header = self.context.object(header_object, offset=offset+attr_base_offset+16, layer_name=layer.name)
+
+                    vollog.debug(f"Attr Type: {attr_header.AttrType}")
+
+                    # If this is not a valid type then exit the loop
+                    if not AttributeTypes(attr_header.AttrType).value:
+                        break
+
+                    # Offset past the headers to the attribute data
+                    attr_data_offset = offset+attr_base_offset+24
```
ikelos: Probably better to get the relative offset from an attribute defined within the ISF?  So something like `self.context.symbols.get_type(<type_name>).relative_child_offset(<member_at_the_start_of_attributes>)`.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,151 @@
...
+                    
+                    # Standard Information Attribute
+                    if attr_header.AttrType == 0x10:
```
ikelos: Again, hardcoded values seem bad, particular if it's an `Enum`, we should check the value rather than the integer value.  This should become much easier once they're defined as ISF enums, rather than just a number (see comment on later file).

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,151 @@
...
+                        attr_data = self.context.object(si_object, offset=attr_data_offset, layer_name=layer.name)
```
ikelos: yapf will pick these up, but our typical style has spaces around the equals for keyword parameters and for function argument defaults.

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -0,0 +1,104 @@
...
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import enum
+
+from volatility3.framework import exceptions, objects, renderers
+from volatility3.framework.objects import utility
+
+
+class AttributeTypes(enum.Enum):
```
ikelos: These can be defined as enums within the ISF file, which will mean that values don't have to keep getting wrapped in their class every time we need their name.  An example of how to define them can be found at https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/symbols/windows/pdb.json#L1507.  After that you can just ask the volatility object for `thing.description` (https://github.com/volatilityfoundation/volatility3/blob/280c91212140ceb320ee59c647378e28f919592e/volatility3/framework/objects/__init__.py#L473).

volatility3/framework/symbols/windows/mft.py
```diff
@@ -0,0 +1,15 @@
...
+from volatility3.framework.symbols import intermed
+from volatility3.framework.symbols.windows.extensions import mft
+
+
+class MFTIntermedSymbols(intermed.IntermediateSymbolTable):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.set_type_class('FILE_NAME_ENTRY', mft.MFTFileName)
```
ikelos: This could probably be replaced with a reference to https://github.com/volatilityfoundation/volatility3/blob/280c91212140ceb320ee59c647378e28f919592e/volatility3/framework/symbols/windows/extensions/__init__.py#L457.

volatility3/framework/symbols/windows/mft.json
```diff
@@ -0,0 +1,371 @@
...
+                    "type": {
+                        "kind": "base",
+                        "name": "unsigned int"
+                      }
+                },
+                "NonResidentFlag": {
+                    "offset": 8,
+                    "type": {
+                        "kind": "base",
+                        "name": "unsigned char"                      }
```
ikelos, line 273: Slightly weird formatting, but I'm ok with it...  5;P

volatility3/framework/symbols/windows/mft.json
```diff
@@ -0,0 +1,371 @@
+{
+    "metadata": {
+        "producer": {
+          "version": "0.0.1",
+          "name": "kevthehermit-by-hand",
+          "comment": "Using structures defined in File System Forensic Analysis pg 353+",
```
ikelos, line 6: Nice!  5:D  Thanks for providing references, good work!  5:)

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -0,0 +1,104 @@
...
+        return signature
+
+
+class MFTFileName(objects.StructType):
+    """This represents an MFT $FILE_NAME Attribute"""
+
+    def get_full_name(self) -> str:
+        output = self.Name.cast("string",
+                                    encoding = "utf16",
+                                    max_length = self.NameLength*2,
```
ikelos: Not sure this is necessary, since it essentially duplicates https://github.com/volatilityfoundation/volatility3/blob/280c91212140ceb320ee59c647378e28f919592e/volatility3/framework/symbols/windows/extensions/__init__.py#L457 (see the windows/mft.py file).

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -0,0 +1,104 @@
...
+                                    errors = "replace")
+        return output
+
+    def get_file_namespace(self) -> str:
```
ikelos: Is this for future expansion?

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -0,0 +1,104 @@
...
+    Directory = 0x10000000
+    IndexView = 0x20000000
+    unknown = None
+
+    @classmethod
+    def _missing_(cls, value):
+        return cls(PermissionFlags.unknown)
+
+
+class MFTEntry(objects.StructType):
```
ikelos, line 8: Seems a little excessive, particularly if the rest of this file (the Enums and the MFTFileName become redundant).

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def generate_timeline(self):
+        for row in self._generator():
+            if row[-1] != 'N/A':
+                filename = row[-1]
+                created = f'File {row[-1]} Created'
+                updated = f'File {row[-1]} Updated'
+                modified = f'File {row[-1]} Modified'
+                accessed = f'File {row[-1]} Accessed'
+                yield (f'File {filename} created', timeliner.TimeLinerType.CREATED, row[7])
```
ikelos: I think it would be more useful to return these all in one go, since they relate to the same item.  Since these are the same file, the item value should match (probably just the filename).  The timeliner plugin will then be able to put them on the same row and sort/duplicate them as necessary.  If the documentation needs updating to make this clearer just lemme know and I'll tart it up appropriately...  5:)

ikelos (commented):
Really nice, happy to see the ISF changes, thanks for those!  5:)

Sadly (and I'm sorry you had to type all the enumerations into one file), the enumerations can be stored directly in the ISF (which also makes accessing the names much easier) and vastly reduces the need for the `extensions/mft.py` file.  The `windows/mft.py` file is fine, mostly because we haven't got the normal windows symbols loaded, but otherwise I'd suggest just cross referencing the kernel's symbol table for `_UNICODE_STRING`.  The timeliner stuff would work, but actually it's better to use the first parameter when yielding, but generally all good, and nearly ready for merging!  Great work!  5:D

ikelos:
Ok, changes to yarascan have been made, so just gimme a shout when you want this reviewed again and hopefully it should be pretty trivial to merge then...  5:)

volatility3/framework/plugins/windows/mftscan.py
kevthehermit: `AttributeTypes(attr_header.AttrType).value` is calculated inside the loop on each pass through. This means I can only make the check and break from inside. 

I could do 

```
valid_header = True

while valid_header:
    attr_header = self.context.object(header_object, offset=offset+attr_base_offset, layer_name=layer.name)
 
   . . . 

    if not AttributeTypes(attr_header.AttrType).value:
        valid_header = False


    . . . 

but that feels like it's just a verbose way of saying `while true` ? Happy to write it like this or  take other suggestions.

File: volatility3/framework/symbols/windows/extensions/mft.py (hunk 26 above)
Comment (kevthehermit) [reply volatility3/framework/symbols/windows/extensions/mft.py]: It is a little bit different, 
The UNICODE_STRING uses `max_length = self.Length`

In this instance, we actually have to double the length of the value we read from the offset. which is why we do a `max_length = self.NameLength*2`

File: volatility3/framework/symbols/windows/extensions/mft.py (hunk 27 above)
Comment (kevthehermit) [reply volatility3/framework/symbols/windows/extensions/mft.py]: Removed, Not needed now this is an enum

File: volatility3/framework/plugins/windows/mftscan.py (hunk 29 above)
Comment (kevthehermit) [reply volatility3/framework/plugins/windows/mftscan.py]: Took me a while to realize the description / info is part of the key to put all the times on to a single row.

File: volatility3/framework/plugins/windows/mftscan.py (hunk 18 above)
Comment (ikelos) [reply volatility3/framework/plugins/windows/mftscan.py]: Mmmm, I was thinking more moving the guard up into the while statement, so something like:

```
valid_header = True
attr_header = self.context.object(header_object,
  offset = offset + attr_base_offset,
  layer_name = layer.name)

while attr_header.is_valid_choice():
    ... everything you did previously ...
    
    # Update the base offset to point to the next attribute
    attr_base_offset += attr_header.Length    
    attr_header = self.context.object(header_object,
      offset = offset + attr_base_offset,
      layer_name = layer.name)
```

File: volatility3/framework/plugins/windows/mftscan.py [hunk 30]
```diff
@@ -0,0 +1,159 @@
...
+                                  conversion.wintime_to_datetime(attr_data.AccessedTime), file_name)
+
+                    # Update the base offset to point to the next attribute
+                    attr_base_offset += attr_header.Length
+
+            except exceptions.PagedInvalidAddressException:
+                pass
+
+    def generate_timeline(self):
+        print("MFT Timeline")
```
ikelos: Debugging statements should be done with `vollog` (I've had these sneak in occasionally, but since we've spotted it, best to remove it please).  5;P

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -0,0 +1,21 @@
...
+#
+
+from volatility3.framework import objects
+
+
+class MFTEntry(objects.StructType):
+    """This represents the base MFT Record"""
+
+    def get_signature(self) -> str:
+        signature = self.Signature.cast('string', max_length = 4, encoding = 'latin-1')
```
ikelos, line 12: I'm ok with this since the `get_full_name` method below is still needed, but generally I'd try to discourage class overrides just for a single method that's only a line long.  So with something like this, where the signature which doesn't really need to be a string, but could be matched against an `int` for example (`== 0xDEADBEEF`), I'd probably do it in the code directly.  As I say, fine here because we need the file anyway...  5:)

volatility3/framework/symbols/windows/mft.json
```diff
@@ -0,0 +1,467 @@
...
+                "File": 1,
+                "Directory": 2,
+                "DirInUse": 3
+            },
+            "size": 1
+        },
+        "PermissionFlagEnum": {
+            "base":"unsigned char",
+            "constants": {
+                "ReadOnly": 1,
```
ikelos, line 102: Much nicer!  Thanks!!!  5:D

volatility3/framework/symbols/windows/mft.py
ikelos: Hmmm, a whole file for this really seems like a waste.  I'd recommend just doing the class override when the table is constructed using a dictionary parameter to the create method (https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/symbols/intermed.py#L220), so:

```
        symbol_table = IntermediateSymbolTable.create(self.context, self.config_path, "windows", "mft", class_types = {'FILE_NAME_ENTRY', mft.MFTFileName)
```

If there end up being lots of additional classes needed for handle MFT structures, then it might be worth making a whole table specifically for them, but otherwise it doesn't seem all that useful...

ikelos (commented):
Much nicer, the enumeration stuff feels a lot cleaner!  You're right about the length thing, slightly disappointing they did things differently to the main OS, but fair enough.  Still a couple places where I think we could trim a file, but looking really nice now.  5:)

volatility3/framework/plugins/windows/mftscan.py
kevthehermit: Silly mistake not to check 😁

kevthehermit:
OK, Think that's everything 🤞🏼 and all the suggested changes implemented.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,165 @@
...
+                # We will update this on each pass in the next loop and use it as the new offset.
+                attr_base_offset = mft_record.FirstAttrOffset
+
+                attr_header = self.context.object(header_object,
+                                                  offset = offset + attr_base_offset,
+                                                  layer_name = layer.name)
+
+                # There is no field that has a count of Attributes
+                # Keep Attempting to read attributes until we get an invalid attr_header.AttrType
+                while attr_header.AttrType in attr_types.choices.values():
```
ikelos: Sorry, there is actually an `attr_header.is_valid_choice()` (https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/objects/__init__.py#L492) function, and it will make it easier for us to change if there becomes a more efficient way of doing it...

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,165 @@
...
+                    vollog.debug(f"Attr Type: {attr_header.AttrType}")
+
+                    # Offset past the headers to the attribute data
+                    attr_data_offset = offset + attr_base_offset + self.context.symbol_space.get_type(
+                        attribute_object).relative_child_offset("Attr_Data")
+
+                    # MFT Flags determine the file type or dir
+                    if mft_record.Flags in mft_flags.choices.values():
```
ikelos: Could be `is_valid_choice()` as well.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,165 @@
...
+                            conversion.wintime_to_datetime(attr_data.UpdatedTime),
+                            conversion.wintime_to_datetime(attr_data.AccessedTime),
+                            renderers.NotApplicableValue(),
+                        )
+
+                    # File Name Attribute
+                    if attr_header.AttrType == attr_types.choices.get('FILE_NAME'):
+                        attr_data = self.context.object(fn_object, offset = attr_data_offset, layer_name = layer.name)
+                        file_name = attr_data.get_full_name()
+                        if attr_data.Flags in permission_flags.choices.values():
```
ikelos: Here too

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,165 @@
...
+                            attr_types.lookup(attr_header.AttrType),
+                            conversion.wintime_to_datetime(attr_data.CreationTime),
+                            conversion.wintime_to_datetime(attr_data.ModifiedTime),
+                            conversion.wintime_to_datetime(attr_data.UpdatedTime),
+                            conversion.wintime_to_datetime(attr_data.AccessedTime),
+                            renderers.NotApplicableValue(),
+                        )
+
+                    # File Name Attribute
+                    if attr_header.AttrType == attr_types.choices.get('FILE_NAME'):
```
ikelos: I think this is essentially equivalent to `attr_header.AttrType.lookup() == 'FILE_NAME'` which I think is a little easier to understand.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,165 @@
...
+                        attr_data = self.context.object(fn_object, offset = attr_data_offset, layer_name = layer.name)
+                        file_name = attr_data.get_full_name()
+                        if attr_data.Flags in permission_flags.choices.values():
+                            permissions = permission_flags.lookup(attr_data.Flags)
```
ikelos: Here you can just do `attr_data.Flags.lookup()`

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,165 @@
...
+                        else:
+                            permissions = hex(attr_data.Flags)
+
+                        yield 1, (format_hints.Hex(attr_data_offset), mft_record.get_signature(),
+                                  mft_record.RecordNumber, mft_record.LinkCount, mft_flag, permissions,
+                                  attr_types.lookup(attr_header.AttrType),
```
ikelos: This also could just be `attr_header.AttrType.lookup()`.  The Enum already knows what type it is, so the object carries around all its possible choices.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,165 @@
...
+                        attr_data = self.context.object(si_object, offset = attr_data_offset, layer_name = layer.name)
+
+                        yield 0, (
+                            format_hints.Hex(attr_data_offset),
+                            mft_record.get_signature(),
+                            mft_record.RecordNumber,
+                            mft_record.LinkCount,
+                            mft_flag,
+                            renderers.NotApplicableValue(),
+                            attr_types.lookup(attr_header.AttrType),
```
ikelos: Here too `attr_header.AttrType.lookup()` would do.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -0,0 +1,165 @@
...
+                while attr_header.AttrType in attr_types.choices.values():
+                    vollog.debug(f"Attr Type: {attr_header.AttrType}")
+
+                    # Offset past the headers to the attribute data
+                    attr_data_offset = offset + attr_base_offset + self.context.symbol_space.get_type(
+                        attribute_object).relative_child_offset("Attr_Data")
+
+                    # MFT Flags determine the file type or dir
+                    if mft_record.Flags in mft_flags.choices.values():
+                        mft_flag = mft_flags.lookup(mft_record.Flags)
```
ikelos: The object carries around its own choices, and if you call lookup without an argument, it passes `self`, so this could just be `mft_record.Flags.lookup()`.

ikelos (commented):
Sooooo close!  There's a couple more efficiencies with how enumerations are used that could make this a bit better.  It's technically fine, but since people may refer to this to figure out how to do stuff, I'd prefer it did it as best as possible...  5:S  After that I think it's merge worthy!  5:D

ikelos (approved):
Looks good, thanks!  5:D

PR #625, opened 2022-01-16
use doc/requirements.txt when building with readthedocs

npetroni:


ikelos (approved):
Looks good and should be the minimum to get the readthedocs to build properly.  5:)

PR #630, opened 2022-01-24
Feature/isf unification

### LINKED ISSUE #618: Vol.py `-s` switch is too complicated
**Describe the bug**
Add a new symbol directory using `-s` doesn't add the files to the linux or mac banner caches, meaning any ISF files found there are ignored.

Putting the same file in `volatility3/symbols` or `volatility3/framework/symbols` allows it to be processed correctly and everything runs fine...

ikelos:
Big, but hopefully unnoticable change to how we handle a generic list of identifiers and ISF files.  Previously this was only used for linux/mac, and the identifiers were the banner, however the guid/age/pdbname could also be used as an identifier for windows ISF files.  This luckily *should* be stored in the ISF files in the metadata section, and the autogeneration code of volatility has always done this.  Previously volatility used the file name and location to identify and load the right ISF for windows.

Now, the various OSes are unified under a single cache and allows symbol files to be dropped under any name in any subdirectory under the symbols directory.  The primary implementation stores the data in a database that's updated each run (the first run will take a while depending on how many ISF files there are).  If the database ever is corrupted, it's wiped and starts again, so it shouldn't be possible for a bad cache to hang volatility, but it's worth checking.  The database *can* return multiple locations for a single identifier (although each location can only store one identifier), but that should result in warnings.|

ISF files are allowed to not contain an identifier, but then they must not have an operating system associated with them.  The changes should not have been relied on, however the sqlcache is versionable to make sure that internal components that use it can check it's the right implementation (the interface isn't versioned, perhaps it should be instead of the concrete implementation?).\

isfinfo now makes use of it, as does the pdbutility code.

I think that covers everything, but this could do with a fair amount of tire kicking.  There's been a lot changed and it could almost certainly do with tweaks, both in terms of bugs and in terms of how it's architected.  Going to wait for review on this rather than the usual week for review...

### DIFF
--- a/volatility3/framework/automagic/symbol_cache.py
+++ b/volatility3/framework/automagic/symbol_cache.py
@@ -104,7 +104,7 @@ class CacheManagerInterface(interfaces.configuration.VersionableInterface):
         """Returns the location of the symbol file given the identifier
 
         Args:
-            identifier: string that uniquely identifies a particular symbolt table
+            identifier: string that uniquely identifies a particular symbol table
             operating_system: optional string to restrict identifiers to just those for a particular operating system
 
         Returns:
@@ -144,6 +144,18 @@ class CacheManagerInterface(interfaces.configuration.VersionableInterface):
         """Returns all identifiers for a particular operating system"""
         pass
 
+    def get_location_statistics(self, location: str) -> Optional[Tuple[int, int, int, int]]:
+        """Returns ISF statistics based on the location
+
+        Returns:
+            A tuple of base_types, types, enums, symbols, or None is location not found"""
+
+    def get_verified(self, location: str) -> bool:
+        """Returns whether a location ISF has been verified against its schema"""
+
+    def set_verified(self, location: str, state: bool = True) -> None:
+        """Sets the verified state of a location based on whether it has been successfully verified against its schema"""
+
 
 class SqliteCache(CacheManagerInterface):
     _required_framework_version = (2, 0, 0)
@@ -163,7 +175,23 @@ class SqliteCache(CacheManagerInterface):
         database = sqlite3.connect(path)
         database.row_factory = sqlite3.Row
         database.cursor().execute(
-            'CREATE TABLE IF NOT EXISTS cache (location TEXT UNIQUE NOT NULL, identifier TEXT, operating_system TEXT, local BOOL, cached DATETIME)')
+            f'CREATE TABLE IF NOT EXISTS database_info (schema_version INT DEFAULT {constants.CACHE_SQLITE_SCEMA_VERSION})')
+        schema_version = database.cursor().execute('SELECT schema_version FROM database_info').fetchone()
+        if not schema_version:
+            database.cursor().execute(f'INSERT INTO database_info VALUES ({constants.CACHE_SQLITE_SCEMA_VERSION})')
+        elif schema_version['schema_version'] == constants.CACHE_SQLITE_SCEMA_VERSION:
+            # All good, so pass and move on
+            pass
+        else:
+            vollog.info(f"Previous cache schema version found: {schema_version['schema_version']}")
+            # TODO: Implement code if the schema changes
+            # Current this should never happen so we start over again
+            database.close()
+            os.unlink(path)
+            return self._connect_storage(path)
+        database.cursor().execute(
+            'CREATE TABLE IF NOT EXISTS cache (location TEXT UNIQUE NOT NULL, identifier TEXT, operating_system TEXT, verified BOOL DEFAULT False,'
+            'stats_base_types INT DEFAULT 0, stats_types INT DEFAULT 0, stats_enums INT DEFAULT 0, stats_symbols INT DEFAULT 0, local BOOL, cached DATETIME)')
         database.commit()
         return database
 
@@ -207,6 +235,25 @@ class SqliteCache(CacheManagerInterface):
             return row['identifier']
         return None
 
+    def get_location_statistics(self, location: str) -> Optional[Tuple[int, int, int, int]]:
+        results = self._database.cursor().execute(
+            'SELECT stats_base_types, stats_types, stats_enums, stats_symbols FROM cache WHERE location = ?',
+            (location,)).fetchall()
+        for row in results:
+            return row['stats_base_types'], row['stats_types'], row['stats_enums'], row['stats_symbols']
+        return None
+
+    def get_verified(self, location: str) -> bool:
+        results = self._database.cursor().execute('SELECT verified FROM cache WHERE location = ?',
+                                                  (location,)).fetchall()
+        for row in results:
+            return row['verified']
+        return False
+
+    def set_verified(self, location: str, state: bool = True) -> None:
+        self._database.cursor().execute('UPDATE cache (verified) VALUES (?) WHERE location = ?',
+                                        (state, location,))
+
     def update(self, progress_callback = None):
         """Locates all files under the symbol directories.  Updates the cache with additions, modifications and removals.
         This also updates remote locations based on a cache timeout.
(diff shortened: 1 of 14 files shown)


ikelos:
I also need to update the documentation at https://volatility3.readthedocs.io/en/latest/complex-plugin.html#writing-using-intermediate-symbol-format-files

ikelos:
And probably go through the code and take out anywhere that makes use of the subpath part of that function.

ikelos:
Anyone had a chance to prod this yet?  It's pretty big so I'd prefer we get it merged before too much else changes the ground beneath it...

ikelos:
This has been sat around for a while, there's been several months for testing.  I'd really like someone to get some eyes on it if possible, but if not then I'll just end up merging it and fixing any issues that arise afterwards, so anyone that wants to give it a look please do so!  5:)

volatility3/framework/plugins/isfinfo.py
```diff
@@ -98,33 +96,52 @@ def check_valid(data):
...
+                    yield (0, (location, valid, num_bases, num_types, num_symbols, num_enums, str(identifier)))
 
     # Try to open the file, load it as JSON, read the data from it
 
     def run(self):
         return renderers.TreeGrid([("URI", str), ("Valid", str),
                                    ("Number of base_types", int), ("Number of types", int), ("Number of symbols", int),
-                                   ("Number of enums", int), ("Windows info", str), ("Linux banner", str),
-                                   ("Mac banner", str)], self._generator())
+                                   ("Number of enums", int), ("Identifying infomration", str)], self._generator())
```
digitalisx: Found typo!
```suggestion
                                   ("Number of enums", int), ("Identifying information", str)], self._generator())
```

volatility3/framework/automagic/symbol_cache.py
```diff
@@ -22,164 +25,392 @@
...
+        idextractors = list(volatility3.framework.class_subclasses(IdentifierProcessor))
+
+        # New or not recently updated
+
+        counter = 0
+        files_to_process = new_locations.union(cache_update)
+        number_files_to_process = len(files_to_process)
+        cursor = self._database.cursor()
+        try:
+            for location in files_to_process:
```
digitalisx: What do you think about using `enumerate`? It has the advantage of not having to declare a separate variable.
```suggestion
            for counter, location in enumerate(files_to_process):
```

digitalisx (commented):
@ikelos Thank you for your work! 🙌
I also tested the PR and it seems to work well.
I leave a minor comment in the code :)

volatility3/framework/automagic/symbol_cache.py
digitalisx: If you use this, you can remove the `counter` variable.

volatility3/framework/automagic/symbol_cache.py
```diff
@@ -188,23 +417,23 @@ def _verify(self) -> bool:
...
-            for banner in self._data[operating_system]:
-                binary_banner = base64.b64decode(banner)
-                file_list = banners.get(binary_banner, [])
-                for value in self._data[operating_system][banner]:
-                    if value not in file_list:
-                        file_list = file_list + [value]
-                    banners[binary_banner] = file_list
+            for identifier in self._data[operating_system]:
+                binary_identifier = base64.b64decode(identifier)
+                file_list = identifiers.get(binary_identifier, [])
```
digitalisx: It seems to be an unused variable.
```suggestion
```

volatility3/framework/configuration/requirements.py
```diff
@@ -406,13 +408,19 @@ def unsatisfied(self, context: interfaces.context.ContextInterface,
...
-        if len(self._version) > 0 and self._component.version[0] != self._version[0]:
-            return {config_path: self}
-        if len(self._version) > 1 and self._component.version[1] < self._version[1]:
+        if not self.matches_required(self._version, self._component.version):
             return {config_path: self}
         context.config[interfaces.configuration.path_join(config_path, self.name)] = True
         return {}
 
+    @classmethod
+    def matches_required(cls, required: Tuple[int, ...], version: Tuple[int, int, int]):
```
digitalisx: Could you provide the return type the same as any other changes?

```suggestion
    def matches_required(cls, required: Tuple[int, ...], version: Tuple[int, int, int]) -> bool:
```

volatility3/framework/automagic/symbol_cache.py
```diff
@@ -22,164 +25,390 @@
...
+
+    def __init__(self, filename: str):
+        super().__init__(filename)
+        try:
+            self._database = self._connect_storage(filename)
+        except sqlite3.DatabaseError:
+            os.unlink(filename)
+            self._database = self._connect_storage(filename)
+
+    def _connect_storage(self, path: str):
```
digitalisx: It seems to be return the `Connection` Object of SQLite3.

```suggestion
    def _connect_storage(self, path: str) -> sqlite3.Connection:
```

volatility3/framework/automagic/symbol_cache.py
```diff
@@ -22,164 +25,390 @@
...
+        """Returns all identifiers for a particular operating system"""
+        pass
+
+    def get_location_statistics(self, location: str) -> Optional[Tuple[int, int, int, int]]:
+        """Returns ISF statistics based on the location
+
+        Returns:
+            A tuple of base_types, types, enums, symbols, or None is location not found"""
+
+    def get_hash(self, location: str) -> bool:
```
digitalisx: I think the return type of hash is not a bool.

```suggestion
    def get_hash(self, location: str) -> Optional[str]:
```

volatility3/framework/automagic/symbol_cache.py
```diff
@@ -22,164 +25,390 @@
...
+            statement += f" WHERE {' AND '.join(additions)}"
+        results = self._database.cursor().execute(statement)
+        for row in results:
+            if row['identifier'] in output and row['identifier'] and row['location']:
+                vollog.debug(
+                    f"Duplicate entry for identifier {row['identifier']}: {row['location']} and {output[row['identifier']]}")
+            output[row['identifier']] = row['location']
+        return output
+
+    def get_identifiers(self, operating_system: Optional[str]):
```
digitalisx: If decision return type on interface, It will be change :)

volatility3/framework/automagic/symbol_cache.py
```diff
@@ -22,164 +25,390 @@
...
+        Returns:
+            A dictionary of identifiers mapped to a location
+        """
+        pass
+
+    def get_identifier(self, location: str) -> Optional[bytes]:
+        """Returns an identifier based on a specific location or None"""
+        pass
+
+    def get_identifiers(self, operating_system: Optional[str]):
```
digitalisx: Is there a reason why only this type hint is missing? :)

volatility3/framework/automagic/symbol_cache.py
```diff
@@ -22,164 +25,390 @@
...
+        if operating_system is not None:
+            statement = 'SELECT location FROM cache WHERE identifier = ? AND operating_system = ?'
+            parameters = (identifier, operating_system)
+        results = self._database.cursor().execute(statement, parameters).fetchall()
+        result = None
+        for row in results:
+            result = row['location']
+        return result
+
+    def get_local_locations(self) -> Generator[str, None, None]:
```
digitalisx, line 218: It seems to be different from the type defined in the interface.

digitalisx (commented):
@ikelos Thank you for applying my suggestion, Additionally I leave small comments on the unification of type hint. It is a recommendation based on Python PEP, not mandatory.. :)

volatility3/framework/automagic/symbol_cache.py
digitalisx: In the interface, it is defined like this: : `List[str]`

ikelos: Thanks, I've changed the interface to an `Iterable` since every Generator is an Iterable, and this does technically return a `Generator`.

volatility3/framework/automagic/symbol_cache.py
ikelos: Heheh, I just forgot, but added the right hints in (I used Lists rather than Iterables, but I may change that in the future).

digitalisx (approved):
@ikelos Thank you for apply my suggestion.
I think that's all I can review.
And this code looks better then before and it's a good change.
I hope the other maintainers think the same.. 🙂

ikelos:
@digitalisx thanks very much for looking it over and improving it.  I'm glad I've now got confidence that it's been checked over at least once, so I'm much happier merging it...  I expect it to get merged at the end of the week, I just need to figure out if I need to post an announcement about the change on slack, given the first run will be quite long when it needs to initially build the cache...

npetroni:
When I run `windows.pslist` against a Windows 10 image on 9be57a43146c17f2d1a6625b328acaba78d43f95, it produces the correct output. However, subsequent runs of the same command seem to re-parse the PDB file. Is that expected? On `develop`, the same sequence of commands does *not*  result in a second parsing of the PDB.

The tell-tale sign is the printing of `Reading TPI layer` in the output.  The second run takes significantly longer on my Mac for `isf-unification` than for `develop`, unless I am doing something wrong (very possible!).

ikelos:
Hmmm, no that's very definitely wrong.  It should have cached it and be pulling it back from the cache directory (although it has been a while since I've looked at the code).  Lemme see if I can recreate it...

ikelos:
Hmmm, so it seems to process it, and update the cache, but the generated JSON never gets written to the directory...  5:S

ikelos:
Well, that's thoroughly disheartening.  When we find a PDB entry, it has an age of say, 1.  We then grab the corresponding pdb file from Microsoft, with a 1 as the age in the filename.  We process it and it says that its age is in fact 2.  We therefore stash its identifier including an age of 2.  When we find the PDB entry in memory again, it's age 1, so we redownload it...  5:\

ikelos:
The pdb name with the age value taken *from* the pdb file doesn't exist on Microsoft servers, so it looks like we've got a mismatch between fields.  Unfortunately, that means even if we sort this, we'll need to regenerate the entire library based off the filenames (doable, but annoying)...  5:S

digitalisx:
When I first checked the operation, I felt that there was no problem.
But when I checked again, the same phenomenon occurred. I couldn't double check it.. 😱
Watching your error tracking, I'll be happy to comment if I can think of anything.. 🙂

ikelos:
Ok, so, confusingly it looks like there's a *second* field called age in the `DBI_HEADER` structure, which is the one we should have been using (instead of the age in the PDB header itself).  I've added a fix, but all previously generated PDB files will be inaccurate and not match correctly (which should just mean they're re-downloaded and updated, possibly taking up some disk space even though they'll no longer be used).  We should really go through the old windows.zip file and update it based off the filename (which used the age value we'd wanted) and also add some more recent kernels in.  The windows.zip file has some historical symbol tables whose data is no longer available, so it is still a useful resource and one worth updating...

Very happy to have people think over the consequences of changing the age value this late in the game...  5:S

digitalisx:
I don't know if it'll help, but... The relatively recent Symbol Tables appear to be organized. :)
(https://github.com/JPCERTCC/Windows-Symbol-Tables)

mischw:
> ... We should really go through the old windows.zip file and update it based off the filename (which used the age value we'd wanted) and also add some more recent kernels in. ...

Hey. Are there any plans to update the symbol tables? Or is any help wanted?

ikelos:
There hasn't been yet, but if you'd like to help we'll happy accept any progress you can make on the task?  The win.zip pack as it stands now is useful because it contains some historical JSON files that were made before Microsoft seems to have changed some of their pdb files to remove some typing information from them.  I'll look over your pull request and we can go from there...  5:)

PR #635, opened 2022-02-02
Add the psaux plugin for Linux command line argument listing

atcuno:


volatility3/framework/plugins/linux/psaux.py
```diff
@@ -0,0 +1,90 @@
...
+                return renderers.UnreadableValue()
+
+            proc_layer = self.context.layers[proc_layer_name]
+
+            # read argv from userland
+            start = task.mm.arg_start
+
+            # get the size of the arguments with sanity checking
+            size_to_read = task.mm.arg_end - task.mm.arg_start
+            if size_to_read < 1 or size_to_read > 4096:
```
ikelos: Not an issue, but I think pyfhon allows this to be written as: `if not (0 < size_to_read <= 4096)`

volatility3/framework/plugins/linux/psaux.py
```diff
@@ -0,0 +1,90 @@
...
+                return renderers.UnreadableValue()
+
+            # attempt to read it all as partial values are invalid and misleading
+            try:
+                argv = proc_layer.read(start, size_to_read)
+            except exceptions.InvalidAddressException:
+                return renderers.UnreadableValue()
+
+            # the arguments are null byte terminated, replace the nulls with spaces
```
ikelos, line 70: Is this reasonable/what investigators would expect and/or does it allow for confusion between parameters?  I don't want to print a null character, and most other characters have other meanings (such as `|`), but if there's an easy way to avoid human confusion then we should take it.  If not, then no problem...  5:)

volatility3/framework/plugins/linux/psaux.py
```diff
@@ -0,0 +1,90 @@
...
+            s = argv.decode().split('\x00')
+            args = " ".join(s)
+        else:
+            # kernel thread
+            # [ ] mimics ps on a live system
+            # also helps identify malware masquerading as a kernel thread, which is fairly common
```
ikelos, line 76: Really appreciate having these comments!  Great stuff, thanks!  5:D

volatility3/framework/plugins/linux/psaux.py
```diff
@@ -0,0 +1,90 @@
...
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+from typing import Optional
+
+from volatility3.framework import symbols, exceptions, renderers, interfaces
+from volatility3.framework.objects import utility
+from volatility3.plugins.linux import pslist
+
+class PsAux(pslist.PsList):
```
ikelos: Errr, this isn't ideal unless you're actually going to use pslist's generator and run methods (which would then likely just give you a copy of pslist).  To prevent breakages, please drop the inheritance and call `list_tasks` from the pslist class directly (which should be a classmethod for this very reason).  You'll need to write your own `get_requirements` method, but you'd have to anyway, because there should be a `VersionRequirement` or `PluginRequirement` (they're practically identical) to make sure pslist does what you expect it to do.


Not inheriting just prevents this breaking if pslist changes in the future.  Strictly there should be a requirement on a specific version of pslist, but if you're going that far, you might as well

ikelos (commented):
Pretty clean, *really* happy to have all the comments, good work there!  5:D  I think the only blocker is inheriting from `pslist` rather than having it as a dependency requirement and just calling into it.

Once that's fixed up, lemme know and we can get it merged...  5:)

volatility3/framework/plugins/linux/psaux.py
atcuno: It is okay. if the plugin shows spaces inside an argument then it means they are associated, otherwise the shell / command would not have executed correctly.

atcuno:
@ikelos I believe I addressed all your feedback in the commit I just made.

ikelos:
Thanks, it looks like it still inherits from pslist, even though you've got a get_requirements method.  Is the inheritance still needed and can it be removed/changed?

atcuno:
> Thanks, it looks like it still inherits from pslist, even though you've got a get_requirements method. Is the inheritance still needed and can it be removed/changed?

fixed. thanks.

PR #641, opened 2022-02-09
Fixed pdb scanning bug

paulkermann:
Added missing `data_offset` to pattern matching result. This fixes an issue in identifying Windows PDB records. 

_Previous versions of this code used to have this fix_. See the change in the yield (line 362), in commit c5c726ab2865966597b647576e20dbf6175b7aa3


ikelos (approved):
Wow, really good spot!  I'm amazed that it didn't affect operation enough for more people to shout?  I guess we were lucky/unlucky that some pdbs still turned up on scan chunk boundaries...  5:S  Thanks very much!

PR #643, opened 2022-02-09
fix pdb mz parsing

ilay122:
The e_lfanew is a uint32 and not a byte.
I have had an issue where the `get_guid_from_mz` would fail because of that.

### DIFF
--- a/volatility3/framework/symbols/windows/pdbutil.py
+++ b/volatility3/framework/symbols/windows/pdbutil.py
@@ -131,8 +131,14 @@ class PDBUtility(interfaces.configuration.VersionableInterface):
         # Check it is actually the MZ header
         if mz_sig != b"MZ":
             return None
-
-        nt_header_start = ord(layer.read(offset + 0x3C, 1))
+        
+        nt_header_start = struct.unpack("<I", layer.read(offset + 0x3C, 4))[0]
+        pe_sig = layer.read(offset + nt_header_start, 2)
+        
+        # Check it is actually the Nt Headers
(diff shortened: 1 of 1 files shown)


volatility3/framework/symbols/windows/pdbutil.py
```diff
@@ -131,8 +131,14 @@ def get_guid_from_mz(cls, context: interfaces.context.ContextInterface, layer_na
         # Check it is actually the MZ header
         if mz_sig != b"MZ":
             return None
-
-        nt_header_start = ord(layer.read(offset + 0x3C, 1))
+        
+        nt_header_start = struct.unpack("<I", layer.read(offset + 0x3C, 4))[0]
```
ikelos, line 135: I'd be a little happier using `int.from_bytes(..., byteorder='little')`, since it doesn't require the use of an additional module or splitting a tuple, but it's not a showstopper...

ikelos (commented):
Ah, you're quite right, thanks for spotting that!  5:)  I'd slightly prefer to use `int.from_bytes` rather than `struct`, but it's not a show stopper.  If you're happy to make the change, or have reason not to, then either way I'm happy to merge it, just let me know.  Thanks!  5:)

volatility3/framework/symbols/windows/pdbutil.py
ilay122: I prefer struct because it is more specific.

ikelos:
Fair enough, thanks again for the submission!  5:)

PR #644, opened 2022-02-10
Object information ChainMap invalid order

paulkermann:
When creating the `self._vol` the input is the new dict with the masked offset and the original `object_info`.
By passing the `object_info` first the `normalized_offset` is not accessible.
Example:
```
context.object("function", offset=0xffffffff80000000, ...).vol.offset == 0xffffffff80000000 # problem
context.object("function", offset=0xffff80000000,     ...).vol.offset == 0xffffffff80000000 # ok
```
By reversing the order we get the expected offset.

### DIFF
--- a/volatility3/framework/interfaces/objects.py
+++ b/volatility3/framework/interfaces/objects.py
@@ -115,7 +115,7 @@ class ObjectInterface(metaclass = abc.ABCMeta):
         mask = context.layers[object_info.layer_name].address_mask
         normalized_offset = object_info.offset & mask
 
-        self._vol = collections.ChainMap({}, object_info, {'type_name': type_name, 'offset': normalized_offset}, kwargs)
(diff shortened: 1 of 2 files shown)


ikelos:
Thanks very much, I'm not sure how we missed that.  This could have some unintended consequences, so I'm trying to figure out if it's worth a version bump so people can tell when the change happened?  @npetroni what are your thoughts?  @atcuno @iMHLv2, have you ever noticed this?  I'm leaning towards merging it, since I think it's what we'd always intended to happen (the code was there, but wasn't functioning properly).

paulkermann:
@ikelos I pushed another fix.
This code assumed the dicts to be at fixed indexes so I just switched them up.

ikelos:
Hmmm, I haven't been able to find an instance where the two don't match, so I don't think this will affect too many existing users, so I'm going to merge and we'll have to see if anything extremely subtle breaks...

Issue #313, opened 2020-08-24
LDR modules


AsafEitani:
**Is your feature request related to a problem? Please describe.**
LDR modules used to detect unlinked dlls, was under volatility 2 mafind.

**Describe the solution you'd like**
Add LDRmodules plugin like in volatility 2.

ikelos:
There may already be functionality present in one of the other plugins, but I'll leave this here until we get confirmation...  5:)

iMHLv2:
Ldrmodules basically diffs the output of vadinfo and dlllist (both of which are present in vol3). We can create ldrmodules for vol3, but probably will not happen until after the first release.

pudii:
@ikelos I think this can be closed as well.

ikelos:
Yep, thanks catching that!  5:)

PR #645, opened 2022-02-11
Implement LDRmodules plugin

pudii:
Implementation of the LDRmodules plugin for volatility3 as it was done in volatility2.
The output and the functionality should be similar.

Solution for issue #313 .

### DIFF
--- a/volatility3/framework/plugins/windows/ldrmodules.py
+++ b/volatility3/framework/plugins/windows/ldrmodules.py
@@ -1,0 +1,97 @@
+from volatility3.framework import interfaces, constants
+from volatility3.framework import renderers, interfaces, exceptions
+from volatility3.framework.configuration import requirements
+from volatility3.framework.objects import utility
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.symbols import intermed
+from volatility3.framework.symbols.windows.extensions import pe
+from volatility3.plugins.windows import pslist, vadinfo
+
+class LdrModules(interfaces.plugins.PluginInterface):
+    _required_framework_version = (2, 0, 0)
+    _version = (1, 0, 0)
+ 
+    @classmethod
+    def get_requirements(cls):
+        return [
+            requirements.ModuleRequirement(name = 'kernel', description = 'Windows kernel', architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
+            requirements.VersionRequirement(name = 'pslist', component = pslist.PsList, version = (2, 0, 0)),
+            requirements.VersionRequirement(name = 'vadinfo', component = vadinfo.VadInfo, version = (2, 0, 0)),
+            requirements.ListRequirement(name = 'pid',
+                                         element_type = int,
+                                         description = "Process IDs to include (all other processes are excluded)",
+                                         optional = True),
+           ]
+
+    def _generator(self, procs):
+
+        pe_table_name = intermed.IntermediateSymbolTable.create(self.context,
+                                                                self.config_path,
+                                                                "windows",
+                                                                "pe",
+                                                                class_types = pe.class_types)
+
+        def filter_function(x: interfaces.objects.ObjectInterface) -> bool:
+            try:
+                return not (x.get_private_memory() == 0 and x.ControlArea)
+            except AttributeError:
+                return False
+
+        filter_func = filter_function   
+
+        for proc in procs:
+            proc_layer_name = proc.add_process_layer()
+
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/ldrmodules.py
```diff
@@ -0,0 +1,97 @@
...
+
+class LdrModules(interfaces.plugins.PluginInterface):
+    _required_framework_version = (2, 0, 0)
+    _version = (1, 0, 0)
+ 
+    @classmethod
+    def get_requirements(cls):
+        return [
+            requirements.ModuleRequirement(name = 'kernel', description = 'Windows kernel', architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
```
ikelos: This should be unnecessary, it's now part of the `ModuleRequirement` (`kernel.symbol_table_name`).

volatility3/framework/plugins/windows/ldrmodules.py
```diff
@@ -0,0 +1,97 @@
...
+            mem_order_mod = dict((mod.DllBase, mod)
+                                for mod in proc.mem_order_modules())
+
+            mapped_files = {}
+            for vad in vadinfo.VadInfo.list_vads(proc, filter_func = filter_func):
+                dos_header = self.context.object(pe_table_name + constants.BANG + "_IMAGE_DOS_HEADER",
+                                   offset = vad.get_start(),
+                                   layer_name = proc_layer_name)
+                try:
+                    if dos_header.e_magic != 0x5A4D:
```
ikelos, line 61: It would be good to get a comment here to say that this is looking for VADs that start with an MZ header.  Should this check the contents of the entire vad, or do modules only ever start at the beginning of a vad?

volatility3/framework/plugins/windows/ldrmodules.py
```diff
@@ -0,0 +1,97 @@
...
+                        continue
+                except exceptions.PagedInvalidAddressException:
+                    continue
+
+                mapped_files[int(vad.get_start())] = str(vad.get_file_name() or '')
```
ikelos: I don't know that this needs to be cast as an int?  It should already be an int that gets returned, but if python says it's unhashable that would be good to know.  I'm just trying to avoid unnecessary casts (that might make other people include them too).  5:)  Same with the `str`.

volatility3/framework/plugins/windows/ldrmodules.py
```diff
@@ -0,0 +1,97 @@
...
+                load_mod = load_order_mod.get(base, None)
+                init_mod = init_order_mod.get(base, None)
+                mem_mod = mem_order_mod.get(base, None)
+
+                yield (0, [int(proc.UniqueProcessId),
+                            str(proc.ImageFileName.cast("string",
+                                                max_length = proc.ImageFileName.vol.count,
+                                                errors = 'replace')),
+                            format_hints.Hex(base),
+                            str(load_mod != None),
```
ikelos: If you don't want to return the actual value of these things (and use a `BaseAbsentValue` for when it's not there), then it would be better to make this a bool type, so that another UI can directly make use of the value if it wants, rather than having to parse a string.

volatility3/framework/plugins/windows/ldrmodules.py
```diff
@@ -0,0 +1,97 @@
...
+                    if dos_header.e_magic != 0x5A4D:
+                        continue
+                except exceptions.PagedInvalidAddressException:
+                    continue
+
+                mapped_files[int(vad.get_start())] = str(vad.get_file_name() or '')
+
+            for base in mapped_files.keys():
+                # Does the base address exist in the PEB DLL lists?
+                load_mod = load_order_mod.get(base, None)
```
ikelos, line 70: If you want to display the value, rather than simply present or not, it would be better/easier to use something descended from `BaseAbsentValue`, since the formatters can decide how to represent a missing value.  If you just want a true/false value then `None` is fine here.

ikelos (commented):
Thanks, this looks pretty good, there's a couple of small comments, but they should all be pretty easy to fix up.  5:)

volatility3/framework/plugins/windows/ldrmodules.py
pudii: True/False is fine here. :)

pudii:
Thanks for the feedback. Should be fixed with the commit [12c3f34](https://github.com/volatilityfoundation/volatility3/pull/645/commits/12c3f340370f5e4dd1cb34d6803154c23164d234).

ikelos (approved):
Ok, looks good, thanks!  5:)

PR #650, opened 2022-02-16
pdbconv improvments

paulkermann:
Only after I finished this PR I saw [this](https://github.com/volatilityfoundation/volatility3/commit/796cf69b205e428385cd63dcc67dc40ce469c4a9) but either way, I added support for parsing some C++ types and a fixed a small bug that could cause fields in a struct  to be skipped.

I will attach the source pdb test and the result ISF so you can try for yourself.
[input.zip](https://github.com/volatilityfoundation/volatility3/files/8079396/input.zip)

Looking forward to hearing your feedback.


volatility3/framework/symbols/windows/pdbconv.py
```diff
@@ -828,7 +867,9 @@ def convert_fields(self, fields: int) -> Dict[Optional[str], Dict[str, Any]]:
             vollog.warning("Fields structure did not contain a list of fields")
             return result
         for field in fields_struct:
-            _, name, member = field
+            leaf_type, name, member = field
+            if leaf_type in [leaf_type.LF_NESTTYPE, leaf_type.LF_METHOD, leaf_type.LF_STMEMBER, leaf_type.LF_ONEMETHOD]:
```
ikelos: Be nice to have a comment here explaining that this is just to ignore these types until we can write them (perhaps a `# TODO` of `# FIXME`?)

volatility3/framework/symbols/windows/pdbconv.py
```diff
@@ -755,22 +772,31 @@ def process_types(self, type_references: Dict[str, int]) -> None:
...
+
     def consume_type(
             self, module: interfaces.context.ModuleInterface, offset: int, length: int
     ) -> Tuple[Tuple[Optional[interfaces.objects.ObjectInterface], Optional[str], Union[
         None, List, interfaces.objects.ObjectInterface]], int]:
         """Returns a (leaf_type, name, object) Tuple for a type, and the number
         of bytes consumed."""
         leaf_type = self.context.object(module.get_enumeration("LEAF_TYPE"),
-                                        layer_name = module._layer_name,
+                                        layer_name = module.layer_name,
```
ikelos: Thanks, not sure how this snuck in, but good spot.  5:)

volatility3/framework/symbols/windows/pdbconv.py
```diff
@@ -803,11 +829,24 @@ def consume_type(
                 else:
                     name = self.parse_string(parsed_obj.name, leaf_type < leaf_type.LF_ST_MAX,
                                              size = remaining - name_offset)
+                    current_consumed = parsed_obj.vol.size + len(name) + 1
```
ikelos: I might be tempted to preinitialize `excess `to 0, and then move the `current_consumed` line down so it happens in both cases

ikelos: This changes the output that can be produced, but in an additive only way, so technically we should bump [VERSION_PATCH](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/constants/__init__.py#L43) (which is used to write the metadata into the JSON).

volatility3/framework/symbols/windows/pdbconv.py
```diff
@@ -755,22 +772,31 @@ def process_types(self, type_references: Dict[str, int]) -> None:
...
+        'LF_NESTTYPE': ('LF_NESTTYPE', True, None),
+        'LF_MFUNC_ID': ('LF_MFUNC_ID', True, None),
+        'LF_MFUNCTION': ('LF_MFUNCTION', False, None),
+        'LF_METHOD': ('LF_METHOD', True, None),
+        'LF_STMEMBER': ('LF_STMEMBER', True, None),
+        'LF_ONEMETHOD': ('LF_ONEMETHOD', True, None),
+        'LF_BCLASS': ('LF_BCLASS', False, None),
     }
 
+    types_to_skip = ["LF_METHODLIST"]
```
ikelos: Errr, be nice to have a comment as to what sorts of things go in this list and why we'd be skipping them.

volatility3/framework/symbols/windows/pdbconv.py
```diff
@@ -548,6 +560,11 @@ def omap_lookup(self, address):
 
     def name_strip(self, name):
         """Strips unnecessary components from the start of a symbol name."""
+        if name.startswith("?") and len(name.split("@@")) == 3:
```
ikelos, line 504: I'd want to think about whether this could create duplicates (since it's essentially going into a dictionary, so only one entry each).  We could key it off the linkage_name instead, but that's a more significant shift, but at the very least we'll need to check for name collisions and avoid overwriting one that already exists with the same name...

ikelos: (This is also in general, I know we already do some name mangling below).  5:)

ikelos (commented):
Awesome!  Thanks very much for this, the pdb conversion code is really ugly, and you're a braver man than I for wading into it (and I wrote the thing!).  5:P  Looks pretty good, just a couple of minor points and it still needs some testing, but hopefully I can get through those quickly.  5:)

volatility3/framework/symbols/windows/pdbconv.py
paulkermann: I will add a comment explaining why we skip those leafs.

volatility3/framework/symbols/windows/pdbconv.py
paulkermann: Much obliged

volatility3/framework/symbols/windows/pdbconv.py
paulkermann: I will bump the minor version.

volatility3/framework/symbols/windows/pdbconv.py
paulkermann: I will add a comment.

volatility3/framework/symbols/windows/pdbconv.py
paulkermann: So what do you think we should do in case of a duplicate?

paulkermann: If a duplicate is detected the non stripped name is used, and a warning is logged. If that warning is triggered in the future and reported by someone we can think about a different solution.

ikelos: Yeah, logging it seems like the best immediate possibility.  Might be worth doing it at warning so we get some visibility.  Since it will only happen at PDB generation we need to shout and scream (it would be worth testing how often we run into that situation, if it's not often we might want to abort so we don't produce broken ISF files)?

ikelos: Using the non-stripped name is likely to get us in trouble down the line if people start to rely on them in some instances.

ikelos: We've opted for the linkage name always being the unstripped one, because the main one will be the one that gets looked up, and we wanted to make that the one people used (although, ideally not with parameters, etc).  So let's still try to use the stripped name, but as I say, in the case of duplicates we need to see if it's common or not.  If not, then bail out, if so then shout and scream and we need to think of how best to architect it (probably with input from @npetroni )...

paulkermann: sure.

paulkermann:
Any more work needs to be done?

volatility3/framework/symbols/windows/pdbconv.py
```diff
@@ -498,8 +514,12 @@ def read_symbol_stream(self):
                 if self._omap_mapping:
                     address = self.omap_lookup(address)
                 stripped_name = self.name_strip(name)
+                if stripped_name in self.symbols:
+                    vollog.warning(f"{stripped_name} already present at symbols, non stripped name {name} will be used!")
+                    stripped_name = name
```
ikelos: This is going to confuse people, we need to be consistent, so I'd issue the warning but then overwrite it.  Having people not know whether to lookup by stripped name or not will just cause chaos.

ikelos (commented):
From your side there's just one small change, but I'm afraid we'll need a bit of time to run some tests on this, so please give us a couple of weeks and then feel free to prod us.  5:)

volatility3/framework/symbols/windows/pdbconv.py
paulkermann: Okay, code now overwrites the `stripped_name` symbol. Log changed accordingly.

ikelos: Awesome, thanks!  5:)

paulkermann:
@ikelos any progress on this?

volatility3/framework/constants/__init__.py
```diff
@@ -40,7 +40,7 @@
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 2  # Number of releases of the library with a breaking change
 VERSION_MINOR = 0  # Number of changes that only add to the interface
-VERSION_PATCH = 2  # Number of changes that do not change the interface
+VERSION_PATCH = 3  # Number of changes that do not change the interface
```
digitalisx: There was a small change in the process of merging the plugins [95a11c3](https://github.com/volatilityfoundation/volatility3/pull/689/commits/95a11c366965b03c4f69b6b75687e600d2dd231f), which resulted in version bump. If the PR is merged, we need to update the version one more time. (`2.0.4`)

paulkermann: I have updated it to version patch 4.

paulkermann:
From what I understood the main conflict in this PR is the handling of C++ symbols.
IE what should we do when there are overloaded functions(global variables can't really be overloaded without a different namespace?).
If we want to do 'real' demangeling then this code is a good reference: https://github.com/moyix/pdbparse/blob/b5b61793e4a457c43a5aef7a0499b959826b2c04/src/undname.c#L1241

It looks like windbg `u` command only accepts demangeled names. Also, if a symbol (after demangeling) appears several twice then the first one is used(the first one found I think).
I think that doing whatever windbg does (and at first defining it) is the right course of action.
With the apis described below we remain versatile with those changes:
1) `get_symbol` would by default return the first thing it found (main use case used by people)(the thing we have right now).
1) `get_exact_symbol` would return a symbol based on the mangled name.

On duplicated symbols we keep the first one found as the "main" one which would be returned from `get_symbol`.
To get a specific symbol one would need to use `get_exact_symbol`.
It is the developer responsibility to see if the symbol he accesses has several overloads and which symbol getter he needs to use.

paulkermann:
From what I understood the main conflict in this PR is the handling of C++ symbols.
IE what should we do when there are overloaded functions(global variables can't really be overloaded without a different namespace?).
If we want to do 'real' demangeling then this code is a good reference: https://github.com/moyix/pdbparse/blob/b5b61793e4a457c43a5aef7a0499b959826b2c04/src/undname.c#L1241

It looks like windbg `u` command only accepts demangeled names. Also, if a symbol (after demangeling) appears several twice then the first one is used(the first one found I think).
I think that doing whatever windbg does (and at first defining it) is the right course of action.
With the apis described below we remain versatile with those changes:
1) `get_symbol` would by default return the first thing it found (main use case used by people)(the thing we have right now).
1) `get_exact_symbol` would return a symbol based on the mangled name.

On duplicated symbols we keep the first one found as the "main" one which would be returned from `get_symbol`.
To get a specific symbol one would need to use `get_exact_symbol`.
It is the developer responsibility to see if the symbol he accesses has several overloads and which symbol getter he needs to use.

In this [link](https://lise.pnfsoftware.com/winpdb/2C5423AE810E58D725A98CB83193CB9180FECB7509367EAAA08F2FC2D89F551D-ntoskrnl.html) you can see the function `RtlStringCbCopyUnicodeString` is present twice (one C++ definition and one C definition).
I suggest that the  C++ one (`0x1403057B0`) would be returned from `get_symbol` because it has the lower address.
To get the other one you would need to use `get_exact_symbol`.

paulkermann:
Hey, any update on this? @ikelos

ikelos:
I'm really sorry, I'm still mulling over how best to handle referencing and interacting with C++ types and I haven't been able to spend time just sitting and thinking through the implications of this.  Unfortunately this isn't a high priority so the scare time I spend on volatility is directed to other issues.  Do please leave the issue open, but bear with us as we want to make sure we completely consider the implications and most maintainable means of supporting new features like this down the road...

PR #653, opened 2022-02-17
Improve slow pdb scanning

paulkermann:
By starting the pdb scan from the initial loader mappings (see [here](https://codemachine.com/articles/x64_kernel_virtual_address_space_layout.html)) we can cut the slow scanning time by a factor.

### DIFF
--- a/volatility3/framework/automagic/pdbscan.py
+++ b/volatility3/framework/automagic/pdbscan.py
@@ -192,9 +192,15 @@ class KernelPDBScanner(interfaces.automagic.AutomagicInterface):
         if not physical:
             layer_to_scan = virtual_layer_name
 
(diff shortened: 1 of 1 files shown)


ikelos:
I would like to keep the full slow scan as a backup, so perhaps parameterize `_method_layer_pdb_scan` so that after the quicker scan, the complete space can still be checked?  I dunno if that's crazy because there'd be no way a valid kernel could ever occur in a lower address, I just want to try everything so people can't say we missed it...  5:)

paulkermann:
@ikelos I have added the optimized scan to run before the original slow scan.

ikelos:
Cool, looks good, thanks!  5:D

Issue #627, opened 2022-01-18
Volatility can't process UNC path


sin4ts:
**Describe the bug**
Volatility3 can't process files which are located on network share and thus provided using UNC path.
All the more, if cached symbols are also located on a network share (for example because volatility itself is stored on a network share), they can't be loaded which cause volatility to crash.
I assume of course one has the correct permission to access the given network share. Thank's to Windows SSO, credentials are not required to be explicitly provided - i.e. one can access the network share using the corresponding UNC path, as if it was a local drive.

**Context**
Volatility Version:  1.0.0 and later
Operating System:  Windows
Python Version:  3.7
Suspected Operating System:  
Command: 
vol.py -f \\\\SHARE\memdump.mem
Or:
vol.y --single-location file://SHARE/memedump.mem
Or:
\\\\SHARE\vol.py -f C:\memdump.mem

**To Reproduce**
- Run vol.py locally with a memory image stored on a network share and provided with a UNC path
Or:
- Run vol.py from a network share with a memory image stored on the local disk

**Expected behavior**
vol.py crashes with the below stacktrace.

WARNING  volatility3.framework.plugins: Automagic exception occurred: urllib.error.URLError: <urlopen error [WinError 3] Le chemin d’accès spécifié est introuvable: '\\share\\dev\\volatility3\\volatility3\\framework\\symbols\\windows\\pdb.json'>
Level 9  volatility3.framework.plugins: Traceback (most recent call last):
  File "C:\Program Files\Python310\lib\urllib\request.py", line 1505, in open_local_file
    stats = os.stat(localfile)
FileNotFoundError: [WinError 3] Le chemin d’accès spécifié est introuvable: '\\share\\dev\\volatility3\\volatility3\\framework\\symbols\\windows\\pdb.json'

**Additional information**
The issue seems to be caused by urllib being unable to process UNC path in volatility3\framework\layers\resources.py (even  if pysmb is installed).
The bug can be mitigated by processing URL starting with file:// with the native "open" function instead of urllib.request.urlopen (with a bit of formatting to convert the URL into a valid UNC path or local file path).

**Full Stacktrace**
Volatility 3 Framework 2.0.0
INFO     volatility3.cli: Volatility plugins path: ['Z:\\dev\\volatility3\\volatility3\\plugins', 'Z:\\dev\\volatility3\\volatility3\\framework\\plugins']
[...]

PR #655, opened 2022-02-20
Add support for UNC paths

paulkermann:
I was trying to run volatility3 against UNC paths from a share and that did not work for me.

I saw this [issue](https://github.com/volatilityfoundation/volatility3/issues/627) and I think this can be merged in the meanwhile as it will work and in the future `single_location.geturl()` should be used instead of `urlunparse` when the bug is fixed.

### DIFF
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -17,6 +17,7 @@ import json
 import logging
 import os
 import sys
+import pathlib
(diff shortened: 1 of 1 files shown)


ikelos:
I've merged in #658 instead, so I'm going to close this off.  Please shout if there's a situation where the two differ and we can investigate...  5:)

PR #657, opened 2022-02-21
Fix read whole module

paulkermann:
When calling `get_guid_from_mz` the whole module is read from memory.

I have a dump where one page in the module was paged out, and therefore an exception was thrown (even though the rest of the pages are valid). With this little fix the executable parsing works correctly and the PDB is fetched.


ikelos:
I'm not generally a fan of papering over missing chunks of data, but in this instance, you're right we only need the identifying information to figure out the pdb, and if that data isn't there it'll be discovered later on, so this is fine (just justifying it for posterity).  Thanks!  5:)

PR #659, opened 2022-02-23
Objects: Don't try to read 0 bytes when unmarshalling

### LINKED ISSUE #652: Null pointer dereference in UNICODE_STRING
Ran into an issue with `UNICODE_STRING`'s `get_string` function, when `self.Buffer` is null.
`self.Buffer.dereference()` fails, and `InvalidAddressException` gets thrown.

I'm aware of commit 00f3b4fb4448cd121d472effcaaf37f3e2a20375; I'd prefer at least a more specific error in this case, rather than a page fault error (which is intended for actual invalid addresses and paged out memory).

This occurrence (`UNICODE_STRING` with NULL buffer and 0 length) is not uncommon in Windows judging by my experience.

I think the function should return `""` if `self.Length` is zero. I'd be happy to implement this change if you'd like.

ikelos:
Closes #652.

This patch fixes unicode `get_string` on 0 length by firstly, constructing the string directly rather than dereferencing to a null pointer and then building a string.  This allows for the second change, which allows construction of a string object (any primitive of 0 length) to avoid reading the layer if the string `max_length` is 0.

### DIFF
--- a/volatility3/framework/objects/__init__.py
+++ b/volatility3/framework/objects/__init__.py
@@ -142,10 +142,9 @@ class PrimitiveObject(interfaces.objects.ObjectInterface):
     def _unmarshall(cls, context: interfaces.context.ContextInterface, data_format: DataFormatInfo,
                     object_info: interfaces.objects.ObjectInformation) -> TUnion[int, float, bool, bytes, str]:
         # Don't try to lookup a 0 length data format, incase it's at an invalid offset.  Length 0 means b''
+        data = b''
         if data_format.length > 0:
             data = context.layers.read(object_info.layer_name, object_info.offset, data_format.length)
-        else:
-            data = b''
         return convert_data_to_value(data, cls._struct_type, data_format)
 
     class VolTemplateProxy(interfaces.objects.ObjectInterface.VolTemplateProxy):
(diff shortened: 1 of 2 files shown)


ikelos:
Will likely commit in a couple of days if no one shouts.

volatility3/framework/objects/__init__.py
```diff
@@ -141,7 +141,11 @@ def __getnewargs_ex__(self):
     @classmethod
     def _unmarshall(cls, context: interfaces.context.ContextInterface, data_format: DataFormatInfo,
                     object_info: interfaces.objects.ObjectInformation) -> TUnion[int, float, bool, bytes, str]:
-        data = context.layers.read(object_info.layer_name, object_info.offset, data_format.length)
+        # Don't try to lookup a 0 length data format, incase it's at an invalid offset.  Length 0 means b''
+        if data_format.length > 0:
```
paulkermann, line 146: I would do something similar to this:
```
data = b''
if data_format.length > 0:
    data = ...
```

ikelos: Thanks, yep, I get what you meant now, I've added that, thanks.  5:)

PR #662, opened 2022-02-25
Wrong assumptions about size limits

moonshadow565:


ikelos:
Thanks for the patch.  Since PE pointer sizes are 32-bit the theoretical PE size limit would be 4Gb, but we cap at 100Mb at the moment to avoid a bad memory acquisition leading to the tool potentially reconstructing several Gb files in error.  We haven't seen many binaries that go over that limit, so we'd be interested in knowing the size of the PE you were analyzing that went over this limit, whether it was a legitimate binary file, and if you know it, what program it is that carries around a binary of that size?

moonshadow565:
> Thanks for the patch. Since PE pointer sizes are 32-bit the theoretical PE size limit would be 4Gb, but we cap at 100Mb at the moment to avoid a bad memory acquisition leading to the tool potentially reconstructing several Gb files in error. We haven't seen many binaries that go over that limit, so we'd be interested in knowing the size of the PE you were analyzing that went over this limit, whether it was a legitimate binary file, and if you know it, what program it is that carries around a binary of that size?

110MB, so just bit over.
Figured 2GB is not so much memory to lose today, considering i have seen games with well over 200MB executables.

ikelos:
Well, the problem would be if someone had a bad memory image, that started saving out lots of 2Gb files because the sizes read from the image were wrong.  Pretty happy upping it to 250Mb if you've seen some at 200Mb though?  Let me know if you just want me to do that or want to update the PR...  5:)

moonshadow565:
No plans to edit PR, you can put 256MB limit if you want :)
Maybe a better option would be to allow for some sort of flag to disable or increase limits like this?

ikelos:
Thanks, I've done both as part of `1b09f20b`, so I'll close this off....

PR #664, opened 2022-02-28
Improving --config and --write-config

paulkermann:
For automation it's better to specify both `--config` and `--write-config`. Under these circumstances, an existing config is used, or will be created.
This PR generates the config file name based on `single_location`.

With this improvement you can run:
```
vol.py -f FILE_LOCATION --config --write-config windows.pslist
```
This will work without an existing config (and will write it anyways).

If you want to use the default config for the FILE_LOCATION without the --write-config you need to write
```
vol.py -f FILE_LOCATION --config="" windows.pslist
```
(Otherwise, argparse will consider `windows.pslist` as the config path)

This improvement is better for automation because the script doesn't need to keep track of whether a configuration was create already or not.

Right now I think this improvement fits the CLI usage best, but in the future this can also be implemented in volshell if requested.
Of course you can still also run the plugin without the `FILE_LOCATION` if the config is specified and config files created in the past will still work.

### DIFF
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -104,8 +104,10 @@ class CommandLine:
                 parser.prog))
         parser.add_argument("-c",
                             "--config",
+                            nargs='?',
                             help = "Load the configuration from a json file",
                             default = None,
+                            const= "",
                             type = str)
         parser.add_argument("--parallelism",
                             help = "Enables parallelism (defaults to off if no argument given)",
@@ -274,6 +276,7 @@ class CommandLine:
         # It has to go here so it can be overridden by single-location if it's defined
         # NOTE: This will *BREAK* if LayerStacker, or the automagic configuration system, changes at all
         ###
+        single_location = None
         if args.file:
             try:
                 single_location = self.location_from_file(args.file)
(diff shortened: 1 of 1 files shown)


ikelos:
I'm not keen on the concept of `config` having two linked meanings, at the moment `config` is the path to read in a config.  If we want to specify one for `--write-config` it should either have a different name or be a parameter to `--write-config`.  `--config` (being for reading configurations) should fail if the named config isn't present.

I also think that a heavyweight automation system would construct its own python harness, rather than calling the CLI direct (volatility 3 was built as a library to enable this) and, if it's not a heavyweight automation system, then a simple shell script could check for the existence of a config and run with `--config` if necessary and `--write-config` if not.  I agree it might be useful to choose the config output file on the CLI, but I think there's better solutions than overloading the use of another flag to resolve that.  I don't there's many plugins that stash much in the config, so writing a config that came from a config isn't all that useful. At the moment, the config is primarily just to reduce the automagic runtime.

paulkermann:
> I'm not keen on the concept of `config` having two linked meanings

It still has only one meaning: a path to a config file. The difference is that if the file is not present it will still work.
I do not use the path in `config` for the `--write-config`. It's just that the `--write-config` output path is now generated based on the image name.
Maybe I can add a warning that the config file was not present so it is more noticeable.

Either way the config file the `--write-config` creates used to have a constant name of `config.json` which other places overwrote (timeliner if I remember correctly and also running the cli `--write-config` with different images). By having the name generated based on `single_location` we avoid this problem.

I think this pull request will aid end users who use the CLI and that the `--config --write-config` is really convenient when you need to run the plugins via the CLI and want to reduce the automagic runtime.

paulkermann:
#676 closes this PR

PR #667, opened 2022-03-07
task_struct get all threads from the parent task.

samuelzurowski:
Hello,

I found myself wanting to go through the thread_nodes of a given process often for Linux images. Thought other people who are developing plugins or other modules may find this useful when dealing with the task_struct. The picture bellow shows what this small function does.

image

### DIFF
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -201,6 +201,24 @@ class task_struct(generic.GenericIntelProcess):
 
             yield (start, end - start)
 
+    def get_thread_nodes(self) -> Iterable[interfaces.objects.ObjectInterface]:
+        """Returns a list of the task_struct based on the list_head 
+        thread_node structure."""
+
+        task_symbol_table_name = self.get_symbol_table_name()
+
+        parent = self.group_leader
+        for task in self.thread_node.to_list(
+            f"{task_symbol_table_name}{constants.BANG}task_struct",
+            "thread_node"
+        ):
+
+            if task.group_leader != parent: continue
+
+            yield task
+
+
+
 
 class fs_struct(objects.StructType):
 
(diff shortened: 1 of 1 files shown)


samuelzurowski:
I modified this to use the same concept for volatility2 using thread_group instead (see [volatility2 threads](https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/overlays/linux/linux.py#L1909)). Eventually going to use this to create a threads plugin for volatility3. 

Also added some comments so people can understand what it is doing.

samuelzurowski:
I made a threads plugin too. Once this is sorted out ill send a PR for linux.threads I've worked on!

ikelos:
Thanks, I'll just check this with our linux plugin chap (@atcuno) that he's happy with this addition.  Ideally there'd be a way to version this bit of code, so people knew that this method would be here or not.  It may just be we have to bump the volatility framework version, but since it's an additive change, we can just bump the minor version number.

atcuno:
@ikelos This looks good to me, and I would like to get it merged ASAP so Gus' code in #600 can be adjusted to use it. In particular, the to_list approach covers smear with 'seen' and always does the try/except on referencing .next:

https://github.com/volatilityfoundation/volatility3/blob/581251cb0547b118ef01687aed81aab0a2f75f6d/volatility3/framework/symbols/linux/extensions/__init__.py#L434

Also, I see this code uses the "f" syntax but I believe our minimum Python version requirement covers this now right? I know we discussed this before, but I wanted to be extra careful.

ikelos:
Yep, f-strings came in in version 3.6, which is now our minimum, so I think we've changed a lot of the code base to use them, should be all good now.  5:)

ikelos (approved):
Ok, I'll allow this one in without versioning, because it's an additive change, but as I say, at some point we should really consider versioning these and requiring specific versions through a requirement.

PR #669, opened 2022-03-08
Fix: typo for some plugins, documents.

digitalisx:
Hello, I found a typo in the document and code, during code analysis,
so I scanned it overall and corrected it.

### DIFF
--- a/volatility3/framework/symbols/windows/mft.json
+++ b/volatility3/framework/symbols/windows/mft.json
@@ -134,7 +134,7 @@
                           "kind": "base",
                           "name": "unsigned char"
                         }
-                      }
+                    }
                 },
                 "UpdateSequenceOffset": {
                     "offset": 4,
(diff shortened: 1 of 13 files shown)


volatility3/framework/symbols/windows/mft.json
```diff
@@ -192,7 +192,7 @@
                         "name": "unsigned int"
                     }
                 },
-                "AlocatedSize": {
```
ikelos, line 195: This one might cause problems if people already rely on it.  I'd need to have a think about changing it.  Probably it's ok, but I need to consider what it might break.

ikelos: This looks to be the only place it's used, so I'll take the risk and change it.

ikelos (approved):
Thanks, this look good.  I hadn't realized I'd made so many typos!  5:P

PR #670, opened 2022-03-09
Feature: implement `MBRScanner` plugin.

digitalisx:
## Description

Among them, I'm implemented (or porting) of MBR Scanner according to the Volatility 3 structure.

It was implemented so that the same results as Volatility 2 can be obtained by referring
to the [existing code](https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/mbrparser.py) (mbrparser of Volatiltiy 2) and the [underlying knowledge](https://volatility-labs.blogspot.com/2012/10/movp-43-recovering-master-boot-records.html).

## Command
Help Command

```
python3 vol.py -h
windows.mbrscan.MBRScan Scans for and parses potential Master Boot Records (MBRs)
```

Run Command

`python3 vol.py -f case.vmem windows.mbrscan`

## Output Example

<details>
<summary>Click to open result</summary>

```
Potential MBR at Physical Offset	Disk Signature	Bootcode md5	Bootcode (FULL) md5	Partition Entries Info	Disasm	Hexdump
0xc6a75fe	3b-83-70-12	a36c5e4f47e84449ff07ed3517b43a31	a2cb9999354eaa8961d94d237866b068	

===== Partition Table #1 =====
Boot Flag: 0x80 (Bootable)
Partition Type: 0x7 (NTFS)
Starting Sector (LBA): 0x800 (2048)
Starting CHS: Cylinder: 0 Head: 32 Sector: 33
Ending CHS: Cylinder: 1023 Head: 254 Sector: 63
Size in Sectors: 0x5000000 (83886080)

===== Partition Table #2 =====
Boot Flag: 0x0 
Partition Type: 0x7 (NTFS)
Starting Sector (LBA): 0x5000800 (83888128)
Starting CHS: Cylinder: 1023 Head: 254 Sector: 63
Ending CHS: Cylinder: 1023 Head: 254 Sector: 63
Size in Sectors: 0x3c00000 (62914560)

===== Partition Table #3 =====
Boot Flag: 0x0 
Partition Type: 0xf (Microsoft Extended, LBA)
Starting Sector (LBA): 0x8c00800 (146802688)
Starting CHS: Cylinder: 1023 Head: 254 Sector: 63
Ending CHS: Cylinder: 1023 Head: 254 Sector: 63
Size in Sectors: 0x627a000 (103260160)

===== Partition Table #4 =====
Boot Flag: 0x0 
Partition Type: 0x0 (Empty)
Starting Sector (LBA): 0x0 (0)
Starting CHS: Cylinder: 0 Head: 0 Sector: 0
Ending CHS: Cylinder: 0 Head: 0 Sector: 0
Size in Sectors: 0x0 (0)
	
0x0:	xor	eax, eax
0x2:	mov	ss, eax
0x4:	mov	esp, 0xc08e7c00
0x9:	mov	ds, eax
0xb:	mov	esi, 0xbf7c00
0x10:	push	es
0x11:	mov	ecx, 0xf3fc0200
0x16:	movsb	byte ptr es:[edi], byte ptr [esi]
0x17:	push	eax
0x18:	push	0xfbcb061c
0x1d:	mov	ecx, 0xbebd0004
0x22:	pop	es
0x23:	cmp	byte ptr [esi], 0
0x27:	jl	0x34
0x29:	jne	0xc583013d
0x2f:	adc	dl, ah
0x31:	int1	
0x32:	int	0x18
0x34:	mov	byte ptr [esi], dl
0x37:	push	ebp
0x38:	mov	byte ptr [esi + 0x11], 5
0x3c:	mov	byte ptr [esi + 0x10], 0
0x40:	mov	ah, 0x41
0x42:	mov	ebx, 0x13cd55aa
0x47:	pop	ebp
0x48:	jb	0x59
0x4a:	cmp	ebx, 0x975aa55
0x50:	test	ecx, 0x3740001
0x56:	inc	byte ptr [esi + 0x10]
0x59:	pushaw	
0x5b:	cmp	byte ptr [esi + 0x10], 0
0x5f:	je	0x87
0x61:	push	0
0x65:	add	byte ptr [eax], al
0x67:	push	word ptr [esi + 8]
0x6b:	push	0x680000
0x70:	jl	0xda
0x72:	add	dword ptr [eax], eax
0x74:	push	0x42b40010
0x79:	mov	dl, byte ptr [esi]
0x7c:	mov	esi, esp
0x7e:	int	0x13
0x80:	lahf	
0x81:	add	esp, 0x10
0x84:	sahf	
0x85:	jmp	0x9b
0x87:	mov	eax, 0xbb0201
0x8c:	jl	0x18
0x8e:	push	esi
0x8f:	add	byte ptr [edx + 0x4e8a0176], cl
0x95:	add	cl, byte ptr [edx + 0x13cd036e]
0x9b:	popaw	
0x9d:	jae	0xbb
0x9f:	dec	byte ptr [esi + 0x11]
0xa2:	jne	0xb0
0xa4:	cmp	byte ptr [esi], 0x80
0xa8:	je	0x80b20138
0xae:	jmp	0x34
0xb0:	push	ebp
0xb1:	xor	ah, ah
0xb3:	mov	dl, byte ptr [esi]
0xb6:	int	0x13
0xb8:	pop	ebp
0xb9:	jmp	0x59
0xbb:	cmp	dword ptr [esi], 0xaa557dfe
0xc1:	jne	0x131
0xc3:	push	dword ptr [esi]
0xc6:	call	0x17750158
0xcb:	cli	
0xcc:	mov	al, 0xd1
0xce:	out	0x64, al
0xd0:	call	0xdfb00158
0xd5:	out	0x60, al
0xd7:	call	0xffb00158
0xdc:	out	0x64, al
0xde:	call	0xb8fb0158
0xe3:	add	byte ptr [ebx + 0x23661acd], bh
0xe9:	sal	byte ptr [ebp + 0x3b], 0x66
0xed:	cmp	ebx, 0x41504354
0xf3:	jne	0x127
0xf5:	cmp	ecx, 0x2c720102
0xfb:	push	0xbb07
0xff:	add	byte ptr [eax], al
0x101:	push	0x200
0x105:	add	byte ptr [eax], al
0x107:	push	8
0x10b:	add	byte ptr [eax], al
0x10d:	push	bx
0x10f:	push	bx
0x111:	push	bp
0x113:	push	0
0x117:	add	byte ptr [eax], al
0x119:	push	0x7c00
0x11d:	add	byte ptr [eax], al
0x11f:	popaw	
0x121:	push	0xcd070000
0x126:	sbb	bl, byte ptr [edx + 0x32]
0x129:	imul	dl
0x12b:	add	byte ptr [eax + eax], bh
0x12f:	int	0x18
0x131:	mov	al, byte ptr [0x8eb07b7]
0x136:	mov	al, byte ptr [0x3eb07b6]
0x13b:	mov	al, byte ptr [0xe43207b5]
0x140:	add	eax, 0xf08b0700
0x145:	lodsb	al, byte ptr [esi]
0x146:	cmp	al, 0
0x148:	je	0x153
0x14a:	mov	ebx, 0xeb40007
0x14f:	int	0x10
0x151:	jmp	0x145
0x153:	hlt	
0x154:	jmp	0x153
0x156:	sub	ecx, ecx
0x158:	in	al, 0x64
0x15a:	jmp	0x15c
0x15c:	and	al, 2
0x15e:	loopne	0x158
0x160:	and	al, 2
0x162:	ret	
0x163:	dec	ecx
0x164:	outsb	dx, byte ptr [esi]
0x165:	jbe	0x1c8
0x167:	insb	byte ptr es:[edi], dx
0x168:	imul	esp, dword ptr [eax + 0x70], 0x69747261
0x170:	je	0x1db
0x172:	outsd	dx, dword ptr [esi]
0x173:	outsb	dx, byte ptr [esi]
0x174:	and	byte ptr [ecx + 0x62], dh
0x178:	insb	byte ptr es:[edi], dx
0x179:	add	byte ptr gs:[ebp + 0x72], al
0x17d:	jb	0x1ee
0x17f:	jb	0x1a1
0x181:	insb	byte ptr es:[edi], dx
0x182:	outsd	dx, dword ptr [esi]
0x183:	popal	
0x184:	imul	ebp, dword ptr fs:[esi + 0x67], 0x65706f20
0x18c:	jb	0x1ef
0x18e:	je	0x1f9
0x190:	outsb	dx, byte ptr [esi]
0x191:	and	byte ptr [bp + di + 0x79], dh
0x195:	jae	0x20b
0x197:	insd	dword ptr es:[edi], dx
0x199:	add	byte ptr [ebp + 0x69], cl
0x19c:	jae	0x211
0x19e:	imul	ebp, dword ptr [esi + 0x67], 0x65706f20
0x1a5:	jb	0x208
0x1a7:	je	0x212
0x1a9:	outsb	dx, byte ptr [esi]
0x1aa:	and	byte ptr [bp + di + 0x79], dh
0x1ae:	jae	0x224
0x1b0:	insd	dword ptr es:[edi], dx
0x1b2:	add	byte ptr [eax], al
0x1b4:	add	byte ptr [ebx + 0x7b], ah	
33 c0 8e d0 bc 00 7c 8e	3.....|.
c0 8e d8 be 00 7c bf 00	.....|..
06 b9 00 02 fc f3 a4 50	.......P
68 1c 06 cb fb b9 04 00	h.......
bd be 07 80 7e 00 00 7c	....~..|
0b 0f 85 0e 01 83 c5 10	........
e2 f1 cd 18 88 56 00 55	.....V.U
c6 46 11 05 c6 46 10 00	.F...F..
b4 41 bb aa 55 cd 13 5d	.A..U..]
72 0f 81 fb 55 aa 75 09	r...U.u.
f7 c1 01 00 74 03 fe 46	....t..F
10 66 60 80 7e 10 00 74	.f`.~..t
26 66 68 00 00 00 00 66	&fh....f
ff 76 08 68 00 00 68 00	.v.h..h.
7c 68 01 00 68 10 00 b4	|h..h...
42 8a 56 00 8b f4 cd 13	B.V.....
9f 83 c4 10 9e eb 14 b8	........
01 02 bb 00 7c 8a 56 00	....|.V.
8a 76 01 8a 4e 02 8a 6e	.v..N..n
03 cd 13 66 61 73 1c fe	...fas..
4e 11 75 0c 80 7e 00 80	N.u..~..
0f 84 8a 00 b2 80 eb 84	........
55 32 e4 8a 56 00 cd 13	U2..V...
5d eb 9e 81 3e fe 7d 55	]...>.}U
aa 75 6e ff 76 00 e8 8d	.un.v...
00 75 17 fa b0 d1 e6 64	.u.....d
e8 83 00 b0 df e6 60 e8	......`.
7c 00 b0 ff e6 64 e8 75	|....d.u
00 fb b8 00 bb cd 1a 66	.......f
23 c0 75 3b 66 81 fb 54	#.u;f..T
43 50 41 75 32 81 f9 02	CPAu2...
01 72 2c 66 68 07 bb 00	.r,fh...
00 66 68 00 02 00 00 66	.fh....f
68 08 00 00 00 66 53 66	h....fSf
53 66 55 66 68 00 00 00	SfUfh...
00 66 68 00 7c 00 00 66	.fh.|..f
61 68 00 00 07 cd 1a 5a	ah.....Z
32 f6 ea 00 7c 00 00 cd	2...|...
18 a0 b7 07 eb 08 a0 b6	........
07 eb 03 a0 b5 07 32 e4	......2.
05 00 07 8b f0 ac 3c 00	......<.
74 09 bb 07 00 b4 0e cd	t.......
10 eb f2 f4 eb fd 2b c9	......+.
e4 64 eb 00 24 02 e0 f8	.d..$...
24 02 c3 49 6e 76 61 6c	$..Inval
69 64 20 70 61 72 74 69	id.parti
74 69 6f 6e 20 74 61 62	tion.tab
6c 65 00 45 72 72 6f 72	le.Error
20 6c 6f 61 64 69 6e 67	.loading
20 6f 70 65 72 61 74 69	.operati
6e 67 20 73 79 73 74 65	ng.syste
6d 00 4d 69 73 73 69 6e	m.Missin
67 20 6f 70 65 72 61 74	g.operat
69 6e 67 20 73 79 73 74	ing.syst
65 6d 00 00 00 63 7b 9a	em...c{.
```

</details>


volatility3/framework/plugins/windows/mbrscan.py
```diff
@@ -0,0 +1,109 @@
...
+                partition_table = self.context.object(partition_table_object, offset = mbr_start_offset, layer_name = layer.name)
+
+                # Extract only BootCode
+                full_mbr = layer.read(mbr_start_offset, mbr_length, pad = True)
+                bootcode = full_mbr[:bootcode_length]
+                
+                if bootcode:
+                    all_zeros = bootcode.count(b"\x00") == len(bootcode)
+
+                if not all_zeros:
```
ikelos, line 80: `all_zeros` may not have  been defined is bootcode is `None` or empty.  Either don't bother with the bootcode check, or set a default value for all_zeros.

volatility3/framework/plugins/windows/mbrscan.py
```diff
@@ -0,0 +1,109 @@
...
+                                                               filename = "mbr",
+                                                               class_types = {
+                                                                'PARTITION_TABLE': mbr.PARTITION_TABLE,
+                                                                'PARTITION_ENTRY': mbr.PARTITION_ENTRY
+                                                               })
+
+        partition_table_object = symbol_table + constants.BANG + "PARTITION_TABLE"
+        
+        # Define Signature and Data Length
+        mbr_signature = b"\x55\xAA"
```
ikelos, line 61: This is a very small signature, are there more checks that can be put in place to ensure we don't hit random sets of 0x200 bytes that start b"\x55\xAA"?  Do MBRs always appear at index 0x0 within a page, etc?

volatility3/framework/plugins/windows/mbrscan.py
```diff
@@ -0,0 +1,109 @@
...
+            
+            except exceptions.PagedInvalidAddressException:
+                pass
+
+    def run(self):
+        return renderers.TreeGrid([
+            ("Potential MBR at Physical Offset", format_hints.Hex),
+            ("Disk Signature", str),
+            ("Bootcode md5", str),
+            ("Bootcode (FULL) md5", str),
```
ikelos: This label seems inaccurate, since it's not the bootcode, but just the full MBR?

volatility3/framework/plugins/windows/mbrscan.py
```diff
@@ -0,0 +1,109 @@
...
+            ("Partition Entries Info", str),
+            ("Disasm", interfaces.renderers.Disassembly),
+            ("Hexdump", format_hints.HexBytes)
```
ikelos: If the disassembly code isn't present, it will return the hexbytes, and if it does, the bytes should be output as part of the disassembly, so I'm not certain having this column makes sense.

volatility3/framework/symbols/windows/extensions/mbr.py
```diff
@@ -0,0 +1,93 @@
...
+                self.DiskSignature[0], 
+                self.DiskSignature[1],
+                self.DiskSignature[2],
+                self.DiskSignature[3]
+        )
+
+class PARTITION_ENTRY(objects.StructType):
+
+    def set_index(self, index:int):
+        """Set Partition Entry Index."""
```
ikelos: This all looks really good, very glad of the docstrings and the type info!  5:D

volatility3/framework/plugins/windows/mbrscan.py
```diff
@@ -0,0 +1,109 @@
...
+                                            partition_table.FirstEntry,
+                                            partition_table.SecondEntry,
+                                            partition_table.ThirdEntry,
+                                            partition_table.FourthEntry
+                                        ]
+                    partition_info = "\n"
+
+                    for index, partition_entry_object in enumerate(partition_entries):
+                        partition_entry_object.set_index(index)
+                        partition_info += str(partition_entry_object)
```
ikelos: This looks like it accumulates a lot of information into a string.  This requires any libraries or programs that consume the output of this plugin to reparse all that data, we're not writing the output for human consumption, we're writing it to be consumed by anyone (programs or humans).

As such, please could you have have additional rows, something along the lines of:

```
ptable1 tabledata1 partition1 partition_attribute1
ptable1 tabledata1 partition2 partition_attribute2
ptable1 tabledata1 partition3 partition_attribute3
ptable2 tabledata2 partition1 partition_attribute1
```
This will allow humans to read the information if needed, but also allows others to build programs that can digest and use the information in its original form?  It's fine to do things like looking up types and converting them to strings, just leave each item of data as its own column, rather than jamming everything into a string with specific formatting.  We've no idea how a web interface may decide to render that line, so much better to keep the data as small and specific as possible.

ikelos (dismissed):
First off, thanks for your contribution, this looks really good!  Good use of symbol tables and enumerations, good use of docstrings and typing!  It's all pretty much great, but it crams an awful lot of data into a single string at the end, including new lines and formatting, which is great specifically for humans reading the CLI, but it limits how useful this plugin could be to other interfaces (such as web interfaces) and to other systems that want to consume specific data without having to reparse the human readable string.

I'm afraid that's a show stopper on this at the moment, but if you split out the giant partition info string into individual rows, each with individual attributes that you want to include, everything else looks great and it should be quick to merge once the changes have been made!  5:)

volatility3/framework/plugins/windows/mbrscan.py
digitalisx: As you know, MBR data consists of three areas.
Among them, the **Partition Table** depends on the user's environment/configuration,
so it is not appropriate to write as a detection feature.
**Boot Code** mostly has the same value,
but it is likely to be modified by malicious code as specified [in the article](https://volatility-labs.blogspot.com/2012/10/movp-43-recovering-master-boot-records.html). 
Considering this, relatively random sets (like False Positive) can be returned due to the short length of the **Signature**, but I think it is the best way to reduce False Negative in memory analysis.
(Actually, the same signature is used for [Volatility 2](https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/mbrparser.py#L282)).
```python
scanner = MBRScanner(needles = ['\x55\xaa'])
            for offset in scanner.scan(address_space):
                PARTITION_TABLE = obj.Object('PARTITION_TABLE', vm = address_space,
                               offset = offset)
```

volatility3/framework/plugins/windows/mbrscan.py
digitalisx: The sources of both data are the same, but I think we need both the **Disassembly** format to track the actual MBR behavior and the **Hexdump** format columns to check data such as MBR Error Messages (Invalid partition tableError loading operating system Missing operating system). [Volatility2 ](https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/mbrparser.py#L453) seems to be dealing with this optionally. What do you think about this? If I misunderstand this feedback, please tell me in more detail.
```python
if not self._config.HEX:
                disasm = self.get_disasm_text(boot_code, start + dis)
                if disasm == "" or self.code_data == None:
                    continue
                boot_code_output = "Disassembly of Bootable Code:\n{0}\n\n".format(disasm)
            else:
                hexstuff = "\n" + "\n".join(["{0:010x}  {1:<48}  {2}".format(o, h, ''.join(c)) for o, h, c in self.Hexdump(boot_code, start)])
                boot_code_output = "Bootable code: \n{0} \n\n".format(hexstuff)
```

volatility3/framework/plugins/windows/mbrscan.py
digitalisx: I fully sympathize with what you said, and I tried to modify it based on this.
The current code was organized in a simple string table, so it was easy to provide a lot of information,
but it was suggested that all the information about the four partitions be provided in separate rows because it takes resources to reuse programs and people.

So, as a result of experimenting with this, there was some inconvenience for the CLI user.
(4 Partition Entries * 10 Info) 40 informations were printed on the CLI as information about the partition table, making it difficult to distinguish information. (I think this is the limitation of the CLI that the more information it contains, the less communicative it is.)

Here are the results of considering the appropriate way to satisfy all of the various use cases. In terms of memory forensics and digital forensics, the need for MBR partition table information is more important as to which partitions exist on the system, whether they are bootable, and how large they are,
so I don't think it's necessary to provide all the information.

Therefore, why don't you yield only the three information (**BootFlag, PartitionType, SizeInSector**) and leave the code for the rest of the information to be called separately?

If I misunderstand this feedback, please tell me in more detail. 😃

volatility3/framework/plugins/windows/mbrscan.py
ikelos: Yep, fair enough, I guess we'll have to see how much the false positives outweigh the false negatives.

volatility3/framework/plugins/windows/mbrscan.py
ikelos: I had thought that the disassembly would contain the hexdump, but perhaps it doesn't.  If you think it's important that the hexdump is always in there, then I'm ok with you leaving it, but just to let you know that users without a disassembler will see the hexdump twice.

volatility3/framework/plugins/windows/mbrscan.py
ikelos: Sure, I think that's ok.  You could hide the rest of the information behind a flag like `--full` or something?  Also be aware that while the CLI output is limited, there are options such as `-r csv` or `-r pretty` that might improve the output.

volatility3/framework/plugins/windows/mbrscan.py
digitalisx: I understand what you're saying, I also think it's more effective to show only Disassembly data and reduce unnecessary elements.

volatility3/framework/plugins/windows/mbrscan.py
digitalisx: That's right. I'll update it based on user comments and new facts that can be confirmed later.

volatility3/framework/symbols/windows/extensions/mbr.py
digitalisx: Thank you for your comment, some code in the file will be changed as the part processed by Enumeration is removed!

volatility3/framework/plugins/windows/mbrscan.py
digitalisx: Thanks for the idea, I think the option you mentioned will help us overcome the limitations of the CLI and provide good data to everyone. I will add related options.

volatility3/framework/plugins/windows/mbrscan.py
frispete: I beg to differ. Check out the example display. It's highly useful to also see the hexdump, since it typically carries strings, that help to identify the source of it. E.g. a Microsoft bootloader looks different to a grub2 bootloader, but those would be hard to differentiate with the disassembly only (that not even tries to separate data from code blocks, etc..). 

I would even like to ask for an (hex) offset column for the hexdump, which would allow to logically syncronize both views.

Hiding this behind a `--full` option is fine of course.

frispete:
Thank you @Digitalisx for this fine work. Much appreciated.

While this is an old structure that is commonly replaced by UEFI nowadays, we still will see MBRs for a couple of decades, I'm sure, and we need to preserve some `applied knowledge` on them as well.

volatility3/framework/plugins/windows/mbrscan.py
digitalisx: When I converted it to CSV format according to the feedback, I got a much better result!
Rendering in CSV format seems to be very good for providing users with a large amount of data in addition to providing it in table format.

**Command**

```
python3 vol.py -f test.vmem -r csv windows.mbrscan --full > test.csv
```

**Screenshot**
스크린샷 2022-03-17 오전 2 31 05

digitalisx:
We discussed a variety of things through feedback, including the validity of signatures, the need for data to be rendered, and how to return data.
We were able to get ideas through meaningful discussions and it was very helpful to me.

I have updated all the suggestions and feedback you gave me!
If there are still any improvements or features you would like to add, please review them! 😊

digitalisx:
Thank you, @frispete 
Your comment was very helpful to me! 

I just made some improvements and migration based on what Volatility developers developed as pioneers. 😅
Anyway, I hope it will be helpful for many people who use Volatility 3.

And I will continue to make efforts to contribute greatly to the DFIR field!

ikelos (commented):
Thanks very much, this all looks good from a code perspective.  I'd just like to pass it by @iMHLv2, who handles most of our windows stuff to make sure he's happy with it.  5:)

I saw a comment you made about keeping the hex data.  If you feel it's valuable then I don't have an issue with it, I just wanted you to be aware that it may appear as a repeated column to those that don't have the disassembler modules installed.

As to adding line offsets to the hex, I assume you just mean relative values.  That would be an update to be made [here](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/cli/text_renderer.py#L27), but be aware that it will be up to other user interfaces as to how they decide to display the hex bytes.  If you would like to update that area of code, please put it in a different pull request so it can be resolved separately.  5:)

Thanks again for your awesome contribution, hopefully we'll get it merged in soon!

frispete:
Hi @ikelos 

I'm watching a couple of projects, but this one stands out because of you. It's a pleasure and a great learning experience to see, how such a big project is handled with such an ease. 

> Thanks very much, this all looks good from a code perspective. I'd just like to pass it by @iMHLv2, who handles most of
> our windows stuff to make sure he's happy with it. 5:)

To be honest, this is about almost all PC, not just windows. Even old Atari ST used that scheme (while the disasm would fail on the bootloader code of course). 

> I saw a comment you made about keeping the hex data. 

Well, that was me.

> If you feel it's valuable then I don't have an issue with it, I just wanted you to be aware that it may appear as a repeated column to those that don't have the disassembler modules installed.

Is there a way to detect that fact? Redundant output should be avoided of course, but as mentioned before and implying the disasm is available, both forms have a pretty good use cases on their own.

> As to adding line offsets to the hex, I assume you just mean relative values. That would be an update to be made [here](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/cli/text_renderer.py#L27), but be aware that it will be up to other user interfaces as to how they decide to display the hex bytes. If you would like to update that area of code, please put it in a different pull request so it can be resolved separately. 5:)

Now this one is playing in the dream league. That would be awesome and a fairly universal addition, but it is beyond the scope of this changeset.

ikelos:
Oh, thanks very much @frispete !  I try to contribute to the kind of project I'd want to volunteer for, and having contributed to a few other projects I know the things I find important.  I'm glad you're enjoying it!  5:)

True, but I wasn't sure the MBR would be loaded into memory by every operating system?  The submission was also made under the windows namespace, so I guess that might have influenced my thoughts.  5;)

Ah, sorry, it's been quite a busy week in the community, so it's a little tricky to track each of the conversations.  For wanting to keep the hex data, as I say, it's doable and I have no issue with it, I just wanted to make sure the ramifications of the decision were clear.

At the moment, you'd need to try importing the disassembler, but that's really a facet of the user interface in use, and there's no real way for the UI to talk to the plugin other than through configuration options.  As I say, I'm not against it, but the output may end up being even more hectic for some users.  5:P

Contributing isn't that scary, and we'd offer advice and commentary on any pull requests (as we're doing here), but at the very least you could file it as an issue so that if we find time and/or someone else wants to come along and contribute they can.  That way the idea won't get forgotten...  5:)

digitalisx:
Hello, For those who participated in the review.
Thank you for comments!

I wanted to respect and follow the opinions of @ikelos, who is experienced and has a direction for this project in this PR and review.

After more consideration, the issue that we now have to resolve seems to be the **processing and appropriateness of data that will be duplicated without modules installed**, rather than whether or not "Hex Dump" is provided.
(I think we have some consensus about providing data! 🙌)

The compromise is to provide Hexdump data, but also add a routine to verify the capstone module installation, such as other plugins ([handles](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/handles.py#L16), [skeleton_key_check](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/skeleton_key_check.py#L31), etc). Before improving the rendering module, I think it is the best way to minimize duplication of user data.

```python
try:
    import capstone
    has_capstone = True 
except ImportError:
    has_capstone = False # Raise error or Go --full routine
```

If you have any comments on this, please thread it. 🙂

P.S.) Can you tell me what it means to **dismissed for a review**? I've only experienced Approve or Request Change through the review, so it's hard to guess how my actions and PR status will change in the future.

ikelos:
Checking for capstone is ok, but since you don't strictly need capstone for the plugin to run, I probably wouldn't recommend doing that.  This is essentially a problem of user display, and the data that the plugin produces contains the right information.  You can do the detection of capstone, but consider if the library were being used with another user interface, having capstone installed may not affect whether the users see what you want them to see or not.  So what I'm trying to suggest is make the plugin output the data you want, and try not to tie it too strongly to the particular UI in use (I'm still hoping one day someone will come along and write a completely incredible UI that's 100 times better than the one we provide).  5:D

As for the review, I tried restarting the review to get it back to just "comment", but it left it as "changes required".  Since it's not completely done yet, I didn't want to mark it as approved, so the only option to get it back to a normal state was to dismiss it (which just means discard the results).  It's absolutely fine and has no bearing on your contribution other than to get the pull request back to a state where it doesn't demand changes in big red letters!  5:)

digitalisx:
Thank you, @ikelos.
I was surprised by your prompt reply. 😅

As you said, capstone is not a necessary plug-in.
Therefore, I agree that the import check logic is not appropriate.

As suggested by @frispete, I modified it to provide HexDump when using the `--full` option.
If we need HexDump, I think it's a little more appropriate to be included in the results of the `--full` option because it's time for analysts to check the MBR data in detail.

And I agree that over time, volatility 3 will increase the contribution of many people and provide better UI and rendering options.

I hope you like the code and decision! 😊

volatility3/framework/plugins/windows/mbrscan.py
```diff
@@ -0,0 +1,241 @@
...
+                            partition_table.FourthEntry.get_ending_sector(),
+                            format_hints.Hex(partition_table.FourthEntry.get_size_in_sectors()),
+                            interfaces.renderers.Disassembly(bootcode, 0, architecture),
+                            format_hints.HexBytes(bootcode)
+                        ))
+                else:
+                    vollog.log(constants.LOGLEVEL_VV, f"Not a valid MBR: Data all zeroed out : {format_hints.Hex(offset)}")
+            
+            except exceptions.PagedInvalidAddressException:
+                pass
```
ikelos: I'd prefer that an exception had a log at the highest level (like `LOGLEVEL_VVVV` so it only shows up when someone really wants to see it), or that a proper test be put in place if this is a thing we expect to run into often.  If it really is an exceptional circumstance I'd prefer not to just ignore the error and move on.

ikelos (commented):
Thanks very much, it's looking a lot better.  I have one more suggestion and I'll accept it in either form, but the output, certainly for `--full` is now very long.  Have you considered outputting the partitions as multiple rows, so something like:

````
Thing | Stuff | Partition Number | Partition Blorb | Partition Wurg
Blarp | Snooo | 1                | Zoop            | Furgle
Blarp | Snooo | 2                | Zweep           | Splink
```
Instead of:
```
Thing | Stuff | Part 1 Blorb | Part 1 Wurg | Part 2 Blorb | Part 2 Wurg
Blarp | Snoo  | Zoop         | Furgle      | Zweep        | Splink
```

(The data is random gibberish, but designed to show where it's repeated and where it changes)

I feel as though each row could be a partition and repeat some common data, than be from a single MBR and be extremely long and verbose...

Let me know which way you prefer and we'll go with that...  5:)

ikelos:
Thanks very much for the update, but I'm not sure I was quite clear in what I'd meant.  The exception shouldn't `continue` or `pass` without letting the user have *some* way of knowing that it's done that (such as by a `LOGLEVEL_VVVV` message, for example).

volatility3/framework/plugins/windows/mbrscan.py
```diff
@@ -0,0 +1,241 @@
...
+                            partition_table.FourthEntry.get_ending_sector(),
+                            format_hints.Hex(partition_table.FourthEntry.get_size_in_sectors()),
+                            interfaces.renderers.Disassembly(bootcode, 0, architecture),
+                            format_hints.HexBytes(bootcode)
+                        ))
+                else:
+                    vollog.log(constants.LOGLEVEL_VVVV, f"Not a valid MBR: Data all zeroed out : {format_hints.Hex(offset)}")
+            
+            except exceptions.PagedInvalidAddressException:
+                continue
```
ikelos, line 164: Sorry, that's not quite what I meant. I think the other error (data all zeroed out) can stay as `LOGLEVEL_VV`, but instead of just being `continue` or `pass` here, it should *also* log that an exception occurred (since it could occur anywhere within the large try block). If we expect that to happen, then we should guard again the specific case we expect, but if we're just trapping when anything goes wrong, then we need to tell the user when something's gone wrong. I hope that makes sense?

digitalisx:
Hello, @ikelos

Thank you for your interest and concern about this pull request.
First of all, I'm sorry! Due to my busy schedule, I was unable to respond quickly to comments after the last push. 😱

This time, I have considered two reviews and tried to improve them.

First, **a lot of the data was moved to the --full option, but I still sympathized with the complexity.**

> Thanks very much, it's looking a lot better. I have one more suggestion and I'll accept it in either form, but the output, certainly for `--full` is now very long. Have you considered outputting the partitions as multiple rows, so something like:
> 
> ```
> Thing | Stuff | Partition Number | Partition Blorb | Partition Wurg
> Blarp | Snooo | 1                | Zoop            | Furgle
> Blarp | Snooo | 2                | Zweep           | Splink
> ```
> Instead of:
> ```
> Thing | Stuff | Part 1 Blorb | Part 1 Wurg | Part 2 Blorb | Part 2 Wurg
> Blarp | Snoo  | Zoop         | Furgle      | Zweep        | Splink
> ```
> 
> (The data is random gibberish, but designed to show where it's repeated and where it changes)
> 
> I feel as though each row could be a partition and repeat some common data, than be from a single MBR and be extremely long and verbose...
> 
> Let me know which way you prefer and we'll go with that...  5:)
> 

That's a very good suggestion. 🙌 The data provided by the partition entry is the same structure, so adding an index column makes it much easier to write data on the table.
(Thanks to the data presented in the example, I was able to understand it quickly. Thank you!)

Therefore, I have made some modifications to how to field data and how to yield it. You can see the following results.

**MBRScanner for standard**
```shell
> python3 vol.py -f case.vmem -r pretty windows.mbrscan
Volatility 3 Framework 2.0.2
  | Potential MBR at Physical Offset | Disk Signature |                     Bootcode MD5 |                     Full MBR MD5 | PartitionIndex | Bootable |                            PartitionType | SectorInSize |     
* |                         0x183a8e |    00-00-18-00 | 8e1cde5dee840efc2e70e701e79f3457 | 7f9d948e15b729ffcef4fd44656d4ca1 |              1 |    False |                                    Empty |   0x4150014d |
* |                         0x183a8e |    00-00-18-00 | 8e1cde5dee840efc2e70e701e79f3457 | 7f9d948e15b729ffcef4fd44656d4ca1 |              2 |    False |                Not Defined PartitionType |     0x433af0 |
* |                         0x183a8e |    00-00-18-00 | 8e1cde5dee840efc2e70e701e79f3457 | 7f9d948e15b729ffcef4fd44656d4ca1 |              3 |    False |                Not Defined PartitionType |   0x68361330 |
* |                         0x183a8e |    00-00-18-00 | 8e1cde5dee840efc2e70e701e79f3457 | 7f9d948e15b729ffcef4fd44656d4ca1 |              4 |    False |                Not Defined PartitionType |   0x2bdea0bf |
```

**MBRScanner for full**
```shell
> python3 vol.py -f case.vmem -r pretty windows.mbrscan --full
Volatility 3 Framework 2.0.2
  | Potential MBR at Physical Offset | Disk Signature |                     Bootcode MD5 |                     Full MBR MD5 | PartitionIndex | Bootable | BootFlag |                            PartitionType | PartitionTypeRaw | StartingLBA | StartingCylinder | StartingCHS | StartingSector | EndingCylinder | EndingCHS | EndingSector | SectorInSize |                                                               Disasm |                         Bootcode
* |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |              1 |    False |     0x48 |                Not Defined PartitionType |             0x9c |  0x446af108 |              404 |         132 |              8 |            424 |        37 |           10 |   0x5a249188 |                                                                      |                                 
* |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |              2 |    False |     0x9c |                         Hidden FAT12,CHS |             0x11 |  0xb009ad47 |              162 |          36 |             49 |             72 |        94 |           20 |   0x141f7a38 |                                                                      |                                 
* |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |              3 |    False |      0x6 |                Not Defined PartitionType |             0xdb |   0x8fc181d |              203 |          34 |             15 |             68 |       104 |           36 |   0x72cffd7b |                                                                      |                                 
* |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |              4 |    False |     0xc6 |                Not Defined PartitionType |             0xc2 |  0x8639a945 |              354 |         225 |             50 |            265 |       192 |           13 |    0x9e900cd |                                                                      |                                 
  |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x0:    add     byte ptr [rbx], al | 00 43 00 33 00 39 00 7e .C.3.9.~
  |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                 0x3:    xor     eax, dword ptr [rax] | 00 31 00 2e 00 4d 00 41 .1...M.A
  |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                 0x5:    cmp     dword ptr [rax], eax | 00 4e 00 00 00 00 00 00 .N......
  |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                                    0x7:    jle     9 | 00 30 00 00 00 38 01 00 .0...8..
  |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                 0x9:    xor     dword ptr [rax], eax | 00 00 00 00 00 00 00 02 ........
  |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                0xb:    add     byte ptr cs:[rbp], cl | 00 1a 01 00 00 18 00 01 ........
  |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0xf:    add     byte ptr [r14], cl | 00 62 4a 00 00 00 00 01 .bJ.....
  |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x13:   add     byte ptr [rax], al | 00 5d e2 bb 3d d0 78 d7 .]..=.x.            
```

**The second is about exceptions and logs.**

> Thanks very much for the update, but I'm not sure I was quite clear in what I'd meant. The exception shouldn't `continue` or `pass` without letting the user have _some_ way of knowing that it's done that (such as by a `LOGLEVEL_VVVV` message, for example).

At first, I thought it was the log level for data zero and feedback on future actions, but I think I misunderstood.

I think the core of this review is the logging and action of exception handling. Handling errors and delivering clear messages to users is very important. Of course, I sympathized with this and proceeded with the appropriate process.

However, my personal concern is that other plugins and codes do not leave separate messages for exceptions or are partial. May I know your intentions for this? (As I was developing using other plugins, I think I developed the same. It would be helpful to contribute to the future if you could provide a good examples or plugins for exception handling.) 🤔

I've made improvements on the two reviews and hope they fit the facts you've suggested by you.

If there is any further improvement, please feel free to leave a thread! 😊

volatility3/framework/plugins/windows/mbrscan.py
```diff
@@ -0,0 +1,169 @@
...
+                    partition_entries = [
+                        partition_table.FirstEntry, partition_table.SecondEntry,
+                        partition_table.ThirdEntry, partition_table.FourthEntry
+                    ]
+
+                    for partition_index, partition_entry_object in enumerate(partition_entries, start=1):
+                        # Output disassembly information and bootcode for each partition entry is inefficient,
+                        # so it can only be processed in the last index.
+                        last_partition_index = len(partition_entries)
+                        bootcode_buf = bootcode if(partition_index == last_partition_index) else b""
```
ikelos: I'd consider a "disk" entry that explicitly is designed to hold information that's only relevant for the disk as a whole.  Including it at the last partition means that programs which filter results may not see a particular partition, and if the number of partitions found changes (or the partition table is empty) the data might not be displayed.

Also, when returning a field that isn't present or isn't applicable, so that when the tool is used as a library programs can figure out whether a field is actually blank or is a value that shouldn't be there, please use one of the types derived from `BaseAbsentValue` such as [NotApplicableValue](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/renderers/__init__.py#L31).  This will show up in the CLI as a `-`, but allows other tools to define how to indicate that the program couldn't, or wasn't expected, to find anything.

So the new example would be something like:

```
Disk blah | Disk wibble | Partition Index | Partition zoob | Partition Fnewp | Bootcode
Something | Whatever    | -               | -              | -               | Disassembly stuff
Something | Whatever    | 1               | Blurgh         | Whacha          | -
Something | Whatever    | 2               | Grang          | Poloom          | -
```
So there is some repetition, but the disassembly is clearly associated with the disk, and the partitions are each individually identified (if you wanted, you could make use of the TreeGrid feature to make the partition rows children of the disk row, by changing the 0 in the `yield` statement to a 1 for the children).

ikelos (commented):
Hi @Digitalisx,

Thanks very much for your patience with the suggestions I'm making.

> Thank you for your interest and concern about this pull request. First of all, I'm sorry! Due to my busy schedule, I was unable to respond quickly to comments after the last push. scream

There's absolutely no rush, take as long as you need to respond, everybody's got other work.  5:)

> Therefore, I have made some modifications to how to field data and how to yield it. You can see the following results.

Awesome, these look much more compact, thanks!  5:D  
 
> **The second is about exceptions and logs.**

> I think the core of this review is the logging and action of exception handling. Handling errors and delivering clear messages to users is very important. Of course, I sympathized with this and proceeded with the appropriate process.

Yep, you're right, it was about ensuring users understand when an exception has occurred.

> However, my personal concern is that other plugins and codes do not leave separate messages for exceptions or are partial. May I know your intentions for this? (As I was developing using other plugins, I think I developed the same. It would be helpful to contribute to the future if you could provide a good examples or plugins for exception handling.) thinking

I understand your concern.  The linux and mac plugins have a tendency to ignore errors to try to continue providing output.  It is up to you, and as I say, if the exception is an expected occurrence, then I'd prefer a specific check for it, but this is your plugin and if you feel those messages aren't helpful then you don't have to include them, they were just a recommendation based on what I noticed.  5:)  The reasoning I've provided is something I'd prefer to have in all plugins, but the plugins are written by many different people, each with their own views on how and whether these things are necessary.  There are some specific things that I'll insist on being changed before going into the code, but passing on an exception isn't really all that bad, so if you'd prefer not to have the message there, then feel free to remove it.  5:)

One thing I would say though, is that you've made it `vollog.debug` which actually shows up after just one or two `-v`s, the `log(constants.LOGLEVEL_VVVV, ...)` only starts showing much later on, and therefore reduces how likely/often users will see the error.  I'd recommend changing the level to that, so that people only see the error if they *really* want to investigate the results they've received...  5:)

Checking through the code, you're right, we're not great at logging when something goes wrong, but here are some examples of when we do:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/verinfo.py#L164
https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/skeleton_key_check.py#L83
https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/memmap.py#L47
https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/envars.py#L77
https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/handles.py#L205
https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/malfind.py#L90
 
> I've made improvements on the two reviews and hope they fit the facts you've suggested by you.

Thanks again for your work so far, the plugin's looking really good now.  With the new output there's one more point I've added but the code's looking good at this point.keep up the great work, we're nearly there (and don't feel rushed to respond, the request isn't going away until you're happy!)...  5;)

volatility3/framework/plugins/windows/mbrscan.py
digitalisx: This is very useful information!
I think it would be a better code than providing empty bytes to avoid displaying data.

digitalisx:
Hello @ikelos,

Great! I was impressed by your detailed explanation.
I'm so happy that there's still a lot to improve and learn! 🙂

First, I read your **explanation of the log and the current status**.

> I understand your concern. The linux and mac plugins have a tendency to ignore errors to try to continue providing output. It is up to you, and as I say, if the exception is an expected occurrence, then I'd prefer a specific check for it, but this is your plugin and if you feel those messages aren't helpful then you don't have to include them, they were just a recommendation based on what I noticed. 5:) The reasoning I've provided is something I'd prefer to have in all plugins, but the plugins are written by many different people, each with their own views on how and whether these things are necessary. There are some specific things that I'll insist on being changed before going into the code, but passing on an exception isn't really all that bad, so if you'd prefer not to have the message there, then feel free to remove it. 5:)

That's right. As you said, determining the importance and level of logs is very difficult because each user has a different perspective. Therefore, many projects and codes, including Volatility3, need to be improved. But I think this will also be improved one day with the contribution of many people! 🙌

As you suggested, I agree to **adjust the log format and level to "VVVV"** to better provide users with error information. (It is helpful for both users and developers to have detailed logs.) That's why I adjusted it.

However, the output error message was written simply based on the various cases and situations you suggested. If you have any better ideas, feel free to tell me!

And secondly, I **replaced the previously empty byte offer with `NotApplicableValue`**. Also, I **changed yield depth according to row attribute**. I think this will create a clearer link in the relationship between information.

The results reflected by referring to other usage examples are as follows.

**MBRScanner for standard**
```shell
> python3 vol.py -f case.vmem -r pretty windows.mbrscan
Volatility 3 Framework 2.0.2
   | Potential MBR at Physical Offset | Disk Signature |                     Bootcode MD5 |                     Full MBR MD5 | PartitionIndex | Bootable |                            PartitionType | SectorInSize |                                                               Disasm
*  |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |            N/A |      N/A |                                      N/A |          N/A |                                                                     
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x0:    add     byte ptr [rbx], al
   |                                  |                |                                  |                                  |                |          |                                          |              |                                 0x3:    xor     eax, dword ptr [rax]
   |                                  |                |                                  |                                  |                |          |                                          |              |                                 0x5:    cmp     dword ptr [rax], eax
   |                                  |                |                                  |                                  |                |          |                                          |              |                                                    0x7:    jle     9
   |                                  |                |                                  |                                  |                |          |                                          |              |                                 0x9:    xor     dword ptr [rax], eax
   |                                  |                |                                  |                                  |                |          |                                          |              |                                0xb:    add     byte ptr cs:[rbp], cl
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0xf:    add     byte ptr [r14], cl
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x13:   add     byte ptr [rax], al
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x15:   add     byte ptr [rax], al
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x17:   add     byte ptr [rax], al
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x19:   xor     byte ptr [rax], al
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x1b:   add     byte ptr [rax], al
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x1d:   cmp     byte ptr [rcx], al
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x1f:   add     byte ptr [rax], al
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x21:   add     byte ptr [rax], al
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x23:   add     byte ptr [rax], al
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x25:   add     byte ptr [rax], al
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x27:   add     al, byte ptr [rax]
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x29:   sbb     al, byte ptr [rcx]
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x2b:   add     byte ptr [rax], al
   |                                  |                |                                  |                                  |                |          |                                          |              |                                   0x2d:   sbb     byte ptr [rax], al
   |                                  |                |                                  |                                  |                |          |                                          |              |                                 0x2f:   add     dword ptr [rax], eax
** |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |              1 |    False |                Not Defined PartitionType |   0x5a249188 |                                                                  N/A
** |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |              2 |    False |                         Hidden FAT12,CHS |   0x141f7a38 |                                                                  N/A
** |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |              3 |    False |                Not Defined PartitionType |   0x8d300285 |                                                                  N/A
** |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |              4 |    False |                Not Defined PartitionType |    0x9e900cd |                                                                  N/A
```

**MBRScanner for full**
```shell
> python3 vol.py -f case.vmem -r pretty windows.mbrscan --full
Volatility 3 Framework 2.0.2
   | Potential MBR at Physical Offset | Disk Signature |                     Bootcode MD5 |                     Full MBR MD5 | PartitionIndex | Bootable | BootFlag |                            PartitionType | PartitionTypeRaw | StartingLBA | StartingCylinder | StartingCHS | StartingSector | EndingCylinder | EndingCHS | EndingSector | SectorInSize |                                                               Disasm |                         Bootcode
*  |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |            N/A |      N/A |      N/A |                                      N/A |              N/A |         N/A |              N/A |         N/A |            N/A |            N/A |       N/A |          N/A |          N/A |                                                                      |                                 
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x0:    add     byte ptr [rbx], al | 00 43 00 33 00 39 00 7e .C.3.9.~
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                 0x3:    xor     eax, dword ptr [rax] | 00 31 00 2e 00 4d 00 41 .1...M.A
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                 0x5:    cmp     dword ptr [rax], eax | 00 4e 00 00 00 00 00 00 .N......
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                                    0x7:    jle     9 | 00 30 00 00 00 38 01 00 .0...8..
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                 0x9:    xor     dword ptr [rax], eax | 00 00 00 00 00 00 00 02 ........
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                0xb:    add     byte ptr cs:[rbp], cl | 00 1a 01 00 00 18 00 01 ........
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0xf:    add     byte ptr [r14], cl | 00 62 4a 00 00 00 00 01 .bJ.....
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x13:   add     byte ptr [rax], al | 00 5d e2 bb 3d d0 78 d7 .]..=.x.
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x15:   add     byte ptr [rax], al | 01 5d e2 bb 3d d0 78 d7 .]..=.x.
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x17:   add     byte ptr [rax], al | 01 5d e2 bb 3d d0 78 d7 .]..=.x.
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x19:   xor     byte ptr [rax], al | 01 5d e2 bb 3d d0 78 d7 .]..=.x.
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x1b:   add     byte ptr [rax], al | 01 00 00 00 00 00 00 00 ........
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x1d:   cmp     byte ptr [rcx], al | 00 00 00 00 00 00 00 00 ........
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x1f:   add     byte ptr [rax], al | 00 20 00 00 00 00 00 00 ........
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x21:   add     byte ptr [rax], al | 00 6c 01 61 00 6d 00 64 .l.a.m.d
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x23:   add     byte ptr [rax], al | 00 36 00 34 00 5f 00 6d .6.4._.m
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x25:   add     byte ptr [rax], al | 00 69 00 63 00 72 00 6f .i.c.r.o
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x27:   add     al, byte ptr [rax] | 00 73 00 6f 00 66 00 74 .s.o.f.t
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x29:   sbb     al, byte ptr [rcx] | 00 2d 00 77 00 69 00 6e .-.w.i.n
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x2b:   add     byte ptr [rax], al | 00 64 00 6f 00 77 00 73 .d.o.w.s
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                   0x2d:   sbb     byte ptr [rax], al | 00 2d 00 68 00 2e 00 2e .-.h....
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                 0x2f:   add     dword ptr [rax], eax | 00 6f 00 79 00 6d 00 65 .o.y.m.e
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                                                      | 00 6e 00 74 00 2d 00 6c .n.t.-.l
   |                                  |                |                                  |                                  |                |          |          |                                          |                  |             |                  |             |                |                |           |              |              |                                                                      | 00 61 00 6e 00 67 00 75 .a.n.g.u                
** |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |              1 |    False |     0x48 |                Not Defined PartitionType |             0x9c |  0x446af108 |              404 |         132 |              8 |            424 |        37 |           10 |   0x5a249188 |                                                                  N/A |                              N/A
** |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |              2 |    False |     0x9c |                         Hidden FAT12,CHS |             0x11 |  0xb009ad47 |              162 |          36 |             49 |             72 |        94 |           20 |   0x141f7a38 |                                                                  N/A |                              N/A
** |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |              3 |    False |      0x6 |                Not Defined PartitionType |             0xdb |   0x8fc181d |              203 |          34 |             15 |             68 |       104 |           36 |   0x8d300285 |                                                                  N/A |                              N/A
** |                         0x122ef5 |    61-da-fc-10 | 5601ec3b98a67fa30de216489b7e419f | 1846159e201da943dafa60ff7e2d46b2 |              4 |    False |     0xc6 |                Not Defined PartitionType |             0xc2 |  0x8639a945 |              354 |         225 |             50 |            265 |       192 |           13 |    0x9e900cd |                                                                  N/A |                              N/A
```
Thank you very much for feeling that most of the reviews respect my thoughts and choices!
If you have any feedback or suggestions regarding this change, please feel free to leave a thread. 😊

ikelos (approved):
Awesome, this looks really good!  Thanks for sticking with me through all the changes, hopefully you're as happy with the results as I am...  5:)

PR #675, opened 2022-03-15
Windows Malfind is_vad_empty error

paulkermann:
The `is_vad_empty` all_zero_page is `str` and not `bytes` as it should be.
Therefore `is_vad_empty` will always return `False`

This fixes that issue.


ikelos (approved):
Good catch, thanks!  5:D

PR #676, opened 2022-03-16
Implement `--save-config <filename>` option to replace `--write-config`

ikelos:
This obsoletes #664.

This adds the ability to specify a filename when writing a configuration.  In order to not confuse users too much, it adds a `--save-config` and changes `--write-config` to be equivalent to `--save-config config.json`.  I've leaned towards safety, because I was always a little uneasy as how the previous implementation would blat over a file (seems a bad idea for a forensics tool).  I have however made it a configuration option so that an installation can be configured not to produce endless backups if the configs are used in a feedback loop.

Not sure if we should set this to False (overwrite files) by default?

### DIFF
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -324,18 +324,13 @@ class CommandLine:
             constructed = plugins.construct_plugin(ctx, automagics, plugin, base_config_path, progress_callback,
                                                    self.file_handler_class_factory())
 
-            backup_filename = constants.BACKUP_EXISTING_CONFIG_OUTPUT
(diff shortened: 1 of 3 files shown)


ikelos:
Based on conversations with the team we've decided to fail on an attempt to overwrite a configuration file if it already exists.  This happens after the automagic runs, so hopefully won't take too much time if it goes wrong.  A simple shell script can be used to get the repetitive functionality wanted by pull request #664, to copy the written config back over the input config.  This saves the possibility of the run-away creation of backup files whilst still fulfilling the requested use case.

PR #677, opened 2022-03-16
Remove: Windows Symbol Initialize duplicate import.

digitalisx:
It seems like you are already importing objects for pool below in a different format.
I have removed the duplicate (unused) code.


ikelos (approved):
Thanks, this looks good!  5:)

PR #683, opened 2022-03-19
Fix: typo for `Disassembly` rendering code comment.

digitalisx:
I found typo error while looking at the rendering module!


ikelos (approved):
Hehehe, thanks for spotting that.  5:)

PR #684, opened 2022-03-21
Fix: operating system comparison syntax for create cache path constant.

digitalisx:
While analyzing the Constant code in Framework, I found that there was a problem with the comparison syntax for create cache path constant. The possible values from the **sys.platform** are limited as [follows](https://docs.python.org/3/library/sys.html#sys.platform):
In order for the code to work according to its intentions, it should be compared to **"win32"** instead of **"windows"**.

### DIFF
--- a/volatility3/framework/constants/__init__.py
+++ b/volatility3/framework/constants/__init__.py
@@ -11,7 +11,6 @@ import os.path
 import sys
 from typing import Optional, Callable
 
-import volatility3.framework.constants.linux
 import volatility3.framework.constants.windows
 
 PLUGINS_PATH = [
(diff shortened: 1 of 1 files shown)


volatility3/framework/constants/__init__.py
```diff
@@ -11,7 +11,6 @@
 import sys
 from typing import Optional, Callable
 
-import volatility3.framework.constants.linux
```
ikelos: Hmmmm, I think this is intended to ensure that the constants.linux namespace is available when just `volatility3.framework.constants` is imported.  Please could you re-instate it?

volatility3/framework/constants/__init__.py
```diff
@@ -63,7 +62,7 @@
 CACHE_PATH = os.path.join(os.path.expanduser("~"), ".cache", "volatility3")
 """Default path to store cached data"""
 
-if sys.platform == 'windows':
+if sys.platform == 'win32':
```
ikelos, line 66: Ah, good catch, thanks!  I think I always make that mistake, thanks!  5:)

ikelos (commented):
Looks good, but it removes an extra line unnecessarily, which I'd rather have stay where it is.  Once that's fixed I'll be happy to merge it, thanks!  5:)

volatility3/framework/constants/__init__.py
digitalisx: Thank you, when I checked again, it was a meaningful import. 😅

digitalisx:
Thank you for your review, I have recovered the code you requested! 🙌

ikelos (approved):
Thanks very much, looks good!  5:D

PR #685, opened 2022-03-23
Fix: `__init__` and description of mac environment plugins.

digitalisx:
The plugin description for the mac environment is incorrect or the contents of the init file are not present.
So I added this. 🙂

### DIFF
--- a/volatility3/framework/plugins/mac/__init__.py
+++ b/volatility3/framework/plugins/mac/__init__.py
@@ -1,0 +1,8 @@
+# This file is Copyright 2022 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+"""All core mac plugins.
+
+These modules should only be imported from volatility3.plugins NOT
+volatility3.framework.plugins
+"""
--- a/volatility3/framework/plugins/mac/ifconfig.py
+++ b/volatility3/framework/plugins/mac/ifconfig.py
@@ -9,7 +9,7 @@ from volatility3.framework.symbols import mac
 
 
 class Ifconfig(plugins.PluginInterface):
-    """Lists loaded kernel modules"""
+    """ Lists network interface information for all devices """
 
     _required_framework_version = (2, 0, 0)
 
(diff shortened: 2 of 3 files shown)


volatility3/framework/plugins/mac/ifconfig.py
```diff
@@ -9,7 +9,7 @@
 
 
 class Ifconfig(plugins.PluginInterface):
-    """Lists loaded kernel modules"""
+    """ Lists network interface information for all devices """
```
ikelos: There's now a space at the start of this comment, might be best to just launch right into the comment?

volatility3/framework/plugins/mac/mount.py
```diff
@@ -12,7 +12,7 @@
 
 class Mount(plugins.PluginInterface):
     """A module containing a collection of plugins that produce data typically
-    foundin Mac's mount command"""
+    founding Mac's mount command"""
```
ikelos: Hehehe, you did indeed find a typo, but what we'd been meaning to write was `found in` rather than `founding`.  5:)

volatility3/framework/plugins/mac/__init__.py
```diff
@@ -0,0 +1,8 @@
+# This file is Copyright 2022 Volatility Foundation and licensed under the Volatility Software License 1.0
```
ikelos, line 1: The `__init__.py` file _shouldn't_ be necessary, because we import all python files under the various plugins directories, but I'm happy to have it for symmetry with the other plugins directories.  Thanks!  5:D

ikelos (commented):
Thanks very much for fixing these things up, you've got a great eye for detail!  5:)

Unfortunately a typo or two slipped into the fix, but once they're corrected I'll be happy to merge it.  5:)

volatility3/framework/plugins/mac/ifconfig.py
digitalisx: As I referred to the description of volatility 2, it seems to have been moved as it is. I modified it!

digitalisx:
Thank you for your review!
If my English was better, it would have been a clean PR. 😅
The code has been updated according to your reviews.

volatility3/framework/plugins/mac/mount.py
digitalisx: Oh, thank you. I think my English was not good enough. 😅

ikelos (approved):
Hehehe, that's ok, we appreciate any contributions particularly if you're happy for us to take the time to help get it right.  5:)  Looks great now, thanks!  5:D

Issue #118, opened 2019-10-31
Missing Plugin -> threads


mrratherford:
As of volatility version 2 there were the three plugins: threads, timers and devicetree. I didn't find these plugins in the new version. Do you have plans to bring them back in or are these plugins by community developers?

mrratherford:
Thanks for your quick reply!

digitalisx:
Hello @mrratherford,
This time, I developed DeviceTree Plugin, one of the requests covered in this issue.
It was finally merged in the `develop` branch, so I would appreciate it if you could give me feedback after using it. 🙌

superponible:
@ikelos I submitted a PR for this.  

Some work based on original implementation at https://github.com/digitalisx/volatility3/tree/feature/timers. 

The byte manipulation functions from https://github.com/volatilityfoundation/volatility3/pull/718/files#diff-e13d09e5ccbff35e090d1c91ba6ba0d3d59548f89b38ce97025d2e203b130975 were brought over.  I put them in utility.py to be available for both if that other PR gets merged. Not sure if that's the right place.

ikelos:
> The byte manipulation functions from https://github.com/volatilityfoundation/volatility3/pull/718/files#diff-e13d09e5ccbff35e090d1c91ba6ba0d3d59548f89b38ce97025d2e203b130975 were brought over. I put them in utility.py to be available for both if that other PR gets merged. Not sure if that's the right place.

Sounds good, but I'll take a look during the review.  Thanks for the heads-up about the prior code, that'll save me scratching my head trying to recall!  5;D

digitalisx:
`Timers` plugin has been implemented well, I can happily close the existing draft PR. :)
Thanks @superponible

PR #689, opened 2022-03-27
Feature: implement `DeviceTree` plugin.

digitalisx:
## Description

Hello, everyone in the community! 😃
After reviewing this  #118, I found that DeviceTree plugin has not yet migrated to 3.
So I'm implemented (or porting) of DeviceTree plugin according to the Volatility 3 structure.

It was implemented so that the same results as Volatility 2 can be obtained by referring to the [existing code](https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/malware/devicetree.py).

## Command
Help Command

```
> python3 vol.py -h
windows.devicetree.DeviceTree Listing tree based on drivers and attached devices in a particular windows memory image.
```

Run Command

`python3 vol.py -f case.vmem windows.devicetree`

## Output Example
```shell
Volatility 3 Framework 2.0.2
Offset  Type    DriverName      DeviceName      DeviceType
0x97065bd32e20  DRV     HDAudBus        N/A     N/A
* 0x97065bd32e20        DEV     HDAudBus        00000070        FILE_DEVICE_SOUND
** 0x97065bd32e20       ATT     HDAudBus        00000071 - \Driver\HdAudAddService      FILE_DEVICE_KS
*** 0x97065bd32e20      ATT     HDAudBus        00000072 - \Driver\ksthunk      FILE_DEVICE_KS
0x97065bd36060  DRV     usbehci N/A     N/A
* 0x97065bd36060        DEV     usbehci USBPDO-2        FILE_DEVICE_BUS_EXTENDER
** 0x97065bd36060       ATT     usbehci 0000006f - \Driver\usbhub       UNKNOWN
0x97065bd369c0  DRV     gencounter      N/A     N/A
* 0x97065bd369c0        DEV     gencounter      VmGenerationCounter     FILE_DEVICE_UNKNOWN
0x97065bd5a300  DRV     e1i65x64        N/A     N/A
* 0x97065bd5a300        DEV     e1i65x64        INTELPRO_{2CF4C691-FB21-4D9A-8CEA-F5AD9232FC61} FILE_DEVICE_NETWORK
```

### DIFF
--- a/volatility3/framework/plugins/windows/devicetree.py
+++ b/volatility3/framework/plugins/windows/devicetree.py
@@ -140,5 +140,9 @@ class DeviceTree(interfaces.plugins.PluginInterface):
 
     def run(self) -> renderers.TreeGrid:
         return renderers.TreeGrid([
-            ("Offset", format_hints.Hex), ("Type", str), ("DriverName", str), ("DeviceName", str), ("DeviceType", str),
+            ("Offset", format_hints.Hex),
+            ("Type", str),
+            ("DriverName", str),
+            ("DeviceName", str),
+            ("DeviceType", str),
         ], self._generator())
--- a/volatility3/framework/plugins/windows/devicetree.py
+++ b/volatility3/framework/plugins/windows/devicetree.py
@@ -94,19 +94,31 @@ class DeviceTree(interfaces.plugins.PluginInterface):
         # Scan the Layer for drivers
         for driver in driverscan.DriverScan.scan_drivers(self.context, kernel.layer_name, kernel.symbol_table_name):
             try:
-                driver_name = driver.get_driver_name()
+                try:
+                    driver_name = driver.get_driver_name()
+                except (ValueError, exceptions.PagedInvalidAddressException):
+                    vollog.log(constants.LOGLEVEL_VVVV,
+                        f"Failed to get Driver name : {driver.vol.offset:x}")
+                    driver_name = renderers.UnparsableValue()
 
                 yield (0, (
                     format_hints.Hex(driver.vol.offset),
                     "DRV",
                     driver_name,
                     renderers.NotApplicableValue(),
+                    renderers.NotApplicableValue(),
                     renderers.NotApplicableValue()
                 ))
 
                 # Scan to get the device information of driver.
                 for device in driver.get_devices():
-                    device_name = device.get_device_name()
+                    try:
+                        device_name = device.get_device_name()
+                    except (ValueError, exceptions.PagedInvalidAddressException):
+                        vollog.log(constants.LOGLEVEL_VVVV,
+                            f"Failed to get Device name : {device.vol.offset:x}")
+                        device_name = renderers.UnparsableValue()
+                    
                     device_type = DEVICE_CODES.get(device.DeviceType, "UNKNOWN")
 
                     yield (1, (
@@ -114,29 +126,35 @@ class DeviceTree(interfaces.plugins.PluginInterface):
                         "DEV",
                         driver_name,
                         device_name,
+                        renderers.NotApplicableValue(),
                         device_type
                     ))
                     
                     # Scan to get the attached devices information of device.
                     for level, attached_device in enumerate(device.get_attached_devices(), start=2):
-                        device_name = attached_device.get_device_name()
+                        try:
+                            device_name = attached_device.get_device_name()
+                        except (ValueError, exceptions.PagedInvalidAddressException):
+                            vollog.log(constants.LOGLEVEL_VVVV,
+                                f"Failed to get Attached Device Name: {attached_device.vol.offset:x}")
+                            device_name = renderers.UnparsableValue()
                         
-                        name = "{} - {}".format(device_name, attached_device.DriverObject.DriverName.get_string())
-                        
+                        attached_device_driver_name = attached_device.DriverObject.DriverName.get_string()
                         attached_device_type = DEVICE_CODES.get(attached_device.DeviceType, "UNKNOWN")
 
                         yield (level, (
                             format_hints.Hex(driver.vol.offset),
                             "ATT",
                             driver_name,
-                            name,
+                            device_name,
+                            attached_device_driver_name,
                             attached_device_type
                         ))
                     
             except(exceptions.PagedInvalidAddressException):
-                vollog.log(constants.LOGLEVEL_VVVV, f"Invalid address identified in drivers and devices: {format_hints.Hex(driver.vol.offset)}")
+                vollog.log(constants.LOGLEVEL_VVVV,
+                    f"Invalid address identified in drivers and devices: {driver.vol.offset:x}")
                 continue
-
 
     def run(self) -> renderers.TreeGrid:
         return renderers.TreeGrid([
@@ -144,5 +162,6 @@ class DeviceTree(interfaces.plugins.PluginInterface):
             ("Type", str),
             ("DriverName", str),
             ("DeviceName", str),
+            ("DriverNameOfAttDevice", str),
             ("DeviceType", str),
         ], self._generator())
--- a/volatility3/framework/plugins/windows/devicetree.py
+++ b/volatility3/framework/plugins/windows/devicetree.py
@@ -77,7 +77,7 @@ vollog = logging.getLogger(__name__)
 class DeviceTree(interfaces.plugins.PluginInterface):
     """Listing tree based on drivers and attached devices in a particular windows memory image."""
 
-    _required_framework_version = (2, 0, 1)
+    _required_framework_version = (2, 0, 3)
     _version = (1, 0, 0)
 
     @classmethod
@@ -121,8 +121,7 @@ class DeviceTree(interfaces.plugins.PluginInterface):
                     for level, attached_device in enumerate(device.get_attached_devices(), start=2):
                         device_name = attached_device.get_device_name()
                         
-                        attached_device_name = "Unparsable Value" if isinstance(device_name, renderers.UnparsableValue) else device_name
-                        name = "{} - {}".format(attached_device_name, attached_device.DriverObject.DriverName.get_string())
+                        name = "{} - {}".format(device_name, attached_device.DriverObject.DriverName.get_string())
                         
                         attached_device_type = DEVICE_CODES.get(attached_device.DeviceType, "UNKNOWN")
 
--- a/volatility3/framework/plugins/windows/devicetree.py
+++ b/volatility3/framework/plugins/windows/devicetree.py
@@ -1,0 +1,145 @@
+# This file is Copyright 2022 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import logging
+
+from typing import Iterator, List, Tuple
+
+from volatility3.framework import constants, renderers, exceptions, interfaces
+from volatility3.framework.configuration import requirements
+from volatility3.framework.renderers import format_hints
+from volatility3.plugins.windows import driverscan
+
+DEVICE_CODES = {
+    0x00000027 : "FILE_DEVICE_8042_PORT",
+    0x00000032 : "FILE_DEVICE_ACPI",
+    0x00000029 : "FILE_DEVICE_BATTERY",
+    0x00000001 : "FILE_DEVICE_BEEP",
+    0x0000002a : "FILE_DEVICE_BUS_EXTENDER",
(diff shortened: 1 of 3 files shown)


volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -351,17 +351,38 @@ def dereference(self) -> interfaces.objects.ObjectInterface:
...
-
+    def get_device_name(self) -> Union[str, interfaces.renderers.BaseAbsentValue]:
+        """Get device's name from the object header."""
+        try:
+            header = self.get_object_header()
+            return header.NameInfo.Name.String  # type: ignore
+        except(ValueError):
+            return renderers.UnparsableValue()
+
+    def get_attached_devices(self) -> interfaces.objects.ObjectInterface:
```
ikelos: This is an addition to the core extensions, as such we should be bumping the `VERSION_PATCH` number, and the `_required_framework_version` should be incremented to require that version at a minimum.  All other plugins should be fine since the change is only additive.

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -351,17 +351,38 @@ def dereference(self) -> interfaces.objects.ObjectInterface:
 class DEVICE_OBJECT(objects.StructType, pool.ExecutiveObject):
     """A class for kernel device objects."""
 
-    def get_device_name(self) -> str:
-        header = self.get_object_header()
-        return header.NameInfo.Name.String  # type: ignore
-
+    def get_device_name(self) -> Union[str, interfaces.renderers.BaseAbsentValue]:
```
ikelos: I understand the reason behind this change, but we very explicitly allow exceptions to be raise up until the point of the plugin, so that the plugin can decide when/whether to catch the exception, rather than the framework.

Also, we only tend to use `BaseAbsentValues` within the plugin, because otherwise trying to use this in the core code means people will need to interpret results in some cases, None for other and now BaseAbsentValues in this case.  As it is, it's not a big deal because this should be throwing an exception, but for other situations returning a `BaseAbsentValue` should only really be done by plugin specific code (so not even shared `classmethod`s should be doing it).  It's strictly for rendering and results, not for internal use inside the code.  It will add more exception handling in other places, but at least then it all happens explicitly and everyone can see what gets caught, when and where...

volatility3/framework/plugins/windows/devicetree.py
```diff
@@ -0,0 +1,145 @@
...
+from volatility3.framework import constants, renderers, exceptions, interfaces
+from volatility3.framework.configuration import requirements
+from volatility3.framework.renderers import format_hints
+from volatility3.plugins.windows import driverscan
+
+DEVICE_CODES = {
+    0x00000027 : "FILE_DEVICE_8042_PORT",
+    0x00000032 : "FILE_DEVICE_ACPI",
+    0x00000029 : "FILE_DEVICE_BATTERY",
+    0x00000001 : "FILE_DEVICE_BEEP",
```
digitalisx, line 18: In general, I think it is appropriate for a structured data blueprint or a specific value enumeration to be a JSON-type Symbol. That way, developers can focus on code flow and logic.
However, in this case, I only need the Device Code, so I inserted it into the code, but it doesn't seem to be my favorite result. If you have any good ideas or cases, please let me know! 🤔

volatility3/framework/plugins/windows/devicetree.py
```diff
@@ -0,0 +1,145 @@
...
+                        driver_name,
+                        device_name,
+                        device_type
+                    ))
+                    
+                    # Scan to get the attached devices information of device.
+                    for level, attached_device in enumerate(device.get_attached_devices(), start=2):
+                        device_name = attached_device.get_device_name()
+                        
+                        attached_device_name = "Unparsable Value" if isinstance(device_name, renderers.UnparsableValue) else device_name
```
ikelos: This line is unnecessary, the plugin should return an `UnparsableValue` rather than converting it into a string.  It's up to the UI how to display that.  For the CLI it's represented by a `-`, but for a web GUI it might be something entirely different.  Making that choice within the plugin takes the possibility away from the UI and isn't returning data, but is returning interpreted results instead.

Yield a row should allow a descendant of `BaseAbsentValue` in any field, so there's no need to make it a string.

ikelos: This line is unnecessary, the plugin should return an `UnparsableValue` rather than converting it into a string.  It's up to the UI how to display that.  For the CLI it's represented by a `-`, but for a web GUI it might be something entirely different.  Making that choice within the plugin takes the possibility away from the UI and isn't returning data, but is returning interpreted results instead.

Yield a row should allow a descendant of `BaseAbsentValue` in any field, so there's no need to make it a string.

volatility3/framework/plugins/windows/devicetree.py
```diff
@@ -0,0 +1,145 @@
...
+                    "DRV",
+                    driver_name,
+                    renderers.NotApplicableValue(),
+                    renderers.NotApplicableValue()
+                ))
+
+                # Scan to get the device information of driver.
+                for device in driver.get_devices():
+                    device_name = device.get_device_name()
+                    device_type = DEVICE_CODES.get(device.DeviceType, "UNKNOWN")
```
ikelos, line 122: This is ok because we really do want to say that the value is unknown, not that we couldn't get it for some reason.

volatility3/framework/plugins/windows/devicetree.py
```diff
@@ -0,0 +1,145 @@
...
+                        yield (level, (
+                            format_hints.Hex(driver.vol.offset),
+                            "ATT",
+                            driver_name,
+                            name,
+                            attached_device_type
+                        ))
+                    
+            except(exceptions.PagedInvalidAddressException):
+                vollog.log(constants.LOGLEVEL_VVVV, f"Invalid address identified in drivers and devices: {format_hints.Hex(driver.vol.offset)}")
```
ikelos: This is the right thing to do, nice work!  5:D

volatility3/framework/plugins/windows/devicetree.py
digitalisx: I had `extensions/__init__` return `UnparasableValue` if I could not get a specific data value for the drivers and devices. However, in this case, there was a phenomenon that the string formatting went into the object format.
```
<volatility3.framework.renderers.UnparsableValue object at 0x104efce80> - \Driver\USBHUB3
```
Therefore, I added the following logic by referring to the netscan plugin.
This requires verification whether the right logic. 😅

volatility3/framework/plugins/windows/devicetree.py
```diff
@@ -0,0 +1,145 @@
...
+import logging
+
+from typing import Iterator, List, Tuple
+
+from volatility3.framework import constants, renderers, exceptions, interfaces
+from volatility3.framework.configuration import requirements
+from volatility3.framework.renderers import format_hints
+from volatility3.plugins.windows import driverscan
+
+DEVICE_CODES = {
```
ikelos, line 14: It might be a bit more work, and I'm ok with leaving this as it is for the moment, but it might be more useful to others to add this to an ISF JSON file as an enumeration?  Ideally it could be tied directly in to the `DEVICE_OBJECT` but I think overriding a single member of a structure isn't really feasible (and probably isn't worth it).

At the moment the list is short enough that I'm ok having it embedded in this file.

volatility3/framework/symbols/windows/extensions/__init__.py
ikelos: This also actually returns a `Generator[ObjectInterface, None, None]` I think, rather than just a single `ObjectInterface`?

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -351,17 +351,38 @@ def dereference(self) -> interfaces.objects.ObjectInterface:
...
-        return header.NameInfo.Name.String  # type: ignore
+    def get_driver_name(self) -> Union[str, interfaces.renderers.BaseAbsentValue]:
+        """Get driver's name from the object header."""
+        try:
+            header = self.get_object_header()
+            return header.NameInfo.Name.String  # type: ignore
+        except(ValueError):
+            return renderers.UnparsableValue()
+
+    def get_devices(self) -> interfaces.objects.ObjectInterface:
```
ikelos: Same again here with the `Generator[ObjectInterface, None, None]`.

ikelos (commented):
Awesome, this looks really good, thanks!  5:)  There's a few little bits and pieces that need tidying and a couple of bigger things that we might think about doing in the future, but for now if we get the small stuff cleaned up I'd be happy to merge it...  5:)

volatility3/framework/plugins/windows/devicetree.py
digitalisx: I will solve this conversations with the under review!

volatility3/framework/plugins/windows/devicetree.py
digitalisx: I will solve this conversations with the under review!

volatility3/framework/plugins/windows/devicetree.py
digitalisx: That's right, if I could expand `_DEVICE_OBJECT` Symbol, I would have considered that option, but I put it in the code because I thought it was impossible with the current method.
If there are any changes or updates in the future, I will try to improve it.

volatility3/framework/plugins/windows/devicetree.py
digitalisx: This review overlaps with the above, so we will proceed with Resolve for this case.

volatility3/framework/symbols/windows/extensions/__init__.py
digitalisx: Yes, I will check this and change it.

volatility3/framework/symbols/windows/extensions/__init__.py
digitalisx: Yes, I will check this and change it. According to the policy, it will be updated to `2.0.3`. I will change the minimum version of DeviceTree to the same.

volatility3/framework/plugins/windows/devicetree.py
ikelos: I see, that's fine, but it's the reason we don't use `BaseAbsentValues` until just before they're used.  This method should be returning None and then the check becomes for whether it's None or not (you can keep the "Unparsable value" if you want, rather than turning the entire thing into an unparsable value, but again, I'd suggest that each of these is a different item of data, and so should make up two columns rather than one in the output (and then the conversion wouldn't need to happen).

volatility3/framework/plugins/windows/devicetree.py
ikelos: Yep, I don't have any good ideas, it would be quite complex to put it in a JSON ISF file just for the sake of an enum lookup, so as I say, in this instance I think you've made the right choice.  5:)

volatility3/framework/symbols/windows/extensions/__init__.py
ikelos: Yep, perfect, thanks.  5:)

ikelos (commented):
Some additional comments, generally I think we're both in agreement on everything (and I misunderstood what was going on in one part).  5:)

volatility3/framework/plugins/windows/devicetree.py
digitalisx: That's right, as I worked on the feedback from the previous review, I thought it would be better to show the data in a separate column rather than merge the text. So if you are working on this and request a review again, please check it out! :)

digitalisx: It seems that the meaning was conveyed well in the review below. Therefore, I will process this review as Resolve.

volatility3/framework/symbols/windows/extensions/__init__.py
digitalisx: I understand what you are saying.
I felt that it was difficult to process errors even during actual work.
This method will return only pure values or objects, and errors that occur when the value is retrieved will be handled by the `_generator`. (There is a little bit more syntax for `try` and `except`. 😅)

digitalisx:
Hello @ikelos,
Thank you for your review! 😊

I thought about the feedback you gave me and changed the code through various experiments.
Some codes have been modified directly for feedback, but others have probably been changed indirectly.
I hope you will like my new proposal.

If you have any comments on this PR, please feel free to leave a thread! 🙌

ikelos (approved):
Thanks very much for all the changes, this looks good!  5:D

PR #694, opened 2022-03-30
Crashdump Writer plugin

paulkermann:
This PR furthers the work done in #472, to correctly create a crashdump (properly loaded in Windbg).
Importantly, this also allows parsing encoded KDBG structs in `framework/plugins/windows/info.py`.


volatility3/framework/plugins/windows/info.py
```diff
@@ -7,11 +7,28 @@
 
 from volatility3.framework import constants, interfaces, layers, symbols
 from volatility3.framework.configuration import requirements
-from volatility3.framework.interfaces import plugins
+from volatility3.framework.interfaces import plugins, configuration
 from volatility3.framework.renderers import TreeGrid
 from volatility3.framework.symbols import intermed
 from volatility3.framework.symbols.windows import extensions
+from volatility3.framework.layers import physical
+import struct
```
digitalisx: I recommend that arrange the modules in alphabetical order.
And recommended that the local package, not the module (API) created by Volatility, be directed to the top.

For Example:
```
import struct

from volatility3.framework.interfaces import configuration, plugins # Point
from volatility3.framework.layers import physical # Point
from volatility3.framework.renderers import TreeGrid
from volatility3.framework.symbols.windows import extensions
```

digitalisx (commented):
I reviewed a very small thing about coding conventions/formats by PEP 8 style!

volatility3/framework/plugins/windows/info.py
paulkermann: I have moved the import

paulkermann:
This PR is rebased on develop so only something like the last 3 commits really matter

volatility3/framework/plugins/windows/info.py
digitalisx: Thank you for reflecting my opinion!

paulkermann:
#718 is now the active PR for this plugin

PR #698, opened 2022-04-09
Feature/qemu pci hole

### LINKED ISSUE #587: QEMU: Holes in physical address space not treated properly

ikelos:
Another solution o #585 and #587, mostly based off the awesome work @cstation did in #613 .

After a comment on the upstream bug, we have a way to determine the architecture, and I've converted that into the appropriate code.  Since @cstation did most of the leg work on this and I'd like to make sure he's appropriately credited, I'm happy for him to either change up #613 branch in a similar way, or submit a new PR, or let me know how he'd like crediting.  5:)

I haven't been able to test all the architecture headers because I only have two examples (`pc-i440fx-eoan` and `pc-i440fx-4.2`), but it should be pretty easy to update/fix as needed if we run into a mistake anywhere.  Please let me know how this works out, and we can merge it (or whatever @cstation comes up with) once everyone's happy.  5:)

### DIFF
--- a/volatility3/framework/layers/qemu.py
+++ b/volatility3/framework/layers/qemu.py
@@ -54,15 +54,15 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
     # and location of the memory gap. Deviating hole sizes could eventually be detected for Linux by e.g. scanning
     # for dmesg entries with a regex like rb'\[mem (0x[0-9a-f]{4,10})-0x[0-9a-f]{4,10}\] available for PCI devices'
 
-    debian_re = r"artful|eoan"
+    distro_re = r"(artful|eoan|rhel[\d\.]+)"
 
     pci_hole_table = {re.compile(r"^pc-i440fx-([23456789]|\d\d+)\.\d$"): (0xe0000000, 0xc0000000, 0x100000000),
-                      re.compile(r"^pc-i440fx-[01].\d$"): (0xe0000000, 0xe0000000, 0x100000000),
-                      re.compile(r"^pc-q35-\d.\d$"): (0xe0000000, 0x80000000, 0x100000000),
-                      re.compile(r"^microvm$"): (0xe0000000, 0xc0000000, 0x100000000),
-                      re.compile(r"^xen$"): (0xe0000000, 0xf0000000, 0x100000000),
-                      re.compile(r"^pc-i440fx-" + debian_re + r"$"): (0xe0000000, 0xc0000000, 0x100000000),
-                      re.compile(r"^pc-q35-" + debian_re + r"$"): (0xe0000000, 0x80000000, 0x100000000),
+                      re.compile(r"^pc-i440fx-[01]\.\d$"): (0xe0000000, 0xe0000000, 0x100000000),
+                      re.compile(r"^pc-q35-\d\.\d$"): (0xb0000000, 0x80000000, 0x100000000),
+                      re.compile(r"^microvm$"): (0xc0000000, 0xc0000000, 0x100000000),
+                      re.compile(r"^xen$"): (0xf0000000, 0xf0000000, 0x100000000),
+                      re.compile(r"^pc-i440fx-" + distro_re + r"$"): (0xe0000000, 0xe0000000, 0x100000000),
+                      re.compile(r"^pc-q35-" + distro_re + r"$"): (0xb0000000, 0x80000000, 0x100000000),
                       }
 
     def __init__(self,
@@ -252,6 +252,7 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
 
     def _fallback_determine_architecture(self) -> str:
         architecture_pattern = rb'pc-(i440fx|q35)-([0-9]{1,2}.[0-9]{1,2}(?:.[0-9]{1,2})?)'
+        old_suffix = "-1.0"
         base_layer = self.context.layers[self._base_layer]
 
         vollog.log(constants.LOGLEVEL_VVVV, "QEVM fallback architecture detection used")
@@ -260,7 +261,7 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
         for offset in base_layer.scan(context = self.context, scanner = res):
             line = base_layer.read(offset, 64)
             regex_results = re.search(architecture_pattern, line)
-            architecture = "pc-" + regex_results.groups()[0].decode()
+            architecture = "pc-" + regex_results.groups()[0].decode() + old_suffix
             return architecture
 
         # If that does not work, look in configuration JSON for devices specific to a certain architecture
@@ -268,10 +269,10 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
         for device in self._configuration.get('devices', []):
             device_name = device.get('vmsd_name', '').lower()
             if 'i440fx' in device_name or 'piix' in device_name:
-                architecture = 'pc-i440fx-2.0'
+                architecture = 'pc-i440fx' + old_suffix
                 break
             elif 'ich9' in device_name:
-                architecture = 'pc-q35-1.0'
+                architecture = 'pc-q35' + old_suffix
                 break
         if architecture:
             return architecture
@@ -282,7 +283,7 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
         for offset in base_layer.scan(context = self.context, scanner = res):
             line = base_layer.read(offset, 64)
             regex_results = re.search(architecture_pattern, line)
-            architecture = "pc-" + regex_results.groups()[0].decode().lower()
+            architecture = "pc-" + regex_results.groups()[0].decode().lower() + old_suffix
             return architecture
 
         vollog.warning("Could not determine QEMU target architecture!")
--- a/volatility3/framework/layers/qemu.py
+++ b/volatility3/framework/layers/qemu.py
@@ -3,9 +3,9 @@
 #
 import functools
 import json
-from typing import Optional, Dict, Any, Tuple, List, Set
+from typing import Any, Dict, List, Optional, Set, Tuple
 
-from volatility3.framework import interfaces, exceptions, constants
+from volatility3.framework import constants, exceptions, interfaces
 from volatility3.framework.layers import segmented
 from volatility3.framework.symbols import intermed
 
@@ -39,6 +39,7 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
                  metadata: Optional[Dict[str, Any]] = None) -> None:
         self._qemu_table_name = intermed.IntermediateSymbolTable.create(context, config_path, 'generic', 'qemu')
         self._configuration = None
+        self._architecture = None
         self._compressed: Set[int] = set()
         self._current_segment_name = b''
         super().__init__(context = context, config_path = config_path, name = name, metadata = metadata)
@@ -139,6 +140,9 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
                 section_len = self.context.object(self._qemu_table_name + constants.BANG + 'unsigned long',
                                                   offset = index,
                                                   layer_name = self._base_layer)
+                self._architecture = self.context.object(self._qemu_table_name + constants.BANG + 'string',
+                                                         offset = index + 4, layer_name = self._base_layer,
+                                                         max_length = section_len)
                 index += 4 + section_len
             elif section_byte == self.QEVM_SECTION_START or section_byte == self.QEVM_SECTION_FULL:
                 section_id = self.context.object(self._qemu_table_name + constants.BANG + 'unsigned long',
--- a/volatility3/framework/layers/qemu.py
+++ b/volatility3/framework/layers/qemu.py
@@ -9,7 +9,7 @@ import struct
 from typing import Any, Dict, List, Optional, Set, Tuple
 
 from volatility3.framework import constants, exceptions, interfaces
-from volatility3.framework.layers import segmented
+from volatility3.framework.layers import scanners, segmented
 from volatility3.framework.symbols import intermed
 
 vollog = logging.getLogger(__name__)
@@ -37,11 +37,32 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
     SEGMENT_FLAG_XBZRLE = 0x40
     SEGMENT_FLAG_HOOK = 0x80
 
-    pci_hole_table = {re.compile(r"^pc-i440fx-\d\.\d$"): (0xc0000000, 0x100000000),
-                      re.compile(r"^pc-1440fx-eoan$"): (0xe0000000, 0x100000000),
-                      re.compile(r"^pc-q35$"): (0x80000000, 0x100000000),
-                      re.compile(r"^microvm$"): (0xc0000000, 0x100000000),
-                      re.compile(r"^xen$"): (0xf0000000, 0x100000000)
+    # See https://qemu.readthedocs.io/en/latest/devel/memory.html for more info
+    #
+    # At least the following values could occur for devices using > 3-4 GB RAM:
+    # +--------------------------------+--------------------------------+------------+-------------+
+    # | Architecture                   | Reference Code                 | Hole Start | Hole End    |
+    # +--------------------------------+--------------------------------+------------+-------------+
+    # | PC i440FX + PIIX "New Default" | qemu/hw/i386/pc_piix.c:98      | 0xc0000000 | 0x100000000 |
+    # | PC i440FX + PIIX "Old Default" | qemu/hw/i386/pc_piix.c:98      | 0xe0000000 | 0x100000000 |
+    # | PC Q35 + ICH9                  | qemu/hw/i386/pc_q35.c:141      | 0x80000000 | 0x100000000 |
+    # | MicroVM                        | qemu/hw/i386/microvm.c:291     | 0xc0000000 | 0x100000000 |
+    # | Xen                            | qemu/hw/i386/xen/xen-hvm.c:248 | 0xf0000000 | 0x100000000 |
+    # +--------------------------------+--------------------------------+------------+-------------+
+    #
+    # For now, we assume that the parameter max-ram-below-4g is not set, since this parameter influences the size
+    # and location of the memory gap. Deviating hole sizes could eventually be detected for Linux by e.g. scanning
+    # for dmesg entries with a regex like rb'\[mem (0x[0-9a-f]{4,10})-0x[0-9a-f]{4,10}\] available for PCI devices'
+
+    debian_re = r"artful|eoan"
+
+    pci_hole_table = {re.compile(r"^pc-i440fx-([23456789]|\d\d+)\.\d$"): (0xe0000000, 0xc0000000, 0x100000000),
+                      re.compile(r"^pc-i440fx-[01].\d$"): (0xe0000000, 0xe0000000, 0x100000000),
+                      re.compile(r"^pc-q35-\d.\d$"): (0xe0000000, 0x80000000, 0x100000000),
+                      re.compile(r"^microvm$"): (0xe0000000, 0xc0000000, 0x100000000),
+                      re.compile(r"^xen$"): (0xe0000000, 0xf0000000, 0x100000000),
+                      re.compile(r"^pc-i440fx-" + debian_re + r"$"): (0xe0000000, 0xc0000000, 0x100000000),
+                      re.compile(r"^pc-q35-" + debian_re + r"$"): (0xe0000000, 0x80000000, 0x100000000),
                       }
 
     def __init__(self,
@@ -65,6 +86,7 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
             raise exceptions.LayerException(name, 'No QEMU magic bytes')
         if header[4:] != b'\x00\x00\x00\x03':
             raise exceptions.LayerException(name, 'Unsupported QEMU version found')
+        vollog.debug("QEVM header found")
 
     def _read_configuration(self, base_layer: interfaces.layers.DataLayerInterface, name: str) -> Any:
         """Reads the JSON configuration from the end of the file"""
@@ -160,13 +182,6 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
                 self._architecture = self.context.object(self._qemu_table_name + constants.BANG + 'string',
                                                          offset = index + 4, layer_name = self._base_layer,
                                                          max_length = section_len)
-                for regex in self.pci_hole_table:
-                    if regex.match(self._architecture):
-                        self._pci_hole_start, self._pci_hole_end = self.pci_hole_table[regex]
-                        vollog.log(constants.LOGLEVEL_VVVV, f"QEVM archicture detected as: {self._architecture}")
-                        break
-                else:
-                    vollog.debug(constants.LOGLEVEL_VVVV, f"QEVM unknown architecture found: {self._architecture}")
                 index += 4 + section_len
             elif section_byte == self.QEVM_SECTION_START or section_byte == self.QEVM_SECTION_FULL:
                 section_id = self.context.object(self._qemu_table_name + constants.BANG + 'unsigned long',
@@ -216,6 +231,62 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
                 pass
             else:
                 raise exceptions.LayerException(self._name, f'QEMU unknown section encountered: {section_byte}')
+
+        # If no architecture has been set, try to determine it using fallback mechanisms
+        if not self._architecture:
+            self._architecture = self._fallback_determine_architecture()
+            if self._architecture is None:
+                vollog.log(constants.LOGLEVEL_VV, f"QEVM architecture could not be determined")
+
+        # Once all segments have been read, determine the PCI hole if any
+        for regex in self.pci_hole_table:
+            if regex.match(self._architecture):
+                self._pci_hole_minimum, self._pci_hole_start, self._pci_hole_end = self.pci_hole_table[regex]
+                if self.maximum_address < self._pci_hole_minimum:
+                    # The PCI hole isn't present because we're below the minimum value
+                    self._pci_hole_start, self._pci_hole_end = 0, 0
+                vollog.log(constants.LOGLEVEL_VVVV, f"QEVM architecture detected as: {self._architecture}")
+                break
+        else:
+            vollog.log(constants.LOGLEVEL_VVVV, f"QEVM unknown architecture found: {self._architecture}")
+
+    def _fallback_determine_architecture(self) -> str:
+        architecture_pattern = rb'pc-(i440fx|q35)-([0-9]{1,2}.[0-9]{1,2}(?:.[0-9]{1,2})?)'
+        base_layer = self.context.layers[self._base_layer]
+
+        vollog.log(constants.LOGLEVEL_VVVV, "QEVM fallback architecture detection used")
+
+        res = scanners.RegExScanner(architecture_pattern)
+        for offset in base_layer.scan(context = self.context, scanner = res):
+            line = base_layer.read(offset, 64)
+            regex_results = re.search(architecture_pattern, line)
+            architecture = "pc-" + regex_results.groups()[0].decode()
+            return architecture
+
+        # If that does not work, look in configuration JSON for devices specific to a certain architecture
+        architecture = None
+        for device in self._configuration.get('devices', []):
+            device_name = device.get('vmsd_name', '').lower()
+            if 'i440fx' in device_name or 'piix' in device_name:
+                architecture = 'pc-i440fx-2.0'
+                break
+            elif 'ich9' in device_name:
+                architecture = 'pc-q35-1.0'
+                break
+        if architecture:
+            return architecture
+
+        # Still haven't found architecture, switch to fallback-method
+        architecture_pattern = rb'Standard PC \((i440FX|Q35)'
+        res = scanners.RegExScanner(architecture_pattern)
+        for offset in base_layer.scan(context = self.context, scanner = res):
+            line = base_layer.read(offset, 64)
+            regex_results = re.search(architecture_pattern, line)
+            architecture = "pc-" + regex_results.groups()[0].decode().lower()
+            return architecture
+
+        vollog.warning("Could not determine QEMU target architecture!")
+        return None
 
     def extract_data(self, index, name, version_id):
         if name == 'ram':
--- a/volatility3/framework/layers/qemu.py
+++ b/volatility3/framework/layers/qemu.py
@@ -61,7 +61,7 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
                       re.compile(r"^pc-q35-\d\.\d$"): (0xb0000000, 0x80000000, 0x100000000),
                       re.compile(r"^microvm$"): (0xc0000000, 0xc0000000, 0x100000000),
                       re.compile(r"^xen$"): (0xf0000000, 0xf0000000, 0x100000000),
-                      re.compile(r"^pc-i440fx-" + distro_re + r"$"): (0xe0000000, 0xe0000000, 0x100000000),
+                      re.compile(r"^pc-i440fx-" + distro_re + r"$"): (0xe0000000, 0xc0000000, 0x100000000),
                       re.compile(r"^pc-q35-" + distro_re + r"$"): (0xb0000000, 0x80000000, 0x100000000),
                       }
 
@@ -251,7 +251,7 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
             vollog.log(constants.LOGLEVEL_VVVV, f"QEVM unknown architecture found: {self._architecture}")
 
     def _fallback_determine_architecture(self) -> str:
-        architecture_pattern = rb'pc-(i440fx|q35)-([0-9]{1,2}.[0-9]{1,2}(?:.[0-9]{1,2})?)'
+        architecture_pattern = rb'pc-(i440fx|q35)-(\d{1,2}\.\d{1,2}|[\w\d\.]+)'
         old_suffix = "-2.0"
         base_layer = self.context.layers[self._base_layer]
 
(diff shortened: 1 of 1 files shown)


cstation:
Thanks for the work done @ikelos! I'm fine with continuing on this branch 👍 

Although this code is quite a bit cleaner, it misses out on different instances. [I've made some images](https://ufile.io/oe3sfeuu) that can be used for testing these cases (expires around May 10th, 2022).
- The decision-point in terms of RAM size on whether or not a hole is present, is not necessarily the start of the hole (e.g. for `>= i440fx-2.0`, the hole starts at `0xc0000000` for RAM sizes `>= 0xe0000000`). File for testing this is `kali-2022.1-live-q35-2800-virsh-dump.qevm.gz`;
- The signature for detecting `q35` does not include a version number, whereas it should. File for testing this is `kali-2022.1-live-q35-2800-virsh-dump.qevm.gz`;
- The `i440fx` "Old Default" is present in `pc-i440fx-1.7` and earlier (hole start at `0xe0000000` instead of `0xc0000000`). This is not implemented yet. File for testing this is `kali-2022.1-live-pc-1.6-4096-virsh-dump.qevm.gz`;
- Moreover, on QEMU installation on Ubuntu, architectures that follow the names of the Ubuntu releases (e.g. `pc-q35-eoan`) are available. For `q35`, this won't give problems once these are successfully detected. For `i440fx`, we should try to find out from which Ubuntu-release on the QEMU-architecture follows the "New Default" (hole start at `0xc0000000` instead of `0xe0000000`). File for testing this is `kali-2022.1-live-pc-artful-4096-qemu-migration.qevm.gz`;
- Finally the biggest and most annoying thing, to make things way worse: the architecture is often not present in the header. [As @dagrh pointed out](https://gitlab.com/qemu-project/qemu/-/issues/940), this only occurs in recent versions. It has a reason that #613 contains a dedicated function with RegEx-scanners and different methods for determining the architecture 😉. Files for testing this are `kali-2022.1-live-pc-\d.*.qevm.gz`. Nevertheless, I did not succeed in getting all cases right either, since the needed clues are sometimes simply absent.

Probably unrelated to the PCI-holes, there are still instances in which migration-files seem to not getting parsed properly. The file `kali-linux-2022.1-live-pc-2.0-4096-qemu-migration-roque.qevm.gz` is a nice example. Once the PCI-hole is detected successfully, Volatility is able to parse the upper part of the process list, but still crashes at some point. Haven't still found out why.

Finally: to provide some insights on how the offsets are established, I personally would include the table with the references to the QEMU source in the code. But that's up to you 😄

ikelos:
Hi @cstation, thanks for giving it a look over and coming up with all those examples!

* Hmmmm, ok, is there simply no pci-hole if ram goes up to 0xd000000, for example?
* No problem, I'll get that fixed up now that I have an example, thanks for that!
* Ok, I thought that might be the case, I can get the regexs to differentiate versions.  Just for clarity, is 1.7 old or new, and if it's old, is there a 1.x that's new or can we just use the first digit to identify?
* Ah, that's what eaon was for!  Weird that the make special hardware just for ubuntu, but ok.   Again, I think we can fashion a regex once we know the oldest identifier that falls on either side of the boundary.
* Yeah, fair enough.  We can include the "hunt down the answer" code, but I'd like to deprecate it one day if possible, and I'd definitely use it as a fallback if the known architecture method fails.  Ideally we're after a series of checks that can produce a string which will match the needed architecture values.

Yeah, that sounds like it might be smear, we get bad images just with normal acquisition (are the QEVM's generated from a saved snapshot, or can they be taken from a running VM and if so, might that cause smear?).

Yeah, happy in include the table in there, sorry I was deleting lines from it as I was filling in the regexs to keep track, but very happy to have it documented.  5:D

I'm not sure when I'll get to this but hopefully in the next week or so.  I'll pass it up for review before it goes in fully...  5:)

cstation:
No problem, it helps a lot when some images are available!
- That's right!
- 👍 
- Everything from 2.0 and up is "New Default", 1.7 and down is "Old Default". So the first digit of the version will do!
- 👍 
- 👍 

The QEVMs are generated from a running VM. I actually haven't got a clue what may cause it.

ikelos:
Ok, some more improvements:

* there's now a ram minimum size value for each arch (currently `0xe0000000` for all archs)
* q35 now takes a version number
* differentiates between `>= pc-i440fx-2.0` and `<= pc-i440fx-1.7`
* there's a list of debian distros, we could probably do a catchall, but for testing they're just a hardcoded list, currently not differentiated and it's assumed they're all > 2.x  equivalent
* the fallback code was moved and slightly jiggered to fit what we're doing
* the useful comment was re-added  5;)

There's still a bunch of work that needs to be done, I wasn't able to get all the imaged working properly but haven't investigated fully yet.  Still, hopefully it's coming along...  5:)

cstation:
Great, thanks for the hard work!

Concerning the ram minimum sizes, the value differs per arch. I was able to find the following values:
- i440fx: `0xe0000000`
- Q35: `0xb0000000`
- MicroVM: `0xc0000000`
- Xen: `0xf0000000`

Hopefully this fixes some of the images for you!

I will try to have a look at the _Ubuntu_ releases to see if they all are >2,x. However, Ubuntu isn't the only distro, [RHEL is having them as well](https://bugzilla.redhat.com/show_bug.cgi?id=1957838). So a catchall seems indeed a proper solution ;)

ikelos:
Hmmmm,

I've made the minimum address changes as suggested (and added rhel support), but I'm still only getting results for:
```
kali-2022.1-live-q35-2800-virsh-dump.qevm
```
and pslist is detecting the linux banner, but failing with an invalid address for the following:
```
kali-linux-2022.1-live-pc-2.0-4096-virsh-dump.qevm
kali-linux-2022.1-live-pc-1.6-4096-virsh-dump.qevm
kali-linux-2022.1-live-pc-artful-4096-qemu-miration.qevm
kali-linux-2022.1-live-pc-2.0-4096-qemu-migration-roque.qevm
```
Not sure what's going wrong, but as best I can tell all the pci hole information is being assigned and working correctly?

volatility3/framework/layers/qemu.py
```diff
@@ -77,16 +113,19 @@ def _get_ram_segments(self, index: int, page_size: int) -> Tuple[List[Tuple[int,
...
+            addr = struct.unpack('>Q', base_layer.read(index, 8))[0]
+
             # Flags are stored in the n least significant bits, where n equals the bit-length of pagesize
             flags = addr & (page_size - 1)
             # addr equals the highest multiple of pagesize <= offset
             # (We assume that page_size is a power of 2)
             addr = addr ^ (addr & (page_size - 1))
             index += 8
 
+            if addr > self._pci_hole_start:
```
cstation: Should be `>=` I think? Since `addr` marks the start of a new block, which would otherwise run into the gap?

volatility3/framework/layers/qemu.py
```diff
@@ -192,6 +235,63 @@ def _load_segments(self):
...
+                    # The PCI hole isn't present because we're below the minimum value
+                    self._pci_hole_start, self._pci_hole_end = 0, 0
+                vollog.log(constants.LOGLEVEL_VVVV, f"QEVM architecture detected as: {self._architecture}")
+                break
+        else:
+            vollog.log(constants.LOGLEVEL_VVVV, f"QEVM unknown architecture found: {self._architecture}")
+
+    def _fallback_determine_architecture(self) -> str:
+        architecture_pattern = rb'pc-(i440fx|q35)-([0-9]{1,2}.[0-9]{1,2}(?:.[0-9]{1,2})?)'
+        old_suffix = "-1.0"
```
cstation: We should use `-2.0` as a default suffix, since most i440fx images will be >= 2.0. Using `old_suffix` will always regard i440fx-images that need the fallback detection method as being the "Old Default".

volatility3/framework/layers/qemu.py
```diff
@@ -192,6 +235,63 @@ def _load_segments(self):
...
+        base_layer = self.context.layers[self._base_layer]
+
+        vollog.log(constants.LOGLEVEL_VVVV, "QEVM fallback architecture detection used")
+
+        res = scanners.RegExScanner(architecture_pattern)
+        for offset in base_layer.scan(context = self.context, scanner = res):
+            line = base_layer.read(offset, 64)
+            regex_results = re.search(architecture_pattern, line)
+            architecture = "pc-" + regex_results.groups()[0].decode() + old_suffix
```
cstation: Don't use the suffix here, since the regex contains the version number. Use `f"pc-{regex_results.groups()[0].decode()-{regex_results.groups()[1].decode()}"` instead.

volatility3/framework/layers/qemu.py
cstation: Moreover, the use of such suffixes is not very neat, since the logging will output that Volatility detected a certain machine version, whereas in many cases we are not sure about the exact version.

volatility3/framework/layers/qemu.py
ikelos: Cool, I'll fix that.

volatility3/framework/layers/qemu.py
ikelos: I agree, but the alternative is allowing the regexs to match without a version and then I guess the first one would take priority?  I thought not having the overlap but explicitly adding the version would be better, but yeah, we don't want to add it for cases where we already know the version number.

volatility3/framework/layers/qemu.py
ikelos: Yeah, that feels really ugly, I'd prefer to not using the grouping and just return the whole match, except I was a little too lazy to figure out how to do it when making the changes.  So I'll turn this into the entire match and hopefully that'll satisfy both of us...  5;)

ikelos (commented):
Thanks for the comments, I'll get the changes made up when I can... 5:)

volatility3/framework/layers/qemu.py
```diff
@@ -192,6 +235,63 @@ def _load_segments(self):
...
+                if self.maximum_address < self._pci_hole_minimum:
+                    # The PCI hole isn't present because we're below the minimum value
+                    self._pci_hole_start, self._pci_hole_end = 0, 0
+                vollog.log(constants.LOGLEVEL_VVVV, f"QEVM architecture detected as: {self._architecture}")
+                break
+        else:
+            vollog.log(constants.LOGLEVEL_VVVV, f"QEVM unknown architecture found: {self._architecture}")
+
+    def _fallback_determine_architecture(self) -> str:
+        architecture_pattern = rb'pc-(i440fx|q35)-([0-9]{1,2}.[0-9]{1,2}(?:.[0-9]{1,2})?)'
```
cstation: I noticed that this should include support for Ubuntu/RHEL-releases as well. Suggestion: `rb'pc-(i440fx|q35)-(\d{1,2}.\d{1,2}|[\w\d\.]+)'`. I do not exactly remember why I put `(?:.[0-9]{1,2})?)` in the last part of the expression.

volatility3/framework/layers/qemu.py
```diff
@@ -33,15 +37,46 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
...
+    # for dmesg entries with a regex like rb'\[mem (0x[0-9a-f]{4,10})-0x[0-9a-f]{4,10}\] available for PCI devices'
+
+    distro_re = r"(artful|eoan|rhel[\d\.]+)"
+
+    pci_hole_table = {re.compile(r"^pc-i440fx-([23456789]|\d\d+)\.\d$"): (0xe0000000, 0xc0000000, 0x100000000),
+                      re.compile(r"^pc-i440fx-[01]\.\d$"): (0xe0000000, 0xe0000000, 0x100000000),
+                      re.compile(r"^pc-q35-\d\.\d$"): (0xb0000000, 0x80000000, 0x100000000),
+                      re.compile(r"^microvm$"): (0xc0000000, 0xc0000000, 0x100000000),
+                      re.compile(r"^xen$"): (0xf0000000, 0xf0000000, 0x100000000),
+                      re.compile(r"^pc-i440fx-" + distro_re + r"$"): (0xe0000000, 0xe0000000, 0x100000000),
```
cstation: Should be `(0xe0000000, 0xc0000000, 0x100000000)`, as we asume all machine versions with a Ubuntu/RHEL-release follow the "New Default"

volatility3/framework/layers/qemu.py
```diff
@@ -192,6 +235,63 @@ def _load_segments(self):
             else:
                 raise exceptions.LayerException(self._name, f'QEMU unknown section encountered: {section_byte}')
 
+        # If no architecture has been set, try to determine it using fallback mechanisms
+        if not self._architecture:
+            self._architecture = self._fallback_determine_architecture()
+            if self._architecture is None:
+                vollog.log(constants.LOGLEVEL_VV, f"QEVM architecture could not be determined")
+
+        # Once all segments have been read, determine the PCI hole if any
```
cstation: When this code is executed before the segments are read, more images will work. I didn't figure out why though 🤔 However, moving this code up will make reading the architecture from the `QEVM_CONFIGURATION`-section useless, since the value is not needed anymore.

ikelos: That suggests that the value from the configuration and the value the fallback is finding don't match up, which is concerning, but at least an avenue for investigation.  I don't have time to dig into it now, but I'll give it a look when I can...

cstation:
> Hmmmm,
> 
> I've made the minimum address changes as suggested (and added rhel support), but I'm still only getting results for:
> 
> ```
> kali-2022.1-live-q35-2800-virsh-dump.qevm
> ```
> 
> and pslist is detecting the linux banner, but failing with an invalid address for the following:
> 
> ```
> kali-linux-2022.1-live-pc-2.0-4096-virsh-dump.qevm
> kali-linux-2022.1-live-pc-1.6-4096-virsh-dump.qevm
> kali-linux-2022.1-live-pc-artful-4096-qemu-miration.qevm
> kali-linux-2022.1-live-pc-2.0-4096-qemu-migration-roque.qevm
> ```
> 
> Not sure what's going wrong, but as best I can tell all the pci hole information is being assigned and working correctly?

For the `kali-linux-2022.1-live-pc-2.0-4096-virsh-dump.qevm`, the problem seems to be that the pci-hole location needs to be established before the segments are read.

For `kali-linux-2022.1-live-pc-1.6-4096-virsh-dump.qevm`, we can't establish the version, since it does not seem to be present in the file. This will causes it to be treated like being >=2.0, which will make it fail. Manually editing the code to make it fallback to <=1.7 will make it work. I do not know how to properly fix this. Is it feasible to try "New Default" first, and if that leads to out-of-bounds reading to switch to "Old Default"? Seems horrible, though. Or we just leave that case, since <=1.7 images seldom occur?

The other two images also fail using #613, though they print a large part of the process list. I do not know exactly if some offsets are still off, or if the images contain some smear.

ikelos:
If we've seen versions that say 1.7, then it stands to reason that any file format that doesn't include the version number is older.  That's why I'd say old_suffix to "-1.0".  If that gets clocked back, will the 1.6 image start working and if so is there any problem with just setting it that way, you seemed to want it bumped to 2.0 in one of your comments?

volatility3/framework/layers/qemu.py
cstation: The values do match. I think it is caused by the fact that the code for reading the segments makes some calls to `self._extract_data()`, which calls `self._get_ram_segments()`, which on its turn needs to know where the PCI-holes are. So reading segments before knowing the hole locations will lead to the images being read as if there are no holes?

cstation:
> If we've seen versions that say 1.7, then it stands to reason that any file format that doesn't include the version number is older. That's why I'd say old_suffix to "-1.0". If that gets clocked back, will the 1.6 image start working and if so is there any problem with just setting it that way, you seemed to want it bumped to 2.0 in one of your comments?

Two different version numbers play a role I think: the type + version of the machine being emulated (e.g. `pc-q35-artful`), and the version of QEMU. The first determines the location of the hole, the latter determines whether or not a version number is included in the migration file (I think). With e.g. QEMU 6.0, one could still emulate machine versions < 6.0. I need to perform some research to see if I can find out when the architecture gets included in the configuration section.

ikelos:
Hmmm, ok.  Well, when we get a list of distros and which versions they marry up with that should be fine.  You're right, we do process the whole when building the segments (which makes sense so we don't have to tinker with the mapping function).  

The downside is that if we calculate it before we've been through all the segments, we don't know the maximum size of the image (which is based on the segments, which... haven't been read yet).  We *could* base it on the size of the underlying file, but feels slightly off (if there's bit segments that aren't mapped, we'll get the wrong size).  Not sure how to fix that one, but it should only impact a minimal number of images, so I've just removed the minimum value for the time being.

I've confirmed that `kali-linux-2022.1-live-pc-2.0-4096-virsh-dump.qevm` works with the code now.

cstation:
> The downside is that if we calculate it before we've been through all the segments, we don't know the maximum size of the image (which is based on the segments, which... haven't been read yet). We _could_ base it on the size of the underlying file, but feels slightly off (if there's bit segments that aren't mapped, we'll get the wrong size). Not sure how to fix that one, but it should only impact a minimal number of images, so I've just removed the minimum value for the time being.

This is done properly in #613 though. Reading the machine type from the configuration section is the neat solution, but it may probably make things in the end messy and complicated I think. If the machine type is present in the configuration section, the first regex scanner from the fallback-method will hit on that in an early stage as well.

ikelos:
True, but that's not actually the problem, even if we did the scans, they need to be run before we have the sections and the sections define the maximum size.  The only other way to do it would be to dynamically add the pci hole whenever a section is used (or, possibly better, calculate all the sections, and then at the end go fix them all up according to the identified hole.  I'll work on doing it that way over the weekend...

cstation:
I tried to say two different, unrelated things in my previous comment, but things got maybe a bit messy. I'm sorry 😞 

1. In #613, I properly retrieve the total RAM size in an easy manner and determine the PCI-hole afterwards. Please have a look at it. (I still think it a nice fix which probably only needs some minor changes.)
2. Trying to read the machine-version from the config-section makes things probably unnecessarily complicated (in my opinion). Same goes for storing the offsets and sizes in a regex-based look-up table (values end up twice in it, need to use `old_suffix`, etc.). However, this is just a matter of style and preference I think 😉

ikelos:
Hehehe, ok, I'll go back and have another look at 1.  Hmm, I didn't recall there was a single item that specifies the segment size, can we make presumptions about the order of the item (will it occur before any of the data segments are processed)?  Seems like a good way of handling it, and it can call the fallback if the architecture hasn't been determined yet, cool.  5:)

For number 2, it does add a little complexity, but it guarantees the right results when that value is there.  I don't mind the slight repeating in the regex if they don't overlap.  I feel it would be more complicate to write a single regex that captures all the values which need a particular hole, rather than just writing two non-overlapping regexs.  I agree it's a matter of style, and I'm happy to make it more efficient, but choices of style will never please everybody...  5;)

ikelos:
Ok, hopefully that now uses the right maximum when deciding whether to use the pci hole or not.  I'm gonna keep the regex table, I think it's flexible enough whilst still being compact enough.  I'm also keeping the fallback code off on its own since eventually the only examples of QEVM formats will encode their architecture and we can just chop out all the heuristics.  If we're happy with this, I can turn off the draft mark, and in a few days we can get it merged...  5:)

volatility3/framework/layers/qemu.py
```diff
@@ -33,15 +37,47 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
...
+    # | PC Q35 + ICH9                  | qemu/hw/i386/pc_q35.c:141      | 0x80000000 | 0x100000000 |
+    # | MicroVM                        | qemu/hw/i386/microvm.c:291     | 0xc0000000 | 0x100000000 |
+    # | Xen                            | qemu/hw/i386/xen/xen-hvm.c:248 | 0xf0000000 | 0x100000000 |
+    # +--------------------------------+--------------------------------+------------+-------------+
+    #
+    # For now, we assume that the parameter max-ram-below-4g is not set, since this parameter influences the size
+    # and location of the memory gap. Deviating hole sizes could eventually be detected for Linux by e.g. scanning
+    # for dmesg entries with a regex like rb'\[mem (0x[0-9a-f]{4,10})-0x[0-9a-f]{4,10}\] available for PCI devices'
+
+    distro_re = r"(artful|eoan|rhel[\d\.]+)"
```
cstation: We could maybe replace this with `r"(\w+|rhel[\d\.]+)"` to support all distro names?

volatility3/framework/layers/qemu.py
```diff
@@ -74,32 +111,43 @@ def _get_ram_segments(self, index: int, page_size: int) -> Tuple[List[Tuple[int,
...
-                    #                                       layer_name = self._base_layer)
+                    total_size = self._context.object(self._qemu_table_name + constants.BANG + 'unsigned long long',
+                                                      offset = index + 1 + namelen,
+                                                      layer_name = self._base_layer)
+                    size_array[base_layer.read(index + 1, namelen)] = total_size
                     index += 1 + namelen + 8
                     namelen = self._context.object(self._qemu_table_name + constants.BANG + 'unsigned char',
                                                    offset = index,
                                                    layer_name = self._base_layer)
+                if size_array.get(b'pc.ram',
```
cstation: Perhaps replace this with `if b'pc.ram' in size_array and size_array[b'pc.ram'] < self._pci_hole_minimum:` for better readability? Moreover, it needs to be `<` instead of `<=` I think?

volatility3/framework/layers/qemu.py
```diff
@@ -192,6 +263,45 @@ def _load_segments(self):
             else:
                 raise exceptions.LayerException(self._name, f'QEMU unknown section encountered: {section_byte}')
 
+    def _fallback_determine_architecture(self) -> str:
+        architecture_pattern = rb'pc-(i440fx|q35)-(\d{1,2}\.\d{1,2}|[\w\d\.]+)'
+        old_suffix = "-2.0"
```
cstation: Rename to `default_suffix`, since it is actually the "New Default" for i440fx that we want to use as default fallback and not the "Old Default".

cstation:
I added my last few comments (I think 😉). I'm planning to do some last testing before the weekend to see if I encounter any last issues. If I haven't made any additional comments by then, I'm fine with merging!

volatility3/framework/layers/qemu.py
ikelos: Yeah, I thought it might be better to get bug reports when one comes out we don't support, but I guess it won't really change how we handle it.

ikelos: I'm not sure the version number won't potentially follow other distros, so I've gone for:
```distro_re = r"(\w+[\d\.]?)"```

File: volatility3/framework/layers/qemu.py (hunk 8 above)
Comment (ikelos) [reply volatility3/framework/layers/qemu.py]: Using defaults is fairly standard throughout our codebase, but I've moved the complex bit into a variable for clarity.

File: volatility3/framework/layers/qemu.py (hunk 9 above)
Comment (ikelos) [reply volatility3/framework/layers/qemu.py]: Yep, good point.  5:)

**Comment by ikelos** [comment]
Cool, thanks for the extra review!  I'll merge it on Sunday if there's been no change here then.  Thanks for all your help (and persistence against my objections) with this!  5:D

File: volatility3/framework/layers/qemu.py [hunk 10]
```diff
@@ -33,15 +37,47 @@ class QemuSuspendLayer(segmented.NonLinearlySegmentedLayer):
...
+    # | PC Q35 + ICH9                  | qemu/hw/i386/pc_q35.c:141      | 0x80000000 | 0x100000000 |
+    # | MicroVM                        | qemu/hw/i386/microvm.c:291     | 0xc0000000 | 0x100000000 |
+    # | Xen                            | qemu/hw/i386/xen/xen-hvm.c:248 | 0xf0000000 | 0x100000000 |
+    # +--------------------------------+--------------------------------+------------+-------------+
+    #
+    # For now, we assume that the parameter max-ram-below-4g is not set, since this parameter influences the size
+    # and location of the memory gap. Deviating hole sizes could eventually be detected for Linux by e.g. scanning
+    # for dmesg entries with a regex like rb'\[mem (0x[0-9a-f]{4,10})-0x[0-9a-f]{4,10}\] available for PCI devices'
+
+    distro_re = r"(\w+[\d\.]?)"
```
cstation: Change to `r"(\w+[\d{1,2}\.]*)"`, since e.g. `pc-q35-rhel7.2.0` should match as well.

volatility3/framework/layers/qemu.py
```diff
@@ -192,6 +263,45 @@ def _load_segments(self):
             else:
                 raise exceptions.LayerException(self._name, f'QEMU unknown section encountered: {section_byte}')
 
+    def _fallback_determine_architecture(self) -> str:
+        architecture_pattern = rb'pc-(i440fx|q35)-(\d{1,2}\.\d{1,2}|[\w\d\.]+)'
```
cstation: Change to `rb'pc-(i440fx|q35)-(\d{1,2}\.\d{1,2}|\w+[\d{1,2}\.]*)'` to match regex on line 57

volatility3/framework/layers/qemu.py
```diff
@@ -74,32 +111,43 @@ def _get_ram_segments(self, index: int, page_size: int) -> Tuple[List[Tuple[int,
...
+                    size_array[base_layer.read(index + 1, namelen)] = total_size
                     index += 1 + namelen + 8
                     namelen = self._context.object(self._qemu_table_name + constants.BANG + 'unsigned char',
                                                    offset = index,
                                                    layer_name = self._base_layer)
+                highest_possible_maximum = max([x[0] for x in self.pci_hole_table.values()]) + 1
+                if size_array.get(b'pc.ram', highest_possible_maximum) < self._pci_hole_minimum:
+                    # Turns off the pci_hole if it's not supposed to be there
+                    vollog.debug(
+                        f"QEVM tunrning off PCI hole due to small image size: {size_array.get(b'pc.ram'):x} < {self._pci_hole_minimum:x}")
```
cstation: Add `0x`'s before the addresses for readability and consistency

volatility3/framework/layers/qemu.py
```diff
@@ -192,6 +263,45 @@ def _load_segments(self):
...
+        for device in self._configuration.get('devices', []):
+            device_name = device.get('vmsd_name', '').lower()
+            if 'i440fx' in device_name or 'piix' in device_name:
+                architecture = 'pc-i440fx' + default_suffix
+                break
+            elif 'ich9' in device_name:
+                architecture = 'pc-q35' + default_suffix
+                break
+        if architecture:
+            return architecture
```
cstation, line 289: We should maybe include a log message here that says we did establish the architecture, but that the version is just a guess. Currently we simply print "yeah, its version `2.0`", but we just don't know.

volatility3/framework/layers/qemu.py
```diff
@@ -192,6 +263,45 @@ def _load_segments(self):
...
+
+        # Still haven't found architecture, switch to fallback-method
+        architecture_pattern = rb'Standard PC \((i440FX|Q35)'
+        res = scanners.RegExScanner(architecture_pattern)
+        for offset in base_layer.scan(context = self.context, scanner = res):
+            line = base_layer.read(offset, 64)
+            regex_results = re.search(architecture_pattern, line)
+            architecture = "pc-" + regex_results.groups()[0].decode().lower() + default_suffix
+            return architecture
```
cstation, line 299: We should maybe include a log message here that says we did establish the architecture, but that the version is just a guess. Currently we simply print "yeah, its version `2.0`", but we just don't know.

cstation:
I did some last testing today and added my last comments / suggestions. Thanks again!

ikelos:
Ok, hopefully that's everything resolved.  Sorry for the delay, but if you're happy with it @cstation, then we can get it merged!  5:)

volatility3/framework/layers/qemu.py
```diff
@@ -73,32 +111,43 @@ def _get_ram_segments(self, index: int, page_size: int) -> Tuple[List[Tuple[int,
...
+                    size_array[base_layer.read(index + 1, namelen)] = total_size
                     index += 1 + namelen + 8
                     namelen = self._context.object(self._qemu_table_name + constants.BANG + 'unsigned char',
                                                    offset = index,
                                                    layer_name = self._base_layer)
+                highest_possible_maximum = max([x[0] for x in self.pci_hole_table.values()]) + 1
+                if size_array.get(b'pc.ram', highest_possible_maximum) < self._pci_hole_minimum:
+                    # Turns off the pci_hole if it's not supposed to be there
+                    vollog.debug(
+                        f"QEVM tunrning off PCI hole due to small image size: 0x{size_array.get(b'pc.ram'):x} < 0x{self._pci_hole_minimum:x}")
```
digitalisx: Thank you for your good work!
Everything is fine, but could you check if your intention is correct about this? :)

```suggestion
                        f"QEVM turning off PCI hole due to small image size: 0x{size_array.get(b'pc.ram'):x} < 0x{self._pci_hole_minimum:x}")
```

ikelos:
You're quite right, thanks for the eagle eyes!  5:)

cstation:
I'm happy! :)

ikelos:
Hurray, thanks for sticking with me in getting it over the line, much appreciated!!!  5:D

digitalisx:
I applaud yours passion for working hard for a long time to solve problem.
It's very interesting to watch the process. 🙂

PR #702, opened 2022-04-11
Add Initial address canonization support

### LINKED ISSUE #699: Non-canonical address masking
Currently volatility3 returns non-canonical results when de-referencing addresses (like pointers). It means by default pointers are masked with their layer's `address_mask`. This makes pointer comparison harder, as it is inconsistent with the way the actual OS and hardware handles addresses. 

I believe we should use complete addresses (non-masked addresses) as the conventional address format, so as not to require additional logic in address handling. It will make pointer comparison more consistent with the way it is implemented in the OS.

I'm willing to implement this solution myself as needed, looking forward to hearing your thoughts.

paulkermann:
This PR fixes #699.
Looking forward for more testing and comments.

### DIFF
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -75,7 +75,7 @@ class Intel(linear.LinearlyMappedLayer):
     @classproperty
     @functools.lru_cache()
     def maximum_address(cls) -> int:
-        return (1 << cls._maxvirtaddr) - 1
+        return (1 << cls._bits_per_register) - 1
 
     @classproperty
     def structure(cls) -> List[Tuple[str, int, bool]]:
@@ -102,6 +102,10 @@ class Intel(linear.LinearlyMappedLayer):
         translated address lives in and the layer_name that the address
         lives in
         """
+        if not self.is_address_canonical(offset):
+            # TODO: raise an exception if the address is not canonical
+            vollog.warning(f"Address {hex(offset)} is not canonical!")
+
         entry, position = self._translate_entry(offset)
 
         # Now we're done
@@ -231,6 +235,7 @@ class Intel(linear.LinearlyMappedLayer):
         This allows translation layers to provide maps of contiguous
         regions in one layer
         """
+        offset = self.get_canonical_address(offset)
         if length == 0:
             try:
                 mapped_offset, _, layer_name = self._translate(offset)
@@ -264,11 +269,31 @@ class Intel(linear.LinearlyMappedLayer):
                 length -= chunk_size
                 offset += chunk_size
 
+            if offset > self.maximum_address:
+                break
+            
+            offset = self.get_canonical_address(offset)
+
     @property
     def dependencies(self) -> List[str]:
         """Returns a list of the lower layer names that this layer is dependent
         upon."""
         return [self._base_layer] + self._swap_layers
+
+    @classmethod
+    def is_address_canonical(cls, address: int) -> bool:
+        """Returns whether the address is canonical or not."""
+        return ((address >> (cls._maxvirtaddr - 1)) & 1) == ((address >> (cls._bits_per_register - 1)) & 1)
+
+    @classmethod
+    def _sign_extend(cls, value: int, bits: int) -> int:
+        sign_bit = 1 << (bits - 1)
+        return (value & (sign_bit - 1)) - (value & sign_bit)
+
+    @classmethod
+    def get_canonical_address(cls, address: int) -> int:
+        """Make the address canonical"""
+        return cls._sign_extend(address, cls._maxvirtaddr) & cls.maximum_address
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -102,10 +102,6 @@ class Intel(linear.LinearlyMappedLayer):
         translated address lives in and the layer_name that the address
         lives in
         """
-        if not self.is_address_canonical(offset):
-            # TODO: raise an exception if the address is not canonical
-            vollog.warning(f"Address {hex(offset)} is not canonical!")
-
         entry, position = self._translate_entry(offset)
 
         # Now we're done
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -282,7 +282,12 @@ class Intel(linear.LinearlyMappedLayer):
 
     @classmethod
     def is_address_canonical(cls, address: int) -> bool:
-        """Returns whether the address is canonical or not."""
+        """Returns whether the address is canonical or not.
+        When a 64-bit linear address is generated to access memory, the processor first
+        confirms that the address is canonical. If the address is not canonical, the memory
+        access causes a fault, and the processor makes no attempt to translate the address.
+        To support LAM48 and LAM57 we only check the most upper bit and the max virtual address bit.
+        """
         return ((address >> (cls._maxvirtaddr - 1)) & 1) == ((address >> (cls._bits_per_register - 1)) & 1)
 
     @classmethod
@@ -292,7 +297,11 @@ class Intel(linear.LinearlyMappedLayer):
 
     @classmethod
     def get_canonical_address(cls, address: int) -> int:
-        """Make the address canonical"""
+        """Make the address canonical.
+        A linear address is canonical only if bits the bits above max virtual address are a sign-extension of
+        the max virtual address bit, which is the uppermost bit used in linear-address translation.
+        To do this we sign extend the address based on the max virtual address bit.
+        """
         return cls._sign_extend(address, cls._maxvirtaddr) & cls.maximum_address
 
     @classmethod
--- a/volatility3/framework/layers/linear.py
+++ b/volatility3/framework/layers/linear.py
@@ -69,3 +69,13 @@ class LinearlyMappedLayer(interfaces.layers.TranslationLayerInterface):
                        sections: Iterable[Tuple[int, int]],
                        linear: bool = True) -> Iterable[interfaces.layers.IteratorValue]:
         return super()._scan_iterator(scanner, sections, linear)
+
+    @classmethod
+    def is_address_canonical(cls, address: int) -> bool:
+        """Returns whether the address is canonical or not."""
+        return True
+
+    @classmethod
+    def get_canonical_address(cls, address: int) -> int:
+        """Make the address canonical"""
+        return address
--- a/volatility3/framework/layers/linear.py
+++ b/volatility3/framework/layers/linear.py
@@ -79,3 +79,4 @@ class LinearlyMappedLayer(interfaces.layers.TranslationLayerInterface):
     def get_canonical_address(cls, address: int) -> int:
         """Make the address canonical"""
         return address
+        
--- a/volatility3/framework/layers/linear.py
+++ b/volatility3/framework/layers/linear.py
@@ -1,8 +1,9 @@
-import functools
+import functools, logging
 from typing import List, Optional, Tuple, Iterable
 
 from volatility3.framework import exceptions, interfaces
 
+vollog = logging.getLogger(__name__)
 
 class LinearlyMappedLayer(interfaces.layers.TranslationLayerInterface):
     """Class to differentiate Linearly Mapped layers (where a => b implies that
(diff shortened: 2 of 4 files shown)


volatility3/framework/plugins/windows/handles.py
```diff
@@ -57,10 +57,11 @@ def _decode_pointer(self, value, magic):
 
         value = value & 0xFFFFFFFFFFFFFFF8
         value = value >> magic
-        # if (value & (1 << 47)):
-        #    value = value | 0xFFFF000000000000
+        
+        kernel = self.context.modules[self.config['kernel']]
+        virtual = kernel.layer_name
```
digitalisx: Typically, the section where you set the context is located at the top of the class method.

volatility3/framework/plugins/windows/handles.py
```diff
@@ -57,10 +57,11 @@ def _decode_pointer(self, value, magic):
```
digitalisx: When I checked whether there was a place that used the same method, I found that the `_decode_pointer` method of `windows.netstat` was derived from `windows.handles`.

https://github.com/volatilityfoundation/volatility3/blob/f51914746428c3ce61b68838389ba0667c08064d/volatility3/framework/plugins/windows/netstat.py#L45

If your improvement is correct, it would be very helpful to revise it.

volatility3/framework/layers/linear.py
```diff
@@ -69,3 +69,14 @@ def _scan_iterator(self,
...
+    @classmethod
+    def is_address_canonical(cls, address: int) -> bool:
+        """Returns whether the address is canonical or not."""
+        return True
+
+    @classmethod
+    def get_canonical_address(cls, address: int) -> int:
+        """Make the address canonical"""
+        return address
+
```
digitalisx, line 87: When I checked the commit, is it the version that hasn't been removed? It looks like the method has already been reflected in another file. Please check it out.

digitalisx (commented):
Hello @paulkermann,
First of all, Thank you for your good work. 🙂

I don't have the right to decide on your PR and opinions, not the Maintainer of this project.
However, I would like to give advice on the code, make it better PR to the Project Maintainers and save their review time.

As a result of reviewing the issue, it seems that there is still a lot of discussion and agreement needed. I hope you get a good conclusion!

volatility3/framework/plugins/windows/handles.py
paulkermann: I will move this up

volatility3/framework/plugins/windows/handles.py
paulkermann: It looks like this function does not do anything similar to the `_decode_pointer` function in handles.
It just essentially removes the 2 lowest bits. Since `_PORT_ASSIGNMENT_ENTRY`'s `Entry`  is of type pointer and not a signed long value there is no real need to add the `get_canonical_address`. Of course adding it will work but nothing will change.

volatility3/framework/layers/linear.py
paulkermann: what do you mean by this?

digitalisx: Looking back, I think the review was conducted with little understanding of the code. No problem!

digitalisx:
@paulkermann Thank you for reflecting on my little review. It would be nice to have in-depth reviews and comments on memory... But I think I have enough knowledge not yet. I always looking forward to your good contribution and hope that your thoughts will be conveyed to Maintainers well. 😊

paulkermann:
With this PR accessing non canonical addresses will cause a python exception.
This can be rectified for backward compatibility if needed.

ikelos:
@Digitalisx, we've discussed this with @paulkermann offline.  It's quite a significant change to the way volatility used to work, and the justification for making such a change hasn't really been made yet.  Accessing non-canonical addresses would need to still work in case it's relied upon by other plugins.  That's one of the reasons I'm very reticent to make this change, it doesn't feel like the limited benefit outweighs the potential costs.

digitalisx:
@ikelos First of all, thank you for letting me know about the offline discussion related this issue!

That's right, I saw your concern mentioned in this PR related issue.
So I checked if there are any improvements or differences in existing codes for the code presented, apart from the actual merge of the requested PRs.
The core area processing with memory is a stage where I am learning by looking at their code, so I am also careful about giving opinions.

As in this case, I have been reviewing issues and PRs these days, or making some comments, and I am often worried that giving wrong information or direction. Please let me know if you think the timing of the review is not appropriate.
(I don't want my passion to harm the community.. 😅)

ikelos:
Hi @Digitalisx, reviewing PRs and so on is fine as long as you're clear about your experience levels and that you're not part of the core team.  We're happy to get community contributions, and happy to have the community review itself as long as everyone's got their eyes open as to what's going on.  If you make mistakes my hope is that you'll be (gently) corrected and the right information wins out...  5:)

paulkermann:
This PR is now backward compatible. Read from non-canonical addresses will log a debug message when an invalid address is being read but will still continue.

volatility3/framework/layers/intel.py
```diff
@@ -102,6 +102,10 @@ def _translate(self, offset: int) -> Tuple[int, int, str]:
         translated address lives in and the layer_name that the address
         lives in
         """
+        if not self.is_address_canonical(offset):
```
ikelos: This will have no effect, since the all places the private method `_translate` are called are now preceded by a call to `get_canonical_address`.

volatility3/framework/layers/linear.py
```diff
@@ -69,3 +74,14 @@ def _scan_iterator(self,
                        sections: Iterable[Tuple[int, int]],
                        linear: bool = True) -> Iterable[interfaces.layers.IteratorValue]:
         return super()._scan_iterator(scanner, sections, linear)
+
+    @classmethod
+    def is_address_canonical(cls, address: int) -> bool:
```
ikelos, line 79: Why is this only on linear layers?  It feels either like it iws either something you might want to be able to call on all layers, or just on Intel?

volatility3/framework/layers/intel.py
```diff
@@ -75,7 +75,7 @@ def minimum_address(cls) -> int:
     @classproperty
     @functools.lru_cache()
     def maximum_address(cls) -> int:
-        return (1 << cls._maxvirtaddr) - 1
+        return (1 << cls._bits_per_register) - 1
```
ikelos, line 78: This is not correct, the maximum virtual address is the maximum address of the layer, not how many bits are used to create a register.  Addresses above the previous maximum value will not map properly.  This is also relied upon behaviour, changing this core constant of the intel layer would require all use of the exposed `maxium_address` value to be examined.

volatility3/framework/plugins/windows/handles.py
```diff
@@ -54,13 +54,14 @@ def _decode_pointer(self, value, magic):
...
 
+        # TODO: Make this logic consistent with how the windows kernel does this.
         value = value & 0xFFFFFFFFFFFFFFF8
-        value = value >> magic
-        # if (value & (1 << 47)):
-        #    value = value | 0xFFFF000000000000
+        value = value >> magic        
 
-        return value
+        return self.context.layers[virtual].get_canonical_address(value)
```
ikelos, line 64: As best I can see, this is the only real change in the request, and it's for a plugin, which can be changed in the plugin.  Essentially, plugins must beware when passing offsets, which is exactly how things are at the moment.

volatility3/framework/layers/intel.py
```diff
@@ -264,12 +269,41 @@ def _mapping(self,
                 length -= chunk_size
                 offset += chunk_size
 
+            if offset > self.maximum_address:
```
ikelos, line 268: This is now only testing whether the address is greater than the largest possible register value.

volatility3/framework/layers/intel.py
```diff
@@ -264,12 +269,41 @@ def _mapping(self,
...
+
+    @classmethod
+    def _sign_extend(cls, value: int, bits: int) -> int:
+        sign_bit = 1 << (bits - 1)
+        return (value & (sign_bit - 1)) - (value & sign_bit)
+
+    @classmethod
+    def get_canonical_address(cls, address: int) -> int:
+        """Make the address canonical.
+        A linear address is canonical only if bits the bits above max virtual address are a sign-extension of
```
ikelos, line 297: Reference please?

volatility3/framework/layers/intel.py
```diff
@@ -264,12 +269,41 @@ def _mapping(self,
...
     def dependencies(self) -> List[str]:
         """Returns a list of the lower layer names that this layer is dependent
         upon."""
         return [self._base_layer] + self._swap_layers
 
+    @classmethod
+    def is_address_canonical(cls, address: int) -> bool:
+        """Returns whether the address is canonical or not.
+        When a 64-bit linear address is generated to access memory, the processor first
+        confirms that the address is canonical. If the address is not canonical, the memory
```
ikelos, line 283: Reference please.

ikelos (changes requested):
As pointed out, this makes a core change to the semantics of the intel layer (`maximum_address`) without good justification.  There need to be clearer examples of the problem trying to be solved, and why this is the best/only place the change can be made.  So far this isn't present, but a number of anecdotal examples have been provided by the author without suitable description or clarity.  This is unlikely to move forward, despite further anecdotes, until a proper explanation of the changes.

volatility3/framework/layers/intel.py
paulkermann: Removed this check

volatility3/framework/layers/intel.py
paulkermann: This check is there to prevent wrap around.

volatility3/framework/layers/intel.py
paulkermann: What do you mean by "will not map properly" ?

volatility3/framework/plugins/windows/handles.py
paulkermann: Actually the problem is that the the plugin _assumes_ that canonical addresses are valid which they are **not**.
With the "backward compatibility" commit the code will work without this change.

volatility3/framework/layers/intel.py
paulkermann: `3.3.7.1 Canonical Addressing` In the intel SDM.
The comment is paraphrasing of the information there and the code is sign-extending the address to its valid canonical form.

volatility3/framework/layers/intel.py
paulkermann: Same as above.
The LAM57 and LAM48 comments are taken from the `Intel® Architecture Instruction Set Extensions Programming Reference` - `CHAPTER 14
LINEAR ADDRESS MASKING (LAM)`.

volatility3/framework/layers/linear.py
paulkermann: In the future the ARM, PPC and others will inherit from this class just like the `Intel` layer.
It is not Intel specific and will work for other architectures.

volatility3/framework/plugins/windows/handles.py
```diff
@@ -54,13 +54,14 @@ def _decode_pointer(self, value, magic):
...
         proper pointer values as well.
         """
+        kernel = self.context.modules[self.config['kernel']]
+        virtual = kernel.layer_name
 
+        # TODO: Make this logic consistent with how the windows kernel does this.
         value = value & 0xFFFFFFFFFFFFFFF8
-        value = value >> magic
-        # if (value & (1 << 47)):
-        #    value = value | 0xFFFF000000000000
```
paulkermann, line 61: This "magic constant" here is to make the address canonical.

ikelos (changes requested):
Have you even tested this code?  I just tried a standard run again a windows 10 image running psscan.  I expected it to take orders of magnitude longer due to increasing the address space that needs scanning from 48-bits to 64-bits, but in fact it breaks during automagic.

You were asked to provide justification for such a significant change and still have yet to do so other than you feel it's a more correct way of operating.  That is not the only consideration but nothing else appears to have been considered in this pull request (such as speed/efficiency or testing of the framework).

I'm going to mark this as closed because we won't be making the intended change, and re-requesting reviews for code that has the same critical flaw will not get it to land.  Apply masking to pointers before use, everything will work, it's no different to applying a sign extension function which would now be required by the patch.

ikelos:
This will not be landing, please consider the requests for justification, testing and ensuring efficiency before submitting this in another form.

ikelos:
For completeness, I found another win10 x64 image and ran `psscan` which gave the following timings:

Without changes: 42.98s user 0.63s system 89% cpu 48.566 total
With changes: 91.39s user 9.34s system 12% cpu 13:38.77 total

That 18x times (an order of magnitude) slower because the address space maximum address has been increased.

Edited to add: this may have been differences between hot/cold disk images, ie one had been cached and the other not.  I don't have the update results after ensuring both were in the cache, but I believe they were comparable.

PR #705, opened 2022-04-17
Add Rich output for tables

### LINKED ISSUE #704: Integrate Rich for output
No problem.

I would like to integrate [Rich](https://github.com/Textualize/rich) as a renderer for output. The volatility output is currently text based and can be very misaligned when it comes to certain output, like TreeGrids that print out asterisks to indicate node depth. Although the `pretty`  renderer option usually works, I think it would look better using Rich. As an example, here is a `pretty` windows.info compared to a `rich` one:

image
image

Here is an example of `pstree` which indicates the node depth.

image
image

**Additional information**
The colors show in the screenshots would be configurable in accordance with [Rich Style docs](https://rich.readthedocs.io/en/stable/style.html#loading-themes). The colors shown in this screenshot were made with these settings:

```
[styles]
repr.call = bold italic #00d8eb
repr.ipv6 = yellow 
repr.number = #008df8
repr.str = bright_blue
table.header = bold italic #00d8eb
table.border = #e1251b
```

daddycocoaman:
Hi. This PR is to address #704 to add Rich output.


.gitignore
```diff
@@ -27,3 +27,12 @@ config*.json
 # Pyinstaller files
 build
 dist
+
+# Environments
```
digitalisx: It is very important to exclude venv-related files through .gitignore in Python projects.
However, each user may have a different folder name, and it is a different change from the big topic of this PR, so I recommend you to suggest it separately with a different PR.

.gitignore
```diff
@@ -27,3 +27,12 @@ config*.json
...
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
```
digitalisx: An empty new line is required for the EOF point.

config/richrenderer.ini
```diff
@@ -0,0 +1,7 @@
+[styles]
+repr.call = bold italic #00d8eb
+repr.ipv6 = yellow 
+repr.number = #008df8
+repr.str = bright_blue
+table.header = bold italic #00d8eb
+table.border = #e1251b
```
digitalisx: An empty new line is required for the EOF point.

digitalisx (commented):
Hello @daddycocoaman,
First of all, Thank you for your good work. 🙂

I don't have the right to decide on your PR and opinions, not the Maintainer of this project.
However, I would like to give advice on the code, make it better PR to the Project Maintainers and save their review time. (And need a solve the conflict!)

It looks much better to me, but it seems like a topic that needs to be reviewed and agreed on whether applying a new module called rich fits the direction of the project.

.gitignore
daddycocoaman: I completely forgot I added that to the .gitignore a few months ago. Definitely separate PR worthy, for sure

config/richrenderer.ini
daddycocoaman: Added

daddycocoaman:
@Digitalisx Thanks for the comments and review. Rich is a pretty great module for output so I do hope the maintainers like it.

For the conflict, I'm not sure how the import order for `typing` got screwed up but I've fixed that. The additional conflict appears to be just the additional pathlib import.

.gitignore
digitalisx: Thank you for the update! However, in the process of removing the venv-related line, the existing new line is removed and marked as a change.
If you don't mind, could you add a newline to the .gitignore?

digitalisx:
@daddycocoaman Thank you for the update. I think we just need to proceed with the process for the last `.gitignore` I added. Good luck!

requirements-minimal.txt
```diff
@@ -1,2 +1,3 @@
 # These packages are required for core functionality.
-pefile>=2017.8.1 #foo
\ No newline at end of file
+pefile>=2017.8.1 #foo
+rich>=12.2.0
```
ikelos, line 3: Thanks, I don't think volatility should fail to work if a user doesn't have Rich installed, this can go in the main requirements, but the renderer should fail gracefully if rich isn't present.

volatility3/cli/text_renderer.py
```diff
@@ -8,8 +8,13 @@
 import string
 import sys
 from functools import wraps
-from typing import Callable, Any, List, Tuple, Dict
+from pathlib import Path
+from typing import Any, Callable, Dict, List, Tuple
 
+from rich import box
```
ikelos, line 14: Ideally this should be in a separate file for just this renderer, and the imports in it should be protected by a `try/catch ImportError` which puts a vollog at debug level to say Rich can't be found.  It's up to you whether to have the rich renderer present, but fail to run (with a vollog warning/error) or simply not include the code if the import fails, but it should definitely be optional.

volatility3/cli/text_renderer.py
```diff
@@ -329,6 +338,56 @@ def tab_stop(self, line: str) -> str:
...
+        pass
+
+    def render(self, grid: interfaces.renderers.TreeGrid) -> None:
+        """Renders each column using Rich Table output.
+
+            Args:
+                grid: The TreeGrid object to render
+        """
+        # Allow user to configure tables as they would like from ini file
+        rich_theme_path = Path(__file__).parents[2] / "config/richrenderer.ini"
```
ikelos, line 356: This definitely needs some consideration.  We don't have a configuration file yet, but this mechanism isn't really the route we'd like to go.  This probably needs a whole extra class/mechanism for defining configuration values (to be used so people can define default options for things like -s) and that'll need a lot of though.  So that the time being I'd like to ask that this be removed if you want it merged in, until we get the config mechanism written and in place so the code can use that.

volatility3/cli/text_renderer.py
```diff
@@ -329,6 +338,56 @@ def tab_stop(self, line: str) -> str:
...
+        if rich_theme_path.exists() and rich_theme_path.is_file():
+            theme = Theme.read(str(rich_theme_path.resolve()))
+        else:
+            theme = None
+
+        console = Console(theme = theme)
+        table = Table(expand = False,
+                      box = box.SIMPLE_HEAVY,
+                      highlight = True,
+                      border_style = theme.styles.get("table.border", None) if theme else None)
```
ikelos, line 366: Would a default theme populated with an empty styles be better?  Seems a bit odd to have a default value and a possible different response if the theme doesn't exist...

volatility3/cli/text_renderer.py
```diff
@@ -329,6 +338,56 @@ def tab_stop(self, line: str) -> str:
...
+        [table.add_column(c.name, overflow = "fold") for c in grid.columns]
```
ikelos, line 367: I'm not keen on single letter variables, I know it's pretty clear, but could you stretch to `col` instead of just `c` please?

ikelos (commented):
Thanks very much for your contribution!  This looks cool, and saves all the mess of our custom `pretty` renderer.  I'm not against including it, but it really needs to be an optional dependency, we just need to figure out how to gracefully let the user know they need rich to use it.  I'd suggest including the renderer, but having it fail if the dependency isn't there.

Also, I'd prefer not to allow the configuration mechanism, since the entire tool needs a config mechanism and this should end up using that once it's been planned and written.

A few change though and it looks like it should be good to merge...  5:)

daddycocoaman:
Hi, 

I'm going to take note of the comments made and revisit this later, particularly after the configuration file you want to implement happens. For now, I'm going to write a separate tool to pipe into. 💚

PR #706, opened 2022-04-20
refs #668 handle windows freed big pools more accurately

### LINKED ISSUE #668: Freed big pools in list_big_pools()
When a big pool is freed the struct's `self.Va` is or'd with `0x1` (as per [Alex Ionescu's blog post](https://www.alex-ionescu.com/?p=231)).

> One thing to be aware of is that the Virtual Address (Va) is OR’ed with a bit to indicate if the allocation is freed or allocated — in other words, you may have duplicate Va’s, some freed, and at most one allocated.

Currently the `BigPools` plugin doesn't check if a big pool was freed (and no `is_freed` method is available for `POOL_TRACKER_BIG_PAGES` in volatility3/framework/symbols/windows/extensions/pool.py). So big pool information returned by the plugin can be deceiving, e.g. a virtual address could fall within the bounds of both a freed big pool and a newly-allocated big pool.

I'd suggest having an `is_freed` method, and also allowing to filter results by their freed status in `list_big_pools()`. 
Alternatively, modify the current `is_valid()` method to also verify `(self.Va & 1) == 0`. That depends on your interpretation of the `is_valid()` functionality.

iMHLv2:
This also adds a `--show-free` option, as freed regions are not listed by default.

### DIFF
--- a/volatility3/framework/plugins/windows/bigpools.py
+++ b/volatility3/framework/plugins/windows/bigpools.py
@@ -21,7 +21,7 @@ class BigPools(interfaces.plugins.PluginInterface):
     """List big page pools."""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 0, 0)
+    _version = (1, 0, 1)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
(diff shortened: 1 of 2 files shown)


ikelos (approved):
Yep, looks good, thanks!  5:)

PR #714, opened 2022-04-25
Improving object type_name resolution

paulkermann:
Getting the struct name or symbol table name for an object is mostly done by using `object.vol.type_name.split(constants.BANG)`, which is not ideal.

I have added 2 members to the vol dict in the object interface:
1) short_name: the actual struct name (_EPROCESS, _PEB, unsigned long long) 
1) table_name: `get_symbol_table_name()` does this but has a long name and has to perform the `split()` internally (it's still there for backwards compatibility).

Also added a util to "try except" for the type classes which appeared a few times in the code.

Finally, I added a util function for objects - `at_layer` - which casts said object to a different layer easily.
e.g see the `_EPROCESS` `get_peb()` function.

This PR is fully backwards compatible as only new APIs and members were added. Nothing was removed.
I personally think this could make the code simpler and more intuitive in the future (and change the `.split(constants.BANG)` convention that is prevalent).

### DIFF
--- a/volatility3/framework/interfaces/objects.py
+++ b/volatility3/framework/interfaces/objects.py
@@ -121,6 +121,7 @@ class ObjectInterface(metaclass = abc.ABCMeta):
             vol_info_dict["table_name"] = table_name
             vol_info_dict["short_name"] = struct_name
         else:
+            vol_info_dict["table_name"] = ""
             vol_info_dict["short_name"] = type_name
         
         self._vol = collections.ChainMap({}, vol_info_dict, object_info, kwargs)
--- a/volatility3/framework/interfaces/objects.py
+++ b/volatility3/framework/interfaces/objects.py
@@ -170,11 +170,6 @@ class ObjectInterface(metaclass = abc.ABCMeta):
                                         size = object_template.size)
         return object_template(context = self._context, object_info = object_info)
 
-    def at_layer(self, new_layer_name) -> 'ObjectInterface':
-        """Returns the same object casted at a different layer.
-        """
-        return self._context.object(self.vol.type_name, offset=self.vol.offset, layer_name=new_layer_name)
-
     def has_member(self, member_name: str) -> bool:
         """Returns whether the object would contain a member called
         member_name.
--- a/volatility3/framework/interfaces/objects.py
+++ b/volatility3/framework/interfaces/objects.py
@@ -116,14 +116,6 @@ class ObjectInterface(metaclass = abc.ABCMeta):
         normalized_offset = object_info.offset & mask
 
         vol_info_dict = {'type_name': type_name, 'offset': normalized_offset}
-        if constants.BANG in type_name:
-            table_name, struct_name = type_name.split(constants.BANG)
-            vol_info_dict["table_name"] = table_name
-            vol_info_dict["short_name"] = struct_name
-        else:
-            vol_info_dict["table_name"] = ""
-            vol_info_dict["short_name"] = type_name
-        
         self._vol = collections.ChainMap({}, vol_info_dict, object_info, kwargs)
         self._context = context
 
@@ -151,7 +143,7 @@ class ObjectInterface(metaclass = abc.ABCMeta):
         """
         if constants.BANG not in self.vol.type_name:
             raise ValueError(f"Unable to determine table for symbol: {self.vol.type_name}")
-        table_name = self.vol.table_name
+        table_name = self.vol.type_name[:self.vol.type_name.index(constants.BANG)]
         if table_name not in self._context.symbol_space:
             raise KeyError(f"Symbol table not found in context's symbol_space for symbol: {self.vol.type_name}")
         return table_name
--- a/volatility3/framework/interfaces/objects.py
+++ b/volatility3/framework/interfaces/objects.py
@@ -115,7 +115,15 @@ class ObjectInterface(metaclass = abc.ABCMeta):
         mask = context.layers[object_info.layer_name].address_mask
         normalized_offset = object_info.offset & mask
 
-        self._vol = collections.ChainMap({}, {'type_name': type_name, 'offset': normalized_offset}, object_info, kwargs)
+        vol_info_dict = {'type_name': type_name, 'offset': normalized_offset}
+        if constants.BANG in type_name:
+            table_name, struct_name = type_name.split(constants.BANG)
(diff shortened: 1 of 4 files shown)


volatility3/framework/interfaces/objects.py
```diff
@@ -115,7 +115,16 @@ def __init__(self, context: 'interfaces.context.ContextInterface', type_name: st
         mask = context.layers[object_info.layer_name].address_mask
         normalized_offset = object_info.offset & mask
 
-        self._vol = collections.ChainMap({}, {'type_name': type_name, 'offset': normalized_offset}, object_info, kwargs)
+        vol_info_dict = {'type_name': type_name, 'offset': normalized_offset}
```
ikelos, line 118: This is a lot of effort because you feel `get_symbol_table_name` is too long.  You have to type it once, but it'll get read far more often than that, so I'm all in favour of long names.  Also IDEs tend to autocomplete, so monkeying with things for the sake of saving a few letters isn't a good enough reason.  You're right though, we should probably make use of that method more (for example in cast).  If you really don't like seeing the split, I might consider a method for getting just the symbol name, but it'll be calculated in the method, not stored as additional data that could get out of sync.

volatility3/framework/interfaces/objects.py
```diff
@@ -142,7 +151,7 @@ def get_symbol_table_name(self) -> str:
         """
         if constants.BANG not in self.vol.type_name:
             raise ValueError(f"Unable to determine table for symbol: {self.vol.type_name}")
-        table_name = self.vol.type_name[:self.vol.type_name.index(constants.BANG)]
```
ikelos: This is in a helper function that you never have to see and there's no good reason to change it.

volatility3/framework/interfaces/objects.py
```diff
@@ -156,7 +165,7 @@ def cast(self, new_type_name: str, **additional) -> 'ObjectInterface':
         """
         # TODO: Carefully consider the implications of casting and how it should work
         if constants.BANG not in new_type_name:
-            symbol_table = self.vol['type_name'].split(constants.BANG)[0]
+            symbol_table = self.get_symbol_table_name()
```
ikelos, line 160: This one is a good catch, and we should probably be using this method more in other places.  It might be worth a grep over the rest of the code looking for a similar `.split(contants.BANG)[0]` pattern.

volatility3/framework/interfaces/objects.py
```diff
@@ -169,6 +178,11 @@ def cast(self, new_type_name: str, **additional) -> 'ObjectInterface':
                                         size = object_template.size)
         return object_template(context = self._context, object_info = object_info)
 
+    def at_layer(self, new_layer_name) -> 'ObjectInterface':
```
ikelos: This is a highly specific method, you can't change the offset when you're doing it and it also doesn't offer any control over `native_layer_name` which is used to allow pointers to point from one layer to another.  This isn't going to make it in.

volatility3/framework/interfaces/symbols.py
```diff
@@ -167,6 +167,20 @@ def set_type_class(self, name: str, clazz: Type[objects.ObjectInterface]) -> Non
         """
         raise NotImplementedError("Abstract method set_type_class not implemented yet.")
 
+    def try_set_type_class(self, name: str, clazz: Type[objects.ObjectInterface]) -> bool:
```
ikelos: I'm kind of ok with this idea, although I'm not keen on the `try_` name but I'm not sure what would be better, maybe something about `optional`?  It just seems as though the method name doesn't explain what happens in the case of failure.

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -574,13 +574,9 @@ def get_peb(self) -> interfaces.objects.ObjectInterface:
...
             raise exceptions.InvalidAddressException(proc_layer_name, self.Peb,
-                                                     f"Invalid address at {self.Peb:0x}")
+                                                     f"Invalid Peb address at {self.Peb:0x}")
 
-        sym_table = self.vol.type_name.split(constants.BANG)[0]
-        peb = self._context.object(f"{sym_table}{constants.BANG}_PEB",
-                                   layer_name = proc_layer_name,
-                                   offset = self.Peb)
-        return peb
+        return self.at_layer(proc_layer_name).Peb
```
ikelos: As the only example to use the new method, this is not worth the changes.  I also think it's a pretty uncommon situation to want an object instantiated at the same offset in a mapping that potentially maps that offset to a totally different place.

ikelos (changes requested):
There are a couple of elements of this that are ok, but there's big chunks that seem to be changing code for the sake of it and/or because of laziness at typing long names.  That's not enough to make me accept the changes, long and clear names were chosen for a reason, so the comments will need to be followed as to what will/won't go in.  Once the changes are made, I'll re-review it.

volatility3/framework/interfaces/objects.py
paulkermann: Removed the changes here

volatility3/framework/interfaces/objects.py
paulkermann: Most of it is in  symbol resolution. I think that symbols should be described by an `ObjectInterface` that contains a name (IE `PsLoadedModuleList`). This would be a normalised way to access them.

volatility3/framework/interfaces/symbols.py
paulkermann: renamed it

volatility3/framework/symbols/windows/extensions/__init__.py
paulkermann: removed the function

volatility3/framework/interfaces/objects.py
paulkermann: removed it

volatility3/framework/interfaces/objects.py
paulkermann: restored it

volatility3/framework/interfaces/objects.py
paulkermann: Should be a different PR though.

ikelos (approved):
Looks much more palatable, thanks for making the requested changes.  5:)

PR #718, opened 2022-04-27
Crashdump writer plugin

paulkermann:
This is a PR is instead of #694 because the rebase on that PR was a bit weird.
This was tested against 32-bit and 64-bit memory dumps.


volatility3/framework/plugins/windows/info.py
```diff
@@ -3,15 +3,32 @@
...
+        if (value > 0xFFFFFFFFFFFFFFFF):
+            value -= 0x10000000000000000
+            value += 1
+    return value
+
+def bswap(value: int) -> int:
+    """A byte-swap instruction in Python"""
+
+    hi, lo = struct.unpack(">II", struct.pack("<Q", value))
+    return (hi << 32) | lo
```
digitalisx: If the `rol` and `swap` methods are used only in `Info` plugin class, could you consider converting them to the `@classmethod` in `Info` class?

volatility3/framework/plugins/windows/writecrashdump.py
```diff
@@ -0,0 +1,140 @@
+import math
```
digitalisx: This module does not appear to be in use.

volatility3/framework/plugins/windows/info.py
```diff
@@ -3,15 +3,32 @@
...
+from volatility3.framework.interfaces import plugins, configuration
 from volatility3.framework.renderers import TreeGrid
 from volatility3.framework.symbols import intermed
 from volatility3.framework.symbols.windows import extensions
+from volatility3.framework.layers import physical
 
+def rol(value: int, count: int) -> int: 
+    """A rotate-left instruction in Python"""
+    
+    for y in range(count):
```
digitalisx: If you do not use variables in loop statements, we recommend using `_` rather than meaningless variables such as `i` or `y`.

volatility3/framework/plugins/windows/info.py
```diff
@@ -3,15 +3,32 @@
 #
 
 import time
+import struct
 from typing import List, Tuple, Iterable
 
 from volatility3.framework import constants, interfaces, layers, symbols
 from volatility3.framework.configuration import requirements
-from volatility3.framework.interfaces import plugins
+from volatility3.framework.interfaces import plugins, configuration
```
digitalisx, line 12: PEP8 recommends sorting modules alphabetically.

volatility3/framework/plugins/windows/writecrashdump.py
```diff
@@ -0,0 +1,140 @@
...
+    _required_framework_version = (2, 0, 0)
+    _version = (2, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.TranslationLayerRequirement(name = 'primary', description = 'Memory layer for the kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name = "nt_symbols", description = "Windows kernel symbols"),
+            requirements.IntRequirement(name = 'block_size',
```
digitalisx: The options in volatility3 cli use `-` Perhaps using `block-size` will bring uniformity to the entire command. This part is mixed for each plugins, so I think it would be good to get the maintainer's opinion.

volatility3/framework/plugins/windows/writecrashdump.py
```diff
@@ -0,0 +1,140 @@
...
+            requirements.VersionRequirement("info", component = info.Info, version = (1, 0, 0))
+        ]
+
+    @classmethod
+    def get_physical_layer_name(cls, context: interfaces.context.ContextInterface, vlayer: interfaces.layers.DataLayerInterface) -> str:
+        return context.config.get(interfaces.configuration.path_join(vlayer.config_path, 'memory_layer'), None)
+
+    @classmethod
+    def write_crashdump(cls, context: interfaces.context.ContextInterface, layer_name: str, symbol_table: str,
+                        open_method: Type[interfaces.plugins.FileHandlerInterface]):
```
digitalisx: Although not necessarily enforced in Python, the return type statement will be of great help to developers.

volatility3/framework/plugins/windows/writecrashdump.py
```diff
@@ -0,0 +1,140 @@
...
+                f.write(physical_layer.read(offset, 0x1000, pad = True))
+
+            # Fix KDBG in the dump if it was encoded
+            decoded_data = context.layers[kdbg.vol.layer_name].read(kdbg.vol.offset, kdbg.size())
+            kdbg_physical_address = primary.translate(kdbg.vol.offset)[0]
+            kdbg_file_location = (header_layer.maximum_address + 1) + kdbg_physical_address - physical_layer.minimum_address
+            f.seek(kdbg_file_location)
+            f.write(decoded_data)
+
+    def _generator(self):
```
digitalisx: Like other plugins, I think it would be nice to have a Return Type here! :)

volatility3/framework/plugins/windows/writecrashdump.py
```diff
@@ -0,0 +1,140 @@
...
+        self.write_crashdump(self.context, self.config['primary'], self.config['nt_symbols'], self._file_handler)
+        yield 0, ('Done',)
+
+    def run(self):
```
digitalisx: I think it'd be nice to have a Return Type here, too!

volatility3/framework/plugins/windows/writecrashdump.py
```diff
@@ -0,0 +1,140 @@
...
+            # Fix KDBG in the dump if it was encoded
+            decoded_data = context.layers[kdbg.vol.layer_name].read(kdbg.vol.offset, kdbg.size())
+            kdbg_physical_address = primary.translate(kdbg.vol.offset)[0]
+            kdbg_file_location = (header_layer.maximum_address + 1) + kdbg_physical_address - physical_layer.minimum_address
+            f.seek(kdbg_file_location)
+            f.write(decoded_data)
+
+    def _generator(self):
+        self.write_crashdump(self.context, self.config['primary'], self.config['nt_symbols'], self._file_handler)
+        yield 0, ('Done',)
```
digitalisx: This is NIT, and completing the creation of the dump file is sufficient. Could you consider providing information related to dumps to the user?

digitalisx (commented):
Thanks to the clean up of the PR, it's very easy to see the code! 🙂
I left a little comment.
I hope this plugin can be usefully.

volatility3/framework/plugins/windows/writecrashdump.py
paulkermann: I have removed it

volatility3/framework/plugins/windows/writecrashdump.py
paulkermann: The function does not return anything though

volatility3/framework/plugins/windows/writecrashdump.py
paulkermann: I have not actually written that code. It is a copy-paste code from @ikelos.
Either way I have checked other plugins and they also many of them don't seem to have a return type for the `_generator` or `_run` functions.

volatility3/framework/plugins/windows/info.py
paulkermann: I don't think we really need to export them though
I feel comfortable with them being here right now and when in the future if they are used by a different person then they should be moved to a common utilities file.

volatility3/framework/plugins/windows/writecrashdump.py
paulkermann: I have removed the functionality of block size chunking.
A different PR in the future could add that.

volatility3/framework/plugins/windows/writecrashdump.py
paulkermann: see [this](https://github.com/volatilityfoundation/volatility3/pull/718#discussion_r859915854)

volatility3/framework/plugins/windows/info.py
paulkermann: I have replaced it with `_`

volatility3/framework/plugins/windows/writecrashdump.py
paulkermann: I think providing information about the dump is the purpose of the other volatility plugins (IE `windows.info`).
This is just to convert the dump and the result is the status

volatility3/framework/plugins/windows/info.py
paulkermann: I have replaced the order

digitalisx:
@paulkermann LGTM, some of the things you explained were understood. Thank you for checking my reviews. 🙂

volatility3/framework/plugins/windows/writecrashdump.py
paulkermann: This plugin does not really have anything to yield other then success or failure.
On failure the plugin will throw an exception(or w/e happens by default when a plugin fails) and on success `Done` will be printed.

digitalisx: All right, I thought that if there was anything to provide metadata related to Crashdump to generate, it would be helpful to display it.
If you think you don't have any meaningful information to provide, this is fine, too!

volatility3/framework/plugins/windows/writecrashdump.py
digitalisx: It's possible like this. (I like all your work, but This comment is intended to utilize the Github Suggestion function.)

```suggestion
    def run(self) -> renderers.TreeGrid:
```

volatility3/framework/plugins/windows/writecrashdump.py
digitalisx:
```suggestion
    def _generator(self) -> Iterator[Tuple]:
```

volatility3/framework/plugins/windows/writecrashdump.py
ikelos: Yeah, we haven't historically, and it won't hold up merging, but if you're happy to have it, so are we.  5:)

paulkermann: I have added the return type for the `run` function

volatility3/framework/plugins/windows/writecrashdump.py
paulkermann: I have added the return type for this function

volatility3/framework/plugins/windows/info.py
```diff
@@ -3,15 +3,32 @@
...
 from volatility3.framework.renderers import TreeGrid
 from volatility3.framework.symbols import intermed
 from volatility3.framework.symbols.windows import extensions
+from volatility3.framework.layers import physical
 
+def rol(value: int, count: int) -> int:
```
ikelos: These two methods feel like they should live in a utility file somewhere, they don't seem specific to the physical layer.  Not a showstopper though...

volatility3/framework/plugins/windows/info.py
```diff
@@ -3,15 +3,32 @@
...
 from volatility3.framework.symbols import intermed
 from volatility3.framework.symbols.windows import extensions
+from volatility3.framework.layers import physical
 
+def rol(value: int, count: int) -> int: 
+    """A rotate-left instruction in Python"""
+    
+    for _ in range(count):
+        value *= 2
+        if (value > 0xFFFFFFFFFFFFFFFF):
```
ikelos: Not a huge fan of fixed sized constants, could we parameterize as `limit` or something this with a default of 64, and then use `(1 << limit) - 1`, just so it could be reused in other situations?

volatility3/framework/plugins/windows/info.py
```diff
@@ -3,15 +3,32 @@
...
+            value -= 0x10000000000000000
+            value += 1
+    return value
+
+def bswap(value: int) -> int:
+    """A byte-swap instruction in Python"""
+
+    hi, lo = struct.unpack(">II", struct.pack("<Q", value))
```
ikelos: Seems a little strange, why not `res, = unpack(">Q", struct.pack("Q", value))`?  Similarly, this could be parameterized with a struct type (`Q`)...

volatility3/framework/plugins/windows/info.py
```diff
@@ -87,6 +103,78 @@ def get_kdbg_structure(cls, context: interfaces.context.ContextInterface, config
...
+        ntkrnlmp = cls.get_kernel_module(context, layer_name, symbol_table)
+        if not ntkrnlmp.has_symbol("KdpDataBlockEncoded"):
+            return False
+
+        KdpDataBlockEncoded_value = ntkrnlmp.object("char", offset=ntkrnlmp.get_symbol("KdpDataBlockEncoded").address)
+
+        return KdpDataBlockEncoded_value != 0
+
+    @classmethod
+    def _decode_encoded_bytes(cls, context: interfaces.context.ContextInterface , offset: int,
```
ikelos: The method name should probably mention KDBG somewhere in there, so it's not just *any* encoded bytes.

volatility3/framework/symbols/windows/extensions/kdbg.py
```diff
@@ -5,13 +5,14 @@
...
 from volatility3.framework import objects
 
-
 class KDDEBUGGER_DATA64(objects.StructType):
 
     def get_build_lab(self):
         """Returns the NT build lab string from the KDBG."""
 
         layer_name = self.vol.layer_name
+        if self.vol.native_layer_name:
```
ikelos, line 14: Good catch, thanks!  5:)

volatility3/framework/symbols/windows/extensions/kdbg.py
```diff
@@ -32,5 +35,8 @@ def get_csdversion(self):
 
         return (csdresult >> 8) & 0xffffffff
 
+    def size(self):
```
ikelos: Seems a little redundant, it doesn't even check for pointer exceptions?  Since this is only used in one place, I'd just put it inline rather than pulling it out like this.

volatility3/framework/plugins/windows/writecrashdump.py
```diff
@@ -0,0 +1,134 @@
...
+        run0.PageCount.write(page_count)
+
+        dump_header.RequiredDumpSpace.write((page_count + 2) * 0x1000)
+
+        # We don't try any form of compression, but just write the data as one large run
+        with open_method('crash.dmp') as f:
+            # We want to include the maxmium address
+            header_data = header_layer.read(0, header_layer.maximum_address + 1)
+            f.write(header_data)
+            for offset in range(physical_layer.minimum_address, physical_layer.maximum_address + 1, 0x1000):
```
ikelos: Might be useful to use `progress_callback` to give the user some idea of how far through they are.

ikelos: Might be useful to use `progress_callback` to give the user some idea of how far through they are.

volatility3/framework/plugins/windows/writecrashdump.py
```diff
@@ -0,0 +1,134 @@
...
+            # Fix KDBG in the dump if it was encoded
+            decoded_data = context.layers[kdbg.vol.layer_name].read(kdbg.vol.offset, kdbg.size())
+            kdbg_physical_address = primary.translate(kdbg.vol.offset)[0]
+            kdbg_file_location = (header_layer.maximum_address + 1) + kdbg_physical_address - physical_layer.minimum_address
+            f.seek(kdbg_file_location)
+            f.write(decoded_data)
+
+    def _generator(self) -> Iterator[Tuple]:
+        self.write_crashdump(self.context, self.config['primary'], self.config['nt_symbols'], self._file_handler)
+        yield 0, ('Done',)
```
ikelos: Might be worth mentioning the output filename, just in case it becomes parameterized in the future, and slightly more useful than just done.  5;)

volatility3/framework/plugins/windows/info.py
```diff
@@ -87,6 +103,78 @@ def get_kdbg_structure(cls, context: interfaces.context.ContextInterface, config
...
+    @classmethod
+    def get_kdbg_structure(cls, context: interfaces.context.ContextInterface, config_path: str, layer_name: str,
+                           symbol_table: str) -> interfaces.objects.ObjectInterface:
+        kernel = cls.get_kernel_module(context, layer_name, symbol_table)
+        kdbg = cls.get_raw_kdbg_structure(context, config_path, layer_name, symbol_table)
+        primary = context.layers[kernel.layer_name]
+        tag_value = kdbg.Header.OwnerTag
+        is_kdbg_encoded = cls.is_kdbg_encoded(context, layer_name, symbol_table)
+
+        if not (is_kdbg_encoded and tag_value != b"KDBG"):
```
ikelos: A little confusing, could we please make this `if tag_value != b"KDBG" and not is_kdb_encoded:`?  Makes it easier to parse than the double negative...

volatility3/framework/plugins/windows/writecrashdump.py
```diff
@@ -0,0 +1,134 @@
...
+        if is_64_bit:
+            crashdump_json = 'crash64'
+            dump_header_name = '_DUMP_HEADER64'
+            valid_dump_suffix = [ord('6'), ord('4')]
+        else:
+            crashdump_json = 'crash'
+            dump_header_name = '_DUMP_HEADER'
+            valid_dump_suffix = [ord('M'), ord('P')]
+
+        config_path = 'whatever'
```
ikelos: Please don't do this, it theoretically could conflict and it definitely undermines good practice.  We don't want someone in the future thinking they can rely on that `config_path`.  Something like https://github.com/volatilityfoundation/volatility3/blob/280c91212140ceb320ee59c647378e28f919592e/volatility3/framework/symbols/generic/__init__.py#L21 would be much better (I'm happy with it being split it out into the API if that helps).

volatility3/framework/plugins/windows/info.py
```diff
@@ -87,6 +103,78 @@ def get_kdbg_structure(cls, context: interfaces.context.ContextInterface, config
...
+            decoded_buffer += struct.pack("Q", entry ^ wait_always)
+
+        return decoded_buffer
+
+    @classmethod
+    def _create_structure_from_bytes(cls, context: interfaces.context.ContextInterface, 
+                                     structure_bytes: bytes, structure_symbol_table: str, structure_name: str, 
+                                     offset: int, native_layer_name: str):
+
+        new_layer_name = f"{offset}_{structure_name}"
```
ikelos: I can't tell whether it's better to have the name start with a number of a string.  I feel drawn towards string first, since the offset is a more specific differentiator.  Otherwise this seems like quite a neat way of handling data modification.

volatility3/framework/plugins/windows/writecrashdump.py
```diff
@@ -0,0 +1,134 @@
...
+class WriteCrashDump(plugins.PluginInterface):
+    """Runs the automagics and writes the output to a crashdump format file"""
+
+    _required_framework_version = (2, 0, 0)
+    _version = (2, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.TranslationLayerRequirement(name = 'primary', description = 'Memory layer for the kernel',
```
ikelos: Do these two requirements need to be separate, or would making them a `ModuleRequirement` make more sense?  I'm kind of ok with either, but if there's no explicit reason for it, I'd prefer we keep things in sync and use a `ModuleRequirement`.

volatility3/framework/plugins/windows/writecrashdump.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @classmethod
+    def get_physical_layer_name(cls, context: interfaces.context.ContextInterface, vlayer: interfaces.layers.DataLayerInterface) -> str:
+        return context.config.get(interfaces.configuration.path_join(vlayer.config_path, 'memory_layer'), None)
+
+    @classmethod
+    def write_crashdump(cls, context: interfaces.context.ContextInterface, layer_name: str, symbol_table: str,
+                        open_method: Type[interfaces.plugins.FileHandlerInterface]):
+        kvo = context.layers[layer_name].config["kernel_virtual_offset"]
+        kernel = context.module(symbol_table, layer_name = layer_name, offset = kvo)
```
ikelos: This would become simpler and not need error checking if the kernel were a `ModuleRequirement` above, otherwise this should have checks for the kvo not existing

volatility3/framework/plugins/windows/writecrashdump.py
```diff
@@ -0,0 +1,134 @@
...
+        dump_header.NumberProcessors.write(number_processors)
+
+        dump_header.DirectoryTableBase.write(primary.config['page_map_offset'])
+        if dump_header.has_member("PaeEnabled"):
+            dump_header.PaeEnabled.write(int(is_pae))
+
+        dump_header.PfnDataBase.write(kdbg.MmPfnDatabase)
+        dump_header.PsLoadedModuleList.write(kdbg.PsLoadedModuleList)
+        dump_header.PsActiveProcessHead.write(kdbg.PsActiveProcessHead)
+        dump_header.DumpType.write(1)
```
ikelos: A comment to explain the value 1 here would be a nice-to-have, but not critical  5:)

volatility3/framework/plugins/windows/writecrashdump.py
ikelos: It's not yet updated, but I think the plan was to have `handler.preferred_filename` be updated to the correct filename when eventually written, so go with that please.

ikelos (commented):
Mostly awesome, a couple of bits and pieces around hardcoded values but otherwise really good!  Nice use of comments and sticking to the existing coding style, thanks!  5:)

volatility3/framework/plugins/windows/writecrashdump.py
paulkermann: added a comment

volatility3/framework/plugins/windows/writecrashdump.py
paulkermann: I have replaced the requirements with a module requirement

volatility3/framework/plugins/windows/writecrashdump.py
paulkermann: Replaced with ModuleRequirement

volatility3/framework/plugins/windows/info.py
paulkermann: replaced the order

volatility3/framework/plugins/windows/writecrashdump.py
paulkermann: added a progress callback

paulkermann: added a progress callback

volatility3/framework/symbols/windows/extensions/kdbg.py
paulkermann: removed it and added it inline

volatility3/framework/plugins/windows/info.py
paulkermann: improved the rol function.

volatility3/framework/plugins/windows/info.py
paulkermann: improved the bswap function

volatility3/framework/plugins/windows/info.py
paulkermann: I also thought about this. I think it is right to move them when they are needed elsewhere as well.

volatility3/framework/plugins/windows/info.py
paulkermann: I have renamed the function

volatility3/framework/plugins/windows/writecrashdump.py
ikelos: Turns out this was copied from my old pull request #472 and now I'm embarrassed that I was so undisciplined back in the day.  Entirely my bad, but it does need changing please.  5:)

paulkermann: I have replaced the `whatever` with a random

volatility3/framework/plugins/windows/info.py
```diff
@@ -166,6 +257,7 @@ def _generator(self):
         for i, layer in self.get_depends(self.context, layer_name):
             yield (0, (layer.name, f"{i} {layer.__class__.__name__}"))
 
+        yield (0, ("Is KDBG encoded: ", str(bool(self.is_kdbg_encoded(self.context, layer_name, symbol_table)))))
```
digitalisx: Because we put a delimiter in the rendering process (using a line or blank if we use Pretty), I think we don't need to put a separate delimiter (`:`) when we yield data.

```suggestion
        yield (0, ("IsKDBGEncoded", str(bool(self.is_kdbg_encoded(self.context, layer_name, symbol_table)))))
```

volatility3/framework/plugins/windows/info.py
```diff
@@ -3,15 +3,35 @@
...
 from volatility3.framework.symbols.windows import extensions
+from volatility3.framework.layers import physical
 
+def rol(value: int, count: int, max_bits: int = 64) -> int: 
+    """A rotate-left instruction in Python"""
+    max_bits_mask = (1 << max_bits) - 1
+    return (value << count % max_bits) & max_bits_mask |  \
+           ((value & max_bits_mask) >> (max_bits - (count % max_bits)))
+
+def bswap_32(value: int):
```
digitalisx: It would be nice if there is unity with `bswap_64`. :)

```suggestion
def bswap_32(value: int) -> int:
```

digitalisx (commented):
Hello, @paulkermann
I'm happy to see that the plugin is getting better through feedback, and I respect your work.
It's cool enough now, but I leave a NIT comment. 🙂

volatility3/framework/plugins/windows/info.py
```diff
@@ -3,15 +3,35 @@
...
+    value &= 0xffffffff
+    value = ((value << 8) & 0xFF00FF00) | ((value >> 8) & 0x00FF00FF)
+
+    return ((value << 16) | (value >> 16)) & 0xffffffff
+
+def bswap_64(value: int) -> int:
+    value &= 0xffffffffffffffff
```
ikelos: So we spent time changing `rol` not to use hardcoded limited numbers, but they've snuck back into bswap_32 and bswap_64 in a big way.  I'm also not sure why the doing pack and unpack with a changed endianness didn't achieve the same goal?  Could we go back to the pack/unpack method and use a flag to choose whether the struct should be `I` for 32 bit of `Q` for 64 please?

ikelos (commented):
Cool, `rol` looks good, but `bswap` seems to have taken a step backwards, I doesn't feel like it needs two separate methods, and it's not clear why it went from using struct to doing the maths manually, but I think that's the last bit to fix up (still pending a review from someone else from core, but otherwise looking good).  5:)

volatility3/framework/plugins/windows/info.py
digitalisx: This review may not be valid because the maintainer has proposed integration of functions below.

volatility3/framework/plugins/windows/info.py
paulkermann: replaced it

volatility3/framework/plugins/windows/info.py
paulkermann: added return type

volatility3/framework/plugins/windows/info.py
paulkermann: made it more readable

digitalisx (approved):
@paulkermann Thank you, I think the review I can advise is over.. 🙂
I hope the other things will be revised well and the maintainers will like it!

ikelos:
@iMHLv2 could you please take a look over the KDBG decoding bits of this?  I haven't done a full review yet, but the more eyes we can get on it the better please.  5:)

ikelos:
Any word on this one guys?

ikelos:
This is still waiting on review by @awalters I'm afraid.

paulkermann:
hey @ikelos, can you merge this plugin?

ikelos:
The need for review hasn't changed I'm afraid.  I don't know the crashdump format well enough to agree to bring this into core and support it long term.

PR #721, opened 2022-04-27
Linux threads Plugin

samuelzurowski:
Hello,

I know volatility2 had a threads plugin. I know in my previous PR mentioned was adding it to one of the Linux ps plugins. However, because I already had this done, I thought I would submit it to see if it's something volatility3 would want. (See https://github.com/volatilityfoundation/volatility3/pull/667)

image

Here are two usage examples. It includes offset, PID, COMM, Thread PID, Thread name, thread offset, and more properties depicted in the screenshot. 

Banner tested against: `Identified banner: b'Linux version 4.18.0-348.7.1.el8_5.x86_64 (mockbuild@kbuilder.bsys.centos.org) (gcc version 8.5.0 20210514 (Red Hat 8.5.0-4) (GCC)) #1 SMP Wed Dec 22 13:25:12 UTC 2021\n\x00`

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -14,6 +14,10 @@ capstone>=3.0.5
 # This is required by plugins that decrypt passwords, password hashes, etc.
 pycryptodome
 
+# This can improve error messages regarding improperly configured ISF files,
+# but is only recommended for development
+# jsonschema>=2.3.0
+
 # This is required for memory acquisition via leechcore/pcileech.
 leechcorepyc>=2.4.0
 


ikelos (approved):
Looks pretty straightforward to me, so just for @atcuno to check he's happy with it.

volatility3/framework/plugins/linux/threads.py
```diff
@@ -0,0 +1,67 @@
...
+from volatility3.framework.renderers import format_hints
+from volatility3.plugins.linux import pslist
+
+
+
+
+class Threads(interfaces.plugins.PluginInterface):
+    """Prints threads of a process"""
+
+    _required_framework_version = (2, 0, 0)
```
digitalisx: I think `task.get_threads` has been added in `2.0.4`, the minimum required framework version for the plugin is likely to be `2.0.4`.

```suggestion
    _required_framework_version = (2, 0, 4)
```

volatility3/framework/plugins/linux/threads.py
```diff
@@ -0,0 +1,67 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+from typing import Callable, Iterable, List, Any
+
+from volatility3.framework import renderers, interfaces
```
digitalisx: In PEP8, it is recommended that modules be sorted alphabetically.

```suggestion
from typing import Any, Callable, Iterable, List

from volatility3.framework import interfaces, renderers
```

volatility3/framework/plugins/linux/threads.py
```diff
@@ -0,0 +1,67 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
```
digitalisx: This is very trivial, but it would be nice if the license is changed to 2022.
```suggestion
# This file is Copyright 2022 Volatility Foundation and licensed under the Volatility Software License 1.0
```

digitalisx (commented):
@samuelzurowski Thank you for your contribution,
I left a few comments on the code.
However, maintainer's has already approved this PR, so feel free to check it out!

volatility3/framework/plugins/linux/threads.py
samuelzurowski: Minor - but I agree to commit the suggestion.

volatility3/framework/plugins/linux/threads.py
samuelzurowski: This is important good catch.

digitalisx (approved):
@samuelzurowski Thank you for considering my little comments!
I hope this will be a good plugin!

volatility3/framework/plugins/linux/threads.py
ikelos: Thanks, it would've been a great catch if I'd have done the version numbers properly  5:P  Sadly, I only did a patch bump, which doesn't get checked.  I should really have bumped the framework to (2, 1, 0) , and then this should depend on the framework being at least (2, 1, 0).  Sorry I didn't get it quite right, but very good catch @digitalisx, thanks!  5:)

ikelos: I've now bumped the develop branch to `(2, 1, 0)`.

digitalisx: Ah, I saw your README change and found out that the version should be `2.0.4`. Thank you again for checking @ikelos, I hope my review is helpful. 🙂

samuelzurowski: > I've now bumped the develop branch to `(2, 1, 0)`.

oops, let me change to 2.1.0
So should `_required_framework_version = (2,1,0)`?

What about `_version` and other parameters like pslist's version? I just want to make sure I understand the versioning methodology correctly.

ikelos: Sure, so roughly it goes as follows:
```Framework <-- _required_framework_version -| plugin <-- _version -| other_plugin```
Your plugin needs the framework to be at least `_required_framework_version` to work, and other plugins can make a `VersionRequirement` on your plugin, saying they need your `_version` to be at least a particular value.  If you haven't changed the way your plugin works, you don't need to do anything.  If you've changed the interface that your plugin exposes, you need to bump MAJOR if it's a breaking change, MINOR if it's an addition and if you just want to tell that a change was made for your records, but it has no external impact, you can bump PATCH.

Hopefully that all makes sense?

samuelzurowski: > Sure, so roughly it goes as follows: `Framework <-- _required_framework_version -| plugin <-- _version -| other_plugin` Your plugin needs the framework to be at least `_required_framework_version` to work, and other plugins can make a `VersionRequirement` on your plugin, saying they need your `_version` to be at least a particular value. If you haven't changed the way your plugin works, you don't need to do anything. If you've changed the interface that your plugin exposes, you need to bump MAJOR if it's a breaking change, MINOR if it's an addition and if you just want to tell that a change was made for your records, but it has no external impact, you can bump PATCH.
> 
> Hopefully that all makes sense?

Okay, I thought so. Just wanted to double-check thanks :)

paulkermann:
Aren't you missing out on single threaded processes? 
We should add `sentinel=False` to the input parameter of the `to_list` function in the `get_threads`.

PR #722, opened 2022-04-27
Fix: typo for `API_CHANGES`.

digitalisx:
Hello, everone in the community! 🙂
A typo was found in the process of updating `API_CHANGES`.

### DIFF
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -6,7 +6,7 @@ When an API feature or function is removed or changed, the major version is bump
 
 2.1.0
 =====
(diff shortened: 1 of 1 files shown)


ikelos:
Heheheh, thanks for that, sadly I'm losing sensation in the fingers in my left hand, so there may be more typos like that but always happy to hear about them if they happen!  5:)

digitalisx:
@ikelos That's okay, any mistake is fine if there's a community to review. I think you will be busy because there are so many requests and reviews these days. I think it would be good if you could continue to be with the community with ease. And I wish you good health!

PR #725, opened 2022-05-05
Add: python virtual environments, memory dump on `.gitignore`.

digitalisx:

When we develop or use Volatility3, it is sometimes difficult because git tracks unintended changes.

A case in point is as follows.
- Use `venv` for install modules
- We can also put the memory dump files to analyze into the volatility folder or create them through plugins.

I've seen cases where contributors who develop Volatility3, including myself, set up and develop themselves in `.gitignore` or accidentally reflect them in PR.
I think updating `.gitignore` will reduce the effort of various users.

I also reflected Python's basic `.gitignore` data officially provided by Github to reflect the most common case.

What do you think of this PR?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,8 +34,6 @@ dist
 env/
 venv/
 ENV/
-env.bak/
-venv.bak/
 
 # Memory dump files
 *.dmp
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,16 @@ config*.json
 # Pyinstaller files
 build
 dist
+
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+
+# Memory dump files
+*.dmp
+*.vmem


.gitignore
```diff
@@ -27,3 +27,16 @@ config*.json
...
 build
 dist
+
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
```
ikelos: These seem quite specific as backup files, will other venv tools create different extensions?  I'm wondering whether we just block `*.bak` or `env.*`...

.gitignore
```diff
@@ -27,3 +27,16 @@ config*.json
...
+venv.bak/
+
+# Memory dump files
+*.dmp
```
ikelos, line 39: This isn't standardized, but I'll allow it since it's common, I just don't want to be adding every file extension someone thinks should be ignored.

ikelos (commented):
So, this is ok, but it's also a slippery slope (where we keep getting suggestions or variants for everyone's personal circumstances).  These are pretty common, so I'm ok with it, but I'm nervous about the `.bak` files, I'm not convinced that tools make those or that they should be ignored?

I'll give it a few days in case you want to make any changes, but if not then I'll commit it.

.gitignore
digitalisx: I've researched a lot of cases. 
`.bak` seems to be the type of folder and file name you specify when you want to back up. It's not generated by the program. `env.` I think it's better to take the form. I think this is a little bit more general and it could cover a lot of forms.

digitalisx: However, I think the dots behind the `env` can be used to distinguish various deployment stages such as `.bak`, `.prod`, `.dev`, etc. Since volatility3 is a difficult case to use right away, I think ignoring the `env` `venv` folder is an improvement.

ikelos: Yeah, that's fair enough.  Thanks for making the changes  5:)

ikelos (approved):
Yep, looks ok, thanks  5:)

PR #730, opened 2022-05-09
Removed unused svcscan import from Callbacks plugin

paulkermann:
The `svcscan` plugin is imported but not used in the `Callbacks` plugin.


ikelos:
Huh, looks like that's been in there for ages, good catch, thanks.  5:)

PR #733, opened 2022-05-10
Fix: typo for `API_CHANGES`, code comment.

digitalisx:
I found some typo while analyze documents and codes.
And I updated `API_CHANEGS` comment for unification.

### DIFF
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -6,15 +6,15 @@ When an API feature or function is removed or changed, the major version is bump
 
 2.1.0
 =====
-Add in the linux `task.get_threads` method added to the API.
+Add in the linux `task.get_threads` method to the API.
 
 2.0.3
 =====
-Add in the windows `DEVICE_OBJECT.get_attached_devices` and `DRIVER_OBJECT.get_devices` method added to the API.
+Add in the windows `DEVICE_OBJECT.get_attached_devices` and `DRIVER_OBJECT.get_devices` methods to the API.
 
 2.0.2
 =====
-Fix the behavior of the offsets returned by the PDB scanner.
+Fix the behaviour of the offsets returned by the PDB scanner.
 
 2.0.0
 =====
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -10,11 +10,11 @@ Add in the linux `task.get_threads` method added to the API.
 
 2.0.3
 =====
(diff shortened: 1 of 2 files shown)


API_CHANGES.md
```diff
@@ -10,11 +10,11 @@ Add in the linux `task.get_threads` method added to the API.
 
 2.0.3
 =====
-`DEVICE_OBJECT.get_attached_devices` and `DRIVER_OBJECT.get_devices` added to the API.
+Add in the windows `DEVICE_OBJECT.get_attached_devices` and `DRIVER_OBJECT.get_devices` method added to the API.
 
 2.0.2
 =====
-Fix the behaviour of the offsets returned by the PDB scanner.
+Fix the behavior of the offsets returned by the PDB scanner.
```
ikelos: This is an American spelling, rather than a typo.  I'm not intending to ensure all documentation is consistent between Amertican/British english, so I'll include this, but if it were the only change I'd reject it.

API_CHANGES.md
```diff
@@ -10,11 +10,11 @@ Add in the linux `task.get_threads` method added to the API.
 
 2.0.3
 =====
-`DEVICE_OBJECT.get_attached_devices` and `DRIVER_OBJECT.get_devices` added to the API.
+Add in the windows `DEVICE_OBJECT.get_attached_devices` and `DRIVER_OBJECT.get_devices` method added to the API.
```
ikelos: Technically there's two items mentioned, so the number of `method` should be plural (`methods`).  Also "Add in the... added to the API" doesn't make much sense.  Happy to make it clearer, but best we change properly so there isn't another follow-up pull request.  5;)

I'd go with:

Add in the windows  `DEVICE_OBJECT.get_attached_devices` and `DRIVER_OBJECT.get_devices` methods to the API.

ikelos (commented):
Thanks for this, one good catch, one close catch and one that's changing a British spelling to an American spelling.  The close catch needs changing up please...  5:)

API_CHANGES.md
digitalisx: All right, this is often the case with global projects, I think I was not familiar with British English. This case is not wrong, so I will revert it.

API_CHANGES.md
digitalisx: All right, I'll change the sentence below the same.

digitalisx:
@ikelos Thank you for your review, I fixed them!

ikelos (approved):
Cool, looks good, thanks!  5:)

PR #734, opened 2022-05-11
Fix: `key_path` and `top_key` compare logic bug for `Certificates` plugin.

digitalisx:
## Description
Hello, everyone in the community! 🙂
An error was detected while using `Certificates`, a plugin useful for obtaining certificate data from the registry.

### Commands
```shell
> python3 vol.py -f case.vmem -r pretty  windows.registry.certificates
Volatility 3 Framework 2.2.0
Formatting...0.00		PDB scanning finished
```

### Error Log
```shell
Traceback (most recent call last):
  File "/Users/donghyunkim/Desktop/volatility3/vol.py", line 10, in <module>
    volatility3.cli.main()
    ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/donghyunkim/Desktop/volatility3/volatility3/cli/__init__.py", line 635, in main
    CommandLine().run()
    ^^^^^^^^^^^^^^^^^^^
  File "/Users/donghyunkim/Desktop/volatility3/volatility3/cli/__init__.py", line 342, in run
    renderers[args.renderer]().render(constructed.run())
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/donghyunkim/Desktop/volatility3/volatility3/cli/text_renderer.py", line 177, in render
    grid.populate(visitor, outfd)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/donghyunkim/Desktop/volatility3/volatility3/framework/renderers/__init__.py", line 212, in populate
    for (level, item) in self._generator:
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/donghyunkim/Desktop/volatility3/volatility3/plugins/windows/registry/certificates.py", line 58, in _generator
    unique_key_offset = key_path.index(top_key) + len(top_key) + 1
                        ^^^^^^^^^^^^^^^^^^^^^^^
ValueError: substring not found
```

As a result of tracking the error, the `key_path` and `top_key` extracted from the memory dump were not properly compared. (substring not found)

```
Level 7  volatility3.plugins.windows.registry.certificates: key_path(\??\C:\Users\dhyun\NTUSER.dat\SOFTWARE\Microsoft\SystemCertificates\Root\ProtectedRoots), top_key(Software\Microsoft\SystemCertificates)
[...]

PR #735, opened 2022-05-12
Plugin: update plugin version, logger, dump options of `Certificates` plugin.

digitalisx:
## Description

As I modified this PR (#734), I checked and updated some improvements to this plugin.

-  Most plugins of Volatility3 can choose the `dump` option if they have data that can be dumped or if they write it to a local file system. However, the `Certificates` plugin does not match the plugin's description that when it runs, it automatically dumps and simply lists against our intentions. So I added the dump option.
- However, given existing plugin improvements, radical changes to certain features and commands can be a problem for other current users and systems. So I added a warning log and encouraged you to use the option. This will be removed one day. (Please let me know if you don't need this step, and I will modify it.)
- The plugin does not have a `KeyError`, so I added it because does not show a log for the case that went over.
- I changed the value to underscore (`_`) for the unused value in the iterator statement.

### DIFF
--- a/volatility3/plugins/windows/registry/certificates.py
+++ b/volatility3/plugins/windows/registry/certificates.py
@@ -1,6 +1,6 @@
 import logging
 import struct
-from typing import List, Iterator, Tuple, Type
+from typing import List, Iterator, Optional, Tuple, Type
 
 from volatility3.framework import constants, exceptions, interfaces, renderers
 from volatility3.framework.configuration import requirements
@@ -45,7 +45,7 @@ class Certificates(interfaces.plugins.PluginInterface):
     def dump_certificate(cls, certificate_data: bytes, hive_offset: int,
                         reg_section: str, key_hash: str,
                         open_method: Type[interfaces.plugins.FileHandlerInterface]) -> \
-                        interfaces.plugins.FileHandlerInterface:
+                        Optional[interfaces.plugins.FileHandlerInterface]:
         try:
             if not isinstance(certificate_data, interfaces.renderers.BaseAbsentValue):
                 dump_name = "{} - {} - {}.crt".format(hive_offset, reg_section, key_hash)
--- a/volatility3/plugins/windows/registry/certificates.py
+++ b/volatility3/plugins/windows/registry/certificates.py
@@ -90,9 +90,6 @@ class Certificates(interfaces.plugins.PluginInterface):
                 except exceptions.SwappedInvalidAddressException as exp:
                     vollog.log(constants.LOGLEVEL_VVVV, f"Required memory at {exp.invalid_address:#x} is inaccessible (swapped)")
                     pass
-                except exceptions.PagedInvalidAddressException as exp:
-                    vollog.log(constants.LOGLEVEL_VVVV, f"Required memory at {exp.invalid_address:#x} is not valid (process exited?)")
-                    pass
 
     def run(self) -> renderers.TreeGrid:
         return renderers.TreeGrid([("Certificate path", str), ("Certificate section", str), ("Certificate ID", str),
--- a/volatility3/plugins/windows/registry/certificates.py
+++ b/volatility3/plugins/windows/registry/certificates.py
@@ -47,7 +47,7 @@ class Certificates(interfaces.plugins.PluginInterface):
                         Optional[interfaces.plugins.FileHandlerInterface]:
         try:
             if not isinstance(certificate_data, interfaces.renderers.BaseAbsentValue):
-                dump_name = "{} - {} - {}.crt".format(hive_offset, reg_section, key_hash)
+                dump_name = "{}-{}-{}.crt".format(hive_offset, reg_section, key_hash)
                 file_handle = open_method(dump_name)
                 file_handle.write(certificate_data)
                 return file_handle
--- a/volatility3/plugins/windows/registry/certificates.py
+++ b/volatility3/plugins/windows/registry/certificates.py
@@ -1,3 +1,4 @@
+import contextlib
 import logging
 import struct
 from typing import List, Iterator, Optional, Tuple, Type
@@ -67,7 +68,7 @@ class Certificates(interfaces.plugins.PluginInterface):
                     "Microsoft\\SystemCertificates",
                     "Software\\Microsoft\\SystemCertificates",
             ]:
-                try:
+                with contextlib.suppress(KeyError, exceptions.SwappedInvalidAddressException):
                     # Walk it
                     node_path = hive.get_key(top_key, return_list = True)
                     for (_depth, is_key, _last_write_time, key_path, _volatility, node) in printkey.PrintKey.key_iterator(hive, node_path, recurse = True):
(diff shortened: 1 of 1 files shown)


volatility3/plugins/windows/registry/certificates.py
```diff
@@ -1,16 +1,19 @@
...
 from volatility3.framework.symbols.windows.extensions.registry import RegValueTypes
 from volatility3.plugins.windows.registry import hivelist, printkey
 
+vollog = logging.getLogger(__name__)
 
 class Certificates(interfaces.plugins.PluginInterface):
     """Lists the certificates in the registry's Certificate Store."""
 
     _required_framework_version = (2, 0, 0)
+    _version = (1, 0, 0)
```
ikelos: This doesn't need a version number because it doesn't expose an interface (classmethods expected to be used by other plugins).  It's fine, it doesn't need to go, just to say that it's not strictly required.

ikelos: If you had wanted to make `dump_data` into part of the API then it probably shouldn't live here if we want to be generic, or it should be named `dump_certificate` or `dump_registry_binary` since they're more descriptive if found in another piece of code far away.  5:)

In both instances it would need to be a `classmethod` meaning it would need to take an `open_method` or similar (refer to other `dump` classmethods for consistency).  5:)

volatility3/plugins/windows/registry/certificates.py
```diff
@@ -48,21 +61,23 @@ def _generator(self) -> Iterator[Tuple[int, Tuple[str, str, str, str]]]:
                 try:
                     # Walk it
                     node_path = hive.get_key(top_key, return_list = True)
-                    for (depth, is_key, last_write_time, key_path, volatility,
-                         node) in printkey.PrintKey.key_iterator(hive, node_path, recurse = True):
+                    for (_, is_key, _, key_path, _, node) in printkey.PrintKey.key_iterator(hive, node_path, recurse = True):
```
ikelos: I *slightly* prefer these to be named (even just `_depth`) because whilst it does make the line longer, it means that anyone referring/copy the line doesn't have to find the original call to know what those hidden returned fields are.  Not a big quibble, but I thought I'd mention it.

ikelos (commented):
Thanks, this mostly seems fine, a couple of minor points, nothing that'll stop it getting merged, but I'll leave a few days in case you would like to make changes, before I merge it.

Just so I document this in as many places as possible (and I know you've already heard this in #734): This plugin is specifically under `volatility3/plugins` rather than `volatility3/framework/plugins` because it's an example of a plugin we've had submitted that we're *not* intending to explicitly support, update or maintain.  This is so that people know the purpose of the `volatility3/plugins` directory, and how they should use it for their own plugins rather than modifying the core ones in `volatility3/framework`.  This is also the reason all imports should be for `volatility3.plugins` rather than `volatility3.framework.plugins`.  I feel we should probably document this somewhere more clearly, but since it came up here, here's where I'm putting the explanation for now...  5;)

digitalisx:
> Thanks, this mostly seems fine, a couple of minor points, nothing that'll stop it getting merged, but I'll leave a few days in case you would like to make changes, before I merge it.
> 
> Just so I document this in as many places as possible (and I know you've already heard this in #734): This plugin is specifically under `volatility3/plugins` rather than `volatility3/framework/plugins` because it's an example of a plugin we've had submitted that we're _not_ intending to explicitly support, update or maintain. This is so that people know the purpose of the `volatility3/plugins` directory, and how they should use it for their own plugins rather than modifying the core ones in `volatility3/framework`. This is also the reason all imports should be for `volatility3.plugins` rather than `volatility3.framework.plugins`. I feel we should probably document this somewhere more clearly, but since it came up here, here's where I'm putting the explanation for now... 5;)

@ikelos First of all, thank you for taking the time to review.

Oh, that's right, I remember the comment you told me last time.
I understood that this plugin is officially the plugin that we keep, but it went into that folder as an example.
Please understand my inexperience.. hehe 😅

All right... Now that I understand the intentions, I probably won't be asking for a review or PR for the code in that path.
But this time, thankfully, you proceeded with the review!

I think it's a good opportunity to give a little better user experience to some users who don't understand the intentions. 🙌
(The plugins shown in the help command do not distinguish them, so the average user thinks that this plugin is also an official plug-in and likely used as an official plugin.)

If you give me time to work on the review, I will solve it and request the review again. 🙂

ikelos:
Sure, there's no rush!  5:)  You're right it doesn't really make the distinction to the user, but it was more intended for developers to be aware.  It is really as an example for others, so I'm happy to update it, I just wanted to set people's expectations that it might not get as much love as plugins we've accepted into core.  5:)

volatility3/plugins/windows/registry/certificates.py
```diff
@@ -63,10 +69,11 @@ def _generator(self) -> Iterator[Tuple[int, Tuple[str, str, str, str]]]:
...
                             if self.config['dump']:
-                                if not isinstance(certificate_data, interfaces.renderers.BaseAbsentValue):
-                                    with self.open("{} - {} - {}.crt".format(hex(hive.hive_offset), reg_section,
-                                                                            key_hash)) as file_data:
-                                        file_data.write(certificate_data)
+                                self.dump_data(certificate_data, hive.hive_offset, reg_section, key_hash)
+                            else:
+                                vollog.warning("Certificates plugin is no longer support automatically dumped, please use the dump option.")
+                                self.dump_data(certificate_data, hive.hive_offset, reg_section, key_hash)
+
```
digitalisx: Originally, there was a code that considered side effects because the dump was automatic, but now required options. However, I wondered if it was appropriate because it was not an official plugin and was unlikely to be managed in the future. Do you think it's necessary?

digitalisx (commented):
@ikelos Thank you, I was able to reflect it quickly because most of the suggestions gave me a reasonable and detailed guide. I left one comment for further discussion.

digitalisx:
After reviewing the code again, I forgot to return the file handle and specify the exact return type.
I will check it carefully again before requesting a review.. 😅

volatility3/plugins/windows/registry/certificates.py
ikelos: Hmmmm, I don't think this is the best way of handling this.  It's a bit strange to say it won't dump the certificates and then dump them anyway, and it will also likely be annoying to people to get warned that you need to turn on a flag, given that's how the plugin (now) works.  I understand the point of informing the user, but I think it'll confuse people in the long run, than the one off surprise of finding out the plugin has changed.

It might help if the output were more significantly changed.  As I say, we don't intend to support this, so this is a purely a suggestion not a requirement for this pull request, but the plugin could try loading the certificate and try extracting the common name?  It can just return an AbsentValue if it can't find read it, but it might be slightly more helpful for users and make the non-dump option more useful?

ikelos (approved):
Yep, looks good, thanks!  5:)  Let me know if you want it merged, or want to do more work on it...

digitalisx:
@ikelos Thank you for your review and suggestion, I will work on your suggestion.
However, due to the schedule, it may take some time. 🙂

volatility3/plugins/windows/registry/certificates.py
digitalisx: @ikelos Thank you for your opinion! Could you explain in more detail?
I think I don't understand all the meaning of my English skills. :(

digitalisx:
While I was working on the documentation a recent, I found that some of the results I worked on had an old form.
Therefore, I changed this to ensure that our code is consistent.
Additionally, exception handling has been enhanced to allow users to see more results.

volatility3/plugins/windows/registry/certificates.py
digitalisx: If you don't have any additional comments on this, you can merge it.
Rather, I think it's an opportunity to focus on what I wrote in other draft PR. :)

digitalisx:
@ikelos  I'm sorry I couldn't keep my word that I would work on additional comments.
Could you merge this when you have time?
I am satisfied with the current version as well as I intended.. :)
I'm going to focus on the remaining Drafts PR and the ones with a little higher priority.

ikelos (approved):
Yep, looks good, thanks  5:)

PR #736, opened 2022-05-14
Fix: `JSON` renderer for `EOF` format.

digitalisx:
## Description

We have a wonderful ability to convert the results of plug-ins used in Volatility3 to `JSON`.
Using this function, some terminals such as `zsh`, did not have a newline(`\n`) indicating `EOF`, so it was confirmed that unnecessary data was printed on the output.
Other renderers may have already been processed during the rendering process or can be resolved using the `print` function, but we have used `outfd.write`, so I have fixed it accordingly.

```shell
            "Offset(V)": 2646155392,
            "PID": 2892,
            "PPID": 2860,
            "SessionId": 1,
            "Threads": 71,
            "Wow64": false,
            "__children": []
          }
        ]
      }
    ]
  }
]%
```

## Result

```shell
            "Offset(V)": 2646155392,
            "PID": 2892,
            "PPID": 2860,
            "SessionId": 1,
            "Threads": 71,
            "Wow64": false,
            "__children": []
          }
        ]
      }
    ]
  }
]
```

### DIFF
--- a/volatility3/cli/text_renderer.py
+++ b/volatility3/cli/text_renderer.py
@@ -345,7 +345,7 @@ class JsonRenderer(CLIRenderer):
 
     def output_result(self, outfd, result):
         """Outputs the JSON data to a file in a particular format"""
-        outfd.write(json.dumps(result, indent = 2, sort_keys = True))
+        outfd.write("{}\n".format(json.dumps(result, indent = 2, sort_keys = True)))
 
     def render(self, grid: interfaces.renderers.TreeGrid):
         outfd = sys.stdout


volatility3/cli/text_renderer.py
```diff
@@ -345,7 +345,7 @@ def get_render_options(self) -> List[interfaces.renderers.RenderOption]:
 
     def output_result(self, outfd, result):
         """Outputs the JSON data to a file in a particular format"""
-        outfd.write(json.dumps(result, indent = 2, sort_keys = True))
+        outfd.write("{}\n".format(json.dumps(result, indent = 2, sort_keys = True)))
```
ikelos, line 348: We do now allow format strings (so `outfd.write(f"{json.dumps(result, indent = 2, sort_keys = True)}\n")` or even just `+ "\n"` but this is fine...  5;)

ikelos (approved):
Yep, looks fine, thanks.  5:)

volatility3/cli/text_renderer.py
digitalisx: Thank you for the good information! 🙌

PR #738, opened 2022-05-15
Remove: unused module for windows `extensions`, `objects` initialize code.

digitalisx:
I found a module that doesn't seem to be used in some codes. If we really don't use it, I think it's okay to remove it.

### DIFF
--- a/volatility3/framework/objects/__init__.py
+++ b/volatility3/framework/objects/__init__.py
@@ -9,7 +9,7 @@ import struct
 from typing import Any, ClassVar, Dict, Iterable, List, Optional, Tuple, Type, Union as TUnion, overload
 
 from volatility3.framework import constants, interfaces
-from volatility3.framework.objects import templates, utility
+from volatility3.framework.objects import templates
 
 vollog = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


ikelos:
Thanks for this, however there are certain places where we import the higher level `extensions` or `objects` for clarity, and expect things like `extensions.kdbg` to be imported automatically (for example: https://github.com/volatilityfoundation/volatility3/blob/0a655d76a8e2d2082bf2df428faff22dda4390da/volatility/framework/plugins/windows/info.py#L84).  This may not be the best way of doing things, but `__init__` file sometimes import submodules to ensure they're accessible directly from the top level.  We don't do this everywhere (for example `volatility3` or `volatility3.framework`) but there are situations where it improves readability.

So, totally fine changes, no real issue with the request, but I'm afraid I'm not going to merge it because automatic importing is pretty useful in these instances (it creates a more descriptive namespace of `extensions.kdbg` rather than just `kdbg` which we might want as a module name in numerous different situations).

digitalisx:
@ikelos Thank you, I don't think I was used to using modules this way. I think I can understand the intentions when I look at the code in the future. Thank you for taking the time to give me good knowledge. 🙂

ikelos:
No problem, we value your contributions so wanted to make sure you understood it reasoning (and had an opportunity to talk it through if you thought the reasoning wasn't that good).  Thanks for accepting it!  5:)

Issue #713, opened 2022-04-24
Off by 1 from using the get_end and get_start functions


paulkermann:
https://github.com/volatilityfoundation/volatility3/blob/7d17c191e5cfacc84e0d355b18dcf6c7ce9f68d0/volatility3/framework/plugins/windows/malfind.py#L59

This should actually be `vad_length = vad.get_end() - vad.get_start() + 1`

`get_end` is the last **valid** address _including_ so there is an off by 1. Reading `vad_length` would yield 1 less byte.
I would add a `get_size` function that would return `vad.get_end() - vad.get_start() + 1` to avoid problems like this in the future.
This problem is also present here https://github.com/volatilityfoundation/volatility3/blob/0c603a20ba877795b3da49f6a7ae0f5721d23e7d/volatility3/framework/plugins/windows/vadinfo.py#L154 as another off by 1 as vad_end is readable.

It also looks like this wrong logic happens in other places. One could also maybe just not subtract 1 from `get_end` but that could probably cause compatibility issues so the first solution is probably better.
However it looks like everybody is using `get_end` as it is not subtracted by 1 so I don't really know.

iMHLv2:
@paulkermann Thanks for the bug report. I pushed a branch with a proposed fix. Historically, even back to vol2, the `get_end()` function has returned the last accessible byte in the range, so I'm keeping that unchanged. I added a `get_size()` method that calculates the total size in bytes and fixed a number of plugins that had off-by-one issues. Anyone relying on `get_end()` to calculate size in custom plugins (outside of the core code base) will still have one-by-one issues, but they can fix it by using the new `get_size()` - IMO that is better than changing the behavior of `get_end()` and potentially breaking things "silently."

digitalisx:
Hello @iMHLv2, Thank you for your good work!
It seems that it is not pushed to the development branch and exists only in issue branch.
Could you tell me the intend or reason why not convert to PR? 🙂
(Additional this changes may require a bump version.)

iMHLv2:
> Hello @iMHLv2, Thank you for your good work! It seems that it is not pushed to the development branch and exists only in issue branch. Could you tell me the intend or reason why not convert to PR? 🙂 (Additional this changes may require a bump version.)

PR is in. I just created a branch initially so people could see the changes and experiment with it.

ikelos:
Looks like this should have been closed when #818 landed.  Feel free to reopen if the issue hasn't been resolved.

PR #741, opened 2022-05-22
Fix off by 1 in Filelayer MaximumAddress

paulkermann:
https://github.com/volatilityfoundation/volatility3/blob/9bfa7187daed41791217e29afb828a5a5ab58e5b/volatility3/framework/layers/physical.py#L115

Closely related to #713.
Maximum address is the last valid address.
File size was returned and not file size minus 1.

### DIFF
--- a/volatility3/framework/layers/physical.py
+++ b/volatility3/framework/layers/physical.py
@@ -118,7 +118,7 @@ class FileLayer(interfaces.layers.DataLayerInterface):
         with self._lock:
             orig = self._file.tell()
             self._file.seek(0, 2)
-            self._size = self._file.tell()
+            self._size = self._file.tell() - 1
             self._file.seek(orig)
         return self._size
 
--- a/volatility3/framework/layers/physical.py
+++ b/volatility3/framework/layers/physical.py
@@ -88,6 +88,7 @@ class FileLayer(interfaces.layers.DataLayerInterface):
         self._accessor = resources.ResourceAccessor()
         self._file_: Optional[IO[Any]] = None
         self._size: Optional[int] = None
+        self._maximum_address: Optional[int] = None
         # Construct the lock now (shared if made before threading) in case we ever need it
         self._lock: Union[DummyLock, threading.Lock] = DummyLock()
         if constants.PARALLELISM == constants.Parallelism.Threading:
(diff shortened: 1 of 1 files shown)


volatility3/framework/layers/physical.py
```diff
@@ -118,7 +118,7 @@ def maximum_address(self) -> int:
         with self._lock:
             orig = self._file.tell()
             self._file.seek(0, 2)
-            self._size = self._file.tell()
+            self._size = self._file.tell() - 1
```
ikelos: Thanks, since this is no longer the size, could we change this variable name to _maximum_address please?

ikelos (commented):
Thanks, I didn't realize seeking to the end move to the location after the last byte of the file.  This looks good, but for clarity I'd like to change the variable name please.

digitalisx:
@ikelos I wonder if this PR also requires minor version bump. 🙂 (same question #743)

ikelos:
I considered it, but I don't think anyone's been relying specifically on that value being one higher.  The API hasn't really changed, it's just returning a more accurate value, but otherwise doing exactly what the documentation says it should have, so I chose to leave it.

volatility3/framework/layers/physical.py
paulkermann: name changed

volatility3/framework/layers/physical.py
```diff
@@ -113,14 +113,15 @@ def _file(self) -> IO[Any]:
     def maximum_address(self) -> int:
         """Returns the largest available address in the space."""
         # Zero based, so we return the size of the file minus 1
-        if self._size:
-            return self._size
+        if self._maximum_address
```
digitalisx: Hello @paulkermann, Is it valid without `:` syntax for `if` statements?
```suggestion
        if self._maximum_address:
```

paulkermann: fixed it

ikelos (approved):
Looks good, thanks  5:)

ikelos:
Oh, hang on, could you initialize `_maximum_address` in the constructor please?  I thought you'd just renamed `_size`, not added a new variable.

paulkermann:
wanted to remain backward compatible so I left the old variable.

ikelos (approved):
Awesome, thanks  5:)

volatility3/framework/layers/physical.py
digitalisx: Thank you always for your quick response. :)

PR #745, opened 2022-05-23
Log resource cache usage

paulkermann:
So it's kinda my bad but I was debugging for like 30 minutes trying to understand why a file I have accessed and changed remotely caused the same results locally.
Just added a debug log to say the cache was used.


ikelos (approved):
Looks ok, sorry it wasn't clearer!

PR #749, opened 2022-05-29
Check return value from is_valid

paulkermann:
`is_valid` returns a `bool`. 
That `bool` is ignored in the `_LIST_ENTRY` parsing function causing invalid entries to be yielded.

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -746,7 +746,10 @@ class LIST_ENTRY(objects.StructType, collections.abc.Iterable):
         trans_layer = self._context.layers[layer]
 
         try:
-            trans_layer.is_valid(self.vol.offset)
+            is_valid = trans_layer.is_valid(self.vol.offset)
+            if not is_valid:
+                return
+
             link = getattr(self, direction).dereference()
         except exceptions.InvalidAddressException:
             return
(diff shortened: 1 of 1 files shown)


volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -762,7 +765,9 @@ def to_list(self,
             obj_offset = link.vol.offset - relative_offset
 
             try:
```
ikelos: I think it used to throw an exception, which if instead it just returns a `bool` means that this exception seems redundant now?  The one above may cover the dereference, but this is definitely just for the `is_valid` call.

ikelos (commented):
Thanks, I forget when we changed this from an exception to a `bool`, but I guess this got missed.  Looks fine, but makes one of the `try/catch` statements redundant.

volatility3/framework/symbols/windows/extensions/__init__.py
paulkermann: Some layers throw and exception, some don't.
I can remove the try catch if you want

ikelos: Well, the point of `is_valid` is that you don't need to do the `try/except`, so I'd prefer to find anything where `is_valid` throws an exception, and instead get it to catch and throw false instead.

ikelos:
Thanks, looks good.  5:)

PR #750, opened 2022-05-29
Add: JSON `EOF` for `config` file. (`cli`, `volshell`)

digitalisx:
## Description

This PR is related to #736.
When I dump the config file in JSON form and read it on the terminal, I checked that EOF does not exist.
So I added EOF data for _volatility_ `cli` and `volshell`

```
  "kernel.symbol_table_name.class": "volatility3.framework.symbols.windows.WindowsKernelIntermedSymbols",
  "kernel.symbol_table_name.isf_url": "file:///Users/donghyunkim/Desktop/volatility3/volatility3/symbols/windows/ntkrpamp.pdb/78C36230DCC38A999E9C6C00C3EC5823-1.json.xz",
  "kernel.symbol_table_name.symbol_mask": 0,
  "physical": false,
  "pid": []
}%
```

## Result

```
  "kernel.symbol_table_name.class": "volatility3.framework.symbols.windows.WindowsKernelIntermedSymbols",
  "kernel.symbol_table_name.isf_url": "file:///Users/donghyunkim/Desktop/volatility3/volatility3/symbols/windows/ntkrpamp.pdb/78C36230DCC38A999E9C6C00C3EC5823-1.json.xz",
  "kernel.symbol_table_name.symbol_mask": 0,
  "physical": false,
  "pid": []
}
```

### DIFF
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -332,6 +332,7 @@ class CommandLine:
                     parser.error(f"Cannot write configuration: file {args.save_config} already exists")
                 with open(args.save_config, "w") as f:
                     json.dump(dict(constructed.build_configuration()), f, sort_keys = True, indent = 2)
+                    f.write("\n")
         except exceptions.UnsatisfiedException as excp:
             self.process_unsatisfied_exceptions(excp)
             parser.exit(1, f"Unable to validate the plugin requirements: {[x for x in excp.unsatisfied]}\n")
--- a/volatility3/cli/volshell/__init__.py
+++ b/volatility3/cli/volshell/__init__.py
@@ -246,6 +246,7 @@ class VolShell(cli.CommandLine):
                     parser.error(f"Cannot write configuration: file {args.save_config} already exists")
                 with open(args.save_config, "w") as f:
                     json.dump(dict(constructed.build_configuration()), f, sort_keys = True, indent = 2)
+                    f.write("\n")
         except exceptions.UnsatisfiedException as excp:
             self.process_unsatisfied_exceptions(excp)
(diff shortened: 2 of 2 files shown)


ikelos:
Thanks,  I don't have anything against this, but it seems excessive.  If the newline characters were necessary, the official JSON python module would add them when writing, so it feels like this addition is just for humans?  As I say, I don't really have a problem with adding them, but I'd feel better having a strong reason for adding extra complexity (even though it's minimal).  Is there a strong reason for needing a blank line at the end of the file?

digitalisx:
@ikelos Thank you for your comment.
I think it can be different depending on the perspective of the individual.

It is recommended to include `EOF` in codes and files according to the `LINE` defined by _[POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206)_. In the old days, we learned to put it in or supplement it because it could cause other results or errors. (That's how I learned it.) 

Of course, I think modern compilers and many operating systems have evolved to the point where we don't have to consider this. But I think the reason why Github and IDE are still correcting this is to show the same results in various user environments.

In conclusion, I suggested PR as part of an effort to show the same results to existing practices and various users.

I think it's probably because of my personality that I think is excessive.
If you think the code is less valuable than maintain of the code, you can always close it.
This is just a suggestion. 🙌

Based on the results, I will follow that direction in the project. 🙂

ikelos:
Well, you've got some kinda of a justification for it, so I'm ok adding it, but I'll probably try to keep a limit on stylistic points like this being committed otherwise we may get people back-and-forthing the decision just because of personal taste...  5:)

digitalisx:
@ikelos Thank you for this decision, and I will consider whether it is personal taste or improvement for everyone that before contribute.. :)

PR #754, opened 2022-05-29
CLI: Add support for a configuration options file

### LINKED ISSUE #716: Implement configuration file for persistent CLI options
Hi, I would like to setup a symbol directory without using everytime the `-s` option. Is the volatilityrc still usable ? what is the default configuration search path ? Where is the documentation on the configuration file as well please ?

ikelos:
This adds support for creating a `~/.config/volatility3/vol.json` or `volshell.json` file to handle standard command line parameters.  The ordering will go:

Command line parameter else
JSON command line parameter else
`Constants.py` file else
Hard coded default

The last two may be slightly conflated, since the hard coded default might be the value from the `constants.py` file.

I'd considered adding the CLI config to the volatility context config, but I think that might muddy the waters, with the UI variables accessible to plugins.  Plugins should not need to be interrogating the UI, the UI should tell the plugin everything it needs to know upfront.

This doesn't immediately deal with parameters that aren't exposed by the CLI, but it does offer the ability for the JSON to store additional options that could be used by other features (such as renderers).  I still feel these options should be exposed programmatically, so different UIs can offer means of setting them (since not all other UIs will use the config file).  I still think this is a reasonable solution, slots in cleanly to the UI and provides expansion capability with a bit more consideration.  JSON might be a little too descriptive, given the CLI technically only takes specific types, but I think it will help with future expansion, and we've got space for type checking on load.

### DIFF
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -106,7 +106,7 @@ class CommandLine:
         )
 
         # Load up system defaults
-        delayed_logs, default_config = self.load_system_defaults('vol.json')
+        delayed_logs, default_config = self.load_system_defaults("vol.json")
 
         parser = volargparse.HelpfulArgParser(
             add_help=False,
@@ -280,7 +280,7 @@ class CommandLine:
 
         if partial_args.cache_path:
             constants.CACHE_PATH = partial_args.cache_path
-        
+
         vollog.info(f"Volatility plugins path: {volatility3.plugins.__path__}")
         vollog.info(f"Volatility symbols path: {volatility3.symbols.__path__}")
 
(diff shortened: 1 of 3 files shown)


volatility3/cli/volshell/__init__.py
```diff
@@ -108,10 +118,33 @@ def run(self):
...
         partial_args, _ = parser.parse_known_args(known_args)
+
+        ### Start up logging
+        if partial_args.log:
+            file_logger = logging.FileHandler(partial_args.log)
+            file_logger.setLevel(1)
+            file_formatter = logging.Formatter(datefmt = '%y-%m-%d %H:%M:%S',
+                                               fmt = '%(asctime)s %(name)-12s %(levelname)-8s %(message)s')
+            file_logger.setFormatter(file_formatter)
+            rootlog.addHandler(file_logger)
```
digitalisx: The `rootlog` variable does not appear to exist in that source file.
```python
rootlog = logging.getLogger()
```

ikelos: Thanks, I didn't spot I pulled that over, but it should now be resolved in commit 9ea1a96

digitalisx: Thank you for apply! :)

volatility3/cli/__init__.py
```diff
@@ -366,6 +376,30 @@ def location_from_file(cls, filename: str) -> str:
...
+        default_config_path = os.path.join(os.path.expanduser("~"), ".config", "volatility3", filename)
+        if sys.platform == 'win32':
+            default_config_path = os.path.join(os.environ.get("APPDATA", os.path.expanduser("~")), "volatility3",
+                                               filename)
+
+        delayed_logs = []
+
+        # Process it if the files exist
+        if os.path.exists(default_config_path):
+            result = json.load(open(default_config_path, 'rb'))
```
digitalisx: What do you think of this way? :)

```suggestion
            with open(default_config_path, 'rb') as config_json:
                result = json.load(config_json)
```

ikelos: Yeah, I like it.  I think I even prefer it, thanks!  5;)

ikelos:
Also should consider adding environment variables into this hierarchy, probably above the JSON file, but below the command line...

ikelos:
This has been sat around and nobody seems to have commented on it further, so I'm going to get it merged in and we'll see if/how it gets used.  The documentation is buried in a man page that isn't linked to the main help yet, but can be found when searching.  For now, I'm ok with that as the best place to document it.

PR #761, opened 2022-06-05
Feature/volshell unicorn support

ikelos:
Changes to volshell to support unicorn support

### DIFF
--- a/volatility3/cli/volshell/__init__.py
+++ b/volatility3/cli/volshell/__init__.py
@@ -7,11 +7,13 @@ import json
 import logging
 import os
 import sys
+from typing import Any, List, Tuple
 
 import volatility3.plugins
 import volatility3.symbols
 from volatility3 import cli, framework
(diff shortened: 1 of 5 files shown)


paulkermann:
I had a case where running some Windows function in Unicorn with addresses retrieved from Volatility3 caused an error in the emulator because there was an indirect call which caused a code flow to hit the CFG bitmap validation. The validation then failed because the addresses were not canonical.
Also, `percpu` context related data caused errors because of address canonization.
Relates to #702.

ikelos:
Hi @paulkermann, that doesn't sound like it relates to this pull request?  It sounds instead as though you're trying to talk about #702 on a completely unrelated pull request by throwing the word unicorn in.  This is counter productive and sprays your comments across the issue tracker rather than keeping your points focussed on the issue under discussion.  It also sets your cause back because you haven't furthered the discussion you've just illustrated what was already discussed in #702.  Please constrain all further commentary about canonization to #702 unless it pertains directly to the issue at hand.

PR #768, opened 2022-06-14
Fix: typo for code comment, requirements name (`windows.modscan`).

digitalisx:
## Description

I fixed typo of code comments.
And fixed name `VersionRequirements` of `windows.modscan` plugin.
Even if the name is different, it does not affect the result, but we revised it for uniformity.

### DIFF
--- a/volatility3/framework/plugins/windows/modscan.py
+++ b/volatility3/framework/plugins/windows/modscan.py
@@ -25,7 +25,7 @@ class ModScan(interfaces.plugins.PluginInterface):
         return [
             requirements.ModuleRequirement(name = 'kernel', description = 'Windows kernel',
                                                      architectures = ["Intel32", "Intel64"]),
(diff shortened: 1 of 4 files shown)


volatility3/framework/plugins/windows/modscan.py
```diff
@@ -25,7 +25,7 @@ def get_requirements(cls):
         return [
             requirements.ModuleRequirement(name = 'kernel', description = 'Windows kernel',
                                                      architectures = ["Intel32", "Intel64"]),
-            requirements.VersionRequirement(name = 'poolerscanner',
+            requirements.VersionRequirement(name = 'poolscanner',
```
ikelos, line 28: This is fine, it's a textual name and isn't really used other than for storage in the config tree.  Should be no issues with changing it.  5:)

ikelos (approved):
Yep, this all looks good, thanks!  5:D

PR #769, opened 2022-06-14
Translate rekall\vtypes to ISF format

paulkermann:
I could not find the `vtypes_to_json.py` file so I wrote one myself.
This utils can be used to translate old vtypes that were hand-written or generated to the new ISF format.


development/rekall_json_to_isf.py
```diff
@@ -0,0 +1,253 @@
+import argparse
+import logging
+import json
+import time
+
+class rekall_types_translator:
+    VERSION = "0.3"
```
digitalisx, line 8: Based on what criteria is the version of the tool designated?

development/rekall_json_to_isf.py
```diff
@@ -0,0 +1,253 @@
...
+    def translate_subtype_string(cls, subtype, base_types):
+        if type(subtype) == str:
+            # Simple string
+            # in base types or local json vtypes
+            if subtype in base_types:
+                return {"kind": "base", "name": subtype}
+            
+            return {"kind": "struct", "name": subtype}
+
+        raise RuntimeError("what is this?", subtype)
```
digitalisx: I think the message of the error is a little kinder, and I need to explain the reason for the error.

development/rekall_json_to_isf.py
```diff
@@ -0,0 +1,253 @@
...
+                type_dict["count"] = array_info["count"]
+                type_dict["subtype"] = self.translate_subtype((array_info["target"], array_info.get("target_args", None)), base_types)
+        elif type_string == "Enumeration":
+            enum_info = type_stuff[1]
+            self.enum_info[enum_info["enum_name"]] = enum_info["target"]
+            type_dict["kind"] = "enum"
+            type_dict["name"] = enum_info["enum_name"]
+        else:
+            logging.error(f"{type_stuff}")
+            raise RuntimeError("How did we get here?", type_stuff)
```
digitalisx: It's the same here. (Error Message)

development/rekall_json_to_isf.py
```diff
@@ -0,0 +1,253 @@
...
+        isf_json["enums"] = self.translate_rekall_enums(self.rekall_json.get("$ENUMS", {}), isf_json["base_types"])
+        isf_json["metadata"] = self.translate_rekall_metadata(self.rekall_json["$METADATA"])
+
+        return isf_json
+
+def main():
+    parser = argparse.ArgumentParser(description='Rekall types to ISF. Output will be written to STDOUT and filepath.isf.json')
+    parser.add_argument('--filepath', dest='filepath', type=str, help='The rekall type json path.')
+    parser.add_argument('--is-64bit', dest='is_64bit', default=True, action="store_true")
+    parser.add_argument('--is-32bit', dest='is_64bit', action="store_false")
```
digitalisx: I see what you mean. However, I think that allowing two conflicting options to be allowed is a possibility of error. Is it possible to adjust this?

development/vtypes_to_isf.py
```diff
@@ -0,0 +1,255 @@
...
+        isf_json["enums"] = self.get_enums(isf_json["base_types"])
+        isf_json["metadata"] = self.get_metadata()
+
+        return isf_json
+
+def main():
+    parser = argparse.ArgumentParser(description='volatility2 vtypes to ISF. Output will be written to STDOUT and filepath.isf.json')
+    parser.add_argument('--filepath', dest='filepath', type=str, help='The vtypes file path.')
+    parser.add_argument('--is-64bit', dest='is_64bit', default=True, action="store_true")
+    parser.add_argument('--is-32bit', dest='is_64bit', action="store_false")
```
digitalisx: It's the same here. (Options)

development/rekall_json_to_isf.py
```diff
@@ -0,0 +1,253 @@
...
+
+    filepath = args.filepath
+    rekall_translator = rekall_types_translator(filepath, args.is_64bit, args.endianness)
+    isf_json_dict = rekall_translator.translate_rekall_json_to_isf()
+
+    with open(f"{filepath}.isf.json", "w") as f:
+        json.dump(isf_json_dict, f, indent=4)
+
+    isf_json = json.dumps(isf_json_dict, indent=4)
+    print(isf_json)
```
digitalisx, line 267: If we convert the `vtype` and print it out, it may be very long..

development/vtypes_to_isf.py
```diff
@@ -0,0 +1,255 @@
...
+
+    filepath = args.filepath
+    vtypes = vtypes_translator(filepath, args.is_64bit, args.endianness)
+    isf_json_dict = vtypes.translate_to_isf()
+
+    with open(f"{filepath}.isf.json", "w") as f:
+        json.dump(isf_json_dict, f, indent=4)
+
+    isf_json = json.dumps(isf_json_dict, indent=4)
+    print(isf_json)
```
digitalisx, line 238: It's the same here. (Print)

digitalisx (commented):
Thank you for your work @paulkermann,
It is a good pleasure to have tools for developers. 🙂

For now, I think this PR is ready for review because there is no draft or separate marker, so I left some comments.

Additionally, in the case of new tools, I think it will take some time to figure out the intentions and the consequences unless you're the one who wrote the code yourself.

It might be hard to say this in the absence of a contribute guide.. 😅
If you don't mind, could you write examples or methods of running in the PR description?

In `volatility`, `vtypes` are in the form of JSON inside python, or the code that merges into the existing `vtype` depending on the operating system, so I'm not sure how to put it in.

development/rekall_json_to_isf.py
paulkermann: This version is just local iterations on my local pc. It does not hold any real meaning

development/rekall_json_to_isf.py
paulkermann: I will rename the error string

development/rekall_json_to_isf.py
paulkermann: Also changed

development/rekall_json_to_isf.py
paulkermann: That's what Stackoverflow said stuff about defaulting boolean arguments with argparse. 
It is compatible with many python versions so I chose this one.

development/rekall_json_to_isf.py
paulkermann: This way you can pipe it to wherever you want to save it. The pdbparse library which used to create the old vtypes worked in the same way.

development/vtypes_to_isf.py
paulkermann: Same as above

development/rekall_json_to_isf.py
```diff
@@ -0,0 +1,253 @@
...
+        metadata["producer"] = {}
+        metadata["producer"]["name"] = "rekall types to ISF"
+        metadata["producer"]["datetime"] = time.ctime()
+        metadata["producer"]["version"] = cls.VERSION
+        metadata["rekall_metadata"] = rekall_metadata
+
+        return metadata
+
+    def translate_rekall_enums(self, rekall_enums, base_types):
+        # assume self.enum_info is initilized; run after `translate_rekall_structs`
```
digitalisx:
```suggestion
        # assume self.enum_info is initialized; run after `translate_rekall_structs`
```

digitalisx (commented):
Additional, found typo for code comment.

development/rekall_json_to_isf.py
paulkermann: In case there are issues with the output it could be traced to a specific version now

development/rekall_json_to_isf.py
paulkermann: Fixed it

paulkermann:
I have done little testing on it. 
[win10_x64_19041_vtypes.py.txt](https://github.com/volatilityfoundation/volatility3/files/8901106/win10_x64_19041_vtypes.py.txt)
[0018A9A7F0334E8D965F310D1653A5452.json.txt](https://github.com/volatilityfoundation/volatility3/files/8901107/0018A9A7F0334E8D965F310D1653A5452.json.txt)

Because it is a developer tool I feel comfortable not doing many tests.
Also, as vtypes are essentially python code, some translations may not work.

digitalisx:
> I have done little testing on it. [win10_x64_19041_vtypes.py.txt](https://github.com/volatilityfoundation/volatility3/files/8901106/win10_x64_19041_vtypes.py.txt) [0018A9A7F0334E8D965F310D1653A5452.json.txt](https://github.com/volatilityfoundation/volatility3/files/8901107/0018A9A7F0334E8D965F310D1653A5452.json.txt)
> 
> Because it is a developer tool I feel comfortable not doing many tests. Also, as vtypes are essentially python code, some translations may not work.

Thank you for providing the file, I added some explanation for others. 🙌

## Command

```shell
> python3 vtypes_to_isf.py --filepath=win10_x64_19041_vtypes.py
```

### Input `vtype` example
```python
ntkrnlmp_types = {
  'LIST_ENTRY64' : [ 0x10, {
    'Flink' : [ 0x0, ['unsigned long long']],
    'Blink' : [ 0x8, ['unsigned long long']],
} ],
  'LIST_ENTRY32' : [ 0x8, {
    'Flink' : [ 0x0, ['unsigned long']],
    'Blink' : [ 0x4, ['unsigned long']],
} ],
  '_LIST_ENTRY' : [ 0x10, {
    'Flink' : [ 0x0, ['pointer64', ['_LIST_ENTRY']]],
    'Blink' : [ 0x8, ['pointer64', ['_LIST_ENTRY']]],
} ],
  '_SINGLE_LIST_ENTRY' : [ 0x8, {
    'Next' : [ 0x0, ['pointer64', ['_SINGLE_LIST_ENTRY']]],
} ],
  '_UNICODE_STRING' : [ 0x10, {
    'Length' : [ 0x0, ['unsigned short']],
...
```

### Output example
```json
"Reserved": {
    "offset": 0,
    "type": {
        "kind": "bitfield",
        "bit_position": 12,
        "bit_length": 4,
        "type": {
            "kind": "base",
            "name": "unsigned long long"
        }
    }
},
"ExtendedRegisterCount": {
    "offset": 0,
    "type": {
        "kind": "bitfield",
        "bit_position": 16,
        "bit_length": 8,
        "type": {
            "kind": "base",
            "name": "unsigned long long"
        }
    }
},
```

development/rekall_json_to_isf.py
digitalisx: I think I found what you [said](https://stackoverflow.com/questions/15008758/parsing-boolean-values-with-argparse/15008806#15008806).
All right, I think it might be different because it's a matter of pattern and style. I agree with your code.

paulkermann:
@digitalisx I think you put the wrong output. The output should be an ISF file and ISF files do not contains the '$CONSTANTS' key.

digitalisx:
> @digitalisx I think you put the wrong output. The output should be an ISF file and ISF files do not contains the '$CONSTANTS' key.

@paulkermann Ah good point, I pasted the wrong data. I fixed it.

digitalisx:
I think we need more evaluate (of maintainers) and testing various perspectives on how this tool will help developers and how well to convert the various types of `volatility` or `rekall` to `volatility3`. I wish your good luck.

paulkermann:
Hey @ikelos! Is there any update on this issue?

ikelos:
Not yet I'm afraid, it really isn't a priority.  The tasks stopping this are: 
* Deciding on how to support both mangled and unmangled names when looking up symbols, while preventing name confusion attacks
* Implementing any necessary name mangling algorithms for volatility

We could just put all the names into one lookup method, but that then leaves opens the chance for intentional confusion attacks and more importantly accidental shadowing (where a mangled and unmangled name conflict and one takes precedence.  It also means that to find a function, people need to know the types for parameters and return types.

We need to decide on how much support to add to volatility (full C++ object/method/attribute support or just object/method names).

Given volatility wasn't really designed with any of these concepts in mind, it's going to be a big undertaking to try to merge them nicely into the existing ecosystem (as it was to do with C and C++, hence even the existence of name mangling).

If you can get enough people together to design it and agree how it should work out might move forward, but the team is spread pretty thin and what little time I can give these days is spent more on maintenance, unfortunately.  I hope that clarifies what the hold up is and why it's not made more progress...

PR #772, opened 2022-06-16
Better page faulting with swap file support

paulkermann:
Handle swap entries more correctly.
I have removed the ` # Handle transition page` thing because transition is handle via `_page_is_valid` which is overrided by the `WindowsMixin` anyways.
@f-block - I hope you can test this PR to see if it works for you. I have tested it using the dump you have provided me and it seems to work. I hope you have a better setup for this.


digitalisx:
Hello @paulkermann, This PR related to #581 ?

f-block:
Hi,

yes, this PR looks/works great! Thx @paulkermann.

paulkermann:
@digitalisx yes this PR is into #581

f-block:
Hi,

just had another look: Why do you make the `PageFileLow` shift in `_get_PageFileLow_shift` dependent on the existence of the `SwizzleBit` ? If in a newer Windows version the `PageFileLow` field get moved around again, this won't work anymore. I think the best way to do it would be to get the field start from the `PageFileLow` field directly. Something like this (didn't test it yet):

```
return mmpte_software_type.vol.members.get("PageFileLow")[1].vol.start_bit
```

Cheers,
Frank

paulkermann:
@f-block thanks'. Did not know the `start_bit` was accessible this way.

volatility3/framework/layers/intel.py
```diff
@@ -354,10 +354,8 @@ def _get_invalid_pte_mask(self, kernel):
     @functools.lru_cache()
     def _get_PageFileLow_shift(self, kernel):
         mmpte_software_type = kernel.get_type("_MMPTE_SOFTWARE")
-        if mmpte_software_type.vol.members.get("SwizzleBit", None) is None:
-            return 1 # The old shift
 
-        return 12 # The new shift
+        return mmpte_software_type.vol.members["PageFileLow"][1].start_bit
```
ikelos, line 366: This is a pretty nasty way of doing it, and tinkers with the internal state of volatility.  I've now added #777 which gives the `child_template` call, so you can do `mmpte_software_type.child_template('PageFileLow').start_bit` instead.  Should be much cleaner and make it easier for us if we ever need to mess with the members mechanism.

volatility3/framework/layers/intel.py
```diff
@@ -338,33 +371,19 @@ def _page_is_valid(entry: int) -> bool:
         return bool((entry & 1) or ((entry & 1 << 11) and not entry & 1 << 10))
 
     def _handle_page_fault(self, layer_name, offset, invalid_bits, entry, description):
+        kernel = self._get_kernel_module()
```
ikelos: This should be memoized, otherwise there's going to be (potentially) a bunch of page fails that retraverse the module list to find the kernel.  I'd consider a `_kernel` property which calculates it first time, and returns it from a variable otherwise.

volatility3/framework/layers/intel.py
```diff
@@ -338,33 +371,19 @@ def _page_is_valid(entry: int) -> bool:
...
-            table = self._get_valid_table(base_address)
-            if table is None:
-                raise exceptions.PagedInvalidAddressException(self.name, offset, position + 1, entry,
-                                                              "Page Fault at entry " + hex(
-                                                                  entry) + " in table " + name)
-
-            # Read the data for the next entry
-            entry_data = table[(index << self._index_shift):(index << self._index_shift) + self._entry_size]
-            super()._translate_entry(offset, invalid_bits, entry)
+        entry &= ~self._get_invalid_pte_mask(kernel)
```
ikelos: I get a little bit touchy about `~` in python when dealing with values that might be different lengths.  I think xor would be equivalent to `& ~` and would be unambiguous for any length of value.

volatility3/framework/layers/intel.py
```diff
@@ -324,6 +324,39 @@ class Intel32e(Intel):
...
+        if kvo is None:
+            return None
+
+        for module_name in self.context.modules:
+            if self.context.modules[module_name].offset == kvo:
+                return self.context.modules[module_name]
+
+        return None
+
+    @functools.lru_cache()
```
ikelos, line 346: The `lru_cache` call should work ok, but I recall some difficulties that were encountered in certain places which is why we opt for manual storage in an instance variable in some places.  If it doesn't cause any problems then that's ok, but any weird behaviour, this would be my first culprit.

ikelos (commented):
I'm still concerned about bringing a requirement on the kernel into the windows mixin.  This will need some thinking about and a lot of testing.

volatility3/framework/layers/intel.py
paulkermann: From my understand the function result is cached per class instance so I also think it should work fine.

volatility3/framework/layers/intel.py
paulkermann: When are you planning on merging this in?

volatility3/framework/layers/intel.py
paulkermann: Replaced it with a xor

volatility3/framework/layers/intel.py
paulkermann: moved `self._kernel` initialization to `Intel` constructor.

ikelos:
The PR's become a bit of a message, please could you rebase it.

volatility3/framework/layers/intel.py
ikelos: After that standard review process, so I'm hoping a week but it could be longer.

volatility3/framework/layers/intel.py
ikelos: The kernel variable wouldn't be used by the `Intel` class, please make it part of the `WindowsMixin`.

paulkermann:
@ikelos the problem is [better-page-faulting](https://github.com/volatilityfoundation/volatility3/tree/feature/better-page-faulting) is behind develop. My branch ([this](https://github.com/paulkermann/volatility3/tree/feature/swap-better-pagefault)) is ahead of develop and not behind

ikelos:
Hmmm, ok.  Either you can roll in the changes from better-page-faulting, rebase it all, and we abandon that PR/branch in favour of this one, or we'll have to wait for it to get merged at some point...

paulkermann:
moved this PR to #778.

PR #775, opened 2022-06-22
Refactor: simplify comparision syntax.

digitalisx:
## Description

I did some refactoring on comparison syntax.
I have modified the comparison that can be improved to make it more intuitive and easier for users to understand.
And `None` type of comparison use [recommend](https://peps.python.org/pep-0008/#programming-recommendations) to `is` or `is not`.


ikelos:
These all look good, thanks!  The `is None` is quite an important one, so thanks for catching those!

PR #774, opened 2022-06-22
Documentation: latest update for `simple-plugin`.

digitalisx:
## Description
Hello, everyone in the community! 🙂

Volatility3 has written wonderful documents to help the community participate.
However, during the update process, the document was not updated, and there are some issues that are problematic to apply based on the last release. In particular, the version of the plugin or framework is a problem in the example of creating a plugin. Therefore, I revised the document a little so that it can develop plugin enough in `2.0.1`.

If you have any better content or suggestions that you would like to add to the document, please leave a thread.

Fixed #773

Issue #767, opened 2022-06-12
Automatic synchronization of `readthedocs` to the latest version.


digitalisx:
Hello, everyone in the community! 🙂
While I was teaching students about `volatility3`, I told them that it would be good to refer to the document,
but it seems that the [documents](https://volatility3.readthedocs.io/en/latest/) is not synchronized with the latest code.
If possible, I think `readthedocs` will be a little more useful if we reflect the newly added or removed contents in connection with Github Actions.
It's very easy to write Github Actions, which deploy `readthedocs` that are being push to repository.
But credentials and permissions are required to link with `readthedocs`.
So I can't contribute this area.
Perhaps, Could you have any plans or intentions for this? :)

ikelos:
Hiya, By default readthedocs takes you to the current stable version (`stable` branch), but you can get the latest development documentation, for the `develop` branch. by going to the `latest` version or visiting https://volatility3.readthedocs.io/en/latest/.  We've chosen this due to the unstable nature of the `develop` branch and wanting to avoid confusion from people who just download the main package and don't make use of git.  If you find that latest is out of date with develop, please let us know.  Otherwise I believe it's functioning as intended...

digitalisx:
Yes @ikelos, your intentions seem very right.
So I think that the `stable` version is up to date as you said!
However, the `develop` version does not seem to be reflected immediately as the content of the `develop` branch..!
Some of the submodules (like linux `psaux`, windows `devicetree`, `mbrscanner`..) that have been reflected since a certain point in time seem to be missing or have not been updated.
I think most document builds by hand. And last [build](https://readthedocs.org/projects/volatility3/builds/) of `develop` is about 2 months ago.
Could you check this please? 🙂

ikelos:
Thanks, that's all looked after by @npetroni, so best to check in with him.  My understanding was that it should be done automatically, but I'm not certain...  5:)

npetroni:
I rebuilt `latest` and added a webhook. Let's see what happens next push to `develop`.

digitalisx:
@npetroni Thank you for your work! I'll check result when push at `develop` branch. :)

digitalisx:
All right, Recently I was lucky enough to get the chance (#768) will change the document to push at `develop` branch. And when I checked the changes, it was successful. If there is no problem with the participants of this issue, we will close the issue in the near future.

ikelos:
No problem with me.  5:)

npetroni:
[...]

PR #776, opened 2022-06-22
Issues/issue773

### LINKED ISSUE #773: Simple Plugin Example Problems
I'm following the documentation found here to make a simple plugin.

https://volatility3.readthedocs.io/en/latest/simple-plugin.html

Before I write my own, I'm just going through the steps in the documentation to make the example. However, I have run into two problems. First, the imports shown in the example are incomplete. The resulting plugin will not run without the following:

```python
from volatility3.plugins.windows import pslist
```

Once those are added, the plugin runs, but then halts with the following error:

```
# vol -f test.raw windows.testplugin
Volatility 3 Framework 2.0.1
Unsatisfied requirement plugins.DllList.pslist: 
Unable to validate the plugin requirements: ['plugins.DllList.pslist']
```

The code that I'm using to test with is cut and pasted together from each step in the example documentation as follows. There are no deviations from the documentation except for the two additional imports identified above.

```python
from volatility3.framework import interfaces
from volatility3.plugins.windows import pslist

class DllList(interfaces.plugins.PluginInterface):
    @classmethod
    def get_requirements(cls):
        return [requirements.TranslationLayerRequirement(name = 'primary',
                                                         description = 'Memory layer for the kernel',
                requirements.SymbolTableRequirement(name = "nt_symbols",
                                                    description = "Windows kernel symbols"),
                requirements.PluginRequirement(name = 'pslist',
                                               plugin = pslist.PsList,
                                               version = (1, 0, 0)),
                requirements.ListRequirement(name = 'pid',
                                             element_type = int,
                                             description = "Process IDs to include (all other processes are excluded)",
                                             optional = True)]

    def run(self):
        filter_func = pslist.PsList.create_pid_filter(self.config.get('pid', None))

        return renderers.TreeGrid([("PID", int),
                                   ("Process", str),
                                   ("Base", format_hints.Hex),
                                   ("Size", format_hints.Hex),
                                   ("Name", str),
                                   ("Path", str)],
                                  self._generator(pslist.PsList.list_processes(self.context,
                                                                               self.config['primary'],
                                                                               self.config['nt_symbols'],
                                                                               filter_func = filter_func)))

    def _generator(self, procs):
        for proc in procs:
            for entry in proc.load_order_modules():
                BaseDllName = FullDllName = renderers.UnreadableValue()
                try:
                    BaseDllName = entry.BaseDllName.get_string()
                    # We assume that if the BaseDllName points to an invalid buffer, so will FullDllName
                    FullDllName = entry.FullDllName.get_string()
                    pass

                yield (0, (proc.UniqueProcessId,
                           proc.ImageFileName.cast("string", max_length = proc.ImageFileName.vol.count,
                                                   errors = 'replace'),
                           format_hints.Hex(entry.DllBase), format_hints.Hex(entry.SizeOfImage),
                           BaseDllName, FullDllName))
```

What is missing from the example to get it to run?

ikelos:
This updates the simple plugin documentation to use the latest version of the framework, and features available with that (such as the `ModuleRequirement`).

Resolves #773 
Obsoletes #774

### DIFF
--- a/doc/source/simple-plugin.rst
+++ b/doc/source/simple-plugin.rst
@@ -30,6 +30,9 @@ to be able to run properly.  Any that are defined as optional need not necessari
 
 ::
 
+        _version = (1, 0, 0)
+        _required_framework_version = (2, 0, 0)
+
         @classmethod
         def get_requirements(cls):
             return [requirements.TranslationLayerRequirement(name = 'primary',
@@ -37,13 +40,13 @@ to be able to run properly.  Any that are defined as optional need not necessari
                                                              architectures = ["Intel32", "Intel64"]),
                     requirements.SymbolTableRequirement(name = "nt_symbols",
                                                         description = "Windows kernel symbols"),
-                    requirements.PluginRequirement(name = 'pslist',
-                                                   plugin = pslist.PsList,
-                                                   version = (1, 0, 0)),
                     requirements.ListRequirement(name = 'pid',
                                                  element_type = int,
                                                  description = "Process IDs to include (all other processes are excluded)",
-                                                 optional = True)]
+                                                 optional = True),
+                    requirements.PluginRequirement(name = 'pslist',
+                                                   plugin = pslist.PsList,
+                                                   version = (1, 0, 0))]
 
 
 This is a classmethod, because it is called before the specific plugin object has been instantiated (in order to know how
@@ -93,26 +96,41 @@ This requirement is also a Complex Requirement and therefore will not be request
 
 ::
 
-    requirements.PluginRequirement(name = 'pslist',
-                                   plugin = pslist.PsList,
-                                   version = (1, 0, 0)),
-
-This requirement indicates that the plugin will make use of another plugin's code, and specifies the version requirements
-on that plugin.  The version is specified in terms of Semantic Versioning, meaning that to be compatible, the major
-versions must be identical and the minor version must be equal to or higher than the one provided.  This requirement
-does not make use of any data from the configuration, even if it were provided, it is merely a functional check before
-running the plugin.
-
-::
-
     requirements.ListRequirement(name = 'pid',
                                  description = 'Filter on specific process IDs',
                                  element_type = int,
-                                 optional = True)
+                                 optional = True),
 
-The final requirement is a List Requirement, populated by integers.  The description will be presented to the user to
+The next requirement is a List Requirement, populated by integers.  The description will be presented to the user to
 describe what the value represents.  The optional flag indicates that the plugin can function without the ``pid`` value
 being defined within the configuration tree at all.
+
+::
+
+    requirements.PluginRequirement(name = 'pslist',
+                                   plugin = pslist.PsList,
+                                   version = (1, 0, 0))
+
+This requirement indicates that the plugin will make use of another plugin's code, and specifies the version requirements
+on that plugin.  The version is specified in terms of Semantic Versioning meaning that, to be compatible, the major
+versions must be identical and the minor version must be equal to or higher than the one provided.  This requirement
+does not make use of any data from the configuration, even if it were provided, it is merely a functional check before
+running the plugin.  To define the version of a plugin, populate the `_version` class variable as a tuple of version
+numbers `(major, minor, patch)`.  So for example:
+
+::
+
+    _version = (1, 0, 0)
+
+The plugin may also require a specific version of the framework, and this also uses Semantic Versioning, and can be
+set by defining the `_required_framework_version`.  The major version should match the version of volatility the plugin
+is to be used with, which at the time of writing would be 2.2.0, and so would be specified as below.  If only features, for example,
+from 2.0.0 are used, then the lowest applicable version number should be used to support the greatest number of
+installations:
+
+::
+
+    _required_framework_version = (2, 0, 0)
 
 Define the `run` method
 -----------------------
--- a/doc/source/simple-plugin.rst
+++ b/doc/source/simple-plugin.rst
@@ -43,7 +43,7 @@ to be able to run properly.  Any that are defined as optional need not necessari
                                                  optional = True),
                     requirements.PluginRequirement(name = 'pslist',
                                                    plugin = pslist.PsList,
-                                                   version = (1, 0, 0))]
+                                                   version = (2, 0, 0))]
 
 
 This is a classmethod, because it is called before the specific plugin object has been instantiated (in order to know how
--- a/doc/source/simple-plugin.rst
+++ b/doc/source/simple-plugin.rst
@@ -134,7 +134,7 @@ being defined within the configuration tree at all.
 
     requirements.PluginRequirement(name = 'pslist',
                                    plugin = pslist.PsList,
-                                   version = (1, 0, 0))
+                                   version = (2, 0, 0))]
 
 This requirement indicates that the plugin will make use of another plugin's code, and specifies the version requirements
 on that plugin.  The version is specified in terms of Semantic Versioning meaning that, to be compatible, the major
--- a/doc/source/simple-plugin.rst
+++ b/doc/source/simple-plugin.rst
@@ -35,11 +35,8 @@ to be able to run properly.  Any that are defined as optional need not necessari
 
         @classmethod
         def get_requirements(cls):
-            return [requirements.TranslationLayerRequirement(name = 'primary',
-                                                             description = 'Memory layer for the kernel',
-                                                             architectures = ["Intel32", "Intel64"]),
-                    requirements.SymbolTableRequirement(name = "nt_symbols",
-                                                        description = "Windows kernel symbols"),
+            return [requirements.ModuleRequirement(name = 'kernel', description = 'Windows kernel',
+                                                   architectures = ["Intel32", "Intel64"]),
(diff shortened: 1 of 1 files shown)


ikelos:
@digitalisx thanks for #774 , hopefully this is a slightly more filled out fix for the documentation issue.  Hopefully this all makes sense, but I'd very much appreciate your eagle eye to see if I've introduced any mistakes or gotten anything wrong...  5;)

doc/source/simple-plugin.rst
```diff
@@ -30,89 +30,132 @@ to be able to run properly.  Any that are defined as optional need not necessari
...
+            return [requirements.ModuleRequirement(name = 'kernel', description = 'Windows kernel',
+                                                   architectures = ["Intel32", "Intel64"]),
                     requirements.ListRequirement(name = 'pid',
                                                  element_type = int,
                                                  description = "Process IDs to include (all other processes are excluded)",
-                                                 optional = True)]
+                                                 optional = True),
+                    requirements.PluginRequirement(name = 'pslist',
+                                                   plugin = pslist.PsList,
+                                                   version = (1, 0, 0))]
```
digitalisx: I think the minimum version of the `pslist` plugin should be `2.0.0`. Setting this to `1.0.0` will result in the error mentioned in the issue (#773).

doc/source/simple-plugin.rst
```diff
@@ -30,89 +30,132 @@ to be able to run properly.  Any that are defined as optional need not necessari
...
+The next requirement is a List Requirement, populated by integers.  The description will be presented to the user to
+describe what the value represents.  The optional flag indicates that the plugin can function without the ``pid`` value
+being defined within the configuration tree at all.
 
 ::
 
     requirements.PluginRequirement(name = 'pslist',
                                    plugin = pslist.PsList,
-                                   version = (1, 0, 0)),
+                                   version = (1, 0, 0))
```
digitalisx: Same point of above review comment.

digitalisx (commented):
@ikelos No problem!
I'm so happy if both the community and the user can go in a better direction.
Thank you for working on the issue on your busy schedule!

I left some comments through the review, could you take a look? :)

doc/source/simple-plugin.rst
digitalisx: If unintended, this would be a new issue.

digitalisx:
Duplicate of #774

doc/source/simple-plugin.rst
digitalisx:
```suggestion
                                   version = (2, 0, 0))
```

doc/source/simple-plugin.rst
digitalisx:
```suggestion
                                                   version = (2, 0, 0))]
```

doc/source/simple-plugin.rst
ikelos: Thanks, I fixed this manually.  The comma was intentional (to show it was copied from above), but actually it should have been a `]`.

digitalisx (approved):
@ikelos Good! Thank you for reflecting my opinion. I hope this document will help more people. :)

ikelos:
Thanks for checking it through, it's much appreciated!  5:D

utkonos:
@ikelos @digitalisx Thank you both!

utkonos:
Followup question: I see that this documentation change was merged into develop. Was this intentional? I am unfamiliar with the documentation release process here, but the published documentation is still the old `simple-plugin.rst`. If this is the correct release process, then disregard and forgive my ignorance.

utkonos:
Also, I'm sorry to be such a pest, but this plugin example still doesn't work. There are three imports that I've identified that need to be added to the example.

```python
from volatility3.framework import renderers, interfaces
from volatility3.plugins.windows import pslist
```

However, I'm still hitting an exception. I'm looking now to find this import in other plugins.

```python
Volatility 3 Framework 2.0.1
  File "/home/vagrant/venv/bin/vol", line 8, in <module>
    sys.exit(main())
  File "/home/vagrant/venv/lib/python3.10/site-packages/volatility3/cli/__init__.py", line 625, in main
    CommandLine().run()
  File "/home/vagrant/venv/lib/python3.10/site-packages/volatility3/cli/__init__.py", line 333, in run
  File "/home/vagrant/venv/lib/python3.10/site-packages/volatility3/plugins/windows/test.py", line 28, in run
    ("Base", format_hints.Hex),
NameError: name 'format_hints' is not defined
```

When someone has read the documentation for the simple plugin, they should not need to find imports to get the example to work. Everything should work when the page has been read and followed.

digitalisx:
Hello @utkonos, Thank you for your continuous tracking of the issue.

Some of the processes and intentions of the document release can be found here. (#767)

And `format_hints` can import this code,


```python
from volatility3.framework.renderers import format_hints
```

That's right, I need to mention that part, but I'm worried about how to appear according to the order of description in the document. If there is no work of the maintainer, I will submit the PR in the same way as last time.

ikelos:
Hi @utkonos, the documentation should match the version that it's released with, and develop tracks with [latest](https://volatility3.readthedocs.io/en/latest/simple-plugin.html).  By default, we go to the latest [stable release](https://volatility3.readthedocs.io/en/stable/simple-plugin.html) whose code doesn't contain the changes and would therefore confuse people if the documentation changed.

The example code wasn't designed to be copied verbatim, and in fact should closely match the existing [`dlllist` plugin](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/dlllist.py) if a copy of the code is needed to work along with.  For complete and functioning plugins, the plugins directory https://github.com/volatilityfoundation/volatility3/tree/develop/volatility3/framework/plugins contains all the examples a user could want.  This example is special because it talks through what each part of the code does.  It was supposed to demonstrate and talk through the key components necessary for writing a plugin, and assumes knowledge of python to be able to fill in parts like the imports, since basic python knowledge is seen as a prerequisite for writing plugins.

Having said all that, I'm happy for the imports to be added as long as there aren't too many that then detract/distract from the key learning points of the example, but the documentation is a teaching resource first and foremost, and copy/pastable code second.

utkonos:
@ikelos That release process is understandable. However, the original issue that I submitted was due to the documentation _not_ being updated to match the released version. The issue that I opened was focused on the published documentation not working for the version that was released (imports totally aside).

utkonos:
One more point: this simple plugin documentation is the very first document that discusses writing a plugin in the documentation. If you want people to know that all the code needed to understand plugins is in that directory, this fact should be somewhere in this documentation. Maybe at the top mention that:

```
The example code wasn't designed to be copied verbatim, and in fact should closely match the existing [dlllist plugin](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/dlllist.py) if a copy of the code is needed to work along with. For complete and functioning plugins, the plugins directory https://github.com/volatilityfoundation/volatility3/tree/develop/volatility3/framework/plugins contains all the examples a user could want.
```

That sentence and the link to the location for a new plugin writer to read the code of existing plugins will fill the gap well enough if the example is not meant to result in a working plugin verbatim.

ikelos:
I understand, unfortunately we can't easily backport fixes for documentation, since the release contains the documentation itself, so to change the documentation we'd need to issue a new release.  I don't think the change warrants a emergency release, but the good news is that we are aiming to roll a new release in September which will therefore come with the fixed up documentation.

Point taken about the documentation, I'll get that updated to reflect that the code isn't complete.

utkonos:
@ikelos Thanks! And sorry if I'm being a pest!

ikelos:
Not at all, thanks for helping us make the project better!  5:D  Hopefully it's clearer from commit `951a0f5d` where to go for samples of the code.  5:)

utkonos:
I've run through the example and my plugin now works nicely. Thanks for all of this. I'm now going to start making some new plugins.

On a side note, I rewrote the plugin a bit more in my style of code. None of which is important for this example. There is one small modification that I did that makes the code more Pythonic. There is a `try` / `except` that can be smoother using `contextlib`. Here is the former code:

```python
try:
    BaseDllName = entry.BaseDllName.get_string()
    # We assume that if the BaseDllName points to an invalid buffer, so will FullDllName
    FullDllName = entry.FullDllName.get_string()
    pass
```

This is how I rewrote it:

```python
with contextlib.suppress(exceptions.InvalidAddressException):
    BaseDllName = entry.BaseDllName.get_string()
    FullDllName = entry.FullDllName.get_string()
```

The comment can be left in as well, if that's important.

```python
# We assume that if the BaseDllName points to an invalid buffer, so will FullDllName
with contextlib.suppress(exceptions.InvalidAddressException):
    BaseDllName = entry.BaseDllName.get_string()
    FullDllName = entry.FullDllName.get_string()
```

ikelos:
Nice, I'll have to try and remember that one.  I'm a little cautious about making it easier for people to skip past exceptions, it's been somewhat of an uphill battle to get people to properly catch and deal with exceptions, but I guess searching for `contextlib.suppress` is as easy as searching for `pass`.  I'll still try and keep a limit on their use because generally when something goes wrong we should be logging it so the user knows that something when wrong and what but as here, there are certain situations where the exception's already been handled.  I'll have a think about going through the codebase sometime and adding it to our style, thanks!  5:)

PR #777, opened 2022-06-23
Core: Add support to templates to get child templates

ikelos:
This adds a new call to the API to stop people rummaging around inside the internal state of volatility.

### DIFF
--- a/volatility3/framework/objects/__init__.py
+++ b/volatility3/framework/objects/__init__.py
@@ -605,7 +605,7 @@ class Array(interfaces.objects.ObjectInterface, collections.abc.Sequence):
         @classmethod
         def child_template(cls, template: interfaces.objects.Template, child: str) -> interfaces.objects.Template:
             """Returns the template of the child member."""
-            if 'subtype' in template.vol and child == 'subtype'@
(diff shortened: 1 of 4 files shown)


volatility3/framework/objects/__init__.py
```diff
@@ -602,6 +602,14 @@ def relative_child_offset(cls, template: interfaces.objects.Template, child: str
                 return 0
             raise IndexError(f"Member not present in array template: {child}")
 
+        @classmethod
+        def child_template(cls, template: interfaces.objects.Template, child: str) -> interfaces.objects.Template:
+            """Returns the template of the child member."""
+            if 'subtype' in template.vol and child == 'subtype'@
```
digitalisx: It seems typo of `if` statement syntax.

```suggestion
            if 'subtype' in template.vol and child == 'subtype':
```

digitalisx (commented):
Thank you for work @ikelos, All looks very good!
I left very small comment about code. :)

volatility3/framework/objects/__init__.py
ikelos: You're quite right, thanks for the spot!  5:)

PR #782, opened 2022-07-03
Core: Convert try/except/pass to contextlib.supress

ikelos:
This is based on a recommendation from @utkonos in #776 to consider using `contextlib.suppress` for exceptions that simply pass.  Note there's a few cases where this hasn't been applied:
* Where multiple exceptions are caught and not all pass.
* Where the UI removes imports (such as smb) when attempting to suppress import errors.


ikelos:
This has been sat here for a while, so before it strays too far from its origin, I'm going to merge it.  It's a very uncontentious patch, but we'll keep an eye out for issues that might be raised from it...

digitalisx:
I think we can delete the merged branch! 😀

PR #783, opened 2022-07-04
Refactor try to reduce size of clause to only what is needed.

utkonos:
Based on feedback, memory_object.get_available_pages() might raise
this type of exception, so it's still inside the try clause.

### DIFF
--- a/volatility3/framework/plugins/windows/dumpfiles.py
+++ b/volatility3/framework/plugins/windows/dumpfiles.py
@@ -63,29 +63,28 @@ class DumpFiles(interfaces.plugins.PluginInterface):
         :return: result status
         """
         filedata = open_method(desired_file_name)
-        try:
-            # Description of these variables:
-            #   memoffset: offset in the specified layer where the page begins
-            #   fileoffset: write to this offset in the destination file
-            #   datasize: size of the page
+        # Description of these variables:
+        #   memoffset: offset in the specified layer where the page begins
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/dumpfiles.py
```diff
@@ -63,29 +63,28 @@ def dump_file_producer(cls, file_object: interfaces.objects.ObjectInterface,
...
-                return filedata
         except exceptions.InvalidAddressException:
             vollog.debug(f"Unable to dump file at {file_object.vol.offset:#x}")
             return None
+        if not bytes_written:
+            vollog.debug(f"No data is cached for the file at {file_object.vol.offset:#x}")
+            return None
+        vollog.debug(f"Stored {filedata.preferred_filename}")
+
+        return filedata
```
digitalisx: That's a pretty NIT comment, and I think the log and its corresponding return syntax are a little bit better placed nearby.

```suggestion

        vollog.debug(f"Stored {filedata.preferred_filename}")
        return filedata
```

digitalisx (commented):
Thank you for your work @utkonos, I left some NIT comment. 🙂

volatility3/framework/plugins/windows/dumpfiles.py
utkonos: Thanks.

ikelos:
This looks good, thanks very much!  5:)

PR #784, opened 2022-07-04
Style changes including yapf according to .style.yapf in package root

utkonos:


ikelos (approved):
Looks good, we don't tend to initialize using `list()`, but instead use type annotations to describe the list (and if possible its contents) before using `= []`, but I'm happy to accept it.

PR #789, opened 2022-07-15
Add testing framework

fgomulka:


.github/workflows/test.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python 3.x
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
```
digitalisx: Volatility 3 recommends `3.6` or higher, so shouldn't we adjust that version to `3.6` as well?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,54 @@
+name: Test Volatility3
+on: [push]
```
digitalisx: Push is good for the test, but it would be great if it is possible to pre-test during the Pull Request stage. Most contributors will test it by turning it around themselves, but if we can test it in a common environment, we will be able to reduce errors. Of course, this may vary depending on the available Github Actions in Organization.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,54 @@
...
+      run: |
+        python setup.py sdist --formats=gztar,zip
+        python setup.py bdist_wheel
+
+    - name: Download images
+      run: |
+        curl -sLO "https://downloads.volatilityfoundation.org/volatility3/images/linux-sample-1.bin.gz"
+        gunzip linux-sample-1.bin.gz
+        curl -sLO "https://downloads.volatilityfoundation.org/volatility3/images/win-xp-laptop-2005-06-25.img.gz"
+        gunzip win-xp-laptop-2005-06-25.img.gz
```
digitalisx, line 33: I think we need to test the relatively up-to-date Windows and Mac images. Is this possible?

test/test_volatility.py
```diff
@@ -0,0 +1,381 @@
...
+
+            path = tempfile.mkdtemp()
+
+            rc, out, err = runvol_plugin("windows.dumpfiles.DumpFiles", image, volatility, python, globalargs=["-o", path], pluginargs=["--virtaddr", addr])
+
+            for file in os.listdir(path):
+                fp = open(os.path.join(path, file), "rb")
+                if hashlib.md5(fp.read()).hexdigest() not in known_files["windows_dumpfiles"][file_name][addr]:
+                    failed_chksms += 1
+                fp.close()
```
digitalisx: `with` syntax does not require a separate close.
```suggestion
                with open(os.path.join(path, file), "rb") as fp:
                    if hashlib.md5(fp.read()).hexdigest() not in known_files["windows_dumpfiles"][file_name][addr]:
                        failed_chksms += 1
```

digitalisx (commented):
Thank you for your work @fgomulka!
I think this is very necessary for the project. I left a few comments to make it better.

.github/workflows/test.yaml
fgomulka: @npetroni , shall I also create a workflow for testing on PRs? Or do you think that we should just perform *this* test on PRs as well as pushes?

npetroni: I think using the same test on PRs as well, same workflow, is sufficient for now.

.github/workflows/test.yaml
npetroni: We are integrating the framework first, then will add more/updated images. If you have a public image to suggest, that would be great too.

.github/workflows/test.yaml
fgomulka: I agree with this, fixed in https://github.com/volatilityfoundation/volatility3/pull/789/commits/f96f004b13c15b3c0334f7ca96b2c0459a8f9cf1

test/test_volatility.py
fgomulka: Fixed in https://github.com/volatilityfoundation/volatility3/pull/789/commits/f96f004b13c15b3c0334f7ca96b2c0459a8f9cf1

.github/workflows/test.yaml
fgomulka: Changed in https://github.com/volatilityfoundation/volatility3/pull/789/commits/f96f004b13c15b3c0334f7ca96b2c0459a8f9cf1

.github/workflows/test.yaml
digitalisx: @npetroni All right, I'm glad you have plans to add more. Then this could be enough for now.

test/known_files.json
```diff
@@ -0,0 +1,19 @@
...
+          "8589f1463422884dbf1411aaad278465"
+        ],
+        "0x81eaf418": [
+          "f7a1ae2060a58f8470b97affdb46dccf",
+          "54fd611021fa784912530b8007545986"
+        ],
+        "0x820588e8": "458efbc8fdb859488a6ab2b200cce809"
+      }
+    }
+  }
```
digitalisx: Could you please add new line?

```suggestion
  }
  
```

test/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+
+    parser.addoption("--image-dir", action="append", default=[],
+        help="path to a directory containing images to test")
+
+def pytest_generate_tests(metafunc):
+    """Parameterize tests based on image names"""
+
+    images = metafunc.config.getoption('image')
+    for d in metafunc.config.getoption('image_dir'):
+        images = images + [os.path.join(d, x) for x in os.listdir(d)]
```
digitalisx: Can we use a clearer variable name?

```suggestion
    for image_dir in metafunc.config.getoption('image_dir'):
        images = images + [os.path.join(image_dir, dir) for dir in os.listdir(image_dir)]
```

digitalisx (commented):
@fgomulka Thank you for accept my suggestion, Last I added some NIT comments.

digitalisx (approved):
@fgomulka Thank you for working on my suggest.  I'm satisfied with your work, and I hope the day will come when we can test more operating systems, plugins, various python versions. :)

.github/workflows/test.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+    - name: Download and Extract symbols
+      run: |
+        cd ./volatility3/symbols
+        curl -sLO https://downloads.volatilityfoundation.org/volatility3/symbols/linux.zip
+        unzip linux.zip
```
ikelos, line 39: Doesn't strictly need unzipping, but no problem if you do.

ikelos: Also note, we're just about to merge #630, which may mean you want to prepopulate the cache?

ikelos (approved):
Cool, looks good and @npetroni is happy, so I'm happy!  Thanks!  5:D

PR #790, opened 2022-07-18
Lint: apply code style by `yapf` style file (`.style.yapf`).

digitalisx:

So far, with the contribution of maintainers and members of various communities, we are completing a wonderful project.

With these diverse people together, we were able to meet better ideas and codes.

However, as the number of people increases, various styles of code exist in the project.
Of course, nothing "absolutely" fits the code style, and it's not always a good idea to force it's

Familiar people make contributions with some consideration, but sometimes I think there are difficulties with new contributors. For example, when you're looking at existing code to make a contribution or trying to find a case where certain phrases are used, I think different code styles are a little bit challenging.

I applied the lint according to the currently designated `yapf` setting file (`.styel.yapf`). It's a lot because it's recursively organizing the internal codes. So I checked the changes and tested them.

```shell
> yapf --style .style.yapf -ir .
```

Most changes are simple, such as removing spaces or relocating long arguments, so there is no problem with code execution. And I think this will help a little bit more in the development of a member of the community.
(If there is a consensus later on, it would be nice to have Github Actions that automatically correct the lint!)

If you have any comments on this, please feel free to leave a comment. 🙌


development/banner_server.py
```diff
@@ -45,13 +45,20 @@ def run(self):
                     # If there's no linux symbols, don't cry about it
                     pass
 
-            new_banners = banner_cache.read_new_banners(context, 'BannerServer', potentials, banner_cache.symbol_name,
-                                                        banner_cache.os, progress_callback = PrintedProgress())
+            new_banners = banner_cache.read_new_banners(context,
```
ikelos, line 48: There's no need to keep the `development` directory to the same standards, but it also doesn't hurt.  5:)

ikelos (commented):
Thanks, this all looks fine but you're right, it does touch a lot of files.

I used to apply this regularly but somewhat gave up since not everybody used it and in particular because it's not stable (running it once and running it twice might yield different results).  I think for the sake of keeping things simple (and until we choose a linter that we really like and can apply automatically) I might have to turn this down I'm afraid.  5:S  I've looked at black which is very strict (no options) but I'm still not sold on no space between keyword arguments (`def thing(a = foo)` rather than `def thing(a=foo)`.  It's not a big deal, so if we find one that most people agree on I'll learn to live with it...

So until we decide on that, I think I'm going to leave this to one side, but if you're keen to find a solution then please have a look around at various linters and how easy they'd be to integrate with github, and we can try adding it in.  I just don't want to make a shift in style that touches so much all that often if possible...  Hope that's ok?

digitalisx:
Long time no see @ikelos,
I checked your thoughts and opinions well.
I will suggest it again when I find a Linter suitable for this project.
Therefore, I will close this PR so that we can focus on the remaining PR and challenges.
Perhaps contributors looking for traces related to Lint can search for this PR at any time. 🙂

ikelos:
Yeah, sorry, I've been a little busy/preoccupied recently, but thought I should check in...  5;)
Thanks for understanding!  5:)

PR #791, opened 2022-07-20
Test: add `pytest` in `requirements-test.txt`, some enhancement.

digitalisx:

We recently merged that a great test (#789) has been added to help contributors before contributing to volatility3.

It was confirmed that it works normally on the actual Github Actions.
However, in order to proceed with the test locally as described in the manual, `pytest` must be installed.

```shell
> py.test ./test/test_volatility.py --volatility=vol.py --image win-xp-laptop-2005-06-25.img -k test_windows
zsh: command not found: py.test
```

Currently, this is written in workflow `yml`, but it is not included in the manual and `requirements-test.txt`.
I thought it should be included in the text file because I think it is an essential module for testing.

Also, I modified `.gitignore` to prevent files that occur when `pytest` is executed locally from being included in the `git`.

If you have any comments on this, please feel free to leave a comment. 🙌


.gitignore
```diff
@@ -38,3 +38,7 @@ ENV/
 # Memory dump files
 *.dmp
 *.vmem
+*.img
```
ikelos, line 41: Looks good, but let's cap it at `img` and `dmp`, otherwise this list will grow to everybody's personal way of naming files.  There's no standard, but we also can't support everybody's naming scheme.

ikelos (approved):
Cool, looks good, thanks!  5:)

.gitignore
digitalisx: All right, I'll remember this.

PR #795, opened 2022-07-28
Fix sqlite3 issues in symbol cache

paulkermann:
I practically had [this](https://stackoverflow.com/questions/36829038/no-such-column-error-in-sqlite3) issue in the symbol cache.
I have replaced instances where we do `where local = True` with `where local = 1`.
Also, it looked like at one instance a non-existent column was used (remote) so I have replaced it with `local` and inverted the boolean (line 333).
It looks like insertion works with "native" python types (like in line 316) but lookup does not.
For me this issue caused pdb parsing and isf creation to happen EVERYTIME even though the isf was at present

### DIFF
--- a/volatility3/framework/automagic/symbol_cache.py
+++ b/volatility3/framework/automagic/symbol_cache.py
@@ -216,7 +216,7 @@ class SqliteCache(CacheManagerInterface):
         return result
 
     def get_local_locations(self) -> Generator[str, None, None]:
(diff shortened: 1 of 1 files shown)


ikelos (approved):
Thanks, in later versions of sqlite (section 2.1 of https://www.sqlite.org/datatype3.html), it does support the literals TRUE and FALSE, but since we support python 3.6 still (which was released in 2016) and the change only came in in 2018, I'm going to accept this.  Thanks!  5:)

Issue #801, opened 2022-08-03
Snappy C Library Requirement Not Specific in README


emabrey:
**Is your feature request related to a problem? Please describe.**
The README does not instruct users to install the Snappy c library for an installation using `requirements.txt`, and failing to have that library installed will cause the command `pip3 install -r requirements.txt` to fail with an error.

**Describe the solution you'd like**
Add instructions to the README to install the Snappy C library. For example: `sudo apt-get install libsnappy-dev` or `sudo yum install libsnappy-devel`

**Describe alternatives you've considered**
It may be possible to create a script which does this automatically, and have the user call that instead of pip directly, but that would be an unnecessary complexity.

**Additional information**
The installation fails with a failure to find `snappy-c.h`.

ikelos:
This seems like an issue with `pip install python-snappy`?  We define the requirement, which should in turn define any requirements it has.  Please provide the failure message, since I don't believe this is a volatility issue, but rather a python-snappy issue.

emabrey:
It isn't a `python-snappy` issue. Installing the C library via the system package manager before the call to pip is the best way to do this. Building the library from source and installing it via pip is a nightmare that is completely avoided by using the pre-existing pre-built packages from the system package manager.

The issue is simply that the installation instructions for `python-snappy` tell you to install the C libraries first, as they are transitive dependencies. Without them, `python-snappy` fails with this error, which is caused because it can't find the header files which are installed alongside the devel libraries. Using the pre-built version of 0.5.3 that comes with Kali is seamless, as that means you don't have to build from source so you don't need the devel headers, but having to build for python-snappy 0.6.0 means that unless you have the C library installed first, you won't install python-snappy:0.6.0.

What the error looks like:

```shell
──(emabrey㉿2603-6010-c000-29e4-0000-0000-0000-1227)-[~/volatility3]
└─$ pip3 list | grep snappy 
python-snappy                  0.5.3
                                                                                                                                                                                                       
                                                                                                                                                                                                       
┌──(emabrey㉿2603-6010-c000-29e4-0000-0000-0000-1227)-[~/volatility3]
└─$ pip3 install -r requirements.txt
Defaulting to user installation because normal site-packages is not writeable
Requirement already satisfied: pefile>=2017.8.1 in /usr/lib/python3/dist-packages (from -r requirements.txt (line 2)) (2022.5.30)
[...]

PR #802, opened 2022-08-03
Add Snappy-C Library Instructions to README

emabrey:
ADD the installation instructions for the snappy-c library, to address `snappy-c.h` related installation error.
Issue: #801

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,18 @@ To enable the full range of Volatility 3 functionality, use a command like the o
 pip3 install -r requirements.txt
 ```
 
+### Snappy Library Required for Full Functionality Installation
+If you recieve an error informing you that `snappy-c.h` cannot be found, then you need to install the library and re-run the previous `pip3` command:
+
+DEB:
+```shell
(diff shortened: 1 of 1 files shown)


ikelos:
Thanks, but I'm not keen on providing fixes for every dependency's failure to install correctly.  Pip should explain which package has failed and why, and it's up to the user to determine why that package failed and install it correctly.  I would be willing to point to a bug report on the python-snappy package indicating how to fix the issue, but I think providing our own explanation for a different package's failure to work in certain conditions, sets the wrong tone, and could easily explode into lots and lots of fixes for lots of packages.

emabrey:
Currently `pip3 install -r requirements.txt` _never_ works unless you install the `snappy-c` C libraries first, because the `snappy-python` package doesn't even attempt to install it under any conditions. This isn't a one-off, the `snappy-python` library requires those c libraries to be installed before you install it, or the installation fails with the header error I originally mentioned.

The library tells you to install the C devel library as a pre-requisite dependency:
image

PR #810, opened 2022-08-15
Bugfix: Improve scan via the pdbname_scan

paulkermann:
I had a case where the `PdbSignatureScanner` did find the right pdb. However due to a single paged out address in the kernel module the `is_valid` check failed which caused volatility not to work on my memory dump.
By adding `maximum_invalid_count` we allow ourselves to continue iterating downwards even if some pages were invalid.
The `maximum_invalid_count` is currently defaulted to 100 which was enough to to fix my issue, however we could increase it if is needed.
Otherwise functionality is the same.
Without this extra code I could not run volatility3 on my Win10 1809 memory dump (2GB ram)

### DIFF
--- a/volatility3/framework/automagic/pdbscan.py
+++ b/volatility3/framework/automagic/pdbscan.py
@@ -212,8 +212,7 @@ class KernelPDBScanner(interfaces.automagic.AutomagicInterface):
                                           start = start_scan_address,
                                           page_size = vlayer.page_size,
                                           pdb_names = kernel_pdb_names,
-                                          progress_callback = progress_callback,
-                                          maximum_invalid_count = constants.windows.PE_MAX_EXTRACTION_SIZE // 0x1000)
+                                          progress_callback = progress_callback)
         for kernel in kernels:
             valid_kernel = test_kernel(physical_layer_name, virtual_layer_name, kernel)
             if valid_kernel is not None:
--- a/volatility3/framework/automagic/pdbscan.py
+++ b/volatility3/framework/automagic/pdbscan.py
@@ -212,7 +212,8 @@ class KernelPDBScanner(interfaces.automagic.AutomagicInterface):
                                           start = start_scan_address,
                                           page_size = vlayer.page_size,
                                           pdb_names = kernel_pdb_names,
-                                          progress_callback = progress_callback)
+                                          progress_callback = progress_callback,
+                                          maximum_invalid_count = constants.windows.PE_MAX_EXTRACTION_SIZE // 0x1000)
         for kernel in kernels:
             valid_kernel = test_kernel(physical_layer_name, virtual_layer_name, kernel)
             if valid_kernel is not None:
--- a/volatility3/framework/symbols/windows/pdbutil.py
+++ b/volatility3/framework/symbols/windows/pdbutil.py
@@ -253,7 +253,8 @@ class PDBUtility(interfaces.configuration.VersionableInterface):
                      pdb_names: List[bytes],
                      progress_callback: constants.ProgressCallback = None,
                      start: Optional[int] = None,
-                     end: Optional[int] = None) -> Generator[Dict[str, Optional[Union[bytes, str, int]]], None, None]:
+                     end: Optional[int] = None,
+                     maximum_invalid_count: int = 100) -> Generator[Dict[str, Optional[Union[bytes, str, int]]], None, None]:
         """Scans through `layer_name` at `ctx` looking for RSDS headers that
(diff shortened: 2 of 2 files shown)


volatility3/framework/automagic/pdbscan.py
```diff
@@ -212,7 +212,8 @@ def _method_layer_pdb_scan(self,
                                           start = start_scan_address,
                                           page_size = vlayer.page_size,
                                           pdb_names = kernel_pdb_names,
-                                          progress_callback = progress_callback)
+                                          progress_callback = progress_callback,
+                                          maximum_invalid_count = constants.windows.PE_MAX_EXTRACTION_SIZE // 0x1000)
```
ikelos: This seems like a bad calculation?  If the PE_MAX_EXTRACTION_SIZE is the maximum size before we stop trying to load a PE file, and we reduce it to the maximum number of pages we could ever extract, this essentially makes the check redundant.  We'll stop because of size before we ever reach the invalid limit.  There's a default, so this value should be removed and the default used unless there's a good reason for bumping it so high?

volatility3/framework/symbols/windows/pdbutil.py
```diff
@@ -253,7 +253,8 @@ def pdbname_scan(cls,
                      pdb_names: List[bytes],
                      progress_callback: constants.ProgressCallback = None,
                      start: Optional[int] = None,
-                     end: Optional[int] = None) -> Generator[Dict[str, Optional[Union[bytes, str, int]]], None, None]:
+                     end: Optional[int] = None,
+                     maximum_invalid_count: int = 100) -> Generator[Dict[str, Optional[Union[bytes, str, int]]], None, None]:
```
ikelos, line 257: This seems alright, since in general we're not using the specific pages of the PE file itself, but now seems like a good time to document the parameters and what they do in the docstring please.  Happy to help define things if needed.

volatility3/framework/symbols/windows/pdbutil.py
```diff
@@ -278,10 +279,15 @@ def pdbname_scan(cls,
                                                               sections = [(start, end - start)]):
             mz_offset = None
             sig_pfn = signature_offset // page_size
+            current_invalid_counter = 0
 
             for i in range(sig_pfn, min_pfn, -1):
-                if not ctx.layers[layer_name].is_valid(i * page_size, 2):
+                if current_invalid_counter > maximum_invalid_count:
```
ikelos, line 293: By not bombing out early, this may increase how long is spent on each pdb file, multiplied by the number of possible pdb files the scanner finds.  That should be limited based on the `pdb_names`, but we'll need to consider it if we ever scan for all pdb files for some reason.

volatility3/framework/symbols/windows/pdbutil.py
ikelos: Is 100 the lowest we can reasonably go?  How many pages are likely to have been paged out of the kernel (in fact, why are kernel pages paged out at all)?  Will this now hit ghost images and/or allow someone to partially load a different kernel early in memory to throw volatility off?  It's not a joke, we've had papers published about tricking volatility's automagic in the past, with tricks as similarly daft, so I'm keen to keep our tolerance to errors as low as possible.

ikelos (commented):
Thanks, at least one change and a some documentation and questions, but otherwise seems reasonable because we typically don't need the entirety of the kernel, we just want to make sure we've got the one that's being run and not some random chunk of memory that someone loaded which looks like a kernel.

volatility3/framework/automagic/pdbscan.py
paulkermann: seems reasonable

volatility3/framework/symbols/windows/pdbutil.py
paulkermann: Some kernel pages can be invalid because they are paged out or discarded and marked invalid. In my image there were 25 invalid pages. Also, this function is layer agnostic and can be used by others (for non kernel layers) where paged can be swapped out.

volatility3/framework/symbols/windows/pdbutil.py
paulkermann: I agree

ikelos (approved):
Since this is specifically for identifying the PDB GUID/age and the contents of the kernel aren't needed for processing in anyway, I think skipping broken pages is acceptable.  We'll merge it (although it won't make it into the .2.4.0 release) and keep an eye on it for any images it may break.

PR #813, opened 2022-08-20
Bugfix: Calling with --cache-path now updates Several variables depen…

### LINKED ISSUE #803: Bug with CACHE_PATH
`constants.CACHE_PATH` is updated if `cache_path` is in the arguments for the cli.
However `LINUX_BANNERS_PATH`, `MAC_BANNERS_PATH` and `IDENTIFIERS_PATH` are not updated accordingly.
This causes those files to be created in the default cache path.
We need to update all those paths as well with the new cache path.

jeffguy:
fixes #803
The problem was that several variables in constants/__init__.py are computed based on constants.CACHE_PATH. One example is LINUX_BANNERS_PATH. Without this patch, if you're doing volatility3 development related to the banner code, you may want to run ./vol.py --cache-path /tmp/volatility-development so you don't have to `rm -rf $VOLATILITY_CACHE_DIR.

Before my patch, constants/__init__.py sets the default value of constants.CACHE_PATH, then computes LINUX_BANNERS_PATH based on that default value. Later, cli/__init__.py updates constants.CACHE_PATH to the value passed as --cache-path, but it doesn't update LINUX_BANNERS_PATH.

My fix moves all initialization that depends on constants.CACHE_PATH into a function (update_cache_path). Then, constants/__init__.py calls this function with the default CACHE_PATH value, so all initialization happens like before. But now, cli/__init__.py can call constants.update_cache_path() to re-do all those CACHE_PATH-related initialization steps based on the new value of CACHE_PATH, thus propagating the value the user asked vol.py to use.

Thanks for your hard work on this useful tool!

### DIFF
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -193,7 +193,7 @@ class CommandLine:
                                             for p in partial_args.symbol_dirs.split(";")] + constants.SYMBOL_BASEPATHS
 
         if partial_args.cache_path:
-            constants.CACHE_PATH = partial_args.cache_path
+            constants.update_cache_path(partial_args.cache_path)
 
         if partial_args.log:
             file_logger = logging.FileHandler(partial_args.log)
--- a/volatility3/framework/constants/__init__.py
+++ b/volatility3/framework/constants/__init__.py
@@ -60,21 +60,24 @@ LOGLEVEL_VVV = 7
 LOGLEVEL_VVVV = 6
 """Logging level for -vvvv"""
 
-CACHE_PATH = os.path.join(os.path.expanduser("~"), ".cache", "volatility3")
-"""Default path to store cached data"""
+def update_cache_path(cache_path):
+    CACHE_PATH = cache_path
+    """Default path to store cached data"""
 
-if sys.platform == 'win32':
-    CACHE_PATH = os.path.join(os.environ.get("APPDATA", os.path.expanduser("~")), "volatility3")
-os.makedirs(CACHE_PATH, exist_ok = True)
+    if sys.platform == 'win32':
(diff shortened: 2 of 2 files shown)


digitalisx:
Hello, @jeffguy
Thank you for your work :)

There may be various opinions, but I think the purpose of `constants` is to make values that can be used frequently accessible from anywhere. I agree with the purpose of updating, but We need to think about whether the logic of judgment should exist in the `constants` file and whether it should be provided as a function.

And it seems that all codes using `constants.CACHE_PATH` have been not changed and not sufficiently verified. I think we need to make some more changes and verify. The scope would have been changed as the variable entered the function It will probably be unavailable for all code using `constants.CACHE_PATH`.

I think it can be merged as soon as possible if the maintainers' opinions on this, improved methods, and sufficient verification are merged.

jeffguy:
You're right. Unfortunately, my development machine is airgapped.

I added and pushed the following line to my personal github:
global CACHE_PATH, LINUX_BANNERS_PATH, MAC_BANNERS_PATH, IDENTIFIERS_PATH

I don't see how to modify the pull request to include that most recent commit.

Regarding design decisions like whether to move these four variables from constants/__init__.py to some more-appropriate section, that's a good question. The fact the --cache-path cli argument exists shows these variables are alreday not being treated as constants, So the code I'm proposing doesn't make the situation worse.

If you'd like someone who better understands the volatility design philosophy to put some more thought into refactoring these variables before we fix this issue 813, and you want me to cancel this pull request, I won't take offense. Let me know and I'll cancel this PR.

ikelos:
This has now been deprecated by the merging on #819.  There's a follow-up #838 that adds warning messages for using the previous variables (which at the moment will instead throw an exception if the old variables are accessed for reading).  As such, I'm going to close this, but feel free to reopen it or submit a new pull request if the issue still exists...

PR #814, opened 2022-08-20
Fix: typo for symbol cache.

digitalisx:
## Description

I found some typo for while analyze code, So I fixed it.


ikelos (approved):
Looks good.  Since I don't think anyone will be dependent upon that variable yet, I'm happy to push the change.  If it had been around longer we'd probably have had to make a new variable and set the old one to it (and deprecate the old one).

PR #817, opened 2022-08-24
BUGFIX - layer given by lower layer is not returned from the stacker

paulkermann:
A `return layer` was missing in the stacker which caused the stacker to scan either way.
I have added a log to indicate that the layer was given to us by the lower layer and return it instead of scanning which was not the intended behavior.


paulkermann:
```
...
Level 9  volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DEBUG    volatility3.framework.automagic.windows: DTB was given to us by base layer: 0x1aa000
...
```
Debug output looks like this

ikelos (approved):
Looks good, I'm not sure why we were letting it fall through, there's no efficiency in doing that.  Good catch, thanks!  5:)

PR #818, opened 2022-08-24
refs #713 add a vad.get_size() method and fix several off-by-one issues with calculating vad size

iMHLv2:
See the comments in #713, in particular this description of the changes:

Historically, even back to vol2, the `get_end()` function has returned the last accessible byte in the range, so I'm keeping that unchanged. I added a `get_size()` method that calculates the total size in bytes and fixed a number of plugins that had off-by-one issues. Anyone relying on `get_end()` to calculate size in custom plugins (outside of the core code base) will still have one-by-one issues, but they can fix it by using the new `get_size()` - IMO that is better than changing the behavior of `get_end()` and potentially breaking things "silently."

### DIFF
--- a/volatility3/framework/constants/__init__.py
+++ b/volatility3/framework/constants/__init__.py
@@ -40,7 +40,7 @@ BANG = "!"
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 2  # Number of releases of the library with a breaking change
 VERSION_MINOR = 4  # Number of changes that only add to the interface
-VERSION_PATCH = 1  # Number of changes that do not change the interface
+VERSION_PATCH = 0  # Number of changes that do not change the interface
 VERSION_SUFFIX = ""
 
 # TODO: At version 2.0.0, remove the symbol_shift feature
--- a/volatility3/framework/constants/__init__.py
+++ b/volatility3/framework/constants/__init__.py
@@ -39,7 +39,7 @@ BANG = "!"
 
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 2  # Number of releases of the library with a breaking change
-VERSION_MINOR = 3  # Number of changes that only add to the interface
+VERSION_MINOR = 4  # Number of changes that only add to the interface
 VERSION_PATCH = 1  # Number of changes that do not change the interface
 VERSION_SUFFIX = ""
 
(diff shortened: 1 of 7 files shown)


volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -234,6 +234,10 @@ def get_end(self):
 
         raise AttributeError("Unable to find the ending VPN member")
 
+    def get_size(self) -> int:
```
ikelos, line 237: Thanks, this is an additive change, which I think therefore requires the minor version number to be bumped to 2.4.0 please (and the four plugins that make use of it to bump their `_required_framework_version`).  Otherwise it looks good, and happy to merge after that...  5:)

ikelos (commented):
Looks good, just needs some versioning please...  5;)

volatility3/framework/symbols/windows/extensions/__init__.py
ikelos: I'd quite like to get this merged before the impending code freeze.  It should just be a case of bumping version numbers and then it's good to merge...  5:)

digitalisx: Hello @ikelos, if he's busy I'm willing to take care of the needs of this branch.
However, in this situation, there seems to be no way to work on the branch where PR is open in this Repository.
Is it possible to work on PR only by the person who suggests it?

ikelos: Hi @digitalisx, I'm afraid so.  You'd need to construct your own pull request which would then be worked on separately (and obsolete the original PR).  Luckily it looks like this has seen some progress and should be good to commit...  5:)

volatility3/framework/constants/__init__.py
```diff
@@ -39,7 +39,7 @@
 
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 2  # Number of releases of the library with a breaking change
-VERSION_MINOR = 3  # Number of changes that only add to the interface
+VERSION_MINOR = 4  # Number of changes that only add to the interface
 VERSION_PATCH = 1  # Number of changes that do not change the interface
```
digitalisx:
```suggestion
VERSION_PATCH = 0  # Number of changes that do not change the interface
```

digitalisx (commented):
Hello @iMHLv2, Thank you for your work!
If my understanding is correct, it should be bumped version to `2.4.0`.
Could you please check this? According to the this PR, the version will be bumped to `2.4.1`.

ikelos:
Looks good, thanks!  5:)

PR #819, opened 2022-08-24
Core: Allow for deprecation of constants gracefully

### LINKED ISSUE #803: Bug with CACHE_PATH

ikelos:
This fixes the issue with cache filepaths need to be constructed.  We now force users of the values to do that manually (since otherwise it's duplication of data that evidently can get out of sync).  This introduces a new means of deprecating constants, which will also come in handy in the future, but it relies on a feature only introduces in 3.7.0 (module level `__getattr__`).  I think 3.7 is old enough now to be acceptable as a minimum version, but it is a concern.

This would fix #803 and replace #813

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -7,10 +7,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Set up Python 3.6
+    - name: Set up Python 3.7
       uses: actions/setup-python@v2
       with:
-        python-version: '3.6'
+        python-version: '3.7'
 
     - name: Install dependencies
       run: |
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -7,10 +7,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Set up Python 3.7
+    - name: Set up Python 3.6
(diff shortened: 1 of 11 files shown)


digitalisx (commented):
Thank you for your work @ikelos!
I found some point need to update for change required python version.
I'm not sure if it's good to be included in this PR or if it's right to be proposed as a separate PR again, but it seems necessary to change it as well. :)

https://github.com/volatilityfoundation/volatility3/blob/2529075977f53a9e3f3a023034bf774aeb9e90a5/setup.py#L37

https://github.com/volatilityfoundation/volatility3/blob/2529075977f53a9e3f3a023034bf774aeb9e90a5/README.md?plain=1#L23

ikelos:
You're quite right, thanks.  5:)

ikelos:
Just taking out the python-3.7 changes for the next release.  The impact should be minimal, but means you may not get warned about using deprecated variables.

PR #826, opened 2022-09-03
Fix: duplicate module for `windows.ldrmodules` plugin.

digitalisx:
I found duplicate module in `windows.ldrmodules` plugin, and I fixed it.


ikelos (approved):
Also good, thank you!  5:)

PR #827, opened 2022-09-03
Fix: duplicate paragraph for `LICENSE`.

digitalisx:
## Description
I found duplicate paragraph in `LICENSE.txt`, and I fixed it.
It is also applied to the foundation's [website](https://www.volatilityfoundation.org/license), so if possible, it will need to be modified.


ikelos (approved):
Yep, looks good, thanks!  5:D

PR #834, opened 2022-09-12
Bugfix CSV format line terminator

paulkermann:
In my windows system formatting to csv via the `-r csv` yielded extra new line feeds when output was redirected to a file.
I found a stackoverflow issue which showed the solution here.


ikelos:
Ok, I'm not sure of the significance of this, but it looks ok to me...

PR #835, opened 2022-09-13
Added new containing address flag to vadinfo plugin

paulkermann:
Hey,
added new flag to `windows.vadinfo` plugin to also check the containing address in a certain VAD


ikelos:
This feels as though the address filter isn't all that useful since it's highly specific.  Could both the address and containing-address be rolled into a single filter?  @iMHLv2 could you comment on this please?

paulkermann:
You think I can make the --address flag also search addresses inside a given VAD?
I added a second flag to not change current API

paulkermann:
I merged the two flags together, as you requested

ikelos (approved):
Looks good, thanks.  5:)

PR #845, opened 2022-10-04
Update cachedump.py

SynacktivCerv:
Change the hash format in order to be compatible with John the Ripper.


ikelos (commented):
Hi there, thanks very much for your pull request, unfortunately it somewhat goes against two principles of volatility which mean it won't get merged unfortunately.

Firstly, this is altering the plugin (which might be run by other plugins, used by other tools) to produce human readable output (ie, only to be used by humans).  I understand it would be really handy for the plugin to give you something you can just drop into john the ripper, but version 3 was explicitly designed to be used by both programs and humans.

Secondly the data you'd be returning in the hash field is actually multiple different pieces of data, namely the username (twice) and the domain, which again is using the plugins for human output without considering how tools and programs that use volatility as a library might best make use of the data.  We try hard to ensure that each individual piece of data is released as a separate item for processing by the consumer, rather than as blobs that may need further processing (even if tools like john already do that processing).

In both instances, the plugin is returning the data it should and what you're asking for is something that can convert the data another more human readable format immediately.  I'd suggest either a custom plugin that runs cachedump and produces different output lines, or just piping the output through a simple sed script?

Sorry we're not going to accept the change, but hopefully you understand how this would affect the program's use and how difficult it would be to cater to everyone's individual preferences for human output.  5:(  Thanks again for your contribution!

ikelos:
Hi there

PR #852, opened 2022-10-25
Windows Driver signing enforcement plugin

atcuno:
A common technique by modern Windows rootkits is disabling Driver Signing Enforcement so that unsigned kernel rootkits can load. Rapid 7 documented many examples here:

https://www.rapid7.com/blog/post/2021/12/13/driver-based-attacks-past-and-present/

To disable DSE, malware will set the `g_CiOptions` global variable in the kernel to 0. 

This plugin locates this variable in memory and then reports its status and value. 

If this global variable is disabled on Windows 10 systems then it means the system is either in developer mode or that a rootkit purposely overwrote the value. Neither would be expected in production settings.

### DIFF
--- a/volatility3/framework/plugins/windows/driver_signing_enforcement.py
+++ b/volatility3/framework/plugins/windows/driver_signing_enforcement.py
@@ -1,0 +1,112 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import logging
+from typing import List
+
+from volatility3.framework import exceptions, renderers, interfaces, symbols
+from volatility3.framework.configuration import requirements
+from volatility3.framework.objects import utility
+from volatility3.plugins.windows import info, modules
+from volatility3.framework.symbols.windows import pdbutil
+
+vollog = logging.getLogger(__name__)
+
+class driver_signing_enforcement(interfaces.plugins.PluginInterface):
+    """Reports the status of Driving Signing Enforcement"""
+
+    _required_framework_version = (2, 0, 0)
+    _version = (1, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(name = 'kernel', description = 'Windows kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.PluginRequirement(name = 'modules', plugin = modules.Modules, version = (1, 0, 0)),
+            requirements.VersionRequirement(name = 'pdbutil', component = pdbutil.PDBUtility, version = (1, 0, 0)),
+        ]
+
+    def _get_ci_dll(self, kernel):
+        """
+        Finds the CI.DLL kernel module. This module is used for
+        Code Integrity enforcement, and is targetted by malware to disable security checks.
+
+        Args:
+            kernel: interfaces.context.ModuleInterface of the kernel
+        Returns:
+            The `_LDR_DATA_TABLE_ENTRY` of CI.DLL
+            None if module cannot be located
+        """
+        for mod in modules.Modules.list_modules(self.context, 
+                                                kernel.layer_name,
+                                                kernel.symbol_table_name):
+     
+            try:
+                module_name = mod.BaseDllName.get_string()
+            except exceptions.InvalidAddressException:
+                continue
+
+            if module_name.lower() == "ci.dll":
+                return mod
+
+        return None
+
+    def _check_win10_64bit(self, kernel):
+        """
+        Checks if the analyzed memory sample is from
+        a Windows 10 64bit system.
+
+        Returns:
+            bool
+        """
+        kuser = info.Info.get_kuser_structure(self.context, 
+                                              kernel.layer_name,
+                                              kernel.symbol_table_name)
+
+        nt_major_version = int(kuser.NtMajorVersion)
+
+        is_64bit = symbols.symbol_table_is_64bit(self.context, kernel.symbol_table_name)
+
+        return nt_major_version >= 6 and is_64bit
+    
+    def _generator(self):
+        """
+        Uses the PDB of CI.DLL to locate g_CiOptions and report its value
+        """
+        kernel = self.context.modules[self.config['kernel']]
+
+        if not self._check_win10_64bit(kernel):
+            vollog.error("This plugin only supports 64-bit Windows 10 or later samples.")
+            return
+
+        ci_mod = self._get_ci_dll(kernel)
+        if ci_mod is None:
+            vollog.error("CI.DLL not found in the kernel module list. Cannot proceed.")
+            return
+
+        # we do not have a fallback mechanism, so if the PDB cannot load then we are done
+        try:
+            ci_symbols = pdbutil.PDBUtility.symbol_table_from_pdb(self.context,
+                                                                  interfaces.configuration.path_join(
+                                                                      self.config_path, 'ci'),
+                                                                  kernel.layer_name,
+                                                                  "ci.pdb",
+                                                                  ci_mod.DllBase,
+                                                                  ci_mod.SizeOfImage)
+
+        except exceptions.VolatilityException as e:
+            vollog.error("Unable to analyze the PDB file for CI.DLL: {}.".format(str(e)))
+            return
+
+        ci_module = self.context.module(ci_symbols, layer_name = kernel.layer_name, offset = ci_mod.DllBase)
+
+        options_address = ci_module.get_absolute_symbol_address("g_CiOptions")
+      
+        options = kernel.object(object_type = "unsigned int", offset = options_address, absolute = True)
+
+        yield (0, (options != 0, options))
+
+    def run(self):
+        return renderers.TreeGrid([("Driver Signing Enforcement", bool), ("g_CiOptions Value", int)],
+                                  self._generator())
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/driver_signing_enforcement.py
```diff
@@ -0,0 +1,125 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import logging
+from typing import List
+
+from volatility3.framework import exceptions, renderers, interfaces, symbols
+from volatility3.framework.configuration import requirements
+from volatility3.framework.objects import utility
```
digitalisx: Found unused module.

```suggestion
```

volatility3/framework/plugins/windows/driver_signing_enforcement.py
```diff
@@ -0,0 +1,125 @@
...
+            requirements.ModuleRequirement(name = 'kernel', description = 'Windows kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.PluginRequirement(name = 'modules', plugin = modules.Modules, version = (1, 0, 0)),
+            requirements.VersionRequirement(name = 'pdbutil', component = pdbutil.PDBUtility, version = (1, 0, 0)),
+        ]
+
+    def _get_ci_dll(self, kernel):
+        """
+        Finds the CI.DLL kernel module. This module is used for
+        Code Integrity enforcement, and is targetted by malware to disable security checks.
```
digitalisx: Found some typo.
```suggestion
        Code Integrity enforcement, and is targeted by malware to disable security checks.
```

volatility3/framework/plugins/windows/driver_signing_enforcement.py
```diff
@@ -0,0 +1,125 @@
...
+https://www.rapid7.com/blog/post/2021/12/13/driver-based-attacks-past-and-present/
+
+To disable DSE, malware will set the `g_CiOptions` global variable in the kernel to 0.
+
+This plugin locates this variable in memory and then reports its status and value.
+
+If this global variable is disabled on Windows 10 systems then it means the system is either in
+developer mode or that a rootkit purposely overwrote the value. Neither would be expected in production settings.
+"""
+class driver_signing_enforcement(interfaces.plugins.PluginInterface):
```
digitalisx: Most of the classes in the plugin that currently exist use Pascal Case, and in fact, the class name also recommends that case. Of course, there are no clear guidelines or criteria for this in the current project. What do you think?

```suggestion
class DriverSigningEnforcement(interfaces.plugins.PluginInterface):
```

digitalisx (commented):
Thank you for implementing a very good plugin. I left my brief comments.

atcuno:
Thank you @digitalisx, I batched committed all your feedback.

volatility3/framework/plugins/windows/driver_signing_enforcement.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def _check_win10_64bit(self, kernel):
+        """
+        Checks if the analyzed memory sample is from
+        a Windows 10 64bit system.
+
+        Returns:
+            bool
+        """
+        kuser = info.Info.get_kuser_structure(self.context,
```
ikelos, line 75: You use the info plugin without having a dependency on it in the requirements.

volatility3/framework/plugins/windows/driver_signing_enforcement.py
```diff
@@ -0,0 +1,124 @@
...
+        a Windows 10 64bit system.
+
+        Returns:
+            bool
+        """
+        kuser = info.Info.get_kuser_structure(self.context, 
+                                              kernel.layer_name,
+                                              kernel.symbol_table_name)
+
+        nt_major_version = int(kuser.NtMajorVersion)
```
ikelos, line 79: Does this need converting to an int?  I'm trying not to commit bad habbits, so only if this *doesn't* behave like an int should we be doing this...

volatility3/framework/plugins/windows/driver_signing_enforcement.py
```diff
@@ -0,0 +1,124 @@
...
+        ci_module = self.context.module(ci_symbols, layer_name = kernel.layer_name, offset = ci_mod.DllBase)
+
+        options_address = ci_module.get_absolute_symbol_address("g_CiOptions")
+      
+        options = kernel.object(object_type = "unsigned int", offset = options_address, absolute = True)
+
+        yield (0, (options != 0, options))
+
+    def run(self):
+        return renderers.TreeGrid([("Driver Signing Enforcement", bool), ("g_CiOptions Value", int)],
```
ikelos, line 123: This feels like it will only return a single row ever, are there other options in the structure it would be useful to return?  Seems a waste to do all the work to pull out this one location, and not stop and see if there's any other global flags/variables we could tell people about?

ikelos (commented):
Looks good, save for a couple trivial points.  Just wondering if there's any way of making it return more output/be more useful than just looking at a single switch?

volatility3/framework/plugins/windows/driver_signing_enforcement.py
```diff
@@ -0,0 +1,124 @@
...
+                module_name = mod.BaseDllName.get_string()
+            except exceptions.InvalidAddressException:
+                continue
+
+            if module_name.lower() == "ci.dll":
+                return mod
+
+        return None
+
+    def _check_win10_64bit(self, kernel):
```
digitalisx, line 67:
```suggestion
    def _check_win10_64bit(self, kernel) -> bool:
```

volatility3/framework/plugins/windows/driver_signing_enforcement.py
```diff
@@ -0,0 +1,124 @@
...
+
+        ci_module = self.context.module(ci_symbols, layer_name = kernel.layer_name, offset = ci_mod.DllBase)
+
+        options_address = ci_module.get_absolute_symbol_address("g_CiOptions")
+      
+        options = kernel.object(object_type = "unsigned int", offset = options_address, absolute = True)
+
+        yield (0, (options != 0, options))
+
+    def run(self):
```
digitalisx, line 122:
```suggestion
    def run(self) -> renderers.TreeGrid:
```

volatility3/framework/plugins/windows/driver_signing_enforcement.py
```diff
@@ -0,0 +1,124 @@
...
+                                              kernel.layer_name,
+                                              kernel.symbol_table_name)
+
+        nt_major_version = int(kuser.NtMajorVersion)
+
+        is_64bit = symbols.symbol_table_is_64bit(self.context, kernel.symbol_table_name)
+
+        return nt_major_version >= 6 and is_64bit
+    
+    def _generator(self):
```
digitalisx, line 85:
```suggestion
    def _generator(self) -> Iterator[Tuple]:
```

volatility3/framework/plugins/windows/driver_signing_enforcement.py
```diff
@@ -0,0 +1,124 @@
...
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(name = 'kernel', description = 'Windows kernel',
+                                                     architectures = ["Intel32", "Intel64"]),
+            requirements.PluginRequirement(name = 'modules', plugin = modules.Modules, version = (1, 0, 0)),
+            requirements.VersionRequirement(name = 'pdbutil', component = pdbutil.PDBUtility, version = (1, 0, 0)),
+        ]
+
+    def _get_ci_dll(self, kernel):
```
digitalisx, line 42:
```suggestion
    def _get_ci_dll(self, kernel) -> Optional[interfaces.objects.ObjectInterface]:
```

digitalisx (commented):
Thank you for apply @atcuno,
I suggested an indication of the return type.
It can help many contributor with a kinder code. :)

volatility3/framework/plugins/windows/driver_signing_enforcement.py
atcuno: I copy/pasted this from netscan as I wasn't sure how to it before:

https://github.com/volatilityfoundation/volatility3/blob/8f083ed1dd287a0f726f48f40d8a129cfa5f24bc/volatility3/framework/plugins/windows/netscan.py#L118

I will update here, but a cleanup of netscan would be good too since its a good example of integrating custom types specific to a version.

volatility3/framework/plugins/windows/driver_signing_enforcement.py
ikelos: I'm not sure Iterator is imported, I'm all for adding the typing information, but it also needs to have the `from typing import Iterator` with it, if that's not already there.  That might be tricky to do in a suggestion, but the suggestion isn't that handy if it can't be automatically applied and needs more work on the file afterwards... 5:P

volatility3/framework/plugins/windows/driver_signing_enforcement.py
ikelos: Same again here for `Optional`.

volatility3/framework/plugins/windows/driver_signing_enforcement.py
```diff
@@ -0,0 +1,124 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import logging
+from typing import List
```
digitalisx, line 5: Thank you for your comment @ikelos,
That's right, I think I forgot to present the necessary modules as I used the function to suggest a better way.
If the proposal below is included, I think it will be more stable.

```suggestion
from typing import Iterator, List, Optional
```

atcuno:
This is now protected by Patch Guard and rootkits reset the value back anyway. I will close this as its not a useful plugin anymore.

PR #873, opened 2022-11-29
Fix: ubuntu `runs-on` version for workflow error.

digitalisx:
## Description
Hello, everyone in the community! :)
A while ago, I saw a test error occurring during the PR (#656) of another contributor.
After analyzing the error log and the user's commit, I checked that there was a problem with our workflow script.
```
> Run actions/setup-python@v2
Version 3.6 was not found in the local cache
Error: Version 3.[6](https://github.com/volatilityfoundation/volatility3/actions/runs/3573910567/jobs/6008531558#step:3:7) with arch x64 not found
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```
Follow the [link](https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json) that guides you through to find Python `3.6` and you will see that `ubuntu-20.04` is the last supported version.

However, the `runs-on` version we specified in [workflow](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#choosing-github-hosted-runners) is `ubuntu-latest`, which means it can also work on `ubuntu-22.04`.

In conclusion, it is assumed that the test, which was originally operated at `ubuntu-20.04`, was changed to `ubuntu-22.04` at some point inside Github. That's probably why an error occurs because Python `3.6` cannot be found for `ubuntu-22.04` environment.

We attempt to update Python's version to `3.7`, which is the minimum required by the recent project, so there is a way to upload the Python version, but given the direction of the project I am currently watching, the project seems not ready yet.

Therefore, I considered lowering the version of ubuntu to be tested in a safe way, which works normally.

### DIFF
--- a/.github/workflows/build-pypi.yml
+++ b/.github/workflows/build-pypi.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.6"]
+        python-version: ["3.6.7"]
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
--- a/.github/workflows/build-pypi.yml
+++ b/.github/workflows/build-pypi.yml
@@ -25,6 +25,8 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
+      env:
+        AGENT_TOOLSDIRECTORY: /opt/hostedtoolcache
 
     - name: Install dependencies
       run: |
--- a/.github/workflows/build-pypi.yml
+++ b/.github/workflows/build-pypi.yml
@@ -15,18 +15,16 @@ on:
 jobs:
 
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: ["3.6.7"]
+        python-version: ["3.6"]
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


ikelos:
Thanks, I'm happy to resolve this either way (by dropping ubuntu down a version or bumping the required python version up to 3.7).  I'll leave it up to @npetroni to decide whether 3.7 is a reasonable minimum yet or not...

npetroni:
Thanks, this is a good find. The explanation makes sense.

I think this patch should go onto the release candidate branch at a minimum. Otherwise, we might see spurious failures on that branch as we prepare for release. We shouldn't bump Python versions for the release that is imminent.

Regarding the overall direction of the project, I wanted to do some research before weighing in. As far as I can tell, there are "older" Linux distributions, which are still supported, that ship with Python 3.6 as the default. For example, RHEL 8. As far as I can tell, most/all of those do have a way to easily install 3.7 or later in addition, it's jut not the default. So I think I can get behind a version bump after this release.

As an integration strategy, I think it would be reasonable to merge this patch into develop (and release branch) then do all of the work necessary to switch to 3.7 (or 3.8?) after this release. We could update the Actions as part of that work.

ikelos:
Ok, I'll cherry pick it once it's been merged (changing the base seemed to pull in other changes).  I think we've already got the bulk of the changes needed to get us up to 3.7 ready in #838, so that should be fairly quick to do too.  Thanks for weighing in @npetroni .  5:)

ikelos:
Ok, these changes were applied to release/v2.4.0 in `commit e694713b`.

PR #877, opened 2022-12-06
Core: Apply black to the entire codebase and add workflow

ikelos:
Just so there's some visibility over the application of black and what it would look like before it goes into the main develop.


ikelos:
Applied manually because the code had changed and it was easier to rerun black than rebase.

PR #879, opened 2022-12-08
Fix: workflow `runs-on` version to `ubuntu-20.04`.

digitalisx:
## Description

This PR has come from this PR (#873).
We can use the minimum python version of bump to `3.7` through recently work.
I try to unify all of workflows because they use the `ubuntu-latest`.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   lint:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - uses: psf/black@stable
--- a/.github/workflows/build-pypi.yml
+++ b/.github/workflows/build-pypi.yml
@@ -15,7 +15,7 @@ on:
 jobs:
 
   build:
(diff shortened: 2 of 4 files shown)


ikelos:
I wasn't sure if we were targetting an old ubuntu intentionally?  Happy to make the change if @npetroni is happy with it...

npetroni:
> I wasn't sure if we were targetting an old ubuntu intentionally? Happy to make the change if @npetroni is happy with it...

I'm not sure if I care which specific Ubuntu version we use, but it does seem possible that a "lesson learned" from using `ubuntu-latest` should be that it can start breaking without warning. I think I'd prefer picking a named Ubuntu version. Moving to 22.04 seems reasonable but I'm also fine with keeping 20.04 since that will be supported for a couple more years and represents the "lagging edge" of our user base.

digitalisx:
@npetroni Thank you for your opinion.
Then, would it be a little better to unify the version of ubuntu used in this project to `20.04`?

ikelos:
Yes please, let's go with `ubuntu-20.04` for both then...

ikelos (approved):

npetroni:
Thanks all! Sounds good.

Issue #830, opened 2022-09-07
Is there any way to dump processes in the Linux environment in volatility3?


resposo:
Hello

In a Windows environment, the --dump option allows process dumps, but it does not work in a Linux environment. Is there a way to solve this?

Please let me know if anyone knows how to dump a process in a Linux environment or if they know how to plug-in

atcuno:
Volatility 3 process enumeration plugins for Linux now support `--dump`

PR #884, opened 2022-12-13
First attempt at adding a --dump option to linux.proc

eve-mem:
Attempt to add a `--dump` option to `linux.proc` in an am to be similar to `windows.vadinfo --dump`.

`linux.proc` works in the same way as before without the `--dump` option. (Note: this still has the repeating file path bug)

```
python vol.py -f linux-sample-1.dmp linux.proc
Volatility 3 Framework 2.4.1
PID     Process Start   End     Flags   PgOff   Major   Minor   Inode   File Path       File output

1       init    0x400000        0x409000        r-x     0x0     8       1       652843  /sbin/sbin/init Disabled
1       init    0x608000        0x609000        rw-     0x8000  8       1       652843  /sbin/sbin/init Disabled
1       init    0x23fb000       0x241c000       rw-     0x0     0       0       0       [heap]  Disabled
1       init    0x7f489741c000  0x7f489741e000  r-x     0x0     8       1       913938  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libdl-2.13.so    Disabled
1       init    0x7f489741e000  0x7f489761e000  ---     0x2000  8       1       913938  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libdl-2.13.so    Disabled
1       init    0x7f489761e000  0x7f489761f000  r--     0x2000  8       1       913938  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libdl-2.13.so    Disabled
1       init    0x7f489761f000  0x7f4897620000  rw-     0x3000  8       1       913938  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libdl-2.13.so    Disabled
1       init    0x7f4897620000  0x7f48977a2000  r-x     0x0     8       1       913923  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libc-2.13.so     Disabled
1       init    0x7f48977a2000  0x7f48979a1000  ---     0x182000        8       1       913923  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libc-2.13.so     Disabled
1       init    0x7f48979a1000  0x7f48979a5000  r--     0x181000        8       1       913923  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libc-2.13.so     Disabled
1       init    0x7f48979a5000  0x7f48979a6000  rw-     0x185000        8       1       913923  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libc-2.13.so     Disabled
1       init    0x7f48979a6000  0x7f48979ab000  rw-     0x0     0       0       0       Anonymous Mapping       Disabled
1       init    0x7f48979ab000  0x7f48979c9000  r-x     0x0     8       1       913958  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libselinux.so.1  Disabled
1       init    0x7f48979c9000  0x7f4897bc8000  ---     0x1e000 8       1       913958  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libselinux.so.1  Disabled
1       init    0x7f4897bc8000  0x7f4897bc9000  r--     0x1d000 8       1       913958  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libselinux.so.1  Disabled
1       init    0x7f4897bc9000  0x7f4897bca000  rw-     0x1e000 8       1       913958  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libselinux.so.1  Disabled
1       init    0x7f4897bca000  0x7f4897bcb000  rw-     0x0     0       0       0       Anonymous Mapping       Disabled
1       init    0x7f4897bcb000  0x7f4897c0a000  r-x     0x0     8       1       913982  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libsepol.so.1    Disabled
1       init    0x7f4897c0a000  0x7f4897e09000  ---     0x3f000 8       1       913982  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libsepol.so.1    Disabled
1       init    0x7f4897e09000  0x7f4897e0a000  r--     0x3e000 8       1       913982  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libsepol.so.1    Disabled
1       init    0x7f4897e0a000  0x7f4897e0b000  rw-     0x3f000 8       1       913982  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libsepol.so.1    Disabled
<SNIP>
```

Adding the `--dump` option will save out these areas to disk. 

```
python vol.py -f linux-sample-1.dmp linux.proc --dump
Volatility 3 Framework 2.4.1
PID     Process Start   End     Flags   PgOff   Major   Minor   Inode   File Path       File output

1       init    0x400000        0x409000        r-x     0x0     8       1       652843  /sbin/sbin/init pid.1.vma.0x400000-0x409000.dmp
1       init    0x608000        0x609000        rw-     0x8000  8       1       652843  /sbin/sbin/init pid.1.vma.0x608000-0x609000.dmp
1       init    0x23fb000       0x241c000       rw-     0x0     0       0       0       [heap]  pid.1.vma.0x23fb000-0x241c000.dmp
1       init    0x7f489741c000  0x7f489741e000  r-x     0x0     8       1       913938  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libdl-2.13.so    pid.1.vma.0x7f489741c000-0x7f489741e000.dmp
1       init    0x7f489741e000  0x7f489761e000  ---     0x2000  8       1       913938  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libdl-2.13.so    pid.1.vma.0x7f489741e000-0x7f489761e000.dmp
1       init    0x7f489761e000  0x7f489761f000  r--     0x2000  8       1       913938  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libdl-2.13.so    pid.1.vma.0x7f489761e000-0x7f489761f000.dmp
1       init    0x7f489761f000  0x7f4897620000  rw-     0x3000  8       1       913938  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libdl-2.13.so    pid.1.vma.0x7f489761f000-0x7f4897620000.dmp
1       init    0x7f4897620000  0x7f48977a2000  r-x     0x0     8       1       913923  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libc-2.13.so     pid.1.vma.0x7f4897620000-0x7f48977a2000.dmp
1       init    0x7f48977a2000  0x7f48979a1000  ---     0x182000        8       1       913923  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libc-2.13.so     pid.1.vma.0x7f48977a2000-0x7f48979a1000.dmp
1       init    0x7f48979a1000  0x7f48979a5000  r--     0x181000        8       1       913923  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libc-2.13.so     pid.1.vma.0x7f48979a1000-0x7f48979a5000.dmp
1       init    0x7f48979a5000  0x7f48979a6000  rw-     0x185000        8       1       913923  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libc-2.13.so     pid.1.vma.0x7f48979a5000-0x7f48979a6000.dmp
1       init    0x7f48979a6000  0x7f48979ab000  rw-     0x0     0       0       0       Anonymous Mapping       pid.1.vma.0x7f48979a6000-0x7f48979ab000.dmp
1       init    0x7f48979ab000  0x7f48979c9000  r-x     0x0     8       1       913958  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libselinux.so.1  pid.1.vma.0x7f48979ab000-0x7f48979c9000.dmp
1       init    0x7f48979c9000  0x7f4897bc8000  ---     0x1e000 8       1       913958  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libselinux.so.1  pid.1.vma.0x7f48979c9000-0x7f4897bc8000.dmp
1       init    0x7f4897bc8000  0x7f4897bc9000  r--     0x1d000 8       1       913958  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libselinux.so.1  pid.1.vma.0x7f4897bc8000-0x7f4897bc9000.dmp
<SNIP>
```

This can still be filtered to pid as normal:

```
python .\vol.py -f .\linux-sample-1.dmp linux.proc --pid 8600 --dump
Volatility 3 Framework 2.4.1
PID     Process Start   End     Flags   PgOff   Major   Minor   Inode   File Path       File output

8600    bash    0x400000        0x4e5000        r-x     0x0     8       1       391688  /bin/bin/bash   pid.8600.vma.0x400000-0x4e5000.dmp
8600    bash    0x6e4000        0x6e5000        r--     0xe4000 8       1       391688  /bin/bin/bash   pid.8600.vma.0x6e4000-0x6e5000.dmp
8600    bash    0x6e5000        0x6ee000        rw-     0xe5000 8       1       391688  /bin/bin/bash   pid.8600.vma.0x6e5000-0x6ee000.dmp
8600    bash    0x6ee000        0x6f4000        rw-     0x0     0       0       0       Anonymous Mapping       pid.8600.vma.0x6ee000-0x6f4000.dmp
8600    bash    0x20cd000       0x210e000       rw-     0x0     0       0       0       [heap]  pid.8600.vma.0x20cd000-0x210e000.dmp
8600    bash    0x7fe78a85f000  0x7fe78a86a000  r-x     0x0     8       1       913940  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libnss_files-2.13.so     pid.8600.vma.0x7fe78a85f000-0x7fe78a86a000.dmp
8600    bash    0x7fe78a86a000  0x7fe78aa69000  ---     0xb000  8       1       913940  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libnss_files-2.13.so     pid.8600.vma.0x7fe78a86a000-0x7fe78aa69000.dmp
8600    bash    0x7fe78aa69000  0x7fe78aa6a000  r--     0xa000  8       1       913940  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libnss_files-2.13.so     pid.8600.vma.0x7fe78aa69000-0x7fe78aa6a000.dmp
8600    bash    0x7fe78aa6a000  0x7fe78aa6b000  rw-     0xb000  8       1       913940  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libnss_files-2.13.so     pid.8600.vma.0x7fe78aa6a000-0x7fe78aa6b000.dmp
8600    bash    0x7fe78aa6b000  0x7fe78aa75000  r-x     0x0     8       1       913936  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libnss_nis-2.13.so       pid.8600.vma.0x7fe78aa6b000-0x7fe78aa75000.dmp
8600    bash    0x7fe78aa75000  0x7fe78ac74000  ---     0xa000  8       1       913936  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libnss_nis-2.13.so       pid.8600.vma.0x7fe78aa75000-0x7fe78ac74000.dmp
8600    bash    0x7fe78ac74000  0x7fe78ac75000  r--     0x9000  8       1       913936  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libnss_nis-2.13.so       pid.8600.vma.0x7fe78ac74000-0x7fe78ac75000.dmp
8600    bash    0x7fe78ac75000  0x7fe78ac76000  rw-     0xa000  8       1       913936  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libnss_nis-2.13.so       pid.8600.vma.0x7fe78ac75000-0x7fe78ac76000.dmp
<snip>
```

It can also be filtered to particular regions with the --address option: 

(Note: address are limited to ints, I would have liked to accept 0x400000 or 4194304 but I think this is probably better handled in requirements rather than the plugin itself?)
```
python vol.py -f linux-sample-1.dmp linux.proc --pid 8600 --dump --address 4194304
Volatility 3 Framework 2.4.1
PID     Process Start   End     Flags   PgOff   Major   Minor   Inode   File Path       File output

8600    bash    0x400000        0x4e5000        r-x     0x0     8       1       391688  /bin/bin/bash   pid.8600.vma.0x400000-0x4e5000.dmp
```

A list of address can be supplied too:
```
python vol.py -f linux-sample-1.dmp linux.proc --pid 8600 --dump --address 4194304, 34394112
Volatility 3 Framework 2.4.1
PID     Process Start   End     Flags   PgOff   Major   Minor   Inode   File Path       File output

8600    bash    0x400000        0x4e5000        r-x     0x0     8       1       391688  /bin/bin/bash   pid.8600.vma.0x400000-0x4e5000.dmp
8600    bash    0x20cd000       0x210e000       rw-     0x0     0       0       0       [heap]  pid.8600.vma.0x20cd000-0x210e000.dmp
```


volatility3/framework/plugins/linux/proc.py
```diff
@@ -35,16 +40,138 @@ def get_requirements(cls):
...
+        Args:
+            task: task object from which to list the vma
+            filter_func: Function to take a vma and return True if it should be filtered out
+
+        Returns:
+            A list of vmas based on the task and filtered based on the filter function
+        """
+        if task.mm:
+            for vma in task.mm.get_mmap_iter():
+                if not filter_func(vma):
```
ikelos: The `filter_func` usually returns true to include the item in the list.  I know this seems contrary to the name filter, which suggests removing items, but it's in keeping with the [python filter function](https://docs.python.org/3/library/functions.html#filter) (and the rest of the code base).  Please flip this and update the documentation.

volatility3/framework/plugins/linux/proc.py
```diff
@@ -35,16 +40,138 @@ def get_requirements(cls):
...
+                    yield vma
+
+    @classmethod
+    def vma_dump(
+        cls,
+        context: interfaces.context.ContextInterface,
+        task: interfaces.objects.ObjectInterface,
+        vma: interfaces.objects.ObjectInterface,
```
ikelos: I'd prefer this weren't passed as a complex type, but as simple integers.  I'm ok with one or two complex parameters where it makes sense (such as task), but this is literally just holding two numbers, and if someone wants to call this without having gone the route you expected, they'll need to construct complex objects just to pass in two numbers.  That should also shift where the AttributeError check occurs.

volatility3/framework/plugins/linux/proc.py
```diff
@@ -35,16 +40,138 @@ def get_requirements(cls):
...
+            vm_start = vma.vm_start
+            vm_end = vma.vm_end
+        except AttributeError:
+            vollog.debug("Unable to find the vm_start and vm_end")
+            return None
+        
+        vm_size = vm_end - vm_start
+        if 0 < maxsize < vm_size:
+            vollog.debug(
+                f"Skip virtual memory dump {vm_start:#x}-{vm_end:#x} due to maxsize limit"
```
ikelos: Should this skip, or return up to maxsize?  I don't mind either, I'd just like it clear so as not to confuse people why things didn't return (particularly since the warning requires debug verbosity to have been turned on to see it).

volatility3/framework/plugins/linux/proc.py
```diff
@@ -35,16 +40,138 @@ def get_requirements(cls):
...
+            )
+            return None
+
+        pid = "Unknown"
```
ikelos: This is somewhat messing with the type of pid to try and get a nice error message.  I'm not sure why, but I'd prefer pid to be None and then checked, since None is a known singleton that can easily be checked elsewhere should it be needed.

volatility3/framework/plugins/linux/proc.py
```diff
@@ -35,16 +40,138 @@ def get_requirements(cls):
...
+        proc_layer = context.layers[proc_layer_name]
+        file_name = f"pid.{pid}.vma.{vm_start:#x}-{vm_end:#x}.dmp"
+        try:
+            file_handle = open_method(file_name)
+            chunk_size = 1024 * 1024 * 10
+            offset = vm_start
+            while offset < vm_start + vm_size:
+                to_read = min(chunk_size, vm_start + vm_size - offset)
+                data = proc_layer.read(offset, to_read, pad=True)
+                if not data:
```
ikelos: This would only happen if to_read were 0 presumably, since we're padding the output?  I'd prefer to check that before hand or remove the check if it's unnecessary, this just sets my spidey sense tingling for some reason...  5;P

volatility3/framework/plugins/linux/proc.py
```diff
@@ -35,16 +40,138 @@ def get_requirements(cls):
...
     def _generator(self, tasks):
+        # build filter for addresses if required
+        address_list = self.config.get("address", [])
+        if address_list == []:
+            # do not filter as no address_list was supplied
+            filter_func = lambda _: False 
+        else:
+            # filter for any vm_start that matches the supplied address config
+            def filter_function(x: interfaces.objects.ObjectInterface) -> bool:
+                return x.vm_start not in address_list
```
ikelos: I think this would also need changing to flip the filter function.

volatility3/framework/plugins/linux/proc.py
```diff
@@ -35,16 +40,138 @@ def get_requirements(cls):
...
             ),
+            requirements.BooleanRequirement(
+                name="dump",
+                description="Extract listed memory segments",
+                default=False,
+                optional=True,
+            ),
+            requirements.ListRequirement(
+                name="address",
+                description="Process virtual memory address to include "
```
ikelos: I'm a little nervous about this.  The description needs to be extremely precise so that no one gets confused.  It should be clear whether this value applies to all output or just dumping, and it should be clear whether we're discussing addresses or ranges (since the name is address, but it then after says all other ranges are excluded).  It sounds as though this is a single address, but base address isn't clear whether it's only the base address of the module or the base (starting) address of a block?  Also, it feels as though there could be some clever code in the plugin that *does* allow any address in a block to be used, which might be handier for people who want to extract just a chunk they're interested in, without having to do extra work to figure out the starting address of the segment they're interested in?

volatility3/framework/plugins/linux/proc.py
ikelos: This would also be where the change to allow segment-internal-addresses to act as filters would be.  A little more complex, but should be doable.

volatility3/framework/plugins/linux/proc.py
```diff
@@ -35,16 +40,138 @@ def get_requirements(cls):
...
+        except Exception as excp:
+            vollog.debug(f"Unable to dump virtual memory {file_name}: {excp}")
+            return None
+
+        return file_handle
+
     def _generator(self, tasks):
+        # build filter for addresses if required
+        address_list = self.config.get("address", [])
+        if address_list == []:
```
ikelos: This is a little ugly, and I'd prefer the default were to be `None` and then we can definitively checked for `None` (or even `not address_list`) just in case the flag was provided but no items given somehow.

ikelos (commented):
Mostly looks good, a couple of comments (filter functions are generally inverted from how the code here works), but pretty good.  All the things I'm finding are pretty big design decisions, rather than little stylistic gotchas.  Keep up the good work!  5:)

This too will require @atcuno 's masterful eye before it can go in...  5;)

volatility3/framework/plugins/linux/proc.py
eve-mem: I think I know how I'd add the any address option. I'll update when that is done.

volatility3/framework/plugins/linux/proc.py
eve-mem: I think that's because when a list option is provided it always returns an empty list even if the option isn't provided. Given your comment maybe that's not intended? 

e.g. this still returns an empty list
```
print(self.config["address"], type(self.config["address"]))
address_list = self.config.get("address", None)
print(address_list, type(address_list))
```

Here are the different case:
```
python vol.py -f linux-sample-1.dmp  linux.proc
Volatility 3 Framework 2.4.1
PID     Process Start   End     Flags   PgOff   Major   Minor   Inode   File Path       File output
[] <class 'list'>
[] <class 'list'>
<snip>

python vol.py -f linux-sample-1.dmp  linux.proc --address
Volatility 3 Framework 2.4.1
PID     Process Start   End     Flags   PgOff   Major   Minor   Inode   File Path       File output
[] <class 'list'>
[] <class 'list'>
<snip>

python vol.py -f linux-sample-1.dmp  linux.proc --address 1, 2, 3
Volatility 3 Framework 2.4.1
PID     Process Start   End     Flags   PgOff   Major   Minor   Inode   File Path       File output
[1, 2, 3] <class 'list'>
[1, 2, 3] <class 'list'>
<snip>
```

I tried looking into why that happens, but got lost in requirements and configuration interfaces.

I will use `not address_list` for the moment.

volatility3/framework/plugins/linux/proc.py
eve-mem: I can't work out a good way to not pass vma as an object. I could pass the start address for a vma and then go through `task.mm.get_mmap_iter()` again to find it. or is there some way of accessing the right one without needing to go through the loop? I must be missing something sorry.

volatility3/framework/plugins/linux/proc.py
eve-mem: Yup - completely unnecessary *faceaplm.gif* - good spot!

eve-mem:
I think I've covered most of the comments apart from the not passing the vma as an object. I'm not sure what to do on that one. Thanks!

eve-mem:
Updated again with an attempt to not pass the vma object, just the `vm_start` and `vm_end` as ints now.

cstation:
Closes #830

volatility3/framework/plugins/linux/proc.py
ikelos: Sorry, that's what I meant, `not address_list` catches both an empty list and `None`, seems better than specifically checking equality against the empty list, thanks.  5:)

volatility3/framework/plugins/linux/proc.py
```diff
@@ -35,16 +41,144 @@ def get_requirements(cls):
...
+                    vollog.debug(
+                        f"Excluded vma at offset {vma.vol.offset:#x} for pid {task.pid} due to filter_func"
+                    )
+
+    @classmethod
+    def vma_dump(
+        cls,
+        context: interfaces.context.ContextInterface,
+        task: interfaces.objects.ObjectInterface,
+        vm_start: int,
```
ikelos, line 103: Thanks, I know the method signature is a bit longer, but it's much easier for arbitrary callers of the classmethod to do so now.

volatility3/framework/plugins/linux/proc.py
ikelos: Looks like you figured out what I was after.  5;)

volatility3/framework/plugins/linux/proc.py
eve-mem: I was actually thinking of a making a generic write to file function as a utility somewhere. 

You either pass it a task object and it works off a new layer for that task, or you pass a layer name and it outputs that. 

Sending an addr and a length so it matches the layer read functions. 

What do you think? Not sure where exactly to put it if I did make it, Linux utilities?

ikelos (commented):
Looks good to me.  I'd still prefer @atcuno to give it a once over ideally but from a python perspective it look good.

volatility3/framework/plugins/linux/proc.py
ikelos: I'm not sure how common a task that would be?  If we only really do it in one plugin, then it might as well live in that plugin, but as a classmethod so other plugins can use it.  I think I'd prefer it to be in a layer rather than a task (since that's slightly more generic) and accept a list of tuples of ranges that should be concatenated together from the layer?  That could then live in just about any plugin at that point, but again, I just don't know how common it is to need that functionality...

eve-mem: That makes sense a lot of sense. I was thinking about it for adding --dump to the pslist plugin but looks like @cstation has that covered. 

I'll leave it for now.

volatility3/framework/plugins/linux/proc.py
```diff
@@ -35,16 +41,144 @@ def get_requirements(cls):
...
+        except exceptions.InvalidAddressException as excp:
+            vollog.debug(
+                "Process {}: invalid address {} in layer {}".format(
+                    pid, excp.invalid_address, excp.layer_name
+                )
+            )
+            return None
+
+        vm_size = vm_end - vm_start
+        if 0 < maxsize < vm_size:
```
atcuno: this check seems complicated and I am not sure why comparing "0 < maxsize" is needed? Could it just be if vm_size >= maxize?

eve-mem: Yup... I can't remember why I thought that would have been better. I'll make it more straight forward.

eve-mem: I've updated this now. I remember my original plan was to verify that the size of the area wasn't negative. That should never happen, but I wanted to check for that case. 

I've now made that check obvious and provide a better error in when that does happen.

volatility3/framework/plugins/linux/proc.py
```diff
@@ -35,16 +41,145 @@ def get_requirements(cls):
...
+        """Lists the Virtual Memory Areas of a specific process.
+
+        Args:
+            task: task object from which to list the vma
+            filter_func: Function to take a vma and return False if it should be filtered out
+
+        Returns:
+            Yields vmas based on the task and filtered based on the filter function
+        """
+        if task.mm:
```
ikelos, line 85: If there is no task.mm (or it throws an exception) we probably want to tell the user that we failed, and why?

ikelos (commented):
Ok, looks pretty good.  Just one more log message and then I think it's good to merge!  5:)

eve-mem:
No mm actually happens quite a lot, every kernel thread for example. I think a debug message wouldn't hurt, but I think a warning would get annoying very quickly. 

I'll add that shortly!

volatility3/framework/plugins/linux/proc.py
ikelos: Ah, my mistake, I thought this was the main generator and so would only ever through once.  Yep, I agree a debug would be better...  5:)

eve-mem: I've added this in now.

eve-mem:
I've added an extra debug message, fixed an issue I spotted with a different debug message, and also added a version number as this plugin has some of useful functions that I think will be useful for other plugins.

eve-mem:
Hello @ikelos and @atcuno - Is there anything else that I should be working on for this pull request? I think I covered everything off, but please feel free to correct me!

atcuno:
Assuming the merge conflict is resolved, this is ready for merge @ikelos

ikelos:
Ok, great.  @eve-mem if you could rebase this, then we should be good to merge it.  5;)

eve-mem:
Awesome, I'll get that done. Thanks.

eve-mem:
I think I've resolved that merge conflict correctly now! Certainly a learning experience for me with git! :D

ikelos (commented):
Looks good, thanks!  5:D

PR #885, opened 2022-12-13
add linux.vmayarascan

eve-mem:
This adds `linux.vmayarascan` plugin which is heavily based on the `windows.vadyarascan` plugin, only swapping windows references to linux and changing from processing vads to vmas.

Supports the same flags as the windows version. 

Example output:
```
python vol.py -f linux-sample-3.dmp linux.vmayarascan --yara-file .\rules.yar
Volatility 3 Framework 2.4.1
Offset  PID     Rule    Component       Value

0x7f4897e27cc0  1       finder  $http   68 74 74 70 3a 2f 2f 77 77 77 2e 64 65 62 69 61 6e 2e 6f 72 67
0x1a4698b       322     finder  $http   68 74 74 70 3a 2f 2f 77 77 77 2e 6c 65 6e 6f 76 6f 2e 63 6f 6d
0x1a46a19       322     finder  $http   68 74 74 70 3a 2f 2f 62 78 69 6e 66 6f 2e 74 6f 73 68 69 62 61 2e 63 6f 6d
0x1a508a1       322     finder  $http   68 74 74 70 3a 2f 2f 77 77 77 2e 63 6f 6d 70 75 74 65 72 73 2e 75 73
0x1a50921       322     finder  $http   68 74 74 70 3a 2f 2f 77 77 77 2e 67 61 74 65 77 61 79 2e 63 6f 6d
0x1a509ea       322     finder  $http   68 74 74 70 3a 2f 2f 77 77 77 2e 6c 65 6e 6f 76 6f 2e 63 6f 6d
0x1a50a78       322     finder  $http   68 74 74 70 3a 2f 2f 62 78 69 6e 66 6f 2e 74 6f 73 68 69 62 61 2e 63 6f 6d
0x1a507fe       322     finder  $https  68 74 74 70 73 3a 2f 2f 77 77 77 2e 64 65 6c 6c 62 61 74 74 65 72 79 70 72 6f 67 72 61 6d 2e 63 6f 6d
0x1a4698b       468     finder  $http   68 74 74 70 3a 2f 2f 77 77 77 2e 6c 65 6e 6f 76 6f 2e 63 6f 6d
0x1a46a19       468     finder  $http   68 74 74 70 3a 2f 2f 62 78 69 6e 66 6f 2e 74 6f 73 68 69 62 61 2e 63 6f 6d
0x1a508a1       468     finder  $http   68 74 74 70 3a 2f 2f 77 77 77 2e 63 6f 6d 70 75 74 65 72 73 2e 75 73
0x1a50921       468     finder  $http   68 74 74 70 3a 2f 2f 77 77 77 2e 67 61 74 65 77 61 79 2e 63 6f 6d
0x1a509ea       468     finder  $http   68 74 74 70 3a 2f 2f 77 77 77 2e 6c 65 6e 6f 76 6f 2e 63 6f 6d
0x1a50a78       468     finder  $http   68 74 74 70 3a 2f 2f 62 78 69 6e 66 6f 2e 74 6f 73 68 69 62 61 2e 63 6f 6d
0x1a507fe       468     finder  $https  68 74 74 70 73 3a 2f 2f 77 77 77 2e 64 65 6c 6c 62 61 74 74 65 72 79 70 72 6f 67 72 61 6d 2e 63 6f 6d
<snip>
```


ikelos:
It's getting a bit late for me, so I'm not going to review this one now, but I haven't forgotten about it...  5;)

volatility3/framework/plugins/linux/vmayarascan.py
```diff
@@ -0,0 +1,121 @@
...
+            requirements.ListRequirement(
+                name="pid",
+                element_type=int,
+                description="Process IDs to include (all other processes are excluded)",
+                optional=True,
+            ),
+        ]
+
+    def _generator(self):
+        kernel = self.context.modules[self.config["kernel"]]
```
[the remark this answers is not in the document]
eve-mem: Fixed this one now.

eve-mem:
Hello @ikelos just wondering if you'll be able to take a look over this one. I'd really appreciate your feedback 🙂 thank you.

volatility3/framework/plugins/linux/vmayarascan.py
```diff
@@ -0,0 +1,120 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
```
ikelos: I think we suggest putting the year the code is written into the copyright, so this would presumably 2023?

volatility3/framework/plugins/linux/vmayarascan.py
```diff
@@ -0,0 +1,120 @@
...
+from volatility3.framework import interfaces, renderers
+from volatility3.framework.configuration import requirements
+from volatility3.framework.renderers import format_hints
+from volatility3.plugins import yarascan
+from volatility3.plugins.linux import pslist
+
+
+class VmaYaraScan(interfaces.plugins.PluginInterface):
+    """Scans all virtual memory areas for tasks using yara."""
+
```
ikelos, line 16: Since this has a staticmethod (ie, one that can be called by other parts of volatility and potentially other plugins), it's best to give this an `_version` number, so that if the API needs changing/updating down the line it can be done easily and spotting breakages in the dependency chain will be easy.  5:)

volatility3/framework/plugins/linux/vmayarascan.py
```diff
@@ -0,0 +1,120 @@
...
+            requirements.ListRequirement(
+                name="pid",
+                element_type=int,
+                description="Process IDs to include (all other processes are excluded)",
+                optional=True,
+            ),
+        ]
+
+    def _generator(self):
+        rules = yarascan.YaraScan.process_yara_options(dict(self.config))
```
ikelos, line 54: If this is processing rules from YaraScan, perhaps it makes sense to use the same requirements as YaraScan, so that changes there get reflected here automatically?  That wouldn't work if there's options in the original that you *don't* want in vmayarascan, but otherwise it feels like it might reduce duplication slightly.

ikelos (commented):
Yep, this looks good!  A couple questions around trying to share as much as possible between plugins, but otherwise looks good.  5:)

volatility3/framework/plugins/linux/vmayarascan.py
eve-mem: Yes, I do want to use the same requirements as YaraScan, just add in an option for a pid too. I'll look into this now.

eve-mem:
@ikelos I've updated based on those comments. 

I've updated to use the requirements directly from the generic yarascan plugin. However I had to remove the TranslationLayerRequirement from the base plugin or the automagic would only find either the TranslationLayerRequirement from yarascan OR the ModuleRequirement from vmayarascan. It would just fill the first one from the list. 

How do you feel about the way I've done that?

eve-mem:
@ikelos what do you think of the way I've handled the automagic issue I was having - is there a better way to do that?

eve-mem:
@ikelos - What do you think of the way I handled the requirements for this? Any comments you have would be greatly appreciated.

ikelos:
Hiya, sorry this slipped off my radar.  Hmmmm, ideally the automagic would be fixed up to find both, but since we'd then technically be asking for two translation layers, the current behaviour is probably correct.  The only other way I can think of dealing with it, is just to look for a symbol requirement (which at the moment gets populated with kernel symbols when requested), so essentially revert to the old way of using requirements.  However you then don't get the module at the end of it, and you don't necessarily get things like the kernel offset anywhere.

I'd also be happy with you splitting the requirements out into two sets, so that the flag parameters are publicly accessible on the yarascan plugin (with a minor version bump for an additive change) and then just adding those to the module requirement here.  Essentially it would result in the same thing, it just feels like building an official division might be a useful way of handling the situation for any other plugins where we get similar reusability issues?  Lemme know what you think, I'm ok with this solution and I'd be ok with change the interface for yarascan, and that's currently the one I'd lean towards, but keen to get your view on it...  5:)

eve-mem:
@ikelos - I think your suggestion about splitting the requirements makes a lot sense. Fix it at the source once rather than over and over again in different plugins. I will have a go at making that work. Thanks again for the advice.

eve-mem:
@ikelos - I've made the change so that the standard yarascan plugin to then allow the new vmayarascan plugin to use its command line options. 

I don't think that `get_yarascan_option_requirements()`  is the best name for the function the returns just the command line options for yarascan, but I couldn't  come up with anything better. 

If you think this is a good way to do it I can also update the windows vadyarascan plugin. That currently maintains it's own list of yarascan rule parsing options. Probably a different PR for that though I suspect.

thomasdangl:
Any update on this one?

eve-mem:
Hello @thomasdangl - just currently waiting on review from the core devs. This makes some changes to some of the base yara scanning parts, so understandably it takes more time to think about. 

If you wanted to give it a test drive you can clone this branch and try it out. (do let me know if you find any bugs!)

eve-mem:
Hey @ikelos - sending you a prod about this PR... sorry!

I was thinking of doing a similar thing with layer requirements for a generic strings plugin (to then be used in window/linux/etc) but if this way of doing it is a bad idea i don't want to go down that path. 

What do you think?

volatility3/framework/plugins/yarascan.py
```diff
@@ -61,19 +61,31 @@ class YaraScan(plugins.PluginInterface):
...
 
     # TODO: When the major version is bumped, take the opportunity to rename the yara_rules config to yara_string
     # or something that makes more sense
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
-        return [
+        """Returns the requirements needed to run yarascan directly, combining the TranslationLayerRequirement
+        and the requirements from get_yarascan_option_requirements."""
+        return cls.get_yarascan_option_requirements() + [
```
ikelos, line 73: Does this handling affect the `windows.vadyarascan` plugin at all and/or should `vadyarascan` be updated to take a `ModuleRequirement` rather than a `TranslationLayerRequirement`?

ikelos (commented):
Sorry about the delay in reviewing again, just need to think about how this might affect `windows.vadyarascan`

volatility3/framework/plugins/yarascan.py
eve-mem: Great question! So currently `windows.vadyarascan` doesn't actually use the `yarascan` plugin at all. The windows one maintains it's own list of requirements to do the command line options, it does use the `yarascanner` scanner but not the plugin. 

If you think that this way of allowing the linux plugin to use the base config options from the `yarascan`  plugin  isn't bad, then I would also look to update the windows to work in the same way. That way if you want to update the `yarascan` command line options there is only one place to do it.

ikelos: Yep, that sounds reasonable.  Could you include the windows yarascan changes in here, or would you like to push this in before doing that as a second pull?

ikelos (approved):
All looks good, now, just waiting to hear whether the windows changes associated with this will be included here or in a new PR...

eve-mem:
Thank you @ikelos. Is it okay to keep this as just the linux side and then I'll do the windows one when I get the time.

I've made this issue https://github.com/volatilityfoundation/volatility3/issues/1047 to track that - very happy if you would assign that to me. (equally very happy for anyone to jump in and work on that too)

ikelos:
Sure, no problem.  5:)

PR #886, opened 2022-12-14
Increase speed of windows.handles and NameInfo

paulkermann:
The main problem was that a module was created for each process and for every object with `NameInfo`
Creating modules is **extremely** slow and by not creating redundant duplicate modules we could use the same one if it already exists.


ikelos (commented):
Ok, so I agree that we're massively inefficient at the moment, and I think I know how it came to be.  We decided to sub out anywhere that was making use of the kernel with a lookup in the modules table, and those that couldn't constructed the kernel themselves.

Looking at this code though, it really doesn't need the full kernel object.  The only reason it's using it is do some maths on the address at best, and we then add the kvo to the address anyway?  I think the way of resolving this is to just lookup the layer's kvo (as we do) and lookup the symbol using the symbol table name.  This will be a module-relative offset (I believe) and we then add the kvo to it.  The whole module thing was to try to do stuff "the right way" in case aslr or anything else shifted the address, rather than having every plugin do their own ASLR.  As I said though, we seem to add the kvo anyway, so it really doesn't change a lot.

So, happy with most of your changes, but not keen on the additional mini-functions and I think we can do without them totally by just asking the context's symbol table (from `self.symbol_table_name`) for the offset (and if necessary doing the relative to absolute ourselves).

paulkermann:
@ikelos are there any more changes you think should be in this PR?

ikelos:
Sorry this has taken me so long to review.  There are couple of bits that were still a concern (not checking whether the free name was actually free or not) so I've been a new PR that trying to pull the module from the context, to try and speed things up.  Please could you test #920 and let me know whether there's anything else needed from this patch?

ikelos:
This should now be largely aleviated by #920, please file a new PR if you feel there are still outstanding optimizations that can be made.

PR #892, opened 2022-12-29
add linux.bash_hash and linux.bash_env plugins

a5hlynx:
Hello

I've just added 2 plugins, linux.bash_hash and linux.bash_env.
Below are the usage examples.

```
$ python3 vol.py -f ../../mem/ubuntu64.lime linux.bash_hash
Volatility 3 Framework 2.4.1
PID     Name    Hits    Command Full Path

1416    bash    1       sudo    /usr/bin/sudo
1416    bash    1       ls      /usr/bin/ls
1476    bash    1       insmod  /usr/sbin/insmod
1476    bash    1       rmmod   /usr/sbin/rmmod
1476    bash    2       swapoff /usr/sbin/swapoff
1476    bash    3       swapon  /usr/sbin/swapon
```

```
$ python3 vol.py -f ../../mem/ubuntu64.lime linux.bash_env
Volatility 3 Framework 2.4.1
PID     Process Vars

1       systemd
2       kthreadd
3       rcu_gp
4       rcu_par_gp
5       kworker/0:0
6       kworker/0:0H
7       kworker/0:1
8       kworker/u256:0
9       mm_percpu_wq
10      ksoftirqd/0
11      rcu_sched
12      migration/0
13      idle_inject/0
14      cpuhp/0

..snip..

```

The former was developed based on volatility2's linux_bash_hash plugin and has the same functionality, and the latter is based on linux_bash_env. So these are more of re-implementation for volatility3 rather than new-development.

I am not quite sure if pull-request is an appropreate way to request for plugin addition like this, but if there is no problem about this, would you add these  please?

### DIFF
--- a/volatility3/framework/plugins/linux/bash_env.py
+++ b/volatility3/framework/plugins/linux/bash_env.py
@@ -1,0 +1,229 @@
+# This file is Copyright 2022 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import struct
+from typing import List
+
+from volatility3.framework import constants, renderers, symbols, exceptions, interfaces
+from volatility3.framework.configuration import requirements
+from volatility3.framework.interfaces import plugins
+from volatility3.framework.objects import utility
+from volatility3.framework.symbols.linux.hash import HashIntermedSymbols
+from volatility3.plugins.linux import pslist
+
+class Var(plugins.PluginInterface):
+    """Recovers a process' dynamic environment variables."""
+    shells = ("/bin/bash", "/bin/dash", "/bin/sh")
+
+    _required_framework_version = (2, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Linux kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.PluginRequirement(
+                name="pslist", plugin=pslist.PsList, version=(2, 0, 0)
+            ),
+            requirements.ListRequirement(
+                name="pid",
+                element_type=int,
+                description="Process IDs to include (all other processes are excluded)",
+                optional=True,
+            ),
+        ]
+
+    def _generator(self, tasks):
+        for task in tasks:
+            proc_layer_name = task.add_process_layer()
+            if not proc_layer_name:
+                yield (0, (task.pid, utility.array_to_string(task.comm), ""))
+                continue
+            varstr = ""
+            for (key, val) in self.bash_environment(
+                task, proc_layer_name, self.context, self.config["kernel"]
+            ):
+                if key or val:
+                    varstr = f"{varstr}{key}={val} "
+
+            yield (0, (task.pid, utility.array_to_string(task.comm), varstr))
+
+    def bash_environment(self, task, proc_layer_name, context, module):
+        for (key, val) in self.dynamic_env(task, proc_layer_name, context, module):
+            yield (key, val)
+
+        for (key, val) in self.shell_variables(
+            task, proc_layer_name, context, module
+        ):
+            yield (key, val)
+
+    def dynamic_env(self, task, proc_layer_name, context, module):
+        symbol_table_name = context.modules[module].symbol_table_name
+        is_32bit = not symbols.symbol_table_is_64bit(
+            context, symbol_table_name
+        )
+        addr_sz = 8
+        pack_format = "Q"
+        if is_32bit:
+            addr_sz = 4
+            pack_format = "I"
+        proc_layer = context.layers[proc_layer_name]
+        addr_cache = {0}
+        for vma in task.mm.get_mmap_iter():
+            if not (vma.vm_file and vma.get_protection() == "rw-"):
+                continue
+            fname = vma.get_name(context, task)
+            if fname.find("ld") == -1 and (
+                not fname.endswith(self.shells)
+            ):
+                continue
+            env_start = 0
+            if not isinstance(vma.vm_start, int) or not isinstance(vma.vm_end, int):
+                continue
+            vma_start = vma.vm_start
+            vma_end = vma.vm_end
+            vma_len = vma_end - vma_start
+            vma_data = proc_layer.read(vma_start, vma_len, pad=True)
+            for off in range(0, vma_len - addr_sz, addr_sz):
+                addrstr = vma_data[off : off + addr_sz]
+                addr, = struct.unpack(pack_format, addrstr)
+
+                if addr in addr_cache:
+                    continue
+                addr_cache.add(addr)
+                if addr:
+                    firstaddrstr = proc_layer.read(addr, addr_sz, pad=True)
+                    firstaddr, = struct.unpack(pack_format, firstaddrstr)
+                    buf = proc_layer.read(firstaddr, 64, pad=True)
+                    eqidx = buf.find(b"=")
+                    if eqidx > 0:
+                        nullidx = buf.find(b"\x00")
+                        if nullidx >= eqidx:
+                            env_start = addr
+
+            if env_start == 0:
+                continue
+            envars = context.object(
+                symbol_table_name + constants.BANG + "array",
+                layer_name=proc_layer_name,
+                offset=env_start,
+                subtype=context.symbol_space.get_type(
+                    symbol_table_name + constants.BANG + "pointer"
+                ),
+                count=256,
+            )
+            for var in envars:
+                if var:
+                    sizes = [8, 16, 32, 64, 128, 256, 384, 512, 1024, 2048, 4096]
+                    good_varstr = None
+                    for size in sizes:
+                        try:
+                            varstr = proc_layer.read(var, size)
+                        except exceptions.InvalidAddressException:
+                            continue
+                        if not varstr:
+                            continue
+                        eqidx = varstr.find(b"=")
+                        idx = varstr.find(b"\x00")
+                        if idx == -1 or eqidx == -1 or idx < eqidx:
+                            continue
+                        good_varstr = varstr
+                        break
+                    if good_varstr:
+                        try:
+                            good_varstr = str(good_varstr[:idx], "utf-8")
+                            key = good_varstr[:eqidx]
+                            val = good_varstr[eqidx + 1 :]
+                            yield (key, val)
+                        except exceptions.InvalidAddressException:
+                            continue
+                        except UnicodeDecodeError:
+                            continue
+                    else:
+                        break
+
+    def shell_variables(self, task, proc_layer_name, context, module):
+        symbol_table_name = context.modules[module].symbol_table_name
+        proc_layer = context.layers[proc_layer_name]
+        is_32bit = not symbols.symbol_table_is_64bit(
+            context, symbol_table_name
+        )
+        addr_sz = 8
+        pack_format = "Q"
+        hash_json_file = "hash64"
+        if is_32bit:
+            addr_sz = 4
+            pack_format = "I"
+            hash_json_file = "hash32"
+        ptr_cache = {0}
+        bash_was_last = False
+        for vma in task.mm.get_mmap_iter():
+            try:
+                if vma.vm_file:
+                    fname = vma.get_name(context, task)
+                    bash_was_last = fname.endswith(self.shells)
+
+                if vma.vm_file or vma.get_protection() != "rw-":
+                    continue
+            except exceptions.InvalidAddressException:
+                continue
+            if bash_was_last == False:
+                continue
+            if not isinstance(vma.vm_start, int) or not isinstance(vma.vm_end, int):
+                continue
+            vma_start = vma.vm_start
+            vma_end = vma.vm_end
+            vma_len = vma_end - vma_start
+            vma_data = proc_layer.read(vma_start, vma_len, pad=True)
+            for off in range(0, vma_len - addr_sz, addr_sz):
+                ptr_test = vma_data[off : off + addr_sz]
+                ptr, = struct.unpack(pack_format, ptr_test)
+                if ptr in ptr_cache:
+                    continue
+                try:
+                    ptr_cache.add(ptr)
+                    ptr_test2 = proc_layer.read(ptr + 20, addr_sz)
+                    if not ptr_test2:
+                        continue
+                except exceptions.InvalidAddressException:
+                    continue
+                ptr2, = struct.unpack(pack_format, ptr_test2)
+                try:
+                    test = proc_layer.read(ptr2 + addr_sz, 4)
+                except exceptions.InvalidAddressException:
+                    continue
+                # this searches for bash_hash_table.nbuckets
+                if not test or test != b"\x40\x00\x00\x00":
+                    continue
+                try:
+                    hash_table_name = HashIntermedSymbols.create(
+                        context, self.config_path, "linux", hash_json_file
+                    )
+                    htable = context.object(
+                        hash_table_name + constants.BANG + "bash_hash_table",
+                        layer_name=proc_layer_name,
+                        offset=ptr2,
+                    )
+                    for ent in htable:
+                        key = utility.array_to_string(ent.key.dereference())
+                        val = utility.array_to_string(ent.data.value.dereference())
+                        yield (key, val)
+                except exceptions.InvalidAddressException:
+                    continue
+            bash_was_last = False
+
+    def run(self):
+        filter_func = pslist.PsList.create_pid_filter(self.config.get("pid", None))
+
+        return renderers.TreeGrid(
+            [("PID", int), ("Process", str), ("Vars", str)],
+            self._generator(
+                pslist.PsList.list_tasks(
+                    self.context, self.config["kernel"], filter_func=filter_func
+                )
+            ),
+        )
--- a/volatility3/framework/plugins/linux/bash_hash.py
+++ b/volatility3/framework/plugins/linux/bash_hash.py
@@ -1,0 +1,111 @@
+# This file is Copyright 2022 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+from typing import List
+
+from volatility3.framework import constants, renderers, symbols, interfaces
+from volatility3.framework.configuration import requirements
+from volatility3.framework.interfaces import plugins
+from volatility3.framework.layers import scanners
+from volatility3.framework.objects import utility
+from volatility3.framework.symbols.linux.hash import HashIntermedSymbols
+from volatility3.plugins.linux import pslist
+
+
+class Hash(plugins.PluginInterface):
+    """Recovers bash hash table from bash process memory."""
+
+    _required_framework_version = (2, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Linux kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.PluginRequirement(
+                name="pslist", plugin=pslist.PsList, version=(2, 0, 0)
+            ),
+            requirements.ListRequirement(
+                name="pid",
+                element_type=int,
+                description="Process IDs to include (all other processes are excluded)",
+                optional=True,
+            ),
+        ]
+
+    def _generator(self, tasks):
+
+        vmlinux = self.context.modules[self.config["kernel"]]
+        is_32bit = not symbols.symbol_table_is_64bit(
+            self.context, vmlinux.symbol_table_name
+        )
+        if is_32bit:
+            hash_json_file = "hash32"
+        else:
+            hash_json_file = "hash64"
+        hash_table_name = HashIntermedSymbols.create(
+            self.context, self.config_path, "linux", hash_json_file
+        )
+
+        for task in tasks:
+            task_name = utility.array_to_string(task.comm)
+            if not task_name == "bash":
+                continue
+
+            proc_layer_name = task.add_process_layer()
+            if not proc_layer_name:
+                continue
+
+            for ent in self.bash_hash_entries(task, proc_layer_name, hash_table_name):
+                yield (
+                    0,
+                    (
+                        task.pid,
+                        task_name,
+                        ent.times_found,
+                        utility.array_to_string(ent.key.dereference()),
+                        utility.array_to_string(ent.data.path.dereference()),
+                    ),
+                )
+
+    def bash_hash_entries(self, task, proc_layer_name, hash_table_name):
+
+        nbuckets_offset = self.context.symbol_space.get_type(
+            hash_table_name + constants.BANG + "bash_hash_table"
+        ).relative_child_offset("nbuckets")
+        proc_layer = self.context.layers[proc_layer_name]
+        # this searches for bash_hash_table.nbuckets
+        for off in proc_layer.scan(
+            self.context,
+            scanners.BytesScanner(b"\x40\x00\x00\x00"),
+            sections=task.get_process_memory_sections(heap_only=True),
+        ):
+            htable = self.context.object(
+                hash_table_name + constants.BANG + "bash_hash_table",
+                offset=off - nbuckets_offset,
+                layer_name=proc_layer_name,
+            )
+            for ent in htable:
+                yield ent
+
+    def run(self):
+        filter_func = pslist.PsList.create_pid_filter(self.config.get("pid", None))
+
+        return renderers.TreeGrid(
+            [
+                ("PID", int),
+                ("Name", str),
+                ("Hits", int),
+                ("Command", str),
+                ("Full Path", str),
+            ],
+            self._generator(
+                pslist.PsList.list_tasks(
+                    self.context, self.config["kernel"], filter_func=filter_func
+                )
+            ),
+        )
--- a/volatility3/framework/symbols/linux/extensions/hash.py
+++ b/volatility3/framework/symbols/linux/extensions/hash.py
@@ -1,0 +1,91 @@
+# This file is Copyright 2022 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+from typing import Dict, Tuple
+
+from volatility3.framework import exceptions, constants
+from volatility3.framework import objects, interfaces
+from volatility3.framework.objects import utility
+
+
+class bash_hash_table(objects.StructType):
+    def __init__(
+        self,
+        context: interfaces.context.ContextInterface,
+        type_name: str,
+        object_info: interfaces.objects.ObjectInformation,
+        size: int,
+        members: Dict[str, Tuple[int, interfaces.objects.Template]],
+    ) -> None:
+
+        super().__init__(
+            context=context,
+            type_name=type_name,
+            object_info=object_info,
+            size=size,
+            members=members,
+        )
+        self.context = context
+        self.layer_name = self.vol.layer_name
+        self.symbol_table_name = self.get_symbol_table_name()
+
+    def is_valid(self):
+        try:
+            bucket_array = self.bucket_array
+            nbuckets = self.nbuckets
+            nentries = self.nentries
+        except exceptions.InvalidAddressException:
+            return False
+
+        return bucket_array and nbuckets == 64 and nentries > 1
+
+    def __iter__(self):
+        if self.is_valid():
+
+            seen = {}
+            bucket_array = self.context.object(
+                self.symbol_table_name + constants.BANG + "array",
+                layer_name=self.layer_name,
+                offset=self.bucket_array,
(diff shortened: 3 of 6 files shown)


volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                count=256,
+            )
+            for var in envars:
+                if var:
+                    sizes = [8, 16, 32, 64, 128, 256, 384, 512, 1024, 2048, 4096]
+                    good_varstr = None
+                    for size in sizes:
+                        try:
+                            varstr = proc_layer.read(var, size)
+                        except (Exception, BaseException):
```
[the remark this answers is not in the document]
ikelos: This probably only needs to catch `exceptions.InvalidAddressException`.  Anything else that goes wrong we should probably pass through for something higher to handle.  Handling `BaseException` directly prevents any higher code from knowing something went wrong (and can catch too much, including some things like Ctrl+C key presses and other important system things).

The specific exceptions you expect to happen should be caught here, and everything else allowed to bubble up.  That goes for every issue found below.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                continue
+            varstr = ""
+            for (key, val) in self.bash_environment(
+                task, pack_format, hash_json_file, proc_layer_name
+            ):
+                if key != "" or val != "":
+                    varstr = varstr + key + "=" + val + " "
+            yield (0, (task.pid, utility.array_to_string(task.comm), varstr))
+
+    def bash_environment(self, task, pack_format, hash_json_file, proc_layer_name):
```
ikelos: If you want any of these methods to be run by other plugins (ie, available externally) then they should have `@classmethod` and convert `self` to `cls`.  Also an `_version` should be defined to allow people to depend on it and you to upgrade it smoothly without breaking things when you might want to change the API.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+        for task in tasks:
+            proc_layer_name = task.add_process_layer()
+            if not proc_layer_name:
+                yield (0, (task.pid, utility.array_to_string(task.comm), ""))
+                continue
+            varstr = ""
+            for (key, val) in self.bash_environment(
+                task, pack_format, hash_json_file, proc_layer_name
+            ):
+                if key != "" or val != "":
```
ikelos: It's a little more pythonic to just say `if not key or not val`, that way they could be None or the empty string or even False, and it's also slightly more human readable.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                    varstr = varstr + key + "=" + val + " "
```
ikelos: Consider using a format string (so `f"{varstr}{key}={val} "`) to reduce the number of operations python's doing.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+
+        for (key, val) in self.shell_variables(
+            task, pack_format, hash_json_file, proc_layer_name
+        ):
+            yield (key, val)
+
+    def dynamic_env(self, task, pack_format, proc_layer_name):
+        proc_layer = self.context.layers[proc_layer_name]
+        addr_cache = {0: 1}
+        addr_sz = 8
```
ikelos, line 69: This feels a little bit nasty, taking the original call (`symbols.symbol_table_is_64bit`, storing it as a pack format, and then converting it to a number at the start of the method?  I'd be much happier if you just passed in the kernel symbol table, and then it does the test and calculates whatever constants it needs then?  That'll save on any external callers (outside of this plugin) needing to do the same thing and convert the value in a pack_format just for that one parameter.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+        if pack_format == "I":
+            addr_sz = 4
+        for vma in task.mm.get_mmap_iter():
+            if not (vma.vm_file and vma.get_protection() == "rw-"):
```
ikelos, line 77: Is this situation expected, or should there be a log message in case someone needs to figure out this has happened?

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                continue
+            fname = vma.get_name(self.context, task)
+            if fname.find("ld") == -1 and (
+                not fname.endswith(("/bin/bash", "/bin/dash", "/bin/sh"))
```
ikelos: Should this be `endswith`, what sort of things might go before it and what happens if those things are long and unexpected?  Also, it feels like this list might need to change down the line, might be better to pull it out and make a class variable up at the top of the plugin somewhere?

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+            ):
+                continue
+            env_start = 0
+            vma_start = int(vma.vm_start)
```
ikelos: These shouldn't need converting to integers, please check that they can't just be used straight off.  It's a bad happy to safety cast something that's already an int, because then people copy it and it's extra code that ends up everywhere (which is how I suspect it ended up here!).  5;P

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+            vma_len = vma_end - vma_start
+            vma_data = proc_layer.read(vma_start, vma_len, pad=True)
+
+            for off in range(0, vma_len - addr_sz, addr_sz):
+                addrstr = vma_data[off : off + addr_sz]
+                addr = struct.unpack(pack_format, addrstr)[0]
+
+                if addr in addr_cache:
+                    continue
+                addr_cache[addr] = 1
```
ikelos: This looks like it can be a set, if it's only being tested for membership?  It doesn't seem as though the contents of the dictionary are necessary?

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+
+                if addr:
+                    firstaddrstr = proc_layer.read(addr, addr_sz, pad=True)
+                    if not firstaddrstr or len(firstaddrstr) != addr_sz:
```
ikelos: Since you've used `pad=True`, the len test seems superflous.  There should be no situation where the length isn't padded out to the right size (whether there is too much or too little data).

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                        continue
+                    firstaddr = struct.unpack(pack_format, firstaddrstr)[0]
```
ikelos: I slightly prefer the look of:
```
firstaddr, = stuct.unpack(pack_format, firstaddrstr)
```
over
```
firstaddr = stuct.unpack(pack_format, firstaddrstr)[0]
```
If you know that the pack_format will only ever return one value.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                    buf = proc_layer.read(firstaddr, 64, pad=True)
+                    if not buf:
```
ikelos: Since this is padded, it *might* return 64 null bytes, I don't know if they'll count as false or not, so again it will likely never trigger the continue.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                        continue
+                    eqidx = buf.find(b"=")
+                    if eqidx > 0:
+                        nullidx = buf.find(b"\x00")
+                        if nullidx >= eqidx:
+                            env_start = addr
+
+            if env_start == 0:
+                continue
+            vmlinux = self.context.modules[self.config["kernel"]]
```
ikelos: To make this an externally callable function, the kernel module name should be passed as a parameter and looked up in the context (also passed as a parameter).

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+            envars = self.context.object(
```
ikelos: Nice, big fan of reusing object parsing rather than trying to do it manually, thank you!  5:D

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                layer_name=proc_layer_name,
+                offset=env_start,
+                subtype=self.context.symbol_space.get_type(
+                    vmlinux.symbol_table_name + constants.BANG + "pointer"
+                ),
+                count=256,
+            )
+            for var in envars:
+                if var:
+                    sizes = [8, 16, 32, 64, 128, 256, 384, 512, 1024, 2048, 4096]
```
ikelos, line 121: This feels slight inefficient?  I'd think you'd maybe read a whole page and then just chop it up based on where the first null occurs?  There's also the issue in this implementation of a string (somehow, maybe bad data) ending up with a null character before an equals sign (so `b"something\x00somethingelse=athirdthing\x00"`).  Doing the check for equals after you've made the cut at `\x00` seems like it would be more robust and be easier to error check.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                        idx = varstr.find(b"\x00")
+                        if idx == -1 or eqidx == -1 or idx < eqidx:
+                            continue
+                        good_varstr = varstr
+                        break
+                    if good_varstr:
+                        try:
+                            good_varstr = str(good_varstr[:idx], "utf-8")
+                            key = good_varstr[:eqidx]
+                            val = good_varstr[eqidx + 1 :]
```
ikelos, line 140: If eqidx is -1, do both the key and val become the entire string?

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                            yield (key, val)
+                        except (Exception, BaseException):
+                            continue
+                    else:
+                        break
+
+    def shell_variables(self, task, pack_format, hash_json_file, proc_layer_name):
+        proc_layer = self.context.layers[proc_layer_name]
+        addr_sz = 8
```
ikelos, line 155: Same as above.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+        if pack_format == "I":
+            addr_sz = 4
+        ptr_cache = {0: 1}
```
ikelos: Again, use a set and test for membership.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+        bash_was_last = False
+
+        for vma in task.mm.get_mmap_iter():
+            try:
+                if vma.vm_file:
+                    fname = vma.get_name(self.context, task)
+                    if fname.endswith(("/bin/bash", "/bin/dash", "/bin/sh")):
```
ikelos: Since this list is used twice, it should definitely be pulled out and become common between them both.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                        bash_was_last = False
+
+                if vma.vm_file or vma.get_protection() != "rw-":
+                    continue
+            except (Exception, BaseException):
+                continue
+
+            if bash_was_last == False:
+                continue
+            vma_start = int(vma.vm_start)
```
ikelos: This looks like fairly similar code to above, is there some way of pulling both of these out into common code?  It's right on the border line of being better to reuse the code, so not saying it's necessary, but just wanted to see how clean it would make things if it were shared...

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                    continue
+                try:
+                    ptr_cache[ptr] = 1
+                    ptr_test2 = proc_layer.read(ptr + 20, addr_sz)
+                    if not ptr_test2:
+                        continue
+                except (Exception, BaseException):
+                    continue
+
+                ptr2 = struct.unpack(pack_format, ptr_test2)[0]
```
ikelos: Again, consider `ptr2, = ...` rather than `ptr2 = ...[0]`

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                try:
+                    test = proc_layer.read(ptr2 + 4, 4), "utf-8"
+                except (Exception, BaseException):
+                    continue
+                if not test or test != b"\x40\x00\x00\x00":
```
ikelos, line 200: This is a very specific value, please add a comment or a docstring to explain where this value came from and when it's here.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                    continue
+                hash_table_name = HashIntermedSymbols.create(
```
ikelos: Feels like this table should be constructed once at the start of use, and then the name passed in when this method is called, otherwise you'll have *hundreds* of these tables floating around, even though the content will be identical.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                    self.context, self.config_path, "linux", hash_json_file
+                )
+                htable = self.context.object(
```
ikelos: Might want some try/catch for `InvalidAddress` exceptions around this object call, depends how certain you are you've already checked the pointer address for validity

volatility3/framework/plugins/linux/bash_hash.py
```diff
@@ -0,0 +1,111 @@
...
+
+        vmlinux = self.context.modules[self.config["kernel"]]
+        is_32bit = not symbols.symbol_table_is_64bit(
+            self.context, vmlinux.symbol_table_name
+        )
+        if is_32bit:
+            hash_json_file = "hash32"
+        else:
+            hash_json_file = "hash64"
+        hash_table_name = HashIntermedSymbols.create(
```
ikelos, line 50: This is good, make the table once and just pass around the name.

volatility3/framework/plugins/linux/bash_hash.py
```diff
@@ -0,0 +1,111 @@
...
+    def bash_hash_entries(self, task, proc_layer_name, hash_table_name):
+
+        nbuckets_offset = self.context.symbol_space.get_type(
+            hash_table_name + constants.BANG + "bash_hash_table"
+        ).relative_child_offset("nbuckets")
+        proc_layer = self.context.layers[proc_layer_name]
+
+        for off in proc_layer.scan(
+            self.context,
+            scanners.BytesScanner(b"\x40\x00\x00\x00"),
```
ikelos, line 84: Again, very specific value, please document how it came to live in this plugin, what it means, etc.  5;)

volatility3/framework/symbols/linux/extensions/hash.py
```diff
@@ -0,0 +1,100 @@
...
+
+from typing import Dict, Tuple
+
+from volatility3.framework import exceptions, constants
+from volatility3.framework import objects, interfaces
+from volatility3.framework.objects import utility
+
+
+class bash_hash_table(objects.StructType):
+    def __init__(
```
ikelos, line 13: This doesn't *look* like the constructor does anything new/special that the super constructor wouldn't do.  Am I missing something here or is this unnecessary?

volatility3/framework/symbols/linux/extensions/hash.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def is_valid(self):
+        try:
+            bucket_array = self.bucket_array
+            nbuckets = self.nbuckets
+            nentries = self.nentries
+        except exceptions.InvalidAddressException:
+            return False
+
+        if not bucket_array:
```
ikelos: Feels like these could be turned into `return bucket_array and nbuckets == 64 and entries >= 1`, but happy to keep it expanded if you feel it's more legible/understandable.  5:)

volatility3/framework/symbols/linux/extensions/hash.py
```diff
@@ -0,0 +1,100 @@
...
+            return False
+
+        return True
+
+    def __iter__(self):
+        if self.is_valid():
+
+            seen = {}
+            bucket_array = self.context.object(
+                self.symbol_table_name + constants.BANG + "array",
```
ikelos, line 48: This can be called at the start of this method, rather than defining the `__init__` just to cache it in the class?

volatility3/framework/symbols/linux/extensions/hash.py
```diff
@@ -0,0 +1,100 @@
...
+                    continue
+                try:
+                    bucket_contents = bucket_ptr.dereference().cast(
+                        self.symbol_table_name + constants.BANG + "bucket_contents"
+                    )
+                    while bucket_contents.times_found > 0 or bucket_contents.next != 0:
+                        try:
+                            if bucket_contents in seen:
+                                break
+                            seen[bucket_contents] = 1
```
ikelos, line 72: This is putting the whole variable into the seen set.  I'm not sure that equality will ever fire for two differently constructed objects, even if the data in them is the same?  Are you trying to check that the offset hasn't been seen, or that the data at `bucket_contents` hasn't been seen?  Either way, this should probably we reworked somehow to check for the right thing.

volatility3/framework/symbols/linux/extensions/hash.py
```diff
@@ -0,0 +1,100 @@
...
+                            pdata = bucket_contents.data
+                            if (
+                                0 <= pdata.flags <= 2
```
ikelos, line 75: This probably also needs a bunch more error checking.  `bucket_contents.data` might points off to anything, so then `pdata.flags` won't necessarily exist as an attribute.

ikelos: I'd move the exception handling out to wrap the contents of the loop, rather than having multiple nested tries all for the same attempt.

ikelos (commented):
First off, thanks for the submission!  5:D  This'll still need checking by our linux chap @atcuno, so I can't give it a direct sign off myself, but I've given the style/code a good look over for you.  So, a lot of little comments but probably only two or three big take aways...

* Firstly, use sets rather than dictionaries when all you're testing is membership.
* Secondly, exception handling should be as high up and cover as much as possible, per individual item being tested.
* Thirdly, a lot of little stylistic bits and pieces.  These aren't show stoppers, but just good practice.  Happy to discuss any of them and for some to get missed out if you'd prefer, but there's quite a lot of them in there at the moment.

volatility3/framework/symbols/linux/extensions/hash.py
a5hlynx: Sorry for my late reply and thank you for your comments.
I will check through your comments and modify/commit the codes.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+        if is_32bit:
+            pack_format = "I"
+            hash_json_file = "hash32"
+        else:
+            pack_format = "Q"
+            hash_json_file = "hash64"
+        for task in tasks:
+            proc_layer_name = task.add_process_layer()
+            if not proc_layer_name:
+                yield (0, (task.pid, utility.array_to_string(task.comm), ""))
```
atcuno, line 44: @ikelos should this empty string be NotApplicableValue instead? If the process doesn't have a layer then its likely a kernel thread, which never has environment variables.

volatility3/framework/plugins/linux/bash_env.py
atcuno: It is expected. This is just filtering to the types of memory regions to scan.

volatility3/framework/plugins/linux/bash_env.py
atcuno: line 140 checks for eqidx == -1

volatility3/framework/plugins/linux/bash_env.py
atcuno: I believe the check on line 140 "idx < eqidx:" handles the null before the =

atcuno: I agree the reading over and over based on the `sizes` list seems inefficient, but just reading the whole page wouldn't be a good fix as 1) `var` is likely not page aligned and 2) the environment can span page boundaries. The code attempts to find the smallest valid read that finds a complete variable key/data pair. It seems like 135 could be a `break` though because if reading a smaller size fails then larger reads will also fail. Thoughts?

volatility3/framework/plugins/linux/bash_env.py
atcuno: also, the 4 lines of if/else could just be collapsed to:

bash_was_last = fname.endswith("tuple of shell paths")

volatility3/framework/plugins/linux/bash_env.py
ikelos: Yep, good spot thanks!  Anywhere that a value isn't applicable, it's best to indicate that with a specific type, so that other interfaces (or programs) can understand and handle it better.

ikelos:
@a5hlynx would you be able to investigate the comments we've made please?  It looks like you're keeping it up to date with develop, but there's been no action on the comments made...

a5hlynx:
> @a5hlynx would you be able to investigate the comments we've made please? It looks like you're keeping it up to date with develop, but there's been no action on the comments made...

Sorry for not responding.
Modified some parts based upon the comments. I will modify the unresolved as soon as I address.

volatility3/framework/plugins/linux/bash_env.py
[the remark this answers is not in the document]
a5hlynx: replaced with exceptions.InvalidAddressException.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                            continue
+                        good_varstr = varstr
+                        break
+                    if good_varstr:
+                        try:
+                            good_varstr = str(good_varstr[:idx], "utf-8")
+                            key = good_varstr[:eqidx]
+                            val = good_varstr[eqidx + 1 :]
+                            yield (key, val)
+                        except (Exception, BaseException):
```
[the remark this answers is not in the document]
a5hlynx: replaced with exceptions.InvalidAddressException and UnicodeDecodeError.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                if vma.vm_file:
+                    fname = vma.get_name(self.context, task)
+                    if fname.endswith(("/bin/bash", "/bin/dash", "/bin/sh")):
+                        bash_was_last = True
+                    else:
+                        bash_was_last = False
+
+                if vma.vm_file or vma.get_protection() != "rw-":
+                    continue
+            except (Exception, BaseException):
```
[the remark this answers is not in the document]
a5hlynx: replaced with exceptions.InvalidAddressException.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                ptr_test = vma_data[off : off + addr_sz]
+                ptr = struct.unpack(pack_format, ptr_test)[0]
+                if ptr in ptr_cache:
+                    continue
+                try:
+                    ptr_cache[ptr] = 1
+                    ptr_test2 = proc_layer.read(ptr + 20, addr_sz)
+                    if not ptr_test2:
+                        continue
+                except (Exception, BaseException):
```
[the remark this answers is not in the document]
a5hlynx: replaced with exceptions.InvalidAddressException.

volatility3/framework/plugins/linux/bash_env.py
```diff
@@ -0,0 +1,227 @@
...
+                    continue
+
+                ptr2 = struct.unpack(pack_format, ptr_test2)[0]
+                try:
+                    test = proc_layer.read(ptr2 + 4, 4), "utf-8"
+                except (Exception, BaseException):
```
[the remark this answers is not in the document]
a5hlynx: replaced with exceptions.InvalidAddressException.

volatility3/framework/symbols/linux/extensions/hash.py
```diff
@@ -0,0 +1,100 @@
...
+                subtype=self.context.symbol_space.get_type(
+                    self.symbol_table_name + constants.BANG + "pointer"
+                ),
+                count=self.nbuckets,
+            )
+
+            for i in range(len(bucket_array)):
+                try:
+                    bucket_ptr = bucket_array[i]
+                except (Exception, BaseException):
```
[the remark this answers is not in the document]
a5hlynx: replaced with exceptions.InvalidAddressException.

volatility3/framework/symbols/linux/extensions/hash.py
```diff
@@ -0,0 +1,100 @@
...
+                                0 <= pdata.flags <= 2
+                                and bucket_contents.times_found > 0
+                            ):
+                                try:
+                                    path = utility.array_to_string(
+                                        bucket_contents.data.path.dereference()
+                                    )
+                                    if path.startswith("/"):
+                                        yield (bucket_contents)
+                                except (Exception, BaseException):
```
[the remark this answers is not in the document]
a5hlynx: replaced with exceptions.InvalidAddressException.

volatility3/framework/symbols/linux/extensions/hash.py
```diff
@@ -0,0 +1,100 @@
...
+                                    continue
+
+                            bucket_contents = bucket_contents.next
+                        except (Exception, BaseException):
```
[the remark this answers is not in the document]
a5hlynx: replaced with exceptions.InvalidAddressException.

volatility3/framework/symbols/linux/extensions/hash.py
```diff
@@ -0,0 +1,100 @@
...
+                            continue
+                except (Exception, BaseException):
```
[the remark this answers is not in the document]
a5hlynx: replaced with exceptions.InvalidAddressException.

volatility3/framework/plugins/linux/bash_env.py
a5hlynx: modified based upon the comment.

volatility3/framework/plugins/linux/bash_env.py
a5hlynx: modified based upon the comment.

volatility3/framework/plugins/linux/bash_env.py
a5hlynx: defined shells = ("/bin/bash", "/bin/dash", "/bin/sh") at the top,
modified endswith(("/bin/bash", "/bin/dash", "/bin/sh")) to endswith(self.shells).

volatility3/framework/plugins/linux/bash_env.py
a5hlynx: removed cast.

volatility3/framework/plugins/linux/bash_env.py
a5hlynx: modified to use set.

volatility3/framework/plugins/linux/bash_env.py
a5hlynx: removed the if statement.

volatility3/framework/plugins/linux/bash_env.py
a5hlynx: modified to firstaddr, = stuct.unpack(pack_format, firstaddrstr)

volatility3/framework/plugins/linux/bash_env.py
a5hlynx: removed the if statement.

volatility3/framework/plugins/linux/bash_env.py
a5hlynx: modified this part based upon the comment

ikelos:
Hiya @a5hlynx, you mentioned six months ago you were going to work on the comments we'd made, and we'd really like to see your code land in the tree.  Would you be able to give it a look and make the changes?  We can get things rolling again if you're happy to address the comments we raised?

a5hlynx:
Sorry for my late reply.

I closed this request once.
I will re-open it when I am ready.

atcuno:
> Sorry for my late reply.
> 
> I closed this request once. I will re-open it when I am ready.

These are important plugins to the framework's parity effort with Volatility 2. Do you know if you would have time to address the feedback within the new few weeks? If not, we will likely assign one of our developers to complete them.

ikelos:
You can mark the pull request as a draft if it's not ready, but that way at least people can see it and find it easily, rather than closing it and making it look like the work in it wasn't necessarily usable?

a5hlynx:
> You can mark the pull request as a draft if it's not ready, but that way at least people can see it and find it easily, rather than closing it and making it look like the work in it wasn't necessarily usable?

Thanks for your suggestion.  I marked it as draft.

PR #894, opened 2023-01-05
add linux.iomem plugin based on vol2 plugin by atcuno

eve-mem:
This plugin is based on the vol2 plugin by @atcuno.

It shows a similar output to running `sudo cat /proc/iomem` on a running system. 

Example output:
```
python vol.py -r pretty -f linux-sample-1.dmp linux.iomem
Volatility 3 Framework 2.4.1
      |                          NAME |      START |          END
*     |                       PCI mem |        0x0 | 0xffffffffff
**    |                      reserved |        0x0 |       0xffff
**    |                    System RAM |    0x10000 |      0x9efff
**    |                      reserved |    0x9f000 |      0x9ffff
**    |               PCI Bus 0000:00 |    0xa0000 |      0xbffff
**    |                     Video ROM |    0xc0000 |      0xc7fff
**    |                   Adapter ROM |    0xca000 |      0xcafff
**    |               PCI Bus 0000:00 |    0xcc000 |      0xcffff
**    |               PCI Bus 0000:00 |    0xd0000 |      0xd3fff
**    |               PCI Bus 0000:00 |    0xd4000 |      0xd7fff
**    |               PCI Bus 0000:00 |    0xd8000 |      0xdbfff
**    |                      reserved |    0xdc000 |      0xfffff
***   |                    System ROM |    0xf0000 |      0xfffff
**    |                    System RAM |   0x100000 |   0x1fedffff
***   |                   Kernel code |  0x1000000 |    0x1359525
***   |                   Kernel data |  0x1359526 |    0x1694dff
***   |                    Kernel bss |  0x172a000 |    0x1807fff
**    |                   ACPI Tables | 0x1fee0000 |   0x1fefefff
**    |     ACPI Non-volatile Storage | 0x1feff000 |   0x1fefffff
**    |                    System RAM | 0x1ff00000 |   0x1fffffff
**    |               PCI Bus 0000:00 | 0x20000000 |   0xfebfffff
***   |                  0000:00:0f.0 | 0x20000000 |   0x20007fff
***   |                  0000:00:10.0 | 0x20008000 |   0x2000bfff
***   |               PCI Bus 0000:22 | 0xe5b00000 |   0xe5bfffff
***   |               PCI Bus 0000:1a | 0xe5c00000 |   0xe5cfffff
<snip>
```

### DIFF
--- a/volatility3/framework/plugins/linux/iomem.py
+++ b/volatility3/framework/plugins/linux/iomem.py
@@ -1,0 +1,139 @@
+# This file is Copyright 2023 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import logging
+from typing import List
+
+from volatility3.framework import renderers, interfaces, exceptions
+from volatility3.framework.configuration import requirements
+from volatility3.framework.objects import utility
+from volatility3.framework.renderers import format_hints
+
+vollog = logging.getLogger(__name__)
+
+
+class IOMem(interfaces.plugins.PluginInterface):
+    """Generates an output similar to /proc/iomem on a running system."""
+
+    _required_framework_version = (2, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Linux kernel",
+                architectures=["Intel32", "Intel64"],
+            )
+        ]
+
+    @classmethod
+    def parse_resource(
+        cls,
+        context: interfaces.context.ContextInterface,
+        vmlinux_module_name: str,
+        resource_offset: int,
+        seen: set = set(),
+        depth: int = 0,
+    ):
+        """Recursively parse from a root resource to find details about all related resources.
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            vmlinux_module_name: The name of the kernel module on which to operate
+            resource_offset: The offset to the resouce to be parsed
+            seen: The set of resource offsets that have already been parsed
+            depth: How deep into the resource structure we are
+
+        Yields:
+            Each row of output
+        """
+        # create the resource object
+        vmlinux = context.modules[vmlinux_module_name]
+        resource = vmlinux.object("resource", resource_offset)
+
+        # extract the information required for this resource
+        name = utility.pointer_to_string(resource.name, 128)
+        start = format_hints.Hex(resource.start)
+        end = format_hints.Hex(resource.end)
+
+        # mark this resource as seen in the seen set. Normally this should not be needed but will protect
+        # against possible infinite loops. Warn the user if an infinite loop would have happened.
+        if resource_offset in seen:
+            vollog.warning(
+                f"The resource object at {resource_offset:#x} '{name}' has already been processed, "
+                "this should not normally occur. No further results from related resources will be "
+                "displayed to protect against infinite loops."
+            )
+            return None
+        else:
+            seen.add(resource_offset)
+
+        # yield information on this resource
+        yield depth, (name, start, end)
+
+        # process child resource if this exists
+        if resource.child != 0:
+            yield from cls.parse_resource(
+                context,
+                vmlinux_module_name,
+                resource.child,
+                seen,
+                depth + 1,
+            )
+
+        # process sibling resource if this exists
+        if resource.sibling != 0:
+            yield from cls.parse_resource(
+                context,
+                vmlinux_module_name,
+                resource.sibling,
+                seen,
+                depth,
+            )
+
+    def _generator(self):
+        """Generates an output similar to /proc/iomem on a running system
+
+        Args:
+            None
+
+        Yields:
+            Each row of output using the parse_resource function
+        """
+
+        # get the kernel module from the current context
+        vmlinux_module_name = self.config["kernel"]
+        vmlinux = self.context.modules[vmlinux_module_name]
+
+        # check that the iomem_resource symbol exists
+        # normally exported in /kernel/resource.c
+        try:
+            iomem_root_offset = vmlinux.get_absolute_symbol_address("iomem_resource")
+        except exceptions.SymbolError:
+            iomem_root_offset = None
+
+        # error if 'iomem_resource' is not found
+        if not iomem_root_offset:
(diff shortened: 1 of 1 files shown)


eve-mem:
Perhaps this plugin should be in `linux.proc`? Keep all of the procfs plugins together?

volatility3/framework/plugins/linux/iomem.py
```diff
@@ -0,0 +1,139 @@
...
+
+        Yields:
+            Each row of output
+        """
+        # create the resource object
+        vmlinux = context.modules[vmlinux_module_name]
+        resource = vmlinux.object("resource", resource_offset)
+
+        # extract the information required for this resource
+        name = utility.pointer_to_string(resource.name, 128)
```
atcuno: this needs to try/except InvalidAddressException

volatility3/framework/plugins/linux/iomem.py
```diff
@@ -0,0 +1,139 @@
...
+            resource_offset: The offset to the resouce to be parsed
+            seen: The set of resource offsets that have already been parsed
+            depth: How deep into the resource structure we are
+
+        Yields:
+            Each row of output
+        """
+        # create the resource object
+        vmlinux = context.modules[vmlinux_module_name]
+        resource = vmlinux.object("resource", resource_offset)
```
atcuno: this needs to try/except InvalidAddressException

atcuno:
@ikelos after the two try/except clauses are added as smear protection then this is good to go from @eve-mem

ikelos:
Cool, thanks for the review @atcuno !  5:)

eve-mem:
@ikelos and @atcuno - thank you very much for the review. Hopefully I've covered everything off now - please let me know if you think of any thing else needed for this plugin.

atcuno:
this looks good now

volatility3/framework/plugins/linux/iomem.py
```diff
@@ -0,0 +1,156 @@
...
+from volatility3.framework.objects import utility
+from volatility3.framework.renderers import format_hints
+
+vollog = logging.getLogger(__name__)
+
+
+class IOMem(interfaces.plugins.PluginInterface):
+    """Generates an output similar to /proc/iomem on a running system."""
+
+    _required_framework_version = (2, 0, 0)
```
ikelos, line 18: Since you're exposing `parse_resource` for others to use, you should also start including a `_version` number so people can write their code without wondering which version of this plugin they'll need to talk to.

volatility3/framework/plugins/linux/iomem.py
```diff
@@ -0,0 +1,156 @@
...
+            resource = vmlinux.object("resource", resource_offset)
+        except exceptions.InvalidAddressException:
+            vollog.warning(
+                f"Unable to create resource object at {resource_offset:#x}. This resource, "
+                "its sibling, and any of it's childern and will be missing from the output."
+            )
+            return None
+
+        # extract the information required for this resource
+        start = format_hints.Hex(resource.start)
```
ikelos: Generally it makes it slightly easier to read and reason about if the format_hint is only applied at the moment of use (if, during the yield statement).  Otherwise someone unfamiliar with the code may wonder why it needs putting through this function/class and waste time because we did the action far away from where it was needed.

volatility3/framework/plugins/linux/iomem.py
```diff
@@ -0,0 +1,156 @@
...
+        end = format_hints.Hex(resource.end)
+
+        # get name with protection against smear as following a pointer
+        try:
+            name = utility.pointer_to_string(resource.name, 128)
+        except exceptions.InvalidAddressException:
```
ikelos, line 67: Good work remembering to handle errors, thanks!  5:)  It's very easy to forget, so it's much appreciated when it's in the places that could easily fail but people don't think about.

volatility3/framework/plugins/linux/iomem.py
```diff
@@ -0,0 +1,156 @@
...
+            )
+
+        # recursively parse the resources starting from the root resource at 'iomem_resource'
+        yield from self.parse_resource(
+            self.context, vmlinux_module_name, iomem_root_offset
+        )
+
+    def run(self):
+        columns = [
+            ("NAME", str),
```
ikelos: I'm not sure we usually all-caps column names?  It feels a little... shouty.  5;P  I think just initial capitalization would fit in with all the other plugins...

volatility3/framework/plugins/linux/iomem.py
```diff
@@ -0,0 +1,156 @@
...
+            iomem_root_offset = None
+
+        # error if 'iomem_resource' is not found
+        if not iomem_root_offset:
+            raise TypeError(
+                "This plugin requires the iomem_resource structure. This structure is not present in the supplied symbol table. This means you are either analyzing an unsupported kernel version or that your symbol table is corrupt."
+            )
+
+        # error if type 'resource' is not found
+        if not vmlinux.has_type("resource"):
```
ikelos, line 152: I might be tempted to do this at the start of run.  It really doesn't make much of a difference, but it's not expected that other plugins would ever call this generator and it feels like "can this plugin run" checks should happen at the start `run` rather than in the generator function?  Happy either way, but figured I'd mention it.  There's so little else in this commit that needs mentioning, really good work!  5:)

ikelos (commented):
Looks good to me, a couple minor bits (some of which are stylistic, so don't need fixing if you don't want) and if @atcuno's happy, then we can merge it.  5:)

volatility3/framework/plugins/linux/iomem.py
eve-mem: That is thanks to @atcuno wisdom! I will bear it in mind for the other plugins.

eve-mem:
@ikelos I've updated the plugin with those comments in mind. Let me know if you think there is anything else at all that needs tweaking.

volatility3/framework/plugins/linux/iomem.py
atcuno: This is the only comment left here. If you are okay with the checks where they are then lets merge it.

eve-mem: I can change where those checks happen, thanks for the reminder.

eve-mem: Ah, looks like I already moved these checks into run. I think this is sorted now?

ikelos:
All looks good, thanks!  5:)

PR #899, opened 2023-01-13
Dump ELFs to file

### LINKED ISSUE #847: Replacement of linux_procdump command
Hey,

We have been using linux_procdump command for dumping the executable of a process. It is not available in volatility3. Please tell the replacement for this plugin.

I will be grateful
Thanks

cstation:
In an attempt to revive #606 (but skipping the part of adding timestamps to the output of `linux.pslist`):

For Linux images, there hasn't been the possibility yet to dump ELFs to a file.
This PR seeks to make this possible, in a way that is as consistent as possible with the behaviour and structure of Volatility 3's `windows.PsList` plugin and Volatility 2's `linux_procdump`.

The `--dump` flag on `linux.PsList` dumps the main executable of the process, whereas on `linux.Elfs`, it dumps all valid ELFs encountered in the memory space of the process. The MD5 hashes of the dumped main executables correspond to the MD5s of the results produced by Volatility 2.

I noticed that this partially seems to have some overlap with #847. Nevertheless, the functionality is different: this PR adds support for dumping to `linux.pslist` (dumping the main executable) and `linux.elfs` (dumping all ELFs), whereas #847 seeks to add support for dumping to `linux.proc` (dumping memory segments related to a process).

Closes #847.


ikelos:
@atcuno could you please take a look through this and make sure the dumping is suitable and on the same lines as what we do with windows.pslist?  I'll give it a python pass over when I get a bit of time...

volatility3/framework/plugins/linux/elfs.py
```diff
@@ -36,9 +42,95 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+            task: the task object whose memory should be output
+            open_method: class to provide context manager for opening the file
+        Returns:
+            An open FileHandlerInterface object containing the complete data for the task or None in the case of failure
+        """
+
+        proc_layer = context.layers[layer_name]
+        file_handle = None
+
+        try:
```
atcuno: Catching a broad Exception over a whole code attempt is frowned up in Volatility 3 code. Looking at the code inside this block, the main concerns for exception would be invalid address exceptions anytime an object is constructed (line 79) or when a pointer is followed/dereferenced. This should be changed.

cstation: You're right! I've removed the broad try-except clause from the code.

volatility3/framework/plugins/linux/elfs.py
```diff
@@ -36,9 +42,92 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+                continue
+
+            start = phdr.p_vaddr
+            size = phdr.p_memsz
+            end = start + size
+
+            # Use complete memory pages for dumping
+            # If start isn't a multiple of 4096, stick to the highest multiple < start
+            # If end isn't a multiple of 4096, stick to the lowest multiple > end
+            if start % 4096:
```
ikelos, line 100: Are these values specific to the elf format or to the layer they're being read from?  If it's a part of the ELF format, then please pull it out as a variable (and do calculations on that variable to find masks) rather than hardcoding the number in lots of places.

volatility3/framework/plugins/linux/elfs.py
```diff
@@ -36,9 +42,92 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+                start = start & ~0xFFF
+
+            if end % 4096:
+                end = (end & ~0xFFF) + 4096
+
+            real_size = end - start
+
+            if real_size < 0 or real_size > 100000000:
```
ikelos: The same with this (arbitrary) limit?  I'm fine with a sanity check value to make sure we're not pulling out anything too large, but please make it a variable so that it could be parameterized in the future if needed for some reason.

ikelos (commented):
Cool, looks good thanks, just a few constants that could do with being put in a single variable somewhere (or calculated from the layer if that maxes sense) and not sure why `_get_task_fields` got pushed back into the generator, but generally all good.  Appreciate the doc strings!  Happy to merge once the constants have been drawn out.  5:)

volatility3/framework/plugins/linux/elfs.py
atcuno: I think the the 4096 is okay here, its the standard page size

volatility3/framework/plugins/linux/elfs.py
cstation: I introduced it to stay as close as possible to the Volatility 2 functionality / behaviour ([link](https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/overlays/linux/linux.py#L1272)). I've changed it to the official size limit of an ELF (4GB).

ikelos: Ok, I prefer specific constants to be pulled out so it's easy to see them and make sure that we don't duplicate things but always work from the same value.  I don't know if it's worth dragging out to the `constants.linux` module, but it's probably worth pulling up to the plugin itself, if not the module.

cstation: Sure! I've put it in the `constants.linux` to mirror what has been done at the Windows 'side' (https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/symbols/windows/extensions/pe.py#L135). If you'd rather prefer it to be in  `elfs.py`,  should I implement it like https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/proc.py#L25?

cstation: @ikelos, could you please provide some directions on finalizing this?

ikelos (approved):
Sorry for the delay in checking this, I think you've addressed all the issues, so I'm going to merge this in.  Thanks very much!

PR #900, opened 2023-01-16
Add debug message for symbol_table_from_offset

paulkermann:
Leave a log message if no GUID is found explicitly


ikelos (approved):
Looks good, we probably should have had that in there from the beginning, thanks!  5:)

robinshen36:
@algora-pbc attempt

Hey! This caught my eye while browsing.

I've been wanting to dig deeper into Security - this looks like the perfect opportunity.

Would probably begin with checking the symbol tables and see how it goes.

What's the budget looking like? Around $240-$400 feels about right, but flexible.

Talk soon hopefully.

PR #904, opened 2023-01-17
Add escapechar to csvwriter

paulkermann:
Otherwise it could throw "Need to escape, but no escapechar set".


ikelos (approved):
Looks good, thanks.

PR #920, opened 2023-03-05
Windows: Memoize part of the pool handling code

ikelos:
This is designed to replace PR #886, it will properly check for an existing symbol table name, but keeps track in a table to speed up creation is a lot are made.  It also memoizes the symbol table, to reduce creation of the table (which the original patch didn't do).  Hopefully this covers all the fixes made.  If so, I'm happy to merge it in...


ikelos:
Looks like it's not ready for the prime-time yet, I'll look into it and see what's causing the problems over the next couple of days.

ikelos:
Ok, one of the changes to using the kernel module was still being given an absolute offset but looking it up as if it were relative.

PR #921, opened 2023-03-08
Fix Linux psscan

eve-mem:
This fixes some issues raised by "M N" on volatility slack channel about  linux.psscan. 

Fixes are:
- handling symbol offsets
- finding layer to scan using dependencies rather than being hard coded
- more debug messages

### DIFF
--- a/volatility3/framework/plugins/linux/psscan.py
+++ b/volatility3/framework/plugins/linux/psscan.py
@@ -112,9 +112,16 @@ class PsScan(interfaces.plugins.PluginInterface):
             if "_sched_class" in symbol:
                 # use canonicalize to set the appropriate sign extension for the addr
                 addr = kernel_layer.canonicalize(vmlinux.get_symbol(symbol).address + vmlinux.offset)
+                packed_addr = struct.pack(pack_format, addr)
+
+                # debug message to show needles being searched for and symbol names
+                vollog.debug(
+                    f"Found a sched_class named {symbol} at offset {hex(addr)}. Will scan for these bytes: {packed_addr.hex()}"
+                )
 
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/linux/psscan.py
```diff
@@ -111,12 +111,33 @@ def scan_tasks(
...
+                vollog.debug(
+                    f"Found a sched_class named {symbol} at offset {hex(addr)}. Will scan for these bytes: {packed_addr.hex()}"
+                )
 
                 # append to needles list the packed hex for searching
-                needles.append(struct.pack(pack_format, addr))
+                needles.append(packed_addr)
+        # find the memory layer to scan
+        if len(kernel_layer.dependencies) > 1:
+            vollog.warning(
```
ikelos, line 128: Not sure this needs to be a warning, but I'm happy leaving it for now.  I don't think the swap functionality (which is what would most likely cause more than one dependency) is used all that often.

volatility3/framework/plugins/linux/psscan.py
```diff
@@ -111,12 +111,33 @@ def scan_tasks(
...
+                f"Kernel layer depends on multiple layers however only {kernel_layer.dependencies[0]} will be scanned by this plugin."
+            )
+        elif len(kernel_layer.dependencies) == 0:
```
ikelos, line 131: This might not be worthwhile, unless the plugin can't scan a flat file without knowledge of the kernel (which I think this implementation can't).

ikelos (approved):
Looks good, just a couple comments in general, but nothing preventing the merge.  Thanks for the fix!  5:)

PR #935, opened 2023-04-07
Fix: typo for `linux.iomem` plugin.

digitalisx:
## Descriptions
I found a typo in the error message and comment on `linux.iomem` plugin that we worked on recently and fixed it. :)


ikelos (approved):
Awesome, thank you!  Happily accepted...  5:D

Issue #930, opened 2023-03-31
Linux paths reported are incorrect in multiple plugins


eve-mem:
**Describe the bug**
The file paths reported in various linux plugins is incorrect. 

This was discussed this pull request by @paulkermann -  https://github.com/volatilityfoundation/volatility3/pull/786

@ikelos, @gcmoreira, and @atcuno discuss that this approach is the best:
```
Instead of building the path for then split it, it would be better to simply do:

dname = dentry.d_name.name_as_str()
```

**Context**
Volatility Version:  Volatility 3 Framework 2.4.2 - commit 60292c2da1efa886d0f46bd720af537a6069a623
Operating System:  N/A
Python Version:  Python 3.11.1
Suspected Operating System:  N/A - various linux kernels
Command:  linux.elfs, linux.lsof, linux.proc

**To Reproduce**

1. Use command any of these plugins: linux.elfs, linux.lsof, linux.proc
2. See issue in reported paths

linux.elfs:
```
Volatility 3 Framework 2.4.2
Progress:  100.00               Stacking attempts finished
PID     Process Start   End     File Path

1       init    0x7f489741c000  0x7f489741e000  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libdl-2.13.so
1       init    0x7f4897620000  0x7f48977a2000  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libc-2.13.so
1       init    0x7f48979ab000  0x7f48979c9000  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libselinux.so.1
1       init    0x7f4897e0b000  0x7f4897e2b000  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/ld-2.13.so
1       init    0x7fff14da9000  0x7fff14daa000  [vdso]
322     udevd   0x7fc909adb000  0x7fc909ae6000  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libnss_files-2.13.so
322     udevd   0x7fc909ce7000  0x7fc909cf1000  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libnss_nis-2.13.so
322     udevd   0x7fc909ef2000  0x7fc909f07000  /lib/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libnsl-2.13.so
[...]

Issue #909, opened 2023-02-11
Maple Trees Replaced MMap - Broken v6.1


sempervictus:
**Describe the bug**
```python
Traceback (most recent call last):
  File "/home/rage/Code/Python/volatility3/vol.py", line 10, in <module>
    volatility3.cli.main()
  File "/tmp/volatility3/volatility3/cli/__init__.py", line 797, in main
    CommandLine().run()
  File "/tmp/volatility3/volatility3/cli/__init__.py", line 445, in run
    renderers[args.renderer]().render(constructed.run())
  File "/tmp/volatility3/volatility3/cli/text_renderer.py", line 193, in render
    grid.populate(visitor, outfd)
  File "/tmp/volatility3/volatility3/framework/renderers/__init__.py", line 241, in populate
    for level, item in self._generator:
  File "/tmp/volatility-gpg/linux/gpg_full.py", line 188, in _generator
    for offset, passphrase, secret_size, plaintext in self.locate_timestamps(self.context, tasks):
  File "/tmp/volatility-gpg/linux/gpg_full.py", line 98, in locate_timestamps
    memory_sections = list(task.get_process_memory_sections(heap_only=False))
  File "/tmp/volatility3/volatility3/framework/symbols/linux/extensions/__init__.py", line 221, in get_process_memory_sections
    for vma in self.mm.get_mmap_iter():
  File "/tmp/volatility3/volatility3/framework/symbols/linux/extensions/__init__.py", line 302, in get_mmap_iter
    if not self.mmap:
  File "/tmp/volatility3/volatility3/framework/objects/__init__.py", line 968, in __getattr__
    raise AttributeError(
AttributeError: StructType has no attribute: symbol_table_name1!mm_struct.mmap
```
^^ its hard to get an mmap iterator when there is no mmap struct member in-play.
```c
struct [mm_struct](https://elixir.bootlin.com/linux/v5.19.17/C/ident/mm_struct) {
	struct {
		struct [vm_area_struct](https://elixir.bootlin.com/linux/v5.19.17/C/ident/vm_area_struct) *[mmap](https://elixir.bootlin.com/linux/v5.19.17/C/ident/mmap);		/* list of VMAs */
```
became
```c
struct [mm_struct](https://elixir.bootlin.com/linux/latest/C/ident/mm_struct) {
	struct {
[...]

PR #941, opened 2023-04-25
Linux: Fix repeating path issue in LinuxUtilities

eve-mem:
This would fix issue https://github.com/volatilityfoundation/volatility3/issues/930

This pull request fixes a bug in many Linux plugins where there are incorrect repeating elements in the paths as reported by volatility.

For example `/sbin/sbin/init` being reported where it should be `/sbin/init`. 

This contains the fix suggested by @gcmoreira on this pull request https://github.com/volatilityfoundation/volatility3/pull/786

It also fixes another smaller bug that also results in repeating elements in the paths, but this is much less common. (This is discussed here https://github.com/volatilityfoundation/volatility3/issues/930#issuecomment-1519446869)

This has been tested on memory dumps with the following Linux kernels: Tested on 3.7.10, 3.9.11, 3.10.108, 3.11.10, 3.12.74, 3.13.11, 3.14.79, 3.15.10, 3.16.85, 3.17.8, 3.18.140, 3.19.8, 4.0.9, 4.1.52, 4.2.8, 4.3.6, 4.4.302, 4.5.7, 4.6.7, 4.7.10, 4.8.17, 4.9.337, 4.10.17, 4.11.12, 4.12.14, 4.13.16, 4.14.309, 4.15.18, 4.16.18, 4.17.19, 4.18.20, 4.19.275, 4.20.17, 5.0.21, 5.1.21, 5.2.21, 5.3.18, 5.4.234, 5.5.19, 5.6.19, 5.7.19, 5.8.18, 5.9.16, 5.10.172, 5.11.22, 5.12.19, 5.13.19, 5.14.21, 5.15.98, 5.16.20, 5.17.15, 5.18.19, 5.19.17, 6.0.19, 6.1.15, 6.2.2, 6.2.7, 6.3.0 (can share if interested)

6.1.15, 6.2.2, 6.2.7, 6.3.0 fail on some plugins due to this issue where the mm_struct changed since 6.1 - discussed in this issue https://github.com/volatilityfoundation/volatility3/issues/909

Here is a zip file with the outputs from each of the test images.

### DIFF
--- a/volatility3/framework/symbols/linux/__init__.py
+++ b/volatility3/framework/symbols/linux/__init__.py
@@ -65,7 +65,7 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
         ret_path: List[str] = []
 
         while dentry != rdentry or vfsmnt != rmnt:
-            dname = dentry.path()
+            dname = dentry.d_name.name_as_str()
             if dname == "":
                 break
 
--- a/volatility3/framework/symbols/linux/__init__.py
+++ b/volatility3/framework/symbols/linux/__init__.py
@@ -74,11 +74,12 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
                 if vfsmnt.get_mnt_parent() == vfsmnt:
                     break
 
+                if vfsmnt.get_mnt_mountpoint() == dentry:
+                     break
+
                 dentry = vfsmnt.get_mnt_mountpoint()
                 vfsmnt = vfsmnt.get_mnt_parent()
-
                 continue
-
             parent = dentry.d_parent
             dentry = parent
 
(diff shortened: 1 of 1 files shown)


eve-mem:
Here is the test results zip file: 
[test-results-issue-930.zip](https://github.com/volatilityfoundation/volatility3/files/11319168/test-results-issue-930.zip)

eve-mem:
This is likely not needed given https://github.com/volatilityfoundation/volatility3/pull/942

gcmoreira:
Hey @eve-mem, sorry for my delayed response and thank you for taking the time to try to fix it. 
Even though the code in this PR fixes the bug you reported in #930, there was another issue in my to-do list to solve. The legacy and a the new **_do_get_path()** implementations were living together in the vol3 code and were being used by different plugins. The bug and the behaviour you reported only happened in the legacy version. The PR #942 unifies everything in the new implementation which replaced **LinuxUtilities._do_get_path()**.

PR #942, opened 2023-04-25
Mountinfo fixes

### LINKED ISSUE #930: Linux paths reported are incorrect in multiple plugins

gcmoreira:
This PR replaces **LinuxUtilities._do_get_path()** with the new **mountinfo _do_get_path()** avoiding duplicate code.

As the bug reported by @eve-mem on issue #930 only happens in the old/legacy version of LinuxUtilities._do_get_path(), this PR also fixes it. It also fixes #786.

It was tested using the **linux.sockstat.Sockstat**, **linux.mountinfo.MountInfo** and **linux.proc.Maps** plugins in the following x86-64 kernel version: 
* 3.13.0
* 4.10.0
* 4.15.0
* 4.18.0
* 4.4.0
* 5.8.0
* 5.10.0

### DIFF
--- a/volatility3/framework/plugins/linux/mountinfo.py
+++ b/volatility3/framework/plugins/linux/mountinfo.py
@@ -12,6 +12,7 @@ from volatility3.framework.interfaces import plugins
 from volatility3.framework.symbols import linux
 from volatility3.plugins.linux import pslist
 
+
 vollog = logging.getLogger(__name__)
 
 MountInfoData = namedtuple(
@@ -140,30 +141,38 @@ class MountInfo(plugins.PluginInterface):
         )
 
     def _get_tasks_mountpoints(
-        self, tasks: Iterable[interfaces.objects.ObjectInterface], per_namespace: bool
+        self, tasks: Iterable[interfaces.objects.ObjectInterface], filtered_by_pids: bool
     ):
-        seen_namespaces = set()
+        seen_mountpoints = set()
         for task in tasks:
             if not (
-                task
-                and task.fs
-                and task.fs.root
-                and task.nsproxy
-                and task.nsproxy.mnt_ns
+                task and
+                task.fs and
+                task.fs.root and
+                task.nsproxy and
+                task.nsproxy.mnt_ns
             ):
-                # This task doesn't have all the information required
+                # This task doesn't have all the information required.
+                # It should be a kernel < 2.6.30
                 continue
 
             mnt_namespace = task.nsproxy.mnt_ns
-            mnt_ns_id = mnt_namespace.get_inode()
-
-            if per_namespace:
-                if mnt_ns_id in seen_namespaces:
-                    continue
-                else:
-                    seen_namespaces.add(mnt_ns_id)
+            try:
+                mnt_ns_id = str(mnt_namespace.get_inode())
+            except AttributeError:
+                mnt_ns_id = renderers.NotAvailableValue()
 
             for mount in mnt_namespace.get_mount_points():
+                # When PIDs are filtered, it makes sense that the user want to
+                # see each of those processes mount points. So we don't filter
+                # by mount id in this case.
+                if not filtered_by_pids:
+                    mnt_id = int(mount.mnt_id)
+                    if mnt_id in seen_mountpoints:
+                        continue
+                    else:
+                        seen_mountpoints.add(mnt_id)
+
                 yield task, mount, mnt_ns_id
 
     def _generator(
@@ -171,13 +180,26 @@ class MountInfo(plugins.PluginInterface):
         tasks: Iterable[interfaces.objects.ObjectInterface],
         mnt_ns_ids: List[int],
         mount_format: bool,
-        per_namespace: bool,
+        filtered_by_pids: bool,
     ) -> Iterable[Tuple[int, Tuple]]:
-        for task, mnt, mnt_ns_id in self._get_tasks_mountpoints(tasks, per_namespace):
-            if mnt_ns_ids and mnt_ns_id not in mnt_ns_ids:
+        warning_shown = False
+        for task, mnt, mnt_ns_id in self._get_tasks_mountpoints(tasks, filtered_by_pids):
+            if (
+                not warning_shown and
+                mnt_ns_ids and
+                isinstance(mnt_ns_id, renderers.NotAvailableValue)
+            ):
+                vollog.warning("Cannot filter by namespace id, it is not available in this kernel.")
+                warning_shown = True
+
+            if (
+                not isinstance(mnt_ns_id, renderers.NotAvailableValue) and
+                mnt_ns_ids and
+                mnt_ns_id not in mnt_ns_ids
+            ):
                 continue
 
-            mnt_info = MountInfo.get_mountinfo(mnt, task, self.context)
+            mnt_info = self.get_mountinfo(mnt, task)
             if mnt_info is None:
                 continue
 
@@ -212,7 +234,7 @@ class MountInfo(plugins.PluginInterface):
                 ]
 
             fields_values = [mnt_ns_id]
-            if not per_namespace:
+            if filtered_by_pids:
                 fields_values.append(task.pid)
             fields_values.extend(extra_fields_values)
 
@@ -228,14 +250,14 @@ class MountInfo(plugins.PluginInterface):
             self.context, self.config["kernel"], filter_func=pid_filter
         )
 
-        columns = [("MNT_NS_ID", int)]
+        columns = [("MNT_NS_ID", str)]
         # The PID column does not make sense when a PID filter is not specified. In that case, the default behavior is
         # to displays the mountpoints per namespace.
         if pids:
             columns.append(("PID", int))
-            per_namespace = False
+            filtered_by_pids = True
         else:
-            per_namespace = True
+            filtered_by_pids = False
 
         if self.config.get("mount-format"):
             extra_columns = [
@@ -262,5 +284,5 @@ class MountInfo(plugins.PluginInterface):
         columns.extend(extra_columns)
 
         return renderers.TreeGrid(
-            columns, self._generator(tasks, mount_ns_ids, mount_format, per_namespace)
+            columns, self._generator(tasks, mount_ns_ids, mount_format, filtered_by_pids)
         )
--- a/volatility3/framework/plugins/linux/mountinfo.py
+++ b/volatility3/framework/plugins/linux/mountinfo.py
@@ -50,6 +50,9 @@ class MountInfo(plugins.PluginInterface):
             requirements.PluginRequirement(
                 name="pslist", plugin=pslist.PsList, version=(2, 0, 0)
             ),
+            requirements.VersionRequirement(
+                name="linuxutils", component=linux.LinuxUtilities, version=(2, 1, 0)
+            ),
             requirements.ListRequirement(
                 name="pids",
                 description="Filter on specific process IDs.",
@@ -86,7 +89,7 @@ class MountInfo(plugins.PluginInterface):
         if not mnt_root:
             return None
 
-        path_root = linux.LinuxUtilities._get_path_mnt(task, mnt)
+        path_root = linux.LinuxUtilities.get_path_mnt(task, mnt)
         if not path_root:
             return None
 
--- a/volatility3/framework/plugins/linux/mountinfo.py
+++ b/volatility3/framework/plugins/linux/mountinfo.py
@@ -160,7 +160,7 @@ class MountInfo(plugins.PluginInterface):
 
             mnt_namespace = task.nsproxy.mnt_ns
             try:
-                mnt_ns_id = str(mnt_namespace.get_inode())
+                mnt_ns_id = mnt_namespace.get_inode()
             except AttributeError:
                 mnt_ns_id = renderers.NotAvailableValue()
 
@@ -256,7 +256,7 @@ class MountInfo(plugins.PluginInterface):
             self.context, self.config["kernel"], filter_func=pid_filter
         )
 
-        columns = [("MNT_NS_ID", str)]
+        columns = [("MNT_NS_ID", int)]
         # The PID column does not make sense when a PID filter is not specified. In that case, the default behavior is
         # to displays the mountpoints per namespace.
         if pids:
--- a/volatility3/framework/plugins/linux/mountinfo.py
+++ b/volatility3/framework/plugins/linux/mountinfo.py
@@ -141,16 +141,18 @@ class MountInfo(plugins.PluginInterface):
         )
 
     def _get_tasks_mountpoints(
-        self, tasks: Iterable[interfaces.objects.ObjectInterface], filtered_by_pids: bool
+        self,
+        tasks: Iterable[interfaces.objects.ObjectInterface],
+        filtered_by_pids: bool,
     ):
         seen_mountpoints = set()
         for task in tasks:
             if not (
-                task and
-                task.fs and
-                task.fs.root and
-                task.nsproxy and
-                task.nsproxy.mnt_ns
+                task
+                and task.fs
+                and task.fs.root
+                and task.nsproxy
+                and task.nsproxy.mnt_ns
             ):
                 # This task doesn't have all the information required.
                 # It should be a kernel < 2.6.30
@@ -183,19 +185,23 @@ class MountInfo(plugins.PluginInterface):
         filtered_by_pids: bool,
     ) -> Iterable[Tuple[int, Tuple]]:
         warning_shown = False
-        for task, mnt, mnt_ns_id in self._get_tasks_mountpoints(tasks, filtered_by_pids):
+        for task, mnt, mnt_ns_id in self._get_tasks_mountpoints(
+            tasks, filtered_by_pids
+        ):
             if (
-                not warning_shown and
-                mnt_ns_ids and
-                isinstance(mnt_ns_id, renderers.NotAvailableValue)
+                not warning_shown
+                and mnt_ns_ids
+                and isinstance(mnt_ns_id, renderers.NotAvailableValue)
             ):
-                vollog.warning("Cannot filter by namespace id, it is not available in this kernel.")
+                vollog.warning(
+                    "Cannot filter by namespace id, it is not available in this kernel."
+                )
                 warning_shown = True
 
             if (
-                not isinstance(mnt_ns_id, renderers.NotAvailableValue) and
-                mnt_ns_ids and
-                mnt_ns_id not in mnt_ns_ids
+                not isinstance(mnt_ns_id, renderers.NotAvailableValue)
+                and mnt_ns_ids
+                and mnt_ns_id not in mnt_ns_ids
             ):
                 continue
 
@@ -284,5 +290,6 @@ class MountInfo(plugins.PluginInterface):
         columns.extend(extra_columns)
 
         return renderers.TreeGrid(
-            columns, self._generator(tasks, mount_ns_ids, mount_format, filtered_by_pids)
+            columns,
+            self._generator(tasks, mount_ns_ids, mount_format, filtered_by_pids),
         )
--- a/volatility3/framework/plugins/linux/mountinfo.py
+++ b/volatility3/framework/plugins/linux/mountinfo.py
@@ -74,7 +74,7 @@ class MountInfo(plugins.PluginInterface):
 
     @classmethod
     def get_mountinfo(
-        cls, mnt, task, context
+        cls, mnt, task
     ) -> Union[
         None, Tuple[int, int, str, str, str, List[str], List[str], str, str, List[str]]
     ]:
@@ -85,7 +85,7 @@ class MountInfo(plugins.PluginInterface):
         if not mnt_root:
             return None
 
-        path_root = linux.LinuxUtilities._get_path_root(context, mnt, task.fs.root)
+        path_root = linux.LinuxUtilities._get_path_mnt(task, mnt)
         if not path_root:
             return None
 
--- a/volatility3/framework/plugins/linux/mountinfo.py
+++ b/volatility3/framework/plugins/linux/mountinfo.py
@@ -184,19 +184,12 @@ class MountInfo(plugins.PluginInterface):
         mount_format: bool,
         filtered_by_pids: bool,
     ) -> Iterable[Tuple[int, Tuple]]:
-        warning_shown = False
+        show_filter_warning = False
         for task, mnt, mnt_ns_id in self._get_tasks_mountpoints(
             tasks, filtered_by_pids
         ):
-            if (
-                not warning_shown
-                and mnt_ns_ids
-                and isinstance(mnt_ns_id, renderers.NotAvailableValue)
-            ):
-                vollog.warning(
-                    "Cannot filter by namespace id, it is not available in this kernel."
-                )
-                warning_shown = True
+            if mnt_ns_ids and isinstance(mnt_ns_id, renderers.NotAvailableValue):
+                show_filter_warning = True
 
             if (
                 not isinstance(mnt_ns_id, renderers.NotAvailableValue)
@@ -245,6 +238,11 @@ class MountInfo(plugins.PluginInterface):
             fields_values.extend(extra_fields_values)
 
             yield (0, fields_values)
+
+        if show_filter_warning:
+            vollog.warning(
+                "Could not filter by mount namespace id. This field is not available in this kernel."
+            )
 
     def run(self):
         pids = self.config.get("pids")
--- a/volatility3/framework/plugins/linux/mountinfo.py
+++ b/volatility3/framework/plugins/linux/mountinfo.py
@@ -9,6 +9,7 @@ from typing import Tuple, List, Iterable, Union
 from volatility3.framework import renderers, interfaces
 from volatility3.framework.configuration import requirements
 from volatility3.framework.interfaces import plugins
+from volatility3.framework.symbols import linux
 from volatility3.plugins.linux import pslist
 
 vollog = logging.getLogger(__name__)
@@ -72,39 +73,8 @@ class MountInfo(plugins.PluginInterface):
         ]
 
     @classmethod
-    def _do_get_path(cls, mnt, fs_root) -> Union[None, str]:
-        """It mimics the Linux kernel prepend_path function."""
-        vfsmnt = mnt.mnt
-        dentry = vfsmnt.get_mnt_root()
-
-        path_reversed = []
-        while dentry != fs_root.dentry or vfsmnt.vol.offset != fs_root.mnt:
-            if dentry == vfsmnt.get_mnt_root() or dentry.is_root():
-                parent = mnt.get_mnt_parent().dereference()
-                # Escaped?
-                if dentry != vfsmnt.get_mnt_root():
-                    return None
-
-                # Global root?
-                if mnt.vol.offset != parent.vol.offset:
-                    dentry = mnt.get_mnt_mountpoint()
-                    mnt = parent
-                    vfsmnt = mnt.mnt
-                    continue
-
-                return None
-
-            parent = dentry.d_parent
-            dname = dentry.d_name.name_as_str()
-            path_reversed.append(dname.strip("/"))
-            dentry = parent
-
-        path = "/" + "/".join(reversed(path_reversed))
-        return path
-
-    @classmethod
     def get_mountinfo(
-        cls, mnt, task
+        cls, mnt, task, context
     ) -> Union[
         None, Tuple[int, int, str, str, str, List[str], List[str], str, str, List[str]]
     ]:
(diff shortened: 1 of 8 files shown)


gcmoreira:
Hey @ikelos , I wonder why in the tests it cannot find the type `mount`, it should be there. Also, it would be good to know in which line/context it's failing.

----------------------------- Captured stdout call -----------------------------
python3 vol.py --single-location linux-sample-1.bin -q linux.lsof.Lsof
stdout:
b'Volatility 3 Framework 2.4.2\n\nPID\tProcess\tFD\tPath\n\n\n'
stderr:
b'Volatility experienced a symbol-related issue:\nsymbol_table_name1!mount: Unknown symbol: mount\n\n\t* An invalid symbol table\n\t* A plugin requesting a bad symbol\n\t* A plugin requesting a symbol from the wrong table\n\nNo further results will be produced\n'

ikelos:
Hmmm, getting more information out of the tests is probably something to check with @npetroni, I think?  We could run the tests with `-vvvvvvv`, but I don't know if that will monkey with any of the expected output or whether they all only read stdout?

gcmoreira:
> Hmmm, getting more information out of the tests is probably something to check with @npetroni, I think? We could run the tests with `-vvvvvvv`, but I don't know if that will monkey with any of the expected output or whether they all only read stdout?

Yeah not sure if that will help. I was thinking how we can make it more verbose so that the contributor knows where is the issue and fix it without having to have this kind of conversation here .

eve-mem:
I don't know if it's the correct thing to do, but you can download the mem dump that's being used to check from here: https://downloads.volatilityfoundation.org/volatility3/images/linux-sample-1.bin.gz (it's shown in the testing logs)

I think if you read the testing logs you can also get the symbols. 

That way you can tinker with it locally. 


Edit: I'm running this against all my test images, if there are any issues I'll let you know and can share the images + symbols.

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -207,7 +177,7 @@ def _generator(
             if mnt_ns_ids and mnt_ns_id not in mnt_ns_ids:
                 continue
 
-            mnt_info = self.get_mountinfo(mnt, task)
+            mnt_info = MountInfo.get_mountinfo(mnt, task, self.context)
```
ikelos: Since we're in the same class, it's fine if this is just `self.`  5;)

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -71,40 +72,9 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
-            path_reversed.append(dname.strip("/"))
-            dentry = parent
-
-        path = "/" + "/".join(reversed(path_reversed))
-        return path
-
     @classmethod
     def get_mountinfo(
-        cls, mnt, task
+        cls, mnt, task, context
```
ikelos: This is technically a change of the API (a non-backwards compatible one), so it needs a bump of the version number to `(2, 0, 0)` and any plugins that make use of it will therefore need their `VersionRequirements` on this bumping too.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -59,83 +59,92 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
 
     framework.require_interface_version(*_required_framework_version)
 
-    # based on __d_path from the Linux kernel
     @classmethod
-    def _do_get_path(cls, rdentry, rmnt, dentry, vfsmnt) -> str:
-        ret_path: List[str] = []
+    def _get_path_file(cls, context, task, filp) -> str:
```
ikelos: Since we've removed `_do_get_path` this should technically involve a full version bump (to `(3,0,0)`) and associated plugins that use it.  Since it's as internal function, I believe it shouldn't have been called anywhere else, so we can forego it which will save a lot of plugins potentially breaking.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -59,83 +59,92 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
...
 
-    # method used by 'older' kernels
-    # TODO: lookup when dentry_operations->d_name was merged into the mainline kernel for exact version
     @classmethod
-    def _get_path_file(cls, task, filp) -> str:
-        rdentry = task.fs.get_root_dentry()
-        rmnt = task.fs.get_root_mnt()
-        dentry = filp.get_dentry()
-        vfsmnt = filp.get_vfsmnt()
+    def do_get_path(cls, rdentry, rmnt, dentry, vfsmnt, context) -> Union[None, str]:
```
ikelos: However, the introduction of `do_get_path` does require a version bump of at least the middle value (so to `(2, 1, 0)`) and any plugins that use this particular method will need to depend on at least that version number.

volatility3/framework/plugins/linux/mountinfo.py
ikelos: It's also a little odd that the context is 3rd, when normally the context would be the first variable of a method.  Given the version number we don't technically need to worry about keeping the existing calling convention, which would fail without the new parameter anyway...

volatility3/framework/symbols/linux/__init__.py
ikelos: Also, it'd be awesome to have docstrings for these, even though they're private functions.  Not a showstopper though, just a nice to have...  5;)

volatility3/framework/symbols/linux/__init__.py
ikelos: Also, our convention has typically been to put context as the very first parameter to a method like this?  Since we don't have any backwards compatibility and we're bumping the version number anyway, it'd be good to fix it now so we don't scratch our heads as to how it came to be down the line please...  5:)

ikelos (commented):
Looks good, thanks!  I'm looking forward to getting this all straightened out so everything works as expected.  There's just a few version bumps needed, and some reordering of parameters to match the rest of the codebase, but otherwise looks great!  5:D

eve-mem:
It looks like b2d33c2cb0cf8535647f77feac35c44bc124ace5 works well and fixes the duplicate path problems as per issue https://github.com/volatilityfoundation/volatility3/issues/930. <3 

There are problems with 6.1+ kernels as expected due to maple tree in the mm_struct (https://github.com/volatilityfoundation/volatility3/issues/909)

I did get some failures on my test set, not sure how critical they are, I know I would personally love to see the old `LinuxUtilities._do_get_path` gone as it makes analysis harder.

Here is the full set of logs from my testing (can share samples + symbols): 
[b2d33c2cb0cf8535647f77feac35c44bc124ace5-testing.zip](https://github.com/volatilityfoundation/volatility3/files/11330512/b2d33c2cb0cf8535647f77feac35c44bc124ace5-testing.zip)


High level these are the issues that popped up:
### sockstat
v3.7.10.elf.gz-sockstat.txt 
```
AttributeError: StructType has no attribute: symbol_table_name1!net.ns
```

v3.17.8.elf.gz-sockstat.txt, v3.18.140.elf.gz-sockstat.txt, v3.19.8.elf.gz-sockstat.txt, v4.0.9.elf.gz-sockstat.txt 
```
AttributeError: StructType has no attribute: symbol_table_name1!bpf_prog.type
```

### mountinfo
v3.7.10.elf.gz-mountinfo.txt 
```
AttributeError: Unable to find mnt_namespace inode
```

gcmoreira:
@ikelos Please have a look now. Wow, it looks like it passed all the testcases, except the linter errr well let's talk about that later not sure about some of the suggestions but happy with massacre ... eh sorry ;) ... I mean happy with changing the code if you feel it is necessary.

As you will find in the commit comments, several issues were fixed in mountinfo adding also support for older kernels.  Also, as other modules use the mountinfo related functions to resolve the paths, unfortunately, the scope goes beyond it, affecting lsof, proc_maps, etc.
Also, I included in this PR other fixes and improvements to the sockstat plugin. It was necessary because some of the testcases were failing there after my previous submission. However, the places were it was failing was not related with the changes I submitted. My theory is that the testcases would have failed if they were executed with the pre-existent code that was already in 'develop' anyway, but not 100%. Could it be? The same happen with the linter suggestion. Some of them are no part of my changes, it's code that was already there.

All in all, you will expect even more mountpoints being extracted which were previously skipped due to other errors. The double parent issues with the paths was also fixed. Finally, support for older kernels in both sockstat and mountinfo plugins was added.

It's still pending the linter suggestions and version updates, could you please give me a hand with which numbers to use in each case?

@eve-mem Thanks for the exhaustive testing. I really liked the kernel version coverage, impressive mate. It would be really good if you can run it again

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -258,14 +250,14 @@ def run(self):
             self.context, self.config["kernel"], filter_func=pid_filter
         )
 
-        columns = [("MNT_NS_ID", int)]
+        columns = [("MNT_NS_ID", str)]
```
ikelos: Why is this a string?  If the value's an integer, it's better to leave the data as an integer and then mark the column that way.  It might affect column sorting or other things in the future...

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -170,41 +141,62 @@ def get_mountinfo(
...
             mnt_namespace = task.nsproxy.mnt_ns
-            mnt_ns_id = mnt_namespace.get_inode()
-
-            if per_namespace:
-                if mnt_ns_id in seen_namespaces:
-                    continue
-                else:
-                    seen_namespaces.add(mnt_ns_id)
+            try:
+                mnt_ns_id = str(mnt_namespace.get_inode())
```
ikelos: Does this need explicitly casting as a string?  It seems it's been added in intentionally, but there's no indication of why that I can see from the code?

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -170,41 +141,62 @@ def get_mountinfo(
...
         mnt_ns_ids: List[int],
         mount_format: bool,
-        per_namespace: bool,
+        filtered_by_pids: bool,
     ) -> Iterable[Tuple[int, Tuple]]:
-        for task, mnt, mnt_ns_id in self._get_tasks_mountpoints(tasks, per_namespace):
-            if mnt_ns_ids and mnt_ns_id not in mnt_ns_ids:
+        warning_shown = False
+        for task, mnt, mnt_ns_id in self._get_tasks_mountpoints(tasks, filtered_by_pids):
+            if (
```
ikelos: This feels like a lot more lines than a main `if` guard testing what needs testing, and then an inner guard printing and setting a flag not to print again about the warning message?  Duplicating `if` clauses often leads to desynchronization...

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -258,14 +250,14 @@ def run(self):
             self.context, self.config["kernel"], filter_func=pid_filter
         )
 
-        columns = [("MNT_NS_ID", int)]
+        columns = [("MNT_NS_ID", str)]
         # The PID column does not make sense when a PID filter is not specified. In that case, the default behavior is
         # to displays the mountpoints per namespace.
         if pids:
             columns.append(("PID", int))
```
ikelos, line 264: Not sure if I mentioned this previously, but modifying the columns returned by a plugin will make it harder to automation to handle gracefully.  Given the column name is only 3 characters, and a NotApplicable absent value will likely get represented by a `-`, it shouldn't add too much to the output to leave it in there?  I'm trying to get people to think of these as data output rather than human read output.  It's not a showstopper, but if you'd be ok changing it, I think it might also make a lot of the code easier, since you won't have to count column numbers anywhere, just always include it.

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -170,41 +141,62 @@ def get_mountinfo(
...
+                mnt_ns_id = str(mnt_namespace.get_inode())
+            except AttributeError:
+                mnt_ns_id = renderers.NotAvailableValue()
 
             for mount in mnt_namespace.get_mount_points():
+                # When PIDs are filtered, it makes sense that the user want to
+                # see each of those processes mount points. So we don't filter
+                # by mount id in this case.
+                if not filtered_by_pids:
+                    mnt_id = int(mount.mnt_id)
```
ikelos, line 175: Will `mnt_id` ever not be an `int`?

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -348,3 +375,30 @@ def container_of(
...
+        Raises:
+            ValueError: If it cannot obtain any module from the symbol table
+
+        Returns:
+            volatility3.framework.contexts.Module: A kernel object (vmlinux)
+        """
+        symbol_table_arr = volobj.vol.type_name.split("!", 1)
+        symbol_table = symbol_table_arr[0] if len(symbol_table_arr) == 2 else None
+
+        module_names = volobj._context.modules.get_modules_by_symbol_tables(symbol_table)
```
ikelos: `._context` is essentially private to whatever volobj is.  This is why context has always been passed in, in order to make sure that the context to be used is explicit.  It would set a bad precedent to just start accessing attributes preceded by `_`.  That an unenforced convention within python that the value is private to its class and descendants.  We *could* expose the context for objects as part of the interface, but the reason for not doing that is that it would be very difficult to make that context readonly, and if people think they've got a context, they might absentmindedly modify it, and then wonder why some objects pick up/reread from the context and others don't.

It was also previously explicit to make people consider where exactly the context came from.  Whilst it's somewhat expected that most runs of a plugin will only use one context, that's not strictly required, and when plugins start running sub plugins, for example, they may end up using different contexts.  As such, this is kind of our check.  Are you sure that the context on which the object was made is the context to be looking within?

Unfortunately this is a little bit of a pandoras box which is why, and I do know it's frustrating, the context was always very obviously passed in, so everyone knew what was going on and what came from where.  I'll have a think about the implications of allowing access to the context outside of the object (technically it's just a reference, so if the original context gets changed, they all get changed anyway), but it worries me so I'll need to think through the ramifications.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -348,3 +375,30 @@ def container_of(
         return vmlinux.object(
             object_type=type_name, offset=container_addr, absolute=True
         )
+
+    @classmethod
+    def get_vmlinux_from_volobj(cls, volobj):
```
ikelos: I understand the intention, but this isn't accurately named.  The return value (with a type annotation if possible), appears to be the module the object's type belongs to.  At the moment, we don't really use application specific symbol tables, but the hope is we will one day, and so for a piece of the core API, this really needs to be named properly from the get go.

ikelos (commented):
Generally looks very good, with docstrings, and very clean... but... suspiciously clean and then I realized it's because the `context` has been unthreaded from all the calls and extracted from the object in question.  As you'll see in the comments, this sadly isn't as straightforward a solution as it may appear, and it's going to take me a bit of time to consider, along with support/review from @npetroni to make sure I'm not a) rushing in or b) being a killjoy unreasonably)  5;P

We've got a meeting of the core team coming up soon (or should do), so I'll try and get people primed and ready to discuss it for then, but it means this can't go through immediately, unfortunately.  You can either wait for the review, or re-thread the context value back through all the functions...

The main other comment was about casting only when it's strictly necessary, but as I say, on the whole very nice, very clean and well documented!  Thanks!  5:D

ikelos:
Also, just had a look at the linter output.  All nitpicky stuff, but we use black to ensure consistency across the codebase, which is why it's run on every PR.  I have my own style (and we use to use a yapf template based on it).  I find it easier to read function parameters with a space between the equals, but yapf would often change its mind and result in multiple recommits when it calculated spacing and line length differently.  As such, we've chosen black and we're trying really hard to stick with it.  It's not perfect and I know some of the things it does seem annoying when they're not how you'd normally do them.  I know some of our devs used to uses spaces for alignment and black blows all that away, but we would appreciate it if you could clear up the issues so the tests pass...

eve-mem:
@gcmoreira testing against my sample set gives me the same results as before. So apart from those particular versions failing it looks like the bulk of it has been fixed. 

Happy to test again when the issues @ikelos has raised have been fixed, and also happy to share the test set if anyone is interested.

gcmoreira:
> @gcmoreira testing against my sample set gives me the same results as before. So apart from those particular versions failing it looks like the bulk of it has been fixed. 
> 
> Happy to test again when the issues @ikelos has raised have been fixed, and also happy to share the test set if anyone is interested. 

@eve-mem Thanks for testing it. Sorry but I didn't get that .. So, if the bulk of it was fixed, it's not the same results as before, I'm confused ... Have you seen it failing with some specific kernel versions or any of the plugins are failing?  Yes, could you please provide the test results to have a look

eve-mem:
@gcmoreira sorry, very badly explained. 

- It fixes the path issue as before, no duplicate paths. Eg. The main reason for this pull request. 

- It has the same errors as before, the same missing symbols etc as i said in the previous comment (https://github.com/volatilityfoundation/volatility3/pull/942#issuecomment-1522967256). E.g. the errors i highlighted in my previous round of testing.  So no changes on that front. It may way be an issue with my samples, I'm happy to share the samples or run some debugging for you if its helpful. 

Here are the logs: 
[a09f77897b24b4b0da06a8b45a353ee730cadd08-testing.zip](https://github.com/volatilityfoundation/volatility3/files/11417978/a09f77897b24b4b0da06a8b45a353ee730cadd08-testing.zip)

volatility3/framework/plugins/linux/mountinfo.py
gcmoreira: Yeah, all the changes above from int() to str() is to take of the NotApplicable formatting to "-". Otherwise, I think I got another exception when the value is an instance of NotApplicable and the column is an int.

gcmoreira:
@eve-mem Only 1/3 set of errors you pointed out in https://github.com/volatilityfoundation/volatility3/pull/942#issuecomment-1522967256 was still failing. The others two were already fixed and didn't show  up in  [a09f77897b24b4b0da06a8b45a353ee730cadd08-testing.zip](https://github.com/volatilityfoundation/volatility3/files/11417978/a09f77897b24b4b0da06a8b45a353ee730cadd08-testing.zip)

So, regarding this:
```
v3.17.8.elf.gz-sockstat.txt, v3.18.140.elf.gz-sockstat.txt, v3.19.8.elf.gz-sockstat.txt, v4.0.9.elf.gz-sockstat.txt
> AttributeError: StructType has no attribute: symbol_table_name1!bpf_prog.type
```

What happened is that the BPF program type had a short life in `bpf_prog_aux::prog_type` in 3.18.140 <= kernels < 4.1.52, and then it was moved and renamed to `bpf_prog::type` in 4.1.52. Also, before  3.18.140 it didn't exist, that explains why it was also failing in 3.17.8.

eve-mem:
@gcmoreira - i must have got my greps wrong, sorry for that. (and why isn't important for me to share the actual logs!) Glad you spotted that. 

Im running your latest commits over it all again. If you have any particular versions youd like to see tested I can add them fairly easily? 

(Assuming its useful, i dont have to test them at all if youd rather not)

gcmoreira:
@eve-mem that coverage is good. We have a gap between kernels >=2.6.30 and the lowest you have... vanilla, but also Red Hat based kernels. They do backports which are often problematic.

eve-mem:
@gcmoreira Happy to say there are zero crashes with the new commits. (I tested [7d65c20cc0b971c862889065d8a90d98e24819c9](https://github.com/volatilityfoundation/volatility3/pull/942/commits/7d65c20cc0b971c862889065d8a90d98e24819c9))

Results for completeness: 
[7d65c20cc0b971c862889065d8a90d98e24819c9-testing.zip](https://github.com/volatilityfoundation/volatility3/files/11423044/7d65c20cc0b971c862889065d8a90d98e24819c9-testing.zip)

gcmoreira:
@ikelos ok, I rolled back the changes related to the context. There still one place where it uses `self._context` but if you check the original version of `_get_real_mnt()` it was already like that. I only reuse the new helper functions to avoid repeated code.

Also, note that `get_vmlinux_from_volobj()` is nothing new. It's actually a merge of what it used to be in `_get_new_sock_pipe_path()` and `_get_real_mnt()`, nothing more.

ikelos:
> @ikelos ok, I rolled back the changes related to the context. There still one place where it uses `self._context` but if you check the original version of `_get_real_mnt()` it was already like that. I only reuse the new helper functions to avoid repeated code.

Thanks very much, `self._context` is fine, because objects are allowed to look at their own contexts, they just shouldn't snoop on `anotherobject._context`.

I absolutely want to use avoid repeated code where possible, so I'm happy to split repeated code out, or having something like `get_vmlinux_from_volobj()` as long as it doesn't access any private members (so parameterizing it to accept a `context` would be fine).  5:)   Just keen to make sure it's named appropriately, so if it doesn't always return `vmlinux` but a module, then let's name it `get_module_from_volobj_type()` or something like that?  5:)

volatility3/framework/plugins/linux/mountinfo.py
ikelos: Columns in the output should always accept anything derived from `BaseAbsentValue`.  If you have a situation where that goes wrong, I'd sooner get that fixed than do casting everywhere.  The point of the `BaseAbsentValue` derived classes is to allow whatever medium is display the results to display then in a way that might make more sense (so a web UI might want to highlight them in red for example).  Not being able to distinguish between an empty value and one that wasn't available for some reason will make trickier...

gcmoreira:
@ikelos I took the liberty of marking some of the threads as resolved to make it clear what's done and what's missing. I hope you don't mind. 

If the tests pass with the latest commit, and you like how I implemented that, it looks like the only one pending is pretty much the versioning. Could you help me with that?

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -115,8 +86,8 @@ def get_mountinfo(
         if not mnt_root:
             return None
 
-        path_root = cls._do_get_path(mnt, task.fs.root)
-        if path_root is None:
+        path_root = linux.LinuxUtilities._get_path_mnt(task, mnt)
```
ikelos: Sorry, just noticed, this is a bad.  5:P  Again, it's a call to a private member, so probably this needs the `_` removing from the front of the function name, and then a `VersionRequirement` adding to plugins like this that call it, on the version that contains it.  Sorry I missed this earlier...

ikelos (commented):
No, that's fine.  I appreciate keeping the bug tidy (as long as people don't try to use it as an excuse to ignore a comment without explanation/discussion).  5;)

As to the version numbers, as best I can tell, LinuxUtilities has had its private interface changed (so `_thing` functions) and had its public interface added to.  As such this line:
https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/symbols/linux/__init__.py#L57
Can be changed from `(2,0,0)` to `(2,1,0)` because it's only an additive change.  All previous plugins that relied on the former functionality will continue to work, so they don't need any changing.  Then any plugin that uses the new public calls (specifically `do_get_path` or `get_module_from_volobj_type`) should have a `VersionRequirement` on the `LinuxUtilities` object, with a requirement of `(2,1,0)`.  That only applies to components that can have requirements, so LinuxUtilities itself (which can't have requirements) should be fine.  As such, I don't *think* anything else needs changing...

Hope that's all good, it's quite dense and there's a lot of code, so it's possible something else slipped through (and is why I prefer doing smaller reviews more often), but hopefully it's all good, so if you're happy, once these changes have gone in, we can merge it.  5:)

gcmoreira:
@ikelos thanks, so if it passes all the testcases we are ready to go.

ikelos (approved):
Yep, looks good, thanks  5:)

Issue #944, opened 2023-04-30
windows.info plugin is doubled and can't be used


rareguy:
I do not know what exactly happened, but when I was developing a plugin for volatility3, `windows.info` was suddenly doubled in plugin listing and when I tried to use it it just confused on which one to use.
![image](https://user-images.githubusercontent.com/33000326/235333078-558ce5b0-7780-4de2-9d75-b6892ef30640.png)
![image](https://user-images.githubusercontent.com/33000326/235333081-68e7abbf-a4a1-4f6b-bbd0-58761a0f0c32.png)
Any help would be appreciated. I do view the content of `windows.info` as a reference to my plugin development, but I did not change any part from it. Need to disable the double detection somehow. Thank you

ikelos:
Hi there, this is because an import check we previously had in place was broken about 10 months ago, so thank you for spotting that!

The issue is that `volatility3.framework.plugins` are not supposed to be imported directly, we've done work to ensure that the user location for plugins (`volatility3.plugins`) includes and overrides the framework plugins, this way users can modify the core plugins without needed to change the original files, so that they can revert to them easily by removing their code from `volatility3.plugins`.  It turns out a fix to help more recent versions of pyinstaller inverted that logic and meant that the warning about importing such files never happened except to pyinstaller.

This has since been fixed, and you should receive a warning if you try to run with an `import volatility3.framework.plugins` in your code anywhere.  Please always uses `volatility3.plugins`, it will include everything from the `volatility3.framework.plugins` namespace, but will prioritize files you put in the volatility3/plugins directory.

Thanks again for identifying the issue (and updating the description!).  5;)

rareguy:
I see, so to clarify, this behavior only applies on `volatility3.plugins`? I see the `volatility3.framework.symbols` still works fine as intended.

ikelos:
I think you have that the wrong way around.  You can only import from `volatility3.plugins`, but it will automatically include everything in `volatility3.framework.plugins`, so for example to make use of `windows.info`, you would do:
```
from voalatility3.plugins.windows import info
```

PR #945, opened 2023-05-01
Add: help comment for windows plugins (`windows.crashinfo`, `windows.ldrmodules`, `windows.statistics`).

digitalisx:
## Description
Hello :), I looked at this issue (#944) and found out that the current volatility version don't provide a description of some plugins (`windows.crashinfo`, `windows.ldrmodules`, `windows.statistics`).
We confirmed in the previous PR (#735) that the `windows.statistics` plugin is a plugin that we do not officially support.
However, aside from the functionality, I suggest that it would be good to modify this for unity.
So I added missing descriptions to this.


ikelos (approved):
Looks good, thanks!  5:)

PR #948, opened 2023-05-04
Added modules name flag

paulkermann:
Hey,
I added a name flag to the windows.modules plugin.

### DIFF
--- a/volatility3/framework/plugins/windows/modules.py
+++ b/volatility3/framework/plugins/windows/modules.py
@@ -42,7 +42,12 @@ class Modules(interfaces.plugins.PluginInterface):
                 default=False,
                 optional=True,
             ),
-            requirements.StringRequirement(name="name", description="module name/sub string", optional=True, default=""),
+            requirements.StringRequirement(
+                name="name",
+                description="module name/sub string",
+                optional=True,
+                default="",
+            ),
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/modules.py
```diff
@@ -42,6 +42,12 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                 default=False,
                 optional=True,
             ),
+            requirements.StringRequirement(
+                name="name",
+                description="module name/sub string",
+                optional=True,
+                default="",
```
ikelos: Could you please use `None` rather than `""`?  They should both fail the check, but it's then easier to differentiate if the user provided a value or not

ikelos (commented):
Hiya, looks all good.  I'd prefer a tiny change, but it's not a show stopper, so just let me know if you're going to/when you've made the fix and I can get it merged.  Thanks!  5:)

volatility3/framework/plugins/windows/modules.py
paulkermann: Fixed :)

ikelos:
Thanks, for fixing it up so quickly  5:)

PR #951, opened 2023-05-07
fix deployment of tests

### LINKED ISSUE #950: packaging issue: deployment of tests
**Describe the bug**

The issue is that the setup.py is deploying the `tests` folder under the root python path and not under this package one.

So it's end up under `/usr/lib/python3.11/site-packages/tests` rather than `/usr/lib/python3.11/site-packages/[volatility3](https://github.com/volatilityfoundation/volatility3)/tests` and so conflicts with other packages having the same issue.

Anyway usually test are not shipped in a release package so the easiest would just to remove them. Else they should be deployed in the children directory. 

It's explained in ArchLinux packaging guidelines for Python: https://wiki.archlinux.org/title/Python_package_guidelines#Test_directory_in_site-package

**Context**
Volatility Version:  N/A
Operating System:  N/A
Python Version:  N/A
Suspected Operating System:  N/A
Command:  N/A

**To Reproduce**
Steps to reproduce the behavior: N/A

**Expected behavior**

No files deployed on python root path.

**Example output**
N/A

**Additional information**
N/A

noraj:
fix #950

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setuptools.setup(
     include_package_data=True,
     exclude_package_data={"": ["development", "development.*"], "development": ["*"]},
     packages=setuptools.find_namespace_packages(
-        exclude=["development", "development.*"]
+        exclude=["development", "development.*", "tests*"]
     ),
     entry_points={
         "console_scripts": [
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setuptools.setup(
     include_package_data=True,
     exclude_package_data={"": ["development", "development.*"], "development": ["*"]},
     packages=setuptools.find_namespace_packages(
-        exclude=["development", "development.*", "tests*"]
+        exclude=["development", "development.*", "test*"]
     ),
     entry_points={
         "console_scripts": [


setup.py
```diff
@@ -40,7 +40,7 @@ def get_install_requires():
     include_package_data=True,
     exclude_package_data={"": ["development", "development.*"], "development": ["*"]},
     packages=setuptools.find_namespace_packages(
-        exclude=["development", "development.*"]
+        exclude=["development", "development.*", "tests*"]
```
ikelos: Should this be `tests*` or `test*`?  I don't think we have a tests directory?

ikelos (approved):
Thanks very much, I'll be happy to apply it, but I just wanted to check if there was a typo in it?

ikelos:
It seems as though doc and build may suffer from the same issue?

setup.py
noraj: You are right, the folder is test

noraj:
> It seems as though doc and build may suffer from the same issue?

Yes it impacts doc folder too.

image

ikelos:
Since this hits docs and other directories, it might be better to convert this to an "explicit include only" thing, where only the volatility3 package is included, rather than saying "everything except"?  What do you think?

noraj:
Yes allow lists are always better to avoid including stuff by error but I don't know if it will require an annoyingly long list of files to include.

ikelos:
Sorry for the delay.  No, I think it just needs the volatility3 directory actually?   That should be it?  Rather a long list of things to exclude?  We can get this merged if it's flipped to directories that should be included rather than defaulting to including everything and then a list of directories to exclude?

noraj:
Sounds good to me

ikelos:
Cool, soooooo... are you gonna change that in the request, or should I close it and just make the changes myself?  Happy to do either depending on what you'd like...  5:)

noraj:
@ikelos Maybe it's better if you try yourself. I guess the include list will maybe require trial and error as only including the [volatility3](https://github.com/volatilityfoundation/volatility3/tree/develop/volatility3) folder may not work. You have a better understanding of the software architecture so you'll probably better think of what need to be included.

ikelos:
My attempt failed and broke things (see #1002).  I'm afraid it's very low priority given how delicate python's setup.py installation mechanism appears to be.  You can see my attempt in #1003, feel free to add comments or help me work on it there.  The concept was to use `find_namespace_packages` but that attempt failed, because `where` tells it which directory to look in (so everything is one directory too high), and just using `includes` failed.

noraj:
I see you finally solved it with `include=["volatility3", "volatility3.*"]`.

ikelos:
Yep, this should require less updating in the future and captures everything we wanted.  Just took way too long to figure it which option needed setting and which didn't... 5:S

Thanks for pointing it out and the proposed patch though!  5:)

PR #957, opened 2023-05-17
Linux capabilities plugin

gcmoreira:
This pull request adds support to process capabilities for Linux.
 
From man(7) capabilities:
 _Starting with kernel 2.2, Linux divides the privileges traditionally associated with superuser into distinct units, known as capabilities, which can be independently enabled and disabled.  Capabilities are a per-thread attribute._

This also means that even though a task is running with a user other than **root**, it can still have elevated privileges for specific operations, or even the same privileges as root.

This pull request exposes a blind spot that could be exploited by threat actors to gain elevated privileges or to maintain access to a system. 

Example 1:
```shell
$ id
uid=1000(vagrant) gid=1000(vagrant) groups=1000(vagrant)

$ cat /etc/shadow
cat: /etc/shadow: Permission denied

$ whereis python
python: /usr/bin/python3.8 /usr/lib/python3.9 /usr/lib/python2.7 /usr/lib/python3.8 /etc/python3.8 /usr/local/lib/python3.8

$ getcap /usr/bin/python3.8

# Suppose that the attacker finds a misconfigured executable, or somehow manages to change the capabilities of an executable using maybe another method of exploitation to, for instance, persist the attack.
$ sudo /usr/sbin/setcap cap_chown=ep /usr/bin/python3.8

$ getcap /usr/bin/python3.8
/usr/bin/python3.8 = cap_chown+ep

$ ls -la /usr/bin/python3.8
-rwxr-xr-x 1 root root 5490488 Nov 26  2021 /usr/bin/python3.8

$ python3.8 -c 'import os;os.chown("/etc/shadow", 1000, 1000);print(open("/etc/shadow").read())'
root:*:19005:0:99999:7:::
daemon:*:19005:0:99999:7:::
...
pollinate:*:19005:0:99999:7:::
vagrant:$6$FiZBMW30DrzV0C2$ha6k/LJjknzyqgJhsbSbS5w8r1DQ1NZ2TyppU5WIyZ0/k/Xrb593I04aItcIx02OqeZXQea/rFOpaOlFsGGzq.:19005:0:99999:7:::
...
```
Example 2: Let's see how to get these capabilities from the memory forensics perspective.
```shell
$ python3.8 
Python 3.8.10 (default, Nov 26 2021, 20:14:08) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.getpid()
1738
>>>
..
# In a new terminal
$ grep Cap /proc/1738/status
CapInh:	0000000000000000
CapPrm:	0000000000000001
CapEff:	0000000000000001
CapBnd:	000000ffffffffff
CapAmb:	0000000000000000

# From user-land, with the **capsh** tool help, we can confirm the effective capabilities of this task.
$ capsh --decode=0x0000000000000001
0x0000000000000001=cap_chown
```
Now generate a memory dump and please allow me to paste the plugin full output:
```shell
$ python3 ./vol.py -f ../ubuntu2004lts64bit.lime -r pretty linux.capabilities.Capabilities 
Volatility 3 Framework 2.4.2
  |            Name |  Tid |  Pid | PPid | EUID | cap_inheritable | cap_permitted | cap_effective | cap_bounding | cap_ambient
* |         systemd |    1 |    1 |    0 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* | systemd-journal |  456 |  456 |    1 |    0 |             0x0 |    0x402800cf |    0x402800cf |   0x402800cf |         0x0
* |   systemd-udevd |  480 |  480 |    1 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* | systemd-network |  481 |  481 |    1 |  100 |          0x3c00 |        0x3c00 |        0x3c00 |       0x3c00 |      0x3c00
* |      multipathd |  558 |  558 |    1 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* | systemd-resolve |  599 |  599 |    1 |  101 |             0x0 |           0x0 |           0x0 |          0x0 |         0x0
* | systemd-timesyn |  601 |  601 |    1 |  102 |       0x2000000 |     0x2000000 |     0x2000000 |    0x2000000 |   0x2000000
* |   VGAuthService |  605 |  605 |    1 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* |        vmtoolsd |  606 |  606 |    1 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* | accounts-daemon |  635 |  635 |    1 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* |            cron |  638 |  638 |    1 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* |     dbus-daemon |  639 |  639 |    1 |  103 |             0x0 |    0x20000000 |    0x20000000 |   0xffffffff |         0x0
* |      irqbalance |  647 |  647 |    1 |    0 |             0x0 |           0x0 |           0x0 |          0x0 |         0x0
* | networkd-dispat |  648 |  648 |    1 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* |        rsyslogd |  650 |  650 |    1 |  104 |             0x0 |           0x0 |           0x0 |   0xffffffff |         0x0
* |  systemd-logind |  655 |  655 |    1 |    0 |             0x0 |    0x4420020f |    0x4420020f |   0x4420020f |         0x0
* |         udisksd |  658 |  658 |    1 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* |             atd |  659 |  659 |    1 |    1 |             0x0 |    0xffffffff |           0x0 |   0xffffffff |         0x0
* |          agetty |  664 |  664 |    1 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* |            sshd |  685 |  685 |    1 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* |         polkitd |  694 |  694 |    1 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* | unattended-upgr |  717 |  717 |    1 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* |            sshd |  985 |  985 |  685 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* |         systemd |  988 |  988 |    1 | 1000 |             0x0 |           0x0 |           0x0 |   0xffffffff |         0x0
* |        (sd-pam) |  989 |  989 |  988 | 1000 |             0x0 |           0x0 |           0x0 |   0xffffffff |         0x0
* |            sshd | 1086 | 1086 |  985 | 1000 |             0x0 |           0x0 |           0x0 |   0xffffffff |         0x0
* |            bash | 1087 | 1087 | 1086 | 1000 |             0x0 |           0x0 |           0x0 |   0xffffffff |         0x0
* |           snapd | 1313 | 1313 |    1 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* |       python3.8 | 1738 | 1738 | 1087 | 1000 |             0x0 |           0x1 |           0x1 |   0xffffffff |         0x0
```
From the above output you can see that python3.8 (1738) is running with effective uid (EUID) 1000 and with effective and permitted "cap_chown" capabilitities. This allows this task to do the same as the example 1 above. 

The plugin allows to filter by pid but also lets to show the individual set of capabilities decoded in human-readable text.
A special word "all" is used as a summary meaning that all the capabilities are enabled. This is inspired in the libcap **cap_to_text()** function.

```shell
$ python3 ./vol.py linux.capabilities.Capabilities --help
Volatility 3 Framework 2.4.2
usage: volatility linux.capabilities.Capabilities [-h] [--pids [PIDS ...]] [--inheritable] [--permitted] [--effective] [--bounding] [--ambient]

options:
  -h, --help         show this help message and exit
  --pids [PIDS ...]  Filter on specific process IDs.
  --inheritable      Show only inheritable capabilities in human-readable strings.
  --permitted        Show only permitted capabilities in human-readable strings.
  --effective        Show only effective capabilities in human-readable strings.
  --bounding         Show only bounding capabilities in human-readable strings.
  --ambient          Show only ambient capabilities in human-readable strings.
```
For instance, to only see the **effective capabilities** decoded as strings you can do the following:
```shell
$ python3 ./vol.py -f ubuntu2004lts64bit.lime -r pretty -s ./test_linux linux.capabilities.Capabilities --effective
Volatility 3 Framework 2.4.2
  |            Name |  Tid |  Pid | PPid | EUID |                                                                                           cap_effective
* |         systemd |    1 |    1 |    0 |    0 |                                                                                                     all
* | systemd-journal |  456 |  456 |    1 |    0 |      chown, dac_override, dac_read_search, fowner, setgid, setuid, sys_ptrace, sys_admin, audit_control
* |   systemd-udevd |  480 |  480 |    1 |    0 |                                                                                                     all
* | systemd-network |  481 |  481 |    1 |  100 |                                                     net_bind_service, net_broadcast, net_admin, net_raw
* |      multipathd |  558 |  558 |    1 |    0 |                                                                                                     all
* | systemd-resolve |  599 |  599 |    1 |  101 |                                                                                                       -
* | systemd-timesyn |  601 |  601 |    1 |  102 |                                                                                                sys_time
* |   VGAuthService |  605 |  605 |    1 |    0 |                                                                                                     all
* |        vmtoolsd |  606 |  606 |    1 |    0 |                                                                                                     all
* | accounts-daemon |  635 |  635 |    1 |    0 |                                                                                                     all
* |            cron |  638 |  638 |    1 |    0 |                                                                                                     all
* |     dbus-daemon |  639 |  639 |    1 |  103 |                                                                                             audit_write
* |      irqbalance |  647 |  647 |    1 |    0 |                                                                                                       -
* | networkd-dispat |  648 |  648 |    1 |    0 |                                                                                                     all
* |        rsyslogd |  650 |  650 |    1 |  104 |                                                                                                       -
* |  systemd-logind |  655 |  655 |    1 |    0 | chown, dac_override, dac_read_search, fowner, linux_immutable, sys_admin, sys_tty_config, audit_control
* |         udisksd |  658 |  658 |    1 |    0 |                                                                                                     all
* |             atd |  659 |  659 |    1 |    1 |                                                                                                       -
* |          agetty |  664 |  664 |    1 |    0 |                                                                                                     all
* |            sshd |  685 |  685 |    1 |    0 |                                                                                                     all
* |         polkitd |  694 |  694 |    1 |    0 |                                                                                                     all
* | unattended-upgr |  717 |  717 |    1 |    0 |                                                                                                     all
* |            sshd |  985 |  985 |  685 |    0 |                                                                                                     all
* |         systemd |  988 |  988 |    1 | 1000 |                                                                                                       -
* |        (sd-pam) |  989 |  989 |  988 | 1000 |                                                                                                       -
* |            sshd | 1086 | 1086 |  985 | 1000 |                                                                                                       -
* |            bash | 1087 | 1087 | 1086 | 1000 |                                                                                                       -
* |           snapd | 1313 | 1313 |    1 |    0 |                                                                                                     all
* |       python3.8 | 1738 | 1738 | 1087 | 1000 |                                                                                                   chown
```
* Tested with kernels 2.6.32, 2.6.38, 3.0.0, 3.2.0, 3.10.0, 3.13.0, 3.16.0, 4.4.0, 4.10.0, 4.11.0, 4.13.0, 4.15.0, 4.18.0, 5.0.0, 5.3.0, 5.4.0, 5.8.0, 5.10.0, 5.13.0


eve-mem:
Hey @gcmoreira just to say I have tested this against my collection of test mem samples and did not get any errors at all. <3

gcmoreira:
> Hey @gcmoreira just to say I have tested this against my collection of test mem samples and did not get any errors at all. <3

Nice, thanks @eve-mem

volatility3/framework/constants/linux/__init__.py
```diff
@@ -234,3 +234,48 @@
     "HIDP",
     "AVDTP",
 )
+
+# Ref: include/uapi/linux/capability.h
+CAPABILITIES = (
```
ikelos, line 239: Awesome!  Thanks for splitting these off into the constants file!  5:D

volatility3/framework/plugins/linux/capabilities.py
```diff
@@ -0,0 +1,228 @@
...
+                optional=True,
+            ),
+        ]
+
+    def _check_capabilities_support(self):
+        """Checks that the framework supports at least as much capabilities as
+        the kernel being analysed. Otherwise, it shows a warning for the
+        developers.
+        """
+        vmlinux = self.context.modules[self.config["kernel"]]
```
ikelos: There's nothing technically wrong with this, but it's probably better to parameterize the kernel, so that references to `self.config` stay in the main generator/run methods.  This will make it easier to turn into a public classmethod if we want in the future and keeps it in line with most other helper functions.  5:)

volatility3/framework/plugins/linux/capabilities.py
```diff
@@ -0,0 +1,228 @@
...
+            str: A string with a comma separated list of decoded capabilities
+        """
+        if isinstance(cap, renderers.NotAvailableValue):
+            return cap
+
+        cap_value = cap.get_capabilities()
+        if cap_value == 0:
+            return "-"
+
+        CAP_FULL = 0xFFFFFFFF
```
ikelos: Could we move CAP_FULL to the constants file please?  Easier to keep things like this all in one place, even if it's incredibly unlikely to change...  5:)

volatility3/framework/plugins/linux/capabilities.py
```diff
@@ -0,0 +1,228 @@
...
+        """Returns a dict with the task basic information along with its capabilities
+
+        Args:
+            task: A task object from where to get the fields.
+
+        Returns:
+            dict: A dict with the task basic information along with its capabilities
+        """
+        task_cred = task.real_cred
+        fields = {
```
ikelos: Hmmm, this feels like quite a specific subdivision and I can't tell if it makes the most sense?  I'm wondering if actually a named_tuple would be better here?  That way new ones can be added in and potentially change the other without distrupting other stuff (as long as everything's accessed using names)?

Other options would be fine grained names, or a set of index constants (ugly) just something so that if the data returned is extended or shifts in some way in the future, we can easily expand/adapt to it...

volatility3/framework/plugins/linux/capabilities.py
```diff
@@ -0,0 +1,228 @@
...
+            "capabilities": [
+                task_cred.cap_inheritable,
+                task_cred.cap_permitted,
+                task_cred.cap_effective,
+                task_cred.cap_bset,
+            ],
+        }
+
+        # Ambient capabilities were added in kernels 4.3.6
+        if task_cred.has_member("cap_ambient"):
```
ikelos, line 155: The named_tuple idea would then also clear this up a little too?

volatility3/framework/plugins/linux/capabilities.py
```diff
@@ -0,0 +1,228 @@
...
+        """Yields a dict for each task containing the task's basic information along with its capabilities
+
+        Args:
+            tasks: An iterable with the tasks to process.
+
+        Yields:
+            Iterable[Dict]: A dict for each task containing the task's basic information along with its capabilities
+        """
+        for task in tasks:
+            if task.is_kernel_thread:
```
ikelos: Could we get a quick comment just to explain why we're filtering out kernel thread here please?

volatility3/framework/plugins/linux/capabilities.py
```diff
@@ -0,0 +1,228 @@
...
+            yield self.get_task_capabilities(task)
+
+    def _generator(
+        self, tasks: Iterable[interfaces.objects.ObjectInterface]
+    ) -> Iterable[Tuple[int, Tuple]]:
+        for fields in self.get_tasks_capabilities(tasks):
+            selected_fields = fields["common"]
+            cap_inh, cap_prm, cap_eff, cap_bnd, cap_amb = fields["capabilities"]
+
+            if self.config.get("inheritable"):
```
ikelos: Ewwwwww...  5:P  So it looks like this is completely changing the fields that are output from this function.  That somewhat goes against how we've done things previously, which is to return all data.  It's not as good for human readability, but it makes everything much cleaner and easier when used as a library.  Since these are binary values, it should only be a T or F (and if it says True/False, I'd sooner introduce a separate pull request that changes that than import all this branching).

I'm ok when it's a new capability in linux might change the capabilties returned, but a system that calls this plugin to populate a database is going to have a hell of time if ithe user's allowed to choose their options, and it's not really terrible if we spit everything back, as long as a user has a fairly easy way to limit the results.  I'd even sooner accept a PR to allow the CLI to specific which columns are printed.  Doing it within the plugin takes the ability to make display/formatting decisions away from the interface (which may need to make decisions like that)...

volatility3/framework/plugins/linux/capabilities.py
```diff
@@ -0,0 +1,228 @@
...
+
+        columns = [
+            ("Name", str),
+            ("Tid", int),
+            ("Pid", int),
+            ("PPid", int),
+            ("EUID", int),
+        ]
+
+        if self.config.get("inheritable"):
```
ikelos: Generally altering columns and/or conditionals around the yield step of the generator are what start me looking closer.  I think there's a couple of very specific places I've let it in, but this is a lot of fields to do it with, so I'd prefer to get this reworked...

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1428,3 +1429,110 @@ def get_type(self):
...
+        if hasattr(cred_val, "val"):
+            # From kernels 3.5.7 on it is a 'kuid_t' type
+            value = cred_val.val
+        elif isinstance(cred_val, objects.Integer):
+            # From at least 2.6.30 and until 3.5.7 it was a 'uid_t' type which was an 'unsigned int'
+            value = cred_val
+        else:
+            raise AttributeError("Kernel struct cred is not supported")
+
+        return int(value)
```
ikelos, line 1462: It *should* be that value *is* an int already, so unless you're strictly doing this cast for type checking reasons that you've run into, this cast shouldn't be here (and I'm undecided about what return type I'd prefer, leaving it as `int` will probably throw off mypy, but should theoretically be accurate.).

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1428,3 +1429,110 @@ def get_type(self):
...
+
+    def get_capabilities(self) -> int:
+        """Returns the capability bitfield value
+
+        Returns:
+            int: The capability bitfield value.
+        """
+        # In kernels 2.6.25.20 the kernel_cap_struct::cap became and array
+        cap_value = self.cap[0] if isinstance(self.cap, objects.Array) else self.cap
+        return int(cap_value & 0xFFFFFFFF)
```
ikelos: Just want to sanity check why this is being anded (and since it is being anded, it must already be an int, and so the int cast is definitely superfluous in this case) and should it be being anded with the `CAP_FULL` value, once it's moved out into the constants file?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1428,3 +1429,110 @@ def get_type(self):
...
+    def enumerate_capabilities(self) -> List[str]:
+        """Returns the list of capability strings.
+
+        Returns:
+            List[str]: The list of capability strings.
+        """
+        capabilities_value = self.get_capabilities()
+        return self.capabilities_to_string(capabilities_value)
+
+    def has_capability(self, capability: str) -> bool:
```
ikelos, line 1522: A lot of this functionality looks like it might already be handled by the [Enumeration](https://github.com/volatilityfoundation/volatility3/blob/8e56cb39d13c22575ba5133ffbc6b83344bed0ba/volatility3/framework/objects/__init__.py#L553) object type, which then stores the constants in the (or an) ISF JSON definition and comes with all this existing functionality: lookup/description/choices/etc.

It would require making a separate JSON file to load the type up and then casting the capability object to that type, but you could also just override the `__new__` and `__init__` methods to provide the data that would come from the ISF to grab it from the constants file instead?

It's not a direct 1-1 mapping from the two, but at the very least it might be good to model the method names on this off of those used for standard volatility enumerations, just for consistency of use.  It looks like this is an array of a capability type, in which case I guess I'd expect each of the array items to be a capability enumeration and then some membership methods in this main class?

Please take a look at the Enumeration class and let me know if you think there's any similarities that could be used, or method names that might help people intuit them when working with the structure.  If not, then no problem, but it feels like there's some commonality there somewhere, so I just want to raise it and investigate...  5:)

ikelos (commented):
Generally pretty good, really appreciate the use of constants, docstrings and typing info!  5:D

The two major points that need considering are all the display flags that actually alter the plugin output, and then the `kernel_cap_struct` which feels like it might be able to tie into the `Enumeration` objects, or at least mimic their methods instead of using new names.  Happy to talk this through and end up being wrong, but I'd like to at least have the discussion for posterity...  5;D  5:P

Thanks!  Otherwise looks like an awesome plugin and I'm glad to see Vol's linux functionality expanding so nicely!  5:D

ikelos:
Any progress on this?

volatility3/framework/plugins/linux/capabilities.py
gcmoreira: Hey @ikelos thanks for the review. Let's start with this comment because depending on what we decide here many of the other comments may no longer apply. 

I'm not sure what you mean with the T/F, there are 5 capabilities categories and each of them can have 32 capabilities so I don't think you mean to have 160 fixed columns with true/false.

If you don't like this plugin changes the columns based on its command line parameters, I see two options here  
1) Following your second paragraph, we can show each of the 5 categories with its capabilities in text format 5 x (chown, dac_override, dac_read_search, etc.). This will be quite wide but as far as I understand there will be in the future a way to generically select which columns to show for any plugin in vol3, right? 

2) The most condensed way to show everything is actually the default output of this plugin which is the first plugin output showed above in the main PR description.
```shell
  |            Name |  Tid |  Pid | PPid | EUID | cap_inheritable | cap_permitted | cap_effective | cap_bounding | cap_ambient
* |         systemd |    1 |    1 |    0 |    0 |             0x0 |    0xffffffff |    0xffffffff |   0xffffffff |         0x0
* | systemd-journal |  456 |  456 |    1 |    0 |             0x0 |    0x402800cf |    0x402800cf |   0x402800cf |         0x0
...
```
But in order to a human being read that, they will need to used the function helpers I also added to this PR, or to use externals tools like the **capsh** decoder  i.e: capsh --decode=0x402800cf .. but I see this very impractical and that's why I included options so that the user can have the desired output.

Last but not least, I don't see why a system that calls this plugin will have a hell of time. Actually, calling this plugin as a library is straight forward. They can call  get_task_capabilities(task) for an individual task or get_tasks_capabilities(tasks) (yes, I will make this a classmethod ;) ) to get the capabilities of all the tasks. I added all the functionality to the task and kernel_cap_struct types, so the plugin practically doesn't do anything special, only iterate the tasks. 
The returned format for each task is a dictionary as below. 
```python
{
            "common": [
                comm,
                pid,
                tgid,
                pid,
                euid,
            ],
            "capabilities": [
                cap_inheritable,
                cap_permitted,
                cap_effective,
                cap_bset,
                cap_ambient
            ],
}
```
Then, for each cap_* it can call cap_foo.enumerate_capabilities() and it gets a list with the respective textual representation of each capability, as it's done in the generator. The plugin options are only taken into account when the generator is executed. A system using this plugin as a library don't need to use the generator.

volatility3/framework/plugins/linux/capabilities.py
gcmoreira: Actually, this function is not being called from anywhere :) .. I removed the call by mistake, but anyway it doesn't make sense to be called from outside this plugin. The purpose of this function is to warn the developers (us) that they need to update the list of capabilities because it detects that the kernel being analyzed has more capabilities than the ones we support in this plugin. Not sure if you want it here anyway. Thoughts?

ikelos: Errr, yeah, it's fine to keep in here if you'd like.  Happy for you to call it or just leave it as something developers can add in when they need (although that's far less likely to happen), or even to just remove it completely (it'll be in the history now anyway), but if it is going to say, I'd like it to be parameterized (just so people don't quote it to us an as example of core code doing things that way, and therefore they should be able to as well).  5;P

volatility3/framework/plugins/linux/capabilities.py
ikelos: Hiya, I didn't realize there were so many possibilities!  5:S  In that instance then it does change things a little. 160 columns would be too many, I agree.  The textual output is probably for the best here, but I would like all the columns output every time (if there's no content, they should return a `BaseAbsentValue`).  This saves a framework from having to parse the TreeGrid and special case specific columns being present or not.

The output will be wide on the CLI, but another UI could display that data differently (a web app may choose to allow it to be expanded, etc).  It should still be easy enough for a human to read if they're names from an enumeration

Choosing columns isn't currently a planned feature for the CLI, but we absolutely could, the point is that's in the hands of the UI, not in the hands of the plugin.  Whilst the plugin does provide specific functions for doing things, a UI should not need to know about them to get the data from volatility.  The hope is that UIs need as little "special casing" for the outputs of plugins as possible (hence not requiring code if the number of columns change).  This is specifically generic UIs, not tools using volatility as library.  Variable column names, having to interpret data, or having to parse/post-process data all go against that idea.  I think commas is the least bad thing we can have in this instance where there is so much dense information that needs displaying.

I hope that all makes sense and you're happy with the reasoning behind it?  I may entirely still have missed something, or not considered the different between cases where everything is turned on and nothing is turned on.  If there's still something you want me to think over or a case I've missed, just let me know, I'm happy to keep talking this through until we get it right.  5:)

( I agree the T/F thing was not the right choice for this data, though, thanks for catching me on it!)  5:D

ikelos:
Thanks for the new comments, you're right about the capabilities, my suggestion would've gone badly, but I'm still keen that all five columns are always present in the output.  I'm also leaning more and more towards setting up an `Enumeration` as being the right way to handle all this.  Gimme a shout if you want a hand getting that in place.  5:)

volatility3/framework/plugins/linux/capabilities.py
```diff
@@ -0,0 +1,207 @@
...
+
+        Returns:
+            str: A string with a comma separated list of decoded capabilities
+        """
+        if isinstance(cap, renderers.NotAvailableValue):
+            return cap
+
+        cap_value = cap.get_capabilities()
+        if cap_value == 0:
+            return "-"
```
ikelos: This should be a `BaseAbsentValue` derived instance please (and can also be `if not cap_value`) just so that UIs that might want to know that there's no real data there can display it differently if needed

volatility3/framework/plugins/linux/capabilities.py
```diff
@@ -0,0 +1,207 @@
...
+        the kernel being analysed. Otherwise, it shows a warning for the
+        developers.
+        """
+
+        vmlinux = context.modules[vmlinux_module_name]
+
+        try:
+            kernel_cap_last_cap = vmlinux.object_from_symbol(symbol_name="cap_last_cap")
+        except exceptions.SymbolError:
+            # It should be a kernel < 3.2
```
ikelos, line 91: Just wanted to check this doesn't need a log message (even just a debug one)?

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Ok, I had this ready waiting you reply :) ... Have a look now. It looks much cleaner. I used [dataclasses](https://docs.python.org/3.7/library/dataclasses.html) introduced in Python 3.7 (all good with vol3 minimum requirements) which are really cool, not like namedtuples.

volatility3/framework/plugins/linux/capabilities.py
```diff
@@ -0,0 +1,207 @@
...
+        if cap_value == 0:
+            return "-"
+
+        if cap_value == CAP_FULL:
+            return "all"
+
+        return ", ".join(cap.enumerate_capabilities())
+
+    @classmethod
+    def get_task_capabilities(cls, task: interfaces.objects.ObjectInterface) -> Dict:
```
ikelos: I think it now returns a `Tuple[TaskData, CapabilitiesData]` instead of a `Dict`?

volatility3/framework/plugins/linux/capabilities.py
gcmoreira: Ok, I had this ready waiting you reply :) ... Have a look now. It looks much cleaner. I used [dataclasses](https://docs.python.org/3.7/library/dataclasses.html) introduced in Python 3.7 (all good with vol3 minimum requirements) which are really cool, not like namedtuples.

volatility3/framework/plugins/linux/capabilities.py
```diff
@@ -0,0 +1,207 @@
...
+from volatility3.framework.configuration import requirements
+from volatility3.framework.interfaces import plugins
+from volatility3.framework.objects import utility
+from volatility3.framework.symbols.linux import extensions
+from volatility3.plugins.linux import pslist
+
+vollog = logging.getLogger(__name__)
+
+
+@dataclass
```
ikelos, line 20: You're right, dataclasses are cool!  I very much approve!  5:D

ikelos (commented):
Awesome!  Sooooooo much cleaner!  I hope you're as happy with the new code as I am!  Really smooth!  5:D

There's still a couple tiny bits that I skipped over until we'd figured out the column output stuff, and if you fill like it there's still the `Enumeration` idea to look at if you'd like, but I'm not gonna push it because it's so much tidier now I'm happy to leave it like this if you are...  5:)

volatility3/framework/plugins/linux/capabilities.py
gcmoreira: It doesn't need a log there. Because, old kernels didn't have this symbol, so in those cases we are unable to check that if we support all the capabilities. Anyway, it doesn't matter, we need this for the latest kernels. The older kernels we already know the capabilities list, it's something I had to check when I wrote it code.

volatility3/framework/plugins/linux/capabilities.py
gcmoreira: And its members are typed, so we can also use in combination with the typing module.

volatility3/framework/plugins/linux/capabilities.py
gcmoreira: In this case, the "-" (or the cap_value == 0) means that there is no capability enabled. It doesn't seem that any of the UnreadableValue, UnparsableValue, NotApplicableValue, NotAvailableValue are correct here. It should show an empty value in this case, it's not an error. What do you think it could be the best option? I can also replace it with an empty string if you think it's better.

volatility3/framework/plugins/linux/capabilities.py
gcmoreira: good catch ;)

gcmoreira:
> ... if you fill like it there's still the `Enumeration` idea to look at if you'd like, but I'm not gonna push it because it's so much tidier now I'm happy to leave it like this if you are... 5:)

Yeah, well the native type is not an enum, it's a bitfield. If you think it's not that bad I would leave it like this 🙏 :)

volatility3/framework/plugins/linux/capabilities.py
gcmoreira: Ok, I pushed a new version with an empty string. If you like we leave it like that.

volatility3/framework/plugins/linux/capabilities.py
gcmoreira: Actually, there were more than that to update. I also removed the type from the docstring, it's redundant and one more place to modify when these things happen.

volatility3/framework/plugins/linux/capabilities.py
ikelos: Hehehe, touche, you got me there, it doesn't really match any of the existing ones.  Given that's the case I'm happy with a `""` or a `"-"` or even a more descriptive `"No capabilities"` or something?  I'm happy for you to pick (you might be tired of having to change it, so don't worry "" is fine if you want!)  It's getting late here, so I'll probably push it tomorrow, feel free to update it or not before then...  5:)

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: I'll let this slide (but only because you've been so good on getting the rest of it all sorted!)  5;P

volatility3/framework/plugins/linux/capabilities.py
gcmoreira: Haha, thanks. That should be everything. Also, all the test passed youhuu, so be free to check it again and merge it whenever you want.

PR #968, opened 2023-06-15
Windows: add `--verbose` option for `ldrmodules` plugin.

### LINKED ISSUE #967: ldrmodules does not display the full path of the DLLs
In volatility 2, as per [Command Reference Mal](https://github.com/volatilityfoundation/volatility/wiki/Command-Reference-Mal#ldrmodules), you can specify the verbose flag -v/--verbose to get the full path of the DLLs found in the 3 PEB lists.  Volatility 3 does not have this functionality, although it sounds to be very valuable in some situations.

digitalisx:
## Description
Hello, everyone in the community! :)
This PR comes from this issue (#967).
It will be meaningful to re-implement the original features of [volatility](https://github.com/volatilityfoundation/volatility/blob/a438e768194a9e05eb4d9ee9338b881c0fa25937/volatility/plugins/malware/malfind.py#L523).

## Examples
```shell
> python3 vol.py -f case.vmem windows.ldrmodules --verbose
Volatility 3 Framework 2.4.2
Pid	Process	Base	InLoad	InInit	InMem	MappedPath	LoadPath	InitPath	MemPath

644	services.exe	0x22ea0970000	False	False	False	\Windows\System32\ko-KR\services.exe.mui	N/A	N/A	N/A
644	services.exe	0x7ff6cee90000	True	False	True	\Windows\System32\services.exe	C:\Windows\system32\services.exe : services.exe	N/A	C:\Windows\system32\services.exe : services.exe
660	lsass.exe	0x7ffd48da0000	True	True	True	\Windows\System32\rsaenh.dll	C:\Windows\system32\rsaenh.dll : rsaenh.dll	C:\Windows\system32\rsaenh.dll : rsaenh.dll	C:\Windows\system32\rsaenh.dll : rsaenh.dll
660	lsass.exe	0x7ffd4a610000	True	True	True	\Windows\System32\rpcrt4.dll	C:\Windows\System32\RPCRT4.dll : RPCRT4.dll	C:\Windows\System32\RPCRT4.dll : RPCRT4.dll	C:\Windows\System32\RPCRT4.dll : RPCRT4.dll
760	svchost.exe	0x7ff760700000	True	False	True	\Windows\System32\svchost.exe	C:\Windows\system32\svchost.exe : svchost.exe	N/A	C:\Windows\system32\svchost.exe : svchost.exe
760	svchost.exe	0x7ffd418e0000	True	True	True	\Windows\System32\AppXDeploymentClient.dll	C:\Windows\System32\AppXDeploymentClient.dll : AppXDeploymentClient.dll	C:\Windows\System32\AppXDeploymentClient.dll : AppXDeploymentClient.dll	C:\Windows\System32\AppXDeploymentClient.dll : AppXDeploymentClient.dll
760	svchost.exe	0x7ffd48da0000	True	True	True	\Windows\System32\rsaenh.dll	C:\Windows\system32\rsaenh.dll : rsaenh.dll	C:\Windows\system32\rsaenh.dll : rsaenh.dll	C:\Windows\system32\rsaenh.dll : rsaenh.dll
```

### DIFF
--- a/volatility3/framework/plugins/windows/ldrmodules.py
+++ b/volatility3/framework/plugins/windows/ldrmodules.py
@@ -10,7 +10,7 @@ class LdrModules(interfaces.plugins.PluginInterface):
     """Lists the loaded modules in a particular windows memory image."""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 0, 0)
+    _version = (1, 0, 1)
 
     @classmethod
     def get_requirements(cls):
@@ -32,6 +32,12 @@ class LdrModules(interfaces.plugins.PluginInterface):
                 description="Process IDs to include (all other processes are excluded)",
                 optional=True,
             ),
+            requirements.BooleanRequirement(
+                name="verbose",
+                description="Print the full paths and base names in verbose mode ",
+                default=False,
+                optional=True,
+            )
         ]
 
     def _generator(self, procs):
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/ldrmodules.py
```diff
@@ -32,6 +32,12 @@ def get_requirements(cls):
                 description="Process IDs to include (all other processes are excluded)",
                 optional=True,
             ),
+            requirements.BooleanRequirement(
+                name="verbose",
```
ikelos, line 36: This feels like it might get confusing given the main CLI parameters of `-v` (which can also be written as `--verbose`).  I'd suggest something like `--detail` or `--extra` or `--extra-information` or even just `--full-path`.

volatility3/framework/plugins/windows/ldrmodules.py
```diff
@@ -82,45 +88,115 @@ def filter_function(x: interfaces.objects.ObjectInterface) -> bool:
...
-                    ],
-                )
+                load = renderers.NotApplicableValue()
+                init = renderers.NotApplicableValue()
+                mem = renderers.NotApplicableValue()
+
+                try:
+                    if load_mod:
+                        load = "{0} : {1}".format(
+                            load_mod.FullDllName.get_string(),
```
ikelos, line 98: Do these two values differ, or is the `BaseDllName` part of the `FullDllName`?  Either way we try hard to avoid textually combining data into a single column, because it means that a program using volatility as a library has to do work to split the data back up into its components.  As such, I'd either change the default column entry to the full one (and forget the flag) or if both bits of information are important, then I'd output a second column.

It's slightly more effort, but it will make it much easier for other code to process the results, and humans will still be able to read it.  Since we have version numbers we can change the default output of a plugin without too much issue (and people can check the version number if they depend on specific columns).

volatility3/framework/plugins/windows/ldrmodules.py
```diff
@@ -82,45 +88,115 @@ def filter_function(x: interfaces.objects.ObjectInterface) -> bool:
...
-            self._generator(
-                pslist.PsList.list_processes(
-                    context=self.context,
-                    layer_name=kernel.layer_name,
-                    symbol_table=kernel.symbol_table_name,
-                    filter_func=filter_func,
-                )
-            ),
-        )
+        if not self.config.get("verbose", True):
```
ikelos, line 160: Changing column outputs is doable, but it potentially makes it tricky for things that use the output of plugins automatically (nothing does that I know of yet, and they could read the data from the `TreeGrid` they get back, but...).  If the extra data isn't too messy, then I'd just output it all, all the time.  If it is, then we can go with the optional columns, but again, we're trying to return things as if we're in a database, so separation of things that are separate and as little duplication as possible...  5:)

ikelos (commented):
Thanks very much, very swift fix to issue #968.  I have a few concerns though, we try not to put separate data items into a single field, and where possible (it's not always feasible) we try to not to change the layout of the table based on an input option.  I'm happy to do that here if it's necessary, but it might just be worth always outputting the full data all the time, depending on how ugly it gets?  Lemme know what you think and we can come up with the best way forward...  5:)

digitalisx:
Thank you for your review @ikelos.
The PR was quick, but I think there are a lot of things to work on while thinking about the contents of the review carefully. The response is a bit late, but this is still on my list of tasks of interest.

volatility3/framework/plugins/windows/ldrmodules.py
digitalisx: All right, I accepted your comments.

PR #970, opened 2023-06-16
Windows: Fix VAD offset canonicalization

### LINKED ISSUE #969: windows.vadinfo.VadInfo shows 48-bit addresses in the `Offset` column on 64-bit systems
Inspecting the output of windows.vadinfo.VadInfo plugin on Windows 10 system I noticed the following.
The parent of node 0xb50ade2b1920 is 0xffffb50ade2b1970 which seems to be a valid 64-bit address, but the offset of it and its parent is provided as a 48-bit address 0xb50ade2b1920  & 0xb50ade2b1970.

windows.vadinfo.VadInfo:
```
Volatility 3 Framework 2.4.1
  |  PID |        Process |         Offset |      Start VPN |        End VPN |  Tag |             Protection | CommitCharge | PrivateMemory |             Parent 
* |    4 |         System | 0xb50ade2b1970 |     0x7ffef000 |     0x7ffeffff | VadS |          PAGE_READONLY |            1 |             1 |                0x0 
* |    4 |         System | 0xb50ade2b1920 |     0x7ffe0000 |     0x7ffe0fff | VadS |          PAGE_READONLY |            1 |             1 | 0xffffb50ade2b1970 
```

If I have my facts straight, canonical addresses on x86-64 hardware have the high 16 bits as the sign-extension of the lower 48 bits. From what I see in the above output, the Offset column (0xb50ade2b1970) is missing the high 16 bits, which make it a non-canonical address. Furthermore. If I'm not mistaken the VAD nodes should reside within the system address space. Converting the 0xb50ade2b1920 to decimal I get: 199,058,281,666,848 which is not part of the upper 128 TIB system space, as the space for 64-bit Windows 10 is from 0xFFFF8000'00000000 to -1. 

Although, I'm quite certain this an issue with the plugin, it may actually be mistake on my part due to lack of knowledge on the subject. If that is the case I apologize and would appreciate some explanation regarding this observation of the discrepancy between the Offset & Parent column.

The memory image is of Windows 10 Build 19041.

ikelos:
Closes #969

### DIFF
--- a/volatility3/framework/plugins/windows/vadinfo.py
+++ b/volatility3/framework/plugins/windows/vadinfo.py
@@ -198,6 +198,7 @@ class VadInfo(interfaces.plugins.PluginInterface):
 
     def _generator(self, procs):
         kernel = self.context.modules[self.config["kernel"]]
+        kernel_layer = self.context.layers[kernel.layer_name]
 
         def passthrough(_: interfaces.objects.ObjectInterface) -> bool:
             return False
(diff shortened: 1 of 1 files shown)


ikelos:
Minimal change and verified by the user, so merging now...

Issue #958, opened 2023-05-18
Volshell: rs() command not working as expected


eve-mem:
**Describe the bug**
When attempting to use the `rs()` command in volshell to run a script it does not appear to work on my system. I have the habit now of just pasting the script into the terminal, but decided I should raise this as a little bug. 

I get the same issues on linux and windows. 

I've attempted to have a quick look for a fix, but it wasn't obvious to me right away so rising the issue to track it.

**Context**
Volatility Version:  Develop branch `8e56cb39d13c22575ba5133ffbc6b83344bed0ba`

Windows:
Operating System:  `Windows 10, 22H2 Build 19045.2965`
Python Version:  Python 3.11.1

Linux:
Operating System:  `Linux voltester 5.10.0-21-amd64 #1 SMP Debian 5.10.162-1 (2023-01-21) x86_64 GNU/Linux`
Python Version:  Python 3.9.2

Suspected Operating System:  N/A
Command:  ```rs('radix.py')```

**To Reproduce**

1. Use command 'python volshell.py -f linux-sample-1.dmp -l`

Auto magic works correctly and all other parts of volshell work as expected. 

2. Then attempt to run a script from within volshell, but it raises an AttributeError. 
```
(layer_name) >>> rs('radix.py')
Running code from file:radix.py

Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "C:\Users\eve\volatility3\volatility3\cli\volshell\generic.py", line 483, in run_script
    io.TextIOWrapper(fp.read(), encoding="utf-8"), symbol="exec"
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'bytes' object has no attribute 'readable'
```

**Expected behavior**
I would expect the script to run in the volshell terminal.

ikelos:
Thanks, turns out this was a typo on my side, I put the `.read` inside the brackets, rather than outside them, but it's now resolved.  Also thanks to @thomasdangl for the quick fix in #975 and making me look harder into the issue.  5:D  Any residual problems please feel free to reopen it.  5:)

PR #975, opened 2023-06-28
Fix file encoding in run_script

thomasdangl:
Addresses problem in https://github.com/volatilityfoundation/volatility3/issues/958.


ikelos:
Hmmm, looks like I screwed around with it back during https://github.com/volatilityfoundation/volatility3/commit/129b92e3eedfbc71b1b1570e0d01c4c22e4a6fc1 and broke it.  5:S  Unfortunately there was an issue with the linting, so I made a similar change manually (fixing the typo I originally made) and pushed that.  Thanks again for your contribution though, your fix was fine and would have resolved the problem (and the pull request spurred me to dig into it)!  5:D

PR #978, opened 2023-07-01
Core: Add type parameter to object_from_symbol

ikelos:
Add in the ability to pass a type name when creating a symbol from an object (since windows symbols don't often have the type information associated with them in the ISF files).


ikelos:
No complaints after two weeks, so we'll add it.

PR #984, opened 2023-07-19
Linux capabilities fixes

gcmoreira:
Several fixes:
* The capabilities array was meant to be a 64bit bitwise.
* Capabilities set has to be tested with the kernel maximum value. We can't use the plugin capabilities set, otherwise, when the plugin list of capabilities has more elements than the kernel being analyzed we will have wrong interpretations when we tried to compress the list of capabilities to "all"
* Supports kernels >= 6.3. They changed the kernel_cap_struct::cap type again to a u64 type.

### DIFF
--- a/volatility3/framework/constants/linux/__init__.py
+++ b/volatility3/framework/constants/linux/__init__.py
@@ -279,5 +279,3 @@ CAPABILITIES = (
     "bpf",
     "checkpoint_restore",
 )
-
-CAP_FULL = 0xFFFFFFFF
--- a/volatility3/framework/plugins/linux/capabilities.py
+++ b/volatility3/framework/plugins/linux/capabilities.py
@@ -7,7 +7,6 @@ from dataclasses import dataclass, astuple, fields
 from typing import Iterable, List, Tuple
 
 from volatility3.framework import interfaces, renderers, exceptions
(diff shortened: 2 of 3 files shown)


gcmoreira:
@ikelos Regarding kernel >= 6.3 support, I couldn't test this properly. It should work but there is no Linux distribution with kernel 6.3 yet, so there is no standard .ddeb package to generate the profile. **Mainland** has the stripped kernel but not the .ddeb.  
I compiled a kernel 6.3 from scratch with debug symbols but for some reason vol3 doesn't like it. I'm still not sure if it's the profile I created or if the 6.3 kernel has other changes that make it fail. I plan to look into it further to see why it's failing but it may take some time. In the meantime, I think it's safe to merge these changes and I will create another PR in the future if I found any issues with it.

ikelos (approved):
Looks good, thanks!

ikelos:
So we've got #985 which might be a 6.3 kernel issue.  Their symbol tables doesn't contain a `kernel_cap_struct` which because it's not marked as optional is throwing off any linux analysis.  It turns out it wasn't showing any errors until the full `-vvvvvvv` was turned on, so that might be your issue with vol3 still not liking it?  Anyway, hopefully you can take a lok and please shout if I can help at all.  5:)

PR #913, opened 2023-02-19
Feature/snappy ctypes

ikelos:
This replaces the python-snappy dependency (which had a number of installation difficulties) with a ctypes wrapper.  The wrapper does not force a dependency on the whole of volatility in order to support AVML files, and provides information if the AVML file format is detected but cannot be processed.  On most linux/mac systems the system package for libsnappy should install it to the right location, on Windows, snappy32.dll/snappy64.dll will need to be installed off [the link](https://snappy.machinezoo.com/#shell) from the [official snappy page](https://google.github.io/snappy/).

This should resolve a number of snappy instruction/snappy installation issues that have plagued the AVML support.

PR #987, opened 2023-07-26
Use python-snappy binding instead of ctypes to make it more portable and shorter

xabrouck:
On macOS, the snappy library couldn't be loaded properly when processing a compressed avml file.
It's easier, shorter and more portable to use the python-snappy binding.
It was tested successfully on a compressed AVML file.


digitalisx:
Thank you for your contribution @xabrouck!
`python-snappy` has been used because it is simple,
but it has been replaced because of the installation issue of many users.
Please take a look at #913.

PR #988, opened 2023-07-27
fix bug in snappy lib loading and make it work on macOS.

xabrouck:
Alternative to #987 as it was bringing back issues from #913


ikelos (approved):
Looks good, thanks.  5:)

PR #995, opened 2023-08-14
Check IoC of dirty bit in PTEs from executable VMAs.

xabrouck:
Currently, the Malfind plugin in volatility3 does not check the dirty bit in PTEs associated with executable VMAs.  When someone modifies running code for example using ptrace (e.g. by writing to /proc/pid/mem), the permissions of the VMA remain r-x (ptrace does not require rwx) and the targeted pages become dirty. This is visible in /proc/pid/smaps where the Private_Dirty of executable VMAs become non zero.

The goal of this patch is to detect such dirty executable pages as it likely means that the target process was debugged at some point or that malicious code was injected.

It could happen that no modification is found when comparing the memory dump of the process with the original executable/libraries.  In that case, it can mean that the malware got uninstalled, i.e. the original code was rewritten on top of the malicious code before the memory dump took place.  Nevertheless, it is still interesting to know that some code got injected at some point in the past.

The downside of this patch is that it makes Malfind slower.

### DIFF
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -110,11 +110,11 @@ class Intel(linear.LinearlyMappedLayer):
     def _page_is_valid(entry: int) -> bool:
         """Returns whether a particular page is valid based on its entry."""
         return bool(entry & 1)
-    
+
     @staticmethod
     def _page_is_dirty(entry: int) -> bool:
         """Returns whether a particular page is dirty based on its entry."""
-        return bool(entry & (1<<6))
+        return bool(entry & (1 << 6))
 
     def canonicalize(self, addr: int) -> int:
         """Canonicalizes an address by performing an appropiate sign extension on the higher addresses"""
@@ -267,7 +267,7 @@ class Intel(linear.LinearlyMappedLayer):
     def is_dirty(self, offset: int) -> bool:
         """Returns whether the page at offset is marked dirty"""
         return self._page_is_dirty(self._translate_entry(offset)[0])
-    
+
     def mapping(
         self, offset: int, length: int, ignore_errors: bool = False
     ) -> Iterable[Tuple[int, int, int, int, str]]:
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -110,6 +110,11 @@ class Intel(linear.LinearlyMappedLayer):
     def _page_is_valid(entry: int) -> bool:
         """Returns whether a particular page is valid based on its entry."""
         return bool(entry & 1)
+    
+    @staticmethod
+    def _page_is_dirty(entry: int) -> bool:
+        """Returns whether a particular page is dirty based on its entry."""
+        return bool(entry & (1<<6))
 
     def canonicalize(self, addr: int) -> int:
         """Canonicalizes an address by performing an appropiate sign extension on the higher addresses"""
@@ -259,6 +264,10 @@ class Intel(linear.LinearlyMappedLayer):
         except exceptions.InvalidAddressException:
             return False
 
+    def is_dirty(self, offset: int) -> bool:
+        """Returns whether the page at offset is marked dirty"""
+        return self._page_is_dirty(self._translate_entry(offset)[0])
+    
     def mapping(
         self, offset: int, length: int, ignore_errors: bool = False
     ) -> Iterable[Tuple[int, int, int, int, str]]:
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -578,7 +578,7 @@ class vm_area_struct(objects.StructType):
         return fname
 
     # used by malfind
-    def is_suspicious(self, proclayer):
+    def is_suspicious(self, proclayer=None):
         ret = False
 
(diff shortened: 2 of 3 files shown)


volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -578,7 +578,7 @@ def get_name(self, context, task):
         return fname
 
     # used by malfind
-    def is_suspicious(self):
+    def is_suspicious(self, proclayer):
```
ikelos: This will be an api change, unless you default it to None and then verify whether `proclayer` has been set before using it...

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -587,6 +587,15 @@
...
         elif flags_str == "r-x" and self.vm_file.dereference().vol.offset == 0:
             ret = True
+        elif "x" in flags_str:
+            for i in range(self.vm_start,self.vm_end,constants.linux.PAGE_SHIFT):
+                try:
+                    if proclayer.is_dirty(i):
+                        vollog.warning(f"Found malicious (dirty+exec) page at {hex(i)} !")
+                        ret = True
+                        break
+                except (exceptions.PagedInvalidAddressException, exceptions.InvalidAddressException):
```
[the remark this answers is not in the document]
ikelos: Also, please use vollog to log out something when an exception occurs, unless it's *expected* to occur.  Either way, a comment would be useful to say why it can be ignored (and it'll also make this messages go away).  5;)

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -587,6 +587,15 @@ def is_suspicious(self):
             ret = True
         elif flags_str == "r-x" and self.vm_file.dereference().vol.offset == 0:
             ret = True
+        elif "x" in flags_str:
+            for i in range(self.vm_start,self.vm_end,constants.linux.PAGE_SHIFT):
+                try:
```
ikelos, line 592: Should we keep trying all the other pages if one fails or can the exception move outside the loop?

ikelos (commented):
Thanks a few gotchas, and please apply black to your submission, to make sure the code is all formatted uniformly.  You can see on the build list why black believes it's failing if you just want to manually fix it up...

volatility3/framework/symbols/linux/extensions/__init__.py
xabrouck: I changed it to stop searching in that VMA if a bad translation occurs.

xabrouck: I added comments and improved logging

volatility3/framework/symbols/linux/extensions/__init__.py
xabrouck: Ok, I've done that.

xabrouck:
@ikelos Any remaining concern?

xabrouck:
There's a corner case with some old binaries/libraries that trigger false positives.  It's explained in
https://stackoverflow.com/questions/76651605/understanding-a-dt-textrel-warning
If there are absolute addresses in a PIE (unusual), it will have to do runtime fixups, which require writing to an executable region.

Here's a small example to reproduce the corner case:
```
$ cat dirtyexec.c
int x=0; int main() { scanf("%d",&x); }

$ gcc -w -mcmodel=large -fno-pie dirtyexec.c -o dirtyexec
/usr/bin/ld: /tmp/ccuBN6g5.o: warning: relocation in read-only section `.text'
/usr/bin/ld: warning: creating DT_TEXTREL in a PIE

$ file dirtyexec
dirtyexec: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=f50788c7e03024bd325756f43eb153d0fb265491, for GNU/Linux 3.2.0, not stripped

$ readelf -r dirtyexec|grep R_X86_64_64
000000001172  000500000001 R_X86_64_64       0000000000001050 scanf@GLIBC_2.2.5 + 0

$ ./dirtyexec
^Z
[1]+  Stopped                 ./dirtyexec

$ egrep "\-|Private_Dirty" /proc/$(pidof dirtyexec)/smaps|head -4|tail -2
55c7a3619000-55c7a361a000 r-xp 00001000 fd:00 6426135                    /home/xabrouck/reloc-test/2/dirtyexec
Private_Dirty:         4 kB 
```

volatility3/framework/layers/intel.py
```diff
@@ -259,6 +264,10 @@ def is_valid(self, offset: int, length: int = 1) -> bool:
         except exceptions.InvalidAddressException:
             return False
 
+    def is_dirty(self, offset: int) -> bool:
```
ikelos, line 267: My slight concern is that we only add this to the intel layer, and not the underlying `TranslationLayerInterface`, but I think we've done that in other places.  I guess it just requires that any time we use it we ensure we know we're working with an intel layer (at the moment, that's likely to be all we're working with, but in the future it could be an arm layer or something else)...

ikelos (approved):
Sorry for the delay in reviewing this.  I'm quite nervous about adding things into the Intel space, but I believe this is ok so I'm going to merge it in.

PR #928, opened 2023-03-24
Add basic support for Linux maple tree struct

eve-mem:
This adds basic support for Linux Maple Trees (from `include/linux/maple_tree.h` and `lib/maple_tree.c`) specifically focusing on the mm_mt tree found within the mm_struct.

This would fix https://github.com/volatilityfoundation/volatility3/issues/909

This does not yet include any smear protection, I wanted to get thoughts on this before doing that in case I've taken the wrong approach. 

This has been tested on 6.2.7, 6.2.2, and 6.1.15 for maple tree support. I have also tested against older kernels to ensure this didn't break anything. (I can share these mem images if useful.)

Here is a sample of the linux.proc plugin using this maple tree parsing and the output `cat /proc/<PID>/maps` on the test VM. They show the same results.

procfs:
```
563a816af000-563a816b4000 r--p 00000000 08:01 1442768                    /usr/bin/nano
563a816b4000-563a816dc000 r-xp 00005000 08:01 1442768                    /usr/bin/nano
563a816dc000-563a816ea000 r--p 0002d000 08:01 1442768                    /usr/bin/nano
563a816ea000-563a816eb000 r--p 0003a000 08:01 1442768                    /usr/bin/nano
563a816eb000-563a816ec000 rw-p 0003b000 08:01 1442768                    /usr/bin/nano
563a816ec000-563a816ed000 rw-p 00000000 00:00 0 
563a82594000-563a826a0000 rw-p 00000000 00:00 0                          [heap]
7ff489c00000-7ff489ee5000 r--p 00000000 08:01 1453964                    /usr/lib/locale/locale-archive
7ff48a08e000-7ff48a091000 r--p 00000000 08:01 1453247                    /usr/lib/x86_64-linux-gnu/libnss_files-2.28.so
7ff48a091000-7ff48a098000 r-xp 00003000 08:01 1453247                    /usr/lib/x86_64-linux-gnu/libnss_files-2.28.so
7ff48a098000-7ff48a09a000 r--p 0000a000 08:01 1453247                    /usr/lib/x86_64-linux-gnu/libnss_files-2.28.so
7ff48a09a000-7ff48a09b000 ---p 0000c000 08:01 1453247                    /usr/lib/x86_64-linux-gnu/libnss_files-2.28.so
7ff48a09b000-7ff48a09c000 r--p 0000c000 08:01 1453247                    /usr/lib/x86_64-linux-gnu/libnss_files-2.28.so
[...]

PR #996, opened 2023-08-17
Linux: fix bug where get_process_memory_sections fails with 6.1+ kernels

eve-mem:
This fixes a bug spotted by alishir on slack. This changes ensures that the `get_process_memory_sections` function in the `task_struct` uses the generic `get_vma_iter` iterator in the `mm_struct` extension - this will then choose either `mmap` or `mm_mt` as needed. 

The `get_vma_iter` function was added in this pull request but this path was missed:  https://github.com/volatilityfoundation/volatility3/pull/928 

Without this plugins that use `get_process_memory_sections` won't work on Linux kernels 6.1+.

### DIFF
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -203,7 +203,7 @@ class task_struct(generic.GenericIntelProcess):
     ) -> Generator[Tuple[int, int], None, None]:
         """Returns a list of sections based on the memory manager's view of
         this task's virtual memory."""
-        for vma in self.mm.get_mmap_iter():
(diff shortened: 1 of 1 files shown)


eve-mem:
I forgot to say which current plugins are affected. At the moment it is just the `linux.bash` plugin that uses `get_process_memory_sections`.

That means at the moment the `linux.bash` plugin only works with Linux kernels prior to 6.1 and will crash here on the modern kernels. https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/symbols/linux/extensions/__init__.py#L401

ikelos (approved):
Sorry for the delay in reviewing this.  All looks good, thanks for spotting it!

PR #997, opened 2023-08-22
Issue 985

### LINKED ISSUE #985: Capabilities plugin breaks linux support for kernels without `kernel_cap_struct`
We've had a report from a user that their linux plugins were failing with the following vebose output:

```
Level 8  volatility3.framework.automagic.stacker: Attempting to stack using LinuxIntelStacker
DEBUG    volatility3.framework.automagic.linux: Identified banner: b'Linux version 6.3.12-200.fc38.x86_64 (mockbuild@ad82227541134b7bb9538baac1587d1c) (gcc (GCC) 13.1.1 20230614 (Red Hat 13.1.1-4), GNU ld version 2.39-9.fc38) #1 SMP PREEMPT_DYNAMIC Thu Jul  6 04:05:18 UTC 2023\n\x00'
Level 7  volatility3.framework.automagic.stacker: Exception during stacking: Symbol type not in LintelStacker1 SymbolTable: kernel_cap_struct
Level 6  volatility3.framework.automagic.stacker: Traceback (most recent call last):
  File "/home/xatom32/Documenti/volatility3/volatility3/volatility3/framework/automagic/stacker.py", line 213, in stack_layer
    new_layer = stacker.stack(context, initial_layer, progress_callback)
  File "/home/xatom32/Documenti/volatility3/volatility3/volatility3/framework/automagic/linux.py", line 72, in stack
    table = linux.LinuxKernelIntermedSymbols(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xatom32/Documenti/volatility3/volatility3/volatility3/framework/symbols/linux/__init__.py", line 32, in __init__
    self.set_type_class("kernel_cap_struct", extensions.kernel_cap_struct)
  File "/home/xatom32/Documenti/volatility3/volatility3/volatility3/framework/symbols/intermed.py", line 60, in _delegate_function
    return getattr(self._delegate, name)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xatom32/Documenti/volatility3/volatility3/volatility3/framework/symbols/intermed.py", line 425, in set_type_class
    raise ValueError(f"Symbol type not in {self.name} SymbolTable: {name}")
ValueError: Symbol type not in LintelStacker1 SymbolTable: kernel_cap_struct
```

This seems to be because the `kernel_cap_struct` isn't present in the symbol table.  At the moment we're not sure whether the symbol table was made correctly, or if the symbol really isn't there.  However, it might be prudent to allow for kernels that are too old or don't have the symbol, to continue loading by marking that symbol structures as optional.  The line in question is [here](https://github.com/volatilityfoundation/volatility3/blob/373d769690bf1f9345a5be68f7555b50fc103152/volatility3/framework/symbols/linux/__init__.py#L32) and could be changed to `optional_set_type_class` but would then require checking in the plugin that the structure exists and graceful error message if it doesn't.

@gcmoreira, any chance you could look into this please?

gcmoreira:
This fixes issue #985. It requires [dwarf2json PR #52](https://github.com/volatilityfoundation/dwarf2json/pull/52) which is now merged into master.

### DIFF
--- a/volatility3/framework/symbols/linux/__init__.py
+++ b/volatility3/framework/symbols/linux/__init__.py
@@ -29,10 +29,11 @@ class LinuxKernelIntermedSymbols(intermed.IntermediateSymbolTable):
         self.set_type_class("files_struct", extensions.files_struct)
         self.set_type_class("kobject", extensions.kobject)
         self.set_type_class("cred", extensions.cred)
-        self.set_type_class("kernel_cap_struct", extensions.kernel_cap_struct)
         # Might not exist in the current symbols
         self.optional_set_type_class("module", extensions.module)
         self.optional_set_type_class("bpf_prog", extensions.bpf_prog)
+        self.optional_set_type_class("kernel_cap_struct", extensions.kernel_cap_struct)
+        self.optional_set_type_class("kernel_cap_t", extensions.kernel_cap_t)
 
         # Mount
         self.set_type_class("vfsmount", extensions.vfsmount)
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -1531,7 +1531,7 @@ class kernel_cap_struct(objects.StructType):
             if len(self.cap) == 1:
                 # At least in the vanilla kernel, from 2.6.24 to 2.6.25
                 # kernel_cap_struct::cap become a two elements array.
-                # However, in some distros or custom kernel can techically
+                # However, in some distros or custom kernel can technically
                 # be _KERNEL_CAPABILITY_U32S = _LINUX_CAPABILITY_U32S_1
                 # Leaving this code here for the sake of ensuring completeness.
                 cap_value = self.cap[0]
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -1472,13 +1472,13 @@ class cred(objects.StructType):
 
 
 class kernel_cap_struct(objects.StructType):
-    # struct kernel_cap_struct was added in kernels 2.5.0
+    # struct kernel_cap_struct exists from 2.1.92 <= kernels < 6.3
     @classmethod
     def get_last_cap_value(cls) -> int:
         """Returns the latest capability ID supported by the framework.
 
         Returns:
-            int: The latest supported capability ID supported by the framework.
+            int: The latest capability ID supported by the framework.
         """
         return len(CAPABILITIES) - 1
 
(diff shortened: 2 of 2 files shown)


gcmoreira:
Hey @ikelos! could we work together to have this merged asap?  There are some conflicts now. Do you want me to update the code?

ikelos (commented):
I'm happy merging this in, but there doesn't seem to be any detection or code for determining which version of dwarf2json was used to generate the symbols?  The comments mentioned this required symbols generated by a particular dwarf2json.  What's the upshot of using old symbols with this?  Will it break, or can it stand as is?  If it's ok with older version of dwarf2json (or at least tells people they need to regenerate the symbols to use this plugin), then I'll get it merged.  If we need to faff with getting stopping particular symbol tables based on versions of the generation tool, then it might still have to wait a bit...

gcmoreira:
@ikelos 
I don't think there is a way to check the dwarf2json version used to generate the ISF from the plugin, right? It would be cool. If that's possible, we should also report the "ISF producer" and its version in the backtrace/crash, so when people copy/paste the output for a bug we already know which version they used. 

Additionally, it would be awesome to have what you mentioned in #985
> We may also need to find a way to let plugins request a specific version of an ISF file

Having said that, this plugin works with kernels < 6.3 without any d2j version requirement. Although, it was only tested with d2j 0.6.0. But for kernels >= 6.3, it requires dwarf2json >= 0.7.0. 

I will test it soon, but reading the code it seems that with a kernel >= 6.3 and the dwarf2json < 0.7.0 it will fail but not mentioning anything about the d2j version. I will see how to do it properly and will update this.
Thanks

ikelos:
I *think* we can ask for metadata on the ISF file, which *should* contain a producer and a version number (but there's nothing to stop people having made manual ones, etc, so we'd have to match on the producer name and then interpret the version number and just accept anything else)...

So specifically for IntermediateSymbolTables (not any other kind of symbol table) there is a metadata field, which contains the information from the ISF, which I believe contains the producer name and a producer version (in string form).  So we *should* be able to figure all that stuff out.  What I might do is commit this fix as is, and then we do a second PR that adds the version checking in once we get it in place, how does that sound?  I think that would leave us in the same position we're in at the moment in the case of failure, but make it more likely to work in some of the situations?

Lemme know if that works, and if so we can get this merged (although the conflict will need fixing up first please).

gcmoreira:
That sounds amazing. Let's do that. I will create a new PR once we are able to check the d2j version. Thanks @ikelos

Issue #981, opened 2023-07-06
Unify Linux plugins representation of pids


eve-mem:
**Is your feature request related to a problem? Please describe.**
The "pid" as shown by different plugins is not consistent. Either displaying the `pid` or `tgid` value from the `tast_struct`.

For example the pslist plugin uses `tgid`:
https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/pslist.py#L83

While malfind will report `pid`:
https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/malfind.py#L83

And in fact if you wanted to use a pid filter for plugins, that is actually using `pid` rather than `tgid`:
https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/pslist.py#L63

**Describe the solution you'd like**
Picking one and sticking with it across the plugins. I'm happy to make the changes in a PR when a choice is made. 

I think the biggest question is what would actually be preferred? 

`tgid` is the most technically correct "pid" to display, but using that value and calling it `pid` in plugin outputs might be confusing to some? It'll catch people out when writing plugins or working with volshell. 

However if all changes to `pid` are swapped to `tgid` and now none of the plugins report a pid result. That will be confusing to people that doesn't yet know what `tgid` is. They might want to find a process by a pid they have from logs etc, but it's now not obvious how to do that in vol.

I'm not sure personally what the right answer is, although I think I lean to changing all plugins to use the `tgid` value but reporting it as `pid`.  We're using "pid" to mean "unique id for a process to use between plugins and different sources of data" and `tgid` does exactly that. It's what `ps` displays after all. It's just slightly confusing as the `task_struct` has a field called `pid` that might seem more natural to use. 

We could go to the extreme of adding a `get_pid()` function to the task_stuct in the Linux extensions to keep it all central? 

This is probably quite a minor point, but I'd personally love to see consistency across the plugins. 

I'd be really interested in your thoughts. Thank you.

ikelos:
I don't know enough about the structures to know the difference between a `tgid` and a `pid`.  I'm happy to defer to @atcuno decision on this front?

ikelos:
There was a suggestion at one point of potentially abstracting the concept of a process *from any operating system* into a single class in some way, but I dunno if anyone has the energy or desire to dig into all the details of doing that (or whether it would be a very useful thing to do)?

eve-mem:
[...]

PR #1000, opened 2023-09-04
Prototype generic process functions

eve-mem:
Hello,

This draft PR is to add some experimental features to provide a generic way of getting basic information from processes regardless of operating system. The idea was discussed in the comments of this issue: https://github.com/volatilityfoundation/volatility3/issues/981

The idea being to provide a set of functions that can be used in plugins to make them easier to make (or volshell), and be consistent across the different operating systems so it's easier to switch between them.

It is very basic, just adding the following functions. Mostly just be shifting code out of various pslist plugins to the extensions so they can used from anywhere. 

| **Function**    | **Windows** | **Linux** | **Mac** |
|-----------------|-------------|-----------|---------|
| get_pid         | 🟢| 🟢| 🟢 |
| get_parent_pid  | 🟢| 🟢| 🟢|
| get_name        | 🟢| 🟢| 🟢|
| get_create_time | ⚫ | 🔴| 🟢|
| get_exit_time   | ⚫| 🔴| 🔴 |

(This is what I mean with the colours if it's not clear - 🟢: Added in this PR, ⚫: Function already existed, 🔴: Not added in this PR)

It was interesting to work on this. I noticed after that the [linux module class](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/symbols/linux/extensions/__init__.py#L28C4-L28C4) also inherits from `GenericIntelProcess` and already has a `get_name` function. 

Also that the windows EPROCESS already had the `get_create_time` and `get_exit_time` functions so I used those names in the generic part. 

I've then modified the pslist plugins for windows/linux/mac to show how this could be used. If this is useful I'm happy to modify the existing plugins to use this, it shouldn't affect how they work - but might make them easier to read. 

I'm not sure if this is the best (or even a good) way of doing this - so I'd be very interested in your thoughts. 

Thanks!


eve-mem:
@ikelos - I'd love your thoughts on this. The idea to have a simple to use way to get key information about a process for plugins. That way there can be a central way to do these lookups within plugins so that they're consistent and if an update needs to be done it only needs to happen in one place. It could be bulked out to include extra bits for a process like getting memory regions etc.

It should hopefully mean that we'd be able to make non-OS specific plugins - assuming it's done in the right way.  Given it's above intel process that might be a good place for it given the arm work at the moment?

Thanks again.

ikelos:
Yep, sorry, it is on my list, just a bit lower down at the moment, sorry!  5:S

volatility3/framework/symbols/mac/extensions/__init__.py
```diff
@@ -108,6 +109,22 @@ def get_process_memory_sections(
...
+        return self.p_ppid
+
+    def get_name(self) -> str:
+        return utility.array_to_string(self.p_comm)
+
+    def get_create_time(self) -> datetime.datetime:
+        start_time_seconds = self.p_start.tv_sec
+        start_time_microseconds = self.p_start.tv_usec
+        return datetime.datetime.fromtimestamp(
+            start_time_seconds + start_time_microseconds / 1e6
```
gcmoreira: We should use aware datetimes here, check https://github.com/volatilityfoundation/volatility3/pull/1222.
For instance, see how `unixtime_to_datetime` is used in the [timespec64](https://github.com/volatilityfoundation/volatility3/blob/e76d51251be922bb364f10f3fc04f7bfe06c759a/volatility3/framework/symbols/linux/extensions/__init__.py#L1837) extension objects.

Additionally, we have talked with @ikelos about moving these functions out of renderes.. see this [FIXME](https://github.com/volatilityfoundation/volatility3/blob/e76d51251be922bb364f10f3fc04f7bfe06c759a/volatility3/framework/renderers/conversion.py#L14) ..  but haven’t gotten around to it yet. Want to be the hero who makes it happen? :)

gcmoreira:
Hey @eve-mem ! It's a great idea to ensure better consistency across the framework.

I think this is also related to your other ticket [Unify Linux plugins representation of pids](https://github.com/volatilityfoundation/volatility3/issues/981) which we should take a decision at some point.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -258,6 +258,18 @@ def get_threads(self) -> Iterable[interfaces.objects.ObjectInterface]:
         ):
             yield task
 
+    def get_pid(self) -> int:
+        """Returns the pid of this process"""
+        return self.tgid
+
+    def get_parent_pid(self) -> int:
+        """Returns the pid of parent of this process"""
+        return self.parent.get_pid() if self.parent else 0
```
gcmoreira: This is something I have in my TO DO list. We should use `task_struct.real_parent` instead for this. For instance, see the [getppid() ](https://elixir.bootlin.com/linux/v6.10.9/source/kernel/sys.c#L989) syscall. A task's `parent` and `real_parent` can differ, for instance when a task is ptracing another. In this case, the tracee's `parent` refers to the tracer, while the `real_parent` remains the original (real) parent task. I know we use `task.parent` for this purpose everywhere in the framework, but that needs to be changed. 

If you are going to replace all instances of `task.parent.tgid` (or `task.parent.pid` pending discussion https://github.com/volatilityfoundation/volatility3/issues/981) with `get_parent_pid()`, I think this is a good opportunity to also fix this issue.

eve-mem:
Thanks both. I completely understand not having time to look at this given how little time i seem to have at the moment. 

Thanks for those pointers @gcmoreira - that's almost exactly why I'd pike something like this PR to exist. Just a single place to have the 'right' answer so we can do it once for all plugins. Not necessarily this version in particular, but the idea at least.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -374,6 +374,35 @@ def get_threads(self) -> Iterable[interfaces.objects.ObjectInterface]:
...
+        # {
+        # 	int pid;
+
+        # 	rcu_read_lock();
+        # 	pid = task_tgid_vnr(rcu_dereference(current->real_parent));
+        # 	rcu_read_unlock();
+
+        # 	return pid;
+        # }
+        return self.real_parent.get_pid() if self.parent else 0
```
gcmoreira: `self.real_parent` in both. Additionally, `parent` and `real_parent` are pointers. So, it could be useful to also check the pointer is readable.
```python
return self.real_parent.get_pid() if self.real_parent and self.real_parent.is_readable() else 0
```

Lastly, there is no need to paste the entire code in the comments. Simply referencing the kernel function name will suffice. Including it could cause confusion, as someone might assume we need synchronization primitives like locks or mutexes, even though thread-safety isn't an issue for us.

volatility3/framework/symbols/linux/extensions/__init__.py
eve-mem: Nice -  I did not know that - I've done this now.

eve-mem:
Hey @ikelos and @gcmoreira,

I was working on the PR for https://github.com/volatilityfoundation/volatility3/issues/981 and it just felt to me that doing this first might make it all a bit easier. 

Assuming that you both think that this is a good idea of course! What do you think?

I also tried to add an explanation of tgid/pid/tid into the readme docs. Not sure what you both make of that?

@gcmoreira thanks for the comments - I think I've got them all now. I'm a little unsure I got the datetime part correct in the mac side - could you check that a little closer?

Thanks both.
:fox_face:

gcmoreira:
Hey @eve-mem good job!
Yeah the aware datetime change is correct.

I think it's a solid idea, but we should keep refining the implementation. There are a few areas that could be improved, and some issues that still need fixing.

- `GenericProcess`: 
  - Should include methods to return all TGID and PID (user PID and TID) and their respective parents. Not only the Thread Group Leader, otherwise we would be burying the threads behind that interface.
  - I don't think the method names do much to clear up the confusion between the kernel PID/TGID/TID/etc. In fact, it feels like we're just shifting the problem from one class to another. Even the docstrings are not helping at all to that purpose. We should be more explicit with the method names. We should choose a point of view and stick to it throughout the framework. If we decide to take the user-space's perspective, I think it should be better to use names like `get_user_pid()`, `get_user_tid()`, and so on. For instance, see below that with the current implementation (in red) even with the comment clarifying it, it's still confusing. This can be much clear if we use more explicit method names and include the user TID (green changes).
 ```diff
-    user_pid = task.get_pid()
-    user_tid = task.pid  # Note that pid is user_tid in Linux
-    user_ppid = task.get_parent_pid()
+    user_pid = task.get_user_pid()
+    user_tid = task.get_user_tid()
+    user_ppid = task.get_user_ppid() 
# or get_parent_user_pid()
```
- Filter: I strongly believe it's a mistake to filter by user_pid (TGID). Actually, with the proposed implementation we are changing the way Volatility3 is working now, since we are currently filtering by user_tid (PID). If we change to use user_pid (TGID) that makes impossible to filter by TID. Also, note that this affects not just the pslist plugin `--filter` argument, which can now be superseded by the more general `--filters` argument and maybe removed, but with the changes in `create_pid_filter()` [here](https://github.com/volatilityfoundation/volatility3/pull/1000/files#diff-b9e2042cc4ddf10930db5c82310d7758cc39475ff8745f73e5285e50bd2553e4R75) this is affecting other current and future plugins which may require to get a specific thread and not the thread group leader. With this change we are introducing a bug and no one in the framework will be able to find a specific thread. I explained this with some examples [here]( https://github.com/volatilityfoundation/volatility3/issues/981#issuecomment-1851479368). I believe it's quite clear, but if it's not, please let me know, and I'll do my best to clarify that explanation.
  - On the contrary, even for users unfamiliar with the meaning of "a pid" in the framework, using user_tid (PID) will yield the expected result, since `user_pid == user_tid` (or in kernel space `task_struct.tgid == task_struct.pid`) for a thread group leader (what is typically referred to as 'a process' in user-space).
- We should also extend these changes to the `create_pid_filter()` method name to stick to the proposed point of view. With the Volatility3's current behavior (filtering by user_tid) it should be renamed to `create_user_tid_filter()`
- Review all occurrences of the term `pid` in the framework and adjust accordingly to align with the proposed perspective.
- Lastly, we should avoid using the abbreviation `upid` for `user_pid`, as it has a specific meaning in the Linux kernel see [here](https://elixir.bootlin.com/linux/v6.11.3/source/include/linux/pid.h#L50). I'm not saying you are already doing this, I just want to flag this in case we feel tempted to use it. I might be the first to slip up on my own words LOL! That's why peer reviews are crucial.

**Comment by eve-mem** [comment]
Thanks @gcmoreira - makes sense I'll work on it. Particularly the pid filtering, it makes sense to me now.

I was hoping that the functions would be the same across the different OSes, I'll change the get_pid to get_user_pid for all of them.

**Comment by eve-mem** [comment]
Hey @gcmoreira - I was updating the linux side of the generic processes as per your suggestion. 

e.g.
```diff
-    user_pid = task.get_pid()
-    user_tid = task.pid  # Note that pid is user_tid in Linux
-    user_ppid = task.get_parent_pid()
+    user_pid = task.get_user_pid()
+    user_tid = task.get_user_tid()
+    user_ppid = task.get_user_ppid() 
# or get_parent_user_pid()
```

It means that then the linux side will have slightly different functions to the other OSes - meaning it's less generic again. I'm hoping it means a plugin could run on both linux/windows/macos without any changes. 

What do you think about having the linux side having  `get_user_pid` and `get_user_tid` as you described, but keeping `get_pid` as a wrapper simple wrapper that then simply calls `get_user_pid`.

It means there's still only one place for the actual logic for `get_user_pid` etc - so a single place for the logic if it ever needed to be updated. It let's people use the generic `get_pid` for generic plugins, and then still allows the use of `get_user_pid` in the actual linux plugins.

Would love to hear your thoughts on whether this simplifies things or adds unnecessary layers!

ikelos:
I think an interface should be generic, with the common usage (so whilst there may be non-user pids and tids and ppids, we kind of all have an intuition of what we mean, and the docstring can explain precisely what's returned, and we can overrload OS specific ones with other OS specific functions if we really need (so people don't have to jump into and out of the handlers to do different things, but the only ones they can rely on across multiplatform methods are the ones on the generic interface.  This will mean choosing what we expect to be "common" carefully, because changing it later will be nigh on impossible...

doc/source/getting-started-linux-tutorial.rst
```diff
@@ -191,3 +191,22 @@ Now to find the commands that were run in the bash shell by using ``linux.bash``
         1733    bash    2020-01-16 14:00:36.000000      sudo apt upgrade
         1733    bash    2020-01-16 14:00:41.000000      chmod +x meterpreter
         1733    bash    2020-01-16 14:00:42.000000      sudo ./meterpreter
+
+PIDs, TIDs, and TGID
+--------------------
+Typically processes are a collection of threads that are working together for a single program to run. On Windows `notepad.exe` would be a process that might contain multiple threads, in task manager you would only see the single process. In Linux you will also see processes split into multiple threads working together, however Linux uses a single `task` structure to represent processes and the threads that make them up.
```
atcuno: Linux doesn't use a single task struct, threads have their own. The code reflects this but this part of the documentation is off:

"uses a single `task` structure to represent processes and the threads"

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -374,6 +374,23 @@ def get_threads(self) -> Iterable[interfaces.objects.ObjectInterface]:
...
+        """Returns the pid of this process"""
+        return self.tgid
+
+    def get_parent_pid(self) -> int:
+        """Returns the pid of parent of this process"""
+        # Uses real_parent rather than parent to match Linux kernel getppid
+        return (
+            self.real_parent.get_pid()
+            if self.real_parent and self.real_parent.is_readable()
+            else 0
```
atcuno: this should be a renderer unavailable instance as 0 is misleading

PR #1001, opened 2023-09-05
Making cred optional as it didn't exist in old 2.6 kernels

xabrouck:
When running volatility3 on an old system with a 2.6 kernel, I get this error message:

`ValueError: Symbol type not in LintelStacker1 SymbolTable: cred`

This patch is to make `cred` optional, which allows volatility to run fine on that old kernel.


ikelos (approved):
Yep, this looks good, sorry for the delay in reviewing!

Issue #1002, opened 2023-09-05
No module named 'volatility3.cli'


ShutdownRepo:
# Describe the bug
`ModuleNotFoundError` error raised when trying to run volatility after a fresh install

# To Reproduce
Error is raised after install with either of the following methods

```bash
python3 -m pipx install git+https://github.com/volatilityfoundation/volatility3
vol --help
```

or

```bash
git -C /opt/tools clone https://github.com/volatilityfoundation/volatility3.git
python3 -m pipx install /opt/tools/volatility3
vol --help
```

# Example error output
```
Traceback (most recent call last):
  File "/root/.local/bin/vol", line 5, in <module>
    from volatility3.cli import main
ModuleNotFoundError: No module named 'volatility3.cli'
```

ShutdownRepo:
I think it was introduced in b4c6b661f01fc3dde54362a4f55be4d89e4cc6e5
This install would work
```bash
git -C /opt/tools/ clone https://github.com/volatilityfoundation/volatility3
git -C /opt/tools/volatility3 revert b4c6b661f01fc3dde54362a4f55be4d89e4cc6e5
python3 -m pipx install /opt/tools/volatility3
vol --help
```

ikelos:
Thanks very much for letting us know.  Could you please test the proposed fix in #1003 and let us know if it resolves the problem?

qu35t-code:
Hey, fix on issues/issue1002 doesn't works

```bash
> git checkout issues/issue1002

Branch 'issues/issue1002' set up to track remote branch 'issues/issue1002' from 'origin'.
Switched to a new branch 'issues/issue1002'

> python3 -m pipx install .
  installed package volatility3 2.5.0, installed using Python 3.9.2
  These apps are now globally available
    - vol
    - volshell
done! ✨ 🌟 ✨

> vol --help
Traceback (most recent call last):
  File "/root/.local/bin/vol", line 5, in <module>
    from volatility3.cli import main
ModuleNotFoundError: No module named 'volatility3.cli'
```


If we revert to this commit it works : b4c6b661f01fc3dde54362a4f55be4d89e4cc6e5

```bash
> git -C /opt/tools/volatility3 revert b4c6b661f01fc3dde54362a4f55be4d89e4cc6e5
[develop 5458cef0] Revert "Core: Include only volatility3 in distributions packages"
 1 file changed, 1 insertion(+), 1 deletion(-)

> python3 -m pipx uninstall volatility3
[...]

PR #1004, opened 2023-09-05
Adding test install workflow for GitHub Actions pipelines

ShutdownRepo:
Added a workflow that makes sure volatility installs and runs as intended for all push and pull request triggers.
Will help avoid issues like https://github.com/volatilityfoundation/volatility3/issues/1002 in the future

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        host: [ ubuntu-latest, macOS-latest, windows-latest ]
+        host: [ ubuntu-latest, windows-latest ]
         python-version: [ "3.7", "3.8", "3.9", "3.10", "3.11" ]
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -1,0 +1,29 @@
+name: Test install Volatility3
+on: [push, pull_request]
+jobs:
+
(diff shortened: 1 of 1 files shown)


ikelos:
@npetroni   Could you please look this over and check it's all ok?

ShutdownRepo:
Sorry for the multiple commits, but we're good to go now
The pipeline runs on a matrix strategy with multiple operating systems and multiple python versions, to make sure volatility installs and runs smoothly everywhere
Here is an example run
image

ShutdownRepo:
Hello there, any news on this? Do you need me to change something?

npetroni:
> Hello there, any news on this? Do you need me to change something?

It looks like this is going to fail on the macOS hosts. Would you like to try to fix it or remove those from the PR?
https://github.com/volatilityfoundation/volatility3/actions/runs/6215300225

ShutdownRepo:
In my opinion, macOS should be removed from the pipeline as long as there is no installation guidance specific to this OS for volatility. The reason it's not building is it's missing a couple dependencies and vol3 docs don't address this. I also think vol3 will be mostly be used by win & linux users and not so much macOS. And even if macOS users want to, the best course of actions for them is probably to use a container. I'm a macOS user myself and that's what I'd probably do.
TL;DR: removing macOS from the list, could be added later on if it's in volatility3's roadmap to have a clear install process for this OS.

npetroni (approved):
This change works for me.

PR #1015, opened 2023-10-08
Fix mac pslist

Abyss-W4tcher:
- Volatility3 version used : 2.5.1
- Memory dump : Mac OS 10.8.5_build-12F45

Executing `mac.pslist` or any plugin needing it results in : 

```sh
$ vol3 [...] mac.pslist
Volatility 3 Framework 2.5.1	Stacking attempts finished                  

PID	Process	Argc	Arguments
  File "/volatility3/vol.py", line 10, in <module>
    volatility3.cli.main()
  File "/volatility3/volatility3/cli/__init__.py", line 790, in main
    CommandLine().run()
  File "/volatility3/volatility3/cli/__init__.py", line 447, in run
  File "/volatility3/volatility3/cli/text_renderer.py", line 193, in render
    grid.populate(visitor, outfd)
  File "/volatility3/volatility3/framework/renderers/__init__.py", line 241, in populate
  File "/volatility3/volatility3/framework/plugins/mac/psaux.py", line 41, in _generator
    for task in tasks:
  File "/volatility3/volatility3/framework/plugins/mac/pslist.py", line 194, in list_tasks_tasks
    for task in queue_entry.walk_list(queue_entry, "tasks", "task"):
  File "/volatility3/volatility3/framework/objects/__init__.py", line 968, in __getattr__
    raise AttributeError(
AttributeError: ClassType has no attribute: symbol_table_name1!queue_head_t.walk_list
```

Here is the symbol from the ISF file : 

```json 
"tasks": {
    "type": {
    "kind": "struct",
    "name": "queue_entry"
    },
    "offset": 40
}
```

After investigating the code, I found out that instantiating `queue_entry` to "queue_entry" type manually gave it the property "walk_list" : 

```diff
diff --git a/volatility3/framework/plugins/mac/pslist.py b/volatility3/framework/plugins/mac/pslist.py
index 88045a27..c0e149fc 100644
--- a/volatility3/framework/plugins/mac/pslist.py
+++ b/volatility3/framework/plugins/mac/pslist.py
@@ -188,7 +188,7 @@ class PsList(interfaces.plugins.PluginInterface):
 
         kernel_layer = context.layers[kernel.layer_name]
 
-        queue_entry = kernel.object_from_symbol(symbol_name="tasks")
+        queue_entry = kernel.object("queue_entry", kernel.get_symbol("tasks").address)
 
         seen: Dict[int, int] = {}
         for task in queue_entry.walk_list(queue_entry, "tasks", "task"):
```

Patch is based on Volatility2 implementation of mac_pslist plugin : 

```python
# https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/mac/pstasks.py#L35
def allprocs(self):
    common.set_plugin_members(self)
    tasksaddr = self.addr_space.profile.get_symbol("_tasks")
    queue_entry = obj.Object("queue_entry", offset = tasksaddr, vm = self.addr_space)
    seen = { tasksaddr : 1 }
    for task in queue_entry.walk_list(list_head = tasksaddr):
        # [...]
```

Fix seems retrocompatible to me, and should not break the existing implementation.

### DIFF
--- a/volatility3/framework/plugins/mac/pslist.py
+++ b/volatility3/framework/plugins/mac/pslist.py
@@ -188,7 +188,7 @@ class PsList(interfaces.plugins.PluginInterface):
 
         kernel_layer = context.layers[kernel.layer_name]
 
-        queue_entry = kernel.object_from_symbol(symbol_name="tasks")
+        queue_entry = kernel.object("queue_entry", kernel.get_symbol("tasks").address)
 
         seen: Dict[int, int] = {}
         for task in queue_entry.walk_list(queue_entry, "tasks", "task"):
--- a/volatility3/framework/plugins/mac/pslist.py
+++ b/volatility3/framework/plugins/mac/pslist.py
@@ -49,9 +49,7 @@ class PsList(interfaces.plugins.PluginInterface):
         ]
 
     @classmethod
-    def get_list_tasks(
-        cls, method: str
-    ) -> Callable[
+    def get_list_tasks(cls, method: str) -> Callable[
         [interfaces.context.ContextInterface, str, Callable[[int], bool]],
         Iterable[interfaces.objects.ObjectInterface],
     ]:
@@ -188,9 +186,7 @@ class PsList(interfaces.plugins.PluginInterface):
 
         kernel_layer = context.layers[kernel.layer_name]
 
-        queue_entry = kernel.object(
-            object_type="queue_entry", offset=kernel.get_symbol("tasks").address
-        )
+        queue_entry = kernel.object_from_symbol(symbol_name="tasks")
 
         seen: Dict[int, int] = {}
         for task in queue_entry.walk_list(queue_entry, "tasks", "task"):
--- a/volatility3/framework/plugins/mac/pslist.py
+++ b/volatility3/framework/plugins/mac/pslist.py
@@ -188,7 +188,9 @@ class PsList(interfaces.plugins.PluginInterface):
 
         kernel_layer = context.layers[kernel.layer_name]
 
-        queue_entry = kernel.object(object_type="queue_entry", offset=kernel.get_symbol("tasks").address)
+        queue_entry = kernel.object(
+            object_type="queue_entry", offset=kernel.get_symbol("tasks").address
+        )
 
         seen: Dict[int, int] = {}
         for task in queue_entry.walk_list(queue_entry, "tasks", "task"):
--- a/volatility3/framework/plugins/mac/pslist.py
+++ b/volatility3/framework/plugins/mac/pslist.py
@@ -188,7 +188,7 @@ class PsList(interfaces.plugins.PluginInterface):
 
         kernel_layer = context.layers[kernel.layer_name]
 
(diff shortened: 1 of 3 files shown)


ikelos:
Hmmm, that feels more like the `object_from_symbol` call not honouring the class_overrides, which is strange, so I'm going to do a little more digging.  There may be another way of resolving this that catches more than just this one instance, but if not the change looks ok...  5:)

ikelos:
Sorry, still intending to look at this!  Hopefully I'll get some time next week to do so...

Abyss-W4tcher:
Using the ISF available [here](https://github.com/Abyss-W4tcher/volatility3-symbols/raw/master/macOS/10.9/macOS_KDK_10.9.3_build-13D65.json.xz), I noticed the expected symbol type `queue_entry` was in fact `queue_head_t` : 

```sh
$ jq '.symbols.tasks' macOS_KDK_10.9.3_build-13D65_formatted.json                                                     
{
  "type": {
    "kind": "struct",
    "name": "queue_head_t"
  },
  "address": 18446743523962958000
}
```

This a strange behaviour, as the corresponding Volatility2 profile specifies `queue_entry` : 
```python
'tasks': [0x28, ['queue_entry']]
```

Both structures point to the same thing, as indicated in the [documentation](https://developer.apple.com/documentation/kernel/queue_head_t) : 

image


I reverted the patch in `pslist` plugin, and added a type class in the Mac symbols init. This should resolve the issue globally, instead of this specific case.

Abyss-W4tcher:
Just added a patch against a problem where the `queue_entry` or `queue_head_t` was walked past the linked list end. Issue : 

```python
# We are comparing an integer to an object :
while (queue_element is not None and queue_element.vol.offset != list_head):
```

This resulted in the following output, even sometimes crashing due to the datetime (`mac-sample-1.bin`) : 


```sh
Volatility 3 Framework 2.6.1
  |         OFFSET |             NAME |       PID | UID | GID |                  Start Time | PPID
* | 0xff800dcfccc0 |      kernel_task |         0 |   0 |   0 | 2014-06-20 21:43:15.140294  |    0
* | 0xff801378ea80 |          launchd |         1 |   0 |   0 | 2014-06-20 21:43:15.141776  |    0
[...]
* | 0xff80155d5e90 | QuickLookSatelli |       759 | 501 |  20 | 2014-07-01 06:23:25.382745  |    1
* | 0xff801324ab80 |                  | 324640768 |   0 |   4 | 4692-01-16 09:25:20.100006  |    0
```

ikelos:
Thanks, I'm reasonably happy with this and this looks like an ok way of dealing with the renaming of a structure, but I'm wondering, given the structure can change, whether it makes more sense to have a separate walker method that takes the structure as a parameter, rather than tying to the structure with multiple names?  So it would become `plugin.walk_queue(queue_struct, params, ...)` rather than `queue_struct.walk_queue(params, ...)`...

Pretty happy to do either, but just want to check if anyone has any strong preferences?  Specifically @atcuno and @gcmoreira since this likely most affects them...

ikelos (approved):
Checked with @atcuno

PR #999, opened 2023-08-30
Windows: Update pslist.py, add friendly option

616c696365:
Hey :) Addition of --friendly option to windows.pslist. It will work the same without the --friendly option. Here's an example when you run it:
```
volatility3 % python3 vol.py -f ../dumps/memory-images/xp-laptop-2005-07-04-1430.img windows.pslist --dump --friendly
Volatility 3 Framework 2.5.0
Progress:  100.00		PDB scanning finished
PID	PPID	ImageFileName	Offset(V)	Threads	Handles	SessionId	Wow64	CreateTime	ExitTime	File output

4	0	System	0x823c87c0	62	1133	N/A	False	N/A	N/A	Error outputting file
400	4	smss.exe	0x8214b020	3	21	N/A	False	2005-07-04 18:17:26.000000 	N/A	400.smss.exe.0x48580000.dmp
456	400	csrss.exe	0x821c11a8	11	551	0	False	2005-07-04 18:17:29.000000 	N/A	456.csrss.exe.0x4a680000.dmp
480	400	winlogon.exe	0x814dc020	18	522	0	False	2005-07-04 18:17:29.000000 	N/A	480.winlogon.exe.0x1000000.dmp
524	480	services.exe	0x815221c8	17	321	0	False	2005-07-04 18:17:30.000000 	N/A	524.services.exe.0x1000000.dmp
536	480	lsass.exe	0x821d8248	20	369	0	False	2005-07-04 18:17:30.000000 	N/A	536.lsass.exe.0x1000000.dmp
680	524	svchost.exe	0x814f0020	19	206	0	False	2005-07-04 18:17:31.000000 	N/A	680.svchost.exe.0x1000000.dmp
760	524	svchost.exe	0x821daa88	10	289	0	False	2005-07-04 18:17:31.000000 	N/A	760.svchost.exe.0x1000000.dmp
800	524	svchost.exe	0x821463a8	75	1558	0	False	2005-07-04 18:17:31.000000 	N/A	800.svchost.exe.0x1000000.dmp
840	524	Smc.exe	0x8216c9b0	22	421	0	False	2005-07-04 18:17:32.000000 	N/A	840.Smc.exe.0x400000.dmp
932	524	svchost.exe	0x81530228	6	93	0	False	2005-07-04 18:17:33.000000 	N/A	932.svchost.exe.0x1000000.dmp
[...]

PR #1023, opened 2023-10-17
Core: Add (optional) sanitization to the FileHandler class

ikelos:
This is a first cut at ensuring filename writes are safe.  This completely excludes unicode filenames, which will not be useful for a large chunk of forensics, which is why it's being made optional to each plugin before we move forward with it.  Whitelists are much more effective than blacklists, but I'm not certain how to allow unicode characters without potentially tripping over something a filesystem handles poorly, so for now its an explicit latin-1 whitelist.


ikelos:
Merging this just to keep the history (it was a little bit naughty for the changes to have been included in #999).  5:)

PR #1027, opened 2023-10-31
fix guid and pdb_name for #1026

### LINKED ISSUE #1026: get_guid_from_mz not compatible with newer versions of PEFile
**Describe the bug**
It looks like this commit in `pefile` might have broken `PDBUtility.get_guid_from_mz` since the signature has changed to follow GUID standards. Fortunately, this means the guid can just be referenced by the new `Signature_String` attribute. (https://github.com/erocarrera/pefile/commit/d0d6aafd6e3fcaec8a07d2dfef2a2b7adda18269), 

**Context**
Volatility Version:  2.5.0
Operating System:  N/A
Python Version:  N/A
Suspected Operating System:  N/A
Command:  

**To Reproduce**

1. Call `PDBUtility.get_guid_from_mz` from another plugin.
2. Watch it fail

**Expected behavior**
Calling `PDBUtility.get_guid_from_mz` is successful.

**Example output**
```
  File "/home/dcm/dev/volatility3/vol.py", line 10, in <module>
    volatility3.cli.main()
  File "/home/dcm/dev/volatility3/volatility3/cli/__init__.py", line 790, in main
    CommandLine().run()
  File "/home/dcm/dev/volatility3/volatility3/cli/__init__.py", line 447, in run
  File "/home/dcm/dev/volatility3/volatility3/cli/text_renderer.py", line 193, in render
    grid.populate(visitor, outfd)
  File "/home/dcm/dev/volatility3/volatility3/framework/renderers/__init__.py", line 245, in populate
  File "/home/dcm/dev/volatility3/development/clr.py", line 199, in _generator
    for strong_name, version in self.get_modules_by_process(
  File "/home/dcm/dev/volatility3/development/clr.py", line 135, in get_modules_by_process
    result = pdbutil.PDBUtility.get_guid_from_mz(context, entry.vol.layer_name, entry.DllBase)
  File "/home/dcm/dev/volatility3/volatility3/framework/symbols/windows/pdbutil.py", line 235, in get_guid_from_mz
    binascii.hexlify(debug_entry.Signature_Data4),
TypeError: a bytes-like object is required, not 'int'
```

**Additional information**
Seems like a simple PR. I can contribute.

daddycocoaman:
Addresses #1026. Uses the Signature_String to grab the GUID and ensures that the pdb_name is only the filename, not the whole absolute path.

### DIFF
--- a/volatility3/framework/symbols/windows/pdbutil.py
+++ b/volatility3/framework/symbols/windows/pdbutil.py
@@ -9,6 +9,7 @@ import lzma
 import os
 import re
 import struct
+from pathlib import PureWindowsPath
 from typing import Any, Dict, Generator, List, Optional, Tuple, Union
 from urllib import parse, request
 
@@ -226,13 +227,12 @@ class PDBUtility(interfaces.configuration.VersionableInterface):
             return None
 
         pdb_name = debug_entry.PdbFileName.decode("utf-8").strip("\x00")
+
+        # Let pathlib do the filename extraction. This will likely always be a Windows path though.
+        pdb_name = PureWindowsPath(pdb_name).name
+
         age = debug_entry.Age
-        guid = "{:08x}{:04x}{:04x}{}".format(
-            debug_entry.Signature_Data1,
-            debug_entry.Signature_Data2,
-            debug_entry.Signature_Data3,
-            binascii.hexlify(debug_entry.Signature_Data4).decode("utf-8"),
-        )
+        guid = debug_entry.Signature_String[:32]  # Removes the Age from the GUID
         return guid, age, pdb_name
 
     @classmethod
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 # The following packages are required for core functionality.
-pefile>=2017.8.1
+pefile>=2023.2.7
 
 # The following packages are optional.
 # If certain packages are not necessary, place a comment (#) at the start of the line.
--- a/requirements-minimal.txt
+++ b/requirements-minimal.txt
@@ -1,2 +1,2 @@
 # These packages are required for core functionality.
-pefile>=2017.8.1 #foo
+pefile>=2023.2.7 #foo
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # The following packages are required for core functionality.
-pefile>=2017.8.1
+pefile>=2023.2.7
 
 # The following packages are optional.
 # If certain packages are not necessary, place a comment (#) at the start of the line.


daddycocoaman:
It might be a good idea to bump pefile to at least v2023.2.7 as well. Thoughts?

eve-mem:
Awesome, thanks for looking into it. Does your comment about it being a windows path mean it would break this from working on Linux etc?

daddycocoaman:
> Awesome, thanks for looking into it. Does your comment about it being a windows path mean it would break this from working on Linux etc? 

Nope. Actually the opposite. I'm dev-ing on Linux. 

It's just that `Path` will automatically resolve to whatever platform you're on and trying to parse a Windows path on Linux with the normal `Path` will fail. The PdbFIleNames from the Microsoft Symbol Server will always be a Windows path, AFAIK, so we have to treat it as a Windows path regardless of what platform the Vol3 user is on. There's a chance it might not always be on Windows but I'll probably be the one to find out soon. 👀

```
>>> import platform
>>> from pathlib import Path
>>> platform.system()
'Linux'

>>> cmd = Path("C:\System32\cmd.exe")
>>> cmd.name
'C:\\System32\\cmd.exe'

>>> cmd.stem
'C:\\System32\\cmd'

>>> cmd.parts
('C:\\System32\\cmd.exe',)

>>> from pathlib import PureWindowsPath
>>> cmd = PureWindowsPath("C:\System32\cmd.exe")
>>> cmd.name
'cmd.exe'

>>> cmd.stem
'cmd'

>>> cmd.parts
('C:\\', 'System32', 'cmd.exe')

>>> from pathlib import PurePath
>>> cmd = PurePath("C:\System32\cmd.exe")
>>> cmd.name
'C:\\System32\\cmd.exe'

>>> cmd.stem
'C:\\System32\\cmd'

>>> cmd.parts
('C:\\System32\\cmd.exe',)
```

eve-mem:
Great! It looks like pathlib is used in other parts of the framework so that shouldn't change anything there. I'm not sure what parts of vol use the function you've fixed so I can't comment on that side of things. 

Equally I've not been able to recreate your issue to do any testing, it looks like you were using a custom plugin. I don't know much about the windows side so maybe it'll be obvious to someone else to test in volshell or similar. 

I'm sure one of the core devs will take a look at this for you soon enough, thanks for being patient.

volatility3/framework/symbols/windows/pdbutil.py
```diff
@@ -226,13 +227,12 @@ def get_guid_from_mz(
...
+        pdb_name = PureWindowsPath(pdb_name).name
+
         age = debug_entry.Age
-        guid = "{:08x}{:04x}{:04x}{}".format(
-            debug_entry.Signature_Data1,
-            debug_entry.Signature_Data2,
-            debug_entry.Signature_Data3,
-            binascii.hexlify(debug_entry.Signature_Data4).decode("utf-8"),
-        )
+        guid = debug_entry.Signature_String[:32]  # Removes the Age from the GUID
```
eve-mem, line 235: Am I correct in thinking that `Signature_String` is new to pefile, and so wont be found in older versions? It's probably wise to either update the requirements file with the minimum version, or maybe add a try/except around it?

eve-mem (commented):
Just added a thought about the version of pefile that might be needed.

volatility3/framework/symbols/windows/pdbutil.py
daddycocoaman: Correct but there's no maximum value set for pefile in the requirements.txt, so anyone installing Vol3 is always getting the newest version anyway.

Might be time to update the minimum versions of dependencies.

eve-mem: Yes sorry that's what I mean, setting a minimum version in the requirements given that it would actually break now if there was an older version would probably be helpful.

ikelos: So whilst also requiring the minimum version on pefile, I just also want to check that the old guid values and the new guid values will match if the same file is provided to both the old code and the new code?  Happy to use the new method (and do less manually ourselves), but I want to ensure it returns exactly the same output...

ikelos (commented):
This looks good, just needs checking the guids match the old code, and adding in the minimum requirement on all the `requirement.txt` files

volatility3/framework/symbols/windows/pdbutil.py
daddycocoaman: As it stands now, the "old code" doesn't actually work at all in this function and probably hasn't since February. I use Rich for tracebacks, so errors are easier to understand but if you look at the ``debug_entry`` local here, you'll see the new pefile structure with 6 fields as opposed to 4, while the current Vol3 code still looks for only the 4 fields.

Current code:
```python
 /home/dcm/dev/volatility3/volatility3/framework/symbols/windows/pdbutil.py:239 in                │
│ get_guid_from_mz                                                                                 │
│                                                                                                  │
│   236 │   │   │   debug_entry.Signature_Data1,                                                   │
│   237 │   │   │   debug_entry.Signature_Data2,                                                   │
│   238 │   │   │   debug_entry.Signature_Data3,                                                   │
│ ❱ 239 │   │   │   binascii.hexlify(debug_entry.Signature_Data4).decode("utf-8"),                 │
│   240 │   │   )                                                                                  │
│   241 │   │   # guid = debug_entry.Signature_String[:32]  # Removes the Age from the GUID        │
│   242 │   │   return guid, age, pdb_name                                                         │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │                  age = 1                                                                     │ │
│ │                  cls = <class 'volatility3.framework.symbols.windows.pdbutil.PDBUtility'>    │ │
│ │              context = <volatility3.framework.contexts.Context object at 0x7f04b1a04fa0>     │ │
│ │           debug_data = <pefile.DebugData object at 0x7f04aa808c10>                           │ │
│ │          debug_entry = <Structure: [CV_INFO_PDB70] 0x15B7F0 0x0 CvSignature: RSDS 0x15B7F4   │ │
│ │                        0x4 Signature_Data1: F9A24309 0x15B7F8 0x8 Signature_Data2: A184      │ │
│ │                        0x15B7FA 0xA Signature_Data3: C71B 0x15B7FC 0xC Signature_Data4: 78   │ │
│ │                        0x15B7FD 0xD Signature_Data5: 85 0x15B7FE 0xE Signature_Data6:        │ │
│ │                        047551E716D6 0x15B804 0x14 Age: 0x1 0x15B808 0x18 PdbFileName:        │ │
│ │                        ntdll.pdb>                                                            │ │
│ │                index = 3                                                                     │ │
│ │                layer = <volatility3.framework.layers.intel.WindowsIntel32e object at         │ │
│ │                        0x7f04aa3bbe80>                                                       │ │
│ │           layer_name = 'layer_name_Process13728_1'                                           │ │
│ │             max_size = 2191360                                                               │ │
│ │               mz_sig = b'MZ'                                                                 │ │
│ │      nt_header_start = 224                                                                   │ │
│ │               offset = 140725005778944                                                       │ │
│ │ optional_header_size = 240                                                                   │ │
│ │             pdb_name = 'ntdll.pdb'                                                           │ │
│ │              pe_data = <pefile.PE object at 0x7f04aa451360>                                  │ │
│ │            pe_header = b'MZ\x90\x00\x03\x00\x00\x00\x04\x00\x00\x00\xff\xff\x00\x00\xb8\x00… │ │
│ │               pe_sig = b'PE'                                                                 │ │
│ │               pefile = <module 'pefile' from                                                 │ │
│ │                        '/home/dcm/dev/volatility3/.venv/lib/python3.10/site-packages/pefile… │ │
│ │        physical_data = b'MZ\x90\x00\x03\x00\x00\x00\x04\x00\x00\x00\xff\xff\x00\x00\xb8\x00… │ │
│ │                 sect = <Structure: [IMAGE_SECTION_HEADER] 0x378 0x0 Name: .reloc 0x380 0x8   │ │
│ │                        Misc: 0x628 0x380 0x8 Misc_PhysicalAddress: 0x628 0x380 0x8           │ │
│ │                        Misc_VirtualSize: 0x628 0x384 0xC VirtualAddress: 0x216000 0x388 0x10 │ │
│ │                        SizeOfRawData: 0x1000 0x38C 0x14 PointerToRawData: 0x20E000 0x390     │ │
│ │                        0x18 PointerToRelocations: 0x0 0x394 0x1C PointerToLinenumbers: 0x0   │ │
│ │                        0x398 0x20 NumberOfRelocations: 0x0 0x39A 0x22 NumberOfLinenumbers:   │ │
│ │                        0x0 0x39C 0x24 Characteristics: 0x42000040>                           │ │
│ │           sizeofHdrs = 4096                                                                  │ │
│ │         virtual_data = b'MZ\x90\x00\x03\x00\x00\x00\x04\x00\x00\x00\xff\xff\x00\x00\xb8\x00… │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
TypeError: a bytes-like object is required, not 'int'
```

If you're more interested in making sure the GUIDs are correct, here's an example of me using the function from a custom plugin and the output of `symchk.exe` from Windows Debugging Tools to check `ntdll` from the same host. The GUIDs match.

Working code:
```
INFO     volatility3.framework.symbols.windows.pdbconv: Download PDB file...
DEBUG    volatility3.framework.symbols.windows.pdbconv: Attempting to retrieve http://msdl.microsoft.com/download/symbols/ntdll.pdb/F9A24309A184C71B7885047551E716D61/ntdll.pdb
```

Symchk.exe:
```
PS C:\Program Files\Microsoft Visual Studio\2022\Preview> symchk.exe /v C:\Windows\System32\ntdll.dll
[SYMCHK] Searching for symbols to C:\Windows\System32\ntdll.dll in path srv*c:\Symbols*http://msdl.microsoft.com/download/symbols
DBGHELP: Symbol Search Path: srv*c:\Symbols*http://msdl.microsoft.com/download/symbols
[SYMCHK] Using search path "srv*c:\Symbols*http://msdl.microsoft.com/download/symbols"
DBGHELP: No header for C:\Windows\System32\ntdll.dll.  Searching for image on disk
DBGHELP: C:\Windows\System32\ntdll.dll - OK
SYMSRV:  BYINDEX: 0x1
         c:\Symbols*http://msdl.microsoft.com/download/symbols
         ntdll.pdb
         F9A24309A184C71B7885047551E716D61
SYMSRV:  PATH: c:\Symbols\ntdll.pdb\F9A24309A184C71B7885047551E716D61\ntdll.pdb
SYMSRV:  RESULT: 0x00000000
DBGHELP: ntdll - public symbols
        c:\Symbols\ntdll.pdb\F9A24309A184C71B7885047551E716D61\ntdll.pdb
```

ikelos: Sorry I wasn't clear enough.  I'm trying to ensure the API is consistent, so I want to make sure that whatever *did* work with the version available in February is what we return now using the new method.  It sounds like it does, I'm just keen to ensure it's verified that the byte ordering and results end up the same.  If they do that's fine, if they don't we need to think about which ordering is best (whether we make the change and what breaks that might cause, or change the ordering so that the results match the previous version).  This may not be an issue at all, I just want to check before we approve it...  5:)

daddycocoaman: I went back to `pefile==2017.8.1` which is what requirements.txt currently has and it didn't work so I jumped to `2019.4.18` at random and it did. 

```
DEBUG    volatility3.framework.symbols.windows.pdbutil: Using symbol library: ntdll.pdb/F9A24309A184C71B7885047551E716D6-1
```

I dunno how often this function gets used or if at all, but it looks like the minimum requirement for pefile has been wrong since vol3 1.0.0. It just happened to work **at the time**.

ikelos (approved):
Ok, since the filenames on the server match up and this pulls down the correct files, the changes seem as though they're transparent from the outside, so I'll approve this.  Thanks for spotting it!

PR #1034, opened 2023-11-10
Linux: update maple tree extension to fix issue #1032

eve-mem:
Hello,

While debugging the maple tree parsing to see what has caused this issue https://github.com/volatilityfoundation/volatility3/issues/1032 for @4n6-fl I realized how I could recreate the issue and fix it. 

For some reason that I don't understand (and I would like to understand so, if this makes sense to you please would you team me) if the `get_slot_iter` function is called twice it will fail, the `seen` set still contains values when the `_parse_maple_tree_node` function starts. It's supposed to be an optional parameter with the default set to an empty set - so I'm not really sure what is going on. It's likely some python I don't understand. 

To recreate the issue it is as simple as opening volshell and trying to to parse the maple tree twice. 
```
(layer_name) >>> [ vma for  vma in ps()[-2].mm.mm_mt.get_slot_iter()]
[154723148639504, 154723148637832, 154723148638592, 154723148639352, 154723353319200, 154723148638744, 154723353319352, 154723353318592, 154723353320872, 154723353318896, 154723353317984, 154723353317680, 154723353317832, 154723353321176, 154723353318744, 154723353318440, 154723148638440, 154723353320720, 154723353319504, 154723148637224, 154723353317376, 154723353319656, 154723353319048, 154723148638896, 154723148636312, 154723148637984, 154723148636160, 154723353318288, 154723148639200, 154723148637376, 154723148639808, 154723148638136, 154723353317528, 154723148636464, 154723353321024, 154723148639048, 154723148636920, 154723148639960, 154723148636616, 154723353320112, 154723148636768, 154723148638288, 154723148637528, 154723148637680]
(layer_name) >>> [ vma for  vma in ps()[-2].mm.mm_mt.get_slot_iter()]
WARNING  volatility3.framework.symbols.linux.extensions: The mte 0x8cb84e6c971e has all ready been seen, no further results will be produced for this node.
WARNING  volatility3.framework.symbols.linux.extensions: The mte 0x8cb84e6c971e has all ready been seen, no further results will be produced for this node.
```

It is the double parsing that causes the bash plugin to have this issue. It does so here https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/bash.py#L82 and https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/bash.py#L92. 

This PR changes the call to `_parse_maple_tree_node` to actually specify the empty set - which seems to do the trick as seen here: 

```
[...]

PR #1035, opened 2023-11-10
Linux: update bash plugin to only get memory sections once

eve-mem:
Hello, 

After seeing that the Linux bash plugin calls the `get_process_memory_sections` function twice to get the same results in https://github.com/volatilityfoundation/volatility3/pull/1034  

I felt like I had to change it to be the smallest most insignificant amount more efficient. It likely makes no difference with older kernels that don't use maple tree structures, but in 6.1+ parsing the tree is at least a little bit more work so it makes sense to only do it once. 

This doesn't change how the plugin works or what any of the functions return, they all work exactly as they did before - perhaps meaning this PR wasn't worth wasting anyone time with....

### DIFF
--- a/volatility3/framework/plugins/linux/bash.py
+++ b/volatility3/framework/plugins/linux/bash.py
@@ -75,11 +75,16 @@ class Bash(plugins.PluginInterface, timeliner.TimeLinerInterface):
 
             bang_addrs = []
 
+            # get task memory sections to be used by scanners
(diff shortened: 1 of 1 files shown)


ikelos (approved):
Nope, that's exactly the kind of efficiency improvement we like!  It may seem small now, but when it's done a LOT, it adds up!  5:)

Issue #832, opened 2022-09-09
linux.check_afinfo.Check_afinfo Plugin Throws AttributeError


mthbrown:
**Describe the bug**
When running the **linux.check_afinfo.Check_afinfo** plugin on a Linux memory dump using a custom **Symbols Table**, it throws an **AttributeError**. Other plugins work without problems

**Context**
Volatility Version:  Volatility 3 Framework 2.3.1 (latest from git as of today)
Operating System:  Ubuntu 20.04.2 LTS
Python Version:  3.6.12
Suspected Operating System:  Ubuntu 22.04 (memory of this machine was captured)
Command:  

```
$ python vol.py -f ~/volatility/mem/Linux64.mem linux.check_afinfo.Check_afinfo
Volatility 3 Framework 2.3.1
Progress:  100.00               Stacking attempts finished                 
Symbol Name     Member  Handler Address
Traceback (most recent call last):
  File "vol.py", line 10, in <module>
    volatility3.cli.main()
  File "/home/ubuntu/vol3/volatility3/volatility3/cli/__init__.py", line 636, in main
    CommandLine().run()
  File "/home/ubuntu/vol3/volatility3/volatility3/cli/__init__.py", line 343, in run
    renderers[args.renderer]().render(constructed.run())
  File "/home/ubuntu/vol3/volatility3/volatility3/cli/text_renderer.py", line 177, in render
    grid.populate(visitor, outfd)
  File "/home/ubuntu/vol3/volatility3/volatility3/framework/renderers/__init__.py", line 212, in populate
    for (level, item) in self._generator:
  File "/home/ubuntu/vol3/volatility3/volatility3/framework/plugins/linux/check_afinfo.py", line 84, in _generator
    for name, member, address in self._check_afinfo(global_var_name, global_var, op_members, seq_members):
  File "/home/ubuntu/vol3/volatility3/volatility3/framework/plugins/linux/check_afinfo.py", line 51, in _check_afinfo
    for hooked_member, hook_address in self._check_members(var.seq_fops, var_name, op_members):
  File "/home/ubuntu/vol3/volatility3/volatility3/framework/objects/__init__.py", line 789, in __getattr__
    raise AttributeError(f"{agg_name} has no attribute: {self.vol.type_name}.{attr}")
AttributeError: StructType has no attribute: symbol_table_name1!tcp_seq_afinfo.seq_fops
```

**To Reproduce**

See above

**Expected behavior**

The plugin should run successfully

[...]

PR #1038, opened 2023-11-13
Linux: add member presence checks to linux.checks_afinfo 

eve-mem:
Hello!

Issue https://github.com/volatilityfoundation/volatility3/issues/832 is tracking the fact that newer kernels (from around 2018) won't have the required members for the hooking checks to be performed in this way. 

This PR adds some checks around the members of the objects in the `_check_afinfo` function so that a more informative warning is displayed rather than simply crashing. Does this seem like a good idea?

### DIFF
--- a/volatility3/framework/plugins/linux/check_afinfo.py
+++ b/volatility3/framework/plugins/linux/check_afinfo.py
@@ -51,10 +51,24 @@ class Check_afinfo(plugins.PluginInterface):
                 yield check, addr
 
     def _check_afinfo(self, var_name, var, op_members, seq_members):
-        for hooked_member, hook_address in self._check_members(
-            var.seq_fops, var_name, op_members
-        ):
-            yield var_name, hooked_member, hook_address
+        # check if object has a least one of the members used for analysis by this function
+        required_members = ["seq_fops", "seq_ops", "seq_show"]
+        for member in required_members:
+            vollog.debug(f"{var_name}: {member} :{var.has_member(member)}")
+        has_required_member = any(
+            [var.has_member(member) for member in required_members]
+        )
+        if not has_required_member:
+            vollog.warning(
+                f"This plugin requires the seq_fops, seq_ops, or seq_show members to be to check for hooks. These members are not present in the {var_name} object at {hex(var.vol.offset)}. This means you are either analyzing an unsupported kernel version or that your symbol table is corrupt."
+            )
+            return
+
+        if var.has_member("seq_fops"):
+            for hooked_member, hook_address in self._check_members(
+                var.seq_fops, var_name, op_members
+            ):
+                yield var_name, hooked_member, hook_address
 
         # newer kernels
         if var.has_member("seq_ops"):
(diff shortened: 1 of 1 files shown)


Abyss-W4tcher:
Hello, do you think redirecting users to https://github.com/AsafEitani/rootkit_plugins/blob/main/plugins/check_seqops.py, or directly merging both plugins would be a solution ?

They work toward the same goal in the end :) 

Regards

eve-mem:
@Abyss-W4tcher - I'd imagine that's certainly a possibility if @AsafEitani wanted to make a PR. My understanding is that @atcuno is looking into the way that this issue should be fixed fundamentally, and it might well be by updating it to do similar things to that seqops plugin. 

I'm not sure that a framework plugin would every redirect users to a community plugin, but I'd let the core devs make that call. Maybe you and I could help more on the community support side of things and point people to these other resources when it pops up (exactly like you did in the issue comments for #832)

🦊 _just a random internet vol user_

volatility3/framework/plugins/linux/check_afinfo.py
```diff
@@ -51,10 +51,24 @@ def _check_members(self, var_ops, var_name, members):
                 yield check, addr
 
     def _check_afinfo(self, var_name, var, op_members, seq_members):
-        for hooked_member, hook_address in self._check_members(
-            var.seq_fops, var_name, op_members
-        ):
-            yield var_name, hooked_member, hook_address
+        # check if object has a least one of the members used for analysis by this function
+        required_members = ["seq_fops", "seq_ops", "seq_show"]
+        for member in required_members:
```
ikelos: Hmmm, I know it's not a big deal, but this is traversing all the elements do output the debug information, and then does it again (effectively) as part of the any operator.  The debug information will probably only be checked if they're all false, so it might not be useful at all?

ikelos (commented):
Looks pretty good to me, just wondering how useful the debugging message is, given it traverse the loop twice every single item its called on, seemingly with not a great value?

volatility3/framework/plugins/linux/check_afinfo.py
eve-mem: Yes, I agree with you. Not the best. 

I wanted to have some difference between "The plugin run and there was no output, because there was nothing hooked" and "The plugin run and there was no output, because the kernel version isn't supported". 

The later might lead to a false sense of security. I'll work up a better way to show that difference.

eve-mem:
Hello @ikelos 

I've tweaked the logic for debug and warning messages. This removes the loops through the same parts twice, and also means that if every check with `_check_afinfo` fails you'll get a single warning message. Previously this change would have some a warning for every single symbol without the right members - which is quite spamy. It would be 6 warnings with the current plugin. 

I may be abusing the `PluginRequirementException` to do it however... what do you think? I originally used an attribute error, but thought catching those might hide other problems in the future somehow. Maybe there is a better way to do this?

I'm just keen to show the difference between "no results, all is fine" and "no results, because the plugin couldn't check".

volatility3/framework/plugins/linux/check_afinfo.py
eve-mem: I've tweaked this now, but might be abusing the `PluginRequirementException` - let me know what you think.

ikelos (approved):
Yep, looks good, thanks!  Much happier about the logging and checks..  5:)

ikelos:
As to the `PluginRequirementException`, I think it's the right one to use, even if there isn't an explicit requirement exception.  Kinda makes me wonder how we'd make an exception like that?  I guess we'd need to check that a kernel symbol containing the kernel version was within certain bounds, but that might not be accurate.  Defining a requirement on each of the required members would be cumbersome too.  I think this works best, I didn't even know we had that exception, but I think it's the right one.  If it causes any problems we can swap it out...  5:)

PR #1039, opened 2023-11-14
Linux extensions patches

Abyss-W4tcher:
- Fix a small error in the kobject class (see https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/overlays/linux/linux.py#L1204 for comparison)

module class :
- Fix leftovers from https://github.com/volatilityfoundation/volatility3/commit/83a8afba6baddf919a4e7fb032854bd354ca216c
- Fix broken implementation of module symbols construction (sources : https://refspecs.linuxbase.org/elf/gabi4+/ch4.symtab.html, https://github.com/TheCodeArtist/elf-parser/blob/master/elf-parser.c#L313) 

Reading symbols of an arbitrary module : 

```python
(layer_name) >>> from volatility3.framework.objects import utility
(layer_name) >>> vmlinux = self.context.modules[self.config["kernel"]]
(layer_name) >>> m = vmlinux.object("module",0xffffc06db040,absolute=True) # lime module
(layer_name) >>> utility.array_to_string(m.name)
'lime'
(layer_name) >>> list(m.get_symbols())
[('.LC0', 281473910153660, 281473910405065), ('setup_tcp', 281473910149120, 281473910405070), ('filp_open', 281473822829024, 281473910405080), ('crypto_alloc_ahash', 281473825133440, 281473910405090), ('strcpy', 281473826031120, 281473910405109), ('digest', 281473910158312, 281473910405313), ('deflate', 281473910153040, 281473910405339), ('path', 281473910158336, 281473910405116), ('localhostonly', 281473910158320, 281473910405121), ('ldigest_write_tcp', 281473910152432, 281473910405135)] # stripped for readability
```

### DIFF
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -105,14 +105,18 @@ class module(generic.GenericIntelProcess):
             yield attr
 
     def get_symbols(self):
-        """Get module symbols"""
+        """Get module symbols
+
+        Yields:
+                A tuple for each symbol containing the symbol name and its corresponding value
+        """
         if symbols.symbol_table_is_64bit(self._context, self.get_symbol_table_name()):
             prefix = "Elf64_"
         else:
             prefix = "Elf32_"
         elf_table_name = intermed.IntermediateSymbolTable.create(
             self._context,
-            self._context.modules["kernel"].config_path,
+            "module",
             "linux",
             "elf",
             native_types=None,
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -134,8 +134,8 @@ class module(generic.GenericIntelProcess):
                 try:
                     sym_offset = self.section_strtab + sym.st_name
                     sym_name = self._context.layers[self.vol.layer_name].read(
-                        sym_offset, sym.st_size
-                    )
+                        sym_offset, 512
+                    )  # 512 is the value of KSYM_NAME_LEN
                 except exceptions.PagedInvalidAddressException:
                     continue
 
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -139,12 +139,12 @@ class module(generic.GenericIntelProcess):
                 except exceptions.PagedInvalidAddressException:
                     continue
 
-                if sym_name:
+                # Stop at first null byte (strtab is a null terminated strings list)
+                sym_name = sym_name.split(b"\x00")[0].decode("latin-1")
+                if sym_name != "":
                     # Normalize sym_value
                     mask = self._context.layers[self.vol.layer_name].address_mask
                     sym_value = sym.st_value & mask
-                    # Stop at first null byte (strtab is a null terminated strings list)
-                    sym_name = sym_name.split(b"\x00")[0].decode("latin-1")
                     yield (sym_name, sym_value, sym_offset)
 
     def get_symbol(self, wanted_sym_name):
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -132,37 +132,37 @@ class module(generic.GenericIntelProcess):
             ),
             count=self.num_symtab + 1,
         )
-
         if self.section_strtab:
             for sym in syms:
+                sym_arr = self._context.object(
+                    self.get_symbol_table_name() + constants.BANG + "array",
+                    layer_name=self.vol.native_layer_name,
+                    offset=self.section_strtab + sym.st_name,
+                )
                 try:
-                    sym_offset = self.section_strtab + sym.st_name
-                    sym_name = self._context.layers[self.vol.layer_name].read(
-                        sym_offset, 512
-                    )  # 512 is the value of KSYM_NAME_LEN
-                except exceptions.PagedInvalidAddressException:
+                    sym_name = utility.array_to_string(
+                        sym_arr, 512
+                    )  # 512 is the value of KSYM_NAME_LEN kernel constant
+                except exceptions.InvalidAddressException:
                     continue
-
-                # Stop at first null byte (strtab is a null terminated strings list)
-                sym_name = sym_name.split(b"\x00")[0].decode("latin-1")
                 if sym_name != "":
-                    # Normalize sym_value
+                    # Normalize sym.st_value offset, which is an address pointing to the symbol value
                     mask = self._context.layers[self.vol.layer_name].address_mask
-                    sym_value = sym.st_value & mask
-                    yield (sym_name, sym_value, sym_offset)
+                    sym_address = sym.st_value & mask
+                    yield (sym_name, sym_address)
 
     def get_symbol(self, wanted_sym_name):
         """Get symbol value for a given symbol name"""
-        for sym_name, sym_value, sym_offset in self.get_symbols():
+        for sym_name, sym_address in self.get_symbols():
             if wanted_sym_name == sym_name:
-                return sym_value
+                return sym_address
 
         return None
 
-    def get_symbol_name_from_value(self, wanted_sym_value):
-        """Get symbol name for a given symbol value"""
-        for sym_name, sym_value, sym_offset in self.get_symbols():
-            if wanted_sym_value == sym_value:
+    def get_symbol_from_address(self, wanted_sym_address):
+        """Get symbol name for a given symbol address"""
+        for sym_name, sym_address in self.get_symbols():
+            if wanted_sym_address == sym_address:
                 return sym_name
 
         return None
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -1154,7 +1154,7 @@ class vfsmount(objects.StructType):
 class kobject(objects.StructType):
     def reference_count(self):
         refcnt = self.kref.refcount
-        if self.has_member("counter"):
+        if refcnt.has_member("counter"):
             ret = refcnt.counter
         else:
             ret = refcnt.refs.counter
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -141,7 +141,7 @@ class module(generic.GenericIntelProcess):
             for sym in syms:
                 yield sym
 
-    def get_symbols_names_and_addresses(self):
+    def get_symbols_names_and_addresses(self) -> Tuple[str, int]:
         """Get names and addresses for each symbol of the module
 
         Yields:
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -71,11 +71,11 @@ class module(generic.GenericIntelProcess):
     def _get_sect_count(self, grp):
         """Try to determine the number of valid sections"""
         arr = self._context.object(
-            self.get_symbol_table().name + constants.BANG + "array",
+            self.get_symbol_table_name() + constants.BANG + "array",
             layer_name=self.vol.layer_name,
             offset=grp.attrs,
             subtype=self._context.symbol_space.get_type(
-                self.get_symbol_table().name + constants.BANG + "pointer"
+                self.get_symbol_table_name() + constants.BANG + "pointer"
             ),
             count=25,
         )
@@ -92,11 +92,11 @@ class module(generic.GenericIntelProcess):
         else:
             num_sects = self._get_sect_count(self.sect_attrs.grp)
         arr = self._context.object(
-            self.get_symbol_table().name + constants.BANG + "array",
+            self.get_symbol_table_name() + constants.BANG + "array",
             layer_name=self.vol.layer_name,
             offset=self.sect_attrs.attrs.vol.offset,
             subtype=self._context.symbol_space.get_type(
-                self.get_symbol_table().name + constants.BANG + "module_sect_attr"
+                self.get_symbol_table_name() + constants.BANG + "module_sect_attr"
             ),
             count=num_sects,
         )
@@ -105,13 +105,14 @@ class module(generic.GenericIntelProcess):
             yield attr
 
     def get_symbols(self):
-        if symbols.symbol_table_is_64bit(self._context, self.get_symbol_table().name):
+        """Get module symbols"""
+        if symbols.symbol_table_is_64bit(self._context, self.get_symbol_table_name()):
             prefix = "Elf64_"
         else:
             prefix = "Elf32_"
         elf_table_name = intermed.IntermediateSymbolTable.create(
-            self.context,
-            self.config_path,
+            self._context,
+            self._context.modules["kernel"].config_path,
             "linux",
             "elf",
             native_types=None,
@@ -119,7 +120,7 @@ class module(generic.GenericIntelProcess):
         )
 
         syms = self._context.object(
-            self.get_symbol_table().name + constants.BANG + "array",
+            self.get_symbol_table_name() + constants.BANG + "array",
             layer_name=self.vol.layer_name,
             offset=self.section_symtab,
             subtype=self._context.symbol_space.get_type(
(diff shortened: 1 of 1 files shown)


volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -105,40 +105,62 @@ def get_sections(self):
...
+        """Get module symbols"""
+        if symbols.symbol_table_is_64bit(self._context, self.get_symbol_table_name()):
             prefix = "Elf64_"
         else:
             prefix = "Elf32_"
         elf_table_name = intermed.IntermediateSymbolTable.create(
-            self.context,
-            self.config_path,
+            self._context,
+            self._context.modules["kernel"].config_path,
```
ikelos: This assumes that the plugin requested a module called `kernel`.  It's unclear why the config path has been changed, that's just a location to store configuration information.  Also, variables that start with an `_` are private, and `self.context` is just a read-only property that provides `self._context` so it's not clear why this change is necessary?  Please don't include changes like this without explaining why they're being made and why they're necessary...

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -105,40 +105,62 @@ def get_sections(self):
...
+                except exceptions.PagedInvalidAddressException:
+                    continue
+
+                # Stop at first null byte (strtab is a null terminated strings list)
+                sym_name = sym_name.split(b"\x00")[0].decode("latin-1")
+                if sym_name != "":
+                    # Normalize sym_value
+                    mask = self._context.layers[self.vol.layer_name].address_mask
+                    sym_value = sym.st_value & mask
+                    yield (sym_name, sym_value, sym_offset)
```
ikelos: This is now yielding a totally different type of result, which is a hard API break and would need version number bumping.  It isn't clear wht this information is being returned separately rather than as a symbol object as it had been.

ikelos (commented):
So, happy with the `get_symbol_table_name()` changes, happy with the `refcnt` change, but the big changes in the middle will need more explanation and detail as to why they're being done that way, what the end goal is (since it's not clear that's the best way of doing it) and I making a change to the API will be a lot messier than otherwise, so if possible please avoid that like the plague.  Unfortunately it doens't look like this code has accurate type signatures yet, but if they were there mypy would be complaining loudly that they no longer matched.  Happy for you to either resubmit the first two small changes and get them merged quickly, or for you to hold them up whilst we talk over the bigger change, but unfortunatly this won't get merged as is...

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: Hello, I am trying to get more familiar with the code base, and even if it effectively works locally for me, it may not be the best in terms of context. I'll be happy to work this out to make it cleaner.


So here basically, the two objects I replaced wheren't accessible anymore from the "self" object, which is why I had to resort to this. One solution would be to make them available to the "module" object, from the class instantiating it.

edit : here is an example, with `m` a module : 

```python
(layer_name) >>> m.context
  File "[...]/volatility3/volatility3/framework/objects/__init__.py", line 968, in __getattr__
    raise AttributeError(
AttributeError: StructType has no attribute: symbol_table_name1!module.context
(layer_name) >>> m.config_path
  File "[...]/volatility3/volatility3/framework/objects/__init__.py", line 968, in __getattr__
    raise AttributeError(
AttributeError: StructType has no attribute: symbol_table_name1!module.config_path
```

Abyss-W4tcher:
Hello again,

EDIT : I moved this part of the comment to the corresponding conversation.

Thanks for your time and review :) !

Regards,
Abyss Watcher.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -105,40 +105,62 @@ def get_sections(self):
...
+                try:
+                    sym_offset = self.section_strtab + sym.st_name
+                    sym_name = self._context.layers[self.vol.layer_name].read(
+                        sym_offset, 512
+                    )  # 512 is the value of KSYM_NAME_LEN
+                except exceptions.PagedInvalidAddressException:
+                    continue
+
+                # Stop at first null byte (strtab is a null terminated strings list)
+                sym_name = sym_name.split(b"\x00")[0].decode("latin-1")
```
eve-mem: Hey @Abyss-W4tcher - Just a minor one, more a function that that might be helpful for you to know than anything else. For the this name part, you might be able to use the `pointer_to_string` utility rather than doing the split for the null byte yourself. It also handles errors nicely for you as well if there are weird characters. 

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/objects/utility.py#L23

🦊 _just a random internet vol user_

Abyss-W4tcher: Hi, thanks for pointing this out to me ! Excluding the use of "self._context" which isn't great, as discussed earlier, could you tell me if it is the right way to do it : 

```python
sym_offset : int = self.section_strtab + sym.st_name
try:
  pointer_to_sym_offset = self._context.object("pointer", sym_offset) # is the use of "cast('pointer')" possible for an int ?
except PagedInvalidAddressException: 
 continue 

sym_name = utility.pointer_to_string(pointer_to_sym_offset, 512)
if sym_name != "": 
 [...]
```

Thanks

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -105,40 +105,62 @@ def get_sections(self):
             yield attr
 
     def get_symbols(self):
-        if symbols.symbol_table_is_64bit(self._context, self.get_symbol_table().name):
+        """Get module symbols"""
```
Abyss-W4tcher: - Add an indication on returned values

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Yeah, ok.  So we don't  expose `_context`, I think to put systems off using them.  Since it's `self`, `self._context` is acceptable (the underscore is indicating it's a private variable, but accessing a private variable via self is ok).  I went back and looked at the git blame to see how that got there.  It looks like the `_` was dropped due to a mistake when attempting to clean some other bits up.

Sadly, accessing the kernel layer is definitely a no-no for the reasons provided.  The config_path was originally set as `""` which is also not good, because if it's hard coded it could conflict.  Most other functions ask for a context and config_path when they need to construct symbol tables  (for example in the kdbg handler):

```
    def get_kdbg_structure(
        cls,
        config_path: str,
        layer_name: str,
        symbol_table: str,
    ) -> interfaces.objects.ObjectInterface:

...

        kdbg_table_name = intermed.IntermediateSymbolTable.create(
            context,
            interfaces.configuration.path_join(config_path, "kdbg"),
            "windows",
            "kdbg",
            native_types=native_types,
            class_types=extensions.kdbg.class_types,
        )
```

Please try to do something along those lines.  It's strictly not important at the moment, but best to code as though it will become important in the future (hence the change from its original `""` to what it is now).

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Sorry, I gave the wrong advice earlier, `self._context` is the right way to do it, given there is no other means of getting the object's context, but it is private, so reaching into another object's `_context` would be bad.  Hope that makes sense...

ikelos: Yes, if you have an integer, you should be able to cast it to any other volatility type.  Pointers can be instantiated even on invalid addresses though, so best to check that integer against the layer using, I think, `is_valid` first...

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: Regarding the fact that the previous implementation, using an undocumented set_cached_strtab function, wasn't working with pretty obvious errors, and with no GitHub issues raised, I started noticing that this code didn't seem to be used anywhere in the codebase.

Doing so, to follow the style of previous Volatility2 implementation of this [section](https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/overlays/linux/linux.py#L994), I decided to correct the logic needed to read symbols :

- iterate over the "ElfXX_Sym" structures stored in "strtab"
- extract each "st_name", which is an offset inside the "strtext" string list
- read [512 bytes](https://elixir.bootlin.com/linux/v6.6/source/include/linux/kallsyms.h#L18) starting from "strtext" offset + "st_name"
- check if read was successful, and if symbol isn't empty

The return value isn't a "sym" object, because it does not provide any useful information for the caller, instead I adapted :
```python
# https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/overlays/linux/linux.py#L1015
if sym_name != "":
                ret_syms.append((str(sym_name), sym_struct.st_value.v()))
```
The st_value is normalized, following : https://refspecs.linuxbase.org/elf/gabi4+/ch4.symtab.html#:~:text=In%20executable%20and%20shared%20object%20files%2C%20st_value.

I also added the absolute symbol offset to the tuple returned, but one can say that is not really interesting, and we could stick with (symbol_name, symbol_value).

Finally, I think this would eventually result in an API break, because right now I do not see how it could be usable (and where it is used as is) ?

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: If config_path is indeed an arbitrary string ("kdbg" seems a convenient choice given the function you provided), I assume naming it `module` in the current context should be right ? 

However, if by "conflict" you mean that the config_path have to be unique for each instantiated module, we could try to concatenate the module address to the "module" string to be sure it is unique.

Abyss-W4tcher:
The two previous commits include more comments, use of utilities functions, changing the config path and a general implementation following more closely the Volatilty2 one.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: `module` would work, but rather than reinstantiating the table for each time this method is called (and potentially making lots and lots of tables), it would be better to give it a known name (long and descriptive).  Ideally you could make the plugin load up the symbols it wants and then just pass them in as a table name, and thus remove the need for this bit of code to decide?

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Sadly, that doesn't matter too much.  If we exposed it, then it's part of the API.  So the options then become:

* Add something extra under a different name that does the right thing (bump the minor number)
* Change this function so it returns the right thing in the same format as the previous one did (bump the patch number)
* Remove the old one entirely and make a new one (or just change this one) with a different API, whilst bumping the major version number of whatever component this is associated with. (bump the major number)

We shouldn't be afraid of bumping version numbers, it just needs to make us think carefully about our API when we design it to avoid doing it too often and confusing everybody.  I have no problem with the options or with bumping the numbers appropriately, and if it isn't used much it should be fine, but we're not the only users of the library, so if we are going to make a change we need to stick to our own rules about making changes...  5:)

ikelos (commented):
Here's some replies.  I don't mind changing how things work, but if we do we need to follow our own rules about such changes.  Since the API is public, someone else may be using it even if we're not, so removing it (or changing how it returns) would require a major version bump.  Adding a new one just requires a minor version bump (and we should add a deprecation notice to the old function).

See [here](https://github.com/volatilityfoundation/volatility3/blob/d56297c4b1cf0f9f4912f4f4158e232c700acb3f/volatility3/framework/plugins/windows/netstat.py#L340) for an example of how we do deprecation.

Abyss-W4tcher:
Hello, I took notice of the different scenarios regarding deprecation. I think the best option is to keep the existing functions inputs and outputs types, and add, in a second time, functions that does what I initially (re)implemented.

To do so, I made a dedicated function to get the `elf_table_name`, which is called by `get_symbols` only once. There aren't multiple "elf" tables created with this logic (thanks for pointing it out).

I kept the `sym` object as return value of `get_symbols()`, and added a function `get_symbols_names_and_addresses()` to do the names and addresses extraction. 

All already existing functions still have the same inputs and outputs types, but have their core logic changed, which I hope should only result in a "patch" number bump :)

Abyss-W4tcher:
Hi👋,

I would like to know if the team had any news or plans regarding this PR. Indeed, I have some plugins, planned to be released as "community plugins", that depend on these edits. 
Of course, there aren't any hurry, but if there is a schedule for this PR review, please let me know :) !

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -92,53 +92,94 @@ def get_sections(self):
...
-        if symbols.symbol_table_is_64bit(self._context, self.get_symbol_table().name):
-            prefix = "Elf64_"
-        else:
-            prefix = "Elf32_"
+    def get_elf_table_name(self):
         elf_table_name = intermed.IntermediateSymbolTable.create(
-            self.context,
-            self.config_path,
+            self._context,
+            "config_name_elf_symbol_table",
```
ikelos, line 110: I'd ditch the `config_name` at the start of this.  It's in the configuration, so the `config_name` is kind of implicit/seems a little odd...

ikelos (commented):
Hiya, generally no complaints.  I'd probably update the config name, but that can be changed without any effect later, so happy to merge this if you don't want to update it now.  Just let me know either way and we'll get it merged...

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -92,53 +92,94 @@ def get_sections(self):
...
             count=self.num_symtab + 1,
         )
         if self.section_strtab:
             for sym in syms:
-                sym.set_cached_strtab(self.section_strtab)
                 yield sym
 
-    def get_symbol(self, wanted_sym_name):
-        """Get value for a given symbol name"""
+    def get_symbols_names_and_addresses(self):
```
ikelos: Could you please type annotate this too (so `-> Tuple[str, int]`)?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -92,53 +92,94 @@ def get_sections(self):
...
-                self.get_symbol_table().name + constants.BANG + "module_sect_attr"
+                self.get_symbol_table_name() + constants.BANG + "module_sect_attr"
             ),
             count=num_sects,
         )
 
         for attr in arr:
             yield attr
 
-    def get_symbols(self):
```
ikelos, line 107: Removing existing functions will require a major version bump as well.  Since you've defined new functions, please just include a deprecation warning and keep these methods around so the API doesn't change.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -92,53 +92,94 @@ def get_sections(self):
...
+                self._elf_table_name + constants.BANG + prefix + "Sym"
             ),
             count=self.num_symtab + 1,
         )
         if self.section_strtab:
             for sym in syms:
-                sym.set_cached_strtab(self.section_strtab)
                 yield sym
 
-    def get_symbol(self, wanted_sym_name):
```
ikelos, line 135: Removing this would also be an API break and require the major version bumping.

ikelos (commented):
Just spotted a few breaking changes.  I thought you were just adding functions, but you took the old ones away, rather than deprecating them.

You can add the warning in like [this](https://github.com/volatilityfoundation/volatility3/blob/4ce42fcac4aa29cbfdc9676b2420a95ed9cff122/volatility3/framework/contexts/__init__.py#L424) and you can even use the new method to produce the results, but the function needs to return the same outputs when given the same inputs as the original.

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: Hi @ikelos, this function still exists, it just moved a few lines under

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: Hi, this function still exists, it just moved a few lines under (too)

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: As I'm not sure to fully understand what you want me to do with the config_name, I'd prefer if you could change it later.

Abyss-W4tcher:
I added type hinting, as requested. Regarding your other comments, I answered directly in the corresponding conversation 👍.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Just change it from `"config_name_elf_symbol_table",` to `"elf_symbol_table",`

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Ah sorry, the diff was confusing...

ikelos (approved):
Cool, I'll make the `config_name` change in a bit then...  5:)

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: Oh, I thought you were talking about the field type, not the string. Ok this makes sense indeed !

Issue #876, opened 2022-12-06
Issue with windows.strings plugin


crocodile-on-the-nile:
**Describe the bug**
windows.strings plugin does not display a message when a specific string is identified in the memory of a process

**Context**
Volatility Version:  Volatility 3 Framework 2.4.1
Operating System:  Windows 10
Python Version:  3.9.9
Suspected Operating System: Windows 10 Build 1809  
Command:  python vol.py -f ..\DESKTOP-N81KBM0-20221121-235651.dmp windows.strings --strings-file=..\extracted_strings.txt

**Details of experiment**

**Step 1:** **Acquiring the memory dump**

On a Windows 10 computer, a text file was opened in _notepad.exe_. The text file contained the target string _mysupersecretfancypassword_. The process ID of _notepad.exe_ is 6896. Another text file opened in  _wordpad.exe_ contained the same target string. For now, the focus is only on notepad.

Memory was acquired from this computer as _DESKTOP-N81KBM0-20221121-235651.dmp_ using _DumpIt.exe_. 

**Step 2:** **Extracting target string with offset from the memory dump**

Strings.exe was used to extract all the strings from this memory dump into a text file. The command used was:

> strings64.exe -o DESKTOP-N81KBM0-20221121-235651.dmp > strings_from_dump.txt

Four lines from _strings_from_dump.txt_ that contained the target string and its offset were copied into _extracted_strings.txt_

The contents of _extracted_strings.txt_ is as follows:
>type ..\extracted_strings.txt
2065948033:mysupersecretfancypassword\par
2201153780:mysupersecretfancypassword
2443052438:mysupersecretfancypassword
2444601355:mysupersecretfancypassword

**Step 3: Attempting to find context for strings in _extracted_strings.txt_**

The goal was to use _windows.strings_ plugin to prove that the strings in _extracted_strings.txt_ can be found within the memory of _notepad.exe_ with PID 6896.

Since the following command took a long time to complete,
> python vol.py -f ..\DESKTOP-N81KBM0-20221121-235651.dmp windows.strings --strings-file=..\extracted_strings.txt

The PID of notepad was included in the previous command, as:
> python vol.py -f ..\DESKTOP-N81KBM0-20221121-235651.dmp windows.strings --pid 6896 --strings-file=..\extracted_strings.txt

**Expected behavior**

The expected output was a message like 'String found'. However, there was only a message indicating that the string search is progressing.

**Example output**

**Output 1:**
[...]

PR #1043, opened 2023-11-19
Fix for windows.strings revmap offsets

kevthehermit:
This PR fixes a bug with `windows.strings` ref: https://github.com/volatilityfoundation/volatility3/issues/876

Output has been compared from Vol2.6 to ensure the correct offsets are being returned now. 

This PR also changes the output to better fit the new renderer separating results into parts instead of a string `.join` that was used before. 

As a significant change to the output, the version has also been bumped to `2.0.0` 

Example of these changes working and compared to vol2.6 can be seen in the issue thread

### DIFF
--- a/volatility3/framework/plugins/windows/strings.py
+++ b/volatility3/framework/plugins/windows/strings.py
@@ -92,7 +92,7 @@ class Strings(interfaces.plugins.PluginInterface):
 
             # We should really take care of this in the revmap generator
             mapping_entry = revmap.get(
-                phys_offset & 0xFFFFFFFFFFFFF000, [("In Unallocated Space", 0)]
+                phys_offset >> 12, [("In Unallocated Space", 0)]
             )
             for item in mapping_entry:
                 region, offset = item
@@ -178,9 +178,9 @@ class Strings(interfaces.plugins.PluginInterface):
             for mapval in layer.mapping(0x0, layer.maximum_address, ignore_errors=True):
                 offset, _, mapped_offset, mapped_size, maplayer = mapval
                 for val in range(mapped_offset, mapped_offset + mapped_size, 0x1000):
-                    cur_set = reverse_map.get(val, set())
+                    cur_set = reverse_map.get(val >> 12, set())
                     cur_set.add(("kernel", val))
-                    reverse_map[offset] = cur_set
+                    reverse_map[offset >> 12] = cur_set
                 if progress_callback:
                     progress_callback(
                         (offset * 100) / layer.maximum_address,
@@ -214,14 +214,14 @@ class Strings(interfaces.plugins.PluginInterface):
                             for val in range(
                                 mapped_offset, mapped_offset + mapped_size, 0x1000
                             ):
-                                cur_set = reverse_map.get(mapped_offset, set())
+                                cur_set = reverse_map.get(mapped_offset >> 12, set())
                                 cur_set.add(
                                     (
                                         f"Process {process.UniqueProcessId}",
                                         mapped_offset,
                                     )
                                 )
-                                reverse_map[offset] = cur_set
+                                reverse_map[offset >> 12] = cur_set
                             # FIXME: make the progress for all processes, rather than per-process
                             if progress_callback:
                                 progress_callback(
--- a/volatility3/framework/plugins/windows/strings.py
+++ b/volatility3/framework/plugins/windows/strings.py
@@ -90,18 +90,26 @@ class Strings(interfaces.plugins.PluginInterface):
         for phys_offset, string in string_list:
             line_count += 1
 
+            # calculate the offset for this string within a 4096 page so
+            # that this offset can be added to mappings which are all
+            # page aligned. This ensures that a string located at phy
+            # add 0x1e64cd20 would carry the 0xd20 to the virtual offsets
+            # displayed in the plugin output. Without this it would show
+            # only the page that the string was found, rather than the
+            # actually addr. 0xFFF is 4095 e.g. all lower bits set.
+            offset_within_page = phys_offset & 0xFFF
+
+
             mapping_entry = revmap.get(
                 phys_offset >> 12,
                 [{"region": "Unallocated", "pid": -1, "offset": 0x00}],
             )
+
             for item in mapping_entry:
                 # Get the full virtual address not just the page start
                 # If the string is in unalloacted memory, we set the offset to 0x00
                 offset = item.get("offset", 0x00)
-                if offset == 0x00:
-                    virtual_address = 0x00
-                else:
-                    virtual_address = (offset & ~0xFFF) | (phys_offset & 0xFFF)
+                virtual_address = offset + offset_within_page
 
                 yield (
                     0,
@@ -114,6 +122,7 @@ class Strings(interfaces.plugins.PluginInterface):
                     ),
                 )
 
+            
             prog = line_count / num_strings * 100
             if round(prog, 1) > last_prog:
                 last_prog = round(prog, 1)
@@ -160,22 +169,46 @@ class Strings(interfaces.plugins.PluginInterface):
         filter = pslist.PsList.create_pid_filter(pid_list)
 
         layer = context.layers[layer_name]
-        # reverse_map: Dict[int, Set[Tuple[str, int]]] = dict()
-        reverse_map: Dict[int, List[Dict[str, Union[str, int]]]] = dict()
+        reverse_map: Dict[int, Set[Tuple[str, int]]] = dict()
         if isinstance(layer, intel.Intel):
             # We don't care about errors, we just wanted chunks that map correctly
             for mapval in layer.mapping(0x0, layer.maximum_address, ignore_errors=True):
-                offset, _, mapped_offset, mapped_size, maplayer = mapval
-                for val in range(mapped_offset, mapped_offset + mapped_size, 0x1000):
-                    cur_set = reverse_map.get(val >> 12, list())
-                    cur_set.append({"region": "kernel", "pid": -1, "offset": val})
+                (
+                    virt_offset,
+                    _virt_size,
+                    phy_offset,
+                    phy_mapping_size,
+                    _phy_layer_name,
+                ) = mapval
 
-                    reverse_map[offset >> 12] = cur_set
+                # for each page within the mapping we need to store the phy_offset and
+                # the matching virt_offset
+                for offset_to_page_within_mapping in range(0, phy_mapping_size, 0x1000):
+                    # calculate the page number for this phy_offset, e.g. the ">> 12"
+                    # drops the bits that would address an offset within the page.
+                    # This means that all offsets within the same page get the same
+                    # physical_page number.
+                    physical_page = (
+                        phy_mapping_size + offset_to_page_within_mapping
+                    ) >> 12
+
+                    # get the existing mappings for this physical page from the
+                    # reverse map set.
+                    cur_set = reverse_map.get(physical_page, list())
+
+                    # add a mapping for this virtual offset, taking care to add the
+                    # offset_to_page_within_mapping to ensure that all pages match correctly.
+                    # Without this the 2nd, 3rd etc pages would all incorrectly map to the same
+                    # virtual offset.
+                    cur_set.append({"region": "Kernel", "pid": -1, "offset": virt_offset + offset_to_page_within_mapping})
+
+                    # store these results back in the reverse_map
+                    reverse_map[physical_page] = cur_set
                 if progress_callback:
                     progress_callback(
-                        (offset * 100) / layer.maximum_address,
-                        "Creating reverse kernel map",
-                    )
+                        (virt_offset * 100) / layer.maximum_address,
+                         "Creating reverse kernel map",
+                     )
 
             # TODO: Include kernel modules
 
@@ -197,27 +230,36 @@ class Strings(interfaces.plugins.PluginInterface):
 
                     proc_layer = context.layers[proc_layer_name]
                     if isinstance(proc_layer, linear.LinearlyMappedLayer):
+                        # this follows the same pattern as the kernel mappings above.
                         for mapval in proc_layer.mapping(
                             0x0, proc_layer.maximum_address, ignore_errors=True
                         ):
-                            mapped_offset, _, offset, mapped_size, maplayer = mapval
-                            for val in range(
-                                mapped_offset, mapped_offset + mapped_size, 0x1000
+                            (
+                                virt_offset,
+                                _virt_size,
+                                phy_offset,
+                                phy_mapping_size,
+                                _phy_layer_name,
+                            ) = mapval
+                            for offset_to_page_within_mapping in range(
+                                0, phy_mapping_size, 0x1000
                             ):
-                                cur_set = reverse_map.get(mapped_offset >> 12, list())
+                                physical_page = (
+                                    phy_offset + offset_to_page_within_mapping
+                                ) >> 12
+                                cur_set = reverse_map.get(physical_page, list())
                                 cur_set.append(
                                     {
-                                        "region": "process",
+                                        "region": "Process",
                                         "pid": process.UniqueProcessId,
-                                        "offset": mapped_offset,
+                                        "offset": virt_offset + offset_to_page_within_mapping,
                                     }
                                 )
-
-                                reverse_map[offset >> 12] = cur_set
+                                reverse_map[physical_page] = cur_set
                             # FIXME: make the progress for all processes, rather than per-process
                             if progress_callback:
                                 progress_callback(
-                                    (offset * 100) / layer.maximum_address,
+                                    (virt_offset * 100) / proc_layer.maximum_address,
                                     f"Creating mapping for task {process.UniqueProcessId}",
                                 )
 
--- a/volatility3/framework/plugins/windows/strings.py
+++ b/volatility3/framework/plugins/windows/strings.py
@@ -18,7 +18,7 @@ vollog = logging.getLogger(__name__)
 class Strings(interfaces.plugins.PluginInterface):
     """Reads output from the strings command and indicates which process(es) each string belongs to."""
 
-    _version = (1, 2, 0)
+    _version = (2, 0, 0)
     _required_framework_version = (2, 0, 0)
     strings_pattern = re.compile(rb"^(?:\W*)([0-9]+)(?:\W*)(\w[\w\W]+)\n?")
 
@@ -47,7 +47,13 @@ class Strings(interfaces.plugins.PluginInterface):
 
     def run(self):
         return renderers.TreeGrid(
-            [("String", str), ("Physical Address", format_hints.Hex), ("Result", str)],
+            [
+                ("String", str),
+                ("Region", str),
+                ("PID", int),
+                ("Physical Address", format_hints.Hex),
+                ("Virtual Address", format_hints.Hex),
+            ],
             self._generator(),
         )
 
@@ -81,22 +87,45 @@ class Strings(interfaces.plugins.PluginInterface):
         last_prog: float = 0
         line_count: float = 0
         num_strings = len(string_list)
-        for offset, string in string_list:
+        for phys_offset, string in string_list:
             line_count += 1
-            try:
-                revmap_list = [
-                    name + ":" + hex(offset) for (name, offset) in revmap[offset >> 12]
-                ]
-            except (IndexError, KeyError):
-                revmap_list = ["FREE MEMORY"]
-            yield (
-                0,
-                (
-                    str(string, "latin-1"),
-                    format_hints.Hex(offset),
-                    ", ".join(revmap_list),
-                ),
+
+            # We should really take care of this in the revmap generator
+            mapping_entry = revmap.get(
+                phys_offset & 0xFFFFFFFFFFFFF000, [("In Unallocated Space", 0)]
             )
+            for item in mapping_entry:
+                region, offset = item
+                if "Process" in region:
+                    location = "Process"
+                    pid = region.split(" ")[1]
+                elif "kernel" in region:
+                    location = "Kernel"
+                    pid = -1
+                elif "In Unallocated Space" in region:
+                    location = "Unallocated Space"
+                    pid = -1
+                else:
+                    location = "Unknown"
+                    pid = -1
+
+                # Get the full virtual address not just the page start
+                if offset == 0:
+                    virtual_address = 0
+                else:
+                    virtual_address = (offset & ~0xFFF) | (phys_offset & 0xFFF)
+
+                yield (
+                    0,
+                    (
+                        str(string, "latin-1"),
+                        location,
+                        int(pid),
+                        format_hints.Hex(phys_offset),
+                        format_hints.Hex(virtual_address),
+                    ),
+                )
+
             prog = line_count / num_strings * 100
             if round(prog, 1) > last_prog:
                 last_prog = round(prog, 1)
@@ -149,9 +178,9 @@ class Strings(interfaces.plugins.PluginInterface):
             for mapval in layer.mapping(0x0, layer.maximum_address, ignore_errors=True):
                 offset, _, mapped_offset, mapped_size, maplayer = mapval
                 for val in range(mapped_offset, mapped_offset + mapped_size, 0x1000):
-                    cur_set = reverse_map.get(val >> 12, set())
-                    cur_set.add(("kernel", offset))
-                    reverse_map[val >> 12] = cur_set
+                    cur_set = reverse_map.get(val, set())
+                    cur_set.add(("kernel", val))
+                    reverse_map[offset] = cur_set
                 if progress_callback:
                     progress_callback(
                         (offset * 100) / layer.maximum_address,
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/strings.py
```diff
@@ -81,22 +87,45 @@ def _generator(self) -> Generator[Tuple, None, None]:
...
-                revmap_list = ["FREE MEMORY"]
-            yield (
-                0,
-                (
-                    str(string, "latin-1"),
-                    format_hints.Hex(offset),
-                    ", ".join(revmap_list),
-                ),
+
+            # We should really take care of this in the revmap generator
```
ikelos: This kind of comment makes me feel like it's a rushed patch...  5;P  If it should be done there, is there a good reason it's being done here instead?

volatility3/framework/plugins/windows/strings.py
```diff
@@ -81,22 +87,45 @@ def _generator(self) -> Generator[Tuple, None, None]:
...
+            mapping_entry = revmap.get(
+                phys_offset & 0xFFFFFFFFFFFFF000, [("In Unallocated Space", 0)]
```
ikelos: I'm still keen to know the difference between this and shifting it by 12 (which should be the page bit size not hardcoded to 12, so the previous version wasn't great either)?  Is this a case where we should be canconcializing?  Are the bits above 0xFFFFFFFFFFFFF000 important?

volatility3/framework/plugins/windows/strings.py
```diff
@@ -81,22 +87,45 @@ def _generator(self) -> Generator[Tuple, None, None]:
...
             )
+            for item in mapping_entry:
+                region, offset = item
+                if "Process" in region:
```
ikelos: String comparisons for differentiation seem horribly prone to typo issues in the future. Could this be turned into an enumeration or just something that contains the information other than in text form?

volatility3/framework/plugins/windows/strings.py
```diff
@@ -81,22 +87,45 @@ def _generator(self) -> Generator[Tuple, None, None]:
...
+                    location = "Process"
+                    pid = region.split(" ")[1]
```
ikelos: Yep, this is definitely icky.  Really, really needs to be a better way of passing this information.

volatility3/framework/plugins/windows/strings.py
```diff
@@ -81,22 +87,45 @@ def _generator(self) -> Generator[Tuple, None, None]:
...
+                    pid = -1
+                else:
+                    location = "Unknown"
+                    pid = -1
+
+                # Get the full virtual address not just the page start
+                if offset == 0:
+                    virtual_address = 0
+                else:
+                    virtual_address = (offset & ~0xFFF) | (phys_offset & 0xFFF)
```
ikelos: Again, these probably ought to be using the layer.mask value or something like that?

volatility3/framework/plugins/windows/strings.py
```diff
@@ -149,9 +178,9 @@ def generate_mapping(
             for mapval in layer.mapping(0x0, layer.maximum_address, ignore_errors=True):
                 offset, _, mapped_offset, mapped_size, maplayer = mapval
                 for val in range(mapped_offset, mapped_offset + mapped_size, 0x1000):
-                    cur_set = reverse_map.get(val >> 12, set())
-                    cur_set.add(("kernel", offset))
-                    reverse_map[val >> 12] = cur_set
+                    cur_set = reverse_map.get(val, set())
+                    cur_set.add(("kernel", val))
+                    reverse_map[offset] = cur_set
```
ikelos: This appears to have changed the index from what had been `val` to `offset`.  I assume that intentional but it looks like it might potentially leave two entries in the reverse_map.  Since this code appears to just be adding to the members of the set, taking the set from one index and putting it back in a (potentially) different index, seems like a bad idea...

volatility3/framework/plugins/windows/strings.py
```diff
@@ -185,11 +214,14 @@ def generate_mapping(
...
+                                cur_set = reverse_map.get(mapped_offset, set())
                                 cur_set.add(
-                                    (f"Process {process.UniqueProcessId}", offset)
+                                    (
+                                        f"Process {process.UniqueProcessId}",
+                                        mapped_offset,
+                                    )
                                 )
-                                reverse_map[mapped_offset >> 12] = cur_set
+                                reverse_map[offset] = cur_set
```
ikelos: Again, the difference between `mapped_offset` and `offset` is concerning here...  5:S

ikelos (changes requested):
Thanks very much for this, but there's a couple of points about it.  Mostly, *why* does dropping the shifting work?  There's a couple bits where we should use the layer mask functions rather than the (quick and easy, but potentially wrong in the future) fixed page sizes, and lastly there's a couple of list updates that potentially double entries up which definitely need a look.  Those, sadly, are showstoppers, so this can't get merged until they're resolved.

volatility3/framework/plugins/windows/strings.py
kevthehermit: Sorry but out of my depth in explaining why I just compared the two versions one works, the other doesn't

volatility3/framework/plugins/windows/strings.py
kevthehermit: Yes they were listing the wrong way round, you can see this in the explanation i left in the issue thread comparing the 2.6 and 3 plugins these two values were being put in the wrong way round, I just switched them

volatility3/framework/plugins/windows/strings.py
kevthehermit: This used to be a `", ".join(revmap_list),` i would have to rewrite the entire revmap parser and i have to be honest i dont know enough about what thats doing so this felt like a safer option

volatility3/framework/plugins/windows/strings.py
kevthehermit: This used to be a `", ".join(revmap_list),` i would have to rewrite the entire revmap parser and i have to be honest i dont know enough about what thats doing so this felt like a safer option

volatility3/framework/plugins/windows/strings.py
kevthehermit: same as above, its what comes from revmap generator, so would have to rewrite the revmapper

ikelos:
Thanks for giving this another pass.  I'm still really uneasy about the indexing of the revmap, and it feels like we're not quite certain how it's working or how it's supposed to work.  Hopefully one of these evenings or on the weekend, I'll get a chance to dig into it and figure out what's going wrong, but this feels too much of a "it just works now" solution and I much prefer understanding how and why and making sure it's doing the right thing (otherwise we'll just have to come back and fix it in the future when that one corner case turns up and takes us hours to figure out)...  5:)

volatility3/framework/plugins/windows/strings.py
```diff
@@ -147,14 +171,46 @@ def generate_mapping(
...
+                    cur_set = reverse_map.get(physical_page, list())
+
+                    # add a mapping for this virtual offset, taking care to add the
+                    # offset_to_page_within_mapping to ensure that all pages match correctly.
+                    # Without this the 2nd, 3rd etc pages would all incorrectly map to the same
+                    # virtual offset.
+                    cur_set.append(
+                        {
+                            "region": "Kernel",
+                            "pid": -1,
```
ikelos, line 204: This should be a `BaseAbsentValue` derived class.  Maybe `NotApplicable`?

volatility3/framework/plugins/windows/strings.py
```diff
@@ -81,22 +87,40 @@ def _generator(self) -> Generator[Tuple, None, None]:
...
+                # If the string is in unalloacted memory, we set the offset to 0x00
+                offset = item.get("offset", 0x00)
+                virtual_address = offset + offset_within_page
+
+                yield (
+                    0,
+                    (
+                        str(string.strip(), "latin-1"),
+                        item.get("region", "Unallocated"),
+                        item.get("pid", -1),
```
ikelos, line 118: This also should default to a `BaseAbsentValue` derived class.

volatility3/framework/plugins/windows/strings.py
```diff
@@ -178,22 +234,37 @@ def generate_mapping(
...
-                                mapped_offset, mapped_offset + mapped_size, 0x1000
+                            (
+                                virt_offset,
+                                _virt_size,
+                                phy_offset,
+                                phy_mapping_size,
+                                _phy_layer_name,
+                            ) = mapval
+                            for offset_to_page_within_mapping in range(
+                                0, phy_mapping_size, 0x1000
```
ikelos, line 249: This is hardcoding the page size (which then agrees with the value 12 also hardcoded everywhere).  That's ok, but I think it would be better to pull them out and make them constants in the plugin (just so they stay the same if the code ever changes).

volatility3/framework/plugins/windows/strings.py
```diff
@@ -147,14 +171,46 @@ def generate_mapping(
...
+                    virt_offset,
+                    _virt_size,
+                    phy_offset,
+                    phy_mapping_size,
+                    _phy_layer_name,
+                ) = mapval
+
+                # for each page within the mapping we need to store the phy_offset and
+                # the matching virt_offset
+                for offset_to_page_within_mapping in range(0, phy_mapping_size, 0x1000):
```
ikelos, line 184: Hard coded 0x1000

volatility3/framework/plugins/windows/strings.py
```diff
@@ -147,14 +171,46 @@ def generate_mapping(
...
+                    # calculate the page number for this phy_offset, e.g. the ">> 12"
+                    # drops the bits that would address an offset within the page.
+                    # This means that all offsets within the same page get the same
+                    # physical_page number.
+                    physical_page = (
+                        phy_mapping_size + offset_to_page_within_mapping
+                    ) >> 12
```
ikelos, line 191: Hard coded 12

volatility3/framework/plugins/windows/strings.py
```diff
@@ -178,22 +234,37 @@ def generate_mapping(
...
+                            ) = mapval
+                            for offset_to_page_within_mapping in range(
+                                0, phy_mapping_size, 0x1000
                             ):
-                                cur_set = reverse_map.get(mapped_offset >> 12, set())
-                                cur_set.add(
-                                    (f"Process {process.UniqueProcessId}", offset)
+                                physical_page = (
+                                    phy_offset + offset_to_page_within_mapping
+                                ) >> 12
```
ikelos, line 253: Hard coded 12

volatility3/framework/plugins/windows/strings.py
```diff
@@ -81,22 +87,40 @@ def _generator(self) -> Generator[Tuple, None, None]:
...
+            # that this offset can be added to mappings which are all
+            # page aligned. This ensures that a string located at phy
+            # add 0x1e64cd20 would carry the 0xd20 to the virtual offsets
+            # displayed in the plugin output. Without this it would show
+            # only the page that the string was found, rather than the
+            # actually addr. 0xFFF is 4095 e.g. all lower bits set.
+            offset_within_page = phys_offset & 0xFFF
+
+            mapping_entry = revmap.get(
+                phys_offset >> 12,
```
ikelos, line 103: Hard coded 12

ikelos (commented):
Something to consider (that unfortunately I only just thought of) is that if there are multiple layers (like a QEMU layer, or AVML layer or similar) then the offsets in the strings file won't map up to the "physical" addresses returned.  I'm not sure how best to deal with that, but it probably need a little consideration.

Options include generating our own strings style tool, or jiggering the virtual to physical map to translate down to the lowest possible layer (`DataLayer`) or adding a debug message pointing out that the mapping will have been inaccurate if the physical layer isn't a `DataLayer`?

ikelos:
There is then also the issue of multiple physical layers (eg, swap space, etc).  At the moment, I think we throw the layer name from the mapping value away.  I don't know whether to hold this PR up to get all these edge cases sorted or just ok it with the BaseAbsentValue and hardcoding changes?

eve-mem:
It would be pretty easy to make our own version of "strings". You could import from the various yarascan plugins using a very basic regex yara signature of X number of printable characters. That'll handle all the address translation etc, and would allow pid filtering etc. 

Nothing stopping a user doing that right now (i know i have myself in the past) so it would really be a convenience function than anything new. 

I could make a rough example if that sounds like it might be useful?

It would mean that if yarascan supports the page files etc, then so does this plugin. It also means it won't work at all unless the user has installed the extra Yara packages.

eve-mem:
Actually if you wanted to go the "own version of the strings tool" type plugin, rather than "reading the results of the actual strings utility" type plugin then you could even use the RegexScanner to make a basic strings type plugin - therefore not needing yara extremely rough example here: https://github.com/eve-mem/volatility3/tree/regex_based_strings (would need to also handle unallocated space, etc, etc)

On e benefit I see with keeping the current "reading the results of the actual strings utility" type plugin is that as a user I can provided a curated list of strings I'm interested in translating - rather than just everything in one go.

ikelos:
Yep, so the curated list is important and valuable (string scanning can be notoriously slow for python over the `strings` program.  It also means we wouldn't get an more efficient algorithms or anything if they invented them, so it'd be nice as a toy and one I'm happy for us to put in contrib, but I'm not sure it's the right way for the main framework (we don't want to do everything ourselves and become a massive rewrite of the forensics world, better to just take input files of a fixed format and process them).  So @eve-mem, if you want to submit your plugin as a PR, please feel free to, but put them in the main plugins directory, rather than `framework/plugins`, similar to the not-officially-supported certificates plugin.

So then the question still exists, do we merge this before doing a rework of the strings plugin, or do we just dive straight into the rework (and use this as motivation to get it written)?  I'm happy with either, but I would still like any changes mentioned above making if we are going to merge this in (I think they're mainly consolidating fixed values so they're changing in one location, or derived from the space itself)...

eve-mem:
It's a really good question @ikelos, right now the strings plugin doesn't really working at all - it would be great to sort that - the changes here sort a lot of that. Thanks to the hard work from @kevthehermit. So at least windows strings works mostly as expected with some edge cases missed. 

If redoing the strings again is going to take "a while" which it feels like it might, e.g. probably not before the next release? Then it makes sense to me to merge this, then rehaul it all more generically covering all the bases in slower time (3-9 months?)

To rehaul would need to design how users input a strings file from a page file etc, so they could provide 5+ strings files which all line up with different layers in vol - so how do we make it easy to do that on the command line etc. That's just one part of it, with other bits that need thinking about too. To me it just feels like it might take a while?

I am just a random (admittedly keen!) user, I'm happy with whatever you and the other core devs think is the best way to approach it. Give it some thought and if there are bits i can help with let me know.

kevthehermit:
I agree with everything here :) 

The current strings plugin is broken, this "patches that" but in all my testing i have been restricting strings to a specific pid, when i try to run it without defining a pid it never completed, I had it running on a 4Gb Win11 dump ~ 145 Pids for 18 hours and it had still not finished with output,

So yes i think it would be better to start again and think about a more sensible way thats also more agnostic against OS. 

But in the mean time this does at least make it usable if not "perfect"

ikelos:
Building the reverse map will take a long time, if you know of a quicker way to determine which memory block belong to which processes, I'm all for it (the maps should all share the kernel, so there may be something we can do there to scan the kernel and then only the differing entries from the process maps?).

I'm happy to merge this, just need those last batch of comments addressed first please...

eve-mem:
Hello @ikelos and @kevthehermit,

TL;DR - if we use an interval tree we might be able to speed up the strings plugin when there are manageable number of strings to lookup, but it means adding an optional dependency to vol. 

Caveat: Lots more testing needed on many other samples!

I was trying to work out different ways to make the creating the rev maps faster. Currently the strings plugin makes a dict that makes every page of physical memory to the matching virtual pages. Once it's made, its extremely fast as it's all just look ups. 

The downside is it can take a long time to create, on the test 512MB windows sample I normally use it takes about 5 minutes. I'm also testing on a16GB sample now but it's taking a very long time to finish so I don't have results for that yet. With @kevthehermit having a sample still creating a reverse map after 18 hours I thought it might be useful to try some other options. 

I've made some rough proof of concept examples here for two different ways that could be possible options (they'll need more work if they seem like good ideas): https://github.com/volatilityfoundation/volatility3/compare/develop...eve-mem:volatility3:windows_strings_intervaltree

It's based on the current changes for this PR, my git-fu is not good enough to let me make a branch from this though before it's committed... 

The changes give three options for creating the reverse maps. The current method, a tree method (using interval tree), and something I've called "ordered".

I'm not suggesting for the actual plugin these options exist, I think we should pick one and stick with it. This just felt like an easy way to share my experiment. 

The tree method uses an interval tree (ref: https://pypi.org/project/intervaltree/ and https://en.wikipedia.org/wiki/Interval_tree) as it's a really lovely way to build a structure that has start points and end points. You simply add all the ranges to the tree and then later you can ask which ranges can be found at point X. The tree takes care of making that all efficient and ensuring you get the right answer.  It would mean adding an optional dependency though. If we wanted to, we could probably make a tree type structure ourselves that would be quite fast. 

The ordered method is my attempt at something not using any extra dependency's. It's very hard to read and is likely extremely inefficient for what it is. The idea is to put all the mappings from all the processes in order, and then place all the strings to lookup in order too. That way as you loop though checking for matches you can begin to remove mappings from the list once they're no longer possible and you can stop early too. However you're still making many many comparisons per string until you find the correct mapping. 

As far as I can tell each method gives the same results.

I run some tests, using each method to look up 1,10, 100, 1,000, 10,000, and 100,000 strings and timed how long they took on this 512MB sample. Just once for each, it would be better to do more rounds and average them out. I didn't use a config either so vol is scanning for the symbols each time. 

Here are the results
image

To me it's pretty clear that the current method is not really affected by the number of strings to lookup. Once that reverse map is created it's extremely fast - the only issue being that it can take a long time to create in the first place. The interval tree method seems to be pretty fast and it's the approach I prefer personally. The crazy "ordered" approach also seems to work, and maybe with some effort we could make an approach to lookup up offsets that wasn't this crazy and was still a little faster.

I'm sure if I added more and more lines to search both the interval tree and the "ordered" method would end up being slower overall as they require more post processing per string, with enough lookups I'd imagine it could end up being significantly faster to stick with the current method. Equally 100,000 is quite a lot of strings - I'd guess that in most cases that would be useful enough - I know that's more than I'd personally need.

With this small memory sample we're just talking about the difference between 40 seconds and 5 minutes which isn't that huge, but with the larger memory samples that initial creation time with the current method is prohibitive. I've managed to run a test with the tree method on my 16GB sample but I've not yet been able to finish the current method. 

I've not looked at multi threading the reverse map - that would help speed things up too. 

What do you both think? Should I do more testing and experiments around this? Any particular method that would be preferable? 

Raw data if it's useful:
| **Lines**  | **Tree**      | **Ordered** | **Current**     |
| --------- | ---------- | ----------- | ----------- |
| 1      | 36.1755014 | 154.8891513 | 324.7267874 |
| 10     | 34.1541335 | 155.8885713 | 315.4460056 |
| 100    | 33.1670267 | 151.8015634 | 313.7029176 |
| 1000   | 34.2455712 | 152.7433688 | 317.4841116 |
| 10000  | 45.241863  | 160.7236966 | 316.5802767 |
| 100000 | 87.3827075 | 202.0599115 | 319.4830912 |

ikelos:
Well, I still think building a reverse mapping will be useful for lots of things.  I'm interested in the speed savings you get from the interval tree, but it sounds like it's reimplementing grep, so I need to figure out what tricks it's doing.  I still think if we do proper some jiggery pokery just involving the page tables, we should be able to much more efficiently build up a suitable structure letting us look up an offset and determine which processes and where that offset maps up to.  Unfortunately I haven't really had time to dedicate to it, so, I guess keep going with this.

I'm not overly keen on a whole dependency for what's presumably quite a small component, particularly if the technology is static and not constantly being improved, however if it's used often and in other places, or it's too complicated to easily write ourselves/verify what we've written then I'd be ok with it.  Hopefully that gives you an idea of my thinking?

This also is a plugin specific dependency, not a framework-wide one, and we can already deal with that as with Yara and the crypt plugins, so again, less of a huge concern...

eve-mem:
Thanks for the quick reply! 

The interval tree is a fast data structure for this kind of problem rather than a grep type thing. I think it was originally made with dates and time spans in mind. It's not actually searching for any of the strings if you see what I mean. I agree with you though about adding a whole dependency just for strings. 

I also thought there might be some cleverness to be had in the page maps, just looking at the top level you can see all the holes where there would be no possible mappings, and you know what virtual address ranges they cover. By making a allow/deny list by looking at the top (or maybe also the second level) for the pages maps per process you should have a, hopefully, quick way of knowing which virtual address will translate. That only helps in the VtoP direction, I've not got any good ideas for PtoV yet - I'll keep thinking! Please shout if you get any good ideas. 

In the mean time this probably isn't worth holding up this PR for - probably better that strings actually gives the correct answer even if it does take a while.

ikelos:
Ok, I'm still not sure I understand how it's faster then, but I'll have a dig through when I get some time.

It was more that we have as many maps as there are processes, but a huge amount of the map should be the same between the kernel map and the process map, although thinking about it, we must already cope with that somehow.   Once we work on unifying processes, we might be able to make the plugin generic too.

In general, my thinking was just take the diff from the kernel map as the "this is really in this process" map?  We could also later apply OS specific knowledge to know what's actually part of the process and what isn't (VADs, etc)?

I'll keep having a think, but I feel sure there are improvements to be made...

eve-mem:
Yes, I keep thinking there is something that can be done too. I have a few more ideas I'll play with, if you hear nothing from me it's because they were terrible  :D

eve-mem:
Hello again @ikelos and @kevthehermit, sorry for spamming this PR. 

I have had some luck with a very basic implementation of a interval tree style way of doing the reverse mappings, but just using basic python therefore not needing any extra dependencies. It can't do all the fancy things a real interval tree can do but it seems to do the part I needed reasonably well. It doesn't try to produce a balanced tree or anything clever like that. 

I've added it to the same experimental branch here: https://github.com/volatilityfoundation/volatility3/compare/develop...eve-mem:volatility3:windows_strings_intervaltree

It's under the `--custom` option if you wanted to try it. It uses the mini `MappingNode` and `MappingTree` classes I've just haphazardly put into the strings plugin. If it looks useful I can tidy it all up.

In terms of speed it seems to be coming pretty close to the real interval tree on the smaller memory sample. 
image

What do you think of this approach?

Raw numbers again if it helps:
| **Lines**  | **Tree**      | **Ordered** | **Current**     | **Custom** |
| --------- | ---------- | ----------- | ----------- | ----------- |
| 1      | 36.1755014 | 154.8891513 | 324.7267874 | 32.2218848 | 
| 10     | 34.1541335 | 155.8885713 | 315.4460056 | 32.3109088 |
| 100    | 33.1670267 | 151.8015634 | 313.7029176 | 33.2424613 |
| 1000   | 34.2455712 | 152.7433688 | 317.4841116 | 34.2646967 | 
| 10000  | 45.241863  | 160.7236966 | 316.5802767 | 42.3558852 |  
| 100000 | 87.3827075 | 202.0599115 | 319.4830912 | 86.6992142 |

ikelos:
So I've had a chance to look at this, and I really like the custom version.  It doesn't require additional modules but gives a a massive speed up (I've tested it out and it's blazingly fast).  It appears to give the same/better results (although, it tends to go through the strings a process at a time, whereas it might be better to generate all the maps and then start outputting so that all string entries are next to each other (although sorting can be done by the UI/later).  I'd ditch the other two methods since they all achieve the same results (and there's no point having a flag between the methods if they produce identical results).

At the moment we throw away the actual layer it was read from (making the assumption there's only one physical layer, which isn't always true, for example swap spaces, etc), but certainly eve's custom one is better than the current strings implementation.

Next steps are to get it its own PR, so it's tidied up and the other methods aren't cluttering it, then once it goes through review we can get it merged, and if it proves @kevthehermit 's works out faster we can study it to see why and what tweaks we might want.  Does that sound ok to everyone?

ikelos:
(Also, I think the custom version can inherit from dataclasses.dataclass and get a bunch of memory saving for free that way).

eve-mem:
Thanks for taking a look @ikelos. I'll have a go at working this into a separate PR and we can go from there.

ikelos:
@eve-mem, I forget, did a separate PR come in for this?  I've pushed a testing branch `eve-string-speedup` for testing.  I'm just concerned this has been waiting a few months and I don't want it to fall off the cliff of forever...  5;P

Happy if you've got one in the werks to review it, but this was a pretty signficant speed-up you and @kevthehermit came up with and I'd like to get it in the tree...

eve-mem:
@ikelos - I never got finished with something that worked in all cases. I had some problems with that speed up code hitting a recursion limit on some windows samples and I've been tinkering away trying to come up with ways around it without much luck yet (both a brain power and not enough time problem).

Using the actual interval tree package seems to work fine - but that's a whole new dependency for quite a niche use.

ikelos:
Why would their intervaltree work, but yours not?  Could you take a look in their code and see if they've got some special recursion handling?  I didn't hit any of these on my spin on your patch (`feature/eve-string-speedup`).  Could you check that and see if it runs into the recursion issue, and if so send me over an example (and/or a sample image to test on it) please?  I can't see any direct recursion (just creating objects within objects)...

eve-mem:
Essentially there is an actual interval tree, while mine is quite a basic tree. I thought I'd be able to cut out lots of parts of an interval tree that aren't needed, and the result is something that's hardly an interval tree at all. 

I'm still keen to get this resolved, a strings plugin that works and is quick enough to use is really important. 

Sorry it's taking me so long!

PR #1044, opened 2023-11-21
Feature: Alternate Data Streams.

forensicxlab:
Hello, 
Just added a plugin next to MFTScan to extract ADS from the carved MFT records.
The ISF was also updated to include missing attributes.

Research : https://www.forensicxlab.com/posts/volads/

Best regards.

### DIFF
--- a/volatility3/framework/plugins/windows/mftscan.py
+++ b/volatility3/framework/plugins/windows/mftscan.py
@@ -224,12 +224,12 @@ class ADS(interfaces.plugins.PluginInterface):
             config_path=self.config_path,
             sub_path="windows",
             filename="mft",
-            class_types={"MFT_ENTRY": mft.MFTEntry,"FILE_NAME_ENTRY": mft.MFTFileName},
+            class_types={"MFT_ENTRY": mft.MFTEntry,"FILE_NAME_ENTRY": mft.MFTFileName, "ATTRIBUTE": mft.MFTAttribute},
         )
 
         # get each of the individual Field Sets
         mft_object = symbol_table + constants.BANG + "MFT_ENTRY"
-        header_object = symbol_table + constants.BANG + "ATTR_HEADER"
+        attribute_object = symbol_table + constants.BANG + "ATTRIBUTE"
         fn_object = symbol_table + constants.BANG + "FILE_NAME_ENTRY"
 
         # Scan the layer for Raw MFT records and parse the fields
@@ -243,58 +243,32 @@ class ADS(interfaces.plugins.PluginInterface):
                 # We will update this on each pass in the next loop and use it as the new offset.
                 attr_base_offset = mft_record.FirstAttrOffset
 
-                attr_header = self.context.object(
-                    header_object,
+                attr = self.context.object(
+                    attribute_object,
                     offset=offset + attr_base_offset,
                     layer_name=layer.name,
                 )
 
                 # There is no field that has a count of Attributes
-                # Keep Attempting to read attributes until we get an invalid attr_header.AttrType
+                # Keep Attempting to read attributes until we get an invalid attr.AttrType
                 file_name = renderers.NotAvailableValue
                 is_ads = False
-                    # The First $DATA Attr is the 'principal' file itself not the ADS
-                while attr_header.AttrType.is_valid_choice:
-                    # Offset past the headers to the attribute data
-                    attr_data_offset = (
-                        offset
-                        + attr_base_offset
-                        + self.context.symbol_space.get_type(
-                            header_object
-                        ).size
-                    )
+                
+                # The First $DATA Attr is the 'principal' file itself not the ADS
+                while attr.Attr_Header.AttrType.is_valid_choice:
 
-                    if attr_header.AttrType.lookup() == "FILE_NAME":
-                        attr_data = self.context.object(
-                            fn_object, offset=attr_data_offset, layer_name=layer.name
-                        )
+                    if attr.Attr_Header.AttrType.lookup() == "FILE_NAME":
+                        attr_data = attr.Attr_Data.cast(fn_object)
                         file_name = attr_data.get_full_name()
-                    
-                    # DATA Attribute (can be ADS or not)
-                    if attr_header.AttrType.lookup() == "DATA":
+                
+                    if attr.Attr_Header.AttrType.lookup() == "DATA":
                         if is_ads:
-                            if not attr_header.NonResidentFlag:
+                            if not attr.Attr_Header.NonResidentFlag:
                                 # Resident files are the most interesting.
-                                if attr_header.NameLength > 0:
-                                    attr_name_offset = (
-                                        offset
-                                        + attr_base_offset
-                                        + attr_header.NameOffset
-                                    )
+                                if attr.Attr_Header.NameLength > 0:
 
-                                    ads_name = self._context.layers[layer.name].read(
-                                        attr_name_offset, attr_header.NameLength*2 , pad=True
-                                    ).decode('utf-16')
-
-                                    attr_content_offset = (
-                                        offset
-                                        + attr_base_offset
-                                        + attr_header.ContentOffset
-                                    )
-                                                                   
-                                    content = self._context.layers[layer.name].read(
-                                        attr_content_offset, attr_header.ContentLength , pad=True
-                                    )
+                                    ads_name = attr.get_resident_filename()
+                                    content =  attr.get_resident_filecontent()
 
                                     # Preparing for Disassembly
                                     architecture = layer.metadata.get("architecture", None)
@@ -303,10 +277,10 @@ class ADS(interfaces.plugins.PluginInterface):
                                     )
 
                                     yield 0, (
-                                        format_hints.Hex(attr_data_offset),
+                                        format_hints.Hex(attr_data.vol.offset),
                                         mft_record.get_signature(),
                                         mft_record.RecordNumber,
-                                        attr_header.AttrType.lookup(),
+                                        attr.Attr_Header.AttrType.lookup(),
                                         file_name,
                                         ads_name,
                                         format_hints.HexBytes(content),
@@ -317,18 +291,17 @@ class ADS(interfaces.plugins.PluginInterface):
                             
                             
                     # If there's no advancement the loop will never end, so break it now
-                    if attr_header.Length == 0:
+                    if attr.Attr_Header.Length == 0:
                         break
 
                     # Update the base offset to point to the next attribute
-                    attr_base_offset += attr_header.Length
+                    attr_base_offset += attr.Attr_Header.Length
                     # Get the next attribute
-                    attr_header = self.context.object(
-                        header_object,
+                    attr = self.context.object(
+                        attribute_object,
                         offset=offset + attr_base_offset,
                         layer_name=layer.name,
                     )
-
     def run(self):
         return renderers.TreeGrid(
             [
--- a/volatility3/framework/plugins/windows/mftscan.py
+++ b/volatility3/framework/plugins/windows/mftscan.py
@@ -258,17 +258,14 @@ class ADS(interfaces.plugins.PluginInterface):
                                     content = attr.get_resident_filecontent()
                                     if content:
                                         # Preparing for Disassembly
+                                        disasm = interfaces.renderers.BaseAbsentValue
                                         architecture = layer.metadata.get(
                                             "architecture", None
                                         )
-
-                                        disasm = (
-                                            interfaces.renderers.Disassembly(
+                                        if architecture:
+                                            disasm = interfaces.renderers.Disassembly(
                                                 content, 0, architecture.lower()
                                             )
-                                            if architecture
-                                            else interfaces.renderers.BaseAbsentValue
-                                        )
                                     else:
                                         content = renderers.NotAvailableValue
                                         disasm = interfaces.renderers.BaseAbsentValue
--- a/volatility3/framework/plugins/windows/mftscan.py
+++ b/volatility3/framework/plugins/windows/mftscan.py
@@ -59,7 +59,7 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
         fn_object = symbol_table + constants.BANG + "FILE_NAME_ENTRY"
 
         # Scan the layer for Raw MFT records and parse the fields
-        for offset, _rule_name, _name, _value in layer.scan(
+        for offset, _, _, _ in layer.scan(
             context=self.context, scanner=yarascan.YaraScanner(rules=rules)
         ):
             with contextlib.suppress(exceptions.PagedInvalidAddressException):
@@ -253,8 +253,9 @@ class ADS(interfaces.plugins.PluginInterface):
 
                 # There is no field that has a count of Attributes
                 # Keep Attempting to read attributes until we get an invalid attr_header.AttrType
-                file_name = ""
-                is_ads = 0
+                file_name = "N/A"
+                is_ads = False
+                    # The First $DATA Attr is the 'principal' file itself not the ADS
                 while attr_header.AttrType.is_valid_choice:
                     # Offset past the headers to the attribute data
                     attr_data_offset = (
@@ -274,7 +275,7 @@ class ADS(interfaces.plugins.PluginInterface):
 
                     # DATA Attribute (can be ADS or not)
                     if attr_header.AttrType.lookup() == "DATA":
-                        if is_ads > 0:
+                        if is_ads:
                             if not attr_header.NonResidentFlag:
                                 # Resident files are the most interesting.
                                 if attr_header.NameLength > 0:
@@ -313,8 +314,7 @@ class ADS(interfaces.plugins.PluginInterface):
                                         disasm,
                                     )
                         else:
-                            # The First Data Attr is the file itself not the ADS
-                            is_ads+= 1
+                            is_ads = True
                             
                             
                     # If there's no advancement the loop will never end, so break it now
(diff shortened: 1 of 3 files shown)


volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -58,7 +59,7 @@ def _generator(self):
         fn_object = symbol_table + constants.BANG + "FILE_NAME_ENTRY"
 
         # Scan the layer for Raw MFT records and parse the fields
-        for offset, _rule_name, _name, _value in layer.scan(
+        for offset, _, _, _ in layer.scan(
```
ikelos: This doesn't improve readability?  I'm guessing an automated tool decided to do it, but the extra characters don't hurt anyone, so please don't take them away...  5:)

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -189,3 +190,158 @@ def run(self):
...
+        mft_object = symbol_table + constants.BANG + "MFT_ENTRY"
+        header_object = symbol_table + constants.BANG + "ATTR_HEADER"
+        attribute_object = symbol_table + constants.BANG + "ATTRIBUTE"
+        fn_object = symbol_table + constants.BANG + "FILE_NAME_ENTRY"
+
+        # Scan the layer for Raw MFT records and parse the fields
+        for offset, _rule_name, _name, _value in layer.scan(
+            context=self.context, scanner=yarascan.YaraScanner(rules=rules)
+        ):
+            with contextlib.suppress(exceptions.PagedInvalidAddressException):
```
ikelos, line 227: Does this happen often?  Would it be useful to developers to have it logged when it does happen or is it really just noise?

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -189,3 +190,158 @@ def run(self):
...
+
+                attr_header = self.context.object(
+                    header_object,
+                    offset=offset + attr_base_offset,
+                    layer_name=layer.name,
+                )
+
+                # There is no field that has a count of Attributes
+                # Keep Attempting to read attributes until we get an invalid attr_header.AttrType
+                file_name = "N/A"
```
ikelos: This should be one of the a `renderers.NotAvailableValue` so that the UI can decide how to display it please.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -189,3 +190,158 @@ def run(self):
...
+                is_ads = False
+                    # The First $DATA Attr is the 'principal' file itself not the ADS
+                while attr_header.AttrType.is_valid_choice:
+                    # Offset past the headers to the attribute data
+                    attr_data_offset = (
```
ikelos: Rather than using `relative_child_offset` to then construct an object, consider using 
```
file_name = asttr_header.Attr_Data.cast(fn_object).get_full_name()
```

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -189,3 +190,158 @@ def run(self):
...
+                        if is_ads:
+                            if not attr_header.NonResidentFlag:
+                                # Resident files are the most interesting.
+                                if attr_header.NameLength > 0:
+                                    attr_name_offset = (
+                                        offset
+                                        + attr_base_offset
+                                        + attr_header.NameOffset
+                                    )
+                                    ads_name = self._context.layers[layer.name].read(
```
ikelos: Again, please cast this a `string` of some kind, rather than reimplementing the string decoding stuff again.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -189,3 +190,158 @@ def run(self):
...
+                                        attr_name_offset, attr_header.NameLength*2 , pad=True
+                                    ).decode('utf-16')
+                                    attr_content_offset = (
+                                        offset
+                                        + attr_base_offset
+                                        + attr_header.ContentOffset
+                                    )                                    
+                                    content = self._context.layers[layer.name].read(
```
ikelos: This also feels a little nasty (I'm not keen on plugins doing direct layer reads), but I'm not sure how to do it more cleanly and this does seem to be unstructured binary data that just needs reading.  I think an array of bytes would just make this uglier, so probably best to keep it as it is.

ikelos (commented):
Thanks very much, this looks like it should make a good addition to the framework!  There's a few bits to look at (mostly using `cast` rather than getting the offset and constructing a new object on top of it), but otherwise looks good.  As it's a windows plugin, I'd still like @iMHLv2 to give it a review too though...

volatility3/framework/plugins/windows/mftscan.py
forensicxlab: Hello,  
My fault when reviewing my code, not an automated tool, I did not modify the right line as it is the same in the MFTScan and ADS 😅

However, I have always used '_' to denote values not used later in the code (a habit from Rust). I thought it was also a good practice in python3 when unpacking values from a tuple, my mistake, you're right about the code readability. 
I will correct that :)

volatility3/framework/plugins/windows/mftscan.py
forensicxlab: Hello, 
Thanks, I will make the change :)

volatility3/framework/plugins/windows/mftscan.py
forensicxlab: It does happen often indeed on the memory dump I have tested. This line is also present in the already existing mftscan.MFTScan plugin. 

When not deleting the exception, the search is stopped when a pagefault occurs : 

```
0xf9800e8705e8  FILE    75201   DATA    Clockres64.exe  Zone.Identifier
5b 5a 6f 6e 65 54 72 61 [ZoneTra
6e 73 66 65 72 5d 0d 0a nsfer]..
5a 6f 6e 65 49 64 3d 33 ZoneId=3
0x0:    pop     rbx
0x1:    pop     rdx
0x2:    outsd   dx, dword ptr [rsi]
0x3:    outsb   dx, byte ptr [rsi]
0x4:    push    rsp
0x6:    jb      0x69
0x8:    outsb   dx, byte ptr [rsi]
0x9:    jae     0x71
0xb:    jb      0x6b
0xe:    or      eax, 0x6e6f5a0a

Page error 0xf9800f68468a in layer primary (Page Fault at entry 0xf8a001c598200400 in page entry)

```

When suppressing the execption more results are carved : 

```
0xf9800e8705e8  FILE    75201   DATA    Clockres64.exe  Zone.Identifier
5b 5a 6f 6e 65 54 72 61 [ZoneTra
6e 73 66 65 72 5d 0d 0a nsfer]..
5a 6f 6e 65 49 64 3d 33 ZoneId=3
0x0:    pop     rbx
0x1:    pop     rdx
0x2:    outsd   dx, dword ptr [rsi]
0x3:    outsb   dx, byte ptr [rsi]
0x4:    push    rsp
0x6:    jb      0x69
0x8:    outsb   dx, byte ptr [rsi]
0x9:    jae     0x71
0xb:    jb      0x6b
0xe:    or      eax, 0x6e6f5a0a
0xf9800fe149f0  FILE    75346   DATA    ctrl2cap.nt4.sys        Zone.Identifier
5b 5a 6f 6e 65 54 72 61 [ZoneTra
6e 73 66 65 72 5d 0d 0a nsfer]..
5a 6f 6e 65 49 64 3d 33 ZoneId=3
0x0:    pop     rbx
0x1:    pop     rdx
0x2:    outsd   dx, dword ptr [rsi]
0x3:    outsb   dx, byte ptr [rsi]
0x4:    push    rsp
0x6:    jb      0x69
0x8:    outsb   dx, byte ptr [rsi]
0x9:    jae     0x71
0xb:    jb      0x6b
0xe:    or      eax, 0x6e6f5a0a
0xf9800fe1c968  FILE    75378   DATA    Eula.txt        Zone.Identifier
5b 5a 6f 6e 65 54 72 61 [ZoneTra
6e 73 66 65 72 5d 0d 0a nsfer]..
5a 6f 6e 65 49 64 3d 33 ZoneId=3
[...]
```

volatility3/framework/plugins/windows/mftscan.py
forensicxlab: I agree, I was also looking for the best way. Looked at other plugins, seems like netstat for example is doing something similar, I figured it might be accepted like that.

```python3
        part_table = context.object(
            net_symbol_table + constants.BANG + "_PARTITION_TABLE",
            layer_name=layer_name,
            offset=part_table_addr,
        )
        part_count = int.from_bytes(
            context.layers[layer_name].read(tcpip_module_offset + part_count_symbol, 1),
            "little",
        )
```

volatility3/framework/plugins/windows/mftscan.py
ikelos: It is good practice, you can see in the original ones their variables names all start with `_`, but you can also provide an indication of what the variable would have held, should people come to copy and paste the code in the future...  5;)

volatility3/framework/plugins/windows/mftscan.py
ikelos: Ok, that's fine, I just wanted to check, thanks  5:)

volatility3/framework/plugins/windows/mftscan.py
ikelos: Yeah, I'd leave it as it is.  It probably is the right way of doing it, it just gave me the creeps so I thought I'd mention it...  5;P

volatility3/framework/plugins/windows/mftscan.py
forensicxlab: No sure I understood, 
The ADS filename and the ADS content (unlike the $FILE_NAME attr) are not at a predictable offset, let me know if I've missed something and your insight on how I could do the string casting in a better way  😊

Best regards.

volatility3/framework/plugins/windows/mftscan.py
forensicxlab: So, I've looked into this:

The folk(s) that have written the mftscan.MFTScan actually did this trick to construct the attr_data object for $FILE_NAME.

Turns out that the "ATTR_HEADER" structure definition in the symbol table was incomplete (missing the ContentLength & ContentOffset + padding of 2 bytes) which is changing the struct size to 24.

Let me know what you think about the modifications made. I have also fixed that in the mftscan.MFTScan plugin. Looks more adapted for me 😊


Best regards.

ikelos: You're right, `mftscan` does it the same way, but it's pretty ugly, so I'm going to look into a pull request fixing that up, but that can happen after this code lands (and I'll include this code in there).

volatility3/framework/plugins/windows/mftscan.py
ikelos: Ugh, so unfortunately it looks like this isn't unreasonable, but I would much prefer to see a custom class (like [`MFTFileName`](https://github.com/volatilityfoundation/volatility3/blob/9f9afbf90fabebbdb8e4c62d022820fdb3b0a260/volatility3/framework/symbols/windows/extensions/mft.py#L16) ) define this, than have it buried in the midst of this code here.

ikelos:
So you can see how I somewhat expected things to be done by looking at the `issues/improve-mftscan` branch, which changes away the manual offset modification and reading of data to instead construction objects and use the volatility object model to access everything.  Hopefully this is much easier/cleaner to read which is why I've been pushing for it...

ikelos (commented):
As mentioned above (sorry, I thought I'd posted this already), it does a lot of things manually which makes a great place for errors to sneak in.  Rather than doing lots of offset maths out here, I'd much prefer to use the object model which is exactly what volatility is designed for.  In the PR #1049 it gives an example of how I was expecting things to run in this...

forensicxlab:
Hello @ikelos, 

I've introduced your fix to the ADS plugin so you can include your fix when this code lands. 
Also added the `MFTAttribute`in a custom class.

I hope you'll be happy with those changes. Learned a lot again about the framework, thank you for all your comments 😊


Best regards.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -189,3 +189,130 @@ def run(self):
...
+                            if not attr.Attr_Header.NonResidentFlag:
+                                # Resident files are the most interesting.
+                                if attr.Attr_Header.NameLength > 0:
+
+                                    ads_name = attr.get_resident_filename()
+                                    content =  attr.get_resident_filecontent()
+
+                                    # Preparing for Disassembly
+                                    architecture = layer.metadata.get("architecture", None)
+                                    disasm = interfaces.renderers.Disassembly(
```
ikelos: How will this respond if the architecture is None?  Does it need wrapping in an exception handler (and then setting this `disasm` to a `BaseAbsentValue` descended class?

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -21,3 +21,25 @@ def get_full_name(self) -> str:
...
+
+class MFTAttribute(objects.StructType):
+    """This represents an MFT ATTRIBUTE"""
+
+    def get_resident_filename(self) -> str:
+        # To get the resident name, we jump to relative name offset and read name length * 2 bytes of data
+        layer = self._context.layers[self.vol.layer_name]
+        attr_name_offset = self.vol.offset + self.Attr_Header.NameOffset
+       
+        return self._context.layers[layer.name].read(
```
ikelos: I feel like it'd be possible to make this a cast to `UNICODE_STRING` with some tweaking to the structure, but I'm ok with this as is if you don't want to.

ikelos: This might end up throwing decoding exceptions which will need catching where it gets called.

ikelos (commented):
Yep, this looks much better!!!  5:D  I'd prefer it if there was a little more error handling in places, but I'd be ok merging this as is if you really don't want to add the error handling, just let me know.  Hopefully you're happier with it too?  I think it's much more readable and easier to make sense of now...  5:)

forensicxlab:
> Yep, this looks much better!!! 5:D I'd prefer it if there was a little more error handling in places, but I'd be ok merging this as is if you really don't want to add the error handling, just let me know. Hopefully you're happier with it too? I think it's much more readable and easier to make sense of now... 5:)

Yeap let's make things clean ! :) I will propose some exception handling based on your remarks in the coming days ;) 
I am happy with it too, much cleaner and logic. Also a good exercise for me for future pull requests 😊 

Many thanks.

ikelos:
Thanks, I'm afraid another commit snuck in that conflicts with your pull request.  If you could fix it up then we can get it merged in...  5:)

ikelos:
Also please ensure the black formatter runs without issue, so the code formatting is consistent with the rest of the codebase...  5:)

forensicxlab:
Hello @ikelos, 
Took some time but I think I have found a way to extract the ADS filename and the ADS content in a pretty way using the object model (no more direct layer reads).
 
Let me know what you think, I have the feeling that this is much cleaner... :) 

Example output.
```
Volatility 3 Framework 2.5.1
Offset  Record Type     Record Number   MFT Type        Filename        ADS Filename    Hexdump Disasm

0xf98001423528  FILE    78477   DATA    PackageManagement_x64 (1).msi   Zone.Identifier
5b 5a 6f 6e 65 54 72 61 [ZoneTra
6e 73 66 65 72 5d 0d 0a nsfer]..
5a 6f 6e 65 49 64 3d 33 ZoneId=3
0x0:    pop     rbx
0x1:    pop     rdx
0x2:    outsd   dx, dword ptr [rsi]
0x3:    outsb   dx, byte ptr [rsi]
0x4:    push    rsp
0x6:    jb      0x69
0x8:    outsb   dx, byte ptr [rsi]
0x9:    jae     0x71
0xb:    jb      0x6b
0xe:    or      eax, 0x6e6f5a0a
```



Best regards.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -189,3 +173,143 @@ def run(self):
...
+                                        # Preparing for Disassembly
+                                        architecture = layer.metadata.get(
+                                            "architecture", None
+                                        )
+
+                                        disasm = (
+                                            interfaces.renderers.Disassembly(
+                                                content, 0, architecture.lower()
+                                            )
+                                            if architecture
```
ikelos: I think this would be easier to read as:

```
disasm = interfaces.renderers.BaseAbsentValue
if architecture:
   disasm = interfaces.renderers.Disassembly(content, 0, architecture.lower())
```

ikelos: Also, I get that our assumption is that ADS is most commonly used for hiding executable code, but could this not give a lot of false positives if the data stores in it is just data?  Also there seems to be no limit on the amount of data that might be disabled?  I assume files that are too large aren't stored in memory, but I just want to know the largest amount of data we might expect to find there, and how it might affect the output in volatility?

ikelos (commented):
Yep, this looks very clean.  Just a syntactical change to make it a little easier to read, and then just a little question about always disassembling what's found in the ADSes...

volatility3/framework/plugins/windows/mftscan.py
forensicxlab: I understand, 
So, the ADS is mainly used legitimately by Microsoft to introduce for example the Zone.Identifier feature, which is a nice digital forensics clear text artefact to identify the origin of a file. This will be for me our first assumption. Indeed, the value of the ZoneID indicate whether a file is from the internet (3), trusted(2), untrusted(4) etc...The referer url might also be present.
This is a nice pivot point which is the reason for the Hexdump view in order for the analyst to check it in clear text.

A derived usage of this feature can indeed be from an adversary trying to hide malicous code like you said (our second assumption). Although there is some false positives when an ADS doesn't contain any clear text data, the analyst might want to check that the name of the ADS is [a common data stream used by Windows](https://flatcap.github.io/linux-ntfs/ntfs/attributes/data.html) and the disam view to identify potential code. But it not a common practice to put code in the ADS. 

The largest amount of data you might expect per record is ~600 bytes or less for resident data (the remaining space left of the MFT entry which total size is 1024 bytes). This cannot be larger for a resident file. I think it will not affect the output in volatility3 😊

ikelos (approved):
Awesome, all looks good, thanks!

PR #1054, opened 2023-12-05
Handling prototype PTEs on Windows x64

thejanit0r:
Added handling of prototype PTEs on Windows x64.

The bug was found while dumping and analyzing user-mode memory regions of `ntdll.dll`.

Info about prototype PTEs:

- https://codemachine.com/articles/prototype_ptes.html
- Windows Internals 7th - P1

### DIFF
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -464,6 +464,27 @@ class WindowsMixin(Intel):
                     entry=excp.entry,
                     swap_offset=swap_offset,
                 )
+            # Handle Prototype PTEs
+            if isinstance(self, WindowsIntel32e):
+                if pbit:           
+                    vaddr = (0xffff << self._maxvirtaddr) | (entry >> 16)
+                    try:
+                        pte_addr = self._translate_swap(self, vaddr, bit_offset)
+                        
+                        # Read and unpack the PTE
+                        entry = self._context.layers.read(self._base_layer, pte_addr[0], 
+                            self._bits_per_register // 8)
+                            
+                        (entry, ) = struct.unpack(self._entry_format, entry)
+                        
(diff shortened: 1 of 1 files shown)


volatility3/framework/layers/intel.py
```diff
@@ -464,6 +464,27 @@ def _translate_swap(
                     entry=excp.entry,
                     swap_offset=swap_offset,
                 )
+            # Handle Prototype PTEs
+            if isinstance(self, WindowsIntel32e):
+                if pbit:           
+                    vaddr = (0xffff << self._maxvirtaddr) | (entry >> 16)
```
ikelos, line 470: Also, are the `0xffff` in order to canonicalize the address?  There's a [separate function](https://github.com/volatilityfoundation/volatility3/blob/a4b927f478eca745a8a4c7ec25c77fa6de49d0df/volatility3/framework/layers/intel.py#L119) for doing that on intel layers...

volatility3/framework/layers/intel.py
```diff
@@ -464,6 +464,27 @@ def _translate_swap(
...
+                    try:
+                        pte_addr = self._translate_swap(self, vaddr, bit_offset)
```
ikelos, line 472: Could this potentially enter an infinite loop given bad data?

ikelos (commented):
Thanks, just a couple of points, but I'd also like to get this checked out by some others to make sure it functions as intended (since it's pretty core code).  I think @iMHLv2 is probably best placed to look at this one...

thejanit0r:
I'll close the PR as I've come across some further issues (e.g. VAD hardware PTE), for which I rewrote some of the code for this PR. Thanks @ikelos for the quick review and feedback. I will follow up with a new PR.

PR #1057, opened 2023-12-08
Linux: Update pslist to fix pstree

eve-mem:
Hello 👋 

When https://github.com/volatilityfoundation/volatility3/pull/899 was merged it added the dump option and removed `_get_task_fields` from the linux pslist plugin. 

This meant that the pstree plugin then stopped working as below:

```
$ python vol.py -f linux-sample-1.dmp linux.pstree
Volatility 3 Framework 2.5.2
Traceback (most recent call last):acking attempts finished
  File "C:\Users\eve\volatility3\vol.py", line 10, in <module>
    volatility3.cli.main()
  File "C:\Users\eve\volatility3\volatility3\cli\__init__.py", line 790, in main
    CommandLine().run()
  File "C:\Users\eve\volatility3\volatility3\cli\__init__.py", line 447, in run
                                      ^^^^^^^^^^^^^^^^^
  File "C:\Users\eve\volatility3\volatility3\framework\plugins\linux\pslist.py", line 205, in run    columns, self._generator(filter_func, include_threads, decorate_comm, dump)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: PsTree._generator() takes from 2 to 4 positional arguments but 5 were given
```

Here I've just tried to add `_get_task_fields` back, and added `_get_file_output` by moving the code the extracts the binaries to it's own function. None of the cleverness there is mine - that is all the hard work from @cstation. This makes the changes to pstree minimal to make it work again. It also makes it look like there are more lines changing here then there their actually are in this PR, it is mostly just shuffling things around. 

I'm not so happy with `_get_task_fields` including the `format_hints.Hex` there rather than at the yield in the _generator. It just made the pstree side a little easier - very happy to work through that though if it doesn't feel correct. 

With the changes pstree works again:
```
$ python vol.py -f linux-sample-1.dmp linux.pstree
Volatility 3 Framework 2.5.2

0x88001f994740  1       1       0       init    Disabled
* 0x88001f085880        322     322     1       udevd   Disabled
** 0x88001c1be080       468     468     322     udevd   Disabled
** 0x88001c3347c0       469     469     322     udevd   Disabled
* 0x88001f665840        1752    1752    1       rpcbind Disabled
* 0x88001bf1d800        1784    1784    1       rpc.statd       Disabled
* 0x88001be1e1c0        1798    1798    1       rpc.idmapd      Disabled
<snip>
```

Thanks again.

### DIFF
--- a/volatility3/framework/plugins/linux/pslist.py
+++ b/volatility3/framework/plugins/linux/pslist.py
@@ -17,7 +17,7 @@ class PsList(interfaces.plugins.PluginInterface):
 
     _required_framework_version = (2, 0, 0)
 
-    _version = (2, 1, 1)
+    _version = (2, 2, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
@@ -78,8 +78,9 @@ class PsList(interfaces.plugins.PluginInterface):
         else:
             return lambda _: False
 
-    def _get_task_fields(
-        self, task: interfaces.objects.ObjectInterface, decorate_comm: bool = False
+    @classmethod
+    def get_task_fields(
+        cls, task: interfaces.objects.ObjectInterface, decorate_comm: bool = False
     ) -> Tuple[int, int, int, str]:
         """Extract the fields needed for the final output
         Args:
@@ -101,7 +102,7 @@ class PsList(interfaces.plugins.PluginInterface):
             elif task.is_user_thread:
                 name = f"{{{name}}}"
 
-        task_fields = (format_hints.Hex(task.vol.offset), pid, tid, ppid, name)
+        task_fields = (task.vol.offset, pid, tid, ppid, name)
         return task_fields
 
     def _get_file_output(self, task: interfaces.objects.ObjectInterface) -> str:
@@ -174,10 +175,10 @@ class PsList(interfaces.plugins.PluginInterface):
             else:
                 file_output = "Disabled"
 
-            offset, pid, tid, ppid, name = self._get_task_fields(task, decorate_comm)
+            offset, pid, tid, ppid, name = self.get_task_fields(task, decorate_comm)
 
             yield 0, (
-                offset,
+                format_hints.Hex(offset),
                 pid,
                 tid,
                 ppid,
--- a/volatility3/framework/plugins/linux/pslist.py
+++ b/volatility3/framework/plugins/linux/pslist.py
@@ -1,7 +1,7 @@
 # This file is Copyright 2021 Volatility Foundation and licensed under the Volatility Software License 1.0
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
-from typing import Any, Callable, Iterable, List
+from typing import Any, Callable, Iterable, List, Tuple
 
 from volatility3.framework import interfaces, renderers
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/linux/pstree.py
```diff
@@ -72,6 +76,12 @@ def yield_processes(pid):
             task = self._tasks[pid]
 
             row = self._get_task_fields(task, decorate_com)
+            if dump:
+                file_output = self._get_file_output(task)
+            else:
+                file_output = "Disabled"
+            row = self._get_task_fields(task, decorate_com)
```
digitalisx: According to CodeQL, `row` needs multiple definitions?

eve-mem: yup! It is pretty silly of me. I'll get it sorted. Thanks.

eve-mem:
Actually i think I'd like to re work this completely so that pstree using pslist as a plugin requirement rather than inheritance, that'll stop something like this again in the future. 

What do you think @ikelos ?

volatility3/framework/plugins/linux/pslist.py
```diff
@@ -104,49 +169,15 @@ def _generator(
...
-                elif task.is_user_thread:
-                    name = f"{{{name}}}"
+                file_output = self._get_file_output(task)
+            else:
+                file_output = "Disabled"
+
+            offset, pid, tid, ppid, name = self._get_task_fields(task, decorate_comm)
 
             yield 0, (
-                format_hints.Hex(task.vol.offset),
```
ikelos, line 149: I think the idea behind this was to make it easier to pass through.  Hex inherits from int, so the only actual impact this has is to change the class so it can be detected later on.  As such it's not that big of a deal, but I'm fine with it being applied later as long as it is applied later (which this patch doesn't appear to do?).

ikelos (commented):
I think the bigger problem here is that `pstree` inherits from `pslist` whereas it might be better for `pstree` to depending on `pslist` using a `Requirement` so that it has to use public interfaces (rather than private ones which may change).  Given this has been pulled out into `_get_task_fields`, it might be better to change this to `get_task_fields`, expose it as a properly parameterized classmethod, and then bump the version numbers appropiately?

eve-mem:
Yes 100% - that's what I meant with my second comment just badly explained. I'll work on that.

eve-mem:
Look like there's a black update that needs to run over the project. I'll do that when i get a moment, that should resolve those checks on this PR.

ikelos:
Ugh, the black checker is installing an alpha version (`24.1a1`), 23.11.0 says everything's fine.  I'm not too keen on using alphas (and requiring all our devs to use alphas), so before you fix anything, lemme figure out if there's a way to prevent alphas being used by the black workflow please....

eve-mem:
@ikelos it looks like you managed to fix those checks now 🙏 thanks!

Syncing my fork was enough to trigger the checks again - and black is now okay with the changes. Thank you for fixing that!

ikelos (approved):
Ok, this looks pretty tidy now, thanks.  5:)

PR #1062, opened 2023-12-14
tolerate duplicate enum values, keep first name that was seen

### LINKED ISSUE #1060: Duplicate Enum value in `bpf_map_type`
Commit [`bpf: Implement cgroup storage available to non-cgroup-attached bpf progs`](https://github.com/torvalds/linux/commit/c4bcfb38a95edb1021a53f2d0356a78120ecfbe4) merged in Linux 6.2-rc1 introduced a duplicate enum value in [`enum bpf_map_type`](https://elixir.bootlin.com/linux/latest/C/ident/bpf_map_type).

Currently Volatility throws an exception in that case, could we maybe handle it more gracefully, e.g., by making the mapping `int -> str | list[str]` or `int -> list[str]` ?

The relevant code seems to be here and it looks like the problem was already anticipated ;)
```python
    @classmethod
    def _generate_inverse_choices(cls, choices: Dict[str, int]) -> Dict[int, str]:
        """Generates the inverse choices for the object."""
        inverse_choices: Dict[int, str] = {}
        for k, v in choices.items():
            if v in inverse_choices:
                # Technically this shouldn't be a problem, but since we inverse cache
                # and can't map one value to two possibilities we throw an exception during build
                # We can remove/work around this if it proves a common issue
                raise ValueError(
                    f"Enumeration value {v} duplicated as {k} and {inverse_choices[v]}"
                )
            inverse_choices[v] = k
        return inverse_choices
```
https://github.com/volatilityfoundation/volatility3/blob/713b5f96dd218eea9c3fdb490a2436a474d22352/volatility3/framework/objects/__init__.py#L599C19-L599C19

vobst:
Proposed solution for https://github.com/volatilityfoundation/volatility3/issues/1060

### DIFF
--- a/volatility3/framework/objects/__init__.py
+++ b/volatility3/framework/objects/__init__.py
@@ -593,17 +593,19 @@ class Enumeration(interfaces.objects.ObjectInterface, int):
         inverse_choices: Dict[int, str] = {}
         for k, v in choices.items():
             if v in inverse_choices:
-                # Technically this shouldn't be a problem, but since we inverse cache
-                # and can't map one value to two possibilities we throw an exception during build
-                # We can remove/work around this if it proves a common issue
-                raise ValueError(
-                    f"Enumeration value {v} duplicated as {k} and {inverse_choices[v]}"
+                vollog.log(
+                    constants.LOGLEVEL_VVV,
+                    f"Enumeration value {v} duplicated as {k}. Keeping name {inverse_choices[v]}",
                 )
+                continue
             inverse_choices[v] = k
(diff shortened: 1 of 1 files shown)


ikelos (approved):
D'oh!  Sorry, I completely missed you submitted this.  Happy to go with yours, but if possible please update the associated `lookup` docstrings as I've done in #1064?

vobst:
added the docstring

ikelos (approved):
Awesome, thank you!  5:)

PR #1069, opened 2023-12-19
Windows: SvcScan Binary Info

dgmcdona:
In the original volatility framework, the svcscan plugin, in addition to gathering information about running services from the services process, also retrieved the ImagePath and ServiceDll attributes from the Windows registry: https://github.com/volatilityfoundation/volatility/blob/a438e768194a9e05eb4d9ee9338b881c0fa25937/volatility/plugins/malware/svcscan.py#L642. 

This PR adds support for that in volatility3. Additionally, during testing, it became clear that the volatility3 svcscan plugin was not working correctly across all versions of Windows. Specifically, `PID` and `Binary` values were not parsing correctly in some instances, instead producing empty values. This was due to changes in the `_SERVICE_RECORD` structure between Windows versions, leading to incorrect offsets for the `ProcessId` and `BinaryPath` members. This PR also adds updated symbol files for the broken versions of Windows, as well as new`OsDistinguisher` instances for correctly detecting those Windows versions where the changes took place. I have tested these new OS versions and symbol files across all relevant versions of Windows to confirm their correctness.

### DIFF
--- a/volatility3/framework/interfaces/context.py
+++ b/volatility3/framework/interfaces/context.py
@@ -87,7 +87,7 @@ class ContextInterface(metaclass=ABCMeta):
         offset: int,
         native_layer_name: str = None,
         **arguments,
-    ):
+    ) -> "interfaces.objects.ObjectInterface":
         """Object factory, takes a context, symbol, offset and optional
         layer_name.
 
--- a/volatility3/framework/plugins/windows/svcscan.py
+++ b/volatility3/framework/plugins/windows/svcscan.py
@@ -4,18 +4,28 @@
 
 import logging
 import os
-from typing import List
+from typing import Dict, List, NamedTuple, Union
 
-from volatility3.framework import interfaces, renderers, constants, symbols, exceptions
+from volatility3.framework import constants, exceptions, interfaces, renderers, symbols
 from volatility3.framework.configuration import requirements
 from volatility3.framework.layers import scanners
 from volatility3.framework.renderers import format_hints
 from volatility3.framework.symbols import intermed
 from volatility3.framework.symbols.windows import versions
 from volatility3.framework.symbols.windows.extensions import services
-from volatility3.plugins.windows import poolscanner, vadyarascan, pslist
+from volatility3.plugins.windows import poolscanner, pslist, vadyarascan
+from volatility3.plugins.windows.registry import hivelist
 
 vollog = logging.getLogger(__name__)
+
+
+ServiceBinaryInfo = NamedTuple(
+    "ServiceBinaryInfo",
+    [
+        ("dll", Union[str, interfaces.renderers.BaseAbsentValue]),
+        ("binary", Union[str, interfaces.renderers.BaseAbsentValue]),
+    ],
+)
 
 
 class SvcScan(interfaces.plugins.PluginInterface):
@@ -42,10 +52,16 @@ class SvcScan(interfaces.plugins.PluginInterface):
             requirements.PluginRequirement(
                 name="vadyarascan", plugin=vadyarascan.VadYaraScan, version=(1, 0, 0)
             ),
+            requirements.PluginRequirement(
+                name="hivelist", plugin=hivelist.HiveList, version=(1, 0, 0)
+            ),
         ]
 
     @staticmethod
-    def get_record_tuple(service_record: interfaces.objects.ObjectInterface):
+    def get_record_tuple(
+        service_record: interfaces.objects.ObjectInterface,
+        binary_info: ServiceBinaryInfo,
+    ):
         return (
             format_hints.Hex(service_record.vol.offset),
             service_record.Order,
@@ -56,6 +72,8 @@ class SvcScan(interfaces.plugins.PluginInterface):
             service_record.get_name(),
             service_record.get_display(),
             service_record.get_binary(),
+            binary_info.binary,
+            binary_info.dll,
         )
 
     @staticmethod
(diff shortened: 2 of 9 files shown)


volatility3/framework/interfaces/context.py
```diff
@@ -87,7 +87,7 @@ def object(
         offset: int,
         native_layer_name: str = None,
         **arguments,
-    ):
+    ) -> "interfaces.objects.ObjectInterface":
```
ikelos, line 90: Nice catch, thanks!  5:D

volatility3/framework/plugins/windows/svcscan.py
```diff
@@ -42,10 +52,16 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
                 name="vadyarascan", plugin=vadyarascan.VadYaraScan, version=(1, 0, 0)
             ),
+            requirements.PluginRequirement(
+                name="hivelist", plugin=hivelist.HiveList, version=(1, 0, 0)
+            ),
         ]
 
     @staticmethod
-    def get_record_tuple(service_record: interfaces.objects.ObjectInterface):
+    def get_record_tuple(
```
ikelos, line 68: This is technically a change in the API, and since binary info is now a required parameter, this is an additive change (so the major plugin number would need a bump, ie `_version = (2, 0, 0)`) and then check any plugins in core that might have depended on it to change their requirements too.  

The other option would be to make `binary_info` have a default value (of `None`, for instance) and then find a way to handle that in the output (although since the return tuple has changed size too, this probably isn't a good plan.

volatility3/framework/plugins/windows/svcscan.py
```diff
@@ -89,25 +107,45 @@ def create_service_table(
         ):
             symbol_filename = "services-xp-2003-x64"
         elif (
-            versions.is_win10_16299_or_later(context=context, symbol_table=symbol_table)
+            versions.is_win10_25398_or_later(context=context, symbol_table=symbol_table)
```
ikelos: This pattern looks repeated.  Perhaps it could be done with a list of tuples  (like `(versions.is_win_blah, "service-win10-blah")?  Massive if statements feel inefficient and more difficult to add new entries to (more difficult than is necessary).

volatility3/framework/plugins/windows/svcscan.py
```diff
@@ -118,6 +156,16 @@ def create_service_table(
             and not is_64bit
         ):
             symbol_filename = "services-win10-15063-x86"
+        elif (
+            versions.is_win10_up_to_15063(context=context, symbol_table=symbol_table)
+            and is_64bit
```
ikelos: Probably need a flag in the tuple to check whether `is_64bit` is necessary too...

volatility3/framework/plugins/windows/svcscan.py
```diff
@@ -150,13 +198,102 @@ def create_service_table(
             native_types=native_types,
         )
 
+    def _get_service_key(self, kernel):
+        for hive in hivelist.HiveList.list_hives(
+            context=self.context,
+            base_config_path=self.config_path,
```
ikelos: This is probably fine, I don't think the config spaces should matter, but I *think* this is supposed to be unique.  I think something like `interfaces.configuration.path_join(self.config_path, "hivelist")` would be fine...

volatility3/framework/plugins/windows/svcscan.py
```diff
@@ -150,13 +198,102 @@ def create_service_table(
...
+            hive_offsets=None,
+        ):
+            # Get ControlSet\Services.
+            try:
+                return hive.get_key(r"CurrentControlSet\Services")
+            except (KeyError, exceptions.InvalidAddressException):
+                try:
+                    return hive.get_key(r"ControlSet001\Services")
+                except (KeyError, exceptions.InvalidAddressException):
+                    pass
```
ikelos: Probably want a debug warning (`constants.DEBUG_VVVV` I think) to let developers know *something* bad happened, just as an extra to returning `None` so that if the result is ignored, developers could still figure out why something that should've happened didn't...

volatility3/framework/plugins/windows/svcscan.py
```diff
@@ -150,13 +198,102 @@ def create_service_table(
...
+
+            return None
+
+    @staticmethod
+    def _get_service_dll(
+        service_key,
+    ) -> Union[str, interfaces.renderers.BaseAbsentValue]:
+        try:
+            param_key = next(
```
ikelos, line 180: Quite an unusual syntax for doing what you're doing but... it works.  No real issue, just interesting to see it in use.  5:)

ikelos (commented):
Thanks, this looks pretty good.  Since the `get_record_tuple` was officially part of the plugin's API there's a little bit of house keeping to ensure backwards compatibility, but the rest of it looks good.  I'd appreciate @iMHLv2 having a look, but if he's busy and doesn't get to it in a week or so, gimme a reminder and I'll merge it in once the housekeeping's been done.  5:)

volatility3/framework/plugins/windows/svcscan.py
dgmcdona: Agreed, this part felt very clunky when I was working on it. It should also be more clear in the future that the order in which the checks are executed is essential as well, I'll add a comment to that effect.

ikelos: If the order's dependent, does that mean one of the checks doesn't actually check what it says it will?  If a function is called `is_win10_25398_or_later` then it should reliably return that, or be augumented with whatever additional checks are necessary to achieve that?  Hopefully I haven't misunderstood, if you're just going through the checks find out the lowest version that passes, that's fine and obvious so everything working as intended...  5:)

volatility3/framework/plugins/windows/svcscan.py
```diff
@@ -150,13 +198,102 @@
             native_types=native_types,
         )
 
+    def _get_service_key(self, kernel):
```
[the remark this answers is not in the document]
ikelos: Could you please fix this up at the end too?  Ideally with some type information in the signature (using `Optional` if you return `None`) and then an explicit return None at the end, if that's a valid response.

volatility3/framework/plugins/windows/svcscan.py
dgmcdona: Apologies if I caused any confusion here. If my understanding is correct, I think you're correct in your interpretation - we have to fall through the checks in order from the newest OS version to the oldest. I just wanted to make it explicit that you can't reorder this list of tuples, otherwise you will likely select a version of the symbols that is too old for the sample that you're analyzing.

dgmcdona:
Thanks for the feedback! I think I have resolved all of the issues that you pointed out. I also retested across many samples to ensure that these fixes didn't cause any breaking changes.

ikelos (commented):
Looks good, thanks for making the requested changes.  5:)

ikelos:
Please can you make sure the changes have been formatted using black, and then I'm happy to get this merged (the failed check will contain a diff if you don't want to install black yourself for some reason)...

dgmcdona:
Ok, I rebased onto the latest `develop` branch, and reformatted using `black` version 24.1.1; I think it should be good to go now.

volatility3/framework/plugins/windows/svcscan.py
```diff
@@ -150,13 +144,111 @@ def create_service_table(
...
+                        objects.StructType, hive.get_key(r"ControlSet001\Services")
+                    )
+                except (KeyError, exceptions.InvalidAddressException):
+                    vollog.log(
+                        constants.LOGLEVEL_VVVV,
+                        "Could not retrieve any control set from SYSTEM hive",
+                    )
+                    pass
+
+            return None
```
ikelos: I think this needs de-indenting one level based on the CodeQL notice at the top, and the pass statement is unecessary.  The rest of it looks good though, fix those up and once the tests have passed we can get it merged, thanks!

ikelos (commented):
Thanks very much, looking really good now, just a couple trivial bits to fix up and then we can merge it!  5:D

volatility3/framework/plugins/windows/svcscan.py
dgmcdona: Done 👍

ikelos:
Looks good, all merged, thanks!

PR #1071, opened 2023-12-21
Windows: Added '--refined' option to windows malfind plugin

hsarkey:
The windows malfind plugin lists process memory ranges that potentially contain injected code. In the original volatility framework, the windows malfind plugin offered a "W" command line option that refined the results to only high confidence matches. The criteria for a high confidence match was either memory regions with an MZ header or that started with a well known opcode combination, ie. PUSH EBP.

This PR adds support for a "--refined" option for volatility3 to match the functionality of the "W" option from the original framework. For this PR, the possible criteria for a high confidence match was updated to include the common opcodes for 32 and 64bit functions ("\x55\x48" and "\x55\x89"),  in addition to the original criteria of the MZ header or the PUSH EBP opcode ("\x55\x8B") from volatility.  I have tested against various memory samples with injected code to validate the functionality of this option.

### DIFF
--- a/volatility3/framework/plugins/windows/malfind.py
+++ b/volatility3/framework/plugins/windows/malfind.py
@@ -46,12 +46,6 @@ class Malfind(interfaces.plugins.PluginInterface):
             requirements.VersionRequirement(
                 name="vadinfo", component=vadinfo.VadInfo, version=(2, 0, 0)
             ),
-            requirements.BooleanRequirement(
-                name="refined",
-                description="Refine the output. Only show regions with an MZ header or that start with well known opcode combinations (i.e. PUSH EBP). WARNING: This can cause you to overlook regions with wiped headers or shell code blocks starting with NOP sleds, etc.. However, it will in general result in less noisy output.",
-                default=False,
-                optional=True,
-            ),
         ]
 
     @classmethod
@@ -144,9 +138,6 @@ class Malfind(interfaces.plugins.PluginInterface):
                 yield vad, data
 
     def _generator(self, procs):
-        # set refined criteria
-        refined_criteria = [b"MZ", b"\x55\x8B", b"\x55\x48", b"\x55\x89"]
-
         # determine if we're on a 32 or 64 bit kernel
         kernel = self.context.modules[self.config["kernel"]]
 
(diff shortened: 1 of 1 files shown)


ikelos:
Thanks, this looks good I'm just wondering if `refined` is descriptive enough?  I'm not sure what would be better, `--high-suspicion` or `--likely` or something along those lines?  Something a bit more descriptive about what the flag actually does...

ikelos:
Ok, so we had a bit of a think about this and in general I'm not keen on flags that alter the output of the plugin because that's technically making the plugin decide what's really a user interface decision (do I display this data or not).  So in this instance, it might be better to add a permanent column that either lists attributes (like "Common opcodes" or "MZ header" or "PUSH ebp") or a score indicating a change of malware (probably a numeric value to allow for future expansion and tweaking, even if at the moment it only ever returns 0 and 100).  We don't yet have filtering capabilities for the CLI, but this may also encourage people to build a better UI for processing the results of volatility.  Returning a list of attributes might get messy on the CLI and be difficult to parse for another program, but might be easier to see from a human readable perspective.  So either of those work, whichever you think works better?

hsarkey:
After thinking about this, I updated to no longer have the "--refined" command line option and instead added a permanent "Notes" column. This column will display "None" unless a process has one of the common header types discussed above, in which case the "Notes" column will display "MZ header", "PE header", or "Function prologue" depending on what criteria are met.

volatility3/framework/plugins/windows/malfind.py
```diff
@@ -141,16 +141,30 @@ def _generator(self, procs):
...
+        # set refined criteria to know when to add to "Notes" column
+        refined_criteria = [b"MZ", b"\x55\x8B", b"\x55\x48", b"\x55\x89"]
+
         is_32bit_arch = not symbols.symbol_table_is_64bit(
             self.context, kernel.symbol_table_name
         )
 
         for proc in procs:
+            # by default, "Notes" column will be set to none
+            notes = "None"
```
ikelos: Rather than a textual value, it may be worthwhile making this something derived from `BaseAbsentValue`, which might allow a future interface to filter our non-responses more easily?

volatility3/framework/plugins/windows/malfind.py
```diff
@@ -141,16 +141,30 @@ def _generator(self, procs):
...
             process_name = utility.array_to_string(proc.ImageFileName)
 
             for vad, data in self.list_injections(
                 self.context, kernel.layer_name, kernel.symbol_table_name, proc
             ):
+                # Check for unique headers and update "Notes" column if criteria is met
+                if data[0:2] in refined_criteria:
```
ikelos, line 165: This list needs keeping in sync with `refined_criteria`, it might be better to make `refined_criteria` a dictionary that listen an initial byte pattern with a value for notes?  It would make this code a little more generic and it would be easier to add future classifiers if needed...

ikelos (commented):
Hiya, I'm much happier with this change, but I think there's a couple bits that could still be tweaked.  Please let me know what you think and once you're happy with any changes you make and we can get it merged.  5:)

hsarkey:
I took your advice and updated the "refined_criteria" list to be a dictionary, so in case it grows larger in the future, it will easier to maintain. Also, instead of "None", I updated to use the "NotApplicableValue" from the renderers package, which has the BaseAbsentValue you mentioned above.

volatility3/framework/plugins/windows/malfind.py
ikelos: It might be worth using `startswith` or something, rather than limiting all future things to exactly 2 bytes, but that's something we can do fairly easily if we decide to extend the dictionary in the future, so not a show stopper.

ikelos (commented):
Yep, looks good now.  Just thinking about the specific `data[0:2]` bit, but not enough to block this.  Just let me know if you want to change it before I merge or if you're happier leaving it as is.

hsarkey:
Hi, I am happy with keeping the data[0:2] for now.  If significant changes are made to the dictionary, this can be easily modified, but I'm not sure when/if that would even happen. Thanks for your help on this!

PR #1076, opened 2024-01-02
Windows: Add additional command line argument info to windows pstree

ckusek15:
In the original volatility framework, the `windows.pstree` plugin offers a  '--verbose' option that displays additional command line argument info of processes in the tree. The fields that are added when this option is used are `audit`, `cmd`, and `path`. `audit` refers to the ImageFileName in the SeAuditProcessCreationInfo struct of a process; and `cmd` and `path` refer to the CommandLine and ImagePathName strings, respectively, in the process parameter information.

This PR adds these additional fields to the volatility3 `windows.pstree` plugin TreeGrid to match the output expected when using the '--verbose' option of the original framework. Manual testing was performed on multiple windows memory images to verify the expected behavior.

### DIFF
--- a/volatility3/framework/plugins/windows/pstree.py
+++ b/volatility3/framework/plugins/windows/pstree.py
@@ -5,7 +5,7 @@ import datetime
 import logging
 from typing import Callable, Dict, Set, Tuple
 
-from volatility3.framework import objects, interfaces, renderers
+from volatility3.framework import objects, interfaces, renderers, exceptions
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints
 from volatility3.plugins.windows import pslist
@@ -132,6 +132,25 @@ class PsTree(interfaces.plugins.PluginInterface):
                 proc.get_exit_time(),
             )
 
+            try:
+                audit = proc.SeAuditProcessCreationInfo.ImageFileName.Name
+                # If 'audit' is set to the empty string, display NotAvailableValue
+                row += (audit.get_string() or renderers.NotAvailableValue(),)
+            except exceptions.InvalidAddressException:
+                row += (renderers.NotAvailableValue(),)
+
+            try:
+                process_params = proc.get_peb().ProcessParameters
+                row += (
+                    process_params.CommandLine.get_string(),
+                    process_params.ImagePathName.get_string(),
+                )
+            except exceptions.InvalidAddressException:
+                row += (
+                    renderers.NotAvailableValue(),
+                    renderers.NotAvailableValue(),
+                )
+
             yield (self._levels[pid] - 1, row)
             for child_pid in self._children.get(pid, []):
                 yield from yield_processes(
@@ -161,6 +180,9 @@ class PsTree(interfaces.plugins.PluginInterface):
                 ("Wow64", bool),
                 ("CreateTime", datetime.datetime),
                 ("ExitTime", datetime.datetime),
+                ("Audit", str),
+                ("Cmd", str),
+                ("Path", str),
             ],
             self._generator(
                 filter_func=pslist.PsList.create_pid_filter(


volatility3/framework/plugins/windows/pstree.py
```diff
@@ -49,6 +49,12 @@ def get_requirements(cls):
                 description="Process ID to include (with ancestors and descendants, all other processes are excluded)",
                 optional=True,
             ),
+            requirements.BooleanRequirement(
+                name="verbose",
```
ikelos: `verbose` is a little confusing because the main tool already has a `-v` and we have yet to have a flag that doubles in the main tool and a plugin, however it does match with the previous version of volatility and it would need to be explicitly spelled out (`--verbose`).  Lemme know if you think `--details` would work, but it's not a big concern and certainly not a show stopper...

volatility3/framework/plugins/windows/pstree.py
```diff
@@ -149,19 +172,24 @@ def run(self):
...
+            (f"Offset{offsettype}", format_hints.Hex),
+            ("Threads", int),
+            ("Handles", int),
+            ("SessionId", int),
+            ("Wow64", bool),
+            ("CreateTime", datetime.datetime),
+            ("ExitTime", datetime.datetime),
+        ]
+
+        if self.config["verbose"]:
```
ikelos: I think there are other plugins where we return a different TreeGrid (ie, different result columns) based on a flag, but I'm a little bit nervous about it, it'll make it slightly harder for frameworks to know what to expect.  Again, not a show stopper, but given it doesn't take much more to find the additional values, perhaps we should just always include them from the get-go?

volatility3/framework/plugins/windows/pstree.py
```diff
@@ -132,6 +138,23 @@ def yield_processes(pid, descendant: bool = False):
                 proc.get_exit_time(),
             )
 
+            if self.config["verbose"]:
+                row += (
+                    proc.SeAuditProcessCreationInfo.ImageFileName.Name.get_string()
```
ikelos: What happens if this throws an InvalidAddressExcption when try to access this?  At the moment everything will stop, would it be better to catch it as just below?

ikelos (commented):
Looks good, just wondering about catching exceptions when accessing the SeAuditProcessCreationInfo?  A couple of other stylistic suggestions too, but they're not strong suggestions, just let me know which way you'd prefer and I'll go with that...

volatility3/framework/plugins/windows/pstree.py
ckusek15: Added a catch for SeAuditProcessCreationInfo throwing an exception with [97ea676](https://github.com/volatilityfoundation/volatility3/pull/1076/commits/97ea6765ea57df54e3a6721c86b00c6ebe7084d9). There is also a case where `SeAuditProcessCreationInfo.ImageFileName.Name` is able to be accessed, but returns the empty string. I kept the logic for checking this case and added a comment

volatility3/framework/plugins/windows/pstree.py
npetroni: @ikelos does adding columns to a TreeGrid for an existing plugin require bumping the plugin interface version? Could clients be affected by receiving unexpected columns without protecting it with a flag?

I agree, making this the default could be a good approach, I just wanted to clarify the ramifications of such a change.

ikelos: To this point, the version number has only been about the internally accessible API (so methods within the class).  We haven't ever told frameworks that plugin output will be fixed for a particular framework (although I know at least one which has written specific handlers for the output of each plugin).  The column names are programatically accessible, so as long as this is additive it should be fine.  We might also want to get in the habit of getting people to bump the patchj version of a plugin any time there is a change (as we should have been doing, but have generally ignored).

We probably need to make a policy call on this at some point, but for now I'm happy with bump or no bump.  If the version number is to be bumped, I'd only do the minor version, because a major version would require all plugins depending on it to have their dependencies updated.

ckusek15: Decided to go the route of always including the additional rows in the TreeGrid. Updated with [5fb7178](https://github.com/volatilityfoundation/volatility3/pull/1076/commits/5fb7178811248af97e3569f72c42b847f6069ff5)

volatility3/framework/plugins/windows/pstree.py
ckusek15: Flag no longer needed. See https://github.com/volatilityfoundation/volatility3/pull/1076#discussion_r1449103902

ckusek15:
Squashed commits with https://github.com/volatilityfoundation/volatility3/pull/1076/commits/a50ebb6014ff11ea165454e0a08b967d01cce9e8

ikelos (approved):
Looks good, thanks very much!

PR #1080, opened 2024-01-10
Linux netfilter hooks plugin

gcmoreira:
It supports the following protocols: INET, IPv4, IPv6, ARP, NETDEV (ingress and egress hooks), BRIDGE and DECNET.

Supported Netfilter hooks implementations:
- kernels < 4.3 (Tested on kernel 3.11.0-26)
- 4.3 <= kernels < 4.9 (Tested on kernel 4.4.0-210)
- 4.9 <= kernels < 4.14 (Tested on kernel 4.13.0-46)
- 4.14 <= kernels < 4.16 (Tested on kernel 4.15.0-169)
- kernels >= 4.16 (Tested on kernel 4.18.0-10 and 5.19.0-50)

Supported NetDev ingress hooks implementation
- 4.2 <= kernels < 4.9  (Tested on kernel 4.4.0-210)
- 4.9 <= kernels < 4.14  (Tested on kernel 4.13.0-46)
- kernels >= 4.14  (Tested on kernel 4.18.0-10)

Supported NetDev egress hooks implementation
- kernels >= 5.16  (Tested on kernel 5.19.0-50)

Example:
```shell
$ python3 ./vol.py -r pretty \
  -f ram-5.19.0-50 \
  linux.netfilter.Netfilter
Volatility 3 Framework 2.5.2
  |     Net NS |  Proto |         Hook |   Priority |        Handler |                  Module | Is Hooked
* | 4026531840 |   IPV4 |  PRE_ROUTING |       -400 | 0xffffc05b5160 |          nf_defrag_ipv4 |      True
* | 4026531840 |   IPV4 |  PRE_ROUTING |       -200 | 0xffffc062a510 |            nf_conntrack |      True
* | 4026531840 |   IPV4 |     LOCAL_IN |          0 | 0xffffc05e7c60 |               nf_tables |      True
* | 4026531840 |   IPV4 |     LOCAL_IN |          0 | 0xffffc05e74a0 |               nf_tables |      True
* | 4026531840 |   IPV4 |     LOCAL_IN | 2147483647 | 0xffffc062b040 |            nf_conntrack |      True
* | 4026531840 |   IPV4 |      FORWARD |          0 | 0xffffc05e7c60 |               nf_tables |      True
* | 4026531840 |   IPV4 |      FORWARD |          0 | 0xffffc05e74a0 |               nf_tables |      True
* | 4026531840 |   IPV4 |    LOCAL_OUT |       -400 | 0xffffc05b5160 |          nf_defrag_ipv4 |      True
* | 4026531840 |   IPV4 |    LOCAL_OUT |       -200 | 0xffffc062ae60 |            nf_conntrack |      True
* | 4026531840 |   IPV4 |    LOCAL_OUT |          0 | 0xffffc05e7c60 |               nf_tables |      True
* | 4026531840 |   IPV4 |    LOCAL_OUT |          0 | 0xffffc05e74a0 |               nf_tables |      True
* | 4026531840 |   IPV4 | POST_ROUTING |       -225 | 0xffffa681c9c0 | [apparmor_ip_postroute] |      True
* | 4026531840 |   IPV4 | POST_ROUTING | 2147483647 | 0xffffc062b040 |            nf_conntrack |      True
* | 4026531840 |   IPV6 |  PRE_ROUTING |       -400 | 0xffffc061b160 |          nf_defrag_ipv6 |      True
* | 4026531840 |   IPV6 |  PRE_ROUTING |       -200 | 0xffffc062aa40 |            nf_conntrack |      True
* | 4026531840 |   IPV6 |     LOCAL_IN |          0 | 0xffffc05e7cf0 |               nf_tables |      True
* | 4026531840 |   IPV6 |     LOCAL_IN |          0 | 0xffffc05e74a0 |               nf_tables |      True
* | 4026531840 |   IPV6 |     LOCAL_IN | 2147483646 | 0xffffc062b130 |            nf_conntrack |      True
* | 4026531840 |   IPV6 |      FORWARD |          0 | 0xffffc05e7cf0 |               nf_tables |      True
* | 4026531840 |   IPV6 |      FORWARD |          0 | 0xffffc05e74a0 |               nf_tables |      True
* | 4026531840 |   IPV6 |    LOCAL_OUT |       -400 | 0xffffc061b160 |          nf_defrag_ipv6 |      True
* | 4026531840 |   IPV6 |    LOCAL_OUT |       -200 | 0xffffc062aa20 |            nf_conntrack |      True
* | 4026531840 |   IPV6 |    LOCAL_OUT |          0 | 0xffffc05e7cf0 |               nf_tables |      True
* | 4026531840 |   IPV6 |    LOCAL_OUT |          0 | 0xffffc05e74a0 |               nf_tables |      True
* | 4026531840 |   IPV6 | POST_ROUTING |       -225 | 0xffffa681c9c0 | [apparmor_ip_postroute] |      True
* | 4026531840 |   IPV6 | POST_ROUTING | 2147483647 | 0xffffc062b130 |            nf_conntrack |      True
* | 4026531840 | NETDEV |      INGRESS |          0 | 0xffffc05e7a60 |               nf_tables |      True
* | 4026531840 | NETDEV |       EGRESS |          0 | 0xffffc05e7a60 |               nf_tables |      True
```

See the full test case suite output:
[vol3_linux_netfilter_output.txt](https://github.com/volatilityfoundation/volatility3/files/13892877/vol3_linux_netfilter_output.txt)

### DIFF
--- a/volatility3/framework/plugins/linux/netfilter.py
+++ b/volatility3/framework/plugins/linux/netfilter.py
@@ -69,13 +69,9 @@ class AbstractNetfilter(ABC):
     NF_MAX_HOOKS = LARGEST_HOOK_NUMBER + 1
 
     def __init__(
-        self,
-        context: interfaces.context.ContextInterface,
-        config: interfaces.configuration.HierarchicalDict,
+        self, context: interfaces.context.ContextInterface, kernel_module_name: str
     ):
         self._context = context
-        self._config = config
-        kernel_module_name = self._config["kernel"]
         self.vmlinux = context.modules[kernel_module_name]
         self.layer_name = self.vmlinux.layer_name
 
@@ -106,7 +102,8 @@ class AbstractNetfilter(ABC):
         Yields:
             The kmsg records. Same as _run()
         """
-        vmlinux = context.modules[config["kernel"]]
+        kernel_module_name = config["kernel"]
+        vmlinux = context.modules[kernel_module_name]
 
         implementation_inst = None  # type: ignore
         for subclass in class_subclasses(cls):
@@ -123,7 +120,9 @@ class AbstractNetfilter(ABC):
                 "Netfilter implementation '%s' matches!",
                 subclass.__name__,
             )
-            implementation_inst = subclass(context=context, config=config)
+            implementation_inst = subclass(
+                context=context, kernel_module_name=kernel_module_name
+            )
             # More than one class could be executed for an specific kernel version
             # For instance: Netfilter Ingress hooks
             yield from implementation_inst._run()
--- a/volatility3/framework/plugins/linux/netfilter.py
+++ b/volatility3/framework/plugins/linux/netfilter.py
@@ -75,13 +75,13 @@ class AbstractNetfilter(ABC):
     ):
         self._context = context
         self._config = config
-        symbol_table = self._config["kernel"]
-        self.vmlinux = context.modules[symbol_table]
+        kernel_module_name = self._config["kernel"]
+        self.vmlinux = context.modules[kernel_module_name]
         self.layer_name = self.vmlinux.layer_name
 
-        modules = lsmod.Lsmod.list_modules(context, symbol_table)
+        modules = lsmod.Lsmod.list_modules(context, kernel_module_name)
         self.handlers = linux.LinuxUtilities.generate_kernel_handler_info(
-            context, symbol_table, modules
+            context, kernel_module_name, modules
         )
 
         self._set_data_sizes()
--- a/volatility3/framework/plugins/linux/netfilter.py
+++ b/volatility3/framework/plugins/linux/netfilter.py
@@ -136,11 +136,13 @@ class AbstractNetfilter(ABC):
         allow  customization of the code to the specific data structure used in a
         particular kernel implementation
 
-            get_hooks_container_by_protocol(net, proto_name)
+            get_hooks_container(net, proto_name, hook_name)
                 It returns the data structure used in a specific kernel implementation
                 to store the hooks for a respective namespace and protocol, basically:
                     For Ingress hooks:
                         network_namespace[] -> net_device[] -> nf_hooks_ingress[]
+                    For egress hooks:
+                        network_namespace[] -> net_device[] -> nf_hooks_egress[]
                     For all the other Netfilter hooks:
                         <= 4.2.8
                             nf_hooks[]
@@ -148,7 +150,7 @@ class AbstractNetfilter(ABC):
                             network_namespace[] -> nf.hooks[]
 
             get_hook_ops(hook_container, proto_idx, hook_idx)
-                Give the 'hook_container' got in get_hooks_container_by_protocol(), it
+                Give the 'hook_container' got in get_hooks_container(), it
                 returns an iterable of 'nf_hook_ops' elements for a respective protocol
                 and hook type.
 
@@ -163,7 +165,7 @@ class AbstractNetfilter(ABC):
         """
         for netns, net in self.get_net_namespaces():
             for proto_idx, proto_name, hook_idx, hook_name in self._proto_hook_loop():
-                hooks_container = self.get_hooks_container_by_protocol(net, proto_name)
+                hooks_container = self.get_hooks_container(net, proto_name, hook_name)
 
                 for hook_container in hooks_container:
                     for hook_ops in self.get_hook_ops(
@@ -270,7 +272,7 @@ class AbstractNetfilter(ABC):
             net_ns_id = net.ns.inum
             yield net_ns_id, net
 
-    def get_hooks_container_by_protocol(self, net, proto_name):
+    def get_hooks_container(self, net, proto_name, hook_name):
         """Returns the data structure used in a specific kernel implementation to store
         the hooks for a respective namespace and protocol.
 
@@ -281,7 +283,7 @@ class AbstractNetfilter(ABC):
         yield net.nf.hooks
 
     def get_hook_ops(self, hook_container, proto_idx, hook_idx):
-        """Given the hook_container obtained from get_hooks_container_by_protocol(), it
+        """Given the hook_container obtained from get_hooks_container(), it
         returns an iterable of 'nf_hook_ops' elements for a corresponding protocol
         and hook type.
 
@@ -338,7 +340,7 @@ class NetfilterImp_to_4_3(AbstractNetfilter):
         netns, net = renderers.NotAvailableValue(), renderers.NotAvailableValue()
         yield netns, net
 
-    def get_hooks_container_by_protocol(self, net, proto_name):
+    def get_hooks_container(self, net, proto_name, hook_name):
         nf_hooks = self.vmlinux.object_from_symbol("nf_hooks")
         if not nf_hooks:
             return
@@ -508,7 +510,7 @@ class NetfilterImp_4_16_to_latest(NetfilterImp_4_14_to_4_16):
             and vmlinux.get_type("netns_nf").has_member("hooks_ipv4")
         )
 
-    def get_hooks_container_by_protocol(self, net, proto_name):
+    def get_hooks_container(self, net, proto_name, hook_name):
         try:
             if proto_name == "IPV4":
                 net_nf_hooks = net.nf.hooks_ipv4
@@ -556,20 +558,19 @@ class AbstractNetfilterNetDev(AbstractNetfilter):
     def subscribed_protocols(self):
         return ("NETDEV",)
 
-    def get_hooks_container_by_protocol(self, net, proto_name):
-        if proto_name != "NETDEV":
-            return
-
+    def get_hooks_container(self, net, proto_name, hook_name):
         net_device_type = self.vmlinux.get_type("net_device")
         net_device_name = self.get_symbol_fullname("net_device")
         for net_device in net.dev_base_head.to_list(net_device_name, "dev_list"):
-            if net_device_type.has_member("nf_hooks_ingress"):
-                # CONFIG_NETFILTER_INGRESS=y
-                yield net_device.nf_hooks_ingress
+            if hook_name == "INGRESS":
+                if net_device_type.has_member("nf_hooks_ingress"):
+                    # CONFIG_NETFILTER_INGRESS=y
+                    yield net_device.nf_hooks_ingress
 
-            if net_device_type.has_member("nf_hooks_egress"):
-                # CONFIG_NETFILTER_EGRESS=y
-                yield net_device.nf_hooks_egress
+            elif hook_name == "EGRESS":
+                if net_device_type.has_member("nf_hooks_egress"):
+                    # CONFIG_NETFILTER_EGRESS=y
+                    yield net_device.nf_hooks_egress
 
 
 class NetfilterNetDevImp_4_2_to_4_9(AbstractNetfilterNetDev):
--- a/volatility3/framework/plugins/linux/netfilter.py
+++ b/volatility3/framework/plugins/linux/netfilter.py
@@ -572,7 +572,7 @@ class AbstractNetfilterNetDev(AbstractNetfilter):
                 yield net_device.nf_hooks_egress
 
 
-class NetfilterIngressImp_4_2_to_4_9(AbstractNetfilterNetDev):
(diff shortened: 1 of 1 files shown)


gcmoreira:
Hey @ikelos/@atcuno this is still awaiting a review :pray:

atcuno:
@ikelos this one has been tested quite a bit and is good from my end. It is a powerful plugin for rootkit detection.

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -0,0 +1,716 @@
...
+    NF_MAX_HOOKS = LARGEST_HOOK_NUMBER + 1
+
+    def __init__(
+        self,
+        context: interfaces.context.ContextInterface,
+        config: interfaces.configuration.HierarchicalDict,
+    ):
+        self._context = context
+        self._config = config
+        symbol_table = self._config["kernel"]
```
ikelos: `symbol_table` is a bit of an odd name, but not a show stopper.  5;P

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -0,0 +1,716 @@
...
+
+    def _set_data_sizes(self):
+        self.ptr_size = self.vmlinux.get_type("pointer").size
+        self.list_head_size = self.vmlinux.get_type("list_head").size
+
+    @classmethod
+    def run_all(
+        cls,
+        context: interfaces.context.ContextInterface,
+        config: interfaces.configuration.HierarchicalDict,
```
ikelos: It looks like `config['kernel']` is the only time this is ever used? I'd be much happier passing through a module_name than a whole hierarchical dict? Was the expectation this might need extending or is there another good reason to pass config through?

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -0,0 +1,716 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+from dataclasses import dataclass, field
+from abc import ABC, abstractmethod
+import logging
+
+from typing import Iterator, List, Tuple
+from volatility3.framework import (
+    class_subclasses,
```
ikelos: This is a function, not a module.  It would be better (and is somewhat the coding standard in volatility) to `import volatility.framework` and then call `framework.class_subclasses` to avoid people accidentally thinking it's defined here and then importing it from here.

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -0,0 +1,716 @@
...
+        config: interfaces.configuration.HierarchicalDict,
+    ):
+        self._context = context
+        self._config = config
+        symbol_table = self._config["kernel"]
+        self.vmlinux = context.modules[symbol_table]
+        self.layer_name = self.vmlinux.layer_name
+
+        modules = lsmod.Lsmod.list_modules(context, symbol_table)
+        self.handlers = linux.LinuxUtilities.generate_kernel_handler_info(
```
ikelos: I believe `LinuxUtilities` is also versioned, so should be one of the plugin's requirements, a `VersionRequirement` please (just to make sure the API won't change out from under you).

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -0,0 +1,716 @@
...
+            context, symbol_table, modules
+        )
+
+        self._set_data_sizes()
+
+    def _set_data_sizes(self):
+        self.ptr_size = self.vmlinux.get_type("pointer").size
```
ikelos, line 80: This isn't a show stopper because this gets called as part of the `__init__` but in general all class variables should be defined only and directly in the initializer, to avoid accesses before they've been set.

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -0,0 +1,716 @@
...
+            implementation_inst = subclass(context=context, config=config)
+            # More than one class could be executed for an specific kernel version
+            # For instance: Netfilter Ingress hooks
+            yield from implementation_inst._run()
+
+        if implementation_inst is None:
+            vollog.error("Unsupported Netfilter kernel implementation")
+
+    def _run(self) -> Iterator[Tuple[int, str, str, int, int, str, bool]]:
+        """Iterates over namespaces and protocols, executing various callbacks that
```
ikelos, line 148: Really good docstrings!  Thanks!  5:D

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -0,0 +1,716 @@
...
+        Netfiler subclass.
+        """
+
+        # Most implementation handlers respond to these protocols, except for
+        # the ingress hook, which specifically handles the 'NETDEV' protocol.
+        # However, there is no corresponding Netfilter hook implementation for
+        # the INET protocol in the kernel. AFAIU, this is used as
+        # 'NFPROTO_INET = NFPROTO_IPV4 || NFPROTO_IPV6'
+        # in other parts of the kernel source code.
+        return ("IPV4", "ARP", "BRIDGE", "IPV6", "DECNET")
```
ikelos, line 260: Feels like this should be defined as a constant at the top of the module?  Any reason this list is different?

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -0,0 +1,716 @@
...
+        symbol_net_name = self.get_symbol_fullname("net")
+        for net in nethead.to_list(symbol_net_name, "list"):
+            net_ns_id = net.ns.inum
+            yield net_ns_id, net
+
+    def get_hooks_container(self, net, proto_name, hook_name):
+        """Returns the data structure used in a specific kernel implementation to store
+        the hooks for a respective namespace and protocol.
+
+        Except for kernels < 4.3, all the implementations use network namespaces.
```
ikelos, line 292: Did I mention how awesome your docstrings are!  5:D

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -0,0 +1,716 @@
...
+        changes almost in every single implementation.
+        """
+        raise NotImplementedError("You must implement this method")
+
+    def get_symbol_fullname(self, symbol_basename: str) -> str:
+        """Given a short symbol or type name, it returns its full name"""
+        return self.vmlinux.symbol_table_name + constants.BANG + symbol_basename
+
+    @staticmethod
+    def get_member_type(
```
ikelos, line 313: This feels a little black magicy...  You're definitely playing around with volatility internals here.  My concern is if there's a change and we add another object with a subtype or something.  I can see it gets used a lot and looks useful, but I worry about how robust it is.  If you're happy with it, I'll go for it, it's just got my spidey sense tingling a little...  5;)

ikelos (commented):
Yep, really nice plugin.  A couple of minor niggles, some bits that make me uneasy but otherwise really good.  And those docstrings!!!  Keep those coming in, they're awesome!!!  5:D

volatility3/framework/plugins/linux/netfilter.py
gcmoreira: good catch

volatility3/framework/plugins/linux/netfilter.py
gcmoreira: ok, ^^ 'from volatility3 import framework'

volatility3/framework/plugins/linux/netfilter.py
gcmoreira: Each class overrides this, so I think it's convenient to keep them at hand, in its own class, for better readability

volatility3/framework/plugins/linux/netfilter.py
gcmoreira: Well, so far I think it's handy as it's like a shortcut to avoid the same formula (x+y+z) that kept showing up everywhere in the code. That immediately made me think that it may need an abstraction. 

On the other hand, the good news is that if your concern becomes true in the future, we hopefully only need to fix this in one spot.

gcmoreira:
@ikelos all the suggestions were applied and it passes all the tests. It looks like it's ready to merge

volatility3/framework/plugins/linux/netfilter.py
ikelos: Ok, makes sense, happy with that.  5:)

volatility3/framework/plugins/linux/netfilter.py
ikelos: Well, this probably does need extracting, maybe adding as a method alongside [has_member](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/interfaces/objects.py#L279)?  That would let each type override it, but I'm wondering how useful it would be for most types, since it only really applies to structs?

ikelos (commented):
Nearly there, just debating over where to put `get_member_type`.  I'm also wondering if there's now enough similarity between `Template`s and `Object`s to pull out a common class between them both?

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -0,0 +1,714 @@
...
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Linux kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.PluginRequirement(
+                name="lsmod", plugin=lsmod.Lsmod, version=(2, 0, 0)
+            ),
+            requirements.VersionRequirement(
```
ikelos, line 694: Hmmm, should this be pulled from the abstract classes that acutally use LinuxUtilities?  Here is ok, but it's kinda split from the actual types, tagging it in there (either by comment or as a helper method) seems like it might avoid issues in the future...

gcmoreira:
@ikelos this PR has been here for a while and I think we've already addressed the most crucial parts. If the remaining suggestions aren't critical, could we please go ahead and merge it as is? We can always make improvements later

volatility3/framework/plugins/linux/netfilter.py
ikelos: This hasn't been addressed yet.  These abstract Netfilter objects are the actual consumers of the `linuxutils` module, meaning that they should check for its presence in `__init__` and that it's the right version, and throw an exception if it's not met.

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -0,0 +1,714 @@
...
+        self._context = context
+        self.vmlinux = context.modules[kernel_module_name]
+        self.layer_name = self.vmlinux.layer_name
+
+        # Set data sizes
+        self.ptr_size = self.vmlinux.get_type("pointer").size
+        self.list_head_size = self.vmlinux.get_type("list_head").size
+
+        modules = lsmod.Lsmod.list_modules(context, kernel_module_name)
+        self.handlers = linux.LinuxUtilities.generate_kernel_handler_info(
```
ikelos, line 102: Before this gets called, we need to know that the version of `LinuxUtilities` is suitable, that means checking it earlier in the `__init__`.  Ideally using the same code that's already there, ie `VersionRequirement.matches_required`.

ikelos (commented):
Sorry, there is still something that needs fixing before it goes live.  At the moment anyone thinking of reusing the NetFilter objects will fail due to the hidden dependency on `linuxutils` they all have.  It should be easy to verify that in the `__init__` before it gets called.  Technically the failure to import should save us from needing to check it at the plugin level too (but best to leave it in the plugin level so people know to check).

volatility3/framework/plugins/linux/netfilter.py
gcmoreira: Noted and fixed. However, the same happens with `lsmod` in line 82 in the above code snippet, right?

ikelos: Yep, that's true.  The checks ideally should happen as part of the code that calls them, not just the one plugin that makes use of them.

gcmoreira: Done. Hope that works. Thanks

ikelos: Yep, I think that should cover it, thanks!  5:)

ikelos (approved):
Ok, the requirement checking looks like it's now done where needed, so this looks good to go in.  Still feel `get_member_type` is a little black magicky, but not a showstopper I think.

Issue #710, opened 2022-04-24
Missing notepad and clipboard plugins from volatility 2


HasegawaAzusa:
Some of the functions of vol2 are not available to me in vol3.
e.g. `vol -f xxx notepad` or `vol -f xxx clipboard`
Even there is no way to view the history of the command line.
This is fatal to forensics.
I am currently unable to use vol3 to complete normal forensic actions, can you please make vol3 compatible with vol2 as soon as possible?
I do love the fast and modular design of vol3 and I hope vol3 will one day replace vol2 in the future.

ikelos:
Thanks for your comment.  The clipboard plugin I don't know a great deal about, but the `notepad` plugin doesn't work in more recent versions of windows (even under volatility 2) because it's based on the way that the heap was handled back in windows 7.  As such, it's unlikely anyone will invest time in replicating a plugin that won't have use in the future.  Figuring out the modern heap handling code is a big task in itself, and there's not a very high priority need for it.  I'm also slightly unclear how not having the contents of a notepad document could be "fatal to forensics", I've only ever seen it used in capture the flag challenges?  Hyperbole doesn't really help the situation.

Volatility 2 is no longer being developed, and doesn't run on python 3.  Python 2 was marked as end of life on 1 Jan 2020.

I've marked this as a plugin-request bug so that volunteers that want additional functionality can look towards adding it, and knowing what plugins have been requested.  We're a very small volunteer group and our time commitments are varied, so I can't say when either of these plugins will be written by the core team, but if you'd like to try your hand at writing them we'll be happy to offer support and advice for merging them into the main codebase...

its5Q:
I've been bored lately, delving deeper in memory forensics, and decided to make a [notepad plugin](https://github.com/its5Q/volatility3-plugins/blob/main/plugins/windows/notepad.py) for volatility3 myself. It doesn't parse any heap structures or anything fancy like that, it uses a pretty simple memory pattern to look for the displayed text in VADs and a huge charset to filter garbage, It can rarely produce false-positives, but it's better than nothing :)

@ikelos, should I make a PR?

ikelos:
Yes please!  We're always happy to review contributions!  I can't say whether it'll get included, but at least if there's a PR people may find it.  If you could put it in the `plugins/windows` directory, rather than `framework/plugins/windows` that's where we put windows plugins that the core team doesn't fully support/look after and those which are a little more experimental.  Thanks!

its5Q:
[...]

PR #1082, opened 2024-01-11
Add new plugin: windows.notepad.Notepad

its5Q:
Adds a new plugin that is an alternative to volatility2's notepad plugin. Instead of parsing heap structures, it uses simple memory pattern matching to find potential places where the displayed text might reside. It can produce false positives, but I've tried to reduce them to a minimum by filtering strings shorter or equal to 8 characters, and using a table of unicode characters seen in CommonCrawl to remove strings that contain unicode characters that are technically valid, but didn't appear even once in the wild web.

Partially addresses issue #710.


volatility3/plugins/windows/notepad.py
```diff
@@ -0,0 +1,707 @@
...
+            self.context,
+            kernel.layer_name,
+            kernel.symbol_table_name
+        )
+
+        for proc in procs:
+            proc_id = "Unknown"
+            try:
+                proc_id = proc.UniqueProcessId
+                proc_layer_name = proc.add_process_layer()
```
ikelos, line 638: I'd prefer proc_layer_name being given a bogus default initalization, much like `proc_id` above, although `proc_id` above should initialize to a number or `None`.  To support the `unknown` name, change the debug message to be `proc_id or 'Unknown'` although note that if `proc_id` is then 0 this will result in the layer being referred to as Unknown.

volatility3/plugins/windows/notepad.py
```diff
@@ -0,0 +1,707 @@
...
+                    protect_values,
+                    vadinfo.winnt_protections
+                )
+                private = int(vad.get_private_memory())
+                tag = vad.get_tag()
+
+                if private and protection == 'PAGE_READWRITE' and vad_size < 16777216 and tag in ('VadS', 'Vadm'):
+                    try:
+                        data = proc_layer.read(vad_start, vad_size, pad=True)
+                        if not data:
```
ikelos, line 662: You're padding, so this will always return a binary string of `vad_size`, so this check should'nt ever fire unless `vad_size` is 0, which you could check before attempting the read, so this doesn't seem useful.

ikelos (commented):
I've got to say, I'm pretty uncomfortable about this plugin?  I wouldn't ever add it to `framework/plugins`.  It feels as though it's entirely heuristic based without any formal forensic analysis or backing.  It might work occasionally but I'm concerned about adding it to what is otherwise a forensic framework that may be relied upon for evidence.  It also feels as though this could be applied to programs other than notepad and still return partial results?  Is this notepad specific or just a general text search through program VADs?

At the moment this is still in limbo and I'm not sure whether it'll ever get added, I need to think over it and discuss it with the others a bit I'm afraid...

volatility3/plugins/windows/notepad.py
```diff
@@ -0,0 +1,707 @@
...
+from volatility3.framework import exceptions, interfaces, renderers
+from volatility3.framework.configuration import requirements
+from volatility3.framework.renderers import format_hints
+from volatility3.plugins.windows import pslist, vadinfo
+
+vollog = logging.getLogger(__name__)
+
+# I need this huge set for filtering out garbage strings, as I rely on a memory pattern to find the displayed text
+# Taken from https://github.com/glmcdona/binary2strings/blob/02f0c983aeede7ed9acbee96a43f9142dc7ba404/src/binary2strings.hpp#L57
+COMMONCRAWL_CHARS = {0x20, 0x65, 0x61, 0x69, 0x6f, 0x74, 0x6e, 0x72, 0x73, 0xa, 0x6c, 0x64, 0x63, 0x75, 0x68, 0x6d,
```
ikelos, line 14: This large a chunk of data doesn't belong in the plugin source, it should be an additional data file.

ikelos (commented):
Forgot to add this first time round, but this much data doesn't belong in the program code, it should be in auxiliary file.  Also pointing out the code where it came from, where the code is just another arbitrary program doesn't really act as evidence that this is a suitable or accurate means of identifying appropriate strings...

volatility3/plugins/windows/notepad.py
its5Q: Oh, I see, that makes sense. It was part of the code from the VadInfo plugin that read the VADs in chunks. I've removed the chunking and forgot to this check, so it wasn't intentional, and even if it would fire, it would break out of the loop over the VAD list, which I certaintly didn't intend to do.

its5Q:
> I've got to say, I'm pretty uncomfortable about this plugin? I wouldn't ever add it to `framework/plugins`. It feels as though it's entirely heuristic based without any formal forensic analysis or backing. It might work occasionally but I'm concerned about adding it to what is otherwise a forensic framework that may be relied upon for evidence. It also feels as though this could be applied to programs other than notepad and still return partial results? Is this notepad specific or just a general text search through program VADs?
> 
> At the moment this is still in limbo and I'm not sure whether it'll ever get added, I need to think over it and discuss it with the others a bit I'm afraid...

Yeah, totally understandable, I agree with you. Being based exclusively on heuristics of where the searched text *might* be, this is not the best approach for a forensics framework that aims to produce reliable output. I plan on studying NT and LFH heap internals so maybe I could figure something out with that, but for now I think it would be best to make this PR a draft. I'm very much a beginner in writing code to work with memory dumps, and I'm eager to learn, so I hope I don't abandon this idea :D

ikelos:
That sounds like the best course of action, that way people can find and use the code if they want it, but it's clearly not part of the main codebase.  Good luck with the heap support!  I believe the volatility 2 notepad plugin only works up to a particular version of windows, but if you can find one that legitimately reads the right structures to get the actual contents then we'd love to have it included as a core plugin...  Thanks very much for your contribution, we're happy to leave it here in draft for people to find and we look forward to seeing what you come up with in the future!  5:D

PR #1083, opened 2024-01-11
Windows: Add regex filtering to dumpfiles

brandon-barnacle:
In volatility2, the windows.dumpfiles plugin allows you to filter the dumped files using a regular expression. This PR adds the same functionality to volatility3.

The regular expression is passed in using --regex=REGEX and all files matching REGEX will be dumped. The --ignore-case flag can be passed to make the search case-insensitive. The search is case-sensitive by default. The matching volatility2 functionality can be found here:
https://github.com/volatilityfoundation/volatility/blob/a438e768194a9e05eb4d9ee9338b881c0fa25937/volatility/plugins/dumpfiles.py#L844

Manual testing was performed on windows memory images across different windows versions to verify the expected output.

### DIFF
--- a/volatility3/framework/plugins/windows/dumpfiles.py
+++ b/volatility3/framework/plugins/windows/dumpfiles.py
@@ -55,11 +55,11 @@ class DumpFiles(interfaces.plugins.PluginInterface):
                 optional=True,
             ),
             requirements.StringRequirement(
-                name="regex", description="Dump files matching REGEX", optional=True
+                name="filter", description="Dump files matching regular expression FILTER", optional=True
             ),
             requirements.BooleanRequirement(
                 name="ignore-case",
-                description="Ignore case in pattern match",
+                description="Ignore case in filter match",
                 default=False,
                 optional=True,
             ),
@@ -218,11 +218,11 @@ class DumpFiles(interfaces.plugins.PluginInterface):
 
     def _generator(self, procs: List, offsets: List):
         kernel = self.context.modules[self.config["kernel"]]
-        if self.config["regex"]:
-            if self.config["ignore-case"]:
-                file_re = re.compile(self.config["regex"], re.I)
-            else:
-                file_re = re.compile(self.config["regex"])
+        file_re = None
+        if self.config["filter"]:
+            flags = re.I if self.config["ignore-case"] else 0
+            file_re = re.compile(self.config["filter"], flags)
+
 
         if procs:
             # The handles plugin doesn't expose any staticmethod/classmethod, and it also requires stashing
@@ -259,7 +259,7 @@ class DumpFiles(interfaces.plugins.PluginInterface):
                         if obj_type == "File":
                             file_obj = entry.Body.cast("_FILE_OBJECT")
 
-                            if self.config["regex"]:
+                            if file_re:
                                 name = file_obj.file_name_with_device()
                                 if isinstance(name, UnreadableValue):
                                     continue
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/dumpfiles.py
```diff
@@ -243,6 +258,14 @@ def _generator(self, procs: List, offsets: List):
                         obj_type = entry.get_object_type(type_map, cookie)
                         if obj_type == "File":
                             file_obj = entry.Body.cast("_FILE_OBJECT")
+
+                            if self.config["regex"]:
```
ikelos: Presumably this could/should just check whether file_re is set?  Going back to the config allows the (admittedly, extremely unlikely) possibility that something changes the config between when you checked it and when you check it again.

volatility3/framework/plugins/windows/dumpfiles.py
```diff
@@ -272,6 +295,13 @@ def _generator(self, procs: List, offsets: List):
                         if not file_obj.is_valid():
                             continue
 
+                        if self.config["regex"]:
```
ikelos: Same as above.

volatility3/framework/plugins/windows/dumpfiles.py
```diff
@@ -53,6 +54,15 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                 description="Dump a single _FILE_OBJECT at this physical address",
                 optional=True,
             ),
+            requirements.StringRequirement(
```
ikelos, line 57: So adding filtering options like this is great, but also needs some consideration.  Is it consistent with other filter options?  `regex` isn't that descriptive a name, I'd prefer `filter` or `match` or something like that, and then let the `--help` describe that it's a regular expression.  Most people will know what regexs are, but not everybody, and it doesn't really describe the purpose of it.

ikelos: Consideration also needs to be used to determine how this option interacts with others?  If people provide `virtaddr` and `regex`, which wins? (or are they both applied?  Does that make sense?)  What about `pid`?  It should be clear and/or obvious which options will interact with each other and which cancel others out.

volatility3/framework/plugins/windows/dumpfiles.py
```diff
@@ -208,6 +218,11 @@ def process_file_object(
 
     def _generator(self, procs: List, offsets: List):
         kernel = self.context.modules[self.config["kernel"]]
+        if self.config["regex"]:
+            if self.config["ignore-case"]:
+                file_re = re.compile(self.config["regex"], re.I)
```
ikelos: Also, `file_re` needs to be initialized.

ikelos (commented):
Generally the coding style looks good, and the changes are minimal, but adding options to a plugin always needs some thinking about, so there's just some questions about the precedence of the command line options and how that would work.

volatility3/framework/plugins/windows/dumpfiles.py
brandon-barnacle: Yes this makes sense! I will change in both spots.

volatility3/framework/plugins/windows/dumpfiles.py
brandon-barnacle: The flag name change definitely makes sense and `--filter` makes a lot of sense, so I think changing to that will be good.

In terms of how it will interact with the other options, I think it makes sense that it cannot interact with the flags that take in an address, but it can work with the `pid` flag. My newest commit addresses this.

ikelos (commented):
Thanks, this looks much better.  Please could you make sure it's formatted by black and then we can get it merged...  5:)

ikelos:
So the rest of the codebase was still formatted against 23.12.0 rather than 24.1.0 (which is what github now uses).  The issues it's finding are from the original code, so this should merge in just fine and I'll fix it up manually if not...

PR #1084, opened 2024-01-14
Symbols: Add support for metadata verification of symbol tables

ikelos:
Ok this adds support for plugins to test that specific tables (or all tables by default) have a symbol file produced by a suitable generator.  This will not warn on unknown generators, and will not warn if the version of the produced file is suitably high.

As an example `kmsg` has been adapted to include the check.  The check is pretty flexible, it somewhat assumes that the version is made of just digits and dots, but this wasn't mandated in schemas.  A new schema has been introduced as version 6.3.0, which now requires a name and version for the producer information (and requires the producer information).  Since it only forces future requirements, its seen as an additive change, hence the schema number only had the MINOR version bumped.

If anyone spots any issues with the tinkering I've done to the symbol tables, please shout.  The idea is that this can be used to essentially deprecate particular versions of symbol tables, but on a plugin-by-plugin basis (and by producer).

### DIFF
--- a/volatility3/framework/constants/__init__.py
+++ b/volatility3/framework/constants/__init__.py
@@ -44,8 +44,8 @@ BANG = "!"
 
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 2  # Number of releases of the library with a breaking change
-VERSION_MINOR = 5  # Number of changes that only add to the interface
-VERSION_PATCH = 2  # Number of changes that do not change the interface
+VERSION_MINOR = 6  # Number of changes that only add to the interface
+VERSION_PATCH = 0  # Number of changes that do not change the interface
 VERSION_SUFFIX = ""
 
 # TODO: At version 2.0.0, remove the symbol_shift feature
--- a/volatility3/framework/plugins/linux/kmsg.py
+++ b/volatility3/framework/plugins/linux/kmsg.py
@@ -15,7 +15,6 @@ from volatility3.framework import (
     renderers,
 )
 from volatility3.framework.configuration import requirements
-from volatility3.framework.interfaces import plugins
 from volatility3.framework.objects import utility
 
 vollog = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


eve-mem:
Based on the example changes to `kmsg` this looks very easy to use from a plugin writing perspective, and helps work around the issue of older symbols nicely with clear warnings to users when it get's tripped. 

Thanks - I think this will be really helpful.

eve-mem:
I've tested this and it works as expected when using an outdated ISF. Nice!
```
$ python vol.py -f linux-sample-1.dmp linux.kmsg
Volatility 3 Framework 2.6.0

Volatility experienced an issue related to a symbol table:
Invalid symbol table, please ensure the ISF table produced by dwarf2json was produced using a version > 0.4.1

        * An invalid symbol table
        * A plugin requesting a bad symbol
        * A plugin requesting a symbol from the wrong table

```

One question I have is how this would work when a plugin isn't actually run, but say one of it's methods is used by a different plugin? At the moment it feels like it would be the case of the different plugin also needs to check the versions itself, and they need to keep in sync. Which is fine - just something to remember.

Do you think there is any way to have that happen automatically somehow? Somehow by importing a plugin that uses these checks also make there way into the new plugin too?

ikelos:
So, if it isn't already, it should be that if pluginA requires pluginB, the requirement on pluginB isn't fulfilled until any requirements it may have are fulfilled.  I don't know that we check for that yet, I think at the moment VersionRequirements just check the versions, but a plugin requirement might also check the sub-requirements...

Currently it doesn't look like we do that it but should be possible to slot it into the PluginRequirement, but we'll need to distinguish between pre-requisites that are pre-requisites of the plugins (like PluginRequirements) versus runtime inputs (standard list information, etc).  So it's a good point (and one I'll open a separate ticket about, #1095 )...

ikelos:
There's no good way to know that the inputs of a classmethod are correct, so the site intending to use the symbol table should check the symbol table is valid to its requirements before they use it.  This may come with a performance hit.

PR #1085, opened 2024-01-15
CLI: Add in support for selective column output

### LINKED ISSUE #1081: Build a UI based way of choosing columns to display
A lot of plugins can return a lot of data, but a user may not want to see all the data produced.  Unless the extra data is extremely computational difficult to produce (and would significantly increase runtime), this is a UI decision and should be something the UI can decide.  The UI would need some kind of a hint to suggest which columns should ordinarily get shown, and which ones are only for those that really want all the data.  As such, a column hint mechanism to determine which columns should be displayed by default should be added to the plugins, and then the CLI should be updated to support this and to filter down to specific named columns...

ikelos:
This adds support for selective column output on the CLI.  Other UIs are welcome to implement their own features.  Still to add to this is the ability for plugins to provide their own list of expected columns (currently, all columns are output by default).

### DIFF
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -4,6 +4,11 @@ API Changes
 When an addition to the existing API is made, the minor version is bumped.
 When an API feature or function is removed or changed, the major version is bumped.
 
+2.6.0
+=====
+Plugins defining treegrid columns can use three-tuples, including a new third "extra" parameter to determine if a column
+should be hidden by default.
+
 2.5.0
 =====
 Add in support for specifying a type override for object_from_symbol
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -233,6 +233,14 @@ class CommandLine:
             default=False,
             action="store_true",
         )
+        parser.add_argument(
+            "--columns",
+            help="Case-insensitive space separated list of prefixes to determine which columns to output if provided (otherwise all columns)",
+            default=None,
+            # action="extend", Can be enabled when python 3.8 is made default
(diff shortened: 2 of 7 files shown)


ikelos:
Still want to do a little work on this to allow plugins to have a default set of columns and then a `--all` flag or something to display all columns from a plugin.  Not sure whether I should make that something passed to the TreeGrid or not...

ikelos:
There's now the ability for plugins to say whether a particular column should be displayed or not (although this is for the UI to decide how to interpret).  The plugin will still output all data, and it's the UI's job to decide what the user sees based on how it decides to display the data.  The CLI will now display all columns except those marked as extra within the plugin (by creating a column with a third boolean parameter set to true).  This means that the tree grid is still defined early and ideally should not be altered between runs of a plugin.  To display *all* columns, pass in `--columns` without any parameters, this must not be the last or only parameter to the CLI, or it will interpret the plugin name as a column name and say that no plugin name has been provided.  This is like all other list parameters, but this is one of the only CLI parameters that accepts a list.  No plugins currently mark any columns as extra.  Since the additional value for making columns is an API change, plugins making use of it must depend on the appropriate version.

I'm still not sure I really like this, I skirted actually implementing renderer options because it would be a lot of work to define them properly, they need to be able to take in user input as well as pass data from the plugin itself in this instance, and it would probably have a significant impact on plugins, whereas at the moment this is quite minimal.  Still uncertain about the whole thing though (and the way of displaying all columns in the CLI is a little awkward).  The good news is it shouldn't impact any other UIs at all.

Issue #161, opened 2019-12-18
Add in Arm64 support


ikelos:
This is a placeholder for requests concerning Arm/Arm64 support.

JRock007:
Hi, I just saw this issue and happen to have an Arm64 Android device, and would love to see support added in Volatility 3!

I built json files from the System.map and module.ko files (available respectively [here](https://www.dropbox.com/s/vuvibzlc1no3ch5/system-map.json?dl=1) and [here](https://www.dropbox.com/s/415scfpy8op4aje/module.json?dl=1)) using `dwarf2json`.

I also uploaded the original [System.map](https://www.dropbox.com/s/kfsi4ccq16g8s7s/System.map?dl=1) and  [module.ko](https://www.dropbox.com/s/ztvjyq0r38rj8g7/module.ko?dl=1) files.

I have a 3Gb memory dump from a physical device running Android 9 (and the kernel 4.9) acquired using LiME that I can upload.

Is there any way this could be implemented, or for me to help?

ikelos:
Hiya, sorry I've taken so long to get back to you, my time's been diverted a bit recently.  Thanks very much for the files you've provided.  Getting the memory image that goes with them would be much appreciated, in the past we've accepted memory images submitted through google drive (you can send it directly to me as mike.auty@gmail.com and just let me know that you're happy with me sharing it with the other Volatility developers or not).

As this placeholder suggests, it is on our list of things we'd like to implement, but there's a few other large tasks (such as compressed memory for windows) which we need to try and work on too, so I'm afraid I can't guarantee how quickly we'll be able to add it...  5:S

JRock007:
Thanks for your answer! Just sent you an email with a link to the memory dump

kohnakagawa:
Hi. I hope volatility3 support for Windows 10 on Arm.

Currently, Windows 10 on Arm is not widely used, but this OS will be gradually used as the Arm64 laptop devices (e.g., Surface Pro X) come. I checked the [symbol tables](https://github.com/volatilityfoundation/volatility3#symbol-tables) for Windows. However, these symbol tables for Windows 10 on Arm seems to be missing.

Is there any way to implement this? If you do not have enough time, I will help.

Thanks in advance!

Blue0fSky:
Hi,

I would like to inquire if support for the arm/arm64 architectures will soon be implemented. Some files in the repository appear to take into account both architectures, but I am unable to analyze a memory dump from a machine running on armv7.

[...]

PR #1036, opened 2023-11-11
Feature : Hibernation Layer and plugins.

forensicxlab:
Hello, 

- Adding a new translation layer and plugins to allow modern Windows (Win8-Win11) hibernation file conversion to a raw memory image
- Adding new codecs : LZ77, LZ77Huffman decompression algorithms.

Details about the feature : https://www.forensicxlab.com/posts/hibernation/

Best regards.

Issue #464, opened 2021-03-04
VM Scan


xSalice:
**Problem:**
Would it be possible to be able to detect nested-vm inside a memory image and use volatility plugins onto those vm? Example: VMWare was running on the host with another windows machine. 

**Solution:**
Add a new volatility layer similar to rekall's vmscan to detect nested vm.

ikelos:
Hiya, so we'd need to think about how you interact with memory images that could be containers for other memory images, or may be just plain memory images themselves.  It's certainly possible already at the moment to access a guest by creating a specific configuration file and then use that to run volatility plugins, but we don't have anything that looks for virtual machines yet.  I don't think a new layer would necessarily solve those problems, but we can certainly investigate it...  5:)

xSalice:
Cool! Maybe what you can do potentially:

- Create a argument to allow for that vm configuration with its EPT to access that guest
- Seperate plugin for vm scanning/finding the EPT?

Thanks for looking into it!

ikelos:
The vmscanning plugin sounds fine, but I'm quite uncomfortable about adding a new command line flag to CLI without a lot more thought of how it might affect the library and other user interfaces, so that will likely take a fair bit longer.  I need to decide whether that should be part of the automagic or not, since it's (at the moment) a reasonably niche requirement and there may be other ways of handling it (for example, there could be a vm-aware automagic that creates the additional layers automatically when it finds them but then how does the plugin know which one to act upon, etc).  I just need to plan out how best to integrate it, but as I say, it's technically possible at the moment with the appropriate configuration object.

xSalice:
Circling back, what kind of configuration would I need as an example to hard code in order to peak into a guest? An example usage of the configuration in the mean time would help tremendously!

ikelos:
Sorry for dropping the ball on this, I've been tinkering with things along these lines but still don't have anything to show yet.

The kind of a config you'd be looking at would be.  The Intel32e layer could be used to do the mapping, although technically it might miss pages that the guest isn't allowed to read but are still mapped (it's a bit weird), but it ought to get you most of the way there.  This is also configured for the more recent kernel requirements, you could change it to the old primary and symbols requirements if necessary, that should still work...

```
{
    "kernel.layer_name.class": "volatility3.framework.layers.intel.Intel32e",
    "kernel.layer_name.memory_layer.class": "volatility3.framework.layers.intel.Intel32e"",
[...]

PR #1088, opened 2024-01-20
[2024] AArch64 support

Abyss-W4tcher:
Hi :wave:,

This PR provides AArch64 integration to the Volatility3 framework, as well as a design rework in the current linux stacker. 

Implementation follows ARM official documentation, and includes essential APIs for higher level code (plugins etc.).

You can follow the roadmap here : 
- https://github.com/volatilityfoundation/volatility3/issues/161
---

My ressources : 
- https://gist.github.com/Abyss-W4tcher/f1833623c975193446315d48c106750e
- https://gist.github.com/Abyss-W4tcher/8442b6b6b85f725158fe7e9b99e507be
---

Testing : 

- RaspberryPi (virtual) : `Linux version 6.1.21`
- Android AVD (virtual) : `Linux version 3.18.94`
- RaspberryPi (physical) : `Linux version 6.1.0-rpi8-rpi-v8`
- Google Pixel 6A (physical) : `Linux version 5.10.157-android13-4-00003-g830b023b88f3-dirty`
- Jetson lime (physical) : `Linux version 4.9.201-tegra`
- NXP LX2160 (physical) : `Linux version 6.1.36`

_Thanks to everyone who took the time to test this PR on their devices !_

---

Unstable plugins : 

- `linux.bash` : 
     - `sections have no size` on Android related samples. 
     - 12/04/24 update : Android does not ship with `bash` by default, as [it's not GNU/Linux running on the system ](https://www.reddit.com/r/AskTechnology/comments/z2vqyt/comment/ixkaf1l/). Either we need a dedicated plugin (`.ash_history` ?), or a check to tell users why the plugin isn't working.
- `linux.lsof` : 
     - `dentry` members can be NULL, and this behaviour isn't handled (more informations in the dedicated Slack thread)

---

Please note that this is still experimental, testing is still going on. If you want to try this PR out, here are the steps : 

- Merge/checkout the PR into your local Volatility3 copy (`git checkout aarch64-support`)
- Make a memory capture of your AArch64 device (see attached ressources)
- Create the ISF against the vmlinux file with `dwarf2json`
- Set the following on top of the plugins you plan to use : 
```python
requirements.ModuleRequirement(
      name="kernel",
      description="Linux kernel",
      architectures=["Intel32", "Intel64", "AArch64"],
)
```

If you have any *real-life* samples to provide, or if you encounter any error, feel free to comment below !


volatility3/framework/automagic/linux.py
```diff
@@ -162,9 +326,12 @@ def find_aslr(
                 init_task.files.cast("long unsigned int")
                 - module.get_symbol("init_files").address
             )
-            kaslr_shift = init_task_address - cls.virtual_to_physical_address(
-                init_task_json_address
-            )
+            if layer_class == arm.AArch64:
```
ikelos: Rather than changing the API, it should be possible to derive this by getting the layer using `layer_name` and then doing an if on the class of the layer.  The if statement feels a little ugly though, since it'll need support for every layer that comes about.  It might be worth separating this out into a class that can be attached to each class, so that a new layer class can be introduced without tinkering with this code in the future?

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,475 @@
...
+        config_path: str,
+        name: str,
+        metadata: Optional[Dict[str, Any]] = None,
+    ) -> None:
+        super().__init__(
+            context=context, config_path=config_path, name=name, metadata=metadata
+        )
+        self._base_layer = self.config["memory_layer"]
+        # self._swap_layers = []  # TODO
+        self._page_map_offset = self.config["page_map_offset"]
```
ikelos, line 103: There are two pages maps as part of AArch64, aren't there?  Which one should people provide here?  Might be better to use the terminology used by the architecture here, rather than reusing values used elsewhere...

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,475 @@
...
+        self._tcr_el1_tnsz = self.config["tcr_el1_tnsz"]
+        self._page_size = self.config["page_size"]
```
ikelos: Is this part of the architecture, or is it dynamic, or is it something pre-defined?

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,475 @@
...
+
+        # Context : TTB0 (user) or TTB1 (kernel)
```
ikelos: So this layer will only cope with one of user or kernel?  That might be complex to handle in the future?

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,475 @@
...
+        }
+        self._ttb_lookups_descriptors = self._determine_ttbs_lookup_descriptors()
+
+        # [1], see D8.3, page 5852
+        self._descriptors_bits = [
+            (
+                49
+                if self._ttbs_granules[ttb] in [4, 16] and self._is_52bits[ttb]
+                else 47,
+                self._ttb_lookups_descriptors[ttb][-1][1],
```
ikelos: It would be handy to have some descriptions of what the array looks like somewhere (maybe in the docstring for `_determine_tbbs_lookup_descriptors`), otherwise it's difficult to tell what the last element would contain.

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,475 @@
...
+        ]
+
+        self._context_maxvirtaddr = self._ttbs_bitsizes[self._virtual_addr_space]
+        self._canonical_prefix = self._mask(
+            (1 << self._bits_per_register) - 1,
+            self._bits_per_register,
+            self._context_maxvirtaddr,
+        )
+
+        if AARCH64_DEBUGGING:
```
ikelos: Thanks, this is handy.  It might be worth parameterizing the layer with a default value, so it still gets intialized normally, but can be tagged on in volshell, etc?

ikelos: It might also be handy to have a method that can be called to print out this information, so this doesn't need to be done at construction time

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,475 @@
...
+            vollog.debug(f"TnSZ values : {self._ttbs_bitsizes}")
+            vollog.debug(f"Page map offset : {hex(self._page_map_offset)}")
+            vollog.debug(f"Descriptors mappings : {self._ttb_lookups_descriptors}")
+
+    def _determine_ttbs_lookup_descriptors(self) -> List[int]:
+        """Returns the bits to extract from a translation address (highs and lows)"""
+        ttb_lookups_descriptors = []
+
+        for ttb, ttb_granule in enumerate(self._ttbs_granules):
+            va_bit_size = self._ttbs_bitsizes[ttb]
```
ikelos: It might be easier to understand if these values are passed in as parameters, rather than references back to attributes that may have already been set?  It would also need doing to `_granules_indexes`, but then this could become a classmethod, and easier to ensure everything is set appropriately and doesn't change unexpectedly.

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,475 @@
...
+            else:
+                raise exceptions.PagedInvalidAddressException(
+                    layer_name=self.name,
+                    invalid_address=virtual_offset,
+                    invalid_bits=low_bit,
+                    entry=descriptor,
+                )
+            level += 1
+
+        if AARCH64_TRANSLATION_DEBUGGING:
```
ikelos: Again, probably better to have a togglable instance variable, rather than a global module variable.

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,475 @@
...
+        return [self._base_layer]
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.TranslationLayerRequirement(
+                name="memory_layer", optional=False
+            ),
+            requirements.IntRequirement(name="page_map_offset", optional=False),
+            requirements.IntRequirement(name="page_map_offset_kernel", optional=False),
```
ikelos: This... doesn't look like it's used anywhere, and anything related to the kernel shouldn't be a requirement of the layer to operate, it should be optional, because the architecture should be able to map with or without a kernel?

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,475 @@
...
+            requirements.IntRequirement(name="tcr_el1_tnsz", optional=False),
```
ikelos: You can actually provide a description field with requirements, which might be quite handy here?

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,475 @@
...
+        upon."""
+        return [self._base_layer]
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.TranslationLayerRequirement(
+                name="memory_layer", optional=False
+            ),
+            requirements.IntRequirement(name="page_map_offset", optional=False),
```
ikelos: Might also be good to provide a description here (as noted later, `Requirement` classes can take a `description` attribute.  Just some clarification on exactly what this would be for arm, since it's not quite as obvious as with intel....

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,475 @@
...
+    @functools.lru_cache()
+    def minimum_address(cls) -> int:
+        return 0
+
+    @classproperty
+    @functools.lru_cache()
+    def maximum_address(cls) -> int:
+        return (1 << cls._maxvirtaddr) - 1
+
+    def __canonicalize(self, addr: int) -> int:
```
ikelos: These two don't appear to get called from anywhere and are strictly private?  Are they necessary?  If they can't be depended on in children of this class, then they don't seem all that useful?

ikelos (commented):
Thanks very much for this, it looks awesome!  I haven't been through it all, I've most concentrated on the arm layer at the moment (and I haven't made my way through all of that).  The comments pointing to the location of useful documentation is massively useful and very much appreciated!  5:D  There's a few little bits that can be cleaned up but on the whole it looks perfectly in line with the rest of volatility, so good work!  It'll take me a while to find time to go through more of it, and check the various references to fully get what's going on. 

I'm trying to figure out the split between `mapping` and `_mapping`, it looks like `_mapping` returns every chunk, and `mapping` coalesces them together is that about right?  I think that's probably for the best, but it feels like it makes for bulky code somehow.  Really excited to see where this codes (and find some Windows-on-arm samples to try it against too!).  5:D

Abyss-W4tcher:
Hi @ikelos, thanks for your quick replies ! 

I will check every of your comment right now, as I have some spare time. Regarding your general concern regarding `mapping`, it is more or less the exact Intel implementation from https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/layers/intel.py#L271. It worked well out of the box for this layer :).

volatility3/framework/layers/arm.py
Abyss-W4tcher: `"page_map_offset_kernel"` is instantied once by the LinuxStacker, when creating the kernel layer. This implies that any other caller will try to stack in user space, without passing `"page_map_offset_kernel"` but only `"page_map_offset"` (it is the same naming as Intel, to avoid rewriting everything). Thanks to how the framework is designed, the class will still be able to access `"page_map_offset_kernel"`, which indicates what is the current context (am I being instantiated for user space or for kernel space ?). See : 

```python
self._virtual_addr_space = (
            self._page_map_offset == self.config["page_map_offset_kernel"]
        )
```
`self._virtual_addr_space = 0` -> we are in user space/context (TTB0)
`self._virtual_addr_space = 1` -> we are in kernel space/context (TTB1)

Additionally, we need to keep track of the context, because of this part : 

```python
        # Check if requested address belongs to the context virtual memory space
        if ttb_selector != self._virtual_addr_space:
            raise exceptions.InvalidAddressException(
                layer_name=self.name,
                invalid_address=virtual_offset,
            )
```

Making the `page_size` or `tcr_el1_tnsz` options optional isn't viable, because they are constant for their whole spaces. In my sense, calculating everything once and for all in the LinuxStacker, allows compatibility with existing LinuxIntel APIs and facilitates higher layers stacking. As pointed out, Volatility will keep track of the previous config options for layers stacking on top of the AArch64 one.

volatility3/framework/layers/arm.py
Abyss-W4tcher: In the Linux Stacker, we are instantiating the kernel Layer. We pass the kernel DTB (Translation Table Base 1 Base Address)  to the AArch64 layer. There is in fact a unique page map for the whole kernel, stored in ARM register TTBR1, and a page map for the user land stored in TTBR0. TTBR0 is not constant, and changes on every process context switch. What this means, is that TTBR0 is changed to the value of the current process DTB (which happens continuously). 

For me, there are only two possible callers to this layer : 

- LinuxStacker (kernel, only once) 
- "Processes"

People should not have to instantiate the kernel layer twice, but only user space layers (programs) ? In another comment I detailed how I implemented a check to facilitate this.

volatility3/framework/layers/arm.py
Abyss-W4tcher: - `_page_map_offset` is the DTB (either kernel or specific to a process). For the kernel, it can be calculated thanks to the existing Intel implementation. For the process, I don't know exactly where, but it is stored in the `task_struct` directly if I'm not mistaken.
- `_tcr_el1_tnsz` : *The size offset of the memory region addressed by TTBR1_EL1. The region size is 2(64-T1SZ) bytes* -> See page 7081 of the doc. It is defined in the kernel config, but we cannot predict it (it is calculated in LinuxStacker).
- `_page_size` : Page size (4, 16 or 64 KB), also defined in the kernel config, but we cannot predict it (it is calculated in LinuxStacker).

volatility3/framework/layers/arm.py
Abyss-W4tcher: This layer needs to be instantiated one time for the kernel, and as many times as there are processes. End users should not have to instantiate the kernel layer twice, and instantiating a process layer is basically the same as the Intel logic (I think mostly of malfind, which I used to validate the processes layers translations).

volatility3/framework/layers/arm.py
Abyss-W4tcher: Noted, I will take a look !

volatility3/framework/layers/arm.py
Abyss-W4tcher: This is a direct inspiration from https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/layers/intel.py#L19, but I guess we could make things differently. This requires editing the code by hand, so it isn't practical indeed.

volatility3/framework/layers/arm.py
Abyss-W4tcher: I see, I can add a direct function parameter instead of relying on the value prior existence. I can make a getter for the `granules_indexes` with a classmethod too :)

ikelos:
Hehehe, yeah, I figured it might be.  That's not too sweet the intel layer isn't overly complicated but it's fine for now.  If you find any ways of simplifying it we can back port then to intel too... 5;D

volatility3/framework/layers/arm.py
Abyss-W4tcher: Noted, I will take a look !

volatility3/framework/layers/arm.py
Abyss-W4tcher: Inspired from : https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/layers/intel.py#L93 and https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/layers/intel.py#L119.  

I realize now that I messed up the visibility of canonicalize, and will fix it. I did not investigated which higher level code was likely to use these functions, but it definetely crashed when they were missing.

volatility3/framework/automagic/linux.py
Abyss-W4tcher: As the whole `find_aslr` function worked out of the box, without the need to use the `virtual_to_physical_address` function, I did this *quick fix* to avoid repeating the whole function (separating in `find_aslr_aarch64` and `find_aslr_intel`). 

But if in the future we have to add some other specific AArch64 code to `find_aslr` (e.g. handle some retrocompatibility), we will eventually do a clear separation. So I guess we could do it directly even if the code will be 99% the same ?

volatility3/framework/layers/arm.py
Abyss-W4tcher: Noted, I will take a look !

Abyss-W4tcher:
Here is what a typical layer instantiation debug looks like : 

Kernel layer instantiated by `LinuxStacker` and a higher stacker : 

```sh
DEBUG    volatility3.framework.automagic.linux: Linux ASLR shift values determined: physical -ffffffbfc7e00000 virtual 172a600000
DEBUG    volatility3.framework.layers.arm: Base layer : Elf64Layer
DEBUG    volatility3.framework.layers.arm: Virtual address space : kernel
DEBUG    volatility3.framework.layers.arm: Virtual addresses space range : ('0xffffffc000000000', '0xffffffffffffffff')
DEBUG    volatility3.framework.layers.arm: Page size : 4
DEBUG    volatility3.framework.layers.arm: T1SZ : 25
DEBUG    volatility3.framework.layers.arm: Page map offset : 0x41963000
DEBUG    volatility3.framework.layers.arm: Translation mappings : [(38, 30), (29, 21), (20, 12)]
DEBUG    volatility3.framework.automagic.linux: Kernel DTB was found at: 0x41963000
DEBUG    volatility3.framework.automagic.linux: AArch64 image found
DEBUG    volatility3.framework.layers.arm: Base layer : memory_layer
DEBUG    volatility3.framework.layers.arm: Virtual address space : kernel
DEBUG    volatility3.framework.layers.arm: Virtual addresses space range : ('0xffffffc000000000', '0xffffffffffffffff')
DEBUG    volatility3.framework.layers.arm: Page size : 4
DEBUG    volatility3.framework.layers.arm: T1SZ : 25
DEBUG    volatility3.framework.layers.arm: Page map offset : 0x41963000
DEBUG    volatility3.framework.layers.arm: Translation mappings : [(38, 30), (29, 21), (20, 12)]
```

Processes layers instantiated by `malfind` : 

```sh
DEBUG    volatility3.framework.layers.arm: Base layer : memory_layer
DEBUG    volatility3.framework.layers.arm: Virtual address space : user
DEBUG    volatility3.framework.layers.arm: Virtual addresses space range : ('0x0', '0x3fffffffff')
DEBUG    volatility3.framework.layers.arm: Page size : 4
DEBUG    volatility3.framework.layers.arm: T0SZ : 25
DEBUG    volatility3.framework.layers.arm: Page map offset : 0x47266000
DEBUG    volatility3.framework.layers.arm: Translation mappings : [(38, 30), (29, 21), (20, 12)]

DEBUG    volatility3.framework.layers.arm: Base layer : memory_layer
DEBUG    volatility3.framework.layers.arm: Virtual address space : user
DEBUG    volatility3.framework.layers.arm: Virtual addresses space range : ('0x0', '0x3fffffffff')
DEBUG    volatility3.framework.layers.arm: Page size : 4
DEBUG    volatility3.framework.layers.arm: T0SZ : 25
DEBUG    volatility3.framework.layers.arm: Page map offset : 0x47928000
DEBUG    volatility3.framework.layers.arm: Translation mappings : [(38, 30), (29, 21), (20, 12)]
```
---
Check out memory samples here [7 days] : 

- Raspberry, kernel 6.1.21-v8 : https://we.tl/t-m2aOCpsSU5
- Android API 27, kernel 3.18.94 : https://we.tl/t-bzFwbARKzZ

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,542 @@
...
+            )
+        except exceptions.InvalidAddressException:
+            return False
+
+    def is_dirty(self, offset: int) -> bool:
+        """Returns whether the page at offset is marked dirty"""
+        return self._page_is_dirty(self._translate_entry(offset)[2])
+
+    @staticmethod
+    def _page_is_dirty(entry: int) -> bool:
```
Abyss-W4tcher: Dirty state is hardware managed only > Armv8.1-A. It is software managed by Linux (bit 55), so we can use it reliably and for all Arm versions.

However, we need to know how Windows manages it (add a class `WindowsMixIn(AArch64)` which overwrites `_page_is_dirty`), or directly use the hardware managed version for any OS, but it won't work for memory dumps  < Armv8.1-A.

`[1], see D8.4.6, page 5877` and `[1], see D8.4.6, page 5877` and https://developer.arm.com/documentation/102376/0200/Access-Flag/Dirty-state

volatility3/framework/automagic/linux.py
```diff
@@ -39,11 +40,10 @@ def stack(
 
         # Bail out by default unless we can stack properly
         layer = context.layers[layer_name]
-        join = interfaces.configuration.path_join
 
-        # Never stack on top of an intel layer
+        # Never stack on top of a linux layer
```
Abyss-W4tcher, line 46: To improve this, why not add an explicit layer property, like `_is_top_layer` inside `volatility3/framework/layers/intel.py#Intel` and `volatility3/framework/layers/arm.py#AArch64` and check with following : 

```sh
        if getattr(layer, "_is_top_layer") and layer._is_top_layer:
            return None
```
 
If we keep the current implementation, we have to change the Linux, Windows and Mac stacker for each new architecture layer.

ikelos: Part of the reason is virtualization, it's possible to have an arm layer inside an intel layer (and it's certainly possible to have an intel layer on top an intel layer).  You're right, it's not really a scalable solution (hence the FIXME right beneath this line), but it's also not a trivial attribute...

Abyss-W4tcher: Which stacker is supposed to do this (LinuxStacker, WindowsIntelStacker and MacIntelStacker blocks it), for example if a VM managed by qemu-system-aarch64 on an Intel host was running when the memory was dumped ?

Will a "VM (qemu) layer" be available from the globals `context.layers` variable too :


- FileLayer 
  - LimeLayer
      - IntelLayer
        - Memory Layer
        - QemuLayer (not sure about this one)
          - IntelLayer
            - Memory Layer

eve-mem: I don't think there is anything automatic in core that would do that, but it would be nice to have. 

There has been this issue from a while ago that talks about that kind of thing.
https://github.com/volatilityfoundation/volatility3/issues/464

Abyss-W4tcher: Thanks for pointing it out :) So in the current state, as it is not implemented, we will continue to strictly refuse stacking on top of Linux and AArch64. Adding an explicit flag on those two *might* be a temporary and more scalable solution ? 

*I leave this as a side note, for a potentiel reader in the future interested in AArch64 hypervisor execution mode (and what it might imply, if treating a layer from the hypervisor point of view) :* https://developer.arm.com/documentation/102412/0103/Privilege-and-Exception-levels/Exception-levels

garanews:
@Abyss-W4tcher I am trying your branch 👀 
Target:
```
uname -a
Linux instance-20221125-1059 5.15.0-1040-oracle #46-Ubuntu SMP Fri Jul 14 21:47:21 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux

lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.3 LTS
Release:        22.04
Codename:       jammy

```

Dump:
```
sudo insmod /var/lib/dkms/lime-forensics/1.9.1-3/5.15.0-1040-oracle/aarch64/module/lime.ko "path=/tmp/arm_ram.lime format=lime"
```

Symbol creation:
```
sudo apt install linux-image-5.15.0-1040-oracle-dbgsym
dwarf2json linux --elf  /usr/lib/debug/boot/vmlinux-5.15.0-1040-oracle >  ubuntu_arm_5.15.0-1040.json 

```


Updated the pslist plugin adding AArch64 and run it:

```
python vol.py -vvvvvvvvvvvvvvv  -f arm_ram.lime linux.pslist

```


OUTPUT:
```
Level 6  volatility3.framework: Importing from the following paths: C:\ioc\volarm\volatility3\volatility3\framework\layers
Level 6  volatility3.framework: Importing from the following paths: C:\ioc\volarm\volatility3\volatility3\framework\layers
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
Level 6  volatility3.framework: Importing from the following paths: C:\ioc\volarm\volatility3\volatility3\framework\layers
Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel.symbol_table_name
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
Level 9  volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList
Level 6  volatility3.framework: Importing from the following paths: C:\ioc\volarm\volatility3\volatility3\framework\layers
Level 6  volatility3.framework: Importing from the following paths: C:\ioc\volarm\volatility3\volatility3\framework\layers
Level 6  volatility3.framework: Importing from the following paths: C:\ioc\volarm\volatility3\volatility3\framework\layers
Level 6  volatility3.framework: Importing from the following paths: C:\ioc\volarm\volatility3\volatility3\framework\layers
Level 6  volatility3.framework: Importing from the following paths: C:\ioc\volarm\volatility3\volatility3\framework\layers
DEBUG    volatility3.framework.automagic.stacker: physical_layer maximum_address: 25763184799
DEBUG    volatility3.framework.automagic.stacker: Stacked layers: ['IntelLayer', 'LimeLayer', 'FileLayer']
INFO     volatility3.framework.automagic: Running automagic: LinuxSymbolFinder
Level 9  volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DEBUG    volatility3.framework.automagic.symbol_finder: Identified banner: b'Linux version 5.15.0-1040-oracle (buildd@bos01-arm64-037) (gcc (Ubuntu 11.3.0-1ubuntu1~22.04.1) 11.3.0, GNU ld (GNU Binutils for Ubuntu) 2.38) #46-Ubuntu SMP Fri Jul 14 21:47:21 UTC 2023 (Ubuntu 5.15.0-1040.46-oracle 5.15.111)\n\x00'
DEBUG    volatility3.framework.automagic.symbol_finder: Using symbol library: file:///C:/ioc/volarm/volatility3/volatility3/symbols/ubuntu_arm_5.15.0-1040.json

DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!assoc_array_ptr
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!netns_ipvs
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!mtd_info
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!can_pkg_stats
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!can_rcv_lists_stats
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!can_dev_rcv_lists
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!mpls_route
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!sctp_mib
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!smc_stats_rsn
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!smc_stats
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!dn_dev
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!garp_port
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!macsec_ops
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!mctp_dev
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!mpls_dev
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!mrp_port
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!tipc_bearer
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!udp_tunnel_nic
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!pcpu_dstats
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!phylink
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!cfg80211_conn
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!cfg80211_cached_keys
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!cfg80211_cqm_config
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!cfg80211_internal_bss
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!sfp
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!libipw_device
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!smc_hashinfo
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!dsa_8021q_context


```



Any suggestion? 👅

Abyss-W4tcher:
Hi @garanews, could we discuss about it in Slack DMs, to avoid filling the PR with comments ?

I'll post a summary here if we get this fixed :)


edit : The branch wasn't correctly merged. Be sure to do `git checkout aarch64-support` in your local copy. 😄Analysis was successful after the merge.

k1nd0ne:
Hello, 

Thank you @Abyss-W4tcher for this great work. 

In case people are also working on creating a `WindowsAArch64` translation layer and support for windows automagics, here is a sample raw memory image from a Windows 11 ARM 64 Bit along with it's ISF : https://drive.proton.me/urls/GCBR4QMHC4#9Z9pXLnEH4cl

Metadata about the source of the image: 
```
Windows 11 Pro - Build 22621.ni_release.220506-1250
VMWare Fusion with Apple M1 Pro Guest Machine.
```

Extraction method: 
- Windows was put into hibernation.
- hiberfil.sys was extracted after mounting the vmdk offline
- hiberfil.sys was converted successfully to a raw memory image using #1036 

Best regards.

EDIT: 

Some interesting information:

 - The `_KPROCESSOR_STATE` structure [is saved inside the original hibernation file (starting at offset 0x00001000)](https://www.cct.lsu.edu/~golden/Papers/sylvehiber.pdf)
 - Inside this structure we can identify the `_KARM64_ARCH_STATE` structure which seems to contains the required information to fill the ARM layer requirements (Tcr_El1, Ttbr0_El1, ...) **for this specific memory image**.
 
By extracting these data, we _**should**_ be able to fill the requirements and test the arm layer...

Here are the information extracted using this methodology: 

```
Variable	Value
Signature	HIBR
Comment	The hibernation file header signature is correct.
PageSize	4096
SystemTime	2024-06-21 05:17:40+00:00
FirstBootRestorePage	15
NumPageForLoader	22528

-------------_KPROCESSOR_STATE---------------
T0SZ: 	17
T1SZ: 	17
EPD0: 	0
Ttbr0_El1 : 	0x400000802a9000
Ttbr1_El1 : 	0x400000802a9800
```

_Work in progress, the values might not be accurate_

Abyss-W4tcher:
Hello, quick update :

- linux.py
	- Each Intel and AArch64 stackers were separated in dedicated classes ;
	- `virtual_to_physical_address` was moved in `LinuxIntelSubStacker`, as it is not relevant for other architectures (ex: offsets aren't constant accross kernels in AArch64) ;
	- `find_aslr` was unified, by removing the specific Intel `virtual_to_physical_address` operation, which was unnecessary as the KASLR shift can be applied to virtual addresses directly ;
	- CPU registers access was implemented for AArch64, which is needed to determine later if an optional specific CPU feature is ON or OFF. As it depends on ISF debug symbols, the lack of information won't prevent the layer to stack.

My will here was to setup a modular environmenent, for additional CPU architectures, which also answers some of the previous comments. Some "core" changes were operated, but I consider those to be necessary. The current implementation implicitely assumes Intel operations and calculations, which is not suitable for scalability.

- arm.py
	- Introduce LinuxMixIn layers ;
	- Implement page dirty state detection, for both hardware and software (Linux) ;
	- Add the `cpu_registers` requirement, which is a string serialized dict (calculated by automagic). This allows to pass a dict requirement without the need to add a new class in the core.

This should help @k1nd0ne kindone for the hardware dirty state management, even though the Windows software implementation is needed for completion (and even to superseed the hardware implementation if needed). The MixIn structure allows to add specific Windows functions and overrides too.

---

Even if this might need (a lot of) time to review, I tried to document every step to speed up the comprehension.

volatility3/framework/automagic/linux.py
ikelos: We do have a vmscan plugin, and in the future will likely generate a config file that will allow virtual machine guests to have volatility plugins run on them.  This likely either by Intel on Intel, or IntelVM (doesn't exist yet) on Intel.  The point is, it is a capability we'd like to leave open for the future, so planning for it now is useful.

volatility3/framework/automagic/linux.py
```diff
@@ -59,65 +61,97 @@ def stack(
...
 
-                layer = layer_class(
-                    context,
-                    config_path=config_path,
-                    name=new_layer_name,
-                    metadata={"os": "Linux"},
-                )
-                layer.config["kernel_virtual_offset"] = aslr_shift
+                linux_arch_stackers = [LinuxIntelSubStacker, LinuxAArch64SubStacker]
+                for linux_arch_stacker in linux_arch_stackers:
```
Abyss-W4tcher, line 95: Here, the stacking is operated blindly. However, we could also rely on symbols to determine the architecture to stack on. 

I identified `idmap_pg_dir` to be arm64 exclusive, and present since the first 3.7 release. Relying on a symbol, avoids "forced exception rounds", which shouts about missing symbols when an AArch64 image is first passed to the Intel stacker.

Panchajanya1999:
Hi, 

i have trying to get volatility3 working for arm64 android and till now I got it partially working on a physical Google Pixel 6A and LAVA Agni 2. 

From the list of unstable plugins, till now I have got `lsof` and `elfs` working properly. I tried rewriting `netstat` but still no luck.

Here is my repo, feel free to look into it!
https://github.com/Panchajanya1999/volatility3-arm64

Thank You! Will update this thread if i find something interesting to share.

[ oh btw, i used LiME to get proper ramdumps from these two devices ]

here are sample files to have a look
[lsof.txt](https://github.com/user-attachments/files/19546605/lsof.txt)
[elfs.txt](https://github.com/user-attachments/files/19546612/elfs.txt)

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,915 @@
...
+            requirements.IntRequirement(
+                name="kernel_virtual_offset", optional=True, description="ASLR offset"
+            ),
+            requirements.StringRequirement(
+                name="kernel_banner",
+                optional=True,
+                description="Kernel unique identifier, including compiler name and version, kernel version, compile time.",
+            ),
+            requirements.IntRequirement(
+                name=AArch64RegMap.TCR_EL1.__name__,
```
ikelos, line 676: Dynamically named options are risky, because a change to that class could cause an API change to this class (because the input options would change name).  I'd much prefer these to take string names or in a worst case scenario, look them up in constants.

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,915 @@
...
+        """
+        lookup_table = {
+            0b01: 16,  # 16kB
+            0b10: 4,  # 4kB
+            0b11: 64,  # 64kB
+        }
+        return cls._table_lookup(value, lookup_table, reverse_lookup)
+
+
+def set_reg_bits(value: int, reg_field: Enum, reg_value: int = 0) -> int:
```
ikelos, line 876: Since this seems to relate only to `AArch64RegFieldValues`, could this be made a method of it (a staticmethod or classmethod would be fine), rather than floating off on it's own?

volatility3/framework/layers/arm.py
ikelos: Ok, so... it looks like we've got the `page_map_offet` (for non kernel spaces) and then a separate option requirement that contains the register pointing to the kernel layer's page table structure?  That's reasonably elegant, and I guess for the kernel space you could use the same value for both the page_table and the TTBR1.  Still trying to think through the ramifications, but so far it seems ok.

I'm wondering why you picked `page_map_offset` for the config option, rather than `TTBR0` or something for clarity?  If it's something that's expected in other places in the code, then I might want to review those places, because I don't think layer are allowed to depend on requirements of a generic layer (and if we want to have that value exposed in all layers, which doesn't make sense because not every layer has a page table) then it should be done with an interface that exposes a read-only property and have that applied to both intel and arm.

I understand we're treating them as two different layers, but I can't tell whether conceptually that makes sense.  On the one hand, both registers can change indepdendently of each other (in theory).  On the other hand, a process label object may well want to read data from the kernel layer, presumably?

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,915 @@
...
+        Example with granule = 4 kB without 52 bits :
+            (47,12)
+        Example with granule = 16 kB and 52 bits :
+            (49,14)
+        """
+
+        # [1], see D8.3, page 5852
+        return (
+            49 if ttb_granule in [4, 16] and is_52bits else 47,
+            ttb_lookup_indexes[-1][1],
```
ikelos, line 204: I get this is trying to get the tail of the list, but it could do with a comment just to say that's what's going on.     This also doesn't match the signature type, which says `ttb_lookup_indexes` is supposed to be `int` whereas `_determine_ttb_lookup_indexes` which is passed into it, returns `List[Tuple[int]]`?  Could we type it properly and add [Napolean](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/) style docstrings throughout the whole class to give a description for what each argument is please?

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,915 @@
...
+        Returns the translated address, the maximum offset within the block and the page
+        the translated address lives in
+        """
+        base_layer = self.context.layers[self._base_layer]
+
+        # [1], see D8.2.4, page 5825
+        ttb_selector = self._mask(virtual_offset, 55, 55)
+
+        # Check if requested address belongs to the virtual memory space context
+        if ttb_selector != self._virtual_addr_space:
```
ikelos, line 250: So this basically says "if we're in the process space and ask for a kernel address, fail immediately".  I get the simplification, but I feel as though that's quite limiting?  Given we're supposed to know where the kernel layer table (TTBR1 reg value) is supposed to be, could we just swap that in for `table_address` and translate it correctly here?  Throwing an error feels like it will make every consumer have to code in support for a double lookup.  I'm not sure the volatility use needs to know the difference (although there should be a convenience function on the layer to allow users to figure out which table an address is part of).

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,915 @@
...
+                if not self._context.layers[layer_name].is_valid(mapped_offset):
+                    raise exceptions.InvalidAddressException(
+                        layer_name=layer_name, invalid_address=mapped_offset
+                    )
+            except exceptions.InvalidAddressException:
+                if not ignore_errors:
+                    raise
+                return None
+            yield offset, length, mapped_offset, length, layer_name
+            return None
```
ikelos, line 426: Is this necessary, or just a copy-pasta from the intel layer?  I assume it's to deal with a `not explicitly returning None` or some such, but I don't think it's necessary (in either layer), since we know length is 0, so the whole following while loop never gets executed.  If you fix this up, it might be worth taking out of the intel layer too in order to keep them in sync...

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,915 @@
...
+                length -= chunk_size
+                offset += chunk_size
+
+    def _get_virtual_addr_range(
+        self,
+    ) -> Tuple[int]:
+        """Returns the virtual address space range for the current context (user or kernel space)"""
+
+        # [2], see source/arch/arm64/include/asm/memory.h#L62
+        if self._virtual_addr_space == 0:
```
ikelos, line 465: This may go away, if we have both kernel and user living in the same single layer.  Still open for debate at the moment, but reducing special handling like this leans a little toward the "everything in a single layer" model...

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,915 @@
...
+        translated address lives in and the layer_name that the address
+        lives in
+        """
+        table_address, position, _ = self._translate_entry(virtual_offset)
+        offset_within_page = self._mask(virtual_offset, position - 1, 0)
+        physical_offset = table_address + offset_within_page
+
+        return physical_offset, 2**position, self._base_layer
+
+    def _translate_entry(self, virtual_offset: int) -> Tuple[int, int, int]:
```
ikelos, line 239: Since this is an internal function, and we get to decide its output type, could we make it a namedtuple, so that we can access values by name, rather than having random indices all over the place?  I think it'll help readability.  I'd suggest it for `mapping` too, but that's a public interface.  Mebbe one day we'll do it for the intel layer too...

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,915 @@
...
+            # DBM == 1 and AP == 0 -> HW dirty state
+            return bool((entry & (1 << 51)) and not (entry & (1 << 7)))
+        else:
+            raise NotImplementedError(
+                "Hardware updates to Access flag and Dirty state in translation tables are not available. Please try using a software based implementation of dirty state management."
+            )
+
+    @property
+    @functools.lru_cache()
+    def page_shift(self) -> int:
```
ikelos, line 530: I haven't checked, but just a note that inside this class, we should always directly use `_page_size_in_bits`, otherwise we're just adding to unnecessary calls into the call stack for no gain.  The only reason to keep this around is to provide a read-only means of accessing the data (which can be useful).

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,915 @@
...
+                """
+                Contiguous pages might not be mapped, but if we specifically ignore errors, we still want to read the n + 1 page as it might be mapped.
+                """
+                if not ignore_errors:
+                    raise
+                # We can jump more if we know where the page fault occured
+                if isinstance(excp, exceptions.PagedInvalidAddressException):
+                    mask = (1 << excp.invalid_bits) - 1
+                else:
+                    mask = (1 << self.page_shift) - 1
```
ikelos, line 450: This is more expensive than just accessing `_page_size_in_bits`.

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,915 @@
...
+                f"Virtual {hex(virtual_offset)} lives in page frame {hex(table_address)} at offset {hex(self._mask(virtual_offset, low_bit-1, 0))} with descriptor {hex(descriptor)}",
+            )
+
+        return table_address, low_bit, descriptor
+
+    @functools.lru_cache(1025)
+    def _get_valid_table(self, base_address: int) -> Optional[bytes]:
+        """Extracts the translation table, validates it and returns it if it's valid."""
+        table = self._context.layers.read(
+            self._base_layer, base_address, self.page_size
```
ikelos, line 337: This adds two function calls to what would otherwise just be a variable lookup.  Please use `self._page_size`.

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,915 @@
...
+
+        Example return value :
+         {'TCR_EL1.TG1': 3, 'TCR_EL1.T1SZ': 12, 'TCR_EL1.TG0': 1,
+          'TCR_EL1.T0SZ': 12, 'TTBR1_EL1.ASID': 0, 'TTBR1_EL1.BADDR': 1092419584,
+          'TTBR1_EL1.CnP': 0}
+        """
+
+        masked_trees = {}
+        for mm_cls_name, mm_cls in inspect.getmembers(AArch64RegMap, inspect.isclass):
+            if issubclass(mm_cls, Enum) and mm_cls_name in registers_values.keys():
```
ikelos, line 595: Probably want an `else` clause with a `vollog` saying why it got there instead of doing what was expected.

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,915 @@
...
+                    high_bit, low_bit = field.value
+                    masked_value = cls._mask(reg_value, high_bit, low_bit)
+                    masked_trees[dot_joined] = masked_value
+        return masked_trees
+
+    def canonicalize(self, addr: int) -> int:
+        """Canonicalizes an address by performing an appropiate sign extension on the higher addresses"""
+        if self._bits_per_register <= self._ttb_bitsize:
+            return addr & self.address_mask
+        elif addr < (1 << self._ttb_bitsize - 1):
```
ikelos, line 608: Why have this elif?  This isn't getting the canonical prefix it might otherwise expect?  Does arm even have address canonicalization?  There have been arm reference comments everywhere else, but I can't see one here?

volatility3/framework/layers/arm.py
```diff
@@ -0,0 +1,915 @@
...
+        """
+        if addr < (1 << self._ttb_bitsize - 1):
+            return addr
+        return addr ^ self._canonical_prefix
+
+    @staticmethod
+    def _mask(
+        value: int, high_bit: int, low_bit: int, shift: Optional[int] = None
+    ) -> int:
+        """Returns the bits of a value between highbit and lowbit inclusive."""
```
ikelos, line 625: I suspect this is identical to intel.  Fiven there's only two, it's probably worth keeping them in their own classes, but we might want to figure out what shared location we *would* move them to, if we were going to?

ikelos (commented):
Ok, finally found time to go through this.  It's actually in a really good place, and I'm sorry I couldn't find the time before to go through it all.  As it is, I've only really gone through the arm layer, and not any of the `automagic/linux.py` changes yet.  I couldn't find any kind of generic arm stacker, which I see is because we're using info from the linux layer to figure out the various values for ARM registers, much like we do with intel.  I'll review that another time (and actually, I'd almost like to the get the arm layer in without the stacker first, just so it's available, and so everything doesn't sit waiting on the stacker.  That would also allow people to share around config files to load up arm layers even though automagic can't do it directly yet...

I will try and get to the linux stacker stuff soon, hopefully in a couple of weeks, but no promises (sadly).  This stuff seems to take me a long time these days...

PR #1092, opened 2024-01-31
Stable

ikelos:
Merge stable back into develop


ikelos:
Resyncs stable back up to develop (mostly black changes, but also a minor removal of the renamed envvars plugin, which shouldn't have hit stable, and the old one worked in the most recent develop too, so... should be fine?

Issue #1089, opened 2024-01-22
linux: `module_layout` is replaced by `module_memory` as of 6.4


vobst:
Currently Volatility fails to determine base addresses and sizes of modules on >= 6.4 kernels. This affects the `lsmod`, `keyboard_notifiers`, and `tty_check`.

For `lsmod`:
```
Traceback (most recent call last):
  File "/opt/vol/volatility3/vol.py", line 10, in <module>
    volatility3.cli.main()
  File "/opt/vol/volatility3/volatility3/cli/__init__.py", line 790, in main
    CommandLine().run()
  File "/opt/vol/volatility3/volatility3/cli/__init__.py", line 447, in run
    renderers[args.renderer]().render(constructed.run())
  File "/opt/vol/volatility3/volatility3/cli/text_renderer.py", line 193, in render
    grid.populate(visitor, outfd)
  File "/opt/vol/volatility3/volatility3/framework/renderers/__init__.py", line 245, in populate
    for level, item in self._generator:
  File "/opt/vol/volatility3/volatility3/framework/plugins/linux/lsmod.py", line 63, in _generator
    mod_size = module.get_init_size() + module.get_core_size()
  File "/opt/vol/volatility3/volatility3/framework/symbols/linux/extensions/__init__.py", line 40, in get_init_size
    raise AttributeError(
AttributeError: module -> get_init_size: Unable to determine .init section size of module
```

For `keyboard_notifiers`:
```
Traceback (most recent call last):
  File "/opt/vol/volatility3/vol.py", line 10, in <module>
    volatility3.cli.main()
  File "/opt/vol/volatility3/volatility3/cli/__init__.py", line 790, in main
    CommandLine().run()
  File "/opt/vol/volatility3/volatility3/cli/__init__.py", line 447, in run
    renderers[args.renderer]().render(constructed.run())
  File "/opt/vol/volatility3/volatility3/cli/text_renderer.py", line 193, in render
    grid.populate(visitor, outfd)
  File "/opt/vol/volatility3/volatility3/framework/renderers/__init__.py", line 245, in populate
    for level, item in self._generator:
  File "/opt/vol/volatility3/volatility3/framework/plugins/linux/keyboard_notifiers.py", line 42, in _generator
    handlers = linux.LinuxUtilities.generate_kernel_handler_info(
  File "/opt/vol/volatility3/volatility3/framework/symbols/linux/__init__.py", line 319, in generate_kernel_handler_info
[...]

PR #1093, opened 2024-02-02
Linux: fix issue 1089 module memory

eve-mem:
Hello all,

This is an attempt to fix this issue https://github.com/volatilityfoundation/volatility3/issues/1089 raised by @vobst.

@vobst kindly points out the change in the module struct which happens in [this commit on linux](https://github.com/torvalds/linux/commit/ac3b43283923440900b4f36ca5f9f0b1ca43b70e#diff-1a8b5118e6eaf5be44fdf51f9bceaf816aec6d96e4a048b8455bcfbca6a680f2) in the issue which saved me a lot of time - thank you.

There are a number of functions in the module extension that need to use a new enum `mod_mem_type`. Rather than have each one attempt to get the enum I added the `_get_mod_mem_type` function so that it only needs to be accessed once and then the results are stored in the module class in vol. It 'works' but I'm not sure if this is the best way to do it? Happy to take suggestions. 

Size also gets broken down a little more in the new structures I've combined them all to match the output of lsmod, but it could be broken down more if that would be helpful. 

These changes alter `get_module_base`, `get_init_size`, `get_module_core`, and `get_module_init` - I wanted to check everywhere they're used. 
- `get_init_size` is used in the linux lsmod, plugin and I think this works correctly
- `get_module_base` is used in `mask_mods_list`, which then I don't see being used itself in the framework
- I don't see `get_module_core` nor `get_module_init` being used in the framework

They could be used by other plugins out there but I'm not really able to test well. 

Here is a sample output showing the core plugins now work on the newer samples:

Banners showing the newer kernel.
```
$ python vol.py -f ubuntu-6.5.elf banners
Volatility 3 Framework 2.5.2
Offset  Banner

0x115820c98     Linux version 6.5.0-15-generic (buildd@bos03-amd64-040) (x86_64-linux-gnu-gcc-12 (Ubuntu 12.3.0-1ubuntu1~22.04) 12.3.0, GNU ld (GNU Binutils for Ubuntu) 2.38) #15~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Jan 12 18:54:30 UTC 2 (Ubuntu 6.5.0-15.15~22.04.1-generic 6.5.3)
0x1bda00100     Linux version 6.5.0-15-generic (buildd@bos03-amd64-040) (x86_64-linux-gnu-gcc-12 (Ubuntu 12.3.0-1ubuntu1~22.04) 12.3.0, GNU ld (GNU Binutils for Ubuntu) 2.38) #15~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC  (Ubuntu 6.5.0-15.15~22.04.1-generic 6.5.3)
0x1bdb7e2e0     Linux version 6.5.0-15-generic (buildd@bos03-amd64-040) (x86_64-linux-gnu-gcc-12 (Ubuntu 12.3.0-1ubuntu1~22.04) 12.3.0, GNU ld (GNU Binutils for Ubuntu) 2.38) #15~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Jan 12 18:54:30 UTC 2 (Ubuntu 6.5.0-15.15~22.04.1-generic 6.5.3)
0x1bfd7ce00     Linux version 6.5.0-15-generic (buildd@bos03-amd64-040) (x86_64-linux-gnu-gcc-12 (Ubuntu 12.3.0-1ubuntu1~22.04) 12.3.0, GNU ld (GNU Binutils for Ubuntu) 2.38) #15~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Jan 12 18:54:30 UTC 2 (Ubuntu 6.5.0-15.15~22.04.1-generic 6.5.3)3)
```

lsmod:
```
$ python vol.py -f ubuntu-6.5.elf linux.lsmod
Volatility 3 Framework 2.5.2
Offset  Name    Size

$ python vol.py -f ubuntu-6.5.elf linux.lsmod
Volatility 3 Framework 2.5.2
Offset  Name    Size

0xffffc06e1d00  tls     151552
0xffffc060a8c0  isofs   61440
0xffffc05fd4c0  binfmt_misc     24576
0xffffc0586140  nls_iso8859_1   12288
0xffffc06c52c0  aesni_intel     356352
0xffffc066a100  crypto_simd     16384
0xffffc065e200  bochs   20480
0xffffc06513c0  drm_vram_helper 24576
0xffffc0644340  cryptd  24576
0xffffc0547080  drm_ttm_helper  12288
0xffffc0626c40  ttm     110592
0xffffc0595880  joydev  32768
0xffffc0552180  input_leds      12288
0xffffc05cf940  drm_kms_helper  270336
0xffffc0560340  serio_raw       20480
0xffffc05561c0  vmgenid 12288
0xffffc054c200  mac_hid 12288
0xffffc047f400  qemu_fw_cfg     24576
0xffffc05c8340  sch_fq_codel    24576
0xffffc05bc100  msr     12288
0xffffc05a8d80  parport_pc      53248
0xffffc05984c0  ppdev   24576
<snip>
```

tty_check
```
$ python vol.py -f ubuntu-6.5.elf linux.tty_check
Volatility 3 Framework 2.5.2
Name    Address Module  Symbol

tty2    0xffffb6a770f0  __kernel__      n_tty_receive_buf
tty6    0xffffb6a770f0  __kernel__      n_tty_receive_buf
```

keyboard_notifiers (no results, as expected, but it didn't crash!)
```
$ python vol.py -f ubuntu-6.5.elf linux.keyboard_notifiers 
Volatility 3 Framework 2.5.2
Address Module  Symbol
```

Here is a sample output showing that the plugins still work on older samples:

banners:
```
$ python vol.py -f linux-sample-1.dmp banners
Volatility 3 Framework 2.5.2
Offset  Banner

0x1400070       Linux version 3.2.0-4-amd64 (debian-kernel@lists.debian.org) (gcc version 4.6.3 (Debian 4.6.3-14) ) #1 SMP Debian 3.2.57-3+deb7u2
```

lsmod:
```
$ python vol.py -f linux-sample-1.dmp linux.lsmod
Volatility 3 Framework 2.5.2
Offset  Name    Size

0xffffa0426260  rfcomm  33700
0xffffa03ad1e0  bnep    17567
0xffffa0323110  binfmt_misc     12957
0xffffa045f1d0  nfsd    216170
0xffffa0411580  nfs     308353
0xffffa02c9000  nfs_acl 12511
0xffffa03cd240  auth_rpcgss     37143
0xffffa03a7370  fscache 36739
0xffffa03c28f0  lockd   67306
0xffffa03942a0  sunrpc  173730
0xffffa034f260  loop    22641
0xffffa03488c0  btusb   17502
0xffffa036d9e0  bluetooth       119455
0xffffa0342340  rfkill  19012
0xffffa03371c0  coretemp        12898
0xffffa024f170  crc32c_intel    12747
0xffffa033c2f0  ghash_clmulni_intel     13130
0xffffa0331100  aesni_intel     50667
0xffffa0317120  aes_x86_64      16843
0xffffa031e2a0  parport_pc      22364
<snip>
```

tty_check:
```
$ python vol.py -f linux-sample-1.dmp linux.tty_check
Volatility 3 Framework 2.5.2
Name    Address Module  Symbol

tty1    0xffff8122b376  __kernel__      n_tty_receive_buf
tty2    0xffff8122b376  __kernel__      n_tty_receive_buf
tty3    0xffff8122b376  __kernel__      n_tty_receive_buf
tty4    0xffff8122b376  __kernel__      n_tty_receive_buf
tty5    0xffff8122b376  __kernel__      n_tty_receive_buf
tty6    0xffff8122b376  __kernel__      n_tty_receive_buf
tty7    0xffff8122b376  __kernel__      n_tty_receive_buf
```

keyboard_notifiers:
```
$ python vol.py -f linux-sample-1.dmp linux.keyboard_notifiers 
Volatility 3 Framework 2.5.2
Address Module  Symbol
```

Lastly thanks to @Abyss-W4tcher for the https://github.com/Abyss-W4tcher/volatility3-symbols repo - it had exactly the ISF I needed saving me a bunch of time!


volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -26,43 +26,89 @@
...
+                    f"Unable to find mod_mem_type enum. This is expected on kernels <6.4 but may cause issues with later kernels"
+                )
+                self.mod_mem_type = None
+
     def get_module_base(self):
-        if self.has_member("core_layout"):
-            return self.core_layout.base
+        self._get_mod_mem_type()
+        if self.mod_mem_type:
+            return self.mem[self.mod_mem_type["MOD_TEXT"]].base
```
ikelos: This is an Enum, so we should be able to ensure these values are always present.  The question is *just in case* do we trust the enum values never to be renamed/removed, do we use a get with a default we choose or do we put in a try/catch for a lookup error (and in both those cases what default do we return)?  5:S

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -26,43 +26,89 @@
 
 
 class module(generic.GenericIntelProcess):
+    def _get_mod_mem_type(self):
```
ikelos: This doesn't seem quite the right way to deal with this?  Feels like a good place for a property (which can then read from a memoized value after the first call).  It saves on having to remember to make the first call and guarantees the attribute will always be there.

ikelos (commented):
Thanks, I'd swap the get method to an attribute but otherwise this looks reasonable for what it does.  Just a little error checking question after that...

volatility3/framework/symbols/linux/extensions/__init__.py
eve-mem: Great idea - I've updated it to use a property. Makes a lot more sense, looks cleaner to me. Have I done the memorized value side of it the right way?

volatility3/framework/symbols/linux/extensions/__init__.py
eve-mem: Another good idea. At first I wanted to use an UnparsableValue from the renderers - but seeing that before it would raise an attribute error if it couldn't find the right values I thought that made more sense? It feels like if the kernel changes the enum enough that the parts we want are not there - than we probably need to update something rather than just a default value? What do you think? It feels like raising the attribute error would highlight exactly where the problem is.

eve-mem:
Hey @ikelos  - I've updated based on those comments - what do you think?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -26,43 +26,113 @@
 
 
 class module(generic.GenericIntelProcess):
+    @property
+    def mod_mem_type(self):
+        """Return the mod_mem_type enum choices if available or None if not"""
+        if not self.has_member("_mod_mem_type"):
```
ikelos: This shouldn't be checking for existence, the `_mod_mem_type` should be defined as None in the initializer technically.  This should be a test of the value to determine if it's been set before or not.

Given that this is just looking up an enumeration, it *shouldn't* be necessary to ever look it up again so it might be inefficient to test for `is None` because we might rerun the check code even after we know it's not possible, however, since it's a dictionary, we can return an empty dictionary which won't pass an `is None` test (allowing the memoization to shortcircuit) but will fail an `if blah.mod_mem_type:` test.  Also, the memoized variable (`_mod_mem_type`) should be initialized in the initializer...

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -26,43 +26,113 @@
...
+            try:
+                vmlinux = linux.LinuxUtilities.get_module_from_volobj_type(
```
ikelos: This seems a very long winded way of achieving the goal?  I'd have expected `context.get_enumeration(self.get_symbol_table_name + constants.BANG + "mod_mem_type")` to work, or even `context.get_enumeration("mod_mem_type")`?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -26,43 +26,113 @@
...
+                    self._context, self
+                )
+                # mod_mem_type and module_memory were added in kernel 6.4 which replaces
+                # module_layout for storing the information around core_layout etc.
+                # see commit ac3b43283923440900b4f36ca5f9f0b1ca43b70e for more information
+                self._mod_mem_type = vmlinux.get_enumeration("mod_mem_type").choices
+            except exceptions.SymbolError:
+                vollog.debug(
+                    f"Unable to find mod_mem_type enum. This is expected on kernels <6.4 but may cause issues with later kernels"
```
ikelos: This message isn't all that clear, do expect that `mod_mem_type` *doesn't* exist on kernels <6.4?  If so we should more clearly say "This error can be ignored for kernels < 6.4" or something of that nature.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -26,43 +26,113 @@
...
+                )
+                self._mod_mem_type = None
+        return self._mod_mem_type
+
     def get_module_base(self):
-        if self.has_member("core_layout"):
-            return self.core_layout.base
+        if self.mod_mem_type:
```
ikelos: Something about this just seems ugly, but I'm not sure of a better way to do it?  I was thinking of something involving default lookups from `mod_mem_type` or from `mem`, but I can't figure a good way of getting the right results returned more efficiently, so this is probably the right way of doing it (even if something about it doesn't quite sit right to me).  5;P

ikelos (commented):
Looking pretty good, still just a couple more pointers at the moment, but definitely closer to what we're after.  5:)

volatility3/framework/symbols/linux/extensions/__init__.py
eve-mem: Ah thanks - I'll give it a go!

volatility3/framework/symbols/linux/extensions/__init__.py
eve-mem: Yeah, that's the message i want to get across. On the old ones this is fine - but on something newer it probably means there is a problem. It might be better to try and read the actual kernel version and then pick a better message depending on what it is.

volatility3/framework/symbols/linux/extensions/__init__.py
eve-mem: I've updated this now to check for `mem` as that's the member which will contain the values. So if that is there it should go down the route of then checking the values from the enum. I think that works better now, it splits out the two problems so it doesn't rely on the `mod_mem_type` being None or an empty dict to choose the right path. (e.g. problem 1 "which kernel version is it, and therefore what member should I read from" and problem 2 "can I extract out the values I need from the enum correctly" being separate)

eve-mem:
Hello @ikelos - I've updated this now. First time using memoization - so hopefully doing it the right way...! 

One, likely dumb, question I have is if I should also update the framework version number at all?

I've updated the logic for choosing the right members which I think works better than how it was before. Let me know what you think. 

e.g. now it is like this:
```python
    def get_module_init(self):
        if self.has_member("mem"):  # kernels 6.4+
            # blah
        elif self.has_member("init_layout"):
            # blah
        elif self.has_member("module_init"):
            # blah
```

rather than how it was before, weirdly accessing this property I'm making to see which way it should flow:

```python
    def get_module_init(self):
        if self.mod_mem_type: # kernels 6.4+
            # blah
        else:
            if self.has_member("init_layout"):
                # blah
            elif self.has_member("module_init"):
                # blah
```

ikelos (approved):
Cool, this looks good, thanks!  5:)

PR #1096, opened 2024-02-07
Linux update mm extension

eve-mem:
Hello 👋 

@gcmoreira pointed out that the current linux pslist plugin used `get_mmap_iter()` when dumping out elfs. This would fail on kernels that use the maple tree structure instead. (First added here https://github.com/volatilityfoundation/volatility3/pull/928)

This PR updates the pslist plugin to use `get_vma_iter` instead. 

It also changes `get_mmap_iter`  and `get_maple_tree_iter` to `_get_mmap_iter` and `_get_maple_tree_iter` to make it more obvious that they should only be directly used if you know what you're doing. 

Thanks!

### DIFF
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -448,6 +448,7 @@ class maple_tree(objects.StructType):
 
 class mm_struct(objects.StructType):
 
+    # TODO: As of version 3.0.0 this method should be removed
     def get_mmap_iter(self) -> Iterable[interfaces.objects.ObjectInterface]:
         """
         Deprecated: Use either get_vma_iter() or _get_mmap_iter().
@@ -478,6 +479,7 @@ class mm_struct(objects.StructType):
             seen.add(link.vol.offset)
             link = link.vm_next
 
+    # TODO: As of version 3.0.0 this method should be removed
     def get_maple_tree_iter(self) -> Iterable[interfaces.objects.ObjectInterface]:
         """
         Deprecated: Use either get_vma_iter() or _get_maple_tree_iter().
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -447,12 +447,14 @@ class maple_tree(objects.StructType):
 
 
 class mm_struct(objects.StructType):
-    def get_mmap_iter(self) -> Iterable[interfaces.objects.ObjectInterface]:
-        """Returns an iterator for the mmap list member of an mm_struct."""
+    def _get_mmap_iter(self) -> Iterable[interfaces.objects.ObjectInterface]:
+        """Returns an iterator for the mmap list member of an mm_struct. Use this only if
+        required, get_vma_iter() will choose the correct _get_maple_tree_iter() or
+        _get_mmap_iter() automatically as required."""
 
         if not self.has_member("mmap"):
             raise AttributeError(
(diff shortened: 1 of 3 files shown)


volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -447,12 +447,14 @@ def _parse_maple_tree_node(
 
 
 class mm_struct(objects.StructType):
-    def get_mmap_iter(self) -> Iterable[interfaces.objects.ObjectInterface]:
-        """Returns an iterator for the mmap list member of an mm_struct."""
+    def _get_mmap_iter(self) -> Iterable[interfaces.objects.ObjectInterface]:
```
ikelos, line 461: This is technically a bigger change in the core API, so would require a version bump (to at least 2.6.0).  However, since this is implemented as a removal of a call, it should be a full version bump so 3.0.0.  To avoid that and only have the smaller version bump, either leave the methods as `get_mmap_iter` or implement a `get_mmap_iter` which simply passes through and adds a deprecation message ([this](https://dev.to/hckjck/python-deprecation-2mof) looks a reasonable guide to follow to do so).  Unfortunately, we have to change the framework version number to track this, and this is why interfaces need careful consideration before being made public...  5;)

ikelos (commented):
Just a point about version bumping, and sadly this will require the whole framework to get bumped because the change is strictly a removal of functionality.

eve-mem:
Yeah... It doesn't feel useful to bump the versions just for this. Before the maple trees existed, only mmap was needed which is why it wasn't private then. Then when adding the maple tree it just followed the current methods. 

I'll update this PR.

ikelos:
Yeah, updating version numbers is a pig, but it's important that we stick to it, otherwise we jeopardize people being able to rely on what we've put in place to make sure everything works (and so they get good error messages when it doesn't).  It was made public, so probably best to keep it public with a deprecation warning.  Hopefully no one will ever hit it, but in case someone does at least we tell them what they need to do.  Sorry it's a bit of a pain...

eve-mem:
No it makes sense to be precious about version numbers - who knows what people have made that might rely on this - better to be safe than sorry. Hopefully this is better now?

I've made the two versions that simply fall through to the generic `get_vma_iter` function, adding a warning using the normal vollog. I'd see some other places in the framework doing similar - what do you think? I can investigate the warnings package if you think it would be useful. 

I haven't bumped the framework version number - should I? Maybe up to 2.5.3 from 2.5.2? 

It would also be nice to somehow flag that when the framework that when it bumps to 3.0.0 then these two can be removed at the same time. I'm not sure what the best way to do that is? What do you think?

ikelos:
Hiya, yep, do please bump the version number, but since it is an additive change, you'll need to bump the MINOR version not just the PATCH number, so it'll be 2.6.0.

I'm in two minds, I don't want users to get panicked by a warning, but at the same time, it shouldn't ever get hit by core, and it will be a good way of letting people know they need to update so I think probably best to keep it as a warning (which will then get seen by anyone if the codepath is ever hit).

Best way of flagging things for removal is to add a `TODO: As of version 3.0.0, remove these methods` or something.  It requires us to be a little bit vigilant, but if they get missed the deprecation warning message should show up and we'll see the comment and know we can remove it, so it'll only litter our codebase for so long...

eve-mem:
Hey @ikelos - I've made those changes now. Let me know if you think there is anything else needed. 🤗

ikelos (approved):
Yep, looks good, thanks!  5:)

PR #1103, opened 2024-02-22
Linux: psscan tiny update so that  memory_layer_name is reused

eve-mem:
Hello - almost not worth anyone's time but I saw this today and felt like I needed to change it. 

In psscan when getting the memory layer it goes via `kernel_layer.dependencies[0]` twice. This is okay enough while the two lines are next to each other - but it might cause a mismatch in the future when it gets split up and multiple "memory_layers" are supported (e.g. ram + swap etc) and we forget to keep them in sync. It made sense to me to just reuse  `memory_layer_name`.

Thanks!


ikelos:
Yep, looks good, thanks for spotting it!  5:D

Issue #565, opened 2021-09-14
Support dumpfiles with offset


wooyin:
After scan file in vmem, it is hard to dump only one file, cause 'FileScan' display offset, but not virtuladdr.
 `python3 vol.py -f test.vmem windows.filescan.FileScan`

I suggest to add 'offset' to support it.
`python3 vol.py -f test.vmem windows.dumpfiles.DumpFiles --offset 0x7fe7fdd0`

iMHLv2:
Hello @wooyin - the dumpfiles plugin already supports dumping files by virtual or physical offset. Is that not working properly?

```
% python3 vol.py windows.dumpfiles -h
Volatility 3 Framework 1.2.1
usage: volatility windows.dumpfiles.DumpFiles [-h] [--pid PID] [--virtaddr VIRTADDR] [--physaddr PHYSADDR]

optional arguments:
  -h, --help           show this help message and exit
  --pid PID            Process ID to include (all other processes are excluded)
  --virtaddr VIRTADDR  Dump a single _FILE_OBJECT at this virtual address
  --physaddr PHYSADDR  Dump a single _FILE_OBJECT at this physical address
```

ikelos:
I think this can be marked as closed?  Feel free to reopen it if you believe your issue is a still a problem...  5:)

ajmeese7:
Doing `--virtaddr` gives me the error `volatility: error: unrecognized arguments: ‑‑virtaddr`

eve-mem:
Hi @ajmeese7 - what is the full command line you use when you get that error?

ajmeese7:
I don't remember now, it was something to the effect of `./vol.py -f memory.dmp -o ./dumped_files windows.dumpfiles.DumpFiles --virtaddr 0x3e1745d0` though. I think I ended up getting my problem resolved by using `--physaddr`, but in the process of debugging I saw this related issue and figured I'd mention that I ran into the same problem.

Crane-Mocker:
> I don't remember now, it was something to the effect of `./vol.py -f memory.dmp -o ./dumped_files windows.dumpfiles.DumpFiles --virtaddr 0x3e1745d0` though. I think I ended up getting my problem resolved by using `--physaddr`, but in the process of debugging I saw this related issue and figured I'd mention that I ran into the same problem.

[...]

PR #1104, opened 2024-02-26
windows.dumpfiles --addr for offset

Crane-Mocker:
use `--addr` for offset, instead of `--virtaddr` and `--physaddr`
ref: issue #565

### DIFF
--- a/test/test_volatility.py
+++ b/test/test_volatility.py
@@ -141,7 +141,7 @@ def test_windows_dumpfiles(image, volatility, python):
                 volatility,
                 python,
                 globalargs=["-o", path],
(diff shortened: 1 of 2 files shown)


Crane-Mocker:
With the new commit, `test_volatility.py` was updated.
Local `pytest` test passed.

Originally, I forgot to modify `test_volatility.py` since I tested it with another img. Thus, `test/test_volatility.py::test_windows_dumpfiles[win-xp-laptop-2005-06-25.img] FAILED [ 42%]` due **stderr** `error: unrecognized arguments: --virtaddr 0x820588e8\n`

PR #1106, opened 2024-02-28
Documentation: Fix up syntax errors involving * character

### LINKED ISSUE #1105: syntaxwarning on python 3.12
**Describe the bug**
Running code with python 3.12 is raising syntaxwarning in https://github.com/volatilityfoundation/volatility3/blob/887747e3ceb1d40af936e8a91d65e05874f4c06a/volatility3/framework/interfaces/plugins.py#L63

**Context**
Volatility Version: develop
Operating System:  docker python:3.12-slim image
Python Version:  3.12
Suspected Operating System:  
Command:  

**To Reproduce**
```
$ python
Python 3.11.5 (main, Sep 11 2023, 13:54:46) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.- ()[]\{\}!$%^:#~?<>,|"
>>> 

$ python3.12
Python 3.12.2 (main, Feb 13 2024, 09:28:52) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.- ()[]\{\}!$%^:#~?<>,|"
<stdin>:1: SyntaxWarning: invalid escape sequence '\{'
>>> 
```

**Additional information**
[from stackoverflow](https://stackoverflow.com/a/77531416/680310)
Using invalid escape sequences in string literals has been [deprecated since Python 3.6](https://docs.python.org/dev/whatsnew/3.6.html#deprecated-python-behavior). Since then, attempting to use an invalid escape sequence has emitted a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning). This often goes unnoticed if you never run Python with [warnings enabled](https://docs.python.org/3/using/cmdline.html#cmdoption-W). DeprecationWarnings are silenced by default.

Python 3.12 [upgraded the DeprecationWarning to a SyntaxWarning](https://docs.python.org/3/whatsnew/3.12.html#other-language-changes). SyntaxWarnings are emitted by the compiler when the code is parsed, not when it's being run, so they cannot be ignored using a runtime warning filter. Unlike DeprecationWarnings, SyntaxWarnings are displayed by default, which is why you're seeing it now. This increase in visibility was intentional. In a future version of Python, using invalid escape sequence in string literals is planned to eventually become a hard SyntaxError.

ikelos:
Fixes #1105.

Changes single escaped star characters to double literal backslashes (so that when later interpreted they come out as actual *s in the documentation).  Also fix up some brackets with a similar problem.  All appears to work fine with python 3.11 and python 3.12.  Also included some miscellaneous documentation fixes.

Unless someone shouts, I'll probably commit this in a couple of days.

### DIFF
--- a/volatility3/framework/interfaces/plugins.py
+++ b/volatility3/framework/interfaces/plugins.py
@@ -60,7 +60,7 @@ class FileHandlerInterface(io.RawIOBase):
     @staticmethod
     def sanitize_filename(filename: str) -> str:
         """Sanititizes the filename to ensure only a specific whitelist of characters is allowed through"""
-        allowed = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.- ()[]\{\}!$%^:#~?<>,|"
+        allowed = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.- ()[]{}!$%^:#~?<>,|"
         result = ""
         for char in filename:
             if char in allowed:
--- a/volatility3/framework/plugins/linux/kmsg.py
+++ b/volatility3/framework/plugins/linux/kmsg.py
@@ -198,7 +198,9 @@ class ABCKmsg(ABC):
 class Kmsg_pre_3_5(ABCKmsg):
     """The kernel ring buffer (log_buf) is a char array that sequentially stores
     log lines, each separated by newline (LF) characters. i.e:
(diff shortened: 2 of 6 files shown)


ikelos:
No complaints, and the changes are minor, so merging.

PR #1107, opened 2024-03-01
Mac process memory dump

### LINKED ISSUE #485: Support for process memory dump on Mac
Volatility version 2 has a number of Mac plugins that supported dumping the memory, in version 3 none of the plugins seem to support it.

Please implement process memory dump support for Mac in Volatility version 3.

I am not aware of any current alternatives, v2 helper scripts to generate symbols for latest MacOS kernels are broken, therefore v2 cannot be used anymore.

**Additional information**
N/A - should be clear from description

Abyss-W4tcher:
Hi, 

This PR adds support for Mac process memory dumping. It is widely inspired from https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/proc.py and previous work from https://github.com/volatilityfoundation/volatility3/pull/884 by @eve-mem.

Only a few tweaks from linux.proc were needed, for example on the try-except block wrapping `vm_start`, which might result in a `vm_start = None`, but still continuing the loop and eventually raising an error on `format_hints.Hex(vma.vm_start)`. 

I also added a useful feature for the `--address` list requirements, which allows to pass integers in the `0x` format. This is more convenient than having to convert the hex result from a plugin to an integer. This is useful for any plugin using a ListRequirement.

Tested on `mac-sample-1.bin` from the VolatilityFoundation.

Closes #485

### DIFF
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -827,7 +827,11 @@ class CommandLine:
                 requirement,
                 volatility3.framework.configuration.requirements.ListRequirement,
             ):
-                additional["type"] = requirement.element_type
+                # Allow a list of integers, specified with convenient 0x hexadecimal format
+                if requirement.element_type == int:
+                    additional["type"] = lambda x: int(x, 0)
+                else:
+                    additional["type"] = requirement.element_type
                 nargs = "*" if requirement.optional else "+"
                 additional["nargs"] = nargs
             elif isinstance(
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -827,11 +827,7 @@ class CommandLine:
                 requirement,
                 volatility3.framework.configuration.requirements.ListRequirement,
             ):
-                # Allow a list of integers, specified with convenient 0x hexadecimal format
-                if requirement.element_type == int:
-                    additional["type"] = lambda x: int(x, 0)
-                else:
-                    additional["type"] = requirement.element_type
+                additional["type"] = requirement.element_type
                 nargs = "*" if requirement.optional else "+"
                 additional["nargs"] = nargs
             elif isinstance(
--- a/volatility3/framework/plugins/mac/proc_maps.py
+++ b/volatility3/framework/plugins/mac/proc_maps.py
@@ -2,17 +2,23 @@
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
 
-from volatility3.framework import renderers, interfaces
+from volatility3.framework import renderers, interfaces, exceptions
 from volatility3.framework.configuration import requirements
 from volatility3.framework.objects import utility
 from volatility3.framework.renderers import format_hints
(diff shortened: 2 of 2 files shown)


eve-mem:
Thanks for that list of ints change, that one had been on my mind for a while.

volatility3/cli/__init__.py
```diff
@@ -827,7 +827,11 @@ def populate_requirements_argparse(
                 requirement,
                 volatility3.framework.configuration.requirements.ListRequirement,
             ):
-                additional["type"] = requirement.element_type
+                # Allow a list of integers, specified with convenient 0x hexadecimal format
```
ikelos: I'd kind of prefer fixes like this that affect everything to be separate pull requests if that's ok?  Just makes it easier to track down changes and all the rest of the code to stand or fall on its own.  This bit would get committed first, but essentially it affects more than just the mac process memory dump stuff.

volatility3/framework/plugins/mac/proc_maps.py
```diff
@@ -31,30 +37,188 @@ def get_requirements(cls):
...
+            # filter for any vm_start that matches the supplied address config
+            def vma_filter_function(task: interfaces.objects.ObjectInterface) -> bool:
+                addrs_in_vma = [
+                    addr
+                    for addr in address_list
+                    if task.links.start <= addr <= task.links.end
+                ]
+
+                # if any of the user supplied addresses would fall within this vma return true
+                if addrs_in_vma:
```
ikelos: This could probably just `return bool(addrs_in_vma)`, but happy enough if you want it to be more explicit.

ikelos (commented):
Cool, looks good with a couple stylistic points but no showstoppers.  The only other question I have is given how similar the code is the linux dumper, should we try to unify them to a single code base, or would that complicate things (or do they actually work differently under the hood)?

Abyss-W4tcher:
Hi, this is 95% close to the linux code honestly. As it was previously reviewed I thought there was no reason to re-implement the wheel.

Only things changing are objects and members names between Linux and Mac. I think this could be unified, in another PR, but if we need to add some specific OS features in the future it might complexify things ?

volatility3/cli/__init__.py
Abyss-W4tcher: Sure, I will open a specific PR !

ikelos:
Yeah, fair enough.  Could you throw in a comment saying this is almost identical code, maybe starting `# TODO` or `# FIXME` please, just so we don't lose sight (or if one of them gets a fix that the other then doesn't)...  5:)

volatility3/framework/plugins/mac/proc_maps.py
Abyss-W4tcher: Done !

ikelos (approved):
Yep, looks reasonable now.  Still want to keep in mind potentially syncing up the dump commands between *nix and macos, but I guess we'll see if they desynchronize in the future...

PR #1110, opened 2024-03-05
Windows: Add support for missing callback types

dgmcdona:
This PR updates the windows.callbacks.Callbacks plugin to support callback types that were present in the original volatility framework but have not yet been added to volatility3. These callback types include:

- `IoRegisterShutdownNotification`
- `IoRegisterFsRegistrationChange`
- `GenericKernelCallback`
- `EventCategoryHardwareProfileChange`
- `EventCategoryDeviceInterfaceChange`
- `EventCategoryTargetDeviceChange`
- `DbgSetDebugPrintCallback`

This required updates to the callbacks JSON symbol files, the creation of a `_SHUTDOWN_PACKET` extension, and updates to the plugin itself. Because it introduces three new requirements (handles, driverirp, and poolscanner), I have incremented the major version number for the callbacks plugin. No other plugins depend on the callbacks plugin at this time, so it was not necessary to increase version numbers in other plugins.

### DIFF
--- a/volatility3/framework/plugins/windows/callbacks.py
+++ b/volatility3/framework/plugins/windows/callbacks.py
@@ -2,16 +2,24 @@
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
 
-import logging
 import contextlib
-from typing import List, Iterable, Tuple, Optional, Union
+import logging
+from typing import Dict, Iterable, List, Optional, Tuple, Union, cast
 
-from volatility3.framework import constants, exceptions, renderers, interfaces, symbols
+from volatility3.framework import (
+    constants,
+    exceptions,
+    interfaces,
+    objects,
+    renderers,
+    symbols,
+)
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints
 from volatility3.framework.symbols import intermed
 from volatility3.framework.symbols.windows import versions
-from volatility3.plugins.windows import ssdt
+from volatility3.framework.symbols.windows.extensions import callbacks
+from volatility3.plugins.windows import driverirp, handles, poolscanner, ssdt
 
 vollog = logging.getLogger(__name__)
 
@@ -20,7 +28,7 @@ class Callbacks(interfaces.plugins.PluginInterface):
     """Lists kernel callbacks and notification routines."""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 0, 0)
+    _version = (2, 0, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
@@ -33,27 +41,154 @@ class Callbacks(interfaces.plugins.PluginInterface):
             requirements.PluginRequirement(
                 name="ssdt", plugin=ssdt.SSDT, version=(1, 0, 0)
             ),
+            requirements.PluginRequirement(
+                name="poolscanner", plugin=poolscanner.PoolScanner, version=(1, 0, 0)
+            ),
+            requirements.PluginRequirement(
+                name="driverirp", plugin=driverirp.DriverIrp, version=(1, 0, 0)
+            ),
+            requirements.PluginRequirement(
+                name="handles", plugin=handles.Handles, version=(1, 0, 0)
+            ),
         ]
 
-    @staticmethod
-    def create_callback_table(
+    @classmethod
+    def create_callback_scan_constraints(
+        cls,
         context: interfaces.context.ContextInterface,
         symbol_table: str,
-        config_path: str,
-    ) -> str:
-        """Creates a symbol table for a set of callbacks.
+        is_vista_or_above: bool,
+    ) -> List[poolscanner.PoolConstraint]:
+        """Creates a list of Pool Tag Constraints for callback objects.
 
         Args:
             context: The context to retrieve required elements (layers, symbol tables) from
-            symbol_table: The name of an existing symbol table containing the kernel symbols
-            config_path: The configuration path within the context of the symbol table to create
+            symbol_table: The name of an existing symbol table containing the symbols / types
+            is_vista_or_above: A boolean indicating whether the OS version is Vista or newer.
 
         Returns:
-            The name of the constructed callback table
+            The list containing the built constraints.
         """
-        native_types = context.symbol_space[symbol_table].natives
-        is_64bit = symbols.symbol_table_is_64bit(context, symbol_table)
-        table_mapping = {"nt_symbols": symbol_table}
+        constraints = cls._create_default_scan_constraints(context, symbol_table)
+        if is_vista_or_above:
+            constraints += cls._create_scan_constraints_vista_and_above(symbol_table)
+        return constraints
+
+    @staticmethod
+    def _create_default_scan_constraints(
+        context: interfaces.context.ContextInterface, symbol_table: str
+    ) -> List[poolscanner.PoolConstraint]:
+
+        shutdown_packet_size = context.symbol_space.get_type(
+            symbol_table + constants.BANG + "_SHUTDOWN_PACKET"
+        ).size
+        generic_callback_size = context.symbol_space.get_type(
+            symbol_table + constants.BANG + "_GENERIC_CALLBACK"
+        ).size
+        notification_packet_size = context.symbol_space.get_type(
+            symbol_table + constants.BANG + "_NOTIFICATION_PACKET"
+        ).size
+
+        return [
+            poolscanner.PoolConstraint(
+                b"IoFs",
+                type_name=symbol_table + constants.BANG + "_NOTIFICATION_PACKET",
+                size=(notification_packet_size, None),
+                page_type=poolscanner.PoolType.NONPAGED
+                | poolscanner.PoolType.PAGED
+                | poolscanner.PoolType.FREE,
+            ),
+            poolscanner.PoolConstraint(
+                b"IoSh",
+                type_name=symbol_table + constants.BANG + "_SHUTDOWN_PACKET",
+                size=(shutdown_packet_size, None),
+                page_type=poolscanner.PoolType.NONPAGED
+                | poolscanner.PoolType.PAGED
+                | poolscanner.PoolType.FREE,
+                index=(0, 0),
+            ),
+            poolscanner.PoolConstraint(
+                b"Cbrb",
+                type_name=symbol_table + constants.BANG + "_GENERIC_CALLBACK",
+                size=(generic_callback_size, None),
+                page_type=poolscanner.PoolType.NONPAGED
+                | poolscanner.PoolType.PAGED
+                | poolscanner.PoolType.FREE,
+            ),
+        ]
+
+    @staticmethod
+    def _create_scan_constraints_vista_and_above(
+        symbol_table: str,
+    ) -> List[poolscanner.PoolConstraint]:
+        """Creates a list of Pool Tag Constraints for callback objects.
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            symbol_table: The name of an existing symbol table containing the symbols / types
+
+        Returns:
+            The list containing the built constraints.
+        """
+
+        return [
+            poolscanner.PoolConstraint(
+                b"DbCb",
+                type_name=symbol_table + constants.BANG + "_DBGPRINT_CALLBACK",
+                size=(0x20, 0x40),
+                page_type=poolscanner.PoolType.NONPAGED
+                | poolscanner.PoolType.PAGED
+                | poolscanner.PoolType.FREE,
+            ),
+            poolscanner.PoolConstraint(
+                b"Pnp9",
+                type_name=symbol_table + constants.BANG + "_NOTIFY_ENTRY_HEADER",
+                size=(0x30, None),
+                page_type=poolscanner.PoolType.NONPAGED
+                | poolscanner.PoolType.PAGED
+                | poolscanner.PoolType.FREE,
+                index=(1, 1),
+            ),
+            poolscanner.PoolConstraint(
+                b"PnpD",
+                type_name=symbol_table + constants.BANG + "_NOTIFY_ENTRY_HEADER",
+                size=(0x40, None),
+                page_type=poolscanner.PoolType.NONPAGED
+                | poolscanner.PoolType.PAGED
+                | poolscanner.PoolType.FREE,
+                index=(1, 1),
+            ),
+            poolscanner.PoolConstraint(
+                b"PnpC",
+                type_name=symbol_table + constants.BANG + "_NOTIFY_ENTRY_HEADER",
+                size=(0x38, None),
+                page_type=poolscanner.PoolType.NONPAGED
+                | poolscanner.PoolType.PAGED
+                | poolscanner.PoolType.FREE,
+                index=(1, 1),
+            ),
+        ]
+
+    @classmethod
+    def create_callback_symbol_table(
+        cls,
+        context: interfaces.context.ContextInterface,
+        nt_symbol_table: str,
+        config_path: str,
+    ) -> str:
+        """Creates a symbol table for kernel callback objects.
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            nt_symbol_table: The name of the table containing the kernel symbols
+            config_path: The config path where to find symbol files
+
+        Returns:
+            The name of the constructed symbol table
+        """
+        native_types = context.symbol_space[nt_symbol_table].natives
+        is_64bit = symbols.symbol_table_is_64bit(context, nt_symbol_table)
+        table_mapping = {"nt_symbols": nt_symbol_table}
 
         if is_64bit:
             symbol_filename = "callbacks-x64"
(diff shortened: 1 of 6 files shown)


volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -2,16 +2,18 @@
...
-import logging
 import contextlib
-from typing import List, Iterable, Tuple, Optional, Union
+import logging
+from typing import Dict, Iterable, List, Optional, Tuple, Union, cast
 
-from volatility3.framework import constants, exceptions, renderers, interfaces, symbols
+from volatility3.framework import constants, exceptions, interfaces, renderers, symbols
 from volatility3.framework.configuration import requirements
+from volatility3.framework.objects import Pointer, StructType
```
ikelos: We prefer to import `objects` and then use `objects.Pointer` or `objects.StructType`.

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -2,16 +2,18 @@
...
 from volatility3.framework.renderers import format_hints
 from volatility3.framework.symbols import intermed
 from volatility3.framework.symbols.windows import versions
-from volatility3.plugins.windows import ssdt
+from volatility3.framework.symbols.windows.extensions import ObjectInterface, callbacks
```
ikelos: And this is exactly the reason why, ObjectInterface is actually defined in `volatility3.framework.interfaces.objects` but we (mistakenly) imported it directly rather than the objects module.  This could lead to someone importing `callbacks.ObjectInterface`...  Please try hard to only ever import the modules (and the correct modules).  I'll likely fix the `symbols.windows.extensions` module to fix this mistake...

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -20,7 +22,7 @@ class Callbacks(interfaces.plugins.PluginInterface):
     """Lists kernel callbacks and notification routines."""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 0, 0)
+    _version = (2, 0, 0)
```
ikelos, line 31: Bumping this is the right thing to do, even if other modules in the core set don't import it.  Someone else somewhere might, so if there's been a non-backwards compatible change, bumping the major number is exactly the right thing to do.  5:)

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -33,27 +35,163 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
             requirements.PluginRequirement(
                 name="ssdt", plugin=ssdt.SSDT, version=(1, 0, 0)
             ),
+            requirements.VersionRequirement(
+                name="poolscanner", component=poolscanner.PoolScanner, version=(1, 0, 0)
```
ikelos: All of these version requirements check the same thing?  Either the individual components need checking (if they have versions) or just do the check once.

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -65,10 +203,131 @@ def create_callback_table(
...
+
+    @classmethod
+    def _process_scanned_callback(
+        cls, memory_object: StructType, type_map: Dict[int, str]
+    ) -> Tuple[
+        Union[str, interfaces.renderers.BaseAbsentValue],
+        int,
+        Union[str, interfaces.renderers.BaseAbsentValue],
+    ]:
+        _, _, obj_type = memory_object.vol.type_name.rpartition(constants.BANG)
```
ikelos: This is clever, but I think he already have tools for doing it?  Should be that [`memory_object.get_symbol_table_name()`](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/interfaces/objects.py#L149) should do it?  It will also handle exceptional cases where the type may not have a symbol table (rare these days, but I believe still technically possible).

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -65,10 +203,131 @@ def create_callback_table(
...
+        if isinstance(memory_object, callbacks._SHUTDOWN_PACKET):
+            callback_type = "IoRegisterShutdownNotification"
+            try:
+                driver = memory_object.DeviceObject.dereference().DriverObject
```
ikelos: This *shouldn't* require dereferencing?  `DeviceObject.DriverObject` should work.

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -65,10 +203,131 @@ def create_callback_table(
...
+                index = driverirp.MAJOR_FUNCTIONS.index("IRP_MJ_SHUTDOWN")
+                callback_address = driver.MajorFunction[index]
+                details = driver.DriverName.String or renderers.UnparsableValue()
+            except exceptions.InvalidAddressException:
+                callback_address = memory_object.vol.offset
+                details = renderers.NotApplicableValue()
```
ikelos, line 281: Good!  Thanks for using the right return types!  5:)

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -65,10 +203,131 @@ def create_callback_table(
...
+        elif obj_type == "_NOTIFICATION_PACKET":
+            callback_type = "IoRegisterFsRegistrationChange"
+            callback_address = memory_object.NotificationRoutine
+            details = renderers.NotApplicableValue()
+        elif obj_type == "_NOTIFY_ENTRY_HEADER":
+            try:
+                driver = memory_object.DriverObject.dereference()
```
ikelos: Again the dereferencing *should* be unnecessary, although here it seems to be being used to check validity (I believe pointer have an `is_valid` method which would avoid the need for a potentially costly exception setup...

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -107,19 +366,22 @@ def list_notify_routines(
             try:
                 symbol_offset = ntkrnlmp.get_symbol(symbol_name).address
             except exceptions.SymbolError:
-                vollog.debug(f"Cannot find {symbol_name}")
+                # vollog.debug(f"Cannot find {symbol_name}")
```
ikelos: Was this just for testing or do you think the message has no value?  Happy to entertain the idea, I genuinely don't know how useful it is, but I prefer knowing when things have gone wrong than just silently skipping over them.  Perhaps moving if from `debug` to `constants.LOGLEVEL_VVVV` would be better?

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -107,19 +366,22 @@ def list_notify_routines(
...
                 count = 64
             else:
                 count = 8
 
-            fast_refs = ntkrnlmp.object(
-                object_type="array",
-                offset=symbol_offset,
-                subtype=ntkrnlmp.get_type("_EX_FAST_REF"),
-                count=count,
+            fast_refs = cast(
```
ikelos, line 386: I've never really used `cast` before, but I'm happy to have more accurate typing information if it helps.  Thanks!  5:)

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -249,7 +514,7 @@ def list_registry_callbacks(
                 "CmpCallBackCount",
                 "CallbackListHead",
             ]
-            vollog.debug("Failed to get registry callbacks!")
+            # vollog.debug("Failed to get registry callbacks!")
```
ikelos: Again, debugging information shouldn't affect normal usage, so I'd prefer it were there for developers to figure out when things go wrong.  If it's too verbose, that's fine and we can look for a better way to get the right information across, but if this was just a forgotten comment then it's be best to reinstate the message if possible (happy to change it to `constants.LOGLEVEL_VVVV` if that'd be better?

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -367,12 +640,16 @@ def list_bugcheck_callbacks(
...
-            yield "KeBugCheckCallbackListHead", callback.CallbackRoutine, component
+            yield (
+                "KeBugCheckCallbackListHead",
+                cast(int, callback.CallbackRoutine),
+                component,
+            )
 
     def _generator(self):
-        kernel = self.context.modules[self.config["kernel"]]
+        kernel = self.context.modules[cast(str, self.config["kernel"])]
```
ikelos: I'm interested to know under what circumstances this wouldn't be seen as a string by the type checker?  I'm all for improving the typing, but I'd prefer it not get out of hand manually casting everything rather than getting the types right at the source?

ikelos (commented):
Thanks, there's a lot to go through here but it generally looks pretty good.  It's quite a significant chunk of functionality though so I'd quite like to defer to @iMHLv2 's opinion to ensure it does what it's supposed and there's no gotchas that I've missed...

volatility3/framework/plugins/windows/callbacks.py
dgmcdona: Thanks, I think I forgot to update this when copy-pasting. There should really be three separate requirements - the `handles.Handles` plugin, the `poolscanner.PoolScanner` plugin, and the list of major functions `driverirp.MAJOR_FUNCTIONS`, the last of which I'm not sure how to define a requirement for, since it's not a class in another module, but just a list of strings. I'm going to define it as a `PluginRequirement` for now, but let me know if there's a better way.

volatility3/framework/plugins/windows/callbacks.py
dgmcdona: I was wondering if I was getting carried away with the type-casting - I was trying to make things as explicit as possible but this may have been overkill, so I'll roll that change back.

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -2,16 +2,25 @@
...
+from volatility3.framework import (
+    constants,
+    exceptions,
+    interfaces,
+    objects,
+    renderers,
+    symbols,
+)
 from volatility3.framework.configuration import requirements
+from volatility3.framework.interfaces.objects import ObjectInterface
```
ikelos: Please try not to import classes directly, but only ever modules.  Importing `ObjectInterface` like this may allow someone to one day write `from callbacks import ObjectInterface` and it then won't be clear that it actually belongs to another module.

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -33,27 +42,161 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            symbol_table: The name of an existing symbol table containing the symbols / types
+
+        Returns:
+            The list containing the built constraints.
+        """
+
+        PNP9_SIZE = 0x30
```
ikelos: I'm generally a fan of having fixed values pulled out and tagged with a comment or put in a variable, but it doesn't feel like these will get used anywhere else, so I'd be tempted to just embed them in the code below?  The code below already kinda hard codes the binary tag values so I'm not sure it's much different?

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -65,10 +208,133 @@ def create_callback_table(
...
+
+        for (
+            _constraint,
+            mem_object,
+            _header,
+        ) in poolscanner.PoolScanner.generate_pool_scan(
+            context, layer_name, nt_symbol_table, constraints
+        ):
+            try:
+                if hasattr(mem_object, "is_valid") and not mem_object.is_valid():
```
ikelos, line 251: What objects *don't* have an `is_valid` method?  Feels like this could be changed to just check `pointers`?

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -65,10 +208,133 @@ def create_callback_table(
...
+        elif type_name == symbol_table + constants.BANG + "_DBGPRINT_CALLBACK":
+            callback_type = "DbgSetDebugPrintCallback"
+            callback_address = memory_object.Function
+            details = renderers.NotApplicableValue()
+        else:
+            raise ValueError("Unexpected object type %s" % type_name)
+
+        return (
+            callback_type,
+            cast(int, callback_address),
```
ikelos: Yeah, probably a little overboard with the casting...  5;)  I mean, it's good and I approve of well typed systems, but... a lot of these things should already be the type they're cast to, and hopefully mypy will one day be able to figure that out, but I'm not sure littering our code with hints will help all that much?

ikelos (commented):
Yep, pretty happy with this, just a few more comments.  5:)

ikelos:
@iMHLv2 could you give this a scan over to check there's nothing subtle I've missed please?

volatility3/framework/plugins/windows/callbacks.py
dgmcdona: The check fails for `StructType` objects if we remove the `hasattr` check. For instance: `AttributeError: StructType has no attribute: callbacks-x641!_NOTIFICATION_PACKET.is_valid`

There is an explicitly defined `is_valid` method in the callbacks extension added as part of this PR, but I don't think any other objects will have an `is_valid` method since they are just `StructType` objects defined in the callbacks symbol files.

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -33,27 +41,155 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
         config_path: str,
     ) -> str:
-        """Creates a symbol table for a set of callbacks.
+        """Creates a symbol table for kernel callback objects.
 
         Args:
             context: The context to retrieve required elements (layers, symbol tables) from
-            symbol_table: The name of an existing symbol table containing the kernel symbols
-            config_path: The configuration path within the context of the symbol table to create
+            layer_name: The name of the layer on which to operate
```
ikelos: `layer_name` isn't part of the function signature for this call... or used within it anywhere?

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -65,10 +201,141 @@ def create_callback_table(
...
+            callback_address = memory_object.Callback
+            details = renderers.NotApplicableValue()
+
+        elif type_name == symbol_table + constants.BANG + "_DBGPRINT_CALLBACK":
+            callback_type = "DbgSetDebugPrintCallback"
+            callback_address = memory_object.Function
+            details = renderers.NotApplicableValue()
+
+        else:
+            raise ValueError("Unexpected object type %s" % type_name)
```
ikelos: Could we use `f"Unexpected object type {type_name}"` for consistency and readability please?

volatility3/framework/plugins/windows/callbacks.py
```diff
@@ -265,7 +538,13 @@ def list_bugcheck_reason_callbacks(
         layer_name: str,
         symbol_table: str,
         callback_table_name: str,
-    ) -> Iterable[Tuple[str, int, str]]:
+    ) -> Iterable[
+        Tuple[
+            str,
+            int,
+            Union[interfaces.objects.ObjectInterface, renderers.UnreadableValue],
```
ikelos: Happy with this, but feels like using the more generic `interfaces.renderers.BaseAbsentValue` might be more useful to avoid changes in the future, since it covers all of them?

volatility3/framework/symbols/windows/extensions/callbacks.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        try:
+            if not (
+                self.Entry.Flink.is_readable()
+                and self.Entry.Blink.is_readable()
+                and self.DeviceObject.is_readable()
+            ):
+                return False
+
+            device = self.DeviceObject.dereference()
```
ikelos: Again, dereference() *shouldn't* need to be explicitly specified.  I'd prefer not to muddy the code with them when they're unnecessary in case people start picking up bad habits and assuming they're necessary.  If it *is* necessary, I'd like to know why in case we can improve the framework...

ikelos (commented):
Thanks very much for your recent changes!  I'm sorry it's taken so long to get reviewed, I'll let @iMHLv2 know that it's still awaiting his approval before it can go in.  From a code perspective though it looks ok to me...  5:)

volatility3/framework/plugins/windows/callbacks.py
dgmcdona: 😄

dgmcdona:
I fixed an additional issue with `IoRegisterShutdownNotification` callbacks on x64 samples. The `_SHUTDOWN_NOTIFICATION` extension class should only be applied on x86 samples; see [here](https://github.com/volatilityfoundation/volatility3/compare/78db3d8457528fd6f98bee6c0c8df2c8948b9d7d..4d6f4bba0d5127f4a78074266091cf03b808b0f9#diff-c8fa55cd12e00fa088afa5f3be4853ee5f91345ab1803bf22b0e081e7a050b8b) and [here](https://github.com/volatilityfoundation/volatility3/compare/78db3d8457528fd6f98bee6c0c8df2c8948b9d7d..4d6f4bba0d5127f4a78074266091cf03b808b0f9#diff-cde9a57b579c0b95ba3254cad67158a93390e44fba2e6d0f6892496b558b1315)

iMHLv2 (approved):
@ikelos This looks good now. Thanks @dgmcdona!

PR #1115, opened 2024-03-23
macOS: Latest kernels support

Abyss-W4tcher:
Hello 👋,

This Pull Request aims to refactor the macOS x86/64 automagic layer stacking, as well as support for a recent kernel functionality named "KernelCache".

These changes only add APIs to the framework, without modifying inputs/outputs of existing ones.

Tested on kernels : 

- `10.9.3_build-13D65`
- `10.12.6_build-16G29`
- `10.15.7_build-19H15`
- `12.0.1_build-21A559`
- `13.6.4_build-22G513`
- `14.0_build-23A5257q`

Justifications are given, in the form of a comment or a link to the official darwin source code.

### KernelCache 

> The kernelcache is a pre-compiled and pre-linked version of the XNU kernel, with essential device drivers and kernel extensions. It is stored in a compressed format and is decompressed into memory during the boot process. The kernelcache facilitates faster booting by having a ready-to-use version of the kernel and essential drivers available, thereby reducing the time and resources that would otherwise be spent loading and dynamically linking these components at startup. 

source (french) : https://book.hacktricks.xyz/v/fr/macos-hardening/macos-security-and-privilege-escalation/mac-os-architecture#kernelcache

This feature impacts symbol resolving, as some need to be shifted with the "classical" KASLR shift, and some with the "vm_kernel_slide" shift. The current shift determining method, via `darwin banner` scanning, was actually calculating the "vm_kernel_slide" shift, from which it wasn't possible to also determine the KASLR shift. The new implementation takes advantage of the `lowGlo` kernel structure, which is consistent and available in old and recent kernels, and reveals the KASLR shift value. Banner scanning was also proven to be a bit slower, and more redundant, in its logic.

As discussed in #1114, a new macOS kernel module was added to the framework, which takes care of determining which slide to use to correctly resolve a symbol. Automagic is able to detect and instantiate it automatically, as well.

### Additional changes

Some variable names and logics were changed in the mac stacker, to be more straight to the point, while keeping the API compatibility.

---

If you possess any macOS X samples, please feel free to merge this PR in your local installation, and give feedback !

### DIFF
--- a/volatility3/framework/automagic/mac.py
+++ b/volatility3/framework/automagic/mac.py
@@ -5,6 +5,7 @@
 import logging
 import os
 import struct
+import re
 from typing import Optional
 
 from volatility3.framework import constants, exceptions, interfaces
@@ -167,19 +168,27 @@ class MacIntelStacker(interfaces.automagic.StackerLayerInterface):
         """Determines the offset of the actual DTB in physical space and its
         symbol offset."""
 
-        banner_major_version, banner_minor_version = [
-            int(x) for x in compare_banner[22:].split(b".")[0:2]
-        ]
+        darwin_signature_pattern = re.compile(
+            rb"Darwin Kernel Version (?P<major>\d{1,3})\.(?P<minor>\d{1,3})\.\d{1,3}:"
+        )
+        banner_versions = darwin_signature_pattern.search(compare_banner)
+        if not banner_versions:
+            vollog.log(
+                constants.LOGLEVEL_V,
+                f"Unable to determine major and minor versions of banner : {compare_banner}",
+            )
+            return 0
+        banner_major_version = int(banner_versions.group("major"))
+        banner_minor_version = int(banner_versions.group("minor"))
 
+        symbol_table_object: mac.MacKernelIntermedSymbols = context.symbol_space[
+            symbol_table
+        ]
         """
         References :
          - https://github.com/apple-open-source/macos/blob/14.3/xnu/osfmk/x86_64/lowmem_vectors.c#L78
          - https://github.com/apple-open-source/macos/blob/14.3/xnu/osfmk/x86_64/lowglobals.h#L50
         """
-
-        symbol_table_object: mac.MacKernelIntermedSymbols = context.symbol_space[
-            symbol_table
-        ]
         lowglo_json_address = symbol_table_object.get_symbol("lowGlo").address
         lowglo_phys_offset = cls.virtual_to_physical_address(lowglo_json_address)
 
--- a/volatility3/framework/automagic/mac.py
+++ b/volatility3/framework/automagic/mac.py
@@ -129,9 +129,7 @@ class MacIntelStacker(interfaces.automagic.StackerLayerInterface):
                 context.config[
                     cls.join(config_path, MacSymbolFinder.banner_config_key)
                 ] = str(banner, "latin-1")
-                symbol_cache.SqliteCache(identifiers_path).get_identifier_dictionary(
-                    operating_system="mac"
-                )
+
                 layer_class = intel.Intel32e
                 # If an mh_fileset_config exists, this means the KernelCache is in an MH_FILESET format
                 mh_fileset_config = context.config.branch(
--- a/volatility3/framework/automagic/mac.py
+++ b/volatility3/framework/automagic/mac.py
@@ -7,7 +7,7 @@ import os
 import struct
 from typing import Optional
 
-from volatility3.framework import constants, exceptions, interfaces, layers
+from volatility3.framework import constants, exceptions, interfaces
 from volatility3.framework.automagic import symbol_cache, symbol_finder
 from volatility3.framework.configuration import requirements
 from volatility3.framework.layers import intel, scanners
--- a/volatility3/framework/automagic/mac.py
+++ b/volatility3/framework/automagic/mac.py
@@ -93,53 +93,34 @@ class MacIntelStacker(interfaces.automagic.StackerLayerInterface):
                     compare_banner_offset=banner_offset,
                     progress_callback=progress_callback,
                 )
-
                 if kaslr_shift == 0:
                     vollog.log(
                         constants.LOGLEVEL_VVV,
-                        f"Invalid kalsr_shift found at offset: {banner_offset}",
+                        f"Unable to calculate a valid KASLR shift for banner : {banner}",
                     )
                     continue
 
-                bootpml4_addr = cls.virtual_to_physical_address(
-                    table.get_symbol("BootPML4").address + kaslr_shift
-                )
-
-                new_layer_name = context.layers.free_layer_name("MacDTBTempLayer")
-                config_path = join("automagic", "MacIntelHelper", new_layer_name)
-                context.config[join(config_path, "memory_layer")] = layer_name
-                context.config[join(config_path, "page_map_offset")] = bootpml4_addr
-
-                layer = layers.intel.Intel32e(
-                    context,
-                    config_path=config_path,
-                    name=new_layer_name,
-                    metadata={"os": "Mac"},
-                )
-
-                idlepml4_ptr = table.get_symbol("IdlePML4").address + kaslr_shift
+                idlepml4_json_address = symbol_table.get_symbol("IdlePML4").address
+                idlepml4_ptr = idlepml4_json_address + kaslr_shift
                 try:
-                    idlepml4_str = layer.read(idlepml4_ptr, 4)
+                    idlepml4_raw = context.layers[layer_name].read(
+                        cls.virtual_to_physical_address(idlepml4_ptr), 4
+                    )
                 except exceptions.InvalidAddressException:
                     vollog.log(
                         constants.LOGLEVEL_VVVV,
-                        f"Skipping invalid idlepml4_ptr: 0x{idlepml4_ptr:0x}",
+                        f"Skipping invalid IdlePML4 pointer: {hex(idlepml4_ptr)}",
                     )
                     continue
 
-                idlepml4_addr = struct.unpack("<I", idlepml4_str)[0]
-
-                tmp_dtb = idlepml4_addr
-
-                if tmp_dtb % 4096:
+                dtb_candidate = struct.unpack("<I", idlepml4_raw)[0]
+                if dtb_candidate % 4096:
                     vollog.log(
                         constants.LOGLEVEL_VVV,
-                        f"Skipping non-page aligned DTB: 0x{tmp_dtb:0x}",
+                        f"Skipping non-page aligned DTB: {hex(dtb_candidate)}",
                     )
                     continue
-
-                dtb = tmp_dtb
-
+                dtb = dtb_candidate
                 # Build the new layer
                 new_layer_name = context.layers.free_layer_name("IntelLayer")
                 config_path = join("automagic", "MacIntelHelper", new_layer_name)
--- a/volatility3/framework/automagic/mac.py
+++ b/volatility3/framework/automagic/mac.py
@@ -78,13 +78,13 @@ class MacIntelStacker(interfaces.automagic.StackerLayerInterface):
             isf_path = mac_banners.get(banner, None)
             if isf_path:
                 table_name = context.symbol_space.free_table_name("MacintelStacker")
-                table = mac.MacKernelIntermedSymbols(
+                symbol_table = mac.MacKernelIntermedSymbols(
                     context=context,
-                    config_path=join("temporary", table_name),
+                    config_path=cls.join("temporary", table_name),
                     name=table_name,
                     isf_url=isf_path,
                 )
-                context.symbol_space.append(table)
+                context.symbol_space.append(symbol_table)
                 kaslr_shift = cls.find_aslr(
                     context=context,
                     symbol_table=table_name,
--- a/volatility3/framework/automagic/mac.py
+++ b/volatility3/framework/automagic/mac.py
@@ -255,9 +255,6 @@ class MacIntelStacker(interfaces.automagic.StackerLayerInterface):
                 config_path = cls.join(
                     MacSymbolFinder.mh_fileset_config_path_prefix, symbol_table
                 )
-                context.config[cls.join(config_path, "mh_fileset_kernel_cache_check")] = (
-                    mh_fileset_kernel_cache_check
-                )
                 context.config[cls.join(config_path, "vm_kernel_slide")] = (
                     vm_kernel_slide_candidate
                 )
@@ -265,7 +262,10 @@ class MacIntelStacker(interfaces.automagic.StackerLayerInterface):
                 context.config[cls.join(config_path, "kernel_end")] = etext
 
             aslr_shift = aslr_shift_candidate
-            vollog.log(constants.LOGLEVEL_VVVV, f"Mac find_aslr found \"vm_kernel_slide\" to be: {hex(vm_kernel_slide_candidate)}")
+            vollog.log(
+                constants.LOGLEVEL_VVVV,
+                f'Mac find_aslr found "vm_kernel_slide" to be: {hex(vm_kernel_slide_candidate)}',
+            )
 
             break
         vollog.log(
--- a/volatility3/framework/automagic/mac.py
+++ b/volatility3/framework/automagic/mac.py
@@ -182,16 +182,41 @@ class MacIntelStacker(interfaces.automagic.StackerLayerInterface):
             if aslr_shift_candidate & 0xFFF != 0:
                 continue
 
-            tmp_aslr_shift = offset - cls.virtual_to_physical_address(
-                version_json_address
+            # We can use the convenient module methods, with virtual addresses in physical space,
+            # because the shift is constant and known.
+            module = context.module(
+                symbol_table,
+                layer_name,
+                aslr_shift_candidate - cls._KERNEL_MIN_ADDRESS,
             )
 
-            major_string = context.layers[layer_name].read(
-                version_major_phys_offset + tmp_aslr_shift, 4
-            )
-            major = struct.unpack("<I", major_string)[0]
+            # https://github.com/apple-open-source/macos/blob/14.3/xnu/osfmk/i386/i386_vm_init.c#L282
+            vm_kernel_slide_candidate = module.object_from_symbol("vm_kernel_slide")
+            mh_fileset_kernel_cache_check = False
+            try:
+                mh_fileset_kernel_cache_check = cls.detect_mh_fileset_kernel_cache(
+                    context,
+                    symbol_table,
+                    layer_name,
+                    aslr_shift_candidate,
+                )
+            except exceptions.SymbolError as e:
+                vollog.log(constants.LOGLEVEL_VVVV, e)
+            except exceptions.InvalidAddressException:
+                continue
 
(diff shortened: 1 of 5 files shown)


volatility3/framework/automagic/mac.py
```diff
@@ -86,63 +93,56 @@ def stack(
                     compare_banner_offset=banner_offset,
                     progress_callback=progress_callback,
                 )
-
                 if kaslr_shift == 0:
                     vollog.log(
                         constants.LOGLEVEL_VVV,
-                        f"Invalid kalsr_shift found at offset: {banner_offset}",
+                        f"Unable to calculate a valid KASLR shift for banner : {banner}",
```
ikelos, line 100: Outputting the banner seems less useful than the offset that was being checked to be a kernel?  Any reason for the change in debug message?

volatility3/framework/automagic/mac.py
```diff
@@ -86,63 +93,56 @@ def stack(
...
-                context.config[join(config_path, "page_map_offset")] = dtb
-                context.config[join(config_path, MacSymbolFinder.banner_config_key)] = (
-                    str(banner, "latin-1")
+                config_path = cls.join("automagic", "MacIntelHelper", new_layer_name)
+                context.config[cls.join(config_path, "memory_layer")] = layer_name
+                context.config[cls.join(config_path, "page_map_offset")] = dtb
+                context.config[
+                    cls.join(config_path, MacSymbolFinder.banner_config_key)
+                ] = str(banner, "latin-1")
+                symbol_cache.SqliteCache(identifiers_path).get_identifier_dictionary(
```
ikelos: The output of this call doesn't seem to be used (and it would require an SqliteCache, which would make it hard for us to make thing cache system more generic in the future).

volatility3/framework/automagic/mac.py
```diff
@@ -86,63 +93,56 @@ def stack(
...
+                    operating_system="mac"
+                )
+                layer_class = intel.Intel32e
+                # If an mh_fileset_config exists, this means the KernelCache is in an MH_FILESET format
+                mh_fileset_config = context.config.branch(
+                    cls.join(MacSymbolFinder.mh_fileset_config_path_prefix, table_name)
                 )
+                if mh_fileset_config != {}:
```
ikelos: `if mh_fileset_config:` feels less ugly than an equality test against an empty dictionary.

volatility3/framework/automagic/mac.py
```diff
@@ -168,80 +168,223 @@ def find_aslr(
...
 
-        version_minor_symbol = symbol_table + constants.BANG + "version_minor"
-        version_minor_json_address = context.symbol_space.get_symbol(
-            version_minor_symbol
-        ).address
-        version_minor_phys_offset = cls.virtual_to_physical_address(
-            version_minor_json_address
-        )
+        banner_major_version, banner_minor_version = [
+            int(x) for x in compare_banner[22:].split(b".")[0:2]
```
ikelos: This feels like it will be fragile, could we split this out into a separate function so it's clearer, and maybe hunt for the version number a little better than hard coding 22 bytes in?  I realize this was in the previous code, but this gives us an opportunity to tidy it up whilst we're here revamping everything.

volatility3/framework/automagic/mac.py
```diff
@@ -168,80 +168,223 @@ def find_aslr(
...
+                    cls.vm_kernel_slide_kernel_cache_calculations(
+                        context,
+                        symbol_table,
+                        layer_name,
+                        aslr_shift_candidate,
+                        vm_kernel_slide_candidate,
+                    )
+                )
+
+            if vm_kernel_slide_candidate & 0xFFF != 0:
```
ikelos, line 238: Should this be the size of a page?  If so, this should be the value from the `layer.page_size - 1`  rather than a hard coded value...

volatility3/framework/automagic/mac.py
```diff
@@ -168,80 +168,223 @@ def find_aslr(
...
 
         return addr
 
+    @classmethod
+    def _lowglo_scan_generator(
+        cls,
+        context: interfaces.context.ContextInterface,
+        layer_name: str,
+        progress_callback: constants.ProgressCallback,
+    ):
```
ikelos: Could we get a return type signature type please?  Looks like it would return a generator of `int`s...

volatility3/framework/automagic/module.py
```diff
@@ -62,5 +62,17 @@ def __call__(
...
                 context.config[offset_config_path] = offset
 
+                # macOS KernelCache module override
+                layer_class_config_path = interfaces.configuration.path_join(
+                    new_config_path, req, "class"
+                )
+                layer_class_str = context.config[layer_class_config_path]
+                if (
+                    layer_class_str
+                    == "volatility3.framework.layers.intel.MacIntelMhFilesetKernelCache"
```
ikelos, line 72: This feels pretty ugly, hard coded strings to classes is... nasty.  I'd almost sooner this imports the class and then asks the class for its full path, so that IDEs/tools can keep it in check if the class ever moves.  I realize the value is set further up, but it's set to something generic rather than a Mac specific value (and the same for the means of determining whether it's a Mac layer or not).

volatility3/framework/contexts/__init__.py
```diff
@@ -462,3 +462,149 @@ def __init__(
         Module.__init__(
             self, context, name, layer_name, offset, symbol_table_name, layer_name
         )
+
+
+class MacOSKernelCacheSupportModule(Module):
```
ikelos, line 468: If you're constructing a new `Module` with addition configuration items, you'll need to override [`build_configuration`](https://github.com/Abyss-W4tcher/volatility3/blob/781d1f00e2964055a97b1ce1e0bb692a73206e78/volatility3/framework/interfaces/context.py#L176C9-L176C28) and should define the `get_requirements` method of the underlying `ConfigurableInterface`.  This has also highlighted that the `Module`/`ModuleInterface` classes don't define this (but should do).

volatility3/framework/contexts/__init__.py
```diff
@@ -462,3 +462,149 @@ def __init__(
...
+        )
+        not_kernelcache_module_list = list(
+            self._context.symbol_space.get_symbols_by_location(
+                offset=offset - self._vm_kernel_slide,
+                size=size,
+                table_name=self.symbol_table_name,
+            )
+        )
+
+        return kernelcache_module_list + not_kernelcache_module_list
```
ikelos: This feels like it has the opportunity to potentially return the same symbol twice, once from each list?  If we have a means to determine which symbol belongs to which list, wouldn't it be better to only return the value if it's from the correct table?

volatility3/framework/contexts/__init__.py
```diff
@@ -462,3 +462,149 @@ def __init__(
...
+
+        pathjoin = interfaces.configuration.path_join
+        context.config[pathjoin(config_path, "vm_kernel_slide")] = context.config[
+            pathjoin(config_path, self.layer_name, "vm_kernel_slide")
+        ]
+        context.config[pathjoin(config_path, "kernel_start")] = (
+            context.config[pathjoin(config_path, self.layer_name, "kernel_start")]
+            & self.context.layers[self.layer_name].address_mask
+        )
+        context.config[pathjoin(config_path, "kernel_end")] = (
```
ikelos, line 496: Rather than being a parameter, this should be based off the module `offset` and `size`.

volatility3/framework/contexts/__init__.py
```diff
@@ -462,3 +462,149 @@ def __init__(
...
+    def __init__(
+        self, context: interfaces.context.ContextInterface, config_path: str, name: str
+    ) -> None:
+        super().__init__(context, config_path, name)
+
+        pathjoin = interfaces.configuration.path_join
+        context.config[pathjoin(config_path, "vm_kernel_slide")] = context.config[
+            pathjoin(config_path, self.layer_name, "vm_kernel_slide")
+        ]
+        context.config[pathjoin(config_path, "kernel_start")] = (
```
ikelos, line 492: @#Rather than being a new config option, this should use the existing `offset` value I believe.

volatility3/framework/layers/intel.py
```diff
@@ -501,3 +501,25 @@ class WindowsIntel32e(WindowsMixin, Intel32e):
 
     def _translate(self, offset: int) -> Tuple[int, int, str]:
         return self._translate_swap(self, offset, self._bits_per_register // 2)
+
+
+class MacIntelMhFilesetKernelCache(Intel32e):
```
ikelos, line 508: I'm not sure why these have been added to the layer, rather than the module?  I think I probably need to review how configurations for modules are saved.

ikelos (commented):
Thanks very much for this, it looks setup the way I was hoping, but it's also quite involved (not quite as involved as the arm64 layer, but getting close).  I've asked @atcuno for a review to make sure the Mac side of things and the slide stuff makes sense.  I've given it a once over and it looks like I didn't implement everything needed for `Module`s properly, so I probably should go fix that first, but this looks very promising.  5:)

volatility3/framework/automagic/mac.py
Abyss-W4tcher: The banner offset isn't relevant anymore, to detect KASLR. This message can be useful, as there are many different banners scanned in a memory sample. KASLR shift will be banner independant, but the validation stage will stop at `version_major` and `version_minor` checks.

volatility3/framework/automagic/mac.py
Abyss-W4tcher: Actually, this is my bad, it isn't relevant here. This might be the result of a wrong manipulation, or copy/paste...

volatility3/framework/automagic/mac.py
Abyss-W4tcher: I made a change to this logic, by leveraging a regex scanner. If you had something else in mind, do not hesitate to inform me.

volatility3/framework/automagic/mac.py
Abyss-W4tcher: The page size might be architecture independant, even if it's 4kB for macOS x86/64, in the future it might be 16kB or 64kB for AArch64. To determine the page size, we would typically need to read a symbol from the kernel (if it even exists), but is a bit of a "snake biting its tail" situation.

This logic would need to be updated at two other places in this file (dtb aligned and aslr shift aligned).

volatility3/framework/automagic/module.py
Abyss-W4tcher: I was trying to do a normal class/Type comparison, when I realized everything in the config was indeed a string. So I decided to mimic : 

https://github.com/volatilityfoundation/volatility3/blob/55dd39f2ba60ffdd2126b7ea011940f0df42815a/volatility3/framework/automagic/module.py#L37-L39

I am currently checking for methods of converting an object to a full string path, which might be cleaner than the previous snippet of code :

`"."join(contexts.Module.__module__, contexts.Module.__qualname__)`. 

But converting `context.config[layer_class_config_path]` string to a Python class might involve dynamic string to object manipulation, which is neither acceptable ?

volatility3/framework/contexts/__init__.py
Abyss-W4tcher: `offset` (a.k.a `kernel_virtual_offset`) is different from `vm_kernel_slide` (in the case of an MH_FILESET KernelCache), and needs to be passed as a config option.

volatility3/framework/layers/intel.py
Abyss-W4tcher: As discussed on Slack, this is the only reliable way I found to have the config options carried from automagic to the global context config (which are later used in the new `MacOSKernelCacheSupportModule`.

volatility3/framework/contexts/__init__.py
Abyss-W4tcher: In the case of a KernelCache, the `offset` will be the start of the machO KernelCache __TEXT, and not the "classic" kernel __TEXT start.

Using these parameters allow to use pre-determined values, instead of calculating them "on the fly", risking to mess things up. 

FYI, here is a typical configuration, right before exiting `mac.py` automagic : 

```json
{
  "kernel_banner": "Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:23 PDT 2021; root:xnu-8019.41.5~1/RELEASE_X86_64\u0000",
  "kernel_end": 18446743524345970688,
  "kernel_start": 18446743524335484928,
  "kernel_virtual_offset": 384188416,
  "memory_layer": "VmwareLayer",
  "page_map_offset": 451420160,
  "vm_kernel_slide": 379650048
}
```

volatility3/framework/contexts/__init__.py
Abyss-W4tcher: I wanted to play it extra safe on this part, but you are right, we can speed up things 👍. This also made me realize, that my function to check if an offset was in the kernel boundaries, wasn't checking if it was absolute. This has been updated.

volatility3/framework/automagic/mac.py
ikelos: Well, given it can be different we should at least parameterize these.  The page size should be dependent on the architecture, so a value taken from the layer (which is why the layer has the page_size value)...

volatility3/framework/automagic/mac.py
ikelos: It does still get passed to `find_aslr` though, so either it is still useful, or the signature for `find_aslr` takes in unnecessary parameters (which should be fixed, but I don't know which version numbers would need bumping to do that safely)...

volatility3/framework/automagic/module.py
ikelos: Yeah, it should end up as a string in the config, but we want to make sure that string relates to an existing class and that if the file gets moved/renamed etc, an IDE doing refactor can spot the break, hence converting it from an object to a string name.  So, yeah, basically convert the python object to the string name (we can make a separate function for clarity) but then compare `layer_class_str == class_string_name(class_we_care_about)` rather than a fixed string.

You shouldn't ever need to convert from the string to a class?

volatility3/framework/contexts/__init__.py
ikelos: `kernel_virtual_offset` is what we've typically passed in to the kernel module, but *all* modules, regardless of which module, have the common values of `offset` and `size`, but for the new module you've constructed you've chosen different names.  When I add those into the `get_requirements` of the `Module`/`ModuleInterface` you'll see that you're defining the same values but with a different name.  I'd prefer we stick with the generic value names that modules already have rather than custom ones for this specific type of module.

volatility3/framework/contexts/__init__.py
ikelos: This is a module, the module takes an offset and a size (or is supposed to).  If it needs additional parameters, that's fine, but from the looks of it, it's just used different parameter names for values that actually mean the same as offset and size.  The offset is where the module appears in virtual memory, and size is how many bytes it is.

volatility3/framework/layers/intel.py
ikelos: I think I'm missing something, these should be parameters on the Module and have nothing to do with the layer/architecture?  Lemme get the `Module`/`ModuleInterface` stuff tidied up and then we can look at what's going on...

ikelos (commented):
Thanks for making some of those changes.  I'm clearly still quite confused about some of the moving pieces.  Specifically why the layer's are getting touched when the only thing that should be impacted are the modules.  I'll get a PR out for the changes this has highlighted need making to the `Module`/`ModuleInterface` and then we can see how everything fits together after that...

volatility3/framework/automagic/mac.py
Abyss-W4tcher: Exactly, some parameters aren't necessary anymore, but it breaks the API...

volatility3/framework/layers/intel.py
Abyss-W4tcher: Basically, we determine useful values in automagic, and I want them to be accessible when we create the modules in automagic/module.py. 

This have nothing to do with the layer indeed, it is just that the stacked layer has a config, and it is saved in the global context, so I can access them later (and it is also saved to the config, which can be exported with `--save-config`)

I'll wait for your update, I think the modules are missing the `get_requirements()` as you pointed out, which might help saving the config to the global context ?

volatility3/framework/automagic/mac.py
Abyss-W4tcher: The `LimeLayer`, `VMWareLayer` , `FileLayer` etc. is embedding the page size ? I don't know how this could be relevant to higher layers (Intel, AArch64), as the page size is typically decided in the kernel config ?

But at least parameterizing it would be more uniform, indeed !

volatility3/framework/automagic/module.py
Abyss-W4tcher: I see, I will change this bit, but after your update, as we might remove the `MacIntelMhFilesetKernelCache` and use something that makes more sense.

ikelos:
Ok, so `ModuleRequirements` differ from `TranslationLayerRequirements` and `SymbolTableRequirements`, in that they didn't have a class and therefore there was only ever expected to be one `Module` class.  Since we're now introducing another potential class, it needs some reworking...

volatility3/framework/contexts/__init__.py
Abyss-W4tcher: *Sorry if the following has too much references, but I prefer to justify what I am proposing.*

`offset` in this case is equivalent to `vm_kernel_slide`, yes, and I can update this.

But `offset` != `kernel_start`, as `kernel_start` is typically `stext` symbol. Doing so `kernel_end` - `kernel_start` is the size of kernel __TEXT, not especially the size of this module, starting from offset `offset`.

As an example, `vm_kernel_slide`/`offset` physical value can be `0x16c10000`, but `kernel_start` is `0x16a10000`.

See, as a quick [reference](https://github.com/apple-open-source/macos/blob/ea4cd5a06831aca49e33df829d2976d6de5316ec/xnu/osfmk/kern/debug.c#L1870) (or even [here](https://github.com/apple-open-source/macos/blob/ea4cd5a06831aca49e33df829d2976d6de5316ec/xnu/osfmk/i386/i386_vm_init.c#L324) for more details) : 

```c
paniclog_append_noflush("Kernel slide:      0x%016lx\n", (unsigned long) vm_kernel_slide);
paniclog_append_noflush("Kernel text base:  %p\n", (void *)vm_kernel_stext);
```

Which in the end, is needed to mimic the following [behaviour](https://github.com/apple-open-source/macos/blob/ea4cd5a06831aca49e33df829d2976d6de5316ec/xnu/osfmk/i386/i386_vm_init.c#L1084) : 


```c
/*
 * Returns true if the address lies in the kernel __TEXT segment range.
 */
bool
kernel_text_contains(vm_offset_t addr)
{
	return vm_kernel_stext <= addr && addr < vm_kernel_etext;
}
```

Abyss-W4tcher: I consider `kernel_start` and `kernel_end` as additional requirements for this specific module.

volatility3/framework/contexts/__init__.py
ikelos: It's actually not the Module/ModuleInterface that needs to define this, but the ModuleRequirement, so this might require changing mac plugins to require a mac module?  5:S

PR #1118, opened 2024-03-24
Bulk out Module support for specific requirements

ikelos:
This should ensure that modules more explicitly state what requirements they have which should also allow that to be recorded more accurately in the configuration when needed.

@Abyss-W4tcher please could you test this (after adding your own requirements to your new Module type) and report back whether the values you need get stored appropriately or not.

### DIFF
--- a/volatility3/framework/contexts/__init__.py
+++ b/volatility3/framework/contexts/__init__.py
@@ -14,6 +14,7 @@ import logging
 from typing import Callable, Iterable, List, Optional, Set, Tuple, Union
 
 from volatility3.framework import constants, interfaces, symbols, exceptions
+from volatility3.framework.configuration import requirements
 from volatility3.framework.objects import templates
 
 vollog = logging.getLogger(__name__)
@@ -225,6 +226,20 @@ class Module(interfaces.context.ModuleInterface):
         context.config[config_path] = return_val.name
         # Add the module to the context modules collection
         return return_val
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        """Returns a list of Requirement objects for this type of layer."""
+        return [
+            requirements.IntRequirement(name="offset", optional=False),
+            requirements.TranslationLayerRequirement(name="layer_name", optional=False),
+            requirements.TranslationLayerRequirement(
+                name="native_layer_name", optional=True
+            ),
+            requirements.SymbolTableRequirement(
+                name="symbol_table_name", optional=False
+            ),
+        ]
 
     def object(
         self,
--- a/volatility3/framework/interfaces/context.py
+++ b/volatility3/framework/interfaces/context.py
@@ -157,6 +157,10 @@ class ModuleInterface(interfaces.configuration.ConfigurableInterface):
         super().__init__(context, config_path)
         self._module_name = name
 
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        """Returns a list of Requirement objects for this type of layer."""
+
     @property
     def _layer_name(self) -> str:
         return self.config["layer_name"]
@@ -175,7 +179,6 @@ class ModuleInterface(interfaces.configuration.ConfigurableInterface):
 
     def build_configuration(self) -> "interfaces.configuration.HierarchicalDict":
         """Builds the configuration dictionary for this specific Module"""
-
         config = super().build_configuration()
 
         config["offset"] = self.config["offset"]


Abyss-W4tcher:
I will try it soon and keep you informed 👍

Abyss-W4tcher:
I have merged these changes, and tried on a basic sample `mac-sample-1.bin`, and adding random non-optional requirements isn't preventing the module from instantiating.

In fact, `get_requirements()` doesn't seem to be called in the first place.

Abyss-W4tcher:
Hi, any update on this feature ?

ikelos:
No, not yet sorry, I need to find time to unpick why `get_requirements` isn't getting called appropriately...

ikelos:
Ok, so my mistake.  It turns out it's not the Module that's used to figure out its requirements, it's the ModuleRequirement which specifies them, meaning for a specific module to be loaded, we can't just create a new `ArmModule` with different requirements, we'd need to define a custom `ArmModuleRequirement` which somewhat does away with the point of trying to keep things modular.  As such this branch is duff and I'm going to close it off.

In the larger picture of getting the Mac stuff in the right way, I'm afraid it's back to the drawing board for a little bit to figure out how to get optional requirements for a ModuleRequiement (I suspect it'll come down to a MacModuleRequirement, which isn't ideal, but may be workable)...

atcuno:
@ikelos we need to revisit this asap for the parity release. 

The initial issue came up where in recent testing we don't support basically any recent / currently supported Mac version due to a mix of non-updated extensions/type handling as well as the shifting symbols.

@Abyss-W4tcher has two PRs to fix both of these problems, but he told me that this pull request here about Module requirements is blocking them. 

Is there a way to get this resolved in the next week or so?

ikelos:
Possibly, but it's a big change to how our module system currently works.  I don't really get how the apple symbols have two shifts, but my concept was to make an AppleModule which is basically a facade for two other modules (with differing shifts) and then when a symbol is requested, choose which one to pull it from (and which offset to automatically use).  I trying to avoid us doing the shifting the way we did with linux originally which would've gotten people very confused when they asked for a symbol and got one back at a weird address.  I've also no idea whether it'd work, and I suspect it would mean people would need to be careful when using it.  I'm not sure I see how else to work this?  The kernel is the kernel is the kernel, it starts at a specific offset, unless there's two symbol tables, I just don't understand what change they've made (and I haven't really been able to build up a good picture from the information I have been told).

This PR went wrong, but essentially, we need to figure out a clear concept of what's happening in apple world, and then model it within volatility.  At the moment, all I've really been told is that there are "two shift values" and I haven't managed to work up a model around that.  If someone can either explain it more clearly and answer questions about it so I can mock up a proper solution, or work up a new concept for `module` that fits the requirements, I can review it.  It's not a small amount of work though.

ikelos:
(This also has been known about for a year, hoping for a two week turn around is... possible but really we need more people keeping an eye on old bugs and things if sudden deadlines are going to appear.)  I don't think this blocks feature parity with volatility 2, volatility 2 wouldn't be able to process these images either, so whilst I'll do my best to get it in the big release, it may have to wait until the release after this one if we can't figure it out before then...

Abyss-W4tcher:
### A bit of context

Apple's XNU kernel embeds the kernel and kernel extensions into three major [Kernel Collections](https://keith.github.io/xcode-man-pages/kmutil.8.html#COLLECTIONS):

- **The boot kext collection**: contains the kernel and all system kexts necessary for starting and bootstrapping the operating system.
- **The system kext collection**: if used, contains all remaining system kexts required by the operating system, and is loaded after boot. It is prelinked against the boot kext collection. Note that on Apple Silicon Macs, there is no system kext collection.
- **The auxiliary kext collection**: if built, contains kexts placed in /Library/Extensions and any other third-party kexts installed on the system. 

Basically, each of them are Mach-O executables, that prelink kernel extensions with the kernel. A specific type of kernel collection used during startup is named `KernelCache`. 

Starting from macOS X 11.0.1, Apple introduced a new type of mach-o format for `KernelCaches` named `MH_Fileset` (MH standing for **M**ach-o **H**eader).  As it names suggests it, it is [a set of related](https://www.df-f.com/blog/tistheseason) mach-o executables which are loaded together in the same address space. 

### Back at work 

I got back into my previous work on the double module setup. After some time spent debugging, I noticed that there was something inherently weird with the symbol addresses from the ISF used. In fact, many symbols residing in specific sections (`__TEXT`, `sHIB`, `__DATA_CONST`) were **each ** needing a different slide from the classic KASLR offset calculated by automagic. 

Generally, we expect all the symbols from an ISF to be slid out by a single offset, but this is not the case here. I currently have an implementation that is capable of detecting and handling most of the symbols, however I don't feel that runtime patching is really efficient. 

### The underlying issue 

We commonly indicate people to use the `Kernel Development Kit` from apple's developer website to extract the required `kernel` and `kernel.dSYM` mach-os. The issue, is that the `kernel` file is not exactly the `KernelCache` leveraged to boot the kernel:

```sh
$ ipsw macho info kernel                 
Magic         = 64-bit MachO
Type          = EXECUTE # <- this is equal to `FILESET` against the required `KernelCache` file
CPU           = Amd64, x86_64
Commands      = 22 (Size: 5000)
Flags         = NoUndefs, PIE
```

Basically, the symbols extracted from `kernel` are lacking some relocation offsets which leads to failure in their resolution without manually fixing them dynamically. 

### The potential solution

So, in the end, this is not really a Vol3 issue, even though the proposed PR patches the issue for most scenarios without people having to generate again all the ISFs (`dwarf2json` is not to blame).

The fun part, is that accessing the `KernelCache` file is not possible from the KDK. I leave out some details, but even with the `KernelCache` in hands there should be additional work to do to extract the right data. A [tool](https://blacktop.github.io/ipsw/blog/kernel-symbolication/) allows to fetch them from an Apple firmware database (like we do for Windows): 

image

Unfortunately, it is not possible to download `amd64` versions of a kernel release, as only `arm64` files are returned (it is tied to physical Apple silicon devices). As `amd64` is not really supposed to be used anymore with macOSX (only in VMs setups), I guess they didn't compile a `KernelCache` for it. 

I haven't found a reliable solution, but maybe most of the work would be to make a new ISF generation procedure that would patch this symbol shifting issue at the root and make vol3's life easier. Once I haven't completely removed this possibility, I think this PR can stay in standby.

I'll keep you updated, and feel free to ask any question in this subject. I don't want to bother you with low level details, but having your point of view on the general design would be much appreciated.

Thanks :)

ikelos:
Really good research, thanks!  It *sounds* as though a suitable ISF would be usable with volatility3 then?  And that the offsets within the ISFs currently generated are incorrect for Macs (not because of dwarf2json, but because they're based on the wrong data, and should be generated using the `KernelCache`)?  The solution proposed here is really ugly and hacks in place a new concept specific to Macs, which seemingly doesn't model what's acutally happening?  I'm more than happy not to merge this, because there were a number of complications that weren't even resolved in this attempt.

The main question is then, how should volatility 3 be modelling this?  If the symbol addresses are incorrect (but otherwise consistent within a memory image) then I'd prefer to go down that route.  If the same set of addresses need to be altered separately per memory image, then we should work on how that change can be determined and applied.

Would it be possible to carve the KernelCache out of the memory image and generate the symbols live (a lot of work, but i'm just investigating possibilities)?  If the KernelCache isn't in the KDK, how does it get onto a live running machine?  Is it generated from data in the KDK, or downloaded on first install?

Abyss-W4tcher:
*Reminder: all these information only apply to x86_64. Apple Silicon ARM64 support will typically require new researches.*

## Memory carving 

I was able to extract `com.apple.kernel` and the `KernelCache` from memory, however interesting sections seem stripped (it's not funny otherwise). I haven't found a `VMCoreInfo` equivalent despite my researches, the closest I found was the `hibernation` but it is quite limited. 

## New tools and techniques to access the `KernelCache`

There is an awesome [Swiss Army Knife](https://github.com/blacktop/ipsw) for macOS researching named `ipsw`. **This tool is portable and can run on Linux, macOS and Windows. Relying on it exclusively for the following procedures eliminates the need to setup a custom environment to manipulate macOS package formats.**

[IPSW](https://en.wikipedia.org/wiki/IPSW) is also the file format used to install iOS, iPadOS, tvOS, HomePod, watchOS, and most recently, macOS firmware for devices equipped with Apple silicon.

I started investigating if we could access the required data from these packaged files, and I think I was able to get exactly what we needed (I'll post the full procedure for better reproducibility): 

```sh
$ ipsw dl appledb --os macOS --build 22F82
• Getting (1/1) IPSW: UniversalMac_13.4.1_22F82_Restore.ipsw
   64.16 MiB / 11.86 GiB [----------------------------------------------------------| ]
# Extract the boot kernel collection
$ ipsw extract UniversalMac_13.4.1_22F82_Restore.ipsw --output extracted/ --files --pattern "BootKernelExtensions\.kc$" 
   • Extracting files matching pattern "BootKernelExtensions\\.kc$"
      • Mounting DMG 096-09661-088.dmg
      • Mounting DMG 096-09706-080.dmg
      • Mounting DMG 096-09648-077.dmg
         • Extracting to extracted/22F82__MacOS/System/Library/KernelCollections/BootKernelExtensions.kc
      • Created extracted/22F82__MacOS/System/Library/KernelCollections/BootKernelExtensions.kc
# Dump symbols from the (com.apple.)kernel symtab embedded inside the KernelCache
$ ipsw --no-color macho info extracted/22F82__MacOS/System/Library/KernelCollections/BootKernelExtensions.kc --fileset-entry com.apple.kernel --symbols > symbols.txt
$ cat symbols.txt
# ...snip...
0xffffff8001119790:  (__DATA,__common) external	_zone_guard_pages	
0xffffff8000218c78:  (__DATA_CONST,__const) external	_zone_info	
0xffffff8001119780:  (__DATA,__common) external	_zone_pages_wired	
0xffffff80004df9d0:  (__TEXT,__text) external	_zone_require
```

The `com.apple.kernel` symbols extracted from the `KernelCache` and listed above are correctly linked/relocated, whereas the ones from KDK's `kernel` aren't (the current ISFs we have). 

### Updated procedure ? 

#### KDK: `kernel` and `kernel.dSYM` 

```sh
$ ipsw dl kdk --build 22F82 --output KDK/
$ ipsw disk dmg KDK/Kernel_Debug_Kit_13.4.1_build_22F82.dmg            
   • 0) Protective Master Boot Record (MBR : 0)
   • 1) GPT Header (Primary GPT Header : 1)
   • 2) GPT Partition Data (Primary GPT Table : 2)
   • 3)  (Apple_Free : 3)     
   • 4) disk image (Apple_HFS : 4)
   • 5)  (Apple_Free : 5)     
   • 6) GPT Partition Data (Backup GPT Table : 6)
   • 7) GPT Header (Backup GPT Header : 7)
$ ipsw disk dmg KDK/Kernel_Debug_Kit_13.4.1_build_22F82.dmg -p 4 KDK/hfs_disk
   • Extracted 'disk image (Apple_HFS : 4)' as KDK/hfs_disk
$ ipsw disk hfs KDK/hfs_disk --pattern ".*" --output KDK/
   • Extracted KDK/KernelDebugKit.pkg
$ ipsw pkg KDK/KernelDebugKit.pkg --pattern "System/Library/Kernels/(kernel$|kernel.dSYM/Contents/Resources/DWARF/kernel$)" --output KDK/
   • Checking for files in KDK.pkg/Payload...
      • Extracting KDK/System/Library/Kernels/kernel.dSYM/Contents/Resources/DWARF/kernel
      • Extracting KDK/System/Library/Kernels/kernel
```

#### KernelCache: "relocated_symbols.txt"

See previously.

#### dwarf2json 

No changes:

`./dwarf2json mac --macho KDK/System/Library/Kernels/kernel.dSYM/Contents/Resources/DWARF/kernel`

### Current tools updates 

A new converter is needed, here are suggestions:
- Update dwarf2json `mac` with an additional flag to override symbols addresses with the ones in `symbols.txt`
- Make a small Python wrapper that patches the ISF manually afterwards by iterating over `symbols.txt`

I like the Python wrapper a bit more, as people building the ISF will have to follow this procedure so adding an extra small step shouldn't be much of a concern. It also allows them to keep both a `non_relocated_ISF.json` and `relocated_ISF.json` on hand.

### Caveats

- This process requires a good bandwidth and enough disk space, but hopefully someone could automate it :). 
- I haven't encountered this scenario yet, but if for some reason a recent release variation decides to not load the kernel as an `MH_FILESET`, then the non-relocated ISF might still be the one to go with. 
- ARM64 might follow this procedure, or something completely different (for future ref: `ipsw dl appledb --kernel` ?)

## Conclusion

I think we could definitely get away by patching this issue at root in the ISF, leaving us only with the plugins PR to merge. A bit more work to construct the required data, but headache-free PRs reviews and debugging :).

## Bonus: [CTF - Compact ANSI-C Type Format](https://sources.debian.org/src/gdb/15.1-1/include/ctf.h/)

I discovered that the KDK's DWARF-stripped `kernel` file (as opposed to `kernel.dSYM`) still embeds type and symbols information, like `BTF` does for Linux. Here is a sample output: 

```
$ ipsw --no-color kernel ctfdump ./KDK/kernel > ctfdump.txt
$ cat ctfdump.txt
# ... snip ...
- Types ----------------------------------------------------------------------

[1] INTEGER unsigned char encoding=CHAR offset=0 bits=8

<2> TYPEDEF u_char refers to 1
<3> INTEGER long encoding=SIGNED offset=0 bits=64

[4] ARRAY (anon) content: 2 index: 3 nelems: 6
[5] INTEGER unsigned short encoding=0x0 offset=0 bits=16

<6> TYPEDEF u_short refers to 5
<7> STRUCT ether_header (14 bytes)
    ether_dhost type=4 off=0x0
    ether_shost type=4 off=0x30
    ether_type  type=6 off=0x60
# ... snip ...
- Data Objects ---------------------------------------------------------------

0xffffff8000c1a204: <unknown> .constructors_used
0xffffff8000c1a20c: <unknown> .destructors_used
0xffffff8000106000: pdpt_entry_t[512] BootPDPT
0xffffff8000105000: pml4_entry_t[512] BootPML4
# ... snip ...
- Functions ------------------------------------------------------------------

0xffffff8000b0035c: void Assert();
0xffffff8000867a20: int CURSIG(proc_t);
0xffffff80003928d0: void Debugger();
0xffffff8000392780: kern_return_t DebuggerTrapWithState(debugger_op, const char *, const char *, va_list *, uint64_t, void *, boolean_t, unsigned long);
0xffffff80003928f0: void DebuggerWithContext();
```

### Would it be worth it to make a ctf2json ?

Given that the KDK currently embeds `kernel.dSYM`, not really. However, the KDK is not available for every kernel release, so this could be the only way to build an ISF in some cases (ex: OSX 10.8 build "12A269" does not have a KDK listed).

ikelos:
Wow, awesome work!  So... this seems similar to me to every linux kernel build having a custom `system.map` file?  If that's the case (and mac has a limited, specific number of kernels, which it seems to), I'd far prefer we just packaged them all up with and deprecate mac ISF generated with an old version of dwarf2json (like we had to do with [rust-confused linux ISF files](https://github.com/volatilityfoundation/volatility3/blob/2ffb3ac6b153cc7bdbfa1337d478231058a0d77b/volatility3/schemas/__init__.py#L78) ).

My concern with taking a data file (replaceable) and turning it into *two* data files (or a data file and then an applied process) is that it'll lead to all kinds of possible confusions and bug reports where we have to ask people how they generated it, and basiacally manually check every failed mac image analysis.  The data produced by dwarf2json is inaccurate, we can flag when that happens and get them to replace it, but trying to monkey patch it out for them seems like it won't really help and will just create more work for us.

It looks like ipsw would allow us to grab a copy of the kernel to be able to generate the types (ie, obviate the need for the KDK at all?).  We could build a tool like pdbconv (that, at least initially, relies on ipsw)?  I'd be much happier writing something that (given disk space) could grab the needed data and generate it live (like windows does) with fallback to "you've got to generate this yourself" if the ipsw binary isn't available?  If we can't then ctf2json does sound interesting.  The more we can get from the memory image itself the better (although we still need to support otherwise, since memory images can be corrupted or otherwise).

It might also be worth us building up this library on the ISF server (which we still need to shifted over onto the volatilityfoundation github at some point)?

Abyss-W4tcher:
Just made a small Python PoC that updates the current ISF with relocated_symbols.txt offsets, and the analysis works again (with the additional plugins PR) :).

I think making a "all-in-one" small (bash ?) utility that builds an ISF given it has `dwarf2json` and `ipsw` in its path would be an excellent idea indeed. We could even add a metadata tag that's debugged out so that we know when people used this recommended script or if it was manual (for valid or invalid reasons).

We could make a package, but honestly the ISF server design scales better (we will have to store the ARM64 ones someday, and there are many CPU SOC variants for each release). If people want to have the sets offline they can just git clone it.

Relying exclusively on IPSW *could* be doable if we make a CTF parser, but honestly I'd prefer to go with the KDK + DWARF way instead of allocating resources to it for now ?

Also, discarding the non-relocated symbols addresses could circle back to this "address mismatch" issue if some kernels do not boot as a KernelCache. Ideally, we could store each version in the same ISF file under two "symbols" key, after a dwarf2json version bump?


-----

I'll also do additional testing to ensure this procedure works for all KDKs + IPSW available, but it's good to determine the design to follow now.

Abyss-W4tcher:
In fact, we could copy the `dwarf2json linux --system-map` argument and processing, with only a small parsing adjustment on the relocated_symbolts.txt format.

ikelos:
As long as we've got a little script, and Mac kernels are in some way finite (unlike Linux ones where anyone can roll their own), it might be we just build the new ones and put them on the ISF server.  I *don't* recall whether I was foresighted enough to include the producer information in the local cache database, but I suppose we'll find out at some point when we come to figure out how to blat over existing ISFs with newer ones...  5:S

Yep, a small bash utility sounds good.  I'm not keen on introducing metadata tags for specific purposes, you have to carry them around forever, better to use a different producer/version as the means of tagging.

Yep, I'm happy to stick with just the ISF server too, more musing outloud, but the requirements (memory and tools) are heavy so probably not worth it.

Yep, reusing dwarf2json's existing capabilities is definitely the way to go.

Abyss-W4tcher:
So I've made bash and Python PoCs that pretty much automate everything, it's working pretty good. However, `ipsw` won't be able to fetch KDKs for < 13.0... I have to resort to another script for old releases but unfortunately I'm not sure it is a good idea to make it public (or outside of volatility's internal scope). One could also store the `KernelCache`, `kernel` and `kernel.dSYM` triplet additionally to the ISFs just in case.

> My concern with taking a data file (replaceable) and turning it into two data files (or a data file and then an applied process) is that it'll lead to all kinds of possible confusions and bug reports where we have to ask people how they generated it, and basiacally manually check every failed mac image analysis. 

Isn't this already the case for Linux bug reports, even if there are no double data files ? A good part of the time, someone just compiled a `5.15.100~22.04.1` instead of a `5.15.100` and we still have to debug it with them ?

What if we simply do `_isf.json` and `_isf_relocated.json`, with a strong indication in the procedure that people should always use the relocated one ? We could also name the non relocated one like `json.nonrelocated.donotuse` ? If we see someone reporting a bug with a naming that doesn't follow this convention, we can immediately ask him if he followed the """official""" procedure and then:
- indicate him the ISF server
- ask him to re-generate using the procedure
- debug with him if this is an edge case where the procedure fails

Just trying to make it flexible, because I don't know how it will be for ARM64 and tweaking the ISF format just for macOS can soon end up in spaghetti structures and bypasses. dwarf2json is not to blame here, as it correctly parses the kernel like `dwarfdump` would.

By the way, would Vol3 try an ISF that ends with `json.nonrelocated.donotuse` ? This could prevent overlaps if someone still puts the two files in the symbols directory.

ikelos:
Oh yeah, I'm only suggesting we generate the necessary ISFs, not anything proprietary.  I think we're ok keeping the scripts that pull the data from apple, because they should require someone to sign up to an agreement before they let them access it (much like we have the scripts that operate on the KDK DMG files).

It is already the case that we have difficulty debugging linux, but that's due to the recompilable nature of linux, so we don't need to add Mac to the mix.  Mac and Windows have essentially fixed (finite) kernels.  I still don't think requiring additional files is a good idea, take (for example) `.vmem` files and the difficulty we have not knowing whether there's an associated `.vmsn` or not.  It's just a really bad design decision.  We can differentiate based on the producer data inside the file, but separating the two makes it impossible to know whether what we have is complete or not.  We could extend the schema design to include an additional shift but... that seems like we're just not capturing the right data?  The kernels (being finite) should only ever need one type or the other, not both, right?

I'm concerned about spaghetti code and bypasses which is why I'm trying hard not to change volatility from accepting things the way it accepts them (particularly since we got confirmation it worked normally when provided the right offsets).  If Mac produces dynamic shifts (ie, multiple different things needed), then we'll need to figure that out and calculate it (like we do for the aslr and kaslr type shifts) based on the memory image, but additional static datafiles really makes no sense...

Abyss-W4tcher:
- My concern in keeping both an ISF with relocated symbols addresses and one without, was that if for some reason the kernel wasn't booting as part of a KernelCache then it wouldn't work. However this is a bit absurd, as I think if the kernel wasn't configured as a KernelCache, then there wouldn't be a `KernelCache` file inside the IPSW, so the relocation calculations wouldn't even trigger.
- To identify that the ISF symbols addresses were relocated, we can't just put: 

```json
{
    "metadata": {
        "mac": {
            "symbols": [
                {
                    "kind": "symtab",
                    "name": "KernelCache",
                    "hash_type": "sha256",
                    "hash_value": "cfcd93289c55e1f3b5ed9704bae47f7d38d0a34bff8113df071bf42fa4be6fe1"
                },
```

as it isn't true that we obtained the data from the `KernelCache` directly. The only data we need from it, are the new location of each section (`__TEXT`, `__DATA`, ...) when the kernel will load in memory (minus KASLR).  However, what if we add an entry under the symbols tree, but with a new type like `"kind":"relocation_reference", "name":"KernelCache" etc.`. Doing so, macOS automagic could log a debug message on each run that indicates if the loaded symbol table was relocated or not, to help debugging ?

ikelos:
So... that sounds a little complicated for a very what-if scenario, and it means deviating the ISF format specifically for one type of operating system, which we haven't yet proven is necessary?  I'm happy to review a mock-up, but no guarantees it'll get accepted, so depends if you want to invest the effort in keeping these two things separate?

Abyss-W4tcher:
As I said I think we can discard the what-if scenario of keeping two ISFs, that was not my concern anymore. Here is what currently happens with my scripts (schematically): 

```sh
$ dwarf2json mac --macho ./kernel.dSYM --macho-syms ./kernel > isf.json
$ python3 relocate.py --kernel-cache ./kernelcache --kernel-non-relocated ./kernel --non-relocated-isf ./isf.json --overwrite-isf
```

The relocate.py script will take care of calculating the offsets between `./kernelcache->kernel["segments"]` and `./kernel["segments"]` and adjust the symbols in the ISF. Of course everything could/will happen inside dwarf2json but that's not really the subject for now. 

And that's it, we would get off with a single ISF. However, to clearly indicate that the symbols were relocated (we absolutely need this process as discussed), I thought we could add a reference to the `KernelCache` inside the ISF to know it was used ? 

If someone shouts about a non-functional ISF, but that we detect in a `-vv` run that dwarf2json wasn't provided the `KernelCache`, we can directly pinpoint the issue.

ikelos:
Ok, thanks, I see what you're saying now.  I think just altering the producer metadata should be ok, unless there's specific information we want to add?  We'll basically blacklist dwarf2json mac ISF files before a specific version, and we'll need to get a version that only accepts the correct relocated versions (somehow) or we'll just swap to a completely different producer and blacklist everything mac that dwarf2json produces if it can't prevent people using the old technique, then it'll be a matter of clearly documenting everywhere the necessary process/data to ensure people produce the right files.  We can warn on old mac dwarf2json ISFs and (hopefully) we can warn on bad data being fed to dwarf2json.  Given the extra data under the mac metadata field, and happy to include it there somehow, I'm just trying to reduce the number of moving parts or bits that need changing for this use case...

Abyss-W4tcher:
The tricky part, is that ISFs before Kernel Cache feature (and more specifically `MH_FILESET`) are still valid. Proof is that analysis for samples < macOS 11.0 still work. Vol3 could output a warning based on appropriate checks, like we did here: 

https://github.com/volatilityfoundation/volatility3/blob/37bdb6cb9959f69063afb45d761383092bf5809c/volatility3/schemas/__init__.py#L141

I don't think we can prevent users from running dwarf2json without `--kernel-cache` (what if they still want the non relocated version for non-vol3 related things ?), but as you said we need to clearly document it at multiple places.

I am trying to avoid cluttering dwarf2json with these kind of adjustments as much as possible, but making an additional script to manually patch the ISF will most likely result in people overlooking it even more than `--kernel-cache`.

Abyss-W4tcher:
I am not sure yet that I'll have as much time as I had this week to allocate in this subject, it might be done later. I'll keep you updated and then I'll start working on a dwarf2json PR, vol3 PR and on the documentation. Then I'll give another look at the draft plugins PR and hopefully we'll support macOS analysis fully again :).

PR #1130, opened 2024-04-18
Windows: Add max PID and thread count options for filtering processes

ckusek15:
This PR adds additional flags, `--maxPid` and `--maxThreads`, to filter down the output of the `windows.psscan` plugin.

Using the `--maxPid` flag followed by an integer `n` will only output process info for processes that have a PID <= `n`.

Using the `--maxThreads` flag followed by an integer `n` will only output process info for processes that have a thread count < `n`.

This PR also adds default cap values for the PID and thread count, with the PID being capped at 1000000 by default and the thread count being capped at 100000. Therefore, a process with a PID > 1000000 and/or thread count >= 100000 will now be filtered out by default. This filters out a lot of the junk process results and users are able to set these values higher/lower using the `--maxPid` and `--maxThreads` flags if desired.

Manual testing was performed on multiple windows memory images to verify the expected behavior.

### DIFF
--- a/volatility3/framework/plugins/windows/psscan.py
+++ b/volatility3/framework/plugins/windows/psscan.py
@@ -57,6 +57,18 @@ class PsScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                 default=False,
                 optional=True,
             ),
+            requirements.IntRequirement(
+                name="maxPid",
+                description="List processes with ID <= maxPid",
+                default=1000000,
+                optional=True,
+            ),
+            requirements.IntRequirement(
+                name="maxThreads",
+                description="List processes with thread count < maxThreads",
+                default=100000,
+                optional=True,
+            ),
         ]
 
     @classmethod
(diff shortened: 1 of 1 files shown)


ikelos:
These feel like options better handled by the UI than by the plugin itself?  Since it's already found the process, this won't save a great deal of time or make things more efficient, and with flags like this added to this plugin, they'd need adding to other plugins and they'd all get very messy very quickly.   I'd suggest trying to add something generic to the UI?  There's already scope for filtering columns for particular values, you could improve the list of column filters in include ones that set a maximum value for fields, and that would then work both for threads and pids, and allow you to set minimums as well...

See [text filters](https://github.com/volatilityfoundation/volatility3/blob/c1f239b8d2171e83e331ad51bf7ff6e48ed04e52/volatility3/cli/text_filter.py) for more information.  Unfortunately I'm going to close this PR since we're definitely not going to be adding flags for all the possible little filters people may want when they have no impact on the performance of the plugin.

PR #1131, opened 2024-04-18
Windows: Add context arguments to windows.vadyarascan.

AJohnson24:
This change updated the windows.vadyarascan plugin to accept --size and --reverse arguments, which control the overall size of the hex dump of the YARA match and the negative offset respectively.  Additionally, the Value column was replaced with Hexdump.

This more closely mirrors the functionality of volatility2.  If the offset specified by --reverse is large enough to cause a read on a page not present in the system, a warning is logged and padded data will be returned.

Manual testing was performed on a windows memory image to verify the expected behavior. The windows test suite was also run.

### DIFF
--- a/volatility3/framework/plugins/windows/vadyarascan.py
+++ b/volatility3/framework/plugins/windows/vadyarascan.py
@@ -5,7 +5,7 @@
 import logging
 from typing import Iterable, List, Tuple
 
-from volatility3.framework import interfaces, renderers
+from volatility3.framework import exceptions, interfaces, renderers
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints
 from volatility3.plugins import yarascan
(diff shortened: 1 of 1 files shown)


ikelos:
This doesn't seem the right place for these flags?  The size of the hexdump *should* somewhat be UI based (I realize the way it's coded at the moment that's quite difficult to do).  This puts it in the hands of the plugin.

As it stands, Volatility is supposed to give you the offset in the dump to help you get to it, it isn't supposed to extract all the data, and the hex dump is just supposed to show people the context of the data retrieved.  I'd fully expect another tool to consume and extract the right amount of data if more were required?  The hexdump is already fairly generous (I think it's 100 bytes by default)...

If we wanted to allow configurable sizes for the return of binary data, we'd need to weave that into the communication channel between the UI and the framework, so that the UI could specify how much data it wants.  That's a much bigger patch than this one, and will take some time to get right.

Also, volatility 3 is not attempting to blindly mirror the functionality of volatility 2.  If there's a reason for doing something, we'll do it, whether or not volatility 2 did should not play a part in the decision.  Volatility 2 did some things right, but also some things wrong.  What we're trying to do is make volatility 3 a good forensics framework, if these help that then fine, if not then we won't do them.

As such, I'm going to close this off, but if you'd like to get a feature like this into volatility, please investigate the configuration system and work up a patch for passing global values (such as requested size for byte outputs) so that all plugins could make use of it.

PR #1141, opened 2024-05-06
Layers - Several fixes around maximum_address and chunk sizes

gcmoreira:
## `maximum_address` fixes in several layers
I think there are still other places to fix but I haven't fixed everything as testing it will be massive. I think we should review all occurrences of `layer.maximum_address` and where we calculate the data size based on that value. There are other places where I suspect it's being used incorrectly. To mention just a few:
- [PdbMSFStream::maximum_address](https://github.com/volatilityfoundation/volatility3/blob/1c5c57e1550f156844543ff2e54660095e75eb2c/volatility3/framework/layers/msf.py#L257) looks like it's also wrongly calculated as `len(self._pages) * self._pdb_layer.page_size` should be the next byte following the `maximum_address`.
- Additionally, there are other places that require double-checking. For instance, everywhere the `maximum_address` is in a while condition but the value is not included in the address range i.e.: [lime](https://github.com/volatilityfoundation/volatility3/blob/1c5c57e1550f156844543ff2e54660095e75eb2c/volatility3/framework/layers/lime.py#L45),  or [avml](https://github.com/volatilityfoundation/volatility3/blob/1c5c57e1550f156844543ff2e54660095e75eb2c/volatility3/framework/layers/avml.py#L102). Technically looks wrong but I think these could still be ok because they are expecting more than 1 byte to read anyway. So it should mean that the layer/file is truncated/bad format/etc.
## CommandLine: file_handler_class_factory ->CLIFileHandler
- Fixed issue when the filename doesn't contain an extension: ~~For instance, if we call the LayerWriter plugin with --output aaa ... it creates a .aaa (hidden filename in linux). Next iterations using the same argument will generate -1.aaa, -2.aaa, etc. instead of aaa, aaa-1, and aaa-2 respectively.~~
- ~~Stored the real filename created in `file_handle._output_filename`. That way we can know the final filename including both the "--output-dir" and the "--output" values and its "-1", "-2", etc (if used). Otherwise, the filename reported will be incorrect (and old version of this file), which could result in inaccurate analysis. _Unfortunately, the final filename is determined when the file is closed, which might lead to some controversy regarding its implementation._~~
- EDIT: Stored the real filename (only the base name) created in `file_handle.preferred_filename`. Otherwise, the filename reported will be incorrect (an old version of this file), which could result in inaccurate analysis. The final filename will contain the layer name. i.e. `primary.raw`, `primary-1.raw`, etc
## LayerWriter plugin. Several fixes
~~- Fixed missing `--output` argument and other bugs, plus some code improvements.~~

### DIFF
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -789,8 +789,16 @@ class CommandLine:
                 if self._file.closed:
                     return None
 
-                self._file.close()
                 output_filename = self._get_final_filename()
+
+                # Update the filename, which may have changed if a file with
+                # the same name already existed. This needs to be done before
+                # closing the file, otherwise FileHandlerInterface will raise
+                # an exception. Also, the preferred_filename setter only allows
+                #  a specific set of characters, where '/' is not in that list
+                self.preferred_filename = os.path.basename(output_filename)
+
+                self._file.close()
                 os.rename(self._name, output_filename)
 
         if direct:
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -716,19 +716,17 @@ class CommandLine:
                 """Gets the final filename"""
                 if output_dir is None:
                     raise TypeError("Output directory is not a string")
+
                 os.makedirs(output_dir, exist_ok=True)
 
-                pref_name_array = self.preferred_filename.split(".")
-                filename, extension = (
-                    os.path.join(output_dir, ".".join(pref_name_array[:-1])),
-                    pref_name_array[-1],
-                )
-                output_filename = f"{filename}.{extension}"
+                output_filename = os.path.join(output_dir, self.preferred_filename)
+                filename, extension = os.path.splitext(output_filename)
 
                 counter = 1
                 while os.path.exists(output_filename):
-                    output_filename = f"{filename}-{counter}.{extension}"
+                    output_filename = f"{filename}-{counter}{extension}"
                     counter += 1
+
                 return output_filename
 
         class CLIMemFileHandler(io.BytesIO, CLIFileHandler):
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -790,8 +790,8 @@ class CommandLine:
                     return None
 
                 self._file.close()
-                output_filename = self._get_final_filename()
-                os.rename(self._name, output_filename)
+                self._output_filename = self._get_final_filename()
+                os.rename(self._name, self._output_filename)
 
         if direct:
             return CLIDirectFileHandler
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -790,8 +790,8 @@ class CommandLine:
                     return None
 
                 self._file.close()
-                self._output_filename = self._get_final_filename()
-                os.rename(self._name, self._output_filename)
+                output_filename = self._get_final_filename()
+                os.rename(self._name, output_filename)
 
         if direct:
             return CLIDirectFileHandler
--- a/volatility3/framework/plugins/layerwriter.py
+++ b/volatility3/framework/plugins/layerwriter.py
@@ -18,7 +18,7 @@ class LayerWriter(plugins.PluginInterface):
     default_block_size = 0x500000
 
     _required_framework_version = (2, 0, 0)
-    _version = (2, 0, 0)
+    _version = (2, 0, 1)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
@@ -81,7 +81,7 @@ class LayerWriter(plugins.PluginInterface):
 
         file_handle = open_method(preferred_name)
         for i in range(0, layer.maximum_address, chunk_size):
-            current_chunk_size = min(chunk_size, layer.maximum_address - i)
+            current_chunk_size = min(chunk_size, layer.maximum_address + 1 - i)
             data = layer.read(i, current_chunk_size, pad=True)
             file_handle.write(data)
             if progress_callback:
--- a/volatility3/framework/plugins/layerwriter.py
+++ b/volatility3/framework/plugins/layerwriter.py
@@ -41,6 +41,7 @@ class LayerWriter(plugins.PluginInterface):
             requirements.StringRequirement(
                 name="output",
                 description="Output filename",
+                optional=True,
             ),
             requirements.ListRequirement(
                 name="layers",
--- a/volatility3/framework/plugins/layerwriter.py
+++ b/volatility3/framework/plugins/layerwriter.py
@@ -38,6 +38,10 @@ class LayerWriter(plugins.PluginInterface):
                 default=False,
                 optional=True,
             ),
+            requirements.StringRequirement(
+                name="output",
+                description="Output filename",
+            ),
             requirements.ListRequirement(
                 name="layers",
                 element_type=str,
--- a/volatility3/framework/plugins/layerwriter.py
+++ b/volatility3/framework/plugins/layerwriter.py
@@ -115,6 +115,10 @@ class LayerWriter(plugins.PluginInterface):
                             progress_callback=self._progress_callback,
                         )
                         file_handle.close()
+
+                        # Update the filename, which may have changed if a file
+                        # with the same name already existed.
+                        output_name = file_handle.preferred_filename
                     except IOError as excp:
                         yield 0, (f"Layer cannot be written to {output_name}: {excp}",)
 
--- a/volatility3/framework/plugins/layerwriter.py
+++ b/volatility3/framework/plugins/layerwriter.py
@@ -115,9 +115,7 @@ class LayerWriter(plugins.PluginInterface):
                         )
                         file_handle.close()
                     except IOError as excp:
-                        yield 0, (
-                            f"Layer cannot be written to {self.config['output_name']}: {excp}",
-                        )
+                        yield 0, (f"Layer cannot be written to {output_name}: {excp}",)
 
                     yield 0, (f"Layer has been written to {output_name}",)
 
--- a/volatility3/framework/plugins/layerwriter.py
+++ b/volatility3/framework/plugins/layerwriter.py
@@ -115,7 +115,6 @@ class LayerWriter(plugins.PluginInterface):
                             progress_callback=self._progress_callback,
                         )
                         file_handle.close()
-                        output_name = file_handle._output_filename
                     except IOError as excp:
                         yield 0, (f"Layer cannot be written to {output_name}: {excp}",)
 
--- a/volatility3/framework/plugins/layerwriter.py
+++ b/volatility3/framework/plugins/layerwriter.py
@@ -119,6 +119,7 @@ class LayerWriter(plugins.PluginInterface):
                             progress_callback=self._progress_callback,
                         )
                         file_handle.close()
+                        output_name = file_handle._output_filename
                     except IOError as excp:
                         yield 0, (f"Layer cannot be written to {output_name}: {excp}",)
 
--- a/volatility3/framework/plugins/layerwriter.py
+++ b/volatility3/framework/plugins/layerwriter.py
@@ -38,11 +38,6 @@ class LayerWriter(plugins.PluginInterface):
                 default=False,
                 optional=True,
             ),
-            requirements.StringRequirement(
-                name="output",
-                description="Output filename",
-                optional=True,
-            ),
             requirements.ListRequirement(
                 name="layers",
                 element_type=str,
--- a/volatility3/framework/plugins/layerwriter.py
+++ b/volatility3/framework/plugins/layerwriter.py
@@ -95,7 +95,7 @@ class LayerWriter(plugins.PluginInterface):
             if not self.config["layers"]:
                 self.config["layers"] = []
                 for name in self.context.layers:
-                    if not self.context.layers[name].metadata.get("mapped", False):
+                    if "mapped" not in self.context.layers[name].metadata:
                         self.config["layers"] = [name]
 
             for name in self.config["layers"]:
@@ -103,7 +103,8 @@ class LayerWriter(plugins.PluginInterface):
                 if name not in self.context.layers:
                     yield 0, (f"Layer Name {name} does not exist",)
                 else:
-                    output_name = self.config.get("output", ".".join([name, "raw"]))
+                    default_output_name = f"{name}.raw"
+                    output_name = self.config.get("output", default_output_name)
                     try:
                         file_handle = self.write_layer(
                             self.context,
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -180,7 +180,7 @@ class Intel(linear.LinearlyMappedLayer):
         position = self._initial_position
         entry = self._initial_entry
 
-        if self.minimum_address > offset > self.maximum_address:
+        if not (self.minimum_address <= offset <= self.maximum_address):
             raise exceptions.PagedInvalidAddressException(
                 self.name,
                 offset,
--- a/volatility3/framework/layers/segmented.py
+++ b/volatility3/framework/layers/segmented.py
@@ -152,7 +152,7 @@ class NonLinearlySegmentedLayer(
             raise ValueError("SegmentedLayer must contain some segments")
         if self._maxaddr is None:
             mapped, _, length, _ = self._segments[-1]
-            self._maxaddr = mapped + length
+            self._maxaddr = mapped + length - 1
         return self._maxaddr
 
     @property


gcmoreira:
I was wrong with e5a5b895771b655d21c36689c33a534034c31e36, it should be fine. Anyway, I think it is a more common way to express this kind of check and is generally considered more readable.

gcmoreira:
Tests:
## Vmem image
### Vol2 
```shell
$ python2 ../volatility2/vol.py imagecopy -f ../ubuntu2004lts64bit/ubuntu2004lts64bit-Snapshot35.vmem -O ubuntu2004lts64bit-Snapshot35.vmem.raw.vol2
Volatility Foundation Volatility Framework 2.6
Writing data (5.00 MB chunks): |.......................................................................................................|
```
### Vol3
```shell
$ python3 ../volatility3/vol.py -f ../ubuntu2004lts64bit/ubuntu2004lts64bit-Snapshot35.vmem LayerWriter --output ubuntu2004lts64bit-Snapshot35.vmem.raw.vol3
Volatility 3 Framework 2.7.0
Status
Progress:   99.61		Writing layer primary                  
Layer has been written to /home/user/vol3-imagecopy/ubuntu2004lts64bit-Snapshot35.vmem.raw.vol3
```

## Test
```shell
-rw------- 1 user user 536870912 May 15  2023 /media/user/ubuntu2004lts64bit/ubuntu2004lts64bit-Snapshot35.vmem
-rw-rw-r-- 1 user user 536870912 May  6 16:09 ubuntu2004lts64bit-Snapshot35.vmem.raw.vol2
-rw------- 1 user user 536870912 May  6 16:09 ubuntu2004lts64bit-Snapshot35.vmem.raw.vol3

$ sha1sum ubuntu2004lts64bit-Snapshot35.vmem.raw.vol2 ubuntu2004lts64bit-Snapshot35.vmem.raw.vol3
48206033995ab862fdf729ed4bf2a8bd104654ab  ubuntu2004lts64bit-Snapshot35.vmem.raw.vol2
48206033995ab862fdf729ed4bf2a8bd104654ab  ubuntu2004lts64bit-Snapshot35.vmem.raw.vol3
```
## Core image
### Vol2 
```shell
$ python2 ../volatility2/vol.py imagecopy -f ./ram-4.10.0-42.core -O ram-4.10.0-42.core.raw.vol2
Volatility Foundation Volatility Framework 2.6
Writing data (5.00 MB chunks): |...........................................................................................................................................................................................................................................|
```
### Vol3
```shell
+ ./vol.py -f ./ram-4.10.0-42.core LayerWriter --output ram-4.10.0-42.core.raw.vol3
Volatility 3 Framework 2.7.0
Status
Progress:   99.98		Writing layer primary                   
Layer has been written to /home/user/vol3-imagecopy/ram-4.10.0-42.core.raw.vol3
```
## Test
```shell
-rw-r--r-- 1 user user 1208166496 May  6 07:51 ./ram-4.10.0-42.core
-rw-rw-r-- 1 user user 4294967296 May  6 16:10 ram-4.10.0-42.core.raw.vol2
-rw------- 1 user user 4294967296 May  6 16:10 ram-4.10.0-42.core.raw.vol3

$ sha1sum ram-4.10.0-42.core.raw.vol2 ram-4.10.0-42.core.raw.vol3
36a504a0f1e203a6d6cb9bc1a4c0657fc3a47c97  ram-4.10.0-42.core.raw.vol2
36a504a0f1e203a6d6cb9bc1a4c0657fc3a47c97  ram-4.10.0-42.core.raw.vol3

```

gcmoreira:
@ikelos LayerWriter plugin needs #1142 to be fixed

volatility3/cli/__init__.py
```diff
@@ -792,8 +790,8 @@ def close(self):
                     return None
 
                 self._file.close()
-                output_filename = self._get_final_filename()
-                os.rename(self._name, output_filename)
+                self._output_filename = self._get_final_filename()
```
ikelos: Self variables should be defined (and set to a default value, even if it's `None`) in the object's `__init__` method.

volatility3/framework/plugins/layerwriter.py
```diff
@@ -95,15 +100,16 @@ def _generator(self):
             if not self.config["layers"]:
                 self.config["layers"] = []
                 for name in self.context.layers:
-                    if not self.context.layers[name].metadata.get("mapped", False):
+                    if "mapped" not in self.context.layers[name].metadata:
```
ikelos, line 98: Not sure why this was changed, but I don't particularly have a problem with it.

volatility3/framework/plugins/layerwriter.py
```diff
@@ -95,15 +100,16 @@ def _generator(self):
...
                         self.config["layers"] = [name]
 
             for name in self.config["layers"]:
                 # Check the layer exists and validate the output file
                 if name not in self.context.layers:
                     yield 0, (f"Layer Name {name} does not exist",)
                 else:
-                    output_name = self.config.get("output", ".".join([name, "raw"]))
+                    default_output_name = f"{name}.raw"
```
ikelos, line 106: Happy to do this for clarity, just wondering if there's any other reason for it?

volatility3/framework/plugins/layerwriter.py
```diff
@@ -114,10 +120,9 @@ def _generator(self):
                             progress_callback=self._progress_callback,
                         )
                         file_handle.close()
+                        output_name = file_handle._output_filename
```
ikelos: Accessing a private member is bad.  If it's supposed to be exposed it should be public.

volatility3/cli/__init__.py
ikelos: I'd almost sooner this sets the preferred filename *just* before the file is actually closed.  I think the intention was that it would act as the filename that had been output, based on https://github.com/volatilityfoundation/volatility3/blob/a66aa020507ec3486ceef5ed73e43614856403d4/volatility3/framework/interfaces/plugins.py#L41.  The comment suggests that this value would be set to the correct filename once the file is closed, could be done on this line instead of creating a new (private) variable.  It looks like we never did that, but that was the intention.

ikelos (commented):
Happy with most of it, just the way of exposing the finally written filename.  I'm happy to change the API if needed, or to use the `preferred_filename` value which was the original intention if it's not too confusing, but it needs to be one of those two options.  Accessing a new uninitialized private variable is a no-no...

volatility3/framework/plugins/layerwriter.py
```diff
@@ -95,15 +100,16 @@ def _generator(self):
...
                         self.config["layers"] = [name]
 
             for name in self.config["layers"]:
                 # Check the layer exists and validate the output file
                 if name not in self.context.layers:
                     yield 0, (f"Layer Name {name} does not exist",)
                 else:
-                    output_name = self.config.get("output", ".".join([name, "raw"]))
+                    default_output_name = f"{name}.raw"
+                    output_name = self.config.get("output", default_output_name)
```
ikelos, line 107: You could use `output_name = self.config.get("output") or default_output_name`

volatility3/framework/plugins/layerwriter.py
```diff
@@ -38,6 +38,11 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                 default=False,
                 optional=True,
             ),
+            requirements.StringRequirement(
```
ikelos: Please also bear in mind, plugins should be UI agnostic.  This is a very specific option to get a particular preferred filename chosen for the output.  It pretty much only applies to the CLI because a web interface would have you download the output and choose the filename at the time you did that.  I really don't want to clutter plugins with cosmetic options that are catering to a single UI because that reinforces that there's only one UI anyone bothers with.  Luckily, this does use the FileHandler interface, so other UIs can gather and decide how to deal with files, but just as a reminder, the filename is just a preferred filename.

gcmoreira: Mh .. just to clarify, I'm not the author of this plugin and it wasn't my idea to add this option. What I did is to fix an option that is missing as the code below is expecting that.

Following the commits, it seems that, for some reason, you removed the LayerWriter plugin option in [Plugins: Make layerwiter more configurable](https://github.com/volatilityfoundation/volatility3/commit/47ff37b07b79a44a30249a243b095a0edfc1845a), but the option is still expected in the code. Not sure if you forgot to remove that as well or what happened there but it looks like something went wrong there.

volatility3/framework/plugins/layerwriter.py
gcmoreira: eehh .. yes, that will work but that's not very Pythonic... Python's get() has already a way to handle a default value and an experience Python developer will assume that behavior. Actually, whoever wrote line 106 has already fallen into this issue. IMO this is bad and prone to errors, as we can see here.

ikelos: It was probably me to be fair.  5;P  I think `or as default` is pythonic, but you're right it's not as clean as it could be and can lead to confusion.  Get and a default works better, but I'm not immediately seeing the issue with 106 that this new code resolves?

volatility3/framework/plugins/layerwriter.py
ikelos: Totally understood, sorry if my wording was personalized.  I was responding to the idea and how it relates to what we have at the moment, not necessarily you in particular, I apologize.  I just started changing my mind as the review went on and by this point I felt quite strongly against the concept I think.  I didn't spot that the plugin expected an output. 

Having said that, since layerwriter can now accept and write out multiple layers at a time, a single output file wouldn't be right, so I don't think giving the user the choice of output name makes sense at all any more.  We should name each layer after the layer's `name` and write it out that way.  The user will then know what the file is finally called from the plugin output (although there's changes needed for that included in this PR I think) so can locate and rename it appropriately.  We don't really have any other easily retrievable information to put in the name (such as filename, since the lowest layer might not be a FileLayer and there is no unified way of extracting a source from it).

volatility3/framework/plugins/layerwriter.py
ikelos: Oh, I see, if name isn't set, then it won't make `.raw`, it'll just make `raw`.  Yep, my bad, the format string is a much better way of doing it.  Not sure it needs a whole extra variable, but happy with that.

ikelos:
In future it might be easier to submit these as separate, smaller PRs, just so that the trivial ones are easy to put through, and don't get held up by any of the others.  The `maximum_address` fix seems really valuable, but is being held up by the output stuff in layerwriter...

gcmoreira:
Cool! I used commits for that. Each commit addresses a specific thing. I always try my best to keep them separated and organized.  Sorry, I thought that way will also work for you. 

There are only two commits ( 6d22347ce6dd0152746564a2d77022ca1b4d9045 and d7aae3a9828ba03bd5531aa385724a954e48501a) that are not ok. Would you be okay with cherry-picking the commits, or would you prefer if I reverted those ones?

In summary:
## Ignore:
- 6d22347ce6dd0152746564a2d77022ca1b4d9045
- d7aae3a9828ba03bd5531aa385724a954e48501a

## Merge
- f116c08a7ec60f62e3ef931de7639e402f421d65 Fix bug in segmented layers i.e. core elf.
- a6c77c488436c7b05040b5bf475be79cb59457f3 Fix bug in LayerWriter
- c7f29936dcfb81bb9ae3fe33e261db2bb7c3be85 Unless you want to remove all the functionality in CLIMemFileHandler::close() this is still a bug.
- 32a5f131fd22d5351eee264fd6e58420a1458e9c Still a bug, self.config['output_name'] doesn't exist if you don't want to merge 6d22347ce6dd0152746564a2d77022ca1b4d9045. Otherwise, it will always print "Layer cannot be written to None: \<exception\>"

## Up to you but I think these improves the Python code readability.
- e5a5b895771b655d21c36689c33a534034c31e36
- 1246b20b6521e11110038fe091af912e2b5f4439

volatility3/framework/plugins/layerwriter.py
gcmoreira: Yeah up to you. Just to avoid unnecessary code. **"x" not in**  already does the get("x", False), which not only the `get()` but also the `False` are not needed.

volatility3/framework/plugins/layerwriter.py
gcmoreira: hm.. that's not a private member (_ vs __), you mean `internal`? Anywyay, you are absolutely right that in the case we want to access to that filename we need to make it public

volatility3/framework/plugins/layerwriter.py
gcmoreira: Well again, just to keep things clear and avoid any extra code. 
IMO instantiate a list to then join the two elements with a dot and instantiate the final string is a bit excessive for just joining two static strings ("." and "raw") to a variable. Anyway, the main reason is clarity  `f"{name}.raw"` vs `".".join([name, "raw"])`

ikelos:
I think I'd prefer you to revert them, just so I can do it through the github UI easily, if that's ok?  I'm ok with all the ones except the two you marked as ignore (but again, I'll want to check through the whole thing as one to make sure it makes sense, if that's ok)...

volatility3/framework/plugins/layerwriter.py
ikelos: Yeah, I meant protected going by java nomenclature, but basically means should only be accessed from within the class...

gcmoreira:
Ok, reverted those commits and redid it using the `preferred_filename` as you suggested. See example below.
It was a bit tricky as `preferred_filename` has a setter. The file has to be still open to change it. Additionally, the character set used cannot contain "/" .. so, it seems it was intended to be the `base name`.

```console
$ ./vol.py -f ./ram-4.10.0-42.core LayerWriter --layer primary 
Volatility 3 Framework 2.7.0
Status
Progress:   99.98               Writing layer primary                   
Layer has been written to primary.raw

$ ./vol.py -f ram-5.8.0-53.vmem LayerWriter 
Volatility 3 Framework 2.7.0
Status
Progress:   99.61               Writing layer primary                  
Layer has been written to primary-1.raw

$ ./vol.py -f ram-5.8.0-53.vmem LayerWriter 
Volatility 3 Framework 2.7.0
Status
Progress:   99.61               Writing layer primary                  
Layer has been written to primary-2.raw

$ rm primary.raw

$ ./vol.py -f ./ram-4.10.0-42.core LayerWriter --layer primary 
Volatility 3 Framework 2.7.0
Status
Progress:   99.98               Writing layer primary                   
Layer has been written to primary.raw
```

gcmoreira:
@ikelos Another thing I noticed and would like to check with you is the handling of the temporary file. Before renaming it to the final filename, it’s created in the current directory instead of the system’s temporary directory. This might have been intentional since these files are usually large, but I'm not entirely sure. Thoughts?

gcmoreira:
Last but not least, do you have a way to test what I mentioned in the description? maybe @atcuno can help with that
- [PdbMSFStream::maximum_address](https://github.com/volatilityfoundation/volatility3/blob/1c5c57e1550f156844543ff2e54660095e75eb2c/volatility3/framework/layers/msf.py#L257) looks like it's also wrongly calculated as len(self._pages) * self._pdb_layer.page_size should be the next byte following the maximum_address.
I think it should be:
```diff
diff --git a/volatility3/framework/layers/msf.py b/volatility3/framework/layers/msf.py
index 8d84a774..6ad6b1fa 100644
--- a/volatility3/framework/layers/msf.py
+++ b/volatility3/framework/layers/msf.py
@@ -253,8 +253,11 @@ class PdbMSFStream(linear.LinearlyMappedLayer):
 
     @property
     def maximum_address(self) -> int:
-        return self.config.get(
-            "maximum_size", len(self._pages) * self._pdb_layer.page_size
+        return (
+            self.config.get(
+                "maximum_size", len(self._pages) * self._pdb_layer.page_size
+            )
+            - 1
         )
 
     @property
```
`maximum_size` and `maximum_address` are not equivalent. `maximum_address = maximum_size - 1`

ikelos (approved):
All looks good now, thanks for putting in the requested changes.  5:)

ikelos:
Lemme know if you're happy for this to be merged as is, or if you want to throw in the fix for maximum_size/maximum_address?

ikelos:
> Ok, reverted those commits and redid it using the `preferred_filename` as you suggested. See example below. It was a bit tricky as `preferred_filename` has a setter. The file has to be still open to change it. Additionally, the character set used cannot contain "/" .. so, it seems it was intended to be the `base name`.

It is, because a path makes no sense if this is being handled by a web user interface, for example.  The whole framework *has* to be seen through a UI agnostic lens.  The way of returning files can be dealt with by a UI however it wants, but we can't make assumptions about the results being tied to a filesystem.

Happy for the temporary file to be constructed in a temporary directory (although kept in memory may be bad depending on how much data someone wants to write to the file).

I don't know enough about the test you're suggesting, but I'm happy to accept your changes if you say they're needed.  Might be best to check with @atcuno, but I'm happier with the patch now.  Thanks for making the requested changes...  5:)

gcmoreira:
Yeah, probably better merge it as is, and we can always create a new ticket for the other changes. Thanks

Issue #542, opened 2021-07-26
timeliner.Timeliner body file is incompatible with log2timeline


BeanBagKing:
**Describe the bug**
The body file created by the timeliner.Timeliner Volatility3 plugin is incompatible with Plaso's "log2timeline.py --parsers="mactime"". This parser seems to expect all (or at least most) columns to have data in them.

**Context**
Volatility Version:  1.0.1
Operating System:  Ubuntu 20.04 (WSL2)
Python Version:  Python 3.8.10
Suspected Operating System:  Windows 10
Command:  python3 vol.py -f physical-memory.img timeliner.Timeliner --create-bodyfile

**To Reproduce**

1. Dump memory and disk (e.g. raw or e01) from a system
2. Use vol.py to create a body file with the memory, e.g. "python3 vol.py -f physical-memory.img timeliner.Timeliner --create-bodyfile"
3. Use log2timeline.py to create a plaso file, e.g. "log2timeline.py plaso.dump diskimage.raw"
4. Note the total number of **events** (not event sources) in the plaso.dump file, "pinfo.py plaso.dump"
5. Attempt to combine the plaso.dump file and the volatility body file, e.g. "log2timeline.py --parsers="mactime" plaso.dump volatility.body"
6. Run "pinfo.py plaso.dump" again and note that the total number of events does not change.

Edit the volatility.body file and add something (such as a 0) into every empty field, e.g.
```
|PsList - Process: 343 services.exe (902309408234)|||||||||1627261200
```
becomes
```
0|PsList - Process: 343 services.exe (902309408234)|0|0|0|0|0|0|0|0|1627261200
```
and attempt lines 4-6 in the steps to reproduce above. Note that this time the total events increases as expected.

**Expected behavior**
The plaso.dump total events increases by the number of events in the volatility body file.

**Screenshots**
n/a

**Additional information**
I'm not aware of any RFC/specification for what columns are required and which are optional in the body file. However, the example provided by Sluthkit (albeit v2) does have all fields completed with some form of apparent placeholder in "empty" fields (note the 0 in the MD5 column here: https://wiki.sleuthkit.org/index.php?title=Body_file)

With that in mind, I'm not positive if Volatility3 is violating any specification by not including mandatory rows, or if log2timeline is not following specification by refusing to ingest body files that don't contain optional rows. Based on what I can find though, I think it would be best if Volatility adjusted their output.

ikelos:
[...]

PR #1144, opened 2024-05-09
Add a default 0 value to macb columns to avoid mactime not reporting …

atcuno:
…timeline entries

I noticed this issue when creating timelines and looking for process related entries in the mactime output. In the current code, the output of process plugins (pslist, psscan) is not included in the output files of body-file processing tools like mactime.
 
I later determined that the root cause was a very similar issue to this old ticket:

https://github.com/volatilityfoundation/volatility3/issues/542

In 542, the fix was to fill in non-timestamp columns that a plugin may not fill in as otherwise mactime (and other body file parsers) will ignore the line and not report it in the final timeline output. Processes are still not reported by mactime with the current vol3 code though as if a plugin does not fill in all columns, such as pslist only filling in the create (birth) time, then vol3 leaves the other columns as empty - making mactime skip them. My patch forces a default of 0 for these entries so the related lines will be included by body file processing tool.

### DIFF
--- a/volatility3/framework/plugins/timeliner.py
+++ b/volatility3/framework/plugins/timeliner.py
@@ -183,16 +183,16 @@ class Timeliner(interfaces.plugins.PluginInterface):
                                     plugin_name,
                                     self._sanitize_body_format(item),
                                     self._text_format(
-                                        times.get(TimeLinerType.ACCESSED, "")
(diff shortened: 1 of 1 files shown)


ikelos (approved):
Yep, looks resonable (they should porbably have been `AbsentValues` anyway, but I'm happy making them 0 if that helps things process them).

PR #1145, opened 2024-05-09
Correctly check for a failed read

atcuno:
A backtrace was triggered in handles testing due to this read() call being checked for None instead of properly in a try/except.


ikelos (approved):
Nice catch, looks good!

PR #1149, opened 2024-05-19
Add a new getcellroutine plugin that reports hooked GetCellRoutine ha…

atcuno:
…ndlers of memory mapped Windows registry hives

This detects real-world rootkits such as https://www.fortinet.com/blog/threat-research/deep-panda-log4shell-fire-chili-rootkits, which is shown here:

```
$ python3 vol.py -r pretty -f crtsys.lime windows.registry.getcellroutine
Volatility 3 Framework 2.7.0
Formatting...0.00		PDB scanning finished                        
  |    Hive Offset |                Hive Name | GetCellRoutine Module | GetCellRoutine Handler
* | 0x9a869403d000 | \REGISTRY\MACHINE\SYSTEM |                crtsys |         0xf800f85d4cf0
```

### DIFF
--- a/volatility3/framework/plugins/windows/registry/getcellroutine.py
+++ b/volatility3/framework/plugins/windows/registry/getcellroutine.py
@@ -1,0 +1,97 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import logging
+from typing import List
+
+from volatility3.framework import constants, exceptions, interfaces, renderers
+from volatility3.framework.configuration import requirements
+from volatility3.framework.renderers import format_hints
+from volatility3.plugins.windows import ssdt
+from volatility3.plugins.windows.registry import hivelist
+
+vollog = logging.getLogger(__name__)
+
+class GetCellRoutine(interfaces.plugins.PluginInterface):
+    """ Reports registry hives with a hooked GetCellRoutine handler """
+
+    _required_framework_version = (2, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Windows kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.PluginRequirement(
+                name="hivelist", plugin=hivelist.HiveList, version=(1, 0, 0)
+            ),
+            requirements.PluginRequirement(
+                name="ssdt", plugin=ssdt.SSDT, version=(1, 0, 0)
+            ),
+     ]
+
+    def _generator(self):
+        kernel = self.context.modules[self.config["kernel"]]
+
+        collection = ssdt.SSDT.build_module_collection(
+            self.context, kernel.layer_name, kernel.symbol_table_name
+        )
+
+        # walk each hive and validate that the GetCellRoutine handler
+        # is inside of the kernel (ntoskrnl)
+        for hive_object in hivelist.HiveList.list_hives(
+            context=self.context,
+            base_config_path=self.config_path,
+            layer_name=kernel.layer_name,
+            symbol_table=kernel.symbol_table_name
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/registry/getcellroutine.py
```diff
@@ -0,0 +1,98 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
```
eve-mem: Extremely unimportant but the 2019 feels like a simple copy and paste mix-up, is it worth changing to 2024?

volatility3/framework/plugins/windows/registry/getcellroutine.py
```diff
@@ -0,0 +1,98 @@
...
+            context=self.context,
+            base_config_path=self.config_path,
+            layer_name=kernel.layer_name,
+            symbol_table=kernel.symbol_table_name,
+        ):
+            hive = hive_object.hive
+
+            try:
+                cellroutine = hive.GetCellRoutine
+            except exceptions.InvalidAddressException:
```
eve-mem, line 57: Do you think it's worth adding a debug log line here to show that this hive is being skipped?

eve-mem (commented):
Always great to see new plugins. 😃 Great work as always @atcuno 

I've made some tiny comments - see what you think. Nothing actually important, but I'm always a fan of extra logging where possible. 

🦊

volatility3/framework/plugins/windows/registry/getcellroutine.py
```diff
@@ -0,0 +1,98 @@
...
+                continue
+
+            module_symbols = list(
+                collection.get_module_symbols_by_absolute_location(cellroutine)
+            )
+
+            if module_symbols:
+                for module_name, _ in module_symbols:
+                    # GetCellRoutine handlers should only be in the kernel
+                    if module_name not in constants.windows.KERNEL_MODULE_NAMES:
```
eve-mem, line 67: Also here actually - is it worth a debug log to say that `ntoskrnl` etc is being skipped? Useful to explain when this plugin doesn't show any results at all.

volatility3/framework/plugins/windows/registry/getcellroutine.py
atcuno: I will fix this. I started by making a copy of userassist.py since it had all the imports needed and call to hivelist

volatility3/framework/plugins/windows/registry/getcellroutine.py
atcuno: I usually don't log if smear is the reason that it would be broken. If it a rootkit set it to a bad address on the running system then the system would crash, which obviously isn't very helpful.

volatility3/framework/plugins/windows/registry/getcellroutine.py
atcuno: Its one of the plugin types like malfind that only show when something malicious is present, so I only wanted it to print when something is off.

eve-mem: Makes perfect sense. I just mean a debug line when -vvvvvvv etc is used. No point in adding it if you think it's not useful of course :)

volatility3/framework/plugins/windows/registry/getcellroutine.py
eve-mem: I do exactly the same thing every time 😅

volatility3/framework/plugins/windows/registry/getcellroutine.py
eve-mem: Makes sense to me 👍

ikelos: Might still be best to `LOGLEVEL_VVVV` just so that someone trying to figure out what's going on can determine that it got skipped?  Normal debug should be `-vvv` so shouldn't show.

ikelos (commented):
Yep, generally looks good, just a thing about the logging but happy to go either way with it, just let me know if you want to throw it in...

PR #1150, opened 2024-05-20
Generic: Add vmscan plugin

### LINKED ISSUE #464: VM Scan

ikelos:
Adds in a vmscan plugin that will search for a VMCS structure using profile information (currently profiles only exist for skylake and haswell architectures, but more can be added).  After we've found the VMCS location, we can determine the page table and load a layer on top of it, but we'll need to figure out how to then let people run plugins against those virtual images.  Probably by a config, but the config system may need rejigging to support partial information (layer) without an prepopulated module or symbol table...

Doesn't need everyone to review it, just not sure who's got the most knowledge of VMCS or Intel VT-d...

### DIFF
--- a/volatility3/framework/plugins/vmscan.py
+++ b/volatility3/framework/plugins/vmscan.py
@@ -1,0 +1,217 @@
+# This file is Copyright 2023 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import enum
+import logging
+import os
+import struct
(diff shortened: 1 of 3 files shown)


eve-mem:
Interesting! Looks like it would close off this issue if merged https://github.com/volatilityfoundation/volatility3/issues/464

PR #1151, opened 2024-05-20
Linux: Replace uses of specific types with the more generic pointer

### LINKED ISSUE #1041: Add support for newer LLVM parsing for Linux - Replace "long unsigned int" with"unsigned long"
Scenario: Memory acquisition and analysis of Google Kubernetes (GKE) nodes as well as any Linux based kernel image that was compiled with newer version of clang+LLVM, see commit: https://github.com/llvm/llvm-project/commit/f6a561c4d6754b13165a49990e8365d819f64c86.
And this failure occurs for when the Linux kernel (both 5.10 and 5.15) is compiled.

The problem occurs when memory acquisition is done on GKE nodes using AVML and then dwarf2json to build a symbols file using vmlinux of the current build_id of the GKE node and active COS version. 
vmlinux is acquired via: curl -O https://storage.googleapis.com/cos-tools/build_id/vmlinux .e.g curl -O https://storage.googleapis.com/cos-tools/17412.101.24/vmlinux

The current version of Volatility3 uses "long unsigned int" at the following places:
https://github.com/volatilityfoundation/volatility3/blob/d56297c4b1cf0f9f4912f4f4158e232c700acb3f/volatility3/framework/automagic/linux.py#L162
https://github.com/volatilityfoundation/volatility3/blob/d56297c4b1cf0f9f4912f4f4158e232c700acb3f/volatility3/framework/plugins/linux/kmsg.py#L71C32-L71C32

Which causes Volatility3 to fail when the command:
python3 volatility3/vol.py -s PATH_TO/dwarf2json_profile.json -f PATH_TO/memory_dump.lime linux.ANY_PLUGIN is run
Banners and isInfo still works but any other Linux plugin that uses the linux framework and related automagic fails and claims that the symbols file or memory dump file is missing.

At the following lines:
https://github.com/volatilityfoundation/volatility3/blob/d56297c4b1cf0f9f4912f4f4158e232c700acb3f/volatility3/framework/automagic/linux.py#L162
https://github.com/volatilityfoundation/volatility3/blob/d56297c4b1cf0f9f4912f4f4158e232c700acb3f/volatility3/framework/plugins/linux/kmsg.py#L71C32-L71C32

Replace: "long unsigned int"
With: "unsigned long"

When testing this locally Volatility3 runs perfectly fine but for the solution above more testing would have to be done to fix this.

This solution was originally discovered by the Google GKE team and I have no knowledge of other solutions to fix this. 

**Additional information**
Add any other information or screenshots about the feature request here.
This issue was first raised here in the Volatility community Slack channel:
https://volatilitycommunity.slack.com/archives/CP9LZ5KD5/p1700060787864299

ikelos:
Attempts to fix #1041 by changing hard coded types into the more general pointer (since the type is usually for a pointer).  This is due to LLVM producing a different base type.  It doesn't necessarily resolve the situation permanently, but it should alleviate the main situations where it arises.

### DIFF
--- a/volatility3/framework/automagic/linux.py
+++ b/volatility3/framework/automagic/linux.py
@@ -159,7 +159,10 @@ class LinuxIntelStacker(interfaces.automagic.StackerLayerInterface):
 
             # This we get for free
             aslr_shift = (
-                init_task.files.cast("pointer")
+                int.from_bytes(
+                    init_task.files.cast("bytes", length=init_task.files.vol.size),
+                    byteorder=init_task.files.vol.data_format.byteorder,
+                )
                 - module.get_symbol("init_files").address
             )
             kaslr_shift = init_task_address - cls.virtual_to_physical_address(
--- a/volatility3/framework/automagic/linux.py
+++ b/volatility3/framework/automagic/linux.py
@@ -159,7 +159,7 @@ class LinuxIntelStacker(interfaces.automagic.StackerLayerInterface):
 
             # This we get for free
             aslr_shift = (
-                init_task.files.cast("long unsigned int")
+                init_task.files.cast("pointer")
                 - module.get_symbol("init_files").address
             )
             kaslr_shift = init_task_address - cls.virtual_to_physical_address(
--- a/volatility3/framework/plugins/linux/kmsg.py
+++ b/volatility3/framework/plugins/linux/kmsg.py
@@ -66,7 +66,7 @@ class ABCKmsg(ABC):
         self._config = config
         self.vmlinux = context.modules[self._config["kernel"]]
         self.layer_name = self.vmlinux.layer_name  # type: ignore
-        self.long_unsigned_int_size = self.vmlinux.get_type("long unsigned int").size
+        self.long_unsigned_int_size = self.vmlinux.get_type("pointer").size
 
     @classmethod
     def run_all(


ikelos:
Could someone on #1041 please test this to see if it resolves the issue at hand?  The other options available are to hard code our own version of that type, or force dwarf2json to inject it, neither of which seem all that clean...

ikelos:
Welp, that didn't work.  Looks like I'll need to investigate what went wrong...
Seems to be returning no results for linux plugins (but not throw an issue).  Should be easy to check the difference between the working version and the patched version.

ikelos:
Ok, turns out this doesn't work for a number of reasons.  Firstly, it's being done on the physical layer, which still comes out with the right pointer size, but the pointer is then limited to the size of the layer it's in, so it gets cropped.  Secondly, the symbol space has been loaded, but the higher layer hasn't been decided yet (because that depends on the init_task, which is what we're trying to make) so there's no way of using it as the native_layer for the pointer.

So now I'm doing a little bit of ugly gymnastics to cast the pointer to bytes (to get the original value) and then read the integer value from it the same way the pointer would, but without the address layer mask.  *Shouldn't* be necessary for the other use because just wants the size, which should match the system's pointer size (not sure if that holds true for 32-bit systems, I'm not sure what `long unsigned int` is defined as in 32-bit systems).  There may still be other places where the core depends on symbols that may not be present in the final symbol table.  Some will be in plugins, but I care less about them than the core.

PR #1164, opened 2024-06-13
Prevent backtrace in ADS scanning when contents cannot be recovered. …

atcuno:
…Fix missing parantheses as well as format_hints call on bad value


ikelos (approved):
Yep, looks good, thanks  5:)

PR #1165, opened 2024-06-14
The notes variable is not reset for each VAD, allowing bleed through …

atcuno:
…of a previously set note value to VADs enumerated afterwards. This patch resets the value for each VAD to avoid the bleed through bug.


ikelos (approved):
Yep, good catch!  5:)

PR #1167, opened 2024-06-19
Add processghosting plugin to detect process ghosting and related tec…

atcuno:
…hniques. Add process filter that gathers only active, non-smeared userland processes

### DIFF
--- a/volatility3/framework/plugins/windows/processghosting.py
+++ b/volatility3/framework/plugins/windows/processghosting.py
@@ -59,7 +59,9 @@ class ProcessGhosting(interfaces.plugins.PluginInterface):
                 file_object = 0
 
             if isinstance(delete_pending, int) and delete_pending not in [0, 1]:
-                vollog.debug(f"Invalid delete_pending value {delete_pending} found for {process_name} {proc.UniqueProcessId}")
+                vollog.debug(
+                    f"Invalid delete_pending value {delete_pending} found for {process_name} {proc.UniqueProcessId}"
+                )
 
             # delete_pending besides 0 or 1 = smear
             if file_object == 0 or delete_pending == 1:
--- a/volatility3/framework/plugins/windows/processghosting.py
+++ b/volatility3/framework/plugins/windows/processghosting.py
@@ -1,0 +1,99 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import logging
+
+from volatility3.framework import interfaces, exceptions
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/windows/processghosting.py
```diff
@@ -0,0 +1,100 @@
...
+                    file_object = proc.ImageFilePointer
+                    delete_pending = file_object.DeletePending
+                except exceptions.InvalidAddressException:
+                    file_object = 0
+
+            # ImageFilePointer equal to 0 means process ghosting or similar techniques were used
+            else:
+                file_object = 0
+
+            # delete_pending besides 0 or 1 = smear
```
ikelos, line 66: Could we get a DEBUG log message to that effect, or would it clutter the log too much (as in, how common is it?)

volatility3/framework/plugins/windows/processghosting.py
```diff
@@ -0,0 +1,100 @@
...
+            if file_object == 0 or delete_pending == 1:
+                path = renderers.UnreadableValue()
+                if file_object:
+                    try:
+                        path = file_object.FileName.String
+                    except exceptions.InvalidAddressException:
```
ikelos: This could probably be turned into a `with contextlib.suppress(exceptions.InvalidAddressException):` since we've already set path before hand?  At the very least, setting path could just be `pass` since it's the same value as it should have been initialized to...

https://docs.python.org/3.5/library/contextlib.html#contextlib.suppress

volatility3/framework/plugins/windows/pslist.py
```diff
@@ -136,6 +136,29 @@ def create_pid_filter(
                 filter_func = lambda x: x.UniqueProcessId not in filter_list
         return filter_func
 
+    @classmethod
+    def create_active_process_filter(
```
ikelos, line 140: Happy with this, adding new filters is the right way of doing things, this should be useful!  5:)

ikelos (commented):
Couple of minor style points, but no show stoppers.  Shout when you're happy with it (or if you're already happy with it) and I'll get it merged...  5:)

volatility3/framework/plugins/windows/processghosting.py
atcuno: Added

volatility3/framework/plugins/windows/processghosting.py
atcuno: fixed

atcuno:
@ikelos this one is ready

ikelos (approved):

ikelos:
Thanks for making those changes!  5:)

PR #1168, opened 2024-06-19
Add svclist and svcdiff plugins. Make svcscan more modular to support…

atcuno:
… inheritance and cleaner code

### DIFF
--- a/volatility3/framework/plugins/windows/svcdiff.py
+++ b/volatility3/framework/plugins/windows/svcdiff.py
@@ -13,7 +13,7 @@
 
 import logging
 
-from volatility3.framework import symbols
+from volatility3.framework import symbols, interfaces
 from volatility3.framework.configuration import requirements
 from volatility3.plugins.windows import svclist, svcscan
 from volatility3.framework.symbols.windows import versions
@@ -25,6 +25,10 @@ class SvcDiff(svcscan.SvcScan):
     """Compares services found through list walking versus scanning to find rootkits"""
 
     _required_framework_version = (2, 4, 0)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._enumeration_method = self.service_diff
 
     @classmethod
     def get_requirements(cls):
@@ -43,19 +47,24 @@ class SvcDiff(svcscan.SvcScan):
             ),
         ]
 
-    def _generator(self):
+    @classmethod
+    def service_diff(
+        cls,
+        context: interfaces.context.ContextInterface,
+        layer_name: str,
+        symbol_table: str,
+        service_table_name: str,
+        service_binary_dll_map,
+        filter_func,
+    ):
         """
         On Windows 10 version 15063+ 64bit Windows memory samples, walk the services list
         and scan for services then report differences
         """
-        kernel, service_table_name, service_binary_dll_map, filter_func = (
-            self.get_prereq_info()
-        )
-
         if not symbols.symbol_table_is_64bit(
-            self.context, kernel.symbol_table_name
+            context, symbol_table
         ) or not versions.is_win10_15063_or_later(
-            context=self.context, symbol_table=kernel.symbol_table_name
+            context=context, symbol_table=symbol_table
         ):
             vollog.warning(
                 "This plugin only supports Windows 10 version 15063+ 64bit Windows memory samples"
@@ -68,8 +77,9 @@ class SvcDiff(svcscan.SvcScan):
 
         # collect unique service names from scanning
         for service in svcscan.SvcScan.service_scan(
-            self.context,
-            kernel,
+            context,
+            layer_name,
+            symbol_table,
             service_table_name,
             service_binary_dll_map,
             filter_func,
@@ -79,8 +89,9 @@ class SvcDiff(svcscan.SvcScan):
 
         # collect services from listing walking
         for service in svclist.SvcList.service_list(
-            self.context,
-            kernel,
+            context,
+            layer_name,
+            symbol_table,
             service_table_name,
             service_binary_dll_map,
             filter_func,
@@ -89,4 +100,4 @@ class SvcDiff(svcscan.SvcScan):
 
         # report services found from scanning but not list walking
         for hidden_service in from_scan - from_list:
-            yield (0, records[hidden_service])
+            yield records[hidden_service]
--- a/volatility3/framework/plugins/windows/svcdiff.py
+++ b/volatility3/framework/plugins/windows/svcdiff.py
@@ -1,0 +1,74 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+# This module attempts to locate skeleton-key like function hooks.
+# It does this by locating the CSystems array through a variety of methods,
+# and then validating the entry for RC4 HMAC (0x17 / 23)
+#
+# For a thorough walkthrough on how the R&D was performed to develop this plugin,
+# please see our blogpost here:
+#
+# https://volatility-labs.blogspot.com/2021/10/memory-forensics-r-illustrated.html
+
+import logging
+
+from volatility3.framework import symbols
+from volatility3.framework.configuration import requirements
+from volatility3.plugins.windows import svclist, svcscan
+from volatility3.framework.symbols.windows import versions
+
+vollog = logging.getLogger(__name__)
+
+class SvcDiff(svclist.SvcList, svcscan.SvcScan):
+    """Compares services found through list walking versus scanning to find rootkits"""
+
+    _required_framework_version = (2, 4, 0)
+
+    @classmethod
+    def get_requirements(cls):
+        # Since we're calling the plugin, make sure we have the plugin's requirements
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Windows kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.VersionRequirement(
+                name="svclist", component=svclist.SvcList, version=(1, 0, 0)
+            ),
+            requirements.VersionRequirement(
+                name="svcscan", component=svcscan.SvcScan, version=(2, 0, 0)
+            ),
+        ]
+
+    def _generator(self):
+        """
+        Finds services by walking the services.exe list on supported Windows 10 versions
+        """
+        kernel = self.context.modules[self.config["kernel"]]
+
+        if not symbols.symbol_table_is_64bit(self.context, kernel.symbol_table_name) or \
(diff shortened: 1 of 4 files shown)


volatility3/framework/plugins/windows/svcdiff.py
```diff
@@ -0,0 +1,83 @@
...
+# and then validating the entry for RC4 HMAC (0x17 / 23)
+#
+# For a thorough walkthrough on how the R&D was performed to develop this plugin,
+# please see our blogpost here:
+#
+# https://volatility-labs.blogspot.com/2021/10/memory-forensics-r-illustrated.html
+
+import logging
+
+from volatility3.framework import symbols
```
ikelos: No typing import... doesn't bode well for some future comments...  5;P

volatility3/framework/plugins/windows/svclist.py
```diff
@@ -0,0 +1,102 @@
...
+from volatility3.framework import interfaces, exceptions, symbols
+from volatility3.framework.configuration import requirements
+from volatility3.framework.symbols.windows import versions
+from volatility3.plugins.windows import svcscan, pslist
+from volatility3.framework.layers import scanners
+
+vollog = logging.getLogger(__name__)
+
+
+class SvcList(svcscan.SvcScan):
```
ikelos, line 18: We don't generally do inheritance, we do `VersionRequirement`s and then call class methods.  It's a bit more boilerplate, but making this explicit means stuff can't change out from under us later on...

volatility3/framework/plugins/windows/svcdiff.py
```diff
@@ -0,0 +1,83 @@
...
+from volatility3.framework.configuration import requirements
+from volatility3.plugins.windows import svclist, svcscan
+from volatility3.framework.symbols.windows import versions
+
+vollog = logging.getLogger(__name__)
+
+
+class SvcDiff(svclist.SvcList, svcscan.SvcScan):
```
ikelos: I was looking for the run method, and then realize this doesn't inherit directly from `PluginInterface`, but uses multiple inheritance!  5:S  This makes individual plugins much more fragile to changes in the underlying code and requires updating multiple plugins depending on the changes.  Unless there's a very good reason, I'd strongly suggest exposing the functions that are needed in each inheritted plugin, and call them explicitly.  It'll also help other plugins reuse them properly and keep the dependency tree functioning properly.

volatility3/framework/plugins/windows/svclist.py
```diff
@@ -0,0 +1,102 @@
...
+        for vad in vad_root.traverse():
+            filename = vad.get_file_name()
+            if isinstance(filename, str) and filename.lower().endswith(
+                "\\services.exe"
+            ):
+                return [(vad.get_start(), vad.get_size())]
+
+        return None
+
+    def service_list(self, service_table_name, service_binary_dll_map, filter_func):
```
ikelos: This should be a `classmethod`, but it needs to have the standard context and the kernel module name added as parameters.

volatility3/framework/plugins/windows/svcscan.py
```diff
@@ -232,28 +232,40 @@ def _get_service_binary_map(
             for service_key in services
         }
 
-    def _generator(self):
-        kernel = self.context.modules[self.config["kernel"]]
+    def enumerate_vista_or_later_header(
```
ikelos, line 243: Again, this isn't following the standard `classmethod` and parameterized context as literally every other core plugin.  Please adapt it to fit in with the way all the other plugins play together...  5;)

volatility3/framework/plugins/windows/svcscan.py
```diff
@@ -232,28 +232,40 @@ def _get_service_binary_map(
...
+        for service_record in service_header.ServiceRecord.traverse():
+            service_info = service_binary_dll_map.get(
+                service_record.get_name(),
+                ServiceBinaryInfo(
+                    renderers.UnreadableValue(), renderers.UnreadableValue()
+                ),
+            )
+            yield self.get_record_tuple(service_record, service_info)
+
+    def service_scan(self, service_table_name, service_binary_dll_map, filter_func):
```
ikelos: Same again, unless it's `run` or `generator` or something that's private/will only ever be used by that specific plugin, it should be parameterized with at least `context` and exposed as a `classmethod` (to make sure self doesn't work accidentally).

ikelos (changes requested):
Happy with the distinguisher and all the code looks usable but the format it's currently in just ignores the coding style for plugins and reusability.  Shouldn't be too hard to rework, but this isn't going in core like this I'm afraid...

volatility3/framework/plugins/windows/svcdiff.py
atcuno: Added more

volatility3/framework/plugins/windows/svclist.py
atcuno: Made this to match the style of modscan/modules and the others and bumped versions and added requirements to match

volatility3/framework/plugins/windows/svcdiff.py
atcuno: Fixed to match modules/modscan

volatility3/framework/plugins/windows/svclist.py
atcuno: service_list and service_scan are classmethods now. svcdiff updated to use them that way + the version requirements matching

atcuno:
This one should be good to go @ikelos

volatility3/framework/plugins/windows/svcdiff.py
```diff
@@ -0,0 +1,92 @@
...
+            ),
+            requirements.VersionRequirement(
+                name="svclist", component=svclist.SvcList, version=(1, 0, 0)
+            ),
+            requirements.VersionRequirement(
+                name="svcscan", component=svcscan.SvcScan, version=(3, 0, 0)
+            ),
+        ]
+
+    def _generator(self):
```
ikelos: Did you mean to call this method `_enumeration_method`?  This can't be the generator, otherwise this and svcscan could get out of sync and it'd be a pig tracing back through everything to figure out where.  Overwriting the `_generator` is a no-go because it could then easily get out of sync with the inherited classes `run` and it somewhat makes spaghetti of the code.  `_enumeration_method` is the defined way that this class of plugins outputs its list of services, so that's how this plugin should work too.

volatility3/framework/plugins/windows/svclist.py
ikelos: Much, *much* better, thanks!  Still not super keen on the inheritance or expanding how much we do it, but here it's not as bad as tinkering with the generator because it's just override the `servicelist` method, similar to modules/modscan, which is designed to be overridden and have a well defined interface.

volatility3/framework/plugins/windows/svclist.py
```diff
@@ -0,0 +1,114 @@
...
+            ):
+                return [(vad.get_start(), vad.get_size())]
+
+        return None
+
+    @classmethod
+    def service_list(
+        cls,
+        context: interfaces.context.ContextInterface,
+        kernel,
```
ikelos: Much prefer passing context and simple types (name lookups) than context and complex types (kernel modules objects with references that'll just be used for lookup themselves anyway).  Any reason we pass in `kernel` rather than `kernel_name` and looking it up in `context.modules`?

volatility3/framework/plugins/windows/svcscan.py
```diff
@@ -306,37 +334,55 @@ def _generator(self):
...
-                                renderers.UnreadableValue(), renderers.UnreadableValue()
-                            ),
-                        )
-                        yield (
-                            0,
-                            self.get_record_tuple(service_record, service_info),
-                        )
+                        yield service_record
+
+    def get_prereq_info(self):
```
ikelos: Could this be parameterized and made a classmethod?  Doesn't look like it's doing anything that's private only (and if it were, it should start with `_`)...

volatility3/framework/plugins/windows/svcscan.py
```diff
@@ -232,30 +236,54 @@ def _get_service_binary_map(
...
+                ),
+            )
+            yield cls.get_record_tuple(service_record, service_info)
 
-        relative_tag_offset = self.context.symbol_space.get_type(
+    @classmethod
+    def service_scan(
+        cls,
+        context: interfaces.context.ContextInterface,
+        kernel,
```
ikelos: See above about passing a simple string and doing the lookup.

ikelos (commented):
Ok, very complex and nearly over the line, but overriding the `_generator` is a no go.  Luckily, it should be trivial to convert it to an `_enumeration_method` and so keep in line with the other two plugins here.

volatility3/framework/plugins/windows/svcscan.py
atcuno: fixed

volatility3/framework/plugins/windows/svcdiff.py
atcuno: Switched to enumeration method and bubbled the self methods to classmethods along the way

volatility3/framework/plugins/windows/svclist.py
atcuno: fixed

volatility3/framework/plugins/windows/svcscan.py
atcuno: fixed

atcuno:
Everything should be addressed @ikelos

volatility3/framework/plugins/windows/svcscan.py
```diff
@@ -140,18 +144,21 @@ def create_service_table(
...
             os.path.join("windows", "services"),
             symbol_filename,
-            class_types=services.class_types,
+            class_types=services_types.class_types,
             native_types=native_types,
         )
 
-    def _get_service_key(self, kernel) -> Optional[objects.StructType]:
+    @classmethod
+    def _get_service_key(
```
ikelos, line 152: If this one is now exposed, we should really remove the `_` from the front of it (so that we don't assume it's part of the private interface, but that changes need version bumps).

ikelos (commented):
Looking much better, one tiny housekeeping thing and then it can go in...  5:)

volatility3/framework/plugins/windows/svcscan.py
atcuno: I fixed this. That method should have been static and the create service table should have _. All those methods called by get_prereq_info should only be called by get_prereq_info so that the prerequisite data is built correctly. If i need to bump versions let me know, but svcscan already has a major version bump in this branch: https://github.com/volatilityfoundation/volatility3/pull/1168/files#diff-430aa6cb7e2040f3039b1de6da0e184992788d6f2cc1795af6f5f171fb774717R42

ikelos (approved):
Cool, thanks for clearing that up.  5:)

PR #1169, opened 2024-06-19
Add new resident data displaying plugin. Scan in correct layer. Condense code and remove duplication. Fix bugs

atcuno:
This MR performs a number of tasks related to the mftscan file and MFT-related functionatlity:

1) Fixes bugs, such as incorrect instantiation of absent values

2) Creates a unified and inheritable code flow for plugins that want to parse MFT records and their attributes.

3) Previously, ADSscan and MFTScan scanned in the kernel virtual address space, which is incorrect and caused in many samples over half the entries to be missed. These scans should occur in the physical address space.

4) Stops printing a disassembly of ADS data, which makes no sense anyway and made the output strange.

4) Adds a new plugin, ResidentData, in the file that hexdumps all of the resident data of a file (the first $DATA attribute). This inherits and uses a combined implementation of $DATA attribute access from the ADS plugin, which displays the second (and third and fourth ..., if present) $DATA attribute. Volatility 2 displayed resident data inline with the regular MFTscan type output, but this was confusing and would break grep results.

### DIFF
--- a/volatility3/framework/plugins/windows/mftscan.py
+++ b/volatility3/framework/plugins/windows/mftscan.py
@@ -84,8 +84,10 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
         mft_object = symbol_table + constants.BANG + "MFT_ENTRY"
         attribute_object = symbol_table + constants.BANG + "ATTRIBUTE"
 
+        record_map = {}
+
         # Scan the layer for Raw MFT records and parse the fields
-        for offset, _, _, _ in layer.scan(
+        for offset, _rule_name, _name, _value in layer.scan(
             context=context, scanner=yarascan.YaraScanner(rules=rules)
         ):
             with contextlib.suppress(exceptions.InvalidAddressException):
@@ -103,7 +105,7 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                 # There is no field that has a count of Attributes
                 # Keep Attempting to read attributes until we get an invalid attr_header.AttrType
                 while attr.Attr_Header.AttrType.is_valid_choice:
-                    for record in attr_callback(mft_record, attr, symbol_table):
+                    for record in attr_callback(record_map, mft_record, attr, symbol_table):
                         yield record
 
                     # If there's no advancement the loop will never end, so break it now
@@ -120,7 +122,7 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                     )
 
     @staticmethod
-    def parse_mft_records(mft_record, attr, symbol_table):
+    def parse_mft_records(record_map, mft_record, attr, symbol_table):
         # MFT Flags determine the file type or dir
         # If we don't have a valid enum, coerce to hex so we can keep the record
         try:
@@ -189,20 +191,26 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
         if attr.Attr_Header.NonResidentFlag:
             return
 
+        # we aren't looking ADS when we want the first data record
+        if return_first_record:
+            ads_name = renderers.NotApplicableValue()
+
+        # skip records without a name if we want ADS entries
+        elif attr.Attr_Header.NameLength == 0:
+            return
+
+        else:
+            # past the first $DATA record, attempt to get the ADS name
+            # NotAvailableValue = > 1st Data, but name was not parsable
+            ads_name = attr.get_resident_filename()
+            if not ads_name:
+                ads_name = renderers.NotAvailableValue()
+
         content = attr.get_resident_filecontent()
         if content:
             content = format_hints.HexBytes(content)
         else:
             content = renderers.NotAvailableValue()
-
-        # past the first $DATA record, attempt to get the ADS name
-        # NotApplicableValue = 1st Data
-        # NotAvailableValue = > 1st Data, but name was not parsable
-        ads_name = renderers.NotApplicableValue()
-        if not return_first_record and attr.Attr_Header.NameLength > 0:
-            ads_name = attr.get_resident_filename()
-            if not ads_name:
-                ads_name = renderers.NotAvailableValue()
 
         yield (
             format_hints.Hex(record_map[mft_record.RecordNumber][2]),
@@ -246,14 +254,15 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
             display_data = False
 
             # first DATA attribute of this record
-            if record_map[rec_num][1] == 0 and return_first_record:
+            if record_map[rec_num][1] == 0:
                 if return_first_record:
                     display_data = True
-                else:
-                    record_map[rec_num][1] = 1
+
+                record_map[rec_num][1] = 1
 
             # at the second DATA attribute of this record
-            elif not return_first_record:
+            elif record_map[rec_num][1] == 1 and not return_first_record:
+                print("at second record")
                 display_data = True
 
             if display_data:
@@ -265,6 +274,7 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
     @classmethod
     def parse_data_records(
         cls,
+        record_map: Dict[int, Tuple[str, int, int]],
         mft_record: interfaces.objects.ObjectInterface,
         attr: interfaces.objects.ObjectInterface,
         symbol_table,
@@ -273,7 +283,6 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
         """
         Callback for parsing data records through enumerate_mft_records
         """
-        record_map = {}
         for record in cls._do_parse_data_records(
             mft_record, attr, symbol_table, record_map, return_first_record
         ):
@@ -343,11 +352,12 @@ class ADS(interfaces.plugins.PluginInterface):
 
     @staticmethod
     def parse_ads_data_records(
+        record_map: Dict[int, Tuple[str, int, int]],
         mft_record: interfaces.objects.ObjectInterface,
         attr: interfaces.objects.ObjectInterface,
         symbol_table,
     ):
-        return MFTScan.parse_data_records(mft_record, attr, symbol_table, False)
+        return MFTScan.parse_data_records(record_map, mft_record, attr, symbol_table, False)
 
     def _generator(self):
         for (
@@ -382,7 +392,7 @@ class ADS(interfaces.plugins.PluginInterface):
 
 
 class ResidentData(interfaces.plugins.PluginInterface):
-    """Scans for Alternate Data Stream"""
+    """Scans for MFT Records with Resident Data"""
 
     _required_framework_version = (2, 7, 0)
 
@@ -406,11 +416,12 @@ class ResidentData(interfaces.plugins.PluginInterface):
 
     @staticmethod
     def parse_first_data_records(
+        record_map: Dict[int, Tuple[str, int, int]],
         mft_record: interfaces.objects.ObjectInterface,
         attr: interfaces.objects.ObjectInterface,
         symbol_table,
     ):
-        return MFTScan.parse_data_records(mft_record, attr, symbol_table, True)
+        return MFTScan.parse_data_records(record_map, mft_record, attr, symbol_table, True)
 
     def _generator(self):
         for (
--- a/volatility3/framework/plugins/windows/mftscan.py
+++ b/volatility3/framework/plugins/windows/mftscan.py
@@ -105,7 +105,9 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                 # There is no field that has a count of Attributes
                 # Keep Attempting to read attributes until we get an invalid attr_header.AttrType
                 while attr.Attr_Header.AttrType.is_valid_choice:
-                    for record in attr_callback(record_map, mft_record, attr, symbol_table):
+                    for record in attr_callback(
+                        record_map, mft_record, attr, symbol_table
+                    ):
                         yield record
 
                     # If there's no advancement the loop will never end, so break it now
@@ -357,7 +359,9 @@ class ADS(interfaces.plugins.PluginInterface):
         attr: interfaces.objects.ObjectInterface,
         symbol_table,
     ):
-        return MFTScan.parse_data_records(record_map, mft_record, attr, symbol_table, False)
+        return MFTScan.parse_data_records(
+            record_map, mft_record, attr, symbol_table, False
+        )
 
     def _generator(self):
         for (
@@ -421,7 +425,9 @@ class ResidentData(interfaces.plugins.PluginInterface):
         attr: interfaces.objects.ObjectInterface,
         symbol_table,
     ):
-        return MFTScan.parse_data_records(record_map, mft_record, attr, symbol_table, True)
+        return MFTScan.parse_data_records(
+            record_map, mft_record, attr, symbol_table, True
+        )
 
     def _generator(self):
         for (
--- a/volatility3/framework/plugins/windows/mftscan.py
+++ b/volatility3/framework/plugins/windows/mftscan.py
@@ -5,6 +5,8 @@ import contextlib
 import datetime
 import logging
 
+from typing import Generator, Iterable
+
 from volatility3.framework import constants, exceptions, interfaces, renderers
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import conversion, format_hints
@@ -23,6 +25,10 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._record_map = {}
+        self.mft_object = None
+        self.attribute_object = None
+        self.si_object = None
+        self.fn_object = None
 
     @classmethod
     def get_requirements(cls):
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -33,8 +37,10 @@ def get_requirements(cls):
             ),
         ]
 
-    def _generator(self):
-        layer = self.context.layers[self.config["primary"]]
+    def enumerate_mft_records(self, attr_callback):
+        phys_layer = self.context.layers[self.config["primary"]].config["memory_layer"]
```
ikelos: Would be nicer to handle `KeyErrors` here (or use `get` so tht failures result in phys_layer being `None` which can be easily checked)..  The requirements should ensure everything works but, give it's two dictionary lookups that could fail, it just feels a little fragile.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -47,100 +53,110 @@ def _generator(self):
...
+                "ATTRIBUTE": mft.MFTAttribute,
+            },
         )
 
         # get each of the individual Field Sets
-        mft_object = symbol_table + constants.BANG + "MFT_ENTRY"
-        attribute_object = symbol_table + constants.BANG + "ATTRIBUTE"
-        si_object = symbol_table + constants.BANG + "STANDARD_INFORMATION_ENTRY"
-        fn_object = symbol_table + constants.BANG + "FILE_NAME_ENTRY"
+        self.mft_object = symbol_table + constants.BANG + "MFT_ENTRY"
```
ikelos: As part of the coding style, these should be defined as part of `__init__` (initialized to `None`) so that accidental accesses don't throw up an `AttributeError`.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -174,126 +190,120 @@ def run(self):
...
-                name="yarascanner", component=yarascan.YaraScanner, version=(2, 0, 0)
-            ),
-        ]
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # which DATA attribute should be displayed
+        self._display_first_data = False
+
+    def _parse_data_record(self, mft_record, attr):
```
ikelos: Could we add typing information to this just to make it easier to spot things going wrong please?  This will return a `Generator` type so the return at the start wouldn't require `Optional` in the type.  Makes it easier to figure out if code using it will throw errors...

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -174,126 +190,120 @@ def run(self):
...
+                format_hints.Hex(self._record_map[mft_record.RecordNumber][2]),
+                mft_record.get_signature(),
+                mft_record.RecordNumber,
+                attr.Attr_Header.AttrType.lookup(),
+                self._record_map[mft_record.RecordNumber][0],
+                ads_name,
+                content,
+            )
+
+    def parse_data_records(self, mft_record, attr):
```
ikelos: This would also be a generator, even though you can't tell until the very end of the function.

ikelos (commented):
Yep, generally looks good.  A couple of minor house keeping things just to keep the coding style in check but otherwise looks good thanks!

volatility3/framework/plugins/windows/mftscan.py
atcuno: @ikelos I can catch keyerror around this but I am confused:

 1) how it would fail?
 2) what would the backup be to get access to the physical layer? Returning None would mean we can't recover records from the physical layer, which would be a bug we need to fix.

Is there example code of how to get the physical layer if the key error hits?

volatility3/framework/plugins/windows/mftscan.py
atcuno: Good catch. Fixed

volatility3/framework/plugins/windows/mftscan.py
atcuno: fixed

volatility3/framework/plugins/windows/mftscan.py
atcuno: Fixed

volatility3/framework/plugins/windows/mftscan.py
ikelos: Well, it's mostly if `primary` is found but a) isn't a layer or b) isn't an intel layer (for some reason, possibly in the future) and therefore doesn't contain a `memory_layer`.  That could happen if the config is loaded from a file and could send us chasing down the wrong path.  This code chains several assumed config items together without any error handling which is a bit brittle.  It'll throw an exception but that's not really a good result for the user.  Chances are we'll never run into them, but handling errors will help us figure out things going wrong that we didn't expect down the line.  The backup would be a nice error message informing the user that something went wrong (or better, how it went wrong).

It's not a huge deal and there's lots of other places where we don't do proper error checking for all conditions, but at a minimum I'd prefer it to be split into at least two lines, so that we can figure out which lookup is failing...

atcuno: I split them up and added wording to request the user to create an issue on the tracker if those lines are triggered since it seems like it would be some major internal bug, so hopefully we can catch and fix it.

ikelos (changes requested):
Mmmm, this is kinda skirting the rules by making the generator and run dummy functions and then cramming all the differences into the `parse_mft_records` method.  5:S  The generator really shouldn't be returning different sized output depending on input flags (or if it should that should definitely all stay within one plugin).

I think the best way for this set of plugins to work would be to have the `enumerate_mft_records` central and used from one place (ideally parameterized and made a classmethod, I'm not sure why it stores local state mid run?) and then the run/generator and the appropriate columns should be defined in the plugins that use them (so `parse_mft_records` gets distributed across the plugins.  That even negate the need for inheritance, but I'll let you evaluate that...

atcuno:
I am going to rewrite this a bit based on discussion from this weekend and what you said in this ticket, but I did want to point out early that each plugin does have its own `run` already based on the columns it produces.

ikelos:
That's fine.  [This](https://github.com/volatilityfoundation/volatility3/pull/1169/files#diff-9d1ebde0bcd73e1b19e022ffacd1274a6b97f144fdcd4de1ed310e42e44b7ea1R337) only has a run but calls a generator from a child.  The problem is pulling some bits from one plugin but redefining other bits itself, which makes an easily missed dependency between the two.  If you're going to make it so that changes in a parent can't easily break the child without the user getting a nice message telling them about the problem, then that would be ideal...

atcuno:
ok @ikelos I reworked all of this to match where we have the staticmethods and classmethods and each plugin has its own run and generator.. removing the inheritance lets the flag about which $DATA to display be part of the parameters and also the callbacks have consistent columns in their return data sets.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -43,104 +69,222 @@ def _generator(self):
...
         # get each of the individual Field Sets
         mft_object = symbol_table + constants.BANG + "MFT_ENTRY"
         attribute_object = symbol_table + constants.BANG + "ATTRIBUTE"
-        si_object = symbol_table + constants.BANG + "STANDARD_INFORMATION_ENTRY"
-        fn_object = symbol_table + constants.BANG + "FILE_NAME_ENTRY"
 
         # Scan the layer for Raw MFT records and parse the fields
-        for offset, _rule_name, _name, _value in layer.scan(
-            context=self.context, scanner=yarascan.YaraScanner(rules=rules)
+        for offset, _, _, _ in layer.scan(
```
ikelos: Not sure why you're removing the descriptive names here.  If anyone comes looking at the code it'll tell them what's returned for the sake of a few characters?  It's ok, but I don't understand why it was done when they were already properly annotated?

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -33,8 +37,30 @@ def get_requirements(cls):
...
+    def enumerate_mft_records(
+        context: interfaces.context.ContextInterface,
+        config: interfaces.configuration.HierarchicalDict,
+        config_path: str,
+        attr_callback,
+    ) -> interfaces.objects.ObjectInterface:
+        try:
+            primary = context.layers[config["primary"]]
+        except KeyError:
+            vollog.error(
```
ikelos, line 58: Yeah, I admit do look a bit excessive, but at least it's clear what's going on now and having it split onto at least two lines will make it clearer if something goes wrong.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -43,104 +69,222 @@ def _generator(self):
...
                     attr_base_offset += attr.Attr_Header.Length
-                    attr = self.context.object(
+                    # Get the next attribute
+                    attr = context.object(
                         attribute_object,
                         offset=offset + attr_base_offset,
                         layer_name=layer.name,
                     )
 
+    @staticmethod
```
ikelos, line 131: Happy with these as static methods, we just need to remember they count as part of the external API if they ever change in the future, they should cause the version number to bump too.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -43,104 +69,222 @@
...
+            record_map[rec_num][2] = attr.Attr_Data.vol.offset
+
+            display_data = False
+
+            # first DATA attribute of this record
+            if record_map[rec_num][1] == 0 and return_first_record:
+                if return_first_record:
+                    display_data = True
+                else:
+                    record_map[rec_num][1] = 1
```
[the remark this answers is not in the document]
ikelos: The scanner makes a good point.  We verify that return_first_record is set in the if statement before this one...  5;P

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -192,108 +341,90 @@ def get_requirements(cls):
...
+            attr_type,
+            file_name,
+            ads_name,
+            content,
+        ) in MFTScan.enumerate_mft_records(
+            self.context, self.config, self.config_path, self.parse_ads_data_records
+        ):
+            yield (
+                0,
+                (offset, rec_type, rec_num, attr_type, file_name, ads_name, content),
```
ikelos, line 376: This looks the same as the `for` input, wouldn't this be easier to write as:

```
for item in MFTScan.enumerate_mft_records(<args>):
  yield (0, item)
```

or have I missed something?

ikelos: Oh I see, this matches later calls which don't use all the return values, ok, that's fine.

ikelos (commented):
Ok, really happy with this.  A couple style points that I just want to check before merging (and one codeQL issue that needs resolving), but looks like it's in a good state.  5:)  Thanks for sorting all the issues!

volatility3/framework/plugins/windows/mftscan.py
atcuno: I actually thought it was a coding bug to have the name like that if it isn't used. Pylint definitely yelled at me about it in the past.

ikelos: Possibly, but I don't think black complains and I prefer it as a way of ensuring people know what the parameters are in case they want to start using them, rather than trying to rely on memory, getting it wrong and spending cycles just because we wanted to drop a few characters in a commit...  5;)

volatility3/framework/plugins/windows/mftscan.py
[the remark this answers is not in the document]
ikelos: This should be fixed before it goes in?  Line 249 checks `return_first_record` and then line 260 checks it again, which will always be true, so the else clause will never get run.  I think either the check on line 249 needs to go, or the else clause on 252/253 needs to go.

ikelos:
Still waiting on a couple of minor things then it should be fine to go in...

volatility3/framework/plugins/windows/mftscan.py
atcuno: changed it back

ikelos:
If you get this past the black formatting, it can be merged now, thanks for the changes...

atcuno:
> If you get this past the black formatting, it can be merged now, thanks for the changes...

fixed now

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -33,8 +37,30 @@ def get_requirements(cls):
...
-        layer = self.context.layers[self.config["primary"]]
+    @staticmethod
+    def enumerate_mft_records(
+        context: interfaces.context.ContextInterface,
+        config: interfaces.configuration.HierarchicalDict,
+        config_path: str,
+        attr_callback,
+    ) -> interfaces.objects.ObjectInterface:
+        try:
+            primary = context.layers[config["primary"]]
```
ikelos: I'd pass in `primary_layer_name`  as a string.  If this gets called from a different plugin the config may not match exactly and then they'll have to construct a fake one just to get to that value.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -43,104 +69,233 @@ def _generator(self):
...
-                            mft_flag,
-                            permissions,
-                            attr.Attr_Header.AttrType.lookup(),
-                            conversion.wintime_to_datetime(attr_data.CreationTime),
-                            conversion.wintime_to_datetime(attr_data.ModifiedTime),
-                            conversion.wintime_to_datetime(attr_data.UpdatedTime),
-                            conversion.wintime_to_datetime(attr_data.AccessedTime),
-                            file_name,
-                        )
+                    for record in attr_callback(
```
ikelos: Looks like you could *probably* use `yield from` here, but honestly this is fine.  More just a point of interest...
https://docs.python.org/3/whatsnew/3.3.html#pep-380

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -192,108 +352,96 @@ def get_requirements(cls):
...
-        layer = self.context.layers[self.config["primary"]]
+        for (
+            offset,
+            rec_type,
+            rec_num,
+            attr_type,
+            file_name,
+            ads_name,
+            content,
+        ) in MFTScan.enumerate_mft_records(
```
ikelos, line 368: This also looks like it could be done as a `yield from`?

ikelos (commented):
Yep, looking pretty good, much happier about the way this works now.  Just one little last gotcha is passing the whole config in rather than just the data you need.  The whole config will be tricky for other plugins to match up.  Otherwise it looks good to go in!  5:D

volatility3/framework/plugins/windows/mftscan.py
atcuno: done

atcuno:
Feedback addressed @ikelos . I added yield from anywhere feasible plus the primary layer name passing.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -43,104 +69,213 @@ def _generator(self):
...
-                    attr = self.context.object(
+                    # Get the next attribute
+                    attr = context.object(
                         attribute_object,
                         offset=offset + attr_base_offset,
                         layer_name=layer.name,
                     )
 
+    @staticmethod
+    def parse_mft_records(record_map, mft_record, attr, symbol_table):
```
ikelos: This could do with some typing information, but not a show stopper...

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -33,8 +37,30 @@ def get_requirements(cls):
...
         ]
 
-    def _generator(self):
-        layer = self.context.layers[self.config["primary"]]
+    @staticmethod
+    def enumerate_mft_records(
+        context: interfaces.context.ContextInterface,
+        config_path: str,
+        primary_layer_name: str,
+        attr_callback,
```
ikelos: Could we type this as a `Callable` at least, if not providing the full function signature please?

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -43,104 +69,213 @@ def _generator(self):
...
+
+        # skip records without a name if we want ADS entries
+        elif attr.Attr_Header.NameLength == 0:
+            return
+
+        else:
+            # past the first $DATA record, attempt to get the ADS name
+            # NotAvailableValue = > 1st Data, but name was not parsable
+            ads_name = attr.get_resident_filename()
+            if not ads_name:
```
ikelos: Feels like this could be `... = attr.get_resident_filename() or renderers...`  but I'm fine if you want to keep it this way.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -43,104 +69,213 @@ def _generator(self):
...
+            content,
+        )
+
+    @classmethod
+    def parse_data_records(
+        cls,
+        record_map: Dict[int, Tuple[str, int, int]],
+        mft_record: interfaces.objects.ObjectInterface,
+        attr: interfaces.objects.ObjectInterface,
+        symbol_table,
```
ikelos: Just for clarity, could we please rename this to `symbol_table_name` indicating it's not the actual symbol table object, and give it type information too?

ikelos (commented):
Looking good now, thanks for all the changes.  Just a variable name change for clarity and then it should be good to go...  5:)

volatility3/framework/plugins/windows/mftscan.py
atcuno: done

volatility3/framework/plugins/windows/mftscan.py
atcuno: done

volatility3/framework/plugins/windows/mftscan.py
atcuno: done

volatility3/framework/plugins/windows/mftscan.py
atcuno: done

atcuno:
this is ready for merge @ikelos

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -33,8 +37,38 @@ def get_requirements(cls):
...
+        try:
+            primary = context.layers[primary_layer_name]
+        except KeyError:
+            vollog.error(
+                "Unable to obtain primary layer for scanning. Please file a bug on GitHub about this issue."
+            )
+            return
+
+        try:
+            phys_layer = primary.config["memory_layer"]
```
ikelos, line 64: This is fine, but as a point for the future, this may not work on arm layers, etc...

ikelos (approved):
Looks good, thanks for fixing everything up!  5:)

PR #1170, opened 2024-06-20
Add new plugin to detect hollowed processes using a variety of techni…

atcuno:
…ques and allowing for easy additions of future detection techniques

### DIFF
--- a/volatility3/framework/plugins/windows/hollowprocesses.py
+++ b/volatility3/framework/plugins/windows/hollowprocesses.py
@@ -1,4 +1,4 @@
-# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
 import logging
@@ -17,14 +17,14 @@ VadInfo = NamedTuple(
     [
         ("protection", str),
         ("path", str),
-   ],
+    ],
 )
 
 DLLInfo = NamedTuple(
     "DLLInfo",
     [
         ("path", str),
-   ],
+    ],
 )
 
 class HollowProcesses(interfaces.plugins.PluginInterface):
@@ -111,15 +111,21 @@ class HollowProcesses(interfaces.plugins.PluginInterface):
     def _check_load_address(self, proc, _, __):
         image_base = self._get_image_base(proc)
         if image_base is not None and image_base != proc.SectionBaseAddress:
-            yield "The ImageBaseAddress reported from the PEB ({:#x}) does not match the process SectionBaseAddress ({:#x})".format(image_base, proc.SectionBaseAddress)
+            yield "The ImageBaseAddress reported from the PEB ({:#x}) does not match the process SectionBaseAddress ({:#x})".format(
+                image_base, proc.SectionBaseAddress
+            )
 
     def _check_exe_protection(self, proc, vads, __):
         base = proc.SectionBaseAddress
 
         if base not in vads:
-            yield "There is no VAD starting at the base address of the process executable ({:#x})".format(base)
+            yield "There is no VAD starting at the base address of the process executable ({:#x})".format(
+                base
+            )
         elif vads[base].protection != "PAGE_EXECUTE_WRITECOPY":
-            yield "Unexpected protection ({}) for VAD hosting the process executable ({:#x}) with path {}".format(vads[base].protection, base, vads[base].path)
+            yield "Unexpected protection ({}) for VAD hosting the process executable ({:#x}) with path {}".format(
+                vads[base].protection, base, vads[base].path
+            )
 
     def _check_dlls_protection(self, _, vads, dlls):
         for dll_base in dlls:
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/hollowprocesses.py
```diff
@@ -0,0 +1,188 @@
...
+
+from volatility3.framework import interfaces, exceptions, constants
+from volatility3.framework import renderers
+from volatility3.framework.configuration import requirements
+from volatility3.framework.objects import utility
+from volatility3.plugins.windows import pslist, vadinfo
+
+vollog = logging.getLogger(__name__)
+
+VadInfo = NamedTuple(
```
ikelos: I think `namedtuple` is canonically defined in `collections` not in `typing`?  I think that's just the type of it...
https://docs.python.org/3/library/collections.html#collections.namedtuple

volatility3/framework/plugins/windows/hollowprocesses.py
```diff
@@ -0,0 +1,188 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import logging
+from typing import NamedTuple
```
ikelos: There should be a bunch of other things here, if typing only has one import then probably we're missing typing information in the plugin.  5:S  Best to annotate as many (simple) methods as possible.  It'll make error checking a lot easier in the future...

volatility3/framework/plugins/windows/hollowprocesses.py
```diff
@@ -0,0 +1,188 @@
...
+            ),
+            requirements.VersionRequirement(
+                name="pslist", component=pslist.PsList, version=(2, 0, 0)
+            ),
+            requirements.VersionRequirement(
+                name="vadinfo", component=vadinfo.VadInfo, version=(2, 0, 0)
+            ),
+        ]
+
+    def _get_vads_map(self, proc):
```
ikelos: Is this significantly different than `get_vad_map` staticmethod in `windows.vadyarascan`?  If so, it might be worth renaming this one to something a bit more specific (the last one is public and part of the API so would need version bumps to change).

volatility3/framework/plugins/windows/hollowprocesses.py
```diff
@@ -0,0 +1,188 @@
...
+        vads = {}
+
+        kernel = self.context.modules[self.config["kernel"]]
```
ikelos, line 75: Probably fine not to have error checking, but just to be aware, there is a world where this may throw an ugly `KeyErrror` for one reason or another.

volatility3/framework/plugins/windows/hollowprocesses.py
```diff
@@ -0,0 +1,188 @@
...
+                    self.context, kernel.layer_name, kernel.symbol_table_name
+                ),
+                vadinfo.winnt_protections,
+            )
+
+            fn = vad.get_file_name()
+            if not fn or not isinstance(fn, str):
+                fn = "<Non-File Backed Region>"
+
+            vads[vad.get_start()] = VadInfo(protection_string, fn)
```
ikelos: VadInfo is a bit confusing because it's the name of the plugin, it's the name of a structure in the plugin (different case) and now it's also a NamedTuple type.  Given it's only really called once, changing the `NamedTuple`'s type seems like the easiest way to avoid confusion.

volatility3/framework/plugins/windows/hollowprocesses.py
```diff
@@ -0,0 +1,188 @@
...
+        if base not in vads:
+            yield "There is no VAD starting at the base address of the process executable ({:#x})".format(
+                base
+            )
+        elif vads[base].protection != "PAGE_EXECUTE_WRITECOPY":
+            yield "Unexpected protection ({}) for VAD hosting the process executable ({:#x}) with path {}".format(
+                vads[base].protection, base, vads[base].path
+            )
+
+    def _check_dlls_protection(self, _, vads, dlls):
```
ikelos: Again, typing would be really useful here.  `vads` is a dict of named tuples that's indexed by `dll_base` which is an offset, but dlls is just a list of NamedTuples that contain only a single attribute?  Very confusing for people new to the codebase to trace through what's going on...

These methods could really do with docstrings/typing and other clarifying information.  Would go a long way to helping understandability...

ikelos (changes requested):
This one definitely needs some explanation/docstrings/typing information before it goes live.  I think it all hangs together and is ok, but it's a mess for anyone else to wade through and figure out what's going on...  5:S

volatility3/framework/plugins/windows/hollowprocesses.py
atcuno: I can change it if needed, but I got the flow for it from here: https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/interfaces/renderers.py#L21-L29

volatility3/framework/plugins/windows/hollowprocesses.py
atcuno: Changed this and the DLLifno name.

volatility3/framework/plugins/windows/hollowprocesses.py
atcuno: I added many more comments and typing. I didn't realize it was so bare before, sorry.

volatility3/framework/plugins/windows/hollowprocesses.py
atcuno: I renamed it since it gets a different set of information.

volatility3/framework/plugins/windows/hollowprocesses.py
atcuno: Added many comments and type information sets.

volatility3/framework/plugins/windows/hollowprocesses.py
ikelos: Fair enough, I don't think that's right either and I'd prefer we went with collections, but it is used elsewhere so I won't quibble.  Maybe one day I'll have a run back over the codebase and change them both...  5:)

volatility3/framework/plugins/windows/hollowprocesses.py
ikelos: Thanks, it's better, but for instance, `vads` and `dlls` here still seem to be missing typing information?

ikelos (commented):
Much improved, just a little extra typing and it's fine to go in...  5:)

ikelos (approved):

ikelos (commented):
Sorry, trying to get request changes set back to just comment.

volatility3/framework/plugins/windows/hollowprocesses.py
atcuno: Ok I added the missing typing.

ikelos (approved):
Looks good, thanks!

PR #1173, opened 2024-06-20
Windows: Bug fixes and additions to modules and modscan

brandon-barnacle:
This pull request adds several changes to the modules and modscan plugins that fix bugs, unify options and operations, and allow for filtering by name and base address in both plugins. Previously, modscan and modules had a substantial amount of partially duplicated code, which led to bugs, such as:
	Only one allowed filtering by --name
	Only one cycled through the session layers to find the one hosting the kernel module
With the new inheritance and combined implementation, the correct and complete implementation applies to both.

This PR also addresses the feedback in
https://github.com/volatilityfoundation/volatility3/pull/1099. In this PR, --name and –base apply to the processing of modules and determine both which plugins are displayed in output as well as which are processed for extraction. This fixes the issue of using --filters, which only filters the final command line output, but leads to every module being extracted.

### DIFF
--- a/volatility3/framework/plugins/windows/modscan.py
+++ b/volatility3/framework/plugins/windows/modscan.py
@@ -2,13 +2,10 @@
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
 import logging
-from typing import Iterable, List, Generator
+from typing import Iterable
 
-from volatility3.framework import renderers, interfaces, exceptions, constants
+from volatility3.framework import interfaces
 from volatility3.framework.configuration import requirements
-from volatility3.framework.renderers import format_hints
-from volatility3.framework.symbols import intermed
-from volatility3.framework.symbols.windows.extensions import pe
 from volatility3.plugins.windows import poolscanner, dlllist, pslist, modules
 
 vollog = logging.getLogger(__name__)
@@ -36,7 +33,7 @@ class ModScan(modules.Modules):
                 name="poolscanner", component=poolscanner.PoolScanner, version=(1, 0, 0)
             ),
             requirements.VersionRequirement(
-                name="modules", component=modules.Modules, version=(1, 1, 0)
+                name="modules", component=modules.Modules, version=(2, 0, 0)
             ),
             requirements.VersionRequirement(
                 name="pslist", component=pslist.PsList, version=(2, 0, 0)
--- a/volatility3/framework/plugins/windows/modscan.py
+++ b/volatility3/framework/plugins/windows/modscan.py
@@ -9,16 +9,20 @@ from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints
 from volatility3.framework.symbols import intermed
 from volatility3.framework.symbols.windows.extensions import pe
-from volatility3.plugins.windows import poolscanner, dlllist, pslist
+from volatility3.plugins.windows import poolscanner, dlllist, pslist, modules
 
 vollog = logging.getLogger(__name__)
 
 
-class ModScan(interfaces.plugins.PluginInterface):
+class ModScan(modules.Modules):
     """Scans for modules present in a particular windows memory image."""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 0, 0)
+    _version = (2, 0, 0)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._enumeration_method = self.scan_modules
 
     @classmethod
     def get_requirements(cls):
@@ -32,6 +36,9 @@ class ModScan(interfaces.plugins.PluginInterface):
                 name="poolscanner", component=poolscanner.PoolScanner, version=(1, 0, 0)
             ),
             requirements.VersionRequirement(
+                name="modules", component=modules.Modules, version=(1, 1, 0)
+            ),
+            requirements.VersionRequirement(
                 name="pslist", component=pslist.PsList, version=(2, 0, 0)
             ),
             requirements.VersionRequirement(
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/windows/modscan.py
```diff
@@ -9,16 +9,20 @@
...
 from volatility3.framework.symbols import intermed
 from volatility3.framework.symbols.windows.extensions import pe
-from volatility3.plugins.windows import poolscanner, dlllist, pslist
+from volatility3.plugins.windows import poolscanner, dlllist, pslist, modules
 
 vollog = logging.getLogger(__name__)
 
 
-class ModScan(interfaces.plugins.PluginInterface):
+class ModScan(modules.Modules):
```
ikelos, line 14: We typically don't use inheritance, but add a VersionRequirement (as you've done) and then make calls to classmethod and exposed functions?  It makes it easier for plugins to know what they can and can't change and keeps things clearer.

volatility3/framework/plugins/windows/modules.py
```diff
@@ -50,49 +59,74 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+                layer_name=session_layer_name,
+            )
+            file_output = "Error outputting file"
+            if file_handle:
+                file_output = file_handle.preferred_filename
+
+        return file_output
+
+    def process_module(self, session_layers, pe_table_name, mod):
+        if self.config["base"] and self.config["base"] != mod.DllBase:
```
ikelos: We try to parameterize things so that `self.config` is only really accessed in the main `run` or `_generator`.  This allows the other methods to be exposed for use (so `process_module` would have a `base` parameter, and then a `classmethod` decorator, so that other plugins that wanted what this plugin does can just depend on its version and make calls to it.  5:)

volatility3/framework/plugins/windows/modules.py
```diff
@@ -22,6 +22,10 @@ class Modules(interfaces.plugins.PluginInterface):
     _required_framework_version = (2, 0, 0)
     _version = (1, 1, 0)
```
ikelos: The version number should be bumped because you're technically adding need methods to the plugin interface.  So this should go to at least `(1, 2, 0)`

volatility3/framework/plugins/windows/modules.py
```diff
@@ -50,49 +59,74 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+        return file_output
+
+    def process_module(self, session_layers, pe_table_name, mod):
+        if self.config["base"] and self.config["base"] != mod.DllBase:
+            return None
+
+        try:
+            BaseDllName = mod.BaseDllName.get_string()
+        except exceptions.InvalidAddressException:
+            BaseDllName = ""
```
ikelos: Should this be a `BaseAbsentValue`?  It helps distinguish between "something's really gone wrong" and "it's just an empty string in the memory".

volatility3/framework/plugins/windows/modules.py
```diff
@@ -50,49 +59,74 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+
+        if self.config["name"] and self.config["name"] not in BaseDllName:
```
ikelos: Again, this should be parameterized.

volatility3/framework/plugins/windows/modules.py
```diff
@@ -50,49 +59,74 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+            return None
+
+        try:
+            FullDllName = mod.FullDllName.get_string()
+        except exceptions.InvalidAddressException:
+            FullDllName = ""
+
+        file_output = "Disabled"
+        if self.config["dump"]:
+            file_output = self.dump_module(session_layers, pe_table_name, mod)
```
ikelos: If this is just using `pe_table_name` and `session_layers` to pass through to the `dump_module` call, then this should do what it's supposed to do, and the person calling it should call `dump_module` directly instead themselves.

volatility3/framework/plugins/windows/modules.py
```diff
@@ -50,49 +59,74 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
-                (
-                    format_hints.Hex(mod.vol.offset),
-                    format_hints.Hex(mod.DllBase),
-                    format_hints.Hex(mod.SizeOfImage),
-                    BaseDllName,
-                    FullDllName,
-                    file_output,
-                ),
-            )
+            record = self.process_module(session_layers, pe_table_name, mod)
```
ikelos: So this could be expanded to call `process_module` and then check the config here as to whether to dump or not.

volatility3/framework/plugins/windows/modules.py
```diff
@@ -50,49 +59,74 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+
     def _generator(self):
         kernel = self.context.modules[self.config["kernel"]]
+
         pe_table_name = intermed.IntermediateSymbolTable.create(
             self.context, self.config_path, "windows", "pe", class_types=pe.class_types
         )
 
-        for mod in self.list_modules(
+        session_layers = list(
```
ikelos: This doesn't strictly look like it's used apart from for dumping, it should probably be protected by a check for the `config['dump']` flag, to avoid unnecessary processing.  Looks like we did it unnecessarily in the past, so that would provide an efficiency boost too!  5:)

ikelos (commented):
Thanks this is much cleaner and a nice way of dealing with the situation.  Happy merging the various bits and pieces, but not through inheritance please.  We generally explicitly expose bits of plugins that are supposed to be used by other plugins, so that each plugin can be changed and the version bumped without having to worry too much about what else might get affected accidentally.  Shouldn't be much work to change things to work that way.  There's a few function signatures that could be updated, and the dumping support pulled out of the `process_module` call, but otherwise looking good, thanks!

atcuno:
Hey @ikelos , I worked with Brandon and I believe we addressed everything, including the class methods, "" strings, and other suggestions. The versions are bumped in both places, so the inheritance will be protected for future changes like we discussed for other plugins.

PR #1174, opened 2024-06-20
Add PEDump plugin

atcuno:
This PR adds the PEDump plugin, which provides the ability for an analyst to fully reconstruct an executable from any address within any process or the kernel.

Currently, Volatility 3 provides the `--dump` option for various plugins, but they are based upon the related executables being referenced from a well known data structure (kernel module list, process list, dlllist, etc.). In investigations involving reflectively loaded (memory only) DLLs and kernel drivers, Volatility 3 currently provides no method to reconstruct these.

Volatility 2 provided separate plugins for this capability (dlldump and moddump), but they are a bit awkward as they would either enumerate all of the DLLs or kernel modules by default, but otherwise have specialized, single purpose use if `--base` was set. It was also extra strange as dlldump was really geared towards DLLs tracked with a LDR_DATA_TABLE_ENTRY structure, which reflectively loaded ones don't get since they load themselves as opposed to using LoadLibary.

For these reasons, I broke the functionality out into its own plugin so people investigating a variety of code injection techniques can quickly get the malicious executables of out process or kernel memory.

### DIFF
--- a/volatility3/framework/plugins/windows/pedump.py
+++ b/volatility3/framework/plugins/windows/pedump.py
@@ -50,13 +50,13 @@ class PEDump(interfaces.plugins.PluginInterface):
             ),
         ]
 
-    def _write_pe(self, pe_table_name, layer_name, proc_offset, pid):
+    def _dump_pe(self, pe_table_name, layer_name, proc_offset, pid, base):
         try:
             file_handle = self.open(
                 "PE.{:#x}.{:d}.{:#x}.dmp".format(
                     proc_offset,
                     pid,
-                    self.config["base"],
+                    base,
                 )
             )
 
@@ -77,23 +77,25 @@ class PEDump(interfaces.plugins.PluginInterface):
         ) as excp:
             vollog.debug(f"Unable to dump dll at offset {self.config['base']}: {excp}")
             return None
+        finally:
+            file_handle.close()
 
-        file_handle.close()
         return file_handle.preferred_filename
 
-    def _dump_kernel(self, _, pe_table_name, session_layers):
+    def _dump_kernel(self, kernel, pe_table_name, base):
+        session_layers = modules.Modules.get_session_layers(
+            self.context, kernel.layer_name, kernel.symbol_table_name
+        )
+
         session_layer_name = modules.Modules.find_session_layer(
-            self.context, session_layers, self.config["base"]
+            self.context, session_layers, base
         )
 
         if session_layer_name:
             system_pid = 4
 
-            file_output = self._write_pe(
-                pe_table_name,
-                session_layer_name,
-                0,
-                system_pid,
+            file_output = self._dump_pe(
+                pe_table_name, session_layer_name, 0, system_pid, base
             )
 
             if file_output:
@@ -103,9 +105,8 @@ class PEDump(interfaces.plugins.PluginInterface):
                 "Unable to find a session layer with the provided base address mapped in the kernel."
             )
 
-    def _dump_processes(self, kernel, pe_table_name, _):
+    def _dump_processes(self, kernel, pe_table_name, base):
         filter_func = pslist.PsList.create_pid_filter(self.config.get("pid", None))
-        kernel = self.context.modules[self.config["kernel"]]
 
         for proc in pslist.PsList.list_processes(
             context=self.context,
@@ -121,11 +122,8 @@ class PEDump(interfaces.plugins.PluginInterface):
             )
             proc_layer_name = proc.add_process_layer()
 
-            file_output = self._write_pe(
-                pe_table_name,
-                proc_layer_name,
-                proc.vol.offset,
-                pid,
+            file_output = self._dump_pe(
+                pe_table_name, proc_layer_name, proc.vol.offset, pid, base
             )
 
             if file_output:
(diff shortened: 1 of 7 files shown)


volatility3/framework/plugins/windows/pedump.py
```diff
@@ -0,0 +1,178 @@
...
+        ) as excp:
+            vollog.debug(f"Unable to dump dll at offset {self.config['base']}: {excp}")
+            return None
+
+        file_handle.close()
+        return file_handle.preferred_filename
+
+    def _dump_kernel(self, _, pe_table_name, session_layers):
+        session_layer_name = modules.Modules.find_session_layer(
+            self.context, session_layers, self.config["base"]
```
ikelos: Ideally this would be passed `base` as a parameter.  Although this is private, threading it that way would make it more reusable by others (and since it is a swappable/reusable function, that seems like it would be useful in the future).

volatility3/framework/plugins/windows/pedump.py
```diff
@@ -0,0 +1,178 @@
...
+                yield pid, proc_name, file_output
+
+    def _generator(self):
+        kernel = self.context.modules[self.config["kernel"]]
+
+        pe_table_name = intermed.IntermediateSymbolTable.create(
+            self.context, self.config_path, "windows", "pe", class_types=pe.class_types
+        )
+
+        if self.config["kernel_module"] and self.config["pid"]:
```
ikelos, line 139: This whole flag or pid thing feels... a little ugly?  The main code in this whole plugin is `write_pe` (which should probably be `dump_pe` to be in keeping with other places in the codebase.  This then chooses whether to call one of two methods, but it specifies the method, and then each method takes one or another bogus unneeded parameter.  It feels a very strange way of structuring the code for a choice of two?  I'd also sooner just accept a DLL location (with better error checking in case you're handed a bad pe), and then add `--dump` flags into various other plugins that might want to write a PE?  Either way, if you're specifying the method and the parameters, there's no need to try and make the function signatures match, at that point there's very little being reused...

volatility3/framework/plugins/windows/pedump.py
```diff
@@ -0,0 +1,178 @@
...
+        except (
+            IOError,
+            exceptions.VolatilityException,
+            OverflowError,
+            ValueError,
+        ) as excp:
+            vollog.debug(f"Unable to dump dll at offset {self.config['base']}: {excp}")
+            return None
+
+        file_handle.close()
```
ikelos: consider using `finally` just to get in the habit?  It means that whether an exception happens or not the file handle gets closed...  https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions

ikelos (changes requested):
Happy with the core function `_write_pe` although it feels like it should be exposed as a `classmethod`, but the two calling functions with a matching function signature is overly complicating things.  Given there's only two choices, just code them up clearly with the parameters each needs and no more.

volatility3/framework/plugins/windows/pedump.py
atcuno: done

volatility3/framework/plugins/windows/pedump.py
atcuno: done

volatility3/framework/plugins/windows/pedump.py
atcuno: I changed it to dump_pe and the other suggestions on the code. I still think having a --kernel flag to specify the kernel gives the most clear instruction for analysts when they find a reflectively loaded kernel module in memory (probably inside volshell) and want to extract it out. Same for exes or DLLs with the --pid option when a PE is found in volshell or similar method.

Volatility 2 does it the way you suggest with throwing --dump into other existing plugins, but it makes the workflow and trying to explain the workflow to users very awkward. For example, Vol2 had dlldump, which by default would walk the list and dump everything, except if base was set, then it didn't actually walk the list at all and just used your given base. And even if the PE file you wanted wasn't a DLL, such as when process hollowing techniques map multiple EXE files into the same process, you still had to use a plugin named for DLLs to then extract the EXE. Similar awkwardness arose trying to use moddump in Vol2 for reflectively loaded kernel drivers (.sys files) that never even had a module structure created for them in kernel memory, since they were directly, memory only loaded by a rootkit. Trying to force that feature into Vol3's dlllist and modules would cause the same headaches and also prevent code like in https://github.com/volatilityfoundation/volatility3/pull/1183 from providing the very useful feature of extracting from a base address only in the processes where the DLL is truly in the real list (like when analyzing in the presence of of Hell's Gate and similar EDR-bypasses) and even with ASLR on since the mapping will be set in the same place on a per boot basis.

By having the separate plugin, analysts can just get properly reconstructed PE files from any address in any context, also with us providing cycling through the session layers automatically when --kernel is set.

There is also good error handling in there. Below you can see with --kernel and --pid used for both valid and invalid addresses. 

```
$ python3 vol.py -f data.lime -o bbb windows.pedump --pid 3692 --base 0x76bf0000
Volatility 3 Framework 2.7.1
PID	Process	File output
3692	msiexec.exe	PE.0x81ed83a0.3692.0x76bf0000.dmp

$ python3 vol.py -f data.lime-o bbb windows.pedump --pid 3692 --base 0x76bf0010
Volatility 3 Framework 2.7.1
PID	Process	File output

$ python3 vol.py -f data.lime -o bbb windows.pedump --kernel --base 0xb1ea9000
Volatility 3 Framework 2.7.1
PID	Process	File output
4	Kernel	PE.0x0.4.0xb1ea9000.dmp

$ python3 vol.py -f data.lime -o bbb windows.pedump --kernel --base 0xb1ea9100
Volatility 3 Framework 2.7.1
PID	Process	File output
```

ikelos (commented):
Ok, I'm reasonably happier with that, I would still expect the various `dump_` method to be parameterized properly and turning into class methods so that other plugins that may want to can call them though.  Seems a little bit restrictive to not have some expose method so that other plugins can make use of the functionality?

ikelos (approved):

ikelos (commented):
Did exactly the same thing again!  5:\  Sorry for the spam, just wanted to comment, not approve or request changes...

ikelos (changes requested):
Actually, for this one I think making them expose does make sense, so I am gonna mark is as request changes...

atcuno:
Closing this and will make a new one against the latest code

PR #1177, opened 2024-06-21
118 missing plugins threads timers

### LINKED ISSUE #118: Missing Plugin -> threads

superponible:
Implementation of timers plugin from volatility2, with column added showing symbols within the module matching the routine address.

<pre>
Offset	DueTime	Period(ms)	Signaled	Routine	Module	Symbol

0xf8057edffc40	0x000000c9:0x2d5e40c1	0	-	0xf8057edf3f40	pdc	-
0xba8ad6587800	0x000000c9:0x5742a45f	0	-	0xf8058011fb00	BTHport	-
0xba8ad4db2180	0x40002763:0x637332b1	0	-	0xf8057c1a6dc0	ntoskrnl	PopFxIdleTimeoutDpcRoutine
0xf8057ca2e820	0x00000053:0x6c0a1d5a	30000	Yes	0xf8057c125800	ntoskrnl	PopCheckForIdleness
0xba8ad66e1740	0x00000053:0x5efcaa90	16	Yes	0xf80582720b80	vm3dmp	-
0xba8ad46b5bfb	0x00000053:0x7528927e	0	-	0xf8057c1b13b0	ntoskrnl	PopThermalZoneDpc
0xf8057ca003e0	0x00000054:0xb44a1ed3	0	-	0xf8057c1429c0	ntoskrnl	CmpLazyFlushDpcRoutine
0xba8ad4d5d540	0x00000053:0x64c056bb	2000	-	0xf8057f1f3150	storport	-
0xf80582306a50	0x00000053:0x64c0598c	0	-	0xf805822c81b0	afd	DEVPKEY_Device_PowerData
0xba8ad67231b8	0x00000053:0x5f376a6c	1000	Yes	0xf8057ebb8c50	NDIS	-
0xba8ad4673e50	0x00000053:0x78e64de4	0	-	0xf8057c184d60	ntoskrnl	ExpCenturyDpcRoutine
0xf8057f4abbd0	0x00000053:0x60eeb712	0	-	0xf8057f4398f0	Ntfs	-
0xba8ad4e241c0	0x40002710:0x0158ab76	0	-	0xf8057c1a6dc0	ntoskrnl	PopFxIdleTimeoutDpcRoutine
0xf8057f2544e0	0x0000005c:0x55733d1d	0	-	0xf8057f1f3d90	storport	-
0xf8057f254620	0x0000005c:0x55733d1d	0	-	0xf8057f1f3d90	storport	-
0xba8ad4e11950	0x40002710:0x0170a209	0	-	0xf8057c1a6dc0	ntoskrnl	PopFxIdleTimeoutDpcRoutine
</pre>

### DIFF
--- a/volatility3/framework/plugins/windows/timers.py
+++ b/volatility3/framework/plugins/windows/timers.py
@@ -7,7 +7,6 @@ import logging
 from typing import Iterator, List, Tuple, Iterable
 
 from volatility3.framework import (
-    exceptions,
     layers,
     renderers,
     interfaces,
--- a/volatility3/framework/plugins/windows/timers.py
+++ b/volatility3/framework/plugins/windows/timers.py
@@ -1,0 +1,264 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import logging
+
+from typing import Iterator, List, Tuple, Iterable
+
+from volatility3.framework import exceptions, layers, renderers, interfaces, constants, symbols
+from volatility3.framework.configuration import requirements
+from volatility3.framework.objects import utility
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.symbols.windows import versions
+from volatility3.plugins.windows import ssdt
+
+vollog = logging.getLogger(__name__)
+
+
+class Timers(interfaces.plugins.PluginInterface):
+    """Print kernel timers and associated module DPCs"""
+
+    _required_framework_version = (2, 0, 0)
+    _version = (1, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Windows kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.PluginRequirement(
+                name="ssdt", plugin=ssdt.SSDT, version=(1, 0, 0)
+            ),
+        ]
+
+    @classmethod
+    def get_kernel_module(
+        cls,
+        context: interfaces.context.ContextInterface,
+        layer_name: str,
+        symbol_table: str,
+    ):
+        """Returns the kernel module based on the layer and symbol_table"""
+        virtual_layer = context.layers[layer_name]
+        if not isinstance(virtual_layer, layers.intel.Intel):
+            raise TypeError("Virtual Layer is not an intel layer")
+
+        kvo = virtual_layer.config["kernel_virtual_offset"]
+
+        ntkrnlmp = context.module(symbol_table, layer_name=layer_name, offset=kvo)
+        return ntkrnlmp
+
+    @classmethod
+    def get_kpcrs(
+        cls,
+        context: interfaces.context.ContextInterface,
+        layer_name: str,
+        symbol_table: str,
+    ) -> interfaces.objects.ObjectInterface:
+        """Returns the KPCR structure for each processor
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            symbol_table: The name of an existing symbol table containing the kernel symbols
+            config_path: The configuration path within the context of the symbol table to create
+
+        Returns:
+            The _KPCR structure for each processor
+        """
+
+        ntkrnlmp = cls.get_kernel_module(context, layer_name, symbol_table)
+        cpu_count_offset = ntkrnlmp.get_symbol("KeNumberProcessors").address
+        cpu_count = ntkrnlmp.object(
+            object_type="unsigned int", layer_name=layer_name, offset=cpu_count_offset
+        )
+        processor_block = ntkrnlmp.object(
+            object_type="pointer",
+            layer_name=layer_name,
+            offset=ntkrnlmp.get_symbol("KiProcessorBlock").address,
+        )
+        processor_pointers = utility.array_of_pointers(
+            context=context,
+            array=processor_block,
+            count=cpu_count,
+            subtype=symbol_table + constants.BANG + "_KPRCB",
+        )
+        for pointer in processor_pointers:
+            kprcb = pointer.dereference()
+            reloff = ntkrnlmp.get_type("_KPCR").relative_child_offset("Prcb")
+            kpcr = context.object(
+                symbol_table + constants.BANG + "_KPCR",
+                offset=kprcb.vol.offset - reloff,
+                layer_name=layer_name,
+            )
+            yield kpcr
+
+    @classmethod
+    def list_timers(
+        cls,
+        context: interfaces.context.ContextInterface,
+        layer_name: str,
+        symbol_table: str,
+    ) -> Iterable[Tuple[str, int, str]]:
+        """Lists all kernel timers.
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            layer_name: The name of the layer on which to operate
+            symbol_table: The name of the table containing the kernel symbols
+
+        Yields:
+            A _KTIMER entry
+        """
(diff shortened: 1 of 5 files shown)


volatility3/framework/plugins/windows/timers.py
```diff
@@ -0,0 +1,266 @@
...
+
+    @classmethod
+    def get_kernel_module(
+        cls,
+        context: interfaces.context.ContextInterface,
+        layer_name: str,
+        symbol_table: str,
+    ):
+        """Returns the kernel module based on the layer and symbol_table"""
+        virtual_layer = context.layers[layer_name]
```
ikelos: It's possible you'll get handed a bogus `layer_name`.  Not sure if you want to check/catch that or just let it throw a key error for people?

volatility3/framework/plugins/windows/timers.py
```diff
@@ -0,0 +1,266 @@
...
+        if not isinstance(virtual_layer, layers.intel.Intel):
+            raise TypeError("Virtual Layer is not an intel layer")
+
+        kvo = virtual_layer.config["kernel_virtual_offset"]
+
+        ntkrnlmp = context.module(symbol_table, layer_name=layer_name, offset=kvo)
+        return ntkrnlmp
+
+    @classmethod
+    def get_kpcrs(
```
ikelos: Is this the best plugin to embed this functionality in?  I recall KPCRs being used for lots of stuff.  I dunno whether a dedicated KPCR plugin with this static classmethod would be better?

volatility3/framework/plugins/windows/timers.py
```diff
@@ -0,0 +1,266 @@
...
+                description="Windows kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.PluginRequirement(
+                name="ssdt", plugin=ssdt.SSDT, version=(1, 0, 0)
+            ),
+        ]
+
+    @classmethod
+    def get_kernel_module(
```
ikelos: This is interesting, rather than using the `kernel` module, this constructs a new module?  The functionality seems fine but... redundant?  `context.modules(self.config['kernel'])` should get you to the same place (parameterizing `self.config['kernel']` of course)?

volatility3/framework/plugins/windows/timers.py
```diff
@@ -0,0 +1,266 @@
...
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            layer_name: The name of the layer on which to operate
+            symbol_table: The name of the table containing the kernel symbols
+
+        Yields:
+            A _KTIMER entry
+        """
+        ntkrnlmp = cls.get_kernel_module(context, layer_name, symbol_table)
```
ikelos: Again, I'd just parameterize a module name rather than rerunning this (and making new modules) each time?

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -994,6 +995,81 @@ def privileges(self):
...
+    def get_signaled(self):
+        if self.Header.SignalState:
+            return "Yes"
+        return "-"
+
+    def get_raw_dpc(self):
+        """Returns the encoded DPC since it may not look like a pointer after encoding"""
+        symbol_table_name = self.get_symbol_table_name()
+        ulonglong_type = self._context.symbol_space.get_type(
+            symbol_table_name + constants.BANG + "unsigned long long"
```
ikelos: There is a question as to whether this type will always exist (recent versions of clang call it something like `long unsigned int` I think?).  It's probably fine and I don't know what else we'd do (discussion is still going on at #1041 ).  If this is a pointer though, then it should be a pointer which will get the size right automatically.

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -994,6 +995,81 @@ def privileges(self):
...
+                offset=ntkrnlmp.get_symbol("KiWaitNever").address,
+            )
+            wait_always = ntkrnlmp.object(
+                object_type="unsigned long long",
+                offset=ntkrnlmp.get_symbol("KiWaitAlways").address,
+            )
+
+            low_byte = (wait_never) & 0xFF
+            entry = utility.rol(self.get_raw_dpc() ^ wait_never, low_byte)
+            swap_xor = self.vol.offset | 0xFFFF000000000000
```
ikelos: Is this mask value specific to the way the code works, or is it a canonicalized address (in which case we have code for that) or is it based on the address space masking (ie, does it work if the processor decides to have more/fewer than normal bits for paging).

ikelos (commented):
Looks good, happy with the bit shifting being added to utility, I think that's the right place for it.  A couple bits and pieces regarding `unsigned long long` which cropped up recently, and more importantly  some bits about masking...

volatility3/framework/symbols/windows/extensions/__init__.py
superponible: @ikelos i believe it's just to get a canonicalized address.  the code from the other MR had a comment about doing it this way until something was merged... but I didn't see that that was merged. where is the code that could handle this rather than doing this `OR`

volatility3/framework/symbols/windows/extensions/__init__.py
superponible: changed to pointer in latest commit

volatility3/framework/plugins/windows/timers.py
superponible: done

volatility3/framework/plugins/windows/timers.py
superponible: done

volatility3/framework/plugins/windows/timers.py
superponible: added kpcrs plugin

volatility3/framework/plugins/windows/timers.py
superponible: no longer there after other refactoring

volatility3/framework/symbols/windows/extensions/__init__.py
superponible: think I found it, check the latest commit

volatility3/framework/symbols/windows/__init__.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, *args, **kwargs) -> None:
         self.set_type_class("_VACB", extensions.VACB)
         self.set_type_class("_POOL_TRACKER_BIG_PAGES", pool.POOL_TRACKER_BIG_PAGES)
         self.set_type_class("_IMAGE_DOS_HEADER", pe.IMAGE_DOS_HEADER)
+        self.set_type_class("_KTIMER", extensions.KTIMER)
```
ikelos, line 42: Will this always be present, even in old versions of windows?  If this symbol isn't present in some old versions then this should be `optional_set_type_class`.  Otherwise this is fine, just want to check before we merge...

ikelos (commented):
Thanks that looks like most everything, just one little point I wanted to check and then this can go in...  5:)

volatility3/framework/symbols/windows/__init__.py
superponible: Yes, it should be in every version. And I've run this against several versions all the way back to XP and it works.

ikelos: Cool, happy with that then, just wanted to check  5:)

volatility3/framework/symbols/windows/extensions/__init__.py
ikelos: Sorry I didn't get back to you sooner, thanks for finding it!

ikelos (approved):
Yep, all looks good now, thanks!

PR #1183, opened 2024-06-26
Windows: Adding name & base filtering for --dump on dlllist

hsarkey:
Adding ability to filter using --name and --base in dlllist. This filtering works with  --dump, so only the filtered results are dumped vs. generic column filtering which still dumps everything regardless of filters applied.


volatility3/framework/plugins/windows/dlllist.py
```diff
@@ -53,6 +54,22 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
                 optional=True,
             ),
+            requirements.StringRequirement(
+                name="name",
+                description="Specify a regular expression to match dll name(s)",
+                optional=True,
+            ),
+            requirements.IntRequirement(
+                name="base",
+                description="Specify a base address",
```
ikelos: This should specify an address space (virtual or physical) when talking about an offset, since it will matter and confuse people if it isn't clearly specified.

ikelos (commented):
This looks ok, my slight concern is we've not got several filtering options which may overlap in usual ways.  `Base` and `offset` seem like they might interact, and if they're both specified they may block a process from being dump for some reason.  Complexity can be troublesome.  Also, it presumably may be better to make it (and possibly offset) into `ListRequirements` so that many can be specified rather than just one?

volatility3/framework/plugins/windows/dlllist.py
atcuno: @ikelos offset is the physical address of the process' `_EPROCESS` structure, and `base` is the virtual address of the DLL to extract from the process (assuming its in the list). They won't conflict. "base address" is a common way to describe where a DLL or exe is loaded in a process or a kernel driver in the kernel.

For the List requirement, for the process offset at least that would be a big haul as plugins use this filter which only accepts one offset as int. So handles and dlllist would also have to change as well as the filter:

https://github.com/volatilityfoundation/volatility3/blob/1e58547aa93e99033e8f361fb8bef422d52231b7/volatility3/framework/plugins/windows/psscan.py#L92

https://github.com/search?q=repo%3Avolatilityfoundation%2Fvolatility3%20create_offset_filter&type=code

I also don't see multiple physical offsets being sent in at once often (or ever?) as that would mean a kernel driver hid more than one process from the process list, which I don't remember seeing before.

ikelos: Ok, happy to keep it as single, but I think it best if the description for both base and offset be updated with indicators to their respective spaces (virtual and physical) just to clarify for people that don't know the terminology as well...

It may still be more useful for `handles` and `dllist` to take lists, but I won't require it as part of this PR, just an observation to make existing plugins better if we think of something that might help...  5:)

atcuno: ok the physical part is in the --offset already:

https://github.com/volatilityfoundation/volatility3/blob/06a2c0c76c3a80a0709bca2cff60d2718b7bb93d/volatility3/framework/plugins/windows/dlllist.py#L54

I can work with Hannah to get the virtual into the base

ikelos (approved):
Thanks for making the change, looks good to go...

PR #1186, opened 2024-06-27
Add timeliner support to userassist

atcuno:
These records hold the time programs last executed along with their path, number of times ran, and other details depending on the OS version. We should include these in the timeliner interface.

### DIFF
--- a/volatility3/framework/plugins/windows/registry/userassist.py
+++ b/volatility3/framework/plugins/windows/registry/userassist.py
@@ -336,7 +336,6 @@ class UserAssist(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterfac
             )
             yield result
 
-
     def generate_timeline(self):
         self._reg_table_name = intermed.IntermediateSymbolTable.create(
             self.context, self._config_path, "windows", "registry"
@@ -345,7 +344,9 @@ class UserAssist(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterfac
         for row in self._generator():
             _depth, row_data = row
             # check the name and the timestamp to not be empty
-            if isinstance(row_data[5], str) and not isinstance(row_data[10], renderers.NotApplicableValue):
+            if isinstance(row_data[5], str) and not isinstance(
+                row_data[10], renderers.NotApplicableValue
+            ):
                 description = f"UserAssist: {row_data[5]} {row_data[2]} ({row_data[7]})"
                 yield (description, timeliner.TimeLinerType.MODIFIED, row_data[10])
 
--- a/volatility3/framework/plugins/windows/registry/userassist.py
+++ b/volatility3/framework/plugins/windows/registry/userassist.py
@@ -17,11 +17,12 @@ from volatility3.framework.layers.registry import RegistryHive
 from volatility3.framework.renderers import conversion, format_hints
 from volatility3.framework.symbols import intermed
 from volatility3.plugins.windows.registry import hivelist
+from volatility3.plugins import timeliner
 
 vollog = logging.getLogger(__name__)
 
 
-class UserAssist(interfaces.plugins.PluginInterface):
+class UserAssist(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
     """Print userassist registry keys and information."""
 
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/registry/userassist.py
```diff
@@ -335,6 +336,20 @@ def _generator(self):
             )
             yield result
 
+    def generate_timeline(self):
+        self._reg_table_name = intermed.IntermediateSymbolTable.create(
```
ikelos: Is it necessary for this to be loaded before the generator is called?  Feels like it either belongs in the generator or should be parameterized rather than a self variable...

volatility3/framework/plugins/windows/registry/userassist.py
```diff
@@ -335,6 +336,20 @@ def _generator(self):
...
+            self.context, self._config_path, "windows", "registry"
+        )
+
+        for row in self._generator():
+            _depth, row_data = row
+            # check the name and the timestamp to not be empty
+            if isinstance(row_data[5], str) and not isinstance(
```
ikelos, line 347: In case these indices change, *strictly* we should be checking the treegrid and interrogating it for which columns it put out and then referencing those specifically.  I assume you're going to find me an existing example that also does it this way, so I'll let it slide, but it is something we should fix across the board.  Fixing this one now would tell us a) how difficult it will be and b) act as a template for changing the others...

ikelos (commented):
Just a couple of points.  The self._reg_table_name I would like to get resolved before this goes in, other one I'm willing to let slide...

volatility3/framework/plugins/windows/registry/userassist.py
atcuno: Yes, this was my first time writing a generate_timeline function, and when searching across the code base - the other previous ones all used the indexes:

https://github.com/volatilityfoundation/volatility3/blob/4f69638d4d067a1fedc85f37cb62b628aa7f3256/volatility3/framework/plugins/windows/pslist.py#L291

https://github.com/volatilityfoundation/volatility3/blob/4f69638d4d067a1fedc85f37cb62b628aa7f3256/volatility3/framework/plugins/windows/sessions.py#L110

I can see how/why some name (string) based way for the column looks cleaner, but at the same time, if the indexes produced change in generator() then it should be enforced for changes in generate_timeline() of the same plugin before accepting.

Also, I am not sure how inspecting the TreeGrid would function in cases where a column name changed or was removed after generate_timeline is written. I would much prefer the index way to backtrace there since its obvious versus some vollog type print that says a column can no longer be found.

ikelos: Yeah, that's reasonable.  Ok, you convinced me.  5;)

volatility3/framework/plugins/windows/registry/userassist.py
atcuno: Yes, it seemed strange to me too as when I was initially testing the plugin I saw where a sample would get userassist output normally but then my timeline code wasn't reporting anything. I then saw the run function had that setup:

https://github.com/volatilityfoundation/volatility3/blob/8525edd3331c6ccd967c8048d61adf5d2c3e1015/volatility3/framework/plugins/windows/registry/userassist.py#L354

So I copied it over to the generate_timeline and it worked.

I thought of moving that `self._reg_table_name` into the generator, but was nervious of side effects since it also seems to play a role in the registry layer:

https://github.com/search?q=repo%3Avolatilityfoundation%2Fvolatility3%20_reg_table_name&type=code

Although the one set in the registry layer doesn't seem referenced from anywhere.

ikelos: Huh, that is a bit strange.  In this plugin at least I'd like to get it moved to the generator.  In the registry layer, we might reference it just by type name and volatility may figure it out, but that's something we can test another time.  Having it saved but not be referenced is fine, getting it loaded into the context is the important part.

atcuno: Fixed and tested.

ikelos (approved):
Cool, looks good thanks!

Issue #1162, opened 2024-06-12
Page error in layer when dumping Windows 10 hashes


arepi-nemui:
**Describe the bug**
After commit e5a5b895771b655d21c36689c33a534034c31e36, volatility is no longer able to run various modules on a memory dump from a Windows 10 system, such as windows.hashdump.Hashdump and windows.info.Info.

**Context**
Volatility Version:  Commit e5a5b895771b655d21c36689c33a534034c31e36 and later
Operating System:  Linux
Python Version:  3.12.3
Suspected Operating System:  Windows 10
Command:  `python vol.py -f memory.dmp -vvv windows.hashdump.Hashdump`

**To Reproduce**

```
git bisect start
git bisect bad develop
git bisect good 55dd39f2
git bisect run python vol.py -f memory.dmp windows.hashdump.Hashdump
```

**Example output**
```
python vol.py -f memory.dmp -vvv windows.hashdump.Hashdump
Volatility 3 Framework 2.7.1
INFO     volatility3.cli: Volatility plugins path: ['/home/ctf/volatility3/volatility3/plugins', '/home/ctf/volatility3/volatility3/framework/plugins']
INFO     volatility3.cli: Volatility symbols path: ['/home/ctf/volatility3/volatility3/symbols', '/home/ctf/volatility3/volatility3/framework/symbols']
INFO     volatility3.framework.automagic: Detected a windows category plugin
INFO     volatility3.framework.automagic: Running automagic: ConstructionMagic
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Hashdump.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Hashdump.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Hashdump.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Hashdump.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Hashdump.kernel
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Hashdump.kernel.layer_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Hashdump.kernel.layer_name
[...]

PR #1190, opened 2024-07-04
Windows - psscan plugin fixes

gcmoreira:
`windows.psscan.PsScan plugin` breaks when dumping (invalid) processes.

On the one hand, this PR allows the `windows.psscan.PsScan` plugin to handle invalid page errors enabling it to continue searching and dumping processes.

On the other hand, it improves the smear memory checks in the EPROCESS class avoiding wasted time processing garbage.

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -22,7 +22,7 @@ from volatility3.framework.interfaces.objects import ObjectInterface
 from volatility3.framework.layers import intel
 from volatility3.framework.renderers import conversion
 from volatility3.framework.symbols import generic
-from volatility3.framework.symbols.windows.extensions import kdbg, pe, pool
+from volatility3.framework.symbols.windows.extensions import pool
 
 vollog = logging.getLogger(__name__)
 
@@ -611,10 +611,21 @@ class EPROCESS(generic.GenericIntelProcess, pool.ExecutiveObject):
 
                 ctime = self.get_create_time()
                 if not isinstance(ctime, datetime.datetime):
+                    # A process must have a creation time
                     return False
 
                 if not (1998 < ctime.year < 2030):
                     return False
+
+                etime = self.get_exit_time()
+                if isinstance(etime, datetime.datetime):
+                    if not (1998 < etime.year < 2030):
+                        return False
+
+                    # Exit time, if available, must be after the creation time
+                    # At this point, we are sure both are datetimes, so let's compare them
+                    if ctime > etime:
+                        return False
 
             # NT pids are divisible by 4
             if self.UniqueProcessId % 4 != 0:
@@ -633,7 +644,11 @@ class EPROCESS(generic.GenericIntelProcess, pool.ExecutiveObject):
             if dtb & ~0xFFF == 0:
                 return False
 
-            ## TODO: we can also add the thread Flink and Blink tests if necessary
+            # Quick smear test on thread Flink and Blink
+            kernel = 0x80000000  # Yes, it's a quick test
+            list_head = self.ThreadListHead
+            if list_head.Flink < kernel or list_head.Blink < kernel:
+                return False
 
         except exceptions.InvalidAddressException:
             return False
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -614,12 +614,13 @@ class EPROCESS(generic.GenericIntelProcess, pool.ExecutiveObject):
                     # A process must have a creation time
                     return False
 
-                if not (1998 < ctime.year < 2030):
+                current_year = datetime.datetime.now().year
+                if not (1998 < ctime.year < current_year + 10):
                     return False
 
                 etime = self.get_exit_time()
                 if isinstance(etime, datetime.datetime):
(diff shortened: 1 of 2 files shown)


gcmoreira:
@atcuno

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -633,7 +644,11 @@ def is_valid(self) -> bool:
             if dtb & ~0xFFF == 0:
                 return False
 
-            ## TODO: we can also add the thread Flink and Blink tests if necessary
+            # Quick smear test on thread Flink and Blink
+            kernel = 0x80000000  # Yes, it's a quick test
```
ikelos, line 649: This... doesn't feel robust, and it also feels like it only works for 32-bit systems?  I mean, it'll work to filter out *some* bad entries, but?  Is there anything we can do about this or does it really need to be this way?

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -611,11 +611,22 @@ def is_valid(self) -> bool:
...
                 if not isinstance(ctime, datetime.datetime):
+                    # A process must have a creation time
                     return False
 
                 if not (1998 < ctime.year < 2030):
                     return False
 
+                etime = self.get_exit_time()
+                if isinstance(etime, datetime.datetime):
+                    if not (1998 < etime.year < 2030):
```
ikelos: Not sure about limiting the max year to 2030?  I fear in 6 years someone's going to use this and complain.  Is there a reason we capped it at > 2030 ?  Could we make it 3000 or something outlandishly far in the future?

ikelos (commented):
Looks good happy with the psscan stuff.  A few questions about some of the checks that were added to the extension though...

volatility3/framework/symbols/windows/extensions/__init__.py
gcmoreira: I agree.  I followed the same approach we used earlier with `creation_time` line 617. 
Anyway, I'm with you; perhaps we should use 2050? The idea is to filter out weird stuff, so I wouldn't go too far beyond that.

volatility3/framework/symbols/windows/extensions/__init__.py
gcmoreira: Yeah, I thought about this before submitting the PR. The idea of this is to do a quick check, it doesn't need to be perfect. Also, that will be the safer way to go supporting both 32/64bit as it only filter out the lowest addresses. 
```python
           if list_head.Flink < kernel or list_head.Blink < kernel:
                return False
```
In 64bit systems, there will be cases where the link_head addresses fall within the user virtual address space or in the canonical hole (invalid), and this filter won't catch them. We could use `0xffff800000000000` in those cases, however, we would likely face similar issues to those you mentioned [here](https://github.com/volatilityfoundation/volatility3/issues/1162#issuecomment-2209502111)

volatility3/framework/symbols/windows/extensions/__init__.py
ikelos: Ok, 2050 is better than 2030 (26 years is better than 6 years), but we probably want to leave a comment somewhere in here pointing out that if people come hunting in this file for a bug, it might be that legitimate values over a specific time are weeded out as being invalid.  Or we could really push the boat out, find the time the program is running and add X years onto that?  That'd at least keep the code from failing ever far off into the future?

volatility3/framework/symbols/windows/extensions/__init__.py
ikelos: Ok, I'd sooner get some bad apples than weed out valid ones, so happy to stick with this.

As per another conversation though, I think that should be 0xffff000000000000 rather than 0xffff800000000000 unless we're intentionally trying to keep out kernel memory.

volatility3/framework/symbols/windows/extensions/__init__.py
gcmoreira: with "the time the program is running" you mean the current/system time?

ikelos: Yeah, so someone with their system time set to 2060 get years up to 2070 as valid?

ikelos: I know it seems a long way off, but I guarantee you no one will think of it except now or when it become or a problem...  5;P  However, we can leave some problems to future us if we're not that fussed...  5:)

gcmoreira: hehe.. yeah.. I know.. I just wonder if it could be an issue with some users running it,let's say, on a VM with wrong time set or something edge case like that.. it's like we are adding a dependency ... but I agree that will be the best in terms of code maintenance.. I like that part

ikelos: Well, ten years in the future is pretty unusual, and anything less than that should be accepted (and most people testing things set their clocks back), so if we go for "current year + 10".  It shouldn't add a dependency and it'll be better than digging into why it won't work in ten years time...  5:)

gcmoreira: It sounds good. Done

ikelos (approved):
Looks good, thanks for making the changes!  5:)

PR #1192, opened 2024-07-06
Implement the fix that was reverted as part of #1162.

ikelos:
This should properly widen the test by a single byte, hopefully solving the previous issue being encountered (where the check wasn't conducted at all).

Hopefully this can be tested in #1162 before it goes live, since it touches a crucial part of the codebase.

### DIFF
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -180,7 +180,9 @@ class Intel(linear.LinearlyMappedLayer):
         position = self._initial_position
         entry = self._initial_entry
 
-        if self.minimum_address > offset > self.maximum_address:
+        if not (
+            self.minimum_address <= (offset & self.address_mask) <= self.maximum_address
+        ):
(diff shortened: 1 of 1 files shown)


arepi-nemui:
Works for me.

ikelos:
Thanks for testing.  @gcmoreira if you're happy with this I'll merge it...

gcmoreira:
@ikelos Sorry, I'm late to this, but I'm happy with your fix. Thanks.

It would be good to have a test case to detect what happened in #1162. Not specifically for Hashdump but for the address translation layer itself. At least make it sure we can resolve the same mappings than before the changes in a new PR, i.e. using translationlayer.mapping() results.

ikelos:
At the moment, the test cases we have are some example images we run some standard plugins against.  If someone would like to implement more rigorous testing, I'm happy to pull it in but the complexity is quite daunting particularly given my limited time to maintain the project.

gcmoreira:
I understand, no problem.

PR #1193, opened 2024-07-07
Initial draft of complete-vad vadyarascan

### LINKED ISSUE #1155: Windows YARA scan across page boundaries
I have a use case where I need to apply a YARA rule that in many cases spans multiple memory pages. I thought that `windows.vadyarascan` was made for this purpose - i.e., that it passes contiguous virtual memory areas belonging to a task to the scanner. However, that's not the case - chunks of 4K are scanned individually. To me this means I might as well have gone and executed YARA manually against the entire dump file (except that it won't tell me what PID it found a match in).

This makes me question the usefulness of YARA scanning in Volatility in general, as it may give you a false sense of security. If you have a YARA rule that generically matches a certain malware family, and you don't know beforehand if any of the strings happen to cross a page boundary (which can always happen by chance), you can never be sure if Volatility just didn't find it or if it's indeed not present.

I'd like a plugin or option that scans processes with the same semantics as if the scan were executed on a live system. Meaning contiguous virtual memory is scanned in one piece (or at least in bigger chunk sizes with a slight overlap for each chunk).

\-

**Additional information**
My input file is an ELF produced by applying VirtualBox's `debugvm` command to a Windows VM.

### LINKED ISSUE #1189: Critical bug in VADs Yara scanning due to scanning only one page at a time
I was putting together a lab guide for an upcoming workshop and ran into a critical bug in the way Volatility processes Yara searches inside of VADs. In particular, we are scanning one page at a time, even for VADs that are beyond a page size. 

I am filing this as a ticket instead of me just fixing it as I am not comfortable enough with changing all the scanning internals. Also, scanning one page at a time is the best way in the physical/file layer since RAM has the pages scattered, but it needs to be scanning entire VADs at a time in process memory as otherwise we will miss valid hits as I will detail here in the ticket. Obviously, that will require skipping VADs over a certain size (1GB is probably a good boundary), since the yara library will need a string holding the entire VAD's contents at once.

Details of the bug are below:

This is a yara rule we have that students use in training against a sample infected with stuxnet:

```
rule s {
    strings:
        $b = "SeLoadDriverPrivilege" wide
        $a = "Realtek Semiconductor" wide
    condition:
        all of them
}
```

"all of them" means both strings have to be present.

I noticed that Volatility 3 did not report any hits with this rule even though Vol2 always has. Digging in, if I made the rule "any of them", which means if either are present, then we got hits:

```
 |   Offset |  PID | Rule | Component |                                                                                                                         Value
* | 0x8cda8c | 1928 | stux |        $a |                                                       53 00 65 00 4c 00 6f 00 61 00 64 00 44 00 72 00 69 00 76 00 65 00 72 00
* | 0x988424 | 1928 | stux |        $b |                                                                52 65 61 6c 74 65 6b 20 53 65 6d 69 63 6f 6e 64 75 63 74 6f 72
* | 0x988498 | 1928 | stux |        $b |                                                                52 65 61 6c 74 65 6b 20 53 65 6d 69 63 6f 6e 64 75 63 74 6f 72
* | 0x988945 | 1928 | stux |        $b | 52 00 65 00 61 00 6c 00 74 00 65 00 6b 00 20 00 53 00 65 00 6d 00 69 00 63 00 6f 00 6e 00 64 00 75 00 63 00 74 00 6f 00 72 00
```

```
>>> binascii.unhexlify("52 65 61 6c 74 65 6b 20 53 65 6d 69 63 6f 6e 64 75 63 74 6f 72".replace(" ",""))
'Realtek Semiconductor'

>>> binascii.unhexlify("53 00 65 00 4c 00 6f 00 61 00 64 00 44 00 72 00 69 00 76 00 65 00 72 00".replace(" ","").replace("00",""))
'SeLoadDriver'
```

But notice that the two strings are on different pages - 0x8cda8c for one and then the 0x98... for the others, but that those pages are inside the same VAD:

```
$ python3 vol.py -q -f stux.raw windows.vadinfo --pid 1928 --address 0x8cda8c
Volatility 3 Framework 2.7.1
PID	Process	Offset	Start VPN	End VPN	Tag	Protection	CommitCharge	PrivateMemory	Parent	File	File output

1928	lsass.exe	0x82232950	0x870000	0x9a7fff	Vad 	PAGE_EXECUTE_READWRITE	0	0	0x82386728	N/A	Disabled

$ python3 vol.py -q -f stux.raw windows.vadinfo --pid 1928 --address 0x988945
Volatility 3 Framework 2.7.1
PID	Process	Offset	Start VPN	End VPN	Tag	Protection	CommitCharge	PrivateMemory	Parent	File	File output

1928	lsass.exe	0x82232950	0x870000	0x9a7fff	Vad 	PAGE_EXECUTE_READWRITE	0	0	0x82386728	N/A	Disabled
```

For Vadyarascan to be usable in real investigations it needs to be sending an entire VADs worth of data to the yara library to scan, otherwise we will constantly have issues like this, and the vast majority of production rules have where more than one string must match for the rule to match. File/Physical layer scanning should stay on the 4096 boundary though so I am hoping there is some easy way to parameterize the chunk sizes.

This is where I verified the address being scanned and the amount of data per call:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/yarascan.py#L45

ikelos:
Ok, here's an initial version of a change to vadyarascan that should scan a vad in its entirety when doing yara scanning.

Essentially it ignores the scanning framework, because we've already got the process layer to make the vads contiguous so we don't need any of the rebuilding/paging that the scanning framework does.  It's a little bit duplicative in places, but not a huge amount and we use the option/rule creation code from yarascan, so this should be ok.  Please could people test it for me...

Closes #1189.
Closes #1155.

### DIFF
--- a/volatility3/framework/plugins/windows/vadyarascan.py
+++ b/volatility3/framework/plugins/windows/vadyarascan.py
@@ -2,10 +2,11 @@
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
 
+import io
 import logging
 from typing import Iterable, List, Tuple
 
-from volatility3.framework import interfaces, renderers
+from volatility3.framework import exceptions, interfaces, renderers
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints
 from volatility3.plugins import yarascan
@@ -18,7 +19,7 @@ class VadYaraScan(interfaces.plugins.PluginInterface):
     """Scans all the Virtual Address Descriptor memory maps using yara."""
 
     _required_framework_version = (2, 4, 0)
-    _version = (1, 0, 1)
+    _version = (1, 1, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
@@ -32,11 +33,8 @@ class VadYaraScan(interfaces.plugins.PluginInterface):
             requirements.PluginRequirement(
                 name="pslist", plugin=pslist.PsList, version=(2, 0, 0)
             ),
-            requirements.VersionRequirement(
-                name="yarascanner", component=yarascan.YaraScanner, version=(2, 0, 0)
-            ),
             requirements.PluginRequirement(
-                name="yarascan", plugin=yarascan.YaraScan, version=(1, 2, 0)
+                name="yarascan", plugin=yarascan.YaraScan, version=(1, 3, 0)
             ),
             requirements.ListRequirement(
                 name="pid",
@@ -59,6 +57,8 @@ class VadYaraScan(interfaces.plugins.PluginInterface):
 
         filter_func = pslist.PsList.create_pid_filter(self.config.get("pid", None))
 
+        sanity_check = 0x1000 * 0x1000 * 0x1000
+
         for task in pslist.PsList.list_processes(
             context=self.context,
             layer_name=kernel.layer_name,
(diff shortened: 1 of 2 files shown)


eve-mem:
Lgtm. I'll do a touch more testing but it seems good. 

Should we also make some changes to the generic yarascan plugin at the same time? Mayne leave that for another PR

dadokkio:
Speaking about yarascan I was testing [yara-x](https://virustotal.github.io/yara-x/) instead of yara-python for matching since it should be faster (even not so much for initial testing), maintained and easier to use.  
I've created a[ fork for this](https://github.com/volatilityfoundation/volatility3/compare/develop...dadokkio:volatility3:develop) trying to keep the same output rendering and it's working also with depending plugins like vadyarascan.

Some considerations: 
- there is no function to obtain matched value so I took that slicing original data
- the compiled yara-python rule is not similar to the one yara-x serialize into file, so you cannot use old compiled rule with this approach. [More info here](https://github.com/VirusTotal/yara-x/issues/155)

Not sure if this could be interesting or not, if needed I can make a pull request.

ikelos:
Thanks, that looks interesting!  I'm not sure how easy to get yara-x installed as a package on various systems would be, but it might be possible to integrate it as an alternative to yara all within the same plugin?

We could use it as the opportunity to change `yara_rules` to `yara_string` as you've done, we'd need to leave in support for the older yara implementations return values.

It looks like you've already done the bulk of the work, it would just be a case of checking which module exists in priority order (yara-x, then yara-python) and using that one (with maybe an implementation adaptor class in between)?

Issue #1040, opened 2023-11-15
  `windows.filescan` reports all files to be of size `216` 


kevthehermit:
**Describe the bug**
  `windows.filescan` reports all files to be of size `216` 

**Context**
Volatility Version:  2.5.0
Operating System:  Ubuntu 22.04
Python Version:  3.10.12
Suspected Operating System:  Windows 10
Command:  `vol -f /mnt/d/Projects/dump.raw windows.filescan `

**To Reproduce**
Run the plugin on any dump for windows 10 

1. Use command `vol -f /mnt/d/Projects/dump.raw windows.filescan `

**Expected behavior**
The correct Size is reported

**Example output**

```
Volatility 3 Framework 2.5.0

Offset	Name	Size

0x910b71893a20	\Windows\System32\svchost.exe	216
0x910b73217210	\$Secure:$SDS:$DATA	216
0x910b732174f0	\$Secure:$SII:$INDEX_ALLOCATION	216
0x910b732177d0	\Windows\System32\drivers\stornvme.sys	216
0x910b73217940	\$Mft::$BITMAP	216
0x910b73217ab0	\:$I30:$INDEX_ALLOCATION	216
0x910b732181e0	\$MftMirr	216
0x910b73218630	\$LogFile	216
0x910b73218a80	\$Mft	216
0x910b73218d60	\Windows\System32\DriverStore\FileRepository\basicdisplay.inf_amd64_fc93ae411c02f280\BasicDisplay.sys	216
0x910b73218ed0	\$BitMap	216
0x910b734020a0	\$Extend\$RmMetadata\$TxfLog:$I30:$INDEX_ALLOCATION	216
0x910b73402380	\Windows\System32\drivers\dumpfve.sys	216
0x910b73402660	\$Extend\$UsnJrnl:$J:$DATA	216
0x910b73402940	\Device\HarddiskVolume3\$Extend\$RmMetadata\$TxfLog\$TxfLog	216
0x910b73402ab0	\$Directory	216
```

**Additional information**
Dump files correctly saves and extracts the correct size of file object

ikelos:
So dumpfiles walks the pages that make up the contents of the stored file, whereas filescan just reads the `Size` member of the fileobject.  This is a short and therefore will not be able to talk about files larger than 65535 bytes anyway.  Looks like to get the accurate size we'll need to do the same thing as dumpfiles to walk the pages...

[...]

PR #1197, opened 2024-07-12
Windows: remove size from filescan output as it is not the file size 

eve-mem:
Hello!

This small PR removes the 'size' column from the windows filescan plugin. It's not the file size... and is always just the object size which isn't very useful. 

@atcuno commented on the issue https://github.com/volatilityfoundation/volatility3/issues/1040#issuecomment-2183518750 suggesting that it should be removed. So this PR is that. 

Thanks for looking!

### DIFF
--- a/volatility3/framework/plugins/windows/filescan.py
+++ b/volatility3/framework/plugins/windows/filescan.py
@@ -14,6 +14,7 @@ class FileScan(interfaces.plugins.PluginInterface):
     """Scans for file objects present in a particular windows memory image."""
 
     _required_framework_version = (2, 0, 0)
+    _version = (1, 0, 1)
 
     @classmethod
     def get_requirements(cls):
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/filescan.py
```diff
@@ -13,7 +13,7 @@
 class FileScan(interfaces.plugins.PluginInterface):
     """Scans for file objects present in a particular windows memory image."""
```
eve-mem, line 15: I need to revert this. I assumed it was the plugin version. It doesn't need framework 2.0.1, 2.0.0 is fine.

ikelos (commented):
Congrats, you hit our first external/internal API change issue.  Technically any UI should introspect this, and no plugin should directly be using the generator, so the version number doesn't need a minor version bump *but* in case something weird goes on, it would be good to know this is different from the previous version.  

This plugin exposes the `scan_files` method and should therefore already have had a `_version` number.  Please can you add one in.   `(1, 0, 0)` would be fine, `(1, 0, 1)` would be more in keeping because when the plugin changes it's good to be able to differentiate them, so the PATCH number should have changed as part of this request (if the `_version` had existed)...  5:)

eve-mem:
@ikelos - I've added that version number. Thanks for explaining it all to me.

eve-mem:
@ikelos - I've added that version number. Thanks for explaining it all to me.

ikelos (approved):
Yep, looks good, thanks!

PR #1201, opened 2024-07-15
Windows: Fix vadyarascan sanity check and bad documentation

### LINKED ISSUE #1200: windows.vadyarascan broken
**Describe the bug**
Thanks for improving VAD scanning, it's much appreciated! However there's a small problem in the new code.

**Context**
Volatility Version:  Latest develop
Operating System:  Ubuntu 22.04
Python Version:  3.10
Suspected Operating System:  Windows 10
Command:  `vol.py -f dumptoscan.elf windows.vadyarascan --yara-file rules.yar`

**To Reproduce**
Run command above and get no results even if the strings are present.

**Additional information**
The problem seems to be that in [this loop](https://github.com/volatilityfoundation/volatility3/blob/ccaa473972edd01c808698dde081e796304d8f11/volatility3/framework/plugins/windows/vadyarascan.py#L69), the second tuple member is in fact the size and not the end address.

Printing `start` and `end` yields the following for me:
```
0x7ff5fdfb0000 0x2001000
0x400000 0x57000
0x80000 0xc9000
0x30000 0x1d000
0x10000 0x1000
0x20000 0x1000
0x60000 0x2000
0x50000 0x4000
0x70000 0x10000
```

ikelos:
Fixes #1200

### DIFF
--- a/volatility3/framework/plugins/windows/vadyarascan.py
+++ b/volatility3/framework/plugins/windows/vadyarascan.py
@@ -56,7 +56,7 @@ class VadYaraScan(interfaces.plugins.PluginInterface):
 
         filter_func = pslist.PsList.create_pid_filter(self.config.get("pid", None))
 
-        sanity_check = 0x1000 * 0x1000 * 0x1000
+        sanity_check = 0x2000 * 0x1000 * 0x1000
 
         for task in pslist.PsList.list_processes(
             context=self.context,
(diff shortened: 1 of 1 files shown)


ikelos:
I'm a little concerned about allowing 2Gb to be constructed in memory.  I'm still investigating building an io file-like object that can allow direct reads so that we don't have to build the image in memory to scan it...

meck-gd:
Note that `0x2000 * 0x1000 * 0x1000` is not 2 GB, this is even bigger than the previous value. 2 GB would be `2 * 1024 * 1024 * 1024`.

> I'm a little concerned about allowing 2Gb to be constructed in memory. I'm still investigating building an io file-like object that can allow direct reads so that we don't have to build the image in memory to scan it...

That would be perfect if possible.

ikelos:
Ok, but then `0x1000 * 0x1000 * 0x1000` was already greater than `2 * 1024 * 1024 * 1024`, so... was no change required?

meck-gd:
> Ok, but then `0x1000 * 0x1000 * 0x1000` was already greater than `2 * 1024 * 1024 * 1024`, so... was no change required?

`0x1000 * 0x1000 * 0x1000` is 64 GB. I have Windows memory dumps with VADs that exceed this size (with probably only a tiny fraction being backed by actual physical memory), causing Python to be OOM-killed.

ikelos:
Ok, sorry I misread this:

> Also, relating to the same feature - if I fix the bug locally, Volatility runs out of memory scanning a 4 GB memory dump.
> 
> Changing the sanity_check size to 2 GB works around that problem. Not sure if that's too restrictive for some use cases.

So what you wanted was the sanity_check lowering from 64 GB to 2GB?  I don't think VADs should get bigger than 2GB (or that if they do the data is sane), so happy to change that.  Would a 1GB single VAD ever be reasonable?

PR #1204, opened 2024-07-18
Add pedump plugin and unified API for extracting PE files to disk

atcuno:
This addresses all the feedback from the previous pedump PR as well as adds a unified API for extracting files. The previous `dump_pe` API in Volatility was specific to `LDR_DATA_TABLE_ENTRY` instances, which is how DLLs and kernel modules are tracked. The changes in this PR provide unified APIs for extracting the LDR instances plus PE files at any base address in process or kernel memory. Kernel modules are extracted in the correct session layer by default. Version number bumped for dlllist and correct requirements in place for pedump.

### DIFF
--- a/volatility3/framework/plugins/windows/dlllist.py
+++ b/volatility3/framework/plugins/windows/dlllist.py
@@ -147,7 +147,7 @@ class DllList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
 
                 file_output = "Disabled"
                 if self.config["dump"]:
-                    file_handle = pedump.PEDump.dump_ldr_entry(
+                    file_output = pedump.PEDump.dump_ldr_entry(
                         self.context,
                         pe_table_name,
                         entry,
@@ -155,9 +155,10 @@ class DllList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                         proc_layer_name,
                         prefix=f"pid.{proc_id}.",
                     )
-                    file_output = "Error outputting file"
-                    if file_handle:
-                        file_output = file_handle
+
+                    if not file_output:
+                        file_output = "Error outputting file"
+
                 try:
                     dllbase = format_hints.Hex(entry.DllBase)
                 except exceptions.InvalidAddressException:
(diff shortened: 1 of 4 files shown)


volatility3/framework/plugins/windows/dlllist.py
```diff
@@ -204,7 +147,7 @@ def _generator(self, procs):
 
                 file_output = "Disabled"
                 if self.config["dump"]:
-                    file_handle = self.dump_pe(
+                    file_handle = pedump.PEDump.dump_ldr_entry(
```
ikelos: This isn't really a file_handle anymore?  Could we get the variable name updated to something more appropriate please?

volatility3/framework/plugins/windows/modules.py
```diff
@@ -63,7 +63,7 @@ def dump_module(self, session_layers, pe_table_name, mod):
         )
         file_output = f"Cannot find a viable session layer for {mod.DllBase:#x}"
         if session_layer_name:
-            file_handle = dlllist.DllList.dump_pe(
+            file_handle = pedump.PEDump.dump_ldr_entry(
```
ikelos: Same here

volatility3/framework/plugins/windows/pedump.py
```diff
@@ -0,0 +1,270 @@
...
+            layer_name = ldr_entry.vol.layer_name
+
+        file_name = "{}{}.{:#x}.{:#x}.dmp".format(
+            prefix,
+            ntpath.basename(name),
+            ldr_entry.vol.offset,
+            ldr_entry.DllBase,
+        )
+
+        return PEDump.dump_pe(
```
ikelos: Wouldn't this just be `cls`?  Just seems a little out of place?

ikelos (commented):
Generally looks really good!  Couple of small questions, but otherwise should be good to ship!  5:)

volatility3/framework/plugins/windows/modules.py
atcuno: fixed

volatility3/framework/plugins/windows/dlllist.py
atcuno: fixed

volatility3/framework/plugins/windows/pedump.py
atcuno: nice catch. Fixed.

atcuno:
All the feedback is addressed now.

PR #1206, opened 2024-07-20
Add the orphan_kernel_threads plugin to help detect rootkits on Windo…

atcuno:
…ws systems

### DIFF
--- a/volatility3/framework/plugins/windows/orphan_kernel_threads.py
+++ b/volatility3/framework/plugins/windows/orphan_kernel_threads.py
@@ -1,0 +1,102 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import logging
+from typing import List, Generator
+
+from volatility3.framework import interfaces, symbols
+from volatility3.framework.configuration import requirements
+from volatility3.plugins.windows import thrdscan, ssdt
+
+vollog = logging.getLogger(__name__)
+
+
+class Threads(thrdscan.ThrdScan):
+    """Lists process threads"""
+
+    _required_framework_version = (2, 4, 0)
+    _version = (1, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        # Since we're calling the plugin, make sure we have the plugin's requirements
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Windows kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.PluginRequirement(
+                name="thrdscan", plugin=thrdscan.ThrdScan, version=(1, 0, 0)
+            ),
+            requirements.PluginRequirement(
+                name="ssdt", plugin=ssdt.SSDT, version=(1, 0, 0)
+            ),
+        ]
+
+    @classmethod
+    def list_orphan_threads(
+        cls,
+        context: interfaces.context.ContextInterface,
+        layer_name: str,
+        symbol_table: str,
+    ) -> Generator[interfaces.objects.ObjectInterface, None, None]:
+        """Yields thread objects of kernel threads that do not map to a module
+
+        Args:
+            kernel
+
+        Returns:
+            A generator of thread objects of orphaned threads
+        """
+        collection = ssdt.SSDT.build_module_collection(
+            context, layer_name, symbol_table
+        )
+
+        # used to filter out smeared pointers
+        if symbols.symbol_table_is_64bit(context, symbol_table):
+            kernel_start = 0xFFFFF80000000000
+        else:
+            kernel_start = 0x80000000
+
+        for thread in thrdscan.ThrdScan.scan_threads(context, layer_name, symbol_table):
+            # we don't want smeared or terminated threads
+            try:
+                proc = thread.owning_process()
+            except AttributeError:
+                continue
+
+            # we only care about kernel threads, 4 = System
+            # previous methods for determining if a thread was a kernel thread
+            # such as bit fields and flags are not stable in Win10+
+            # so we check if the thread is from the kernel itself or one its child
+            # kernel processes (MemCompression, Regsitry, ...)
+            if proc.UniqueProcessId != 4 and proc.InheritedFromUniqueProcessId != 4:
+                continue
+
+            if thread.StartAddress < kernel_start:
+                continue
+
+            module_symbols = list(
+                collection.get_module_symbols_by_absolute_location(thread.StartAddress)
+            )
+
+            # alert on threads that do not map to a module
+            if not module_symbols:
+                yield thread
+
+    def _generator(self):
+        """
+        A simple wrapper around list_orphan_threads that reuses run from thrdscan
+        """
+        kernel = self.context.modules[self.config["kernel"]]
+
+        for thread in self.list_orphan_threads(
+            context=self.context,
+            layer_name=kernel.layer_name,
+            symbol_table=kernel.symbol_table_name,
+        ):
+            info = self.gather_thread_info(thread)
+            if info:
+                yield (0, info)


volatility3/framework/plugins/windows/orphan_kernel_threads.py
```diff
@@ -0,0 +1,102 @@
...
+        Returns:
+            A generator of thread objects of orphaned threads
+        """
+        collection = ssdt.SSDT.build_module_collection(
+            context, layer_name, symbol_table
+        )
+
+        # used to filter out smeared pointers
+        if symbols.symbol_table_is_64bit(context, symbol_table):
+            kernel_start = 0xFFFFF80000000000
```
ikelos, line 59: Not a big fan of hard coded constants.  If this is always the kernel_start, should it not belong in a constants file?

volatility3/framework/plugins/windows/orphan_kernel_threads.py
```diff
@@ -0,0 +1,102 @@
...
+from typing import List, Generator
+
+from volatility3.framework import interfaces, symbols
+from volatility3.framework.configuration import requirements
+from volatility3.plugins.windows import thrdscan, ssdt
+
+vollog = logging.getLogger(__name__)
+
+
+class Threads(thrdscan.ThrdScan):
```
ikelos, line 15: Again with the inheritance...

volatility3/framework/plugins/windows/orphan_kernel_threads.py
```diff
@@ -0,0 +1,102 @@
...
+
+            module_symbols = list(
+                collection.get_module_symbols_by_absolute_location(thread.StartAddress)
+            )
+
+            # alert on threads that do not map to a module
+            if not module_symbols:
+                yield thread
+
+    def _generator(self):
```
ikelos, line 89: And again with overriding the generator.

This means the run and generator are separated across plugins and could lead to breakage.  This should be its own plugin and `gather_thread_info` should be class method.  Then the run method should be duplicated.

ikelos (changes requested):
Oh man, this looks like it was gonna be easy, I was really close to clicking merge, and then I spotted inheritance, again.  We *cannot* split `run` and `generator`, I *am* ok with duplicating column descriptions to ensure that the plugin output doesn't suddenly change unintentionally and I do want us to find a way to do this sensibly.

Given `gather_thread_info` seems to directly generate the output line, there should probably also be a classmethod in the Threads plugin that generates/defines the TreeGrid.  Then run will essentially be reduced to a single line, and we both get what we're after, maximum reuse minimum breakage...

volatility3/framework/plugins/windows/orphan_kernel_threads.py
atcuno: @ikelos if `run` changes in ThrdScan then the version would bump, which would then make this plugin break, which as discussed on the call is what we *want* to happen. The whole reason for the version requirements is to force plugins to break if they are re-using functionality that changes. The lack of version requirements is what led to so many bugs and inconsistencies in output in Vol2. The way I have it with using the requirements correctly also makes Vol3 consistent for years to come as if the run() changes in ThrdScan, then every plugin that re-uses it will have the same output instead of something like threads has one set of columns, thrdscan has another set of columns, then orphan thread has a different set. Removing copy/paste code and having as much shared code as possible is the only way we will avoid the copy/paste bugs, breakage, and inconsistency of Volatility 2. I thought we had settled this on the call so I wrote the plugins to match. All version requirements are set in the plugins so that the breakage is guaranteed in inheriting plugins.

atcuno: Also, for reference, you approved the same code pattern (inheriting run) three weeks ago in threads when I did the original combination: https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/threads.py#L63

ikelos: As mentioned in the other PR, the version number only changes for internal API changes because it's assumed that the external interface can be introspected, meaning changes to `_generator` or the `TreeGrid` can be made *without* necessitating a version bump, and certainly without a major version bump (which is what would be required to get this plugin to break on a requirement if a change were made).  Without forcing the external API into our versioning system, there is no way for the underlying plugin to know that it may affect children if its external interface (run/generator) changes and it'd be very easy for a reviewer to miss the change and let it slide.

As noted, I raised issues at the time concerning introspection but I relented in the interests of not putting too many blocks in place.  I've reviewed this carefully and have given my reasons for why it's a bad design pattern.  It's a bad design pattern in the Threads plugin too and I wish I'd investigated it further to find a suitable good design pattern (such as the method svcscan/svcdiff used, which is far better).  It really shouldn't add too much extra code to expose methods that are expected to be shared.  `run`/`generate` changes cannot be split across plugins.  I agreed with you that there may be some benefits to inheritance in particular cases, but that wasn't an agreement that it's entirely harmless in all circumstances and can now litter the tree.  I don't mind some inheritance, and you'll see on other PRs I've checked today I haven't complained and agreed with the design where they won't run into this issue.

You can do the inheritance (as indicated by the svcscan/svcdiff plugins), but alter the parent plugin's generator to iterate through a "make this kind of thing" method with a well defined signature, changes to which will be obvious to affect others (and therefore cause a major version bump if they're not compatible).

atcuno: That makes sense. I will make threads/thrdscan/orphan threads work with the enumeration_method flow like services and modules.

atcuno: @ikelos I need some guidance here as when I went to change this over, I ran into a roadblock, and when I went to see how related plugins do it I found all sorts of inconsistencies. I am up for writing the code to switch the affected plugins to all be consistent, but wanted to run it by you first.

The problem/issue:

There are attributes of processes (handles, VADs, environment variables, threads, etc.), that plugins specific to the attribute will want to enumerate all of for every process, but then we also have situations where a plugin just wants to enumerate the attribute of one specific process. 

Sticking with threads, right now threads.py exposes list_threads, which accepts one process instance as a parameter and returns the threads of that one process. This is currently used by suspicious_threads and will be used by others in the future:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/suspicious_threads.py#L167

From my view, having this 'one process -> one set of attributes' API is essential given all the plugins that will want to operate looking for malicious activity against one attribute at a time.

Trying to fit this into `enumeration_method` was tricky though given that the base threads.py plugin wants to enumerate for all processes. My first thought was to add a classmethod such as `list_threads_for_all_processes` or similarly named, that walked the process list and called `list_threads`. This would be what `_enumeration_method` is set to.

I went to check how other, related plugins handle this situation though and everything was inconsistent, so I got stuck.

For example, vadinfo exposes `list_vads`, which calls an extension to get the vads, and does it one process at a time:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/vadinfo.py#L111-L127

But then malfind, which should be a consumer of `list_vads` just calls `traverse` on its own instead of `list_vads` and walks the process list directly:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/malfind.py

Handles is even stranger as it doesn't even expose a class method for enumerating handles:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/handles.py#L402

And then envars is probably the worst as it just gathers the data directly inside of `_generator` using an extension, which obviously extensions can't be versioned and gives nothing for other plugins to use:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/envars.py#L208

So is my idea of keeping `list_threads` how it is and then a `list_threads_for_all_processes` the most ideal way? Only plugins with inheritance/`enumeration_method` will need the "all processes" APIs, which I don't think the other process ones will ever need, so they will keep the singular form like `list_vads`. If that way makes sense then I can get the threads plugins all updated. 

Also, the Black Hat class starts in exactly 13 days, and I have alot of other stuff to do before then too to prep, so I can't switch malfind, envars, etc. all over before Black Hat, but definitely could put on my list to do before the October event.

ikelos: Hiya, unfortunately I think that's an upshot of our development process, lots of people submit their plugins which they want in the codebase, and if the code works there's really only so much tweaking I can ask them to do.  We haven't documented exactly how these things are supposed to be done and previously we probably didn't have enough to make getting them all the same worthwhile.

I agree it would be best to get them all consistent and get the design pattern we intend to use in place for all of them, so that people that come copying plugins copy the right ones, and don't use one slightly less good one as a claim to get theirs into the codebase without a little extra work.  5:)

We also need to consider plugin dependency complexity.  If the actual process for completing the plugin diverts to go through lots of little plugins to do something trivially, then it might be too much.  `List_vads` is tricky because it's really just a wrapper around traverse, but it throws in filtration.  Applying filtration to return a single item from a collection might be inefficient though, so trying to specify a one size fits all will be really tricky.

It does sounds like the right step is to have:

Run through collection of collections (such as processes) -> run through a single item -> do the thing that's actually wanted to that single item.

Then plugins that want to make use of all the other machinery, can do so but slightly less efficiently, or they can implement their own block in that step and call back in further down.  I don't think we should prevent that, because it may be way quicker for a plugin to get to a single process (particularly if it just instantiates one at an offset, rather than running through a list to find it).

So yeah, I think the most generic/reusable would be to keep `list_threads` acting on a single process, and then make an outer function that calls it once for each pid based on a generic filter callable, if that works for you?

No problem about blackhat, I'm just trying not to be the hold up.  We can make a ticket for it and assign you it, my only worry is that there won't be the same level of enthusiasm to complete them that the time pressure of Blackhat seems to be having...  5;)

atcuno: October is actually the most pressure :)

PR #1207, opened 2024-07-21
Add in threads that only provides an implmentation method

ikelos:
Reworks threads from using an altered generator to something more generic which helps make the dependency easier to see and reason about.

### DIFF
--- a/volatility3/framework/plugins/windows/threads.py
+++ b/volatility3/framework/plugins/windows/threads.py
@@ -38,7 +38,7 @@ class Threads(thrdscan.ThrdScan):
                 optional=True,
             ),
             requirements.PluginRequirement(
-                name="thrdscan", plugin=thrdscan.ThrdScan, version=(1, 0, 0)
+                name="thrdscan", plugin=thrdscan.ThrdScan, version=(1, 1, 0)
             ),
         ]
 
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/windows/threads.py
```diff
@@ -18,6 +18,9 @@ class Threads(thrdscan.ThrdScan):
     _required_framework_version = (2, 4, 0)
     _version = (1, 0, 0)
 
+    def __init__(self):
+        self.implementation = self.list_process_threads
```
atcuno, line 22: missing call to super(), which should be before the self.implementation line so it doesn't get overwritten

atcuno:
This looks nice. I left one comment just for the missing super(), also would need to send a parameter to generator now: https://github.com/volatilityfoundation/volatility3/blob/5e96327cb0851fe75d1c7a4cb9ef27641bf119f7/volatility3/framework/plugins/windows/thrdscan.py#L107

volatility3/framework/plugins/windows/threads.py
ikelos: Good catch, thanks!

ikelos:
Yep, very good catch, thanks!

PR #1208, opened 2024-07-22
yara-x support for yarascan

dadokkio:
This pull adds support for [yara-x ](https://virustotal.github.io/yara-x/) library.
If the library is not available it still uses yara-python library as before.

Changes:
- yara-file parameters is now yara_string
- yara_compiled_file is used also for yara-x even if in this case it's a [serialized file](https://github.com/VirusTotal/yara-x/blob/d064e427ad4f3a0f01cf4778d03d3b253727516f/py/src/lib.rs#L466) and not the old compiled 

Question:
- Not sure how to manage requirements, at the moment I've added both deps in them.
- It could be useful to have an env variable to override yara-x default?

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -3,6 +3,7 @@ sphinx>=4.0.0,<7
 sphinx_autodoc_typehints>=1.4.0
 sphinx-rtd-theme>=0.4.3
 
+yara-python
 yara-x
 pycryptodome
 pefile
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -3,6 +3,6 @@ sphinx>=4.0.0,<7
 sphinx_autodoc_typehints>=1.4.0
 sphinx-rtd-theme>=0.4.3
 
-yara-python
+yara-x
 pycryptodome
 pefile
--- a/test/requirements-testing.txt
+++ b/test/requirements-testing.txt
@@ -5,6 +5,7 @@ pefile>=2017.8.1 #foo
 # If certain packages are not necessary, place a comment (#) at the start of the line.
 
 # This is required for the yara plugins
+yara-python>=3.8.0
 yara-x>=0.5.0
 
 pytest>=7.0.0
--- a/test/requirements-testing.txt
+++ b/test/requirements-testing.txt
@@ -5,6 +5,6 @@ pefile>=2017.8.1 #foo
 # If certain packages are not necessary, place a comment (#) at the start of the line.
 
 # This is required for the yara plugins
-yara-python>=3.8.0
+yara-x>=0.5.0
 
 pytest>=7.0.0
--- a/volatility3/framework/plugins/windows/vadyarascan.py
+++ b/volatility3/framework/plugins/windows/vadyarascan.py
@@ -33,7 +33,7 @@ class VadYaraScan(interfaces.plugins.PluginInterface):
                 name="pslist", plugin=pslist.PsList, version=(2, 0, 0)
             ),
             requirements.PluginRequirement(
-                name="yarascan", plugin=yarascan.YaraScan, version=(1, 3, 0)
+                name="yarascan", plugin=yarascan.YaraScan, version=(2, 0, 0)
             ),
             requirements.ListRequirement(
                 name="pid",
@@ -73,26 +73,43 @@ class VadYaraScan(interfaces.plugins.PluginInterface):
                     )
                     continue
 
-                for match in rules.match(data=layer.read(start, size, True)):
-                    if yarascan.YaraScan.yara_returns_instances():
-                        for match_string in match.strings:
-                            for instance in match_string.instances:
+                if not yarascan.YaraScan._yara_x:
+                    for match in rules.match(data=layer.read(start, size, True)):
+                        if yarascan.YaraScan.yara_returns_instances():
+                            for match_string in match.strings:
+                                for instance in match_string.instances:
+                                    yield 0, (
+                                        format_hints.Hex(instance.offset + start),
+                                        task.UniqueProcessId,
+                                        match.rule,
+                                        match_string.identifier,
+                                        instance.matched_data,
+                                    )
+                        else:
+                            for offset, name, value in match.strings:
+                                yield 0, (
+                                    format_hints.Hex(offset + start),
+                                    task.UniqueProcessId,
+                                    match.rule,
+                                    name,
+                                    value,
+                                )
+                else:
+                    data = layer.read(start, size, True)
+                    results = rules.scan(data)
+                    for match in results.matching_rules:
+                        for match_string in match.patterns:
+                            for instance in match_string.matches:
                                 yield 0, (
                                     format_hints.Hex(instance.offset + start),
                                     task.UniqueProcessId,
-                                    match.rule,
+                                    f"{match.namespace}.{match.identifier}",
                                     match_string.identifier,
-                                    instance.matched_data,
+                                    data[
+                                        instance.offset : instance.offset
+                                        + instance.length
+                                    ],
                                 )
-                    else:
-                        for offset, name, value in match.strings:
-                            yield 0, (
-                                format_hints.Hex(offset + start),
-                                task.UniqueProcessId,
-                                match.rule,
-                                name,
-                                value,
-                            )
 
     @staticmethod
     def get_vad_maps(
--- a/volatility3/framework/plugins/windows/vadyarascan.py
+++ b/volatility3/framework/plugins/windows/vadyarascan.py
@@ -73,8 +73,9 @@ class VadYaraScan(interfaces.plugins.PluginInterface):
                     )
                     continue
 
+                data = layer.read(start, size, True)
                 if not yarascan.YaraScan._yara_x:
-                    for match in rules.match(data=layer.read(start, size, True)):
+                    for match in rules.match(data=data):
                         if yarascan.YaraScan.yara_returns_instances():
                             for match_string in match.strings:
                                 for instance in match_string.instances:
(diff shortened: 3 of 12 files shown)


volatility3/framework/plugins/yarascan.py
```diff
@@ -56,16 +64,71 @@
...
 
+    @staticmethod
+    def get_rule(rule):
+        return yara.compile(
+            sources={"n": f"rule r1 {{strings: $a = {rule} condition: $a}}"}
+        )
+
+    @staticmethod
+    def from_compiled_file(filepath):
+        return yara.load(file=resources.ResourceAccessor().open(filepath, "rb"))
```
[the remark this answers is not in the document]
ikelos: Please use:

```
with resources.ResourceAccessor().open(filepath, "rb") as fp:
  yara.load(fp)
```

This will ensure the file is closed after it's no longer needed.

volatility3/framework/plugins/yarascan.py
```diff
@@ -99,13 +162,15 @@ def get_yarascan_option_requirements(
...
+                optional=True,
             ),
             requirements.URIRequirement(
-                name="yara_file", description="Yara rules (as a file)", optional=True
+                name="yara_file",
+                description="Yara rules (as a file)",
+                optional=True,
             ),
-            # This additional requirement is to follow suit with upstream, who feel that compiled rules could potentially be used to execute malicious code
-            # As such, there's a separate option to run compiled files, as happened with yara-3.9 and later
```
ikelos, line 108: Please leave these comments in place, they help us understand why there's a separate option for files.

volatility3/framework/plugins/yarascan.py
```diff
@@ -56,16 +64,71 @@ def __call__(
...
+YaraScanner = YaraXScanner if USE_YARA_X else YaraPythonScanner
+
 
 class YaraScan(plugins.PluginInterface):
     """Scans kernel memory using yara rules (string or file)."""
 
     _required_framework_version = (2, 0, 0)
     _version = (1, 3, 0)
 
-    # TODO: When the major version is bumped, take the opportunity to rename the yara_rules config to yara_string
```
ikelos, line 66: That's fine, but you then haven't bumped the plugin.  It should really become `(2, 0, 0)` for a significant bump and dependent plugins would then also need their requirements bumping.  This is to ensure that we've tested that the plugins can all work with each other...

volatility3/framework/plugins/yarascan.py
```diff
@@ -119,40 +184,26 @@ def get_yarascan_option_requirements(
             ),
         ]
 
-    @classmethod
-    def yara_returns_instances(cls) -> bool:
```
ikelos, line 123: This is a classmethod that may be used in other plugins (which is why it was pulled out as a separate function in the first place), please avoid changing things without good reason.

ikelos: Removing this, in itself, would be reason to bump the version number to `(2, 0, 0)` let alone adding yara-x support.

volatility3/framework/plugins/yarascan.py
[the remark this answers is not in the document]
ikelos: Please do this for all four of the code segments identified by the CodeQL scanning...

test/requirements-testing.txt
```diff
@@ -6,5 +6,6 @@ pefile>=2017.8.1 #foo
 
 # This is required for the yara plugins
 yara-python>=3.8.0
+yara-x>=0.5.0
```
ikelos, line 9: Is there a way to say either or?  A user shouldn't need both at once.

volatility3/framework/plugins/yarascan.py
```diff
@@ -13,19 +13,30 @@
...
+        import yara
+
+        if tuple(int(x) for x in yara.__version__.split(".")) < (3, 8):
+            raise ImportError
+
+        vollog.info("Using yara-python module")
+
+    except ImportError:
+        vollog.info(
+            "Python Yara (>3.8.0) module not found, plugin (and dependent plugins) not available"
```
ikelos: This should probably also offer that yara-x can be used as well (so "Neither yara-x not python yara (3.8.0) was found...")...

doc/requirements.txt
```diff
@@ -4,5 +4,6 @@ sphinx_autodoc_typehints>=1.4.0
 sphinx-rtd-theme>=0.4.3
 
 yara-python
+yara-x
```
ikelos, line 7: I think for the document requirements this is ok.

ikelos (changes requested):
Thanks very much, but it looks like it only considers `yarascan` and not the `windows.vadyarascan` or `linux.vmayarascan` both of which rely on this plugin (albeit it much less in their latest versions).  The version number also requires bumping for several reasons.  Otherwise this looks quite nice and I'm keen to get it in the tree.  5:)

dadokkio:
Thanks for the feedback, all of them should be fixed.
For requirements I don't think you can have both of them in a OR way with a single requirements, so I don't have any suggestion on this. 
From the pipeline I see also that is failing with python 3.7 (yara-x requires 3.8) , not sure if also this can be managed somehow.

ikelos:
Hiya, thanks for fixing those up, to bump the minimum requirement of volatility to python 3.8, you'd need to alter the four following lines.  If you include the changes in your pull request, the pipeline for this request should run against the newer version after this pull.  I think we could maybe make yara and yara-x "extras" meaning users could choose which of them had which required dependencies, but again I'll need to ask the team before we can get this pushed through...

https://github.com/volatilityfoundation/volatility3/blob/05aa81b63cc95289e4989915187a58df5dfcd08c/volatility3/framework/__init__.py#L10

https://github.com/volatilityfoundation/volatility3/blob/05aa81b63cc95289e4989915187a58df5dfcd08c/.github/workflows/test.yaml#L9

https://github.com/volatilityfoundation/volatility3/blob/05aa81b63cc95289e4989915187a58df5dfcd08c/.github/workflows/build-pypi.yml#L21

https://github.com/volatilityfoundation/volatility3/blob/05aa81b63cc95289e4989915187a58df5dfcd08c/.github/workflows/install.yml#L11

ikelos:
Also having bumped the version of yarascan correctly, both `vmayarascan` and `vadyarascan` will need their plugin requirement versions updating to make sure their dependencies are met.  In vmayarascan this also requires the YaraScanner requirement being updated (and therefore choosing whether to use python or yara-x).  It might be worth building a single unified YaraScanner class that can then decide which implementation to use based on which one was available on the system at the time?

dadokkio:
it should be the line:

`YaraScanner = YaraXScanner if USE_YARA_X else YaraPythonScanner`

I ended using two different classes in order to avoid the check in all the functions, but let me know if this could be improved.
Since the __init__ and the generated data are the same other plugins (vadyarascan, vmayarascan) can use it without noticing. I've tested it with vadyarascan plugin adding and removing yara-x library without any issue.

For the _version, Should I bump YaraXScanner and YaraPythonScanner version (to 2.1.0?) and change all depending plugins (vmayarascan, mftscan and vadyarascan) requirements accordingly?

eg:
image

ikelos:
Ah, you're quite right, sorry I missed that.  It will somewhat duplicate the versioning, rather than just having inside the scan function contain an if that decides which of the different lines to use?  Otherwise, yes, you'll need to update both the scanner versions and all three of the plugins that use it.

dadokkio:
Ok, I created a base class to keep common info (like version) shared between yaraxscanner e yarapythonscanner.
I've also bumped plugin requirements and done some other little fix that I forgot on them.

I didn't tested vmayarascan cause I really don't know which parameters must be pass to use it.

volatility3/framework/plugins/windows/vadyarascan.py
```diff
@@ -73,26 +73,43 @@ def _generator(self):
...
                                 yield 0, (
-                                    format_hints.Hex(instance.offset + start),
+                                    format_hints.Hex(offset + start),
                                     task.UniqueProcessId,
                                     match.rule,
+                                    name,
+                                    value,
+                                )
+                else:
+                    data = layer.read(start, size, True)
```
ikelos: This is shared between both implementations, as are many of the return values.  It would be nice if we could shrink the differences to just the loop (but it's not critical so if can't be done easily, don't worry about it).

volatility3/framework/plugins/yarascan.py
```diff
@@ -56,15 +69,62 @@ def __call__(
                 for offset, name, value in match.strings:
                     yield (offset + data_offset, match.rule, name, value)
 
+    @staticmethod
```
ikelos, line 82: All three of these static methods seem to be the same across the two classes, shouldn't they just be part of the base class?

ikelos (commented):
Thanks for tidying this up, but I'm still a little confused why there's two scanner classes, rather than just one with a toggle?  That means that the other plugins automatically get the benefit of yara-x without changing anything, and the implementation is decided based on which yara could be imported?  I wasn't expecting the other plugins to actually be changed, I though just the YaraScanner class would decide internally how it was going to do the scanning? (and so would only affect the vmayarascan, since vadyarascan calls yara directly).  I'd leave vadyarascan for another pull request some time.  This PR is about getting vadyarascan working...

ikelos:
It seems as though the PIP package requires python-3.8 rather than 3.7.3.  Please can you include the bump to 3.8 (following this commit as a guide) so we can see what it would look like and if anything breaks.

https://github.com/volatilityfoundation/volatility3/commit/61e3e894eea28a8e0368c43f0fecb1b5031e0d2b

Also there are some outstanding comments which are awaiting a response?

ikelos:
Just checking in...  We can get this merged once that change is made, it's currently what's making the tests break, so if you can get that fixed we can pull this into the main tree.  5:)

ikelos (approved):
Thanks, this looks good now.  Let's get it in the main tree...  5:)

PR #1209, opened 2024-07-22
Windows: Adds Amcache plugin

dgmcdona:
This adds a plugin for parsing information about installed executables and applications from the `Amcache.hve` registry hive. 

The original volatility plugin supported parsing of `Amcache.hve` hives on Windows 8 systems; this PR includes that functionality, as well as new support for Windows 10+ systems. 

On both Windows 8 and Windows 10 systems, there are analogous 'program' and 'file' type entries. These can be found under the following registry keys:
- Windows 8: `Programs`/`File`
- Windows 10: `InventoryApplication`/`InventoryApplicationFile`

In both cases, `File` entries can be correlated with `Program` entries via a `ProgramId` string. Because a single `Program` can have multiple `files` associated with its installation, this plugin renders them in a tree format, with a `Program` entry being the parent of the `File` entries when the relationship can be determined. Otherwise, `File` entries are treated as standalone and rendered at the top level of the tree. 

It makes sense that there would be standalone `File` entries, since the scheduled task that populates the Amcache periodically scans a fixed set of folders for any PE files, and adds them to the Amcache when found.

On Windows 10 systems, there is a third registry key (`Root\\InventoryDriverBinary`) which contains records of installed drivers. These entries are parsed and rendered at the top level as well.

Where possible, `Program` and `File` entries share fields, and in cases where a field may only apply to one or the other, a `renderers.NotApplicableValue` is used to populate that field.

### DIFF
--- a/volatility3/framework/plugins/windows/amcache.py
+++ b/volatility3/framework/plugins/windows/amcache.py
@@ -357,13 +357,7 @@ class Amcache(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
             install_time = _get_datetime_utc_epoch_value(
                 values, val_enum.InstallTime.value
             )
-            _version = _get_string_value(values, val_enum.Version.value)
-
-            if isinstance(_version, str):
-                if isinstance(product, str):
-                    product = f"{product} {_version}"
-                else:
-                    product = f"UNKNOWN {_version}"
+            version = _get_string_value(values, val_enum.Version.value)
 
             yield program_id, _AmcacheEntry(
                 AmcacheEntryType.Program.name,
@@ -373,6 +367,7 @@ class Amcache(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                 ),
                 install_time=install_time,
                 product_name=product,
+                product_version=version,
             )
 
     @classmethod
@@ -494,11 +489,11 @@ class Amcache(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
 
             # Depending on the Windows version, the key name will be either the name
             # of the driver, or its SHA1 hash.
-            if "/" in binary_key.get_name():
-                driver_name = binary_key.get_name()
+            if "/" in str(binary_key.get_name()):
+                driver_name = str(binary_key.get_name())
                 sha1_hash = _get_string_value(values, val_enum.DriverId.name)
             else:
-                sha1_hash = binary_key.get_name()
+                sha1_hash = str(binary_key.get_name())
                 driver_name = _get_string_value(values, val_enum.DriverName.name)
 
             if isinstance(sha1_hash, str):
@@ -646,6 +641,7 @@ class Amcache(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                 ("SHA1", str),
                 ("Service", str),
                 ("ProductName", str),
+                ("ProductVersion", str),
             ],
             (
                 (indent, dataclasses.astuple(entry))
--- a/volatility3/framework/plugins/windows/amcache.py
+++ b/volatility3/framework/plugins/windows/amcache.py
@@ -1,0 +1,646 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import dataclasses
+import datetime
+import enum
+import logging
+import itertools
+from typing import Dict, Iterable, Iterator, List, Optional, Tuple, Union
+
+from volatility3.framework import interfaces, renderers
+from volatility3.framework.configuration import requirements
+from volatility3.framework.layers import registry
+from volatility3.framework.renderers import conversion
+from volatility3.framework.symbols.windows.extensions import registry as reg_extensions
+from volatility3.plugins import timeliner
+from volatility3.plugins.windows.registry import hivelist
+
+vollog = logging.getLogger(__name__)
+
+#######################################################################
+# More information about the following enums can be found in the report
+# 'Analysis of the AmCache` by Blanche Lagny, 2019
+#######################################################################
+
+
+class Win8FileValName(enum.Enum):
+    """
+    An enumeration that creates a helpful mapping of opaque Windows 8 Amcache
+    'File' subkey value names to their human-readable equivalent.
+    """
+
+    ProgramID = "100"
+    SHA1Hash = "101"
+    Product = "0"
+    Company = "1"
+    Size = "6"
+    SizeOfImage = "7"
+    PEHeaderChecksum = "9"
+    LastModTime = "11"  # REG_QWORD FILETIME
+    CreateTime = "12"  # REG_QWORD FILETIME
+    Path = "15"
+    LastModTime2 = "17"  # REG_QWORD FILETIME
+    Version = "d"
+    CompileTime = "f"  # REG_QWORD UNIX EPOCH
+
+
+class Win8ProgramValName(enum.Enum):
+    """
+    An enumeration that creates a helpful mapping of opaque Windows 8 Amcache
+    'Program' subkey value names to their human-readable equivalent.
+    """
+
+    Product = "0"
+    Version = "1"
+    Publisher = "2"
+    InstallTime = "a"
+    MSIProductCode = "11"
+    MSIPackageCode = "12"
+    ProductCode = "f"
+    PackageCode = "10"
+
+
+class Win10InvAppFileValName(enum.Enum):
+    """
+    An enumeration containing the most useful Windows 10 Amcache
+    'InventoryApplicationFile' subkey value names.
+    """
+
+    FileId = "FileId"
+    LinkDate = "LinkDate"
+    LowerCaseLongPath = "LowerCaseLongPath"
+    ProductName = "ProductName"
+    ProductVersion = "ProductVersion"
+    ProgramID = "ProgramId"
+    Publisher = "Publisher"
+
+
+class Win10InvAppValName(enum.Enum):
+    """
+    An enumeration containing the most useful Windows 10 Amcache
+    'InventoryApplication' subkey value names.
+    """
+
+    InstallDate = "InstallDate"
+    Name = "Name"
+    Publisher = "Publisher"
+    RootDirPath = "RootDirPath"
+    Version = "Version"
+
+
+class Win10DriverBinaryValName(enum.Enum):
+    """
+    An enumeration containing the most useful Windows 10 Amcache
+    'InventoryDriverBinary' subkey value names.
+    """
+
+    DriverId = "DriverId"
+    DriverName = "DriverName"
+    DriverCompany = "DriverCompany"
+    Product = "Product"
+    Service = "Service"
+    DriverTimeStamp = "DriverTimeStamp"
+
+
+class AmcacheEntryType(enum.IntEnum):
+    Driver = 1
+    Program = 2
+    File = 3
+
+
+NullableString = Union[str, None, interfaces.renderers.BaseAbsentValue]
+NullableDatetime = Union[datetime.datetime, None, interfaces.renderers.BaseAbsentValue]
+
+
+@dataclasses.dataclass
+class _AmcacheEntry:
+    """
+    A class containing all information about an entry from the Amcache registry hive.
+    Because all values could potentially be paged out of memory or malformed, they are all
+    a union between their expected value and `interfaces.renderers.BaseAbsentValue`.
+    """
+
+    entry_type: str
+    path: NullableString = renderers.NotApplicableValue()
+    company: NullableString = renderers.NotApplicableValue()
+    last_modify_time: NullableDatetime = renderers.NotApplicableValue()
+    last_modify_time_2: NullableDatetime = renderers.NotApplicableValue()
+    install_time: NullableDatetime = renderers.NotApplicableValue()
+    compile_time: NullableDatetime = renderers.NotApplicableValue()
+    sha1_hash: NullableString = renderers.NotApplicableValue()
+    service: NullableString = renderers.NotApplicableValue()
+    product_name: NullableString = renderers.NotApplicableValue()
+
+
+def _entry_sort_key(entry_tuple: Tuple[NullableString, _AmcacheEntry]) -> str:
+    """Sorts entries by program ID. This is broken out as a function here
+    to ensure consistency in sorting between the `group_by` and `sorted` function
+    invocations.
+    """
+    program_id, _ = entry_tuple
+    key = program_id if isinstance(program_id, str) else ""
+    return key
+
+
+def _get_string_value(
+    values: Dict[str, reg_extensions.CM_KEY_VALUE], name: str
+) -> NullableString:
+    try:
+        value = values[name]
+    except KeyError:
+        return renderers.NotAvailableValue()
+
+    data = value.decode_data()
+    if not isinstance(data, bytes):
+        return renderers.UnparsableValue()
+
+    return data.decode("utf-16le", errors="replace").rstrip("\u0000")
+
+
+def _get_datetime_filetime_value(
+    values: Dict[str, reg_extensions.CM_KEY_VALUE], name: str
+) -> NullableDatetime:
+    try:
+        value = values[name]
+    except KeyError:
+        return renderers.NotAvailableValue()
+
+    data = value.decode_data()
+    if not isinstance(data, int):
+        return renderers.UnparsableValue()
+
+    return conversion.wintime_to_datetime(data)
+
+
+def _get_datetime_utc_epoch_value(
+    values: Dict[str, reg_extensions.CM_KEY_VALUE], name: str
+) -> NullableDatetime:
+    try:
+        value = values[name]
+    except KeyError:
+        return renderers.NotAvailableValue()
+
+    data = value.decode_data()
+    if not isinstance(data, (int, float)):
+        return renderers.UnparsableValue()
+
+    try:
+        return datetime.datetime.fromtimestamp(float(data), datetime.timezone.utc)
+    except (ValueError, OverflowError, OSError):
+        return renderers.UnparsableValue()
+
+
+def _get_datetime_str_value(
+    values: Dict[str, reg_extensions.CM_KEY_VALUE], name: str
+) -> NullableDatetime:
+    try:
+        value = values[name]
+    except KeyError:
+        return renderers.NotAvailableValue()
+
+    data = value.decode_data()
+    if not isinstance(data, int):
+        return renderers.UnparsableValue()
+
+    if isinstance(data, str):
+        try:
+            return datetime.datetime.strptime(data, "%m/%d/%Y %H:%M:%S")
+        except ValueError:
+            return renderers.UnparsableValue()
+    else:
+        return renderers.UnparsableValue()
+
+
+class Amcache(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
+    """Scans for windows services."""
+
+    _required_framework_version = (2, 0, 0)
+    _version = (1, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        # Since we're calling the plugin, make sure we have the plugin's requirements
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Windows kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.PluginRequirement(
+                name="hivelist", plugin=hivelist.HiveList, version=(1, 0, 0)
+            ),
+        ]
+
+    def generate_timeline(
+        self,
+    ) -> Iterator[Tuple[str, timeliner.TimeLinerType, datetime.datetime]]:
+        for _, entry in self._generator():
+            if isinstance(entry.last_modify_time, datetime.datetime):
+                yield f"Amcache: {entry.entry_type} {entry.path} registry key modified", timeliner.TimeLinerType.MODIFIED, entry.last_modify_time
+            if isinstance(entry.last_modify_time_2, datetime.datetime):
+                yield f"Amcache: {entry.entry_type} {entry.path} STANDARD_INFORMATION create time", timeliner.TimeLinerType.CREATED, entry.last_modify_time_2
+            if isinstance(entry.install_time, datetime.datetime):
+                yield f"Amcache: {entry.entry_type} {entry.path} installed", timeliner.TimeLinerType.CREATED, entry.install_time
+            if isinstance(entry.compile_time, datetime.datetime):
+                yield f"Amcache: {entry.entry_type} {entry.path} compiled (PE metadata)", timeliner.TimeLinerType.MODIFIED, entry.compile_time
+
+    @classmethod
+    def get_amcache_hive(
+        cls,
+        context: interfaces.context.ContextInterface,
+        config_path: str,
+        kernel: interfaces.context.ModuleInterface,
+    ) -> Optional[registry.RegistryHive]:
+        """Retrieves the `Amcache.hve` registry hive from the kernel module, if it can be located."""
+        return next(
+            hivelist.HiveList.list_hives(
+                context=context,
+                base_config_path=interfaces.configuration.path_join(
+                    config_path, "hivelist"
+                ),
+                layer_name=kernel.layer_name,
+                symbol_table=kernel.symbol_table_name,
+                filter_string="amcache",
+            ),
+            None,
+        )
+
+    @classmethod
+    def parse_file_key(
+        cls, file_key: reg_extensions.CM_KEY_NODE
+    ) -> Iterator[Tuple[NullableString, _AmcacheEntry]]:
+        """Parses File entries from the Windows 8 `Root\\File` key.
+
+        :param programs_key: The `Root\\File` registry key.
+
+        :return: An iterator of tuples, where the first member is the program ID string for
+        correlating `Root\\Program` entries, and the second member is the `AmcacheEntry`.
+        """
+
+        for file_entry_key in itertools.chain(
+            *(key.get_subkeys() for key in file_key.get_subkeys())
+        ):
+            vollog.debug(f"Checking Win8 File key {file_entry_key.get_name()}")
+            values = {
+                str(value.get_name()): value
+                for value in file_entry_key.get_values()
+                if value.get_name() in [key.value for key in Win8FileValName]
+            }
+
+            program_id = _get_string_value(values, Win8FileValName.ProgramID.value)
+            path = _get_string_value(values, Win8FileValName.Path.value)
+            company = _get_string_value(values, Win8FileValName.Company.value)
+            last_mod_time = _get_datetime_filetime_value(
+                values, Win8FileValName.LastModTime.value
+            )
+            last_mod_time_2 = _get_datetime_filetime_value(
+                values, Win8FileValName.LastModTime2.value
+            )
+            install_time = _get_datetime_filetime_value(
+                values, Win8FileValName.CreateTime.value
+            )
+            compile_time = _get_datetime_utc_epoch_value(
+                values, Win8FileValName.CompileTime.value
+            )
+            sha1_hash = _get_string_value(values, Win8FileValName.SHA1Hash.value)
+            vollog.debug(f"Found sha1hash {sha1_hash}")
+            product_name = _get_string_value(values, Win8FileValName.Product.value)
+
+            yield program_id, _AmcacheEntry(
+                AmcacheEntryType.File.name,
+                path=path,
+                company=company,
+                last_modify_time=last_mod_time,
+                last_modify_time_2=last_mod_time_2,
+                install_time=install_time,
+                compile_time=compile_time,
+                sha1_hash=(
+                    sha1_hash.lstrip("0000")
+                    if isinstance(sha1_hash, str)
+                    else sha1_hash
+                ),
+                product_name=product_name,
+            )
+
+    @classmethod
+    def parse_programs_key(
+        cls, programs_key: reg_extensions.CM_KEY_NODE
+    ) -> Iterator[Tuple[str, _AmcacheEntry]]:
+        """Parses Program entries from the Windows 8 `Root\\Programs` key.
+
+        :param programs_key: The `Root\\Programs` registry key.
+
+        :return: An iterator of tuples, where the first member is the program ID string for
+        correlating `Root\\File` entries, and the second member is the `AmcacheEntry`.
+        """
+        for program_key in programs_key.get_subkeys():
+            values = {
+                str(value.get_name()): value
+                for value in program_key.get_values()
+                if value.get_name() in [key.value for key in Win8ProgramValName]
+            }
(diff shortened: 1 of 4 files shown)


volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,605 @@
...
+        if amcache is None:
+            return
+
+        try:
+            yield from indented(
+                self.parse_driver_binary_key(
+                    amcache.get_key("Root\\InventoryDriverBinary")  # type: ignore
+                )
+            )
+        except KeyError:
```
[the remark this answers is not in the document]
dgmcdona: fixed

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,605 @@
...
+            ),
+        ):
+            files_indent = 0
+
+            if isinstance(program_id, str):
+                try:
+                    program_entry = programs.pop(program_id.strip().strip("\u0000"))
+                    yield (0, program_entry)
+                    files_indent = 1
+                except KeyError:
```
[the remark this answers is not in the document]
dgmcdona: fixed

atcuno:
@ikelos can this one get reviewed before it gets stale?

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,603 @@
...
+)
+
+
+def _try_get_value(
+    values: List[reg_extensions.CM_KEY_VALUE],
+    name: str,
+    expected_type: reg_extensions.RegValueTypes,
+) -> Any:
+    try:
+        value = next(
```
ikelos: So I can't quite see why this is using next rather than just a 0 index?  It looks like you're just using the head of the list because every path from this next returns (meaning you never get to the second item)?  I don't specifically have anything against it, but it makes me wonder what's going on (and why it's provided a list of values) if it only ever uses the first one?  I'd have gone with a list comprehension and a 0 index (if the list has any contents)...

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,603 @@
...
+    ]:
+        if isinstance(value, int):
+            return value
+        else:
+            return renderers.UnparsableValue()
+
+    return value
+
+
+def _sort_entries(entry_tuple):
```
ikelos: Feels as though this could be a lambda, but again not a show stopper, and it offers clarity, but if it's gonna be a new function then proper typing (and mabye even a docstring) please.  Just because it's short doesn't mean it needs no explaining...

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,603 @@
...
+            if isinstance(last_mod, datetime.datetime):
+                yield f"Amcache: {entry_type} {path} registry key modified", timeliner.TimeLinerType.MODIFIED, last_mod
+            if isinstance(last_mod_2, datetime.datetime):
+                yield f"Amcache: {entry_type} {path} STANDARD_INFORMATION create time", timeliner.TimeLinerType.CREATED, last_mod_2
+            if isinstance(install_time, datetime.datetime):
+                yield f"Amcache: {entry_type} {path} installed", timeliner.TimeLinerType.CREATED, install_time
+            if isinstance(compile_time, datetime.datetime):
+                yield f"Amcache: {entry_type} {path} compiled (PE metadata)", timeliner.TimeLinerType.MODIFIED, compile_time
+
+    @classmethod
```
ikelos, line 250: This is usually the first method, but I'm not gonna be picky...

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,603 @@
...
+    def get_amcache_hive(
+        cls,
+        context: interfaces.context.ContextInterface,
+        config_path: str,
+        kernel: interfaces.context.ModuleInterface,
+    ) -> Optional[registry.RegistryHive]:
+        """Retrieves the `Amcache.hve` registry hive from the kernel module, if it can be located."""
+        return next(
```
ikelos, line 258: Again with the next.  Happy you're consistent but it still makes me look around for a list that doesn't seem to exist...  5:P

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,603 @@
...
+                program_id,
+                path,
+                company,
+                product,
+                sha1_hash,
+                last_mod_raw,
+                last_mod_2_raw,
+                create_time_raw,
+            ) = (
+                _try_get_value(values, "100", reg_extensions.RegValueTypes.REG_SZ),
```
ikelos: Are these values and their meanings documented somewhere?  Might be nice to have an enum or a dictionary just for humans to translate what the number is to what the value means.  Just feels like like ordered lists of specific numbers like this are fragile, and keeping a separate clear table somewhere makes it easier to tinker or change without breaking everything like a giant game of jenga...  5;)

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,603 @@
...
+                _try_get_value(values, "1", reg_extensions.RegValueTypes.REG_SZ),
+                _try_get_value(values, "0", reg_extensions.RegValueTypes.REG_SZ),
+                _try_get_value(values, "101", reg_extensions.RegValueTypes.REG_SZ),
+                _try_get_value(values, "11", reg_extensions.RegValueTypes.REG_SZ),
+                _try_get_value(values, "17", reg_extensions.RegValueTypes.REG_SZ),
+                _try_get_value(values, "12", reg_extensions.RegValueTypes.REG_SZ),
+            )
+
+            create_time = (
+                conversion.wintime_to_datetime(create_time_raw)
```
ikelos: I feel like we should now have type hints that do this (eg, HexOrAbsent), but reasonably ok with this.  Just wondering if there's someway to make them more compact or shrink the duplication, but I'm not convinced there is save for a full subfunction, so this is probably fine.

ikelos (commented):
Ok, generally looks ok, a bunch of comments but most of them are stylistic.  The most important one from my perspective would be to get those arbitrary number keys looked up from a dictionary or an enum somewhere, just so that mere humans can see what's supposed to be going on...

volatility3/framework/plugins/windows/amcache.py
dgmcdona: The advantage of `next` with a generator comprehension over using a list comprehension with a zero-index is that creating a complete list of values, even a filtered one, using a list comprehension requires performing the `.get_name().lower().strip().strip("\u0000")` operations on every value, even those in the iterator beyond the one we're looking for. Using `next` with a generator expression only performs those operations as needed, and stops evaluating once the desired value is found.

ikelos: You're right about the efficiency.  I still find standalone calls to `next` a little unusual (rather than a `for...: break`) but that's ok, it just threw me a little (and made me wonder why we stopped after the first one, what are the others used for, but a for loop would have the same effect))?

volatility3/framework/plugins/windows/amcache.py
dgmcdona: I've created enums for each set of Amcache entry value value names, and added a reference in a comment for those who want to seek out more information

volatility3/framework/plugins/windows/amcache.py
dgmcdona: I split it out into a new function since it's essential that the implementations match for `group_by` and `sorted`, added a comment to it plus type hints.

dgmcdona:
@ikelos I made some additions to the `CM_KEY_VALUE` class to hopefully help future plugin authors avoid common pitfalls when dealing with the registry. The `get_string_value`/`get_binary_value`/`get_int_value` methods are meant to:
- Remind plugin authors that registry value data may not be of the type that they would typically expect when retrieving it, and they should be prepared to handle that case. They have the option of providing a default value, or handling the new exception class themselves when it is raised. The new exception contains the actual value type of the data that they are attempting to access.
- Gracefully handle the two major failure cases. These are that the value may be paged out or otherwise absent, in which case those methods return `None`, or that the value is not of the type they expect, in which case an `UnexpectedValueTypeException` is raised or their supplied default value is returned. In my case, I have set the default to `renderers.UnparsableValue()` since there's nothing else I can do with the data at that point.

It worked well in this case, but I'm open to any suggestions you might have for how to improve this interface.

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -5,9 +5,9 @@
 import enum
 import logging
 import struct
-from typing import Iterable, Optional, Union
+from typing import Iterable, List, Optional, Union
 
-from volatility3.framework import constants, exceptions, interfaces, objects
+from volatility3.framework import T, constants, exceptions, interfaces, objects
```
ikelos: Please define a new `TypeVar` here, there seems little reason to reuse the only from the framework?  Its purpose is to make sure that the input class and the output class of a function are the same type.

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -52,6 +75,24 @@ class RegKeyFlags(enum.IntEnum):
     KEY_VIRTUAL_STORE = 0x200
 
 
+class UnexpectedValueTypeException(Exception):
```
ikelos: I wondering if this should live here or in `exceptions`?  Which do you think?  Either way I suspect it should probably inherit from `VolatilityException` rather than just a standard `Exception`.

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -132,6 +173,103 @@ def get_full_key_name(self) -> str:
...
+            RegValueTypes.REG_LINK,
+            RegValueTypes.REG_MULTI_SZ,
+            RegValueTypes.REG_BINARY,
+            RegValueTypes.REG_FULL_RESOURCE_DESCRIPTOR,
+            RegValueTypes.REG_RESOURCE_LIST,
+            RegValueTypes.REG_RESOURCE_REQUIREMENTS_LIST,
+        ]:
+            return data
+        if self.get_type() == RegValueTypes.REG_NONE:
+            return b""
```
ikelos: So, should we be returning an empty (binary) string here, rather than an actual python `None`?  Do we use `None` as an error condition in this?  Generally empty strings meaning "nothing" are hard to distinguish from strings that are intentionally empty, so I'd prefer to return None if we could, but it's there's a good reason I'll listen to it...

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -132,6 +173,103 @@ def get_full_key_name(self) -> str:
...
+
+    def get_type(self) -> RegValueTypes:
+        """Get the type of the registry value"""
+        return RegValueTypes(self.Type)
+
+    def decode_data(self) -> Union[int, bytes]:
+        """Properly decodes the data associated with the value node"""
+        # Determine if the data is stored inline
+        datalen = self.DataLength
+        data = b""
```
ikelos: See later on where I talk about blank strings.

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -152,9 +290,173 @@ def get_subkeys(self) -> Iterable[interfaces.objects.ObjectInterface]:
             subkey_node = hive.get_cell(self.SubKeyLists[index]).u.KeyIndex
             yield from self._get_subkeys_recursive(hive, subkey_node)
 
+    def get_int_value(
+        self, name: str, default: Optional[T] = None
```
ikelos: Should the default also have to an `int` (or an `Optional[int]`) here?

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -152,9 +290,173 @@ def get_subkeys(self) -> Iterable[interfaces.objects.ObjectInterface]:
...
+        :raises: UnexpectedValueTypeException if default is not supplied or is None
+        """
+        try:
+            value = self.get_value_data(name, INTEGER_TYPES)
+        except UnexpectedValueTypeException:
+            if default is not None:
+                return default
+            raise
+
+        return value if isinstance(value, int) else None
```
ikelos: When would `get_value_data(whatever, INTEGER_TYPES)` return something that wasn't an integer (or None)?  I really get the feeling it should be possible to return `None`?

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -152,9 +290,173 @@ def get_subkeys(self) -> Iterable[interfaces.objects.ObjectInterface]:
...
+            )
+        except UnexpectedValueTypeException:
+            if default is not None:
+                return default
+            raise
+
+        return value if isinstance(value, str) else None
+
+    def get_multi_string_value(
+        self, name: str, default: Optional[T] = None
```
ikelos: So, default doesn't have to be a `List` here?

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -152,9 +290,173 @@ def get_subkeys(self) -> Iterable[interfaces.objects.ObjectInterface]:
...
+        try:
+            value = self.get_value_data(name, [RegValueTypes.REG_MULTI_SZ])
+        except UnexpectedValueTypeException:
+            if default is not None:
+                return default
+            raise
+
+        return (
+            value
+            if isinstance(value, list) and all(isinstance(val, str) for val in value)
```
ikelos: Don't list iterators always need to be in square brackets?  Fine if not, just stood out to me a little is all.  5:)

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -152,9 +290,173 @@ def get_subkeys(self) -> Iterable[interfaces.objects.ObjectInterface]:
...
+        :returns: The value data, if found and matches the optional expected type, or None if the value was not found, or could not be read.
+        :raises: UnexpectedValueTypeException if the value data is not one of the optional user-specified types.
+        """
+
+        try:
+            value = next(
+                (
+                    val
+                    for val in self.get_values()
+                    if val.get_name().lower().strip().strip("\u0000") == name.lower()
```
ikelos: Is case definitely not important?  Presumably you could theoretically get two values that differ only in case?  Just want to check that this is absolutely valid for what we're doing?

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -170,7 +472,7 @@ def _get_subkeys_recursive(
         elif signature == "lh" or signature == "lf":
             listjump = 2
         elif node.vol.type_name.endswith(constants.BANG + "_CM_KEY_NODE"):
-            yield node
+            yield node  # type: ignore
```
ikelos: I'm not a big fan of `# type: ignore` because it often means "we don't want to think about this" and quickly becomes "we've forgotten about this".  What are the circumstances here where we're not yield the right type?

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,690 @@
...
+    sha1_hash: NullableString = renderers.NotApplicableValue()
+    service: NullableString = renderers.NotApplicableValue()
+    product_name: NullableString = renderers.NotApplicableValue()
+
+    def __post_init__(self):
+        # Loop through the fields
+        for field in dataclasses.fields(self):
+            if getattr(self, field.name) is None:
+                # Map 'None' values that have been explicitly set to `NotAvailableValue()`
+                setattr(self, field.name, renderers.NotAvailableValue())
```
ikelos: I worry that this is using renderer values (which are typically only for the renderer to figure out how to display them) in code that may be used by plugins (thus requiring the plugins to have to deal with them too).  Generally these values are only supplanted in over None values just before they're yielded into TreeGrid.  I'm happy that this is a separate class, and its only purpose may be for displaying output, but if it's like to get used in normal python code, then it may be better to stick to python's `None` type, and then making the yield statement, something like `this or renderers.NotAvailableValue`?

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,690 @@
...
+
+
+def _entry_sort_key(entry_tuple: Tuple[str, AmcacheEntry]) -> str:
+    """Sorts entries by program ID. This is broken out as a function here
+    to ensure consistency in sorting between the `group_by` and `sorted` function
+    invocations.
+    """
+    program_id, _ = entry_tuple
+    key = program_id if isinstance(program_id, str) else ""
```
ikelos, line 144: Why are we returning a blank string on what essentially looks like failure?  Shouldn't we be distinguishing it somehow (such as returning `None`)?

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,690 @@
...
+        :return: An iterator of tuples, where the first member is the program ID string for
+        correlating `Root\\Program` entries, and the second member is the `AmcacheEntry`.
+        """
+
+        def _handle_timestamp(value) -> NullableDatetime:
+            return (
+                conversion.wintime_to_datetime(value)
+                if isinstance(value, int)
+                else (
+                    renderers.NotAvailableValue()
```
ikelos: Ok, distinguishing between these two states does seems worthwhile, so I'm happy to be easier on using them/passing them around.

ikelos (commented):
Cool, seems good.  I'm somewhat willing to relent on my suggestion of keeping the AbsentValues until the rendering because you do actually seem to be distinguishing on types.  I still thing you could leave out the default argument and then just provide `value or UnparsableValue(),` in the yield statement, but if the default parameter is useful I won't complain.  Just a couple little house keeping bits, but generally a very clean and tidy plugin with very few deviations for the normal coding style for volatility.  Good work!  5:D

volatility3/framework/symbols/windows/extensions/registry.py
dgmcdona: Yeah, registry value names are case insensitive, so this is the behavior we want.

volatility3/framework/symbols/windows/extensions/registry.py
dgmcdona: I just used it there because my language server was complaining about `ObjectInterface` not being an instance of `CM_KEY_NODE`, since it doesn't understand that it's going to have that extension class created instead. I'll remove this and use a `cast` for clarity. We've talked about too much casting being overkill in the past, so if you feel like that's the case here, I'm happy to remove both the cast and the comment.

volatility3/framework/plugins/windows/amcache.py
dgmcdona: `None` isn't comparable to `str` values, and this is being used as a sort key, so we have to return a `str`.

volatility3/framework/plugins/windows/amcache.py
dgmcdona: I was trying to find a good way to differentiate on the things that are 'definitively missing' vs things that have an unexpected value type or could not be recovered due to `InvalidPagedAddressException`s and I'm still not 100% sure I'm happy with this since, as you said, it feels like it would be best to let the consumers of these things deal with the renderers.

volatility3/framework/symbols/windows/extensions/registry.py
dgmcdona: I'm used to seeing `REG_NONE` values being treated as empty binary containers, so I chose this return type as it was just consistent with my prior experience; I just did a sanity check across a couple of test hives to make sure that this will always be the case. I think we can safely return `None` here, I'll make the change.

volatility3/framework/symbols/windows/extensions/registry.py
dgmcdona: The idea behind using `T` here instead of int was that it gives consumers the ability to decide what type they want to return in the case of missing values. I definitely didn't want `renderers` this deep in the code, so this was my attempt to solve that same problem again.

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,690 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
```
ikelos: The copyright should be the year the code was first added, so 2024 please.

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,690 @@
...
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import dataclasses
+import datetime
+import enum
+import logging
+from itertools import groupby, chain
```
ikelos: Please just import `itertools` and then reference the methods as `itertools.groupby` or `groupby.chain`.  This prevents people trying to import it from this module rather than where it's actually defined.

volatility3/framework/symbols/windows/extensions/registry.py
ikelos: Well, ignoring is definitely bad, but casting feels like somewhat ignoring the problem too, but I guess we'll go with cast.

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -132,6 +175,103 @@ def get_full_key_name(self) -> str:
...
+            RegValueTypes.REG_LINK,
+            RegValueTypes.REG_MULTI_SZ,
+            RegValueTypes.REG_BINARY,
+            RegValueTypes.REG_FULL_RESOURCE_DESCRIPTOR,
+            RegValueTypes.REG_RESOURCE_LIST,
+            RegValueTypes.REG_RESOURCE_REQUIREMENTS_LIST,
+        ]:
+            return data
+        if self.get_type() == RegValueTypes.REG_NONE:
+            return None
```
ikelos: Returning None rather than `b""` is better, but unfortunately is also an API change, so strictly this needs a the MAJOR version bumping (which I suspect will be of the whole framework), which will be a pain (because every plugin will need updating)...  5:S

Are we making use of the None return value in particular?

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -132,6 +175,103 @@ def get_full_key_name(self) -> str:
         return "\\".join(reversed(output))
 
 
+class CM_KEY_VALUE(objects.StructType):
```
ikelos: It;s really hard to see what's changed when this has (mostly) just moved position in the file).  I realize it's been done to allow for typing, but I'd almost sooner split it into two PRs, one where the type is in quotes for delayed resolution, and then another shifting the code about...  5:S

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -52,6 +77,24 @@ class RegKeyFlags(enum.IntEnum):
     KEY_VIRTUAL_STORE = 0x200
 
 
+class UnexpectedValueTypeException(exceptions.VolatilityException):
```
ikelos: I wondering if this shouldn't live in `exceptions` rather than off on its own here?

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,690 @@
...
+                driver_name = binary_key.get_string_value(
+                    valName.DriverName.name,
+                    renderers.UnparsableValue(),
+                )
+
+            if isinstance(sha1_hash, str):
+                sha1_hash = sha1_hash[4:] if sha1_hash.startswith("0000") else sha1_hash
+
+            company, product, service, last_write_time, driver_timestamp = (
+                binary_key.get_string_value(
```
ikelos: This makes for very long-winded code, because we can't use `None` as the default and then change `None` to `UnparseableValue()` using `or UnparseableValue()` even though that's the majority of the cases here (since `None` as the default causes it to throw the exception)...  Not certain what to do here but it feels like having to specify a default value every time is not ideal.

I'm leaning to just returning `None` whenever it can't deal with a value and the people could call `get_value_data` if they needed it to throw the wrong_type exception.  Otherwise the `get_binary_data` methods, etc, are all just glorified ways of catching the exception?

ikelos: That would then allow all the `get_x_data(thing, UnparsableValues())` calls to be swapped out for `or UnparsableValue()` at the time the variable is spat out of the generator.

ikelos (commented):
Hiya, sorry the unparsablevalues are causing such a headache.  Looking better now, but there's a couple changes that would require bumping version numbers (one of them the framework MAJOR version number, so I'd prefer to avoid that if possible).

The `get_x_data()` methods are a little cumbersome, and I think it would be better if they always returned `None` when we're not interested in the `UnexpectedType` exception (and then people can call `get_value_data()` when they actually need the exception.  That might also remove a lot of excess default values that could be handled as more succinctly in the generator return.  Otherwise just a couple bits and pieces, but I do appreciate you taking the time to help me make sure that the API is all good for the future as well as this particular use-case...  5:)

volatility3/framework/symbols/windows/extensions/registry.py
dgmcdona: Moved it back to it's original position 👍

volatility3/framework/plugins/windows/amcache.py
dgmcdona: I think I was trying too hard to avoid the requirements of the language (handling exceptions) in order to prevent future runtime errors. It's probably better to go back to the way things were before, for the most part, and plugin authors will just have to be careful about checking values types in their own code.

volatility3/framework/symbols/windows/extensions/registry.py
dgmcdona: Not particularly; I'll revert this.

volatility3/framework/plugins/windows/amcache.py
dgmcdona: Fixed

volatility3/framework/plugins/windows/amcache.py
dgmcdona: Fixed

dgmcdona:
After thinking things over, I decided to roll back the additional methods that I had originally planned to add to the registry extension `CM_KEY_VALUE`/`CM_KEY_NODE` classes, as they were possibly introducing more complexity than they were solving problems. As far as differentiating on missing vs unparsable values, I have moved that logic into helper functions in the `amcache` module. 

I also removed the changes that would have required a major framework version bump.

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -249,6 +282,10 @@ def get_name(self) -> interfaces.objects.ObjectInterface:
         self.Name.count = namelength
         return self.Name.cast("string", max_length=namelength, encoding="latin-1")
 
+    def get_type(self) -> RegValueTypes:
```
ikelos, line 276: Additions like this should still bump the MINOR version number of the framework, but luckily nothing else will have to change (any plugins that call these methods though will have to depend on that later framework version).

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,646 @@
...
+    try:
+        value = values[name]
+    except KeyError:
+        return renderers.NotAvailableValue()
+
+    data = value.decode_data()
+    if not isinstance(data, int):
+        return renderers.UnparsableValue()
+
+    if isinstance(data, str):
```
ikelos, line 208: This seems... strange?  Does the registry record datetime values as... strings?!?!  Without a timezone?!?!?

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,646 @@
...
+        :return: An iterator of tuples, where the first member is the program ID string for
+        correlating `Root\\InventoryApplicationFile` entries, and the second member is the `AmcacheEntry`.
+        """
+        for program_key in inv_app_key.get_subkeys():
+            program_id = program_key.get_name()
+
+            values = {
+                str(value.get_name()): value
+                for value in program_key.get_values()
+                if value.get_name() in [key.value for key in Win10InvAppValName]
```
ikelos: Almost certainly a small point, but this list comprehension looks like it could be done once outside the loop and then just referenced in the loop?

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,646 @@
...
+            install_date = _get_datetime_str_value(
+                values, Win10InvAppValName.InstallDate.value
+            )
+            last_mod = conversion.wintime_to_datetime(
+                program_key.LastWriteTime.QuadPart
+            )
+
+            product: str = name if isinstance(name, str) else "UNKNOWN"  # type: ignore
+            if isinstance(version, str):
+                product += " " + version
```
ikelos: Should we not be returning `version` as a separate field, so another plugin could parse it easily?  I guess they're both strings, it just feels like if they're separate now, why combine them rather than pass them through as two fields?

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,646 @@
...
+        for file_key in inv_app_file_key.get_subkeys():
+
+            vollog.debug(
+                f"Parsing Win10 InventoryApplicationFile key {file_key.get_name()}"
+            )
+
+            values = {
+                str(value.get_name()): value
+                for value in file_key.get_values()
+                if value.get_name() in [key.value for key in valName]
```
ikelos: Again, this list comprehension could be done before the loop even starts rather than every run of a loop inside a loop...

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,646 @@
...
+        :return: An iterator of `AmcacheEntry`s
+        """
+        for binary_key in driver_binary_key.get_subkeys():
+
+            valName = Win10DriverBinaryValName
+
+            values = {
+                str(value.get_name()): value
+                for value in binary_key.get_values()
+                if value.get_name() in [key.value for key in valName]
```
ikelos: Same again

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -0,0 +1,646 @@
...
+        cls, driver_binary_key: reg_extensions.CM_KEY_NODE
+    ) -> Iterator[_AmcacheEntry]:
+        """Parses information about installed drivers from the `Root\\InventoryDriverBinary` registry key.
+
+        :param driver_binary_key: The `Root\\InventoryDriverBinary` registry key
+        :return: An iterator of `AmcacheEntry`s
+        """
+        for binary_key in driver_binary_key.get_subkeys():
+
+            valName = Win10DriverBinaryValName
```
ikelos: This doesn't need to be reset each time round the loop, feels like it could be set once before the loop starts...

ikelos (commented):
Cool, just gave it another quick once over.  It does need a version bump, but luckily nothing overly dramatic!  Otherwise there's a little optimization I think we could do but it's pretty much good to go.  The version bump (on the framework, and in the plugin `required_framework_version` are all that'd be needed).

volatility3/framework/plugins/windows/amcache.py
dgmcdona: It sure does :) This Windows component encodes timestamps in at least 3 different formats (FILETIME, unix epoch, this string format)

ikelos (approved):
Awesome, thanks for fixing up those last points!  5:)

PR #1212, opened 2024-07-23
Add pe_symbols plugin for user specified symbol resolution as well as…

atcuno:
… programmatic access

Examples of using the plugin:

```
$ python3 vol.py --config config.json windows.pe_symbols --source processes --module ntdll.dll --symbol NtProtectVirtualMemory
Volatility 3 Framework 2.7.2
Module	Symbol	                           Address
ntdll.dll	NtProtectVirtualMemory  0x7ff8ef460060

$ python3 vol.py --config config.json windows.pe_symbols --source kernel --module ntoskrnl.exe --symbol ZwOpenThread
Volatility 3 Framework 2.7.2
Module        Symbol       Address
ntoskrnl.exe  ZwOpenThread 0xf801082660d0
```

It is common to see raw addresses when using dis() in volshell or running plugins like vadyarascan and malfind. This provides a quick way for user's to resolve symbols, which otherwise is very painful.. dump the executable -> throw into IDA or Ghidra or parse with a PE parsing tool. 

The plugin also provides the API that quite a few other plugins coming for Black Hat, DEFCON, and October will use.

Note that we don't need to specifically list which process a DLL resolved a symbol in as the DLLs will be at the same address in each process per boot, even with ASLR on.

### DIFF
--- a/volatility3/framework/plugins/windows/pe_symbols.py
+++ b/volatility3/framework/plugins/windows/pe_symbols.py
@@ -262,7 +262,10 @@ class PESymbols(interfaces.plugins.PluginInterface):
         layer_name: str,
         symbol_table: str,
         wanted_symbols: Dict[str, List[str]],
-        module_collector: Callable[[interfaces.context.ContextInterface, str, str, Dict[str, List[str]]], Dict[str, Tuple[str, int]]]
+        module_collector: Callable[
+            [interfaces.context.ContextInterface, str, str, Dict[str, List[str]]],
+            Dict[str, Tuple[str, int]],
+        ],
     ) -> Dict[str, Tuple[str, int]]:
         """
         Loops through each method of symbol analysis until each wanted symbol is found
--- a/volatility3/framework/plugins/windows/pe_symbols.py
+++ b/volatility3/framework/plugins/windows/pe_symbols.py
@@ -112,6 +112,29 @@ class PESymbols(interfaces.plugins.PluginInterface):
     ):
         """
         Attempts to resolve the symbols in `wanted_symbols` through export table analysis
+        wanted_symbols is a dictionary of lower case DLL names, whose values are a list of symbols
+        to resolve per-DLL. Example from apihooks:
+
+        userland_apis = {
+            "wininet.dll": [
+                "HttpSendRequestA",
+                "HttpSendRequestW",
+                "HttpSendRequestExA",
+                ...
+                ],
+
+            "kernel32.dll": [
+                "GetProcAddress",
+                "LoadLibrary",
+                "LoadLibraryA",
+                "LoadLibraryExA",
+                ...
+                ],
+        }
+
+        found_symbols is populated with symbols as they are resolved.
+        It holds a dictionary of DLL names and its values are a list of
+        (function name, runtime address) of each symbol found
         """
         pe_table_name = intermed.IntermediateSymbolTable.create(
             context, config_path, "windows", "pe", class_types=pe.class_types
--- a/volatility3/framework/plugins/windows/pe_symbols.py
+++ b/volatility3/framework/plugins/windows/pe_symbols.py
@@ -1,0 +1,437 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+
+import io
+import logging
+from typing import Dict, Tuple, Optional, List
+
+import pefile
+
+from volatility3.framework import interfaces, exceptions
+from volatility3.framework import renderers, constants
+from volatility3.framework.configuration import requirements
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.symbols import intermed
+from volatility3.framework.symbols.windows import pdbutil
+from volatility3.framework.symbols.windows.extensions import pe
+from volatility3.plugins.windows import pslist, vadinfo, modules
+
+vollog = logging.getLogger(__name__)
+
+# used for special handling of the kernel PDB file. See later notes
+os_module_name = "ntoskrnl.exe"
+
+
+class PESymbols(interfaces.plugins.PluginInterface):
+    """Prints symbols in PE files in process and kernel memory"""
+
+    _required_framework_version = (2, 7, 0)
+
+    @classmethod
+    def get_requirements(cls):
+        # Since we're calling the plugin, make sure we have the plugin's requirements
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Windows kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.VersionRequirement(
+                name="pslist", component=pslist.PsList, version=(2, 0, 0)
+            ),
+            requirements.VersionRequirement(
+                name="vadinfo", component=vadinfo.VadInfo, version=(2, 0, 0)
+            ),
+            requirements.VersionRequirement(
+                name="modules", component=modules.Modules, version=(2, 0, 0)
+            ),
+            requirements.VersionRequirement(
+                name="pdbutil", component=pdbutil.PDBUtility, version=(1, 0, 0)
+            ),
+            requirements.ChoiceRequirement(
+                name="source",
+                description="Where to resolve symbols.",
+                choices=["kernel", "processes"],
+                optional=False,
+            ),
+            requirements.StringRequirement(
+                name="module",
+                description='Module in which to resolve symbols. Use "ntoskrnl.exe" to resolve in the base kernel executable.',
+                optional=False,
+            ),
+            requirements.StringRequirement(
+                name="symbol",
+                description="Symbol name to resolve",
+                optional=False,
+            ),
+        ]
+
+    @staticmethod
+    def _get_pefile_obj(
+        context, pe_table_name: str, layer_name: str, base_address: int
+    ) -> Optional[pefile.PE]:
+        """
+        Attempts to pefile object from the bytes of the PE file
+
+        Args:
+            pe_table_name: name of the pe types table
+            layer_name: name of the process layer
+            base_address: base address of the module
+
+        Returns:
+            the constructed pefile object
+        """
+        pe_data = io.BytesIO()
+
+        try:
+            dos_header = context.object(
+                pe_table_name + constants.BANG + "_IMAGE_DOS_HEADER",
+                offset=base_address,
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/pe_symbols.py
```diff
@@ -0,0 +1,437 @@
...
+        for proc_layer_name, vad_start, _ in vads:
+            # we need a valid PE with an export table
+            pe_module = PESymbols._get_pefile_obj(
+                context, pe_table_name, proc_layer_name, vad_start
+            )
+            if not pe_module:
+                continue
+
+            pe_module.parse_data_directories(
+                directories=[pefile.DIRECTORY_ENTRY["IMAGE_DIRECTORY_ENTRY_EXPORT"]]
```
ikelos, line 155: Can this throw a KeyError, or does the `hasattr` check below catch all the possible issues?

volatility3/framework/plugins/windows/pe_symbols.py
```diff
@@ -0,0 +1,437 @@
...
+            )
+            if not hasattr(pe_module, "DIRECTORY_ENTRY_EXPORT"):
+                continue
+
+            # walk the export table looking for symbols of interest
+            for export in pe_module.DIRECTORY_ENTRY_EXPORT.symbols:
+                # AttributeError throws on empty or ordinal-only exports
+                try:
+                    func = export.name.decode("ascii")
+                except AttributeError:
```
ikelos, line 165: This is to catch if `name` doesn't exist?

volatility3/framework/plugins/windows/pe_symbols.py
```diff
@@ -0,0 +1,437 @@
...
+    def _find_through_exports(
+        context,
+        config_path: str,
+        mod_name: str,
+        vads: Tuple[int, int, int],
+        wanted_symbols: Dict[str, List[str]],
+        found_symbols: Dict[str, Tuple[str, int]],
+    ):
+        """
+        Attempts to resolve the symbols in `wanted_symbols` through export table analysis
```
ikelos, line 116: Could we get a description of the args please?  Just trying to figure out the format of `wanted_symbols` and `found_symbols` without hunting for the call to it...

volatility3/framework/plugins/windows/pe_symbols.py
```diff
@@ -0,0 +1,437 @@
...
+        return pe_ret
+
+    @staticmethod
+    def _find_through_exports(
+        context,
+        config_path: str,
+        mod_name: str,
+        vads: Tuple[int, int, int],
+        wanted_symbols: Dict[str, List[str]],
+        found_symbols: Dict[str, Tuple[str, int]],
```
ikelos, line 113: Feels like this should be return value, rather than an input parameter?

ikelos: It does make it a little easier to combine the two outputs of the methods but...  it feels slightly odd that the methods don't return any intermediate state?

volatility3/framework/plugins/windows/pe_symbols.py
```diff
@@ -0,0 +1,437 @@
...
+                    continue
+
+                if func in wanted_symbols[mod_name]:
+                    address = export.address
+
+                    if mod_name not in found_symbols:
+                        found_symbols[mod_name] = []
+
+                    found_symbols[mod_name].append((func, vad_start + address))
+                    wanted_symbols[mod_name].remove(func)
```
ikelos, line 175: This is altering wanted_symbols that's a mutable type (so will affect future uses of the variable).  Is that right/intentional?

volatility3/framework/plugins/windows/pe_symbols.py
```diff
@@ -0,0 +1,437 @@
...
+                        proc_layer_name,
+                        pdb_name,
+                        vad_start,
+                        vad_size,
+                    )
+
+                    if mod_symbols:
+                        break
+
+                # this exception is expected when the PDB can't be found or downloaded
```
ikelos, line 222: Thank you!  Definitely appreciate comments like this!

volatility3/framework/plugins/windows/pe_symbols.py
```diff
@@ -0,0 +1,437 @@
...
+            if not wanted_symbols[mod_name]:
+                break
+
+    @staticmethod
+    def find_function_offsets(
+        context,
+        config_path: str,
+        layer_name: str,
+        symbol_table: str,
+        module_collector,
```
ikelos: Type signature of this callable would be good please...

volatility3/framework/plugins/windows/pe_symbols.py
```diff
@@ -0,0 +1,437 @@
...
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.symbols import intermed
+from volatility3.framework.symbols.windows import pdbutil
+from volatility3.framework.symbols.windows.extensions import pe
+from volatility3.plugins.windows import pslist, vadinfo, modules
+
+vollog = logging.getLogger(__name__)
+
+# used for special handling of the kernel PDB file. See later notes
+os_module_name = "ntoskrnl.exe"
```
ikelos: Uhhh, could this not live in `constants.windows` or just... somewhere rather than floating in the middle of a plugin file?

volatility3/framework/plugins/windows/pe_symbols.py
```diff
@@ -0,0 +1,437 @@
...
+                        proc_modules[filename] = []
+
+                    proc_modules[filename].append(
+                        (proc_layer_name, vad.get_start(), vad.get_size())
+                    )
+
+        return proc_modules
+
+    @staticmethod
+    def resolve_wanted_process_symbols(
```
ikelos, line 394: Have you considered using partials instead of defining two functions that just sub in a particular input parameter?  You'd shift module collector to by the last parameter, then do:

```
from functools import partial
find_offsets = partial(PESymbols.find_function_offsets, context, config_path, layer_name, symbol_table, wanted_symbols)
```
and then you could call either `find_offsets(PESymbols.get_wanted_process_modules)` or `find_offsets(PESymbols.get_wanted_kernel_modules)`, and if you added methods later on, they'd also be able to fit into the same system?

volatility3/framework/plugins/windows/pe_symbols.py
ikelos: If the module_collector were the last parameter, this could be curried/made a partial function very easily (see a bit later on).

ikelos (commented):
Cool, pretty big but I think it all makes sense.  There's a couple questions, a little more description around function parameters/arguments would be nice, but no major show stoppers!  A little uneasy about mutable types being thrown around as input parameters to get the output (it's quite a C style of coding), but if they're documented clearly that they're output parameters and may be modified it should be ok...

volatility3/framework/plugins/windows/pe_symbols.py
atcuno: Yes

volatility3/framework/plugins/windows/pe_symbols.py
atcuno: the hasattr is how the API says to do / is in the examples.

volatility3/framework/plugins/windows/pe_symbols.py
atcuno: yes, PE files have named and unnamed (ordinal only) exports. the attributeerror is the way shown to deal with unnamed ones.

volatility3/framework/plugins/windows/pe_symbols.py
atcuno: they are just worker functions for helping track symbol resolution.

volatility3/framework/plugins/windows/pe_symbols.py
atcuno: I will move it inside the class. Its just the convention I used since the kernel file can have multiple names even in the same build (see the comments about the PDB searching).

volatility3/framework/plugins/windows/pe_symbols.py
atcuno: I never used partials before. I could explore this method after Black Hat as some plugins for the Black Hat <-> October timeframe will have a similar setup. It definitely seems cleaner

volatility3/framework/plugins/windows/pe_symbols.py
atcuno: added

atcuno:
ok @ikelos I added examples in the comments for the code, and then addressed the rest here in github comments.

volatility3/framework/plugins/windows/pe_symbols.py
ikelos: Ok

volatility3/framework/plugins/windows/pe_symbols.py
ikelos: Mmmmm, note sure I'm a massive fan of pefile's way of doing things, but who am I to complain...  5:)

volatility3/framework/plugins/windows/pe_symbols.py
ikelos: Before this goes live, and the API becomes fixed (since this is a static method, therefore considered public and part of the API), could we move reorder the parameters please?

volatility3/framework/plugins/windows/pe_symbols.py
ikelos: Could you add a `FIXME` comment please, to ensure the things we want to do before October happen?

volatility3/framework/plugins/windows/pe_symbols.py
ikelos: Ok, I don't believe it changes in this plugin, so... feels like a constant.  Happy for it to live somewhere you're happy with, but the module isn't the right place.

ikelos:
Ok, just a few little bits, should be easy to sort and then we can put this in...

volatility3/framework/plugins/windows/pe_symbols.py
atcuno: fixed

volatility3/framework/plugins/windows/pe_symbols.py
atcuno: moved into the class.

atcuno:
Address feedback

atcuno:
@ikelos hold one this one.

PR #1213, opened 2024-07-23
Improvement: Linux/lsof

forensicxlab:
Hello, 

This PR is an improvement for the already existing LSOF plugin.

It is adding the information about the inodes and the linked permissions + file timestamps. 
Because there is timestamps, it was also added to the Timeliner.

**Current lsof output:**

```
  |  PID |         Process |  FD |        Path
* |    1 |         systemd |   0 |          /dev/null
* |    1 |         systemd |   1 |           /dev/null
* |    1 |         systemd |   2 |          /dev/null
* |    1 |         systemd |   3 |          /dev/kmsg
* |    1 |         systemd |   4 |          anon_inode:[8960]
* |    1 |         systemd |   5 |         anon_inode:[8960]
* |    1 |         systemd |   6 |         anon_inode:[8960]
* |    1 |         systemd |   7 |         /sys/fs/cgroup
...
```

**Proposed lsof output:**

```
  |  PID |         Process |  FD |               Path |      Inode |       Mode |                  LastChange |                  LastModify |                LastAccessed |     Size
* |    1 |         systemd |   0 |          /dev/null |          4 | crw-rw-rw- | 2022-09-02 14:19:51.000000  | 2022-09-02 14:19:51.000000  | 2022-09-02 14:19:51.000000  |        0
* |    1 |         systemd |   1 |          /dev/null |          4 | crw-rw-rw- | 2022-09-02 14:19:51.000000  | 2022-09-02 14:19:51.000000  | 2022-09-02 14:19:51.000000  |        0
* |    1 |         systemd |   2 |          /dev/null |          4 | crw-rw-rw- | 2022-09-02 14:19:51.000000  | 2022-09-02 14:19:51.000000  | 2022-09-02 14:19:51.000000  |        0
* |    1 |         systemd |   3 |          /dev/kmsg |         10 | crw-r--r-- | 2022-09-02 14:19:51.000000  | 2022-09-02 14:19:51.000000  | 2022-09-02 14:19:51.000000  |        0
* |    1 |         systemd |   4 |  anon_inode:[8960] |       8960 | ?rw------- | 2022-09-02 14:19:45.000000  | 2022-09-02 14:19:45.000000  | 2022-09-02 14:19:45.000000  |        0
* |    1 |         systemd |   5 |  anon_inode:[8960] |       8960 | ?rw------- | 2022-09-02 14:19:45.000000  | 2022-09-02 14:19:45.000000  | 2022-09-02 14:19:45.000000  |        0
* |    1 |         systemd |   6 |  anon_inode:[8960] |       8960 | ?rw------- | 2022-09-02 14:19:45.000000  | 2022-09-02 14:19:45.000000  | 2022-09-02 14:19:45.000000  |        0
* |    1 |         systemd |   7 |     /sys/fs/cgroup |          1 | dr-xr-xr-x | 2022-09-02 14:37:03.000000  | 2022-09-02 14:37:03.000000  | 2022-09-02 14:19:49.000000  |        0
```

Many thanks,

Regards.

### DIFF
--- a/volatility3/framework/plugins/linux/lsof.py
+++ b/volatility3/framework/plugins/linux/lsof.py
@@ -135,7 +135,6 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
     def generate_timeline(self):
         pids = self.config.get("pid", None)
         symbol_table = self.config["kernel"]
-        filter_func = pslist.PsList.create_pid_filter(self.config.get("pid", None))
         for row in self._generator(pids, symbol_table):
             _depth, row_data = row
             description = f'Process {row_data[1]} ({row_data[0]}) Open "{row_data[3]}"'
--- a/volatility3/framework/plugins/linux/lsof.py
+++ b/volatility3/framework/plugins/linux/lsof.py
@@ -1,4 +1,4 @@
-# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
 """A module containing a collection of plugins that produce data typically
@@ -12,16 +12,17 @@ from volatility3.framework.interfaces import plugins
 from volatility3.framework.objects import utility
 from volatility3.framework.symbols import linux
 from volatility3.plugins.linux import pslist
+from volatility3.plugins import timeliner
 
 vollog = logging.getLogger(__name__)
 
 
-class Lsof(plugins.PluginInterface):
+class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
     """Lists all memory maps for all processes."""
 
     _required_framework_version = (2, 0, 0)
 
-    _version = (1, 1, 0)
+    _version = (2, 0, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
@@ -46,7 +47,7 @@ class Lsof(plugins.PluginInterface):
         ]
 
     @classmethod
-    def list_fds(
+    def list_fds_and_inodes(
         cls,
         context: interfaces.context.ContextInterface,
         symbol_table: str,
@@ -67,27 +68,38 @@ class Lsof(plugins.PluginInterface):
             )
 
             for fd_fields in fd_generator:
-                yield pid, task_comm, task, fd_fields
+                fd_num, filp, full_path = fd_fields
+                inode_metadata = linux.LinuxUtilities.get_inode_metadata(context, filp)
+                try:
+                    inode_num, file_size, imode, ctime, mtime, atime = next(
+                        inode_metadata
+                    )
+                except Exception as e:
+                    vollog.warning(
+                        f"Can't get inode metadata for file descriptor {fd_num}: {e}"
+                    )
+                    continue
+                yield pid, task_comm, task, fd_num, filp, full_path, inode_num, imode, ctime, mtime, atime, file_size
 
     def _generator(self, pids, symbol_table):
         filter_func = pslist.PsList.create_pid_filter(pids)
-        fds_generator = self.list_fds(
+        fds_generator = self.list_fds_and_inodes(
             self.context, symbol_table, filter_func=filter_func
         )
-
-        for pid, task_comm, _task, fd_fields in fds_generator:
-            (
-                fd_num,
-                _filp,
-                full_path,
-                inode_num,
-                imode,
-                ctime,
-                mtime,
-                atime,
-                file_size,
-            ) = fd_fields
-
+        for (
+            pid,
+            task_comm,
+            task,
+            fd_num,
+            filp,
+            full_path,
+            inode_num,
+            imode,
+            ctime,
+            mtime,
+            atime,
+            file_size,
+        ) in fds_generator:
             fields = (
                 pid,
                 task_comm,
@@ -113,22 +125,20 @@ class Lsof(plugins.PluginInterface):
             ("Path", str),
             ("Inode", int),
             ("Mode", str),
-            ("LastChange", datetime.datetime),
-            ("LastModify", datetime.datetime),
-            ("LastAccessed", datetime.datetime),
+            ("Changed", datetime.datetime),
+            ("Modified", datetime.datetime),
+            ("Accessed", datetime.datetime),
             ("Size", int),
         ]
         return renderers.TreeGrid(tree_grid_args, self._generator(pids, symbol_table))
 
     def generate_timeline(self):
+        pids = self.config.get("pid", None)
+        symbol_table = self.config["kernel"]
         filter_func = pslist.PsList.create_pid_filter(self.config.get("pid", None))
-        for row in self._generator(
-            pslist.PsList.list_tasks(
-                self.context, self.config["kernel"], filter_func=filter_func
-            )
-        ):
+        for row in self._generator(pids, symbol_table):
             _depth, row_data = row
-            description = f'Process {row_data[1]} ({row_data[0]}) Open "{row_data[4]}"'
-            yield description, timeliner.TimeLinerType.CHANGED, row_data[5]
-            yield description, timeliner.TimeLinerType.MODIFIED, row_data[6]
-            yield description, timeliner.TimeLinerType.ACCESSED, row_data[7]
+            description = f'Process {row_data[1]} ({row_data[0]}) Open "{row_data[3]}"'
+            yield description, timeliner.TimeLinerType.CHANGED, row_data[6]
+            yield description, timeliner.TimeLinerType.MODIFIED, row_data[7]
+            yield description, timeliner.TimeLinerType.ACCESSED, row_data[8]
--- a/volatility3/framework/plugins/linux/lsof.py
+++ b/volatility3/framework/plugins/linux/lsof.py
@@ -21,7 +21,6 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
     """Lists all memory maps for all processes."""
 
     _required_framework_version = (2, 0, 0)
-
     _version = (2, 0, 0)
 
     @classmethod
@@ -53,7 +52,7 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
         symbol_table: str,
         filter_func: Callable[[int], bool] = lambda _: False,
     ):
-        linuxutils_symbol_table = None  # type: ignore
+        linuxutils_symbol_table = None
         for task in pslist.PsList.list_tasks(context, symbol_table, filter_func):
             if linuxutils_symbol_table is None:
                 if constants.BANG not in task.vol.type_name:
@@ -71,21 +70,17 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
                 fd_num, filp, full_path = fd_fields
                 inode_metadata = linux.LinuxUtilities.get_inode_metadata(context, filp)
                 try:
-                    inode_num, file_size, imode, ctime, mtime, atime = next(
+                    inode_num, itype, file_size, imode, ctime, mtime, atime = next(
                         inode_metadata
                     )
                 except Exception as e:
                     vollog.warning(
                         f"Can't get inode metadata for file descriptor {fd_num}: {e}"
                     )
-                    # Yield NotAvailableValue for each field in case of an exception
-                    inode_num = renderers.NotAvailableValue()
-                    file_size = renderers.NotAvailableValue()
-                    imode = renderers.NotAvailableValue()
-                    ctime = renderers.NotAvailableValue()
-                    mtime = renderers.NotAvailableValue()
-                    atime = renderers.NotAvailableValue()
-                yield pid, task_comm, task, fd_num, filp, full_path, inode_num, imode, ctime, mtime, atime, file_size
+                    inode_num = itype = file_size = imode = ctime = mtime = atime = (
+                        renderers.NotAvailableValue()
+                    )
+                yield pid, task_comm, task, fd_num, filp, full_path, inode_num, itype, imode, ctime, mtime, atime, file_size
 
     def _generator(self, pids, symbol_table):
         filter_func = pslist.PsList.create_pid_filter(pids)
@@ -100,6 +95,7 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
             filp,
             full_path,
             inode_num,
+            itype,
             imode,
             ctime,
             mtime,
@@ -112,6 +108,7 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
                 fd_num,
                 full_path,
                 inode_num,
+                itype,
                 imode,
                 ctime,
                 mtime,
@@ -130,6 +127,7 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
             ("FD", int),
             ("Path", str),
             ("Inode", int),
+            ("Type", str),
             ("Mode", str),
             ("Changed", datetime.datetime),
             ("Modified", datetime.datetime),
@@ -144,6 +142,6 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
         for row in self._generator(pids, symbol_table):
             _depth, row_data = row
             description = f'Process {row_data[1]} ({row_data[0]}) Open "{row_data[3]}"'
-            yield description, timeliner.TimeLinerType.CHANGED, row_data[6]
-            yield description, timeliner.TimeLinerType.MODIFIED, row_data[7]
-            yield description, timeliner.TimeLinerType.ACCESSED, row_data[8]
+            yield description, timeliner.TimeLinerType.CHANGED, row_data[7]
+            yield description, timeliner.TimeLinerType.MODIFIED, row_data[8]
+            yield description, timeliner.TimeLinerType.ACCESSED, row_data[9]
--- a/volatility3/framework/plugins/linux/lsof.py
+++ b/volatility3/framework/plugins/linux/lsof.py
@@ -18,10 +18,10 @@ vollog = logging.getLogger(__name__)
 
 
 class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
-    """Lists all memory maps for all processes."""
+    """Lists open files for each processes."""
 
     _required_framework_version = (2, 0, 0)
-    _version = (2, 0, 0)
+    _version = (1, 2, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
--- a/volatility3/framework/plugins/linux/lsof.py
+++ b/volatility3/framework/plugins/linux/lsof.py
@@ -78,7 +78,13 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
                     vollog.warning(
                         f"Can't get inode metadata for file descriptor {fd_num}: {e}"
                     )
-                    continue
+                    # Yield NotAvailableValue for each field in case of an exception
+                    inode_num = renderers.NotAvailableValue()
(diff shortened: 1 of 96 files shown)


eve-mem:
@forensicxlab really cool! I know I'd find this extra useful. I think it would make sense to up the version for the lsof plugin given its changing. 

Do you think it might also be worth adding some try/except for memory read issues in `files_descriptors_for_process`? It might end up bombing out if if can't read something and hold up the rest of the results. 

If you have the time it would also be good to add a test case for these new bits too - just so they don't get broken by mistake in the future.

eve-mem:
Actually the changes to `files_descriptors_for_process` might need a bump to the actual framework number, just in case there are private plugins making use of it that this will break. (Is that right @ikelos ?)

forensicxlab:
Hello @eve-mem!

Thank you for your advice, that would be indeed a good thing to implement such watchdogs. 

I will work on that. Waiting of ikelos comments as well so I can streamline the modifications :)

volatility3/framework/plugins/linux/lsof.py
```diff
@@ -76,14 +76,59 @@ def _generator(self, pids, symbol_table):
...
+        for row in self._generator(
+            pslist.PsList.list_tasks(
+                self.context, self.config["kernel"], filter_func=filter_func
+            )
+        ):
+            _depth, row_data = row
+            description = f'Process {row_data[1]} ({row_data[0]}) Open "{row_data[4]}"'
+            yield description, timeliner.TimeLinerType.CHANGED, row_data[5]
+            yield description, timeliner.TimeLinerType.MODIFIED, row_data[6]
+            yield description, timeliner.TimeLinerType.ACCESSED, row_data[7]
```
forensicxlab: Woups, row_data selection is wrong there. I will correct that.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -265,8 +266,26 @@ def files_descriptors_for_process(
         for fd_num, filp in enumerate(fds):
             if filp != 0:
                 full_path = LinuxUtilities.path_for_file(context, task, filp)
-
-                yield fd_num, filp, full_path
+                dentry = filp.get_dentry()
```
ikelos: Ok, so this is now a non-additive API change, and it would mean that the version number for this needs incrementing to `(3, 0, 0)` on line 64 (not the whole framework @eve-mem but yes, something used by lots of plugins and therefore would require bumping lots of core plugins requirements and potentially break some external plugins).

Another option would be to define an additional function but I'd then only return the `inode_num, imode, ctime, mtime, atime and file_size` since the rest can be recovered from the original call.  It's a little bit of code duplication, but you could actually extract the start of this out to a private method so they'd share the same code?  Doing that would then be an additive change, and the version would only need to go up to `(2, 2, 0)` and all the existing plugins would continue working just fine.

ikelos (commented):
Yep, @eve-mem was right on the money.  This looks good, but it is making a fundamental change to the `LinuxUtilities` class which is used by many different plugins, and so if you alter an existing function in a non-backwards compatible way then the major version needs bumping and all the plugins that needed it need checking/fixing to make sure they work correctly.  The other option is to shove the starting code into a share private method, and then just expose the additions under another function and that *would* be an additive change in which case every thing carries on working as is (although line 38 should also be updated to reflect the version of LinuxUtilities this plugin would need).  In either case, the version of LinuxUtilities needs to go up (line 64 of the file it's defined in)...

volatility3/framework/symbols/linux/__init__.py
forensicxlab: Hello @ikelos, 
Thank you for your guidance. 

I'll probably go with option two which indeed implies code reuse but will maximize compatibility with any other existing plugins. 

I'll update the code and propose a new implementation. 

Regards.

forensicxlab:
Hello @eve-mem, @ikelos, 

Implemented some changes.  Hope this is better :) 

Also updated the lsof plugin version to `(2,0,0)`.

The watchdogs implemented here ensures that even if the inode's metadata can't be fetch for some reason, the important data will still be available: 

Example of a simulated exception when fetching inode metadata:
```
* | 4545 |     Web Content |  51 |            /:[461] |     - |    - |       - |        - |        - |    -
* | 4545 |     Web Content |  52 |            /:[462] |     - |    - |       - |        - |        - |    -
* | 4545 |     Web Content |  53 |            /:[463] |     - |    - |       - |        - |        - |    -
* | 4545 |     Web Content |  54 |            /:[464] |     - |    - |       - |        - |        - |    -
* | 4545 |     Web Content |  55 |            /:[496] |     - |    - |       - |        - |        - |    -
* | 4675 |      (spatcher) |   0 |          /dev/null |     - |    - |       - |        - |        - |    -
* | 4675 |      (spatcher) |   1 |    socket:[135771] |     - |    - |       - |        - |        - |    -
* | 4675 |      (spatcher) |   2 |    socket:[135771] |     - |    - |       - |        - |        - |    -
```

gcmoreira:
Good idea @forensicxlab, however, IMO in `LinuxUtilities`  is not the best place to do this.

It would be better if you add the getters, as get_* or as a property like below, in the `inode` (extension) object. 

The `inode` extension object hasn't been created yet, but it should be like the code below. 

volatility3/framework/symbols/linux/extensions/__init__.py
```python
class inode(objects.StructType):
...
    @property
    def mode(self) -> str:
        return stat.filemode(self.i_mode)
```

Check  the LinuxKernelIntermedSymbols class in `volatility3/framework/symbols/linux/__init__.py`, you need to add:
```python
class LinuxKernelIntermedSymbols(intermed.IntermediateSymbolTable):
...
        self.set_type_class("inode", extensions.inode)
```

That way, you will be able to do this:
```python
>>> inode
<inode symbol_table_name1!inode: layer_name @ 0x9a460b483538 #632>

>>> inode.mode
'-rw-rw-r--'
```
Also, this approach reduces the number of variables/fields you need to manage in your code, requiring only to obtain the `inode` object.

Now, regarding the inode timestamps.
1) Your code is missing the nsecs. This is very important for timelines.
2) We should use `aware` datetimes to represent times in UTC.

We have a function in `conversion::unixtime_to_datetime()` [here](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/renderers/conversion.py#L28) which does (2), but you still need to include the nsecs (1) to the datetime.

@ikelos: Please see the warning in [datetime.utcfromtimestamp()](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp). I think we should update `unixtime_to_datetime()` and `wintime_to_datetime()` to use `datetime.fromtimestamp(timestamp, tz=timezone.utc)` instead.

To sum up, I think the best approach is to include the datetime conversion in `timespec64` as an extension object.

EDIT: Actually, it's a little bit more complex, since before kernels 4.18 the inode member types for i_[amc]time were `timespec`. Fortunately, the implementation remains consistent for both cases, we can fix this with a simple hack.
```python
class timespec64(objects.StructType):
    def to_datetime(self) -> datetime:
        """Returns the respective aware datetime"""

        dt = conversion.unixtime_to_datetime(self.tv_sec + self.tv_nsec / 1e9)
        return dt
```
in the LinuxKernelIntermedSymbols class, both types pointing to the same extension class.
```
        self.optional_set_type_class("timespec64", extensions.timespec64)
        self.optional_set_type_class("timespec", extensions.timespec64)
```
Thus, the `inode` type is not the only one that will benefit from this, but all kernel types using `timespec64` will.

```python

# kernel 6.2.0
>>> inode.i_ctime
<timespec64 symbol_table_name1!timespec64 (.i_ctime): layer_name @ 0x9a460b4835b0 #16>

>>> inode.i_ctime.to_datetime()
datetime.datetime(2023, 8, 4, 11, 19, 2, 308002, tzinfo=datetime.timezone.utc)

# kernel 4.4.0
>>> inode.i_ctime
<timespec64 symbol_table_name1!timespec (.i_ctime): layer_name @ 0x880035a28590 #16>

>>> inode.i_ctime.to_datetime()
datetime.datetime(2020, 8, 8, 3, 4, 11, 130266, tzinfo=datetime.timezone.utc)
```
Alternatively, you can also add a wrapper/helper in a inode extension, like this:
```python
class inode(objects.StructType):
    @property
    def change_time(self) -> datetime:
        return self.i_ctime.to_datetime()
```
```python
>>> inode.change_time
datetime.datetime(2023, 8, 4, 11, 19, 2, 308002, tzinfo=datetime.timezone.utc)

>>> str(inode.change_time)
'2023-08-04 11:19:02.308002+00:00'
```

forensicxlab:
Thank you @gcmoreira for taking the time to review this PR!

This approach looks indeed very solid. I'll update the code accordingly in the coming days :) 

Regards

gcmoreira:
@forensicxlab It’s up to you whether you want to include the following in the current effort but I noticed there are some stuff missing in this plugin:
- Add the device/superblock major:minor column. This is important to understand where the inode lives.
- Include threads in the task loop (and the respective TID column). For this we should call `list_tasks(..., include_threads=True)` . Threads may or may not share the file descriptor table with the thread group leader, depending on whether the CLONE_FILES flag is included in the clone() syscall. Also, once started, a thread can unshare the fd table. Refer to the unshare() libc syscall wrapper man page, unshare(2).

Again, it's up to you, otherwise I can also add this in a later PR.

gcmoreira:
@ikelos https://github.com/volatilityfoundation/volatility3/pull/1222 fixes the issue I mentioned above plus other issues with datetime exceptions.

@forensicxlab I edited the `timespec64` implementation above as I realized it's not needed to call timedelta and add microseconds to **unixtime_to_datetime**.

forensicxlab:
Hello, @gcmoreira, 
Challenge accepted. 
I'll may ask for you advise from time to time as the implementation goes forward. 

Thank you for the PR fixing the datetime issues :) 

Kind regards.

gcmoreira:
@forensicxlab  bad news,  please check these:
v6.6 https://github.com/torvalds/linux/commit/13bc24457850583a2e7203ded05b7209ab4bc5ef
v6.7 https://github.com/torvalds/linux/commit/12cd44023651666bd44baa36a5c999698890debb
v6.11 https://github.com/torvalds/linux/commit/3aa63a569c64e708df547a8913c84e64a06e7853

Leave it to me, as I'm already implementing this for a PR I'm currently working on. I will push this soon.

forensicxlab:
@gcmoreira Alright no problem. Will your PR include the enhancement to the LSOF plugin ? 

Regards.

gcmoreira:
@forensicxlab that PR will include the changes in the inode extension object from where you can take the datetimes for this PR.

Alternatively, and this is up to @ikelos, we could merge this PR as is, knowing it won't work for +6.6, and incorporate the other changes in a future PR. Given the likely delay between the review and fix cycle and the final merge, this approach might be more efficient

ikelos:
Ideally I'd prefer to review/include finished code that works correctly?  If the other changes get reviewed and mention this then re-reviewing it shouldn't take too long.

forensicxlab:
Hello @ikelos, @gcmoreira 
I don't mind waiting for the inode extension PR to be merged. 
Then, I'll update this one with the appropriate modifications so we can have a nice/clean code to review before merging anything that might transform into a raised issue later.

Good luck and day to you both :) 

Regards.

gcmoreira:
@forensicxlab Whenever you get a chance, could you please add the changes from https://github.com/volatilityfoundation/volatility3/pull/1230 to your branch and make sure everything works? Cheers

gcmoreira:
@forensicxlab https://github.com/volatilityfoundation/volatility3/pull/1230 was just merged, have fun ;)!

forensicxlab:
Hello @gcmoreira, @ikelos,

First of all, thank you for this merge and quick updates.

I have updated the code. 

Sample output:
```
Volatility 3 Framework 2.8.0
  |  PID |         Process |  FD |               Path |      Inode | Type |       Mode |                        Changed |                       Modified |                       Accessed |     Size
* |    1 |         systemd |   0 |          /dev/null |          4 |  CHR | crw-rw-rw- | 2022-09-02 12:19:51.158540 UTC | 2022-09-02 12:19:51.158540 UTC | 2022-09-02 12:19:51.158540 UTC |        0
* |    1 |         systemd |   1 |          /dev/null |          4 |  CHR | crw-rw-rw- | 2022-09-02 12:19:51.158540 UTC | 2022-09-02 12:19:51.158540 UTC | 2022-09-02 12:19:51.158540 UTC |        0
* |    1 |         systemd |   2 |          /dev/null |          4 |  CHR | crw-rw-rw- | 2022-09-02 12:19:51.158540 UTC | 2022-09-02 12:19:51.158540 UTC | 2022-09-02 12:19:51.158540 UTC |        0
* |    1 |         systemd |   3 |          /dev/kmsg |         10 |  CHR | crw-r--r-- | 2022-09-02 12:19:51.158540 UTC | 2022-09-02 12:19:51.158540 UTC | 2022-09-02 12:19:51.158540 UTC |        0
* |    1 |         systemd |   4 |  anon_inode:[8960] |       8960 |    ? | ?rw------- | 2022-09-02 12:19:45.560000 UTC | 2022-09-02 12:19:45.560000 UTC | 2022-09-02 12:19:45.560000 UTC |        0
* |    1 |         systemd |   5 |  anon_inode:[8960] |       8960 |    ? | ?rw------- | 2022-09-02 12:19:45.560000 UTC | 2022-09-02 12:19:45.560000 UTC | 2022-09-02 12:19:45.560000 UTC |        0
* |    1 |         systemd |   6 |  anon_inode:[8960] |       8960 |    ? | ?rw------- | 2022-09-02 12:19:45.560000 UTC | 2022-09-02 12:19:45.560000 UTC | 2022-09-02 12:19:45.560000 UTC |        0
* |    1 |         systemd |   7 |     /sys/fs/cgroup |          1 |  DIR | dr-xr-xr-x | 2022-09-02 12:37:03.039364 UTC | 2022-09-02 12:37:03.039364 UTC | 2022-09-02 12:19:49.296726 UTC |        0
```

I feel like leaving the get_inode_metadata in LinuxUtilities is cleaner. Maybe other plugins could use it in the future ? 

Let me know what you think and if I should not use the LinuxUtilities to extract the inode metadata, performing this operation directly in the plugin as a simple function.

Kind regards.

gcmoreira:
@forensicxlab sorry, but I insist it shouldn't be in `LinuxUtilities`, plus even if you move that function to your code, you need to fix a couple of things there and in the code using that function.
- Not sure why you are doing that but there's no reason for `get_inode_metadata()` to be a generator, plus what your are doing with `next()` in the plugin.
- You need to be more specific with the exceptions you catch; using 'except Exception as e:' isn't good practice. Your linter should already be warning you about this. 
- You should use `or renderers.NotAvailableValue()` instead of `or "?"`
- Lastly, in the case you decide to move that function to your plugin and keep that as a public class method (which is debatable), you need to add the typing info to the return value and complete the docstring.

Apart from that, I think you should split the fd information and the inode information in two functions/generators. I would keep the `list_fds()` function as it was in the previous plugin and on top of that add your code. That will make your changes clearer and easier to review and merge.
In _generator(), you will end up two nested 'for' statements, the outer loop with the list_fds() and in inner loop the new code doing the inode stuff, and from here you will yield the fields. 
You merged everything in one function and that's why you ended up with a generator and `for` statement with too many fields. It's always better practice to have smaller functions just doing specific tasks and not the other way around.

forensicxlab:
Thank you for your guidance, I'll update the implementation. 

Regards.

forensicxlab:
Hi @gcmoreira, 

I addressed the modification based on your guidance. Hope it's better and that you'll find this acceptable :) 

I made the choice to move the  `get_inode_metadata` function to the plugin while we don't have any other use-case that justify putting it into the LinuxUtilities. 

Thanks.

Kind regards.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -67,7 +67,7 @@ def __init__(self, *args, **kwargs) -> None:
 class LinuxUtilities(interfaces.configuration.VersionableInterface):
     """Class with multiple useful linux functions."""
 
-    _version = (2, 2, 0)
+    _version = (2, 1, 0)
```
ikelos: The version should go up, not down and given you're removing an API (get_inode_metadata), it will mean the version needs to go to `(3,0,0)` and will break all the existing linux plugins.  Might be worth leaving it, but including a `DecprecationException` so people know not to use it in the future?

volatility3/framework/plugins/linux/lsof.py
```diff
@@ -45,14 +45,37 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+                        itype,
+                        inode_object.i_size,
+                        inode_object.get_file_mode(),
+                        inode_object.get_change_time(),
+                        inode_object.get_modification_time(),
+                        inode_object.get_access_time(),
+                    )
+        except (exceptions.InvalidAddressException, AttributeError) as e:
+            vollog.warning(f"Can't get inode metadata: {e}")
+        return tuple(renderers.NotAvailableValue() for _ in range(7))
```
ikelos: This doesn't seem worthwhile, it doesn't contain an offset, so there's literally no information in the result.  Feels like it would be better to return `None` and check the result when called.

volatility3/framework/plugins/linux/lsof.py
```diff
@@ -69,21 +92,75 @@ def list_fds(
...
+    def list_fds_and_inodes(
+        cls,
+        context: interfaces.context.ContextInterface,
+        symbol_table: str,
+        filter_func: Callable[[int], bool] = lambda _: False,
+    ):
+        for pid, task_comm, task, (fd_num, filp, full_path) in cls.list_fds(
+            context, symbol_table, filter_func
+        ):
+            inode_metadata = cls.get_inode_metadata(filp)
```
ikelos, line 105: So I'd do a check for `None` here and then populate it with AbsentValues.  The function may get called as part of other computations, so returning None is easier to work with than a tuple of AbsentValues (which are really only supposed to be for the UI to know not to render anything, rather than to be passed around or used for computations).

ikelos (commented):
Just very minor changes from me, it's looking much better now, thanks!  5:)

volatility3/framework/symbols/linux/__init__.py
ikelos: This was only in one commit in the PR, not the PR as a whole, sorry for the noise.

forensicxlab:
Implemented the changes, should be good to go 👍 

Many thanks to you three for the review :)

Regards

volatility3/framework/plugins/linux/lsof.py
```diff
@@ -1,27 +1,27 @@
...
 
 
-class Lsof(plugins.PluginInterface):
+class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
     """Lists all memory maps for all processes."""
 
     _required_framework_version = (2, 0, 0)
-
-    _version = (1, 1, 0)
+    _version = (2, 0, 0)
```
ikelos: Linux.sockstat uses `lsof` 1.1.0.  Please could you verify that sockscan works with your changes and bump the lsof dependency version for it in this PR please?

ikelos: So yeah, this might not *actually* require a MAJOR version bump, only a MINOR one (so `(1,2,0)`), since the currently exposed API methods stay the same, it's just the `_generator` that changes.  In that case you won't need to update `sockstat`, otherwise you will.  It's your call which option you choose, but generally not bumping a MAJOR version without reason breaks fewer people's plugins...  5;)

Then it looks ready to go in.  5:)

volatility3/framework/plugins/linux/lsof.py
```diff
@@ -1,27 +1,27 @@
...
 from volatility3.framework.symbols import linux
 from volatility3.plugins.linux import pslist
+from volatility3.plugins import timeliner
 
 vollog = logging.getLogger(__name__)
 
 
-class Lsof(plugins.PluginInterface):
+class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
     """Lists all memory maps for all processes."""
```
gcmoreira: Oops! Could you do us a favor and update the plugin's docstring?

forensicxlab:
Hello, 
Should be ok now :) 

Regards.

Issue #1146, opened 2024-05-09
Handles unable to process sample Volatility 2 reports correctly. UnionType error?


atcuno:
**Describe the bug**
Running windows.handles on a memory sample that Volatility 2 supports fully causes a strange backtrace in Vol3:

```
# python3.8 vol.py -f data.lime windows.handles --pid 3704
Volatility 3 Framework 2.7.0
Progress:  100.00		PDB scanning finished                        
PID	Process	Offset	HandleValue	Type	GrantedAccess	Name
Traceback (most recent call last):
  File "/home/analyst/vol3/volatility3/framework/plugins/windows/handles.py", line 91, in _get_item
    if not self.context.layers[virtual].is_valid(handle_table_entry.Object):
  File "/home/analyst/vol3/volatility3/framework/objects/__init__.py", line 971, in __getattr__
    raise AttributeError(
AttributeError: UnionType has no attribute: symbol_table_name1!_HANDLE_TABLE_ENTRY.Object

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "vol.py", line 10, in <module>
    volatility3.cli.main()
  File "/home/analyst/vol3/volatility3/cli/__init__.py", line 871, in main
    CommandLine().run()
  File "/home/analyst/vol3/volatility3/cli/__init__.py", line 469, in run
    renderer.render(grid)
  File "/home/analyst/vol3/volatility3/cli/text_renderer.py", line 198, in render
    grid.populate(visitor, outfd)
  File "/home/analyst/vol3/volatility3/framework/renderers/__init__.py", line 245, in populate
    for level, item in self._generator:
  File "/home/analyst/vol3/volatility3/framework/plugins/windows/handles.py", line 378, in _generator
    for entry in self.handles(object_table):
  File "/home/analyst/vol3/volatility3/framework/plugins/windows/handles.py", line 348, in handles
    for handle_table_entry in self._make_handle_array(TableCode, table_levels):
  File "/home/analyst/vol3/volatility3/framework/plugins/windows/handles.py", line 311, in _make_handle_array
    for x in self._make_handle_array(entry, level - 1, depth):
  File "/home/analyst/vol3/volatility3/framework/plugins/windows/handles.py", line 323, in _make_handle_array
    item = self._get_item(entry, handle_value)
  File "/home/analyst/vol3/volatility3/framework/plugins/windows/handles.py", line 111, in _get_item
    raise AttributeError(
[...]

PR #1214, opened 2024-07-24
Windows: update handles plugin to use a default SAR value of 0x10 

eve-mem:
Hello 👋 

This should close off https://github.com/volatilityfoundation/volatility3/issues/1146 

It adds returning a default value of 0x10 when attempting to locate a SAR value for decoding pointers. When this does happen a warning is produced to let the user know. 

This helps resolve a problem when the sar value cannot be calculated, and allows the plugin to continue normally in most cases. I believe in the past vol2 just used a hard coded value, so these changes will help the plugin stay at that same level if there are problems reading the sar value. If there are no problems reading the disassembly then the plugin still uses those values which will be more accurate.   

@atcuno hope this fixes the problem for good now. (At least until the default changes...!)

🦊

### DIFF
--- a/volatility3/framework/plugins/windows/handles.py
+++ b/volatility3/framework/plugins/windows/handles.py
@@ -142,6 +142,7 @@ class Handles(interfaces.plugins.PluginInterface):
         pointers in the _HANDLE_TABLE_ENTRY which allows us to find the
         associated _OBJECT_HEADER.
         """
+        DEFAULT_SAR_VALUE = 0x10  # to be used only when decoding fails
 
         if self._sar_value is None:
             if not has_capstone:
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/handles.py
```diff
@@ -25,7 +25,7 @@ class Handles(interfaces.plugins.PluginInterface):
     """Lists process open handles."""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 0, 1)
```
ikelos, line 28: Hmmmm, this is a tricky one.  It does technically alter the output of `find_sar_value`, but that isn't directly expose, so I think this does only class as an internal change and so the version number you've chosen looks right.

ikelos (approved):
Ok, cool, seems reasonable and thought out.  The version number bump looks right, so... think it's good to go?  Ship it!  5:D

Issue #591, opened 2021-11-19
windows.psscan: Error only when using --physical


garanews:
**vol -vvvvvvvvvvvvv -f banking-malware.vmem windows.psscan**
```
Level 6  volatility3.framework.objects: Void size requested
Level 6  volatility3.framework.objects: Void size requested
Level 6  volatility3.framework.objects: Void size requested
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_EPROCESS_QUOTA_BLOCK
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_PAGEFAULT_HISTORY
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_PSP_CPU_QUOTA_APC
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_JOB_ACCESS_STATE
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_ACTIVATION_CONTEXT_DATA
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_FLS_CALLBACK_INFO
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_ASSEMBLY_STORAGE_MAP
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_SCSI_REQUEST_BLOCK

3488    636     WmiPrvSE.exe    0x7cf2fb00      16      329     0       False   2020-11-15 03:24:41.000000      N/A     Disabled
1832    500     dllhost.exe     0x7d203930      21      195     0       False   2020-11-15 03:24:29.000000      N/A     Disabled
3200    500     SDXHelper.exe   0x7d24d2f0      16      302     1       False   2021-02-09 00:51:12.000000      N/A     Disabled
3196    500     SDXHelper.exe   0x7d24da30      16      302     1       False   2021-02-09 00:51:12.000000      N/A     Disabled
3028    360     conhost.exe     0x7d2508c0      2       33      0       False   2021-02-09 00:51:13.000000      N/A     Disabled
2044    500     dllhost.exe     0x7d2735e0      19      207     0       False   2020-11-15 03:24:30.000000      N/A     Disabled
2104    500     VSSVC.exe       0x7d2e5060      6       119     0       False   2020-11-15 03:24:31.000000      N/A     Disabled
[...]

PR #1215, opened 2024-07-25
Windows psscan issue 591 - Fix offsets when --physical is used

eve-mem:
Hello 👋 

This PR aims to resolve issue where @garanews found an issue with windows psscan when using the --physical flag. 

The main issue is covered in my comment here: https://github.com/volatilityfoundation/volatility3/issues/591#issuecomment-1864171385
> It looks like that the windows poolscanner will in some cases (I think modern windows version) scan a 'virtual' layer rather than a 'physical' one. That means when that the objects that get made are created on this virtual layer. However when the scanning happens on a 'physical' the object is built on that physical layer, with it's native layer being correctly set to the virtual one.
> 
> It means if we read out pointers etc from the object the native layer being set correct means we get the right answer.
> 
> However it means that `object.vol.offset` is a reference to the physical layer - and that's why @garanews was having this issue.

I would recommend reading the comments on the issue (https://github.com/volatilityfoundation/volatility3/issues/591) as @ikelos raised some other good points on other ways to resolve this problem. It may be that this approach is not the best. 

Thanks!
🦊


volatility3/framework/plugins/windows/psscan.py
```diff
@@ -261,40 +261,62 @@
...
+            # in 'virtual' memory.
+            if proc.vol.layer_name == kernel.layer_name:
+                # proc is already in a virtual mem, so a new object is not needed. it means
+                # that if physical addresses are requested in the output then proc.vol.offset
+                # cannot be used because it will be virtual, so the mapping is needed.
+                vproc = proc
+                if self.config["physical"]:
+                    # the display should be physical addresses, so proc cannot be used. The
+                    # mappings are needed to find where it would be physically.
+                    offset = (_, _, offset, _, _) = list(
```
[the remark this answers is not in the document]
ikelos: Does this syntax even work?

volatility3/framework/plugins/windows/psscan.py
```diff
@@ -261,40 +261,62 @@ def _generator(self):
...
-                        self.context,
-                        kernel.symbol_table_name,
-                        pe_table_name,
-                        vproc,
-                        self.open,
-                    )
+                #  renderers.UnreadableValue()
+            else:
+                # proc is in virtual mem, so a new object needs to be creatd.
+                vproc = self.virtual_process_from_physical(
```
ikelos, line 293: This is no longer wrapped in a handler for `InvalidAddressExceptions`, will that cause a problem if it goes wrong?

volatility3/framework/plugins/windows/psscan.py
```diff
@@ -261,40 +261,62 @@ def _generator(self):
...
-            if not self.config["physical"]:
-                offset = proc.vol.offset
-            else:
-                (_, _, offset, _, _) = list(
-                    memory.mapping(offset=proc.vol.offset, length=0)
-                )[0]
+            # format offset for display, but catch errors when UnreadableValue
+            # cannot be formatted
+            try:
+                display_offset = format_hints.Hex(offset)
```
ikelos, line 325: We could set the offset to None rather than `UnreadableValue` and then the test will be simpler/more efficient than setting up an exception handler to catch a known/expected exception?

ikelos (commented):
So a optimizations and a syntax issue, but otherwise looks reasonable.  Sorry for the delay in reviewing...

volatility3/framework/plugins/windows/psscan.py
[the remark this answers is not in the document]
eve-mem: I think it did - but it really doesn't make sense :sweat_smile:

volatility3/framework/plugins/windows/psscan.py
eve-mem: Yup - that's a much better way to do it.

eve-mem:
Sigh - My git-fu is failing on this one.

PR #1217, opened 2024-07-25
Windows: update handles plugin sar warnings to use DEFAULT_SAR_VALUE var

eve-mem:
Hello 👋 

I'm very sorry but I missed something from the previous PR https://github.com/volatilityfoundation/volatility3/pull/1214

The warnings don't use the variable DEFAULT_SAR_VALUE when they should. If it's changed in the future the warnings would become out of sync and this could cause some confusion. 

🦊


ikelos (approved):
Ok, good catch, thanks!

PR #1219, opened 2024-07-28
Added psxview plugin

qpalzmz112:
First attempt at contributing a plugin to Volatility3.

### DIFF
--- a/volatility3/framework/plugins/windows/psxview.py
+++ b/volatility3/framework/plugins/windows/psxview.py
@@ -17,17 +17,14 @@ vollog = logging.getLogger(__name__)
 
 
 class PsXView(plugins.PluginInterface):
-    """Lists all processes found via 6 of the methods described in \"The Art of Memory Forensics,\" which may help
+    """Lists all processes found via four of the methods described in \"The Art of Memory Forensics,\" which may help
     identify processes that are trying to hide themselves. I recommend using -r pretty if you are looking at this
     plugin's output in a terminal."""
 
     # I've omitted the desktop thread scanning method because Volatility3 doesn't appear to have the funcitonality
     # which the original plugin used to do it.
 
-    # I don't think it's worth including the sessions method either because both the original psxview plugin
-    # and Volatility3's sessions plugin begin with the list of processes found by PsList.
-    # The original psxview plugin's session code essentially just filters the pslist for processes
-    # whose session ID is not None. I've matched this in my code, but again, it doesn't seem worth including.
+    # The sessions method is omitted because it begins with the list of processes found by Pslist anyway.
 
     # Lastly, I've omitted the pspcid method because I could not for the life of me get it to work. I saved the
     # code I do have from it, and will happily share it if anyone else wants to add it.
@@ -64,7 +61,7 @@ class PsXView(plugins.PluginInterface):
             ),
             requirements.BooleanRequirement(
                 name="physical-offsets",
-                description="List processes with phyiscall offsets instead of virtual offsets.",
+                description="List processes with physical offsets instead of virtual offsets.",
                 optional=True,
             ),
         ]
--- a/volatility3/framework/plugins/windows/psxview.py
+++ b/volatility3/framework/plugins/windows/psxview.py
@@ -92,7 +92,7 @@ class PsXView(plugins.PluginInterface):
         layer_name = kernel.layer_name
 
         try:
-            _, _, offset, _, _ = list(
+            _original_offset, _original_length, offset, _length, _layer_name = list(
                 self.context.layers[layer_name].mapping(offset=offset, length=0)
             )[0]
         except exceptions.PagedInvalidAddressException:
@@ -190,15 +190,15 @@ class PsXView(plugins.PluginInterface):
         # print results
 
         # list of lists of offsets
-        todo_offsets = [list(processes[source].keys()) for source in processes]
+        offsets = [list(processes[source].keys()) for source in processes]
 
         # flatten to one list
-        todo_offsets = sum(todo_offsets, [])
+        offsets = sum(offsets, [])
 
         # remove duplicates
-        todo_offsets = set(todo_offsets)
+        offsets = set(offsets)
 
-        for offset in todo_offsets:
+        for offset in offsets:
             proc = None
 
             in_sources = {src: False for src in processes}
--- a/volatility3/framework/plugins/windows/psxview.py
+++ b/volatility3/framework/plugins/windows/psxview.py
@@ -4,19 +4,28 @@ from volatility3.framework import constants, exceptions
 from volatility3.framework.interfaces import plugins
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints, TreeGrid
-from volatility3.plugins.windows import handles, info, pslist, psscan, sessions, thrdscan
+from volatility3.plugins.windows import (
+    handles,
+    info,
+    pslist,
+    psscan,
+    sessions,
+    thrdscan,
+)
 
 vollog = logging.getLogger(__name__)
 
+
 class PsXView(plugins.PluginInterface):
-    """Lists all processes found via 6 of the methods described in \"The Art of Memory Forensics,\" which may help 
+    """Lists all processes found via 6 of the methods described in \"The Art of Memory Forensics,\" which may help
     identify processes that are trying to hide themselves. I recommend using -r pretty if you are looking at this
     plugin's output in a terminal."""
-    # I've omitted the desktop thread scanning method because Volatility3 doesn't appear to have the funcitonality
-    # which the original plugin to do it. 
 
-    # I don't think it's worth including the sessions method either because both the original psxview plugin 
-    # and Volatility3's sessions plugin begin with the list of processes found by PsList. 
+    # I've omitted the desktop thread scanning method because Volatility3 doesn't appear to have the funcitonality
+    # which the original plugin to do it.
+
+    # I don't think it's worth including the sessions method either because both the original psxview plugin
+    # and Volatility3's sessions plugin begin with the list of processes found by PsList.
     # The original psxview plugin's session code essentially just filters the pslist for processes
     # whose session ID is not None. I've matched this in my code, but again, it doesn't seem worth including.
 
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,279 @@
...
+                name="identify-expected",
+                description='In the plugin\'s output, replace false with \
+                                                normal where false is the expected result for a Windows machine running normally. \
+                                                Keep in mind that this plugin uses simple checks to identify "normal" behavior, \
+                                                so you may want to double-check the legitimacy of these processes yourself.',
+                optional=True,
+            ),
+            requirements.BooleanRequirement(
+                name="physical-offsets",
+                description="List processes with phyiscall offsets instead of virtual offsets.",
```
ikelos: Small typo, `phyiscall` should be `physical`.

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,279 @@
...
+            ),
+        ]
+
+    def proc_name_to_string(self, proc):
+        return proc.ImageFileName.cast(
+            "string", max_length=proc.ImageFileName.vol.count, errors="replace"
+        )
+
+    def is_ascii(self, str):
+        return str.split(".")[0].isalnum()
```
ikelos: It's not clear what happens if the string starts with a `.`, or if there's ascii characters, then a `.` and then non-ascii characters?  This would probably be better as `all([char in string.ascii_lowercase + "." for char in str.lower()])` to make sure the string only container letters you want?

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,279 @@
...
+        if self.config["physical-offsets"]:
+            return offset
+
+        kernel = self.context.modules[self.config["kernel"]]
+        layer_name = kernel.layer_name
+
+        try:
+            offset = list(
+                self.context.layers[layer_name].mapping(offset=offset, length=0)
+            )[0][2]
```
ikelos: This is a little hard for people reading the code to verify, it would be better if you indexed the first element, but unpacked the second (so `_, _, offset, _, _ = list(self.context.layers[layer_name].mapping(offset=offset, length=0))[0]`)

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,279 @@
...
+        except:
+            # already have physical address
+            pass
```
ikelos: I'm not sure that this failing guarantees the offset was physical, I'd probably let the exception bubble up if you're not sure whether you're right or not.

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,279 @@
...
+
+        return offset
+
+    def proc_list_to_dict(self, tasks):
+        return {self.translate_offset(proc.vol.offset): proc for proc in tasks}
```
ikelos: Wow, that's a new one on me, I never tried... dictionary comprehensions?  I've always made a list of tuples and then called dict on it (so `dict([(self.translate_offset(proc.vol.offset), proc) for proc in tasks])`.  This seems to work, but it did take me a minute to read it because it's so unusual...  5:)

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,279 @@
...
+
+                except exceptions.InvalidAddressException:
+                    vollog.log(
+                        constants.LOGLEVEL_VVV, "Cannot access eprocess object table"
+                    )
+
+        ret = self.filter_garbage_procs(ret)
+        return self.proc_list_to_dict(ret)
+
+    def check_session(self, pslist_procs):
```
ikelos: None of these methods are really worth exposing to the outside world, or can be reused easily, so it would be better to prefix them with `_` to show that they're supposed to be private.

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,279 @@
...
+                    name = self.proc_name_to_string(proc)
+
+                    exit_time = proc.get_exit_time()
+                    if type(exit_time) != datetime.datetime:
+                        exit_time = ""
+                    else:
+                        exit_time = str(exit_time)
+
+                    in_sources = {
+                        src: str(offset in processes[src]) for src in processes
```
ikelos: This seems a bit strange, you go through every list, but for each item in every list (if it hasn't been seen) you go through all of the sources again, looking for the offset.  Wouldn't it make sense to gather all the offsets first, and then for each offset, see which sources it's found in?

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,279 @@
...
+                    }
+
+                    if self.config["identify-expected"]:
+                        f = "False"
+                        n = "Normal"
+
+                        if in_sources["pslist"] == f:
```
ikelos: This is checking whether in_sources contains the string "False"?  This is a pretty weird way of testing things, I think I'd feel happier if these could be split out into checks for each process gathering method.

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,279 @@
...
+                            if exit_time != "":
+                                in_sources["pslist"] = n
+
+                        if in_sources["thrdscan"] == f:
+                            if exit_time != "":
+                                in_sources["thrdscan"] = n
+
+                        if in_sources["csrss"] == f:
+                            if name.lower() in ["system", "smss.exe", "csrss.exe"]:
```
ikelos: This hardcoded list of processes that are guaranteed safe is a little worrying...  Why are processes with these names safe, what's to stop malware naming itself this way and getting marked as normal?

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,279 @@
...
+                name="thrdscan", component=thrdscan.ThrdScan, version=(1, 0, 0)
+            ),
+            requirements.VersionRequirement(
+                name="handles", component=handles.Handles, version=(1, 0, 0)
+            ),
+            requirements.VersionRequirement(
+                name="sessions", component=sessions.Sessions, version=(0, 0, 0)
+            ),
+            requirements.BooleanRequirement(
+                name="identify-expected",
```
ikelos: Check that this does get stored in the config with a `-` rather than a `_`

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,279 @@
...
+            requirements.VersionRequirement(
+                name="thrdscan", component=thrdscan.ThrdScan, version=(1, 0, 0)
+            ),
+            requirements.VersionRequirement(
+                name="handles", component=handles.Handles, version=(1, 0, 0)
+            ),
+            requirements.VersionRequirement(
+                name="sessions", component=sessions.Sessions, version=(0, 0, 0)
+            ),
+            requirements.BooleanRequirement(
```
ikelos: How useful is this functionality?  It seems fairly arbitrary and not forensically sound, so I'm not overly keen in including it in the plugin.  I can see this is pretty much copied from the volatility 2 plugin, but it still doesn't sit too well.  Not sure what to do about it though, so for now this is just me getting nervous about it.

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,279 @@
...
+        processes["pslist"] = self.check_pslist(kdbg_list_processes)
+        processes["psscan"] = self.check_psscan(layer_name, symbol_table)
+        processes["thrdscan"] = self.check_thrdscan()
+        processes["csrss"] = self.check_csrss_handles(
+            kdbg_list_processes, layer_name, symbol_table
+        )
+        processes["sessions"] = self.check_session(kdbg_list_processes)
+
+        seen_offsets = set()
+        for source in processes:
```
ikelos: The original plugin goes through the list by offset and then by source.  I think that makes more sense, see more a little later on...

ikelos (commented):
Thanks very much, this looks like it'll be a good addition.  Some things I'm a little nervous about (how we phrase the "False/Normal" stuff) and the logic seems a little inefficient and circuitous (each offset per source, rather than sources per offset, given that's how the output will likely be displayed).  So some points to cover but definitely on the right path...

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: Good point, I'll revisit this. Do you understand/agree with the intent of this code?

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: Got it, will revisit this.

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: It's actually my first time using a dictionary comprehension. Hard to decide when the tradeoff of readability vs conciseness is worth it though

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: I think I see your point. I'll respond to it in another comment getting at the same issue.

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: This is something I didn't feel great about either. I copied it from the original Psxview, but to be honest I don't think the feature should be in the plugin at all, for the same reasons you mentioned - it could easily be misleading.

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: I ran the plugin with the command-line arguments with no issues.

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: I agree about it not being great. My starting point with the plugin was to try and keep as much functionality from the original psxview as possible, which included some of the questionable parts. I think the plugin would be just as good without this functionality.

volatility3/framework/plugins/windows/psxview.py
ikelos: Sure, but I don't think it achieves that intent.

qpalzmz112: Agreed. Just wanted to make sure before I went and changed it

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: I agree. However, if I let the exception bubble up, it will page fault on the memory dumps I'm using to test. The address causing the page faults appears similar to the physical offsets reported by the original Psxview, so I assumed it was already physical. I'm not sure that assumption makes sense, but either way it seems necessary to not let the exception bubble up here. Do you have any other ideas on how to handle it?

ikelos: If you do want to catch it, at the very least there should be a debug log to say that something went wrong and then skip over it (and the exception should be more specific, probably an invalid address exception or something (to avoid catching other problems and ignoring them).

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: I redid this code with the valid characters being uppercase and lowercase letters, period, and space. I can't actually find what characters are valid online though - do you know if I'm missing anything? Also, I incorporated the rest of your feedback into the commit too.

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: I missed this in the commit I just added. Will look at it again

volatility3/framework/plugins/windows/psxview.py
ikelos: This looks like it's still outstanding?

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def _translate_offset(self, offset):
+        if not self.config["physical-offsets"]:
+            return offset
+
+        kernel = self.context.modules[self.config["kernel"]]
+        layer_name = kernel.layer_name
+
+        try:
+            _, _, offset, _, _ = list(
```
ikelos: If possible I prefer to show what the other values to make it easier for people that want to tinker (so `_original_offset, _original_length, offset, _length, _layer_name`, but I'm not gonna be a stickler about it...  5;)

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,254 @@
...
+        # list of lists of offsets
+        todo_offsets = [list(processes[source].keys()) for source in processes]
+
+        # flatten to one list
+        todo_offsets = sum(todo_offsets, [])
+
+        # remove duplicates
+        todo_offsets = set(todo_offsets)
+
+        for offset in todo_offsets:
```
ikelos: Much cleaner, thanks!  5:D

ikelos (commented):
Yep, looks really good now.  One typo still outstanding and an optional change, but after that we can get it merged, thanks!  5:)

volatility3/framework/plugins/windows/psxview.py
ikelos: I don't but I think you've mostly covered what `isalnum()` would've done and we can always supplement it later if it's missing some.

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: ooops my bad

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: :D

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: hmm that is true but since the code only uses offset here I personally prefer to not have extra variables assigned. Especially because people can just go to the mapping definition which shows what the other values are

ikelos: It's not assigning variables since they start with `_` (anymore than just using `_`) it would just save people needing to go elsewhere to get the information, but fair enough, as I said, not a showstopper.

ikelos (approved):
Looks good, thanks!

volatility3/framework/plugins/windows/psxview.py
qpalzmz112: I didn't know about that syntax! Could be worth adding if only to showcase some obscure (only to me?) syntax

ikelos: Well, happy for you to add it before I merge?  I try to make sure it's used in most calls to mapping, just because it returns so many values and they're often discarded.  I'm having difficulty finding the official python reference, but it seems `_` is just a convention and gets treated just like any other variable, but we say that it won't get used (and type hinters know that, allowing the variable to be reused with different types without issue).  As I say, I just prefer having the name so people referring to the code can more clearly see what's going on.

qpalzmz112: Just added it!

PR #1221, opened 2024-07-28
Renderers: Allow BaseAbsentValues in value results

### LINKED ISSUE #1216: format_hints throws TypeError when using UnreadableValue
When format_hints parses as UnreadableValue it can throw a TypeError which stops a plugin continues. 

The expected result is that the format_hints does not cause a crash, but instead allow the UnreadableValue etc through so that they are displayed in the results. 

Here is a volshell example:
```python
(primary) >>> from volatility3.framework.renderers import format_hints, UnreadableValue
(primary) >>> offset = UnreadableValue()
(primary) >>> format_hints.Hex(offset) 
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'UnreadableValue'
(primary) >>> offset = 1
(primary) >>> format_hints.Hex(offset)
1
```

This is also discussed briefly here https://github.com/volatilityfoundation/volatility3/issues/591#issuecomment-2247551726

ikelos:
Fixes #1216

### DIFF
--- a/volatility3/framework/constants/_version.py
+++ b/volatility3/framework/constants/_version.py
@@ -1,10 +1,8 @@
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 2  # Number of releases of the library with a breaking change
-VERSION_MINOR = 7  # Number of changes that only add to the interface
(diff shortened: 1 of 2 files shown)


eve-mem:
This would work for me. 👍

PR #1222, opened 2024-07-29
Renderers time conversion fixes

gcmoreira:
In the context of [volatilityfoundation/volatility3!1213](https://github.com/volatilityfoundation/volatility3/pull/1213) a few issues with time conversion in the framework were identified.

## Aware datetimes
See warning note on https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp:

> Because naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing a specific timestamp in UTC is by calling datetime.fromtimestamp(timestamp, tz=timezone.utc).

Additionaly, `datetime.utcfromtimestamp()` is deprecated since 3.12

## Wrong exceptions

Since version 3.3 **utcfromtimestamp()** and **fromtimestamp()** Python **datetime** module raises `OverflowError` instead of `ValueError`. As of today, Volatility3 requires Python 3.7.3 so we should only include `OverflowError`

### DIFF
--- a/volatility3/framework/plugins/timeliner.py
+++ b/volatility3/framework/plugins/timeliner.py
@@ -105,7 +105,9 @@ class Timeliner(interfaces.plugins.PluginInterface):
         data = item[1]
 
         def sortable(timestamp):
-            max_date = datetime.datetime(day=1, month=12, year=datetime.MAXYEAR)
+            max_date = datetime.datetime(
+                day=1, month=12, year=datetime.MAXYEAR, tzinfo=datetime.timezone.utc
+            )
             if isinstance(timestamp, interfaces.renderers.BaseAbsentValue):
                 return max_date
             return timestamp
--- a/volatility3/framework/renderers/conversion.py
+++ b/volatility3/framework/renderers/conversion.py
@@ -19,7 +19,7 @@ def wintime_to_datetime(
         return renderers.NotApplicableValue()
     unix_time = unix_time - 11644473600
     try:
-        return datetime.datetime.utcfromtimestamp(unix_time)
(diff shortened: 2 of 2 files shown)


gcmoreira:
hm what?
```python
    return datetime.datetime.fromtimestamp(unix_time, datetime.timezone.utc)
ValueError: year -28 is out of range
```
Weird, that is NOT what the [Python 3.7 standard library documentation](https://docs.python.org/3.7/library/datetime.html#datetime.datetime.fromtimestamp) says. It should not raise **ValueError** in Python 3.7.17.

> Changed in version 3.3: Raise [OverflowError](https://docs.python.org/3/library/exceptions.html#OverflowError) instead of [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) if the timestamp is out of the range of values supported by the platform C localtime() or gmtime() functions. Raise [OSError](https://docs.python.org/3/library/exceptions.html#OSError) instead of [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) on localtime() or gmtime() failure.

@ikelos please note that the test cases are using Python 3.7.17, while the GitHub repository specifies Python 3.7.3 or later. Although the difference is minimal, there is a untested gap there, we should consider updating either the test cases interpreter or the repository requirements for consistency. We should also consider running the test cases with both the minimum supported and latest Python versions to detect deprecation and ensure compatibility.

gcmoreira:
Cool, all test cases pass! I've ensured consistency in exception checking for both wintime_to_datetime and unixtime_to_datetime functions.

ikelos (approved):
Looks good, nice catch, thanks!

PR #1224, opened 2024-07-29
Windows: Fixes bad structure member in callbacks

dgmcdona:
This fixes a bug in the x64 callbacks symbols. The `NotificationRoutine`
is currently an `unsigned int` instead of a void pointer. This prevents
the correct mapping of the notification routine to the kernel module
that contains it.

Also bumps the version number for the netstat plugin's `modules`
requirement - it didn't get updated after #1173 was merged.


atcuno:
@ikelos this was found in some large scale testing. Verified as fixed with the patch.

ikelos (approved):
Ok, if you're happy I'm happy

PR #1225, opened 2024-07-29
Orphan threads v2

atcuno:


atcuno:
@ikelos This is v2 of the orphan thread plugin using the new API. I also fixed some bugs that showed up in plugin testing of threads and thrdscan that were not apparent during code review.

The only remaining issue from your first review was the use of hardcoded kernel addresses, but that is still there for now as its not something I have time to fix before Black Hat. Also, the fix will not just be in this plugin, but all the kernel analyzing plugins (callbacks, timers, threads, driverirp, etc.) need a generic fix for this as we have seen some obviously smeared pointers being reported by the plugins in our testing.

So the plugin now is in mergable state and the kernel pointer smear checks will get a generic fix across plugins after Black Hat.

volatility3/framework/plugins/windows/orphan_kernel_threads.py
```diff
@@ -0,0 +1,94 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
```
ikelos: 2024 now I think?

volatility3/framework/plugins/windows/orphan_kernel_threads.py
```diff
@@ -0,0 +1,94 @@
...
+    @classmethod
+    def list_orphan_kernel_threads(
+        cls,
+        context: interfaces.context.ContextInterface,
+        module_name: str,
+    ) -> Generator[interfaces.objects.ObjectInterface, None, None]:
+        """Yields thread objects of kernel threads that do not map to a module
+
+        Args:
+            kernel
```
ikelos: This is not the right "args"...  5:P

volatility3/framework/plugins/windows/orphan_kernel_threads.py
```diff
@@ -0,0 +1,94 @@
...
+        module = context.modules[module_name]
+        layer_name = module.layer_name
+        symbol_table = module.symbol_table_name
+
+        collection = ssdt.SSDT.build_module_collection(
+            context, layer_name, symbol_table
+        )
+
+        # used to filter out smeared pointers
+        if symbols.symbol_table_is_64bit(context, symbol_table):
```
ikelos, line 67: I'll let it go, but please add in a `FIXME` comment, so it doesn't get lost/forgotten...

ikelos (commented):
Cool, fine with that just a couple quick documentation bits and then it can go in...

atcuno:
Feedback addressed.

ikelos (approved):
Looks good, thanks

PR #1227, opened 2024-07-30
Windows: Updates netscan with new symbol file

dgmcdona:
`netscan` was missing coverage for Windows 10 Build 20348, causing
owners and create times for `_TCP_ENDPOINTS` to be missing. This adds a
symbol file and the necessary version check in the netscan plugin.
Testing confirms that this returns the correct creation time and owner
process.


ikelos (approved):
Cool, thanks for testing it out first.  I've no idea whether the symbols are right, but I trust you...  5:)

PR #1229, opened 2024-07-31
Feature/argcomplete

ikelos:
Adds (optional) argcomplete support to volatility3

### DIFF
--- a/volatility3/cli/volargparse.py
+++ b/volatility3/cli/volargparse.py
@@ -99,6 +99,7 @@ class HelpfulArgParser(argparse.ArgumentParser):
         # return the number of arguments matched
         return len(match.group(1))
 
-    def _check_value(self, action, value):
+    def _check_value(self, action: argparse.Action, value: Any) -> None:
         if not isinstance(action, HelpfulSubparserAction):
             return super()._check_value(action, value)
+        return None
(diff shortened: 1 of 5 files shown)


volatility3/cli/volargparse.py
```diff
@@ -100,3 +98,8 @@ def _match_argument(self, action, arg_strings_pattern) -> int:
 
         # return the number of arguments matched
         return len(match.group(1))
+
+    def _check_value(self, action: argparse.Action, value: Any) -> None:
+        if not isinstance(action, HelpfulSubparserAction):
+            return super()._check_value(action, value)
```
gcmoreira: IMO it's not clear what will happen here:
The `_check_value()` function, see [here](https://github.com/python/cpython/blob/main/Lib/argparse.py#L2546), doesn't specify the typing info for the return value. So, to understand what's happening here you have to go to the Python standard library source code.

I would add a comment indicating that an ArgumentError() exception will be raised if the action class is incorrect. Additionally, I would remove the return statement, ensuring that the function always returns None.

```python
    def _check_value(self, action: argparse.Action, value: Any) -> None:
        if not isinstance(action, HelpfulSubparserAction):
-            return super()._check_value(action, value)
+            super()._check_value(action, value)
        return None
```

gcmoreira:
@ikelos it looks good to me, just one minor observation. Thanks a lot for helping me out with this.

ikelos:
Ok, all updated.  Thanks for looking it over!  5:)  If no one else shouts, I'll probably merge it on the weekend...

PR #1230, opened 2024-08-01
Linux: Add inode, timespec, and timespec64 object extensions

gcmoreira:
Add inode, timespec, and timespec64 object extensions to support different kernel versions, ensuring also we will get aware datetimes when using them.

The changes in this PR are part of a larger PR I'm currently working on and are also required to address the issues detailed in https://github.com/volatilityfoundation/volatility3/pull/1213. 
References:
- https://github.com/volatilityfoundation/volatility3/pull/1222
- v6.6 https://github.com/torvalds/linux/commit/13bc24457850583a2e7203ded05b7209ab4bc5ef
- v6.7 https://github.com/torvalds/linux/commit/12cd44023651666bd44baa36a5c999698890debb
- v6.11 https://github.com/torvalds/linux/commit/3aa63a569c64e708df547a8913c84e64a06e7853

### DIFF
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -1839,7 +1839,7 @@ class inode(objects.StructType):
         """Returns the inode number"""
         return int(self.i_ino)
 
-    def ___time_member_to_datetime(self, member) -> datetime:
+    def _time_member_to_datetime(self, member) -> datetime:
         if self.has_member(f"{member}_sec") and self.has_member(f"{member}_nsec"):
             # kernels >= 6.11 it's i_*_sec -> time64_t and i_*_nsec -> u32
             # Ref Linux commit 3aa63a569c64e708df547a8913c84e64a06e7853
@@ -1865,7 +1865,7 @@ class inode(objects.StructType):
         Returns:
             A datetime with the inode's last access time
         """
-        return self.___time_member_to_datetime("i_atime")
+        return self._time_member_to_datetime("i_atime")
 
     def get_modification_time(self) -> datetime:
         """Returns the inode's last modification time
@@ -1875,7 +1875,7 @@ class inode(objects.StructType):
             A datetime with the inode's last data modification time
         """
 
-        return self.___time_member_to_datetime("i_mtime")
+        return self._time_member_to_datetime("i_mtime")
 
     def get_change_time(self) -> datetime:
         """Returns the inode's last change time
@@ -1884,7 +1884,7 @@ class inode(objects.StructType):
         Returns:
             A datetime with the inode's last change time
         """
-        return self.___time_member_to_datetime("i_ctime")
+        return self._time_member_to_datetime("i_ctime")
 
     def get_file_mode(self) -> str:
         """Returns the inode's file mode as string of the form '-rwxrwxrwx'.
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -1780,34 +1780,42 @@ class inode(objects.StructType):
         # pointer, will easily cause an integer overflow here.
         return self.i_ino > 0 and self.i_count.counter >= 0
 
+    @property
     def is_dir(self) -> bool:
         """Returns True if the inode is a directory"""
         return stat.S_ISDIR(self.i_mode) != 0
 
+    @property
     def is_reg(self) -> bool:
         """Returns True if the inode is a regular file"""
         return stat.S_ISREG(self.i_mode) != 0
 
+    @property
     def is_link(self) -> bool:
         """Returns True if the inode is a symlink"""
         return stat.S_ISLNK(self.i_mode) != 0
 
+    @property
     def is_fifo(self) -> bool:
         """Returns True if the inode is a FIFO"""
         return stat.S_ISFIFO(self.i_mode) != 0
 
+    @property
     def is_sock(self) -> bool:
         """Returns True if the inode is a socket"""
         return stat.S_ISSOCK(self.i_mode) != 0
 
+    @property
     def is_block(self) -> bool:
         """Returns True if the inode is a block device"""
         return stat.S_ISBLK(self.i_mode) != 0
 
+    @property
     def is_char(self) -> bool:
         """Returns True if the inode is a char device"""
         return stat.S_ISCHR(self.i_mode) != 0
 
+    @property
     def is_sticky(self) -> bool:
         """Returns True if the sticky bit is set"""
         return (self.i_mode & stat.S_ISVTX) != 0
@@ -1818,19 +1826,19 @@ class inode(objects.StructType):
         Returns:
             The inode type name
         """
-        if self.is_dir():
+        if self.is_dir:
             return "DIR"
-        elif self.is_reg():
+        elif self.is_reg:
             return "REG"
-        elif self.is_link():
+        elif self.is_link:
             return "LNK"
-        elif self.is_fifo():
+        elif self.is_fifo:
             return "FIFO"
-        elif self.is_sock():
+        elif self.is_sock:
             return "SOCK"
-        elif self.is_char():
+        elif self.is_char:
             return "CHR"
-        elif self.is_block():
+        elif self.is_block:
             return "BLK"
         else:
             return None
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -4,10 +4,13 @@
 
 import collections.abc
 import logging
+import stat
+from datetime import datetime
 import socket as socket_module
 from typing import Generator, Iterable, Iterator, Optional, Tuple, List
 
 from volatility3.framework import constants, exceptions, objects, interfaces, symbols
+from volatility3.framework import renderers
 from volatility3.framework.constants.linux import SOCK_TYPES, SOCK_FAMILY
 from volatility3.framework.constants.linux import IP_PROTOCOLS, IPV6_PROTOCOLS
 from volatility3.framework.constants.linux import TCP_STATES, NETLINK_PROTOCOLS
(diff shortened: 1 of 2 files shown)


volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1761,3 +1764,132 @@ def get_capabilities(self) -> int:
...
+class timespec64(objects.StructType):
+    def to_datetime(self) -> datetime:
+        """Returns the respective aware datetime"""
+
+        dt = renderers.conversion.unixtime_to_datetime(self.tv_sec + self.tv_nsec / 1e9)
+        return dt
+
+
+class inode(objects.StructType):
+    def is_valid(self) -> bool:
```
ikelos, line 1778: Would all these be better as properties?  Can't tell, but we kind of had an unofficial rule that additions to a Struct that did stuff started `get_` and otherwise they were made properties.  These looks ideal for making into properties?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1761,3 +1764,132 @@ def get_capabilities(self) -> int:
...
+        elif self.is_fifo():
+            return "FIFO"
+        elif self.is_sock():
+            return "SOCK"
+        elif self.is_char():
+            return "CHR"
+        elif self.is_block():
+            return "BLK"
+        else:
+            return renderers.UnparsableValue()
```
ikelos: This is specificially for the UI, so I'm not keen on them bleeding into the core.  If someone wants to use these types for something completely different, this should be a standard python type like None, and then verified before use to be put into the results (and turned into a `BaseAbsentValue` descendant).

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1761,3 +1764,132 @@ def get_capabilities(self) -> int:
...
+
+    def get_inode_number(self) -> int:
+        """Returns the inode number"""
+        return int(self.i_ino)
```
ikelos: Is this not already an int?  Just wondering why the overly cautious casting?  Is it so it can go straight into the results (a volatility int should still be ok there)?  I prefer not casting unnecessarily because other people see it and then they can unnecessarily because they think they have to, and soon everything is littered with unnecessary casts and it becomes much more difficult to read.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1761,3 +1764,132 @@ def get_capabilities(self) -> int:
...
+
+    def ___time_member_to_datetime(self, member) -> datetime:
```
ikelos: `__` at the start has a specific meaning within python.  Unlike just `_` which informally means private, but actually has no effect, `__` I think prevents inheritance of the class and does alter the way python works.

I don't know if this is to mimic a linux function name and I'm all for making the function names as similar to linux as possible, but when they conflict with the way python functions I don't think we can keep them in sync.  If this isn't supposed to be called outside the class it should only have one `_` (so the class could still be inherited by someone and function) and if it is supposed to be called outside the class, then ideally it wouldn't start with any `_`.

ikelos (commented):
Thanks, mostly looks fine, but one function starts with ` __` which should only be used in very specific circumstances, there's a return value that pulls in pits of the user interface essentially, and should instead by using standard python types, and the non-`get_` functions should be turned into properties please...

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -4,10 +4,13 @@
 
 import collections.abc
 import logging
+import stat
+from datetime import datetime
 import socket as socket_module
 from typing import Generator, Iterable, Iterator, Optional, Tuple, List
 
 from volatility3.framework import constants, exceptions, objects, interfaces, symbols
+from volatility3.framework import renderers
```
ikelos: This shouldn't be necessary (and it's because of trying to return a `BaseAbsentValue` descendant.  Please remove this when you get rid of the absent value return below.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: cool, I will convert those to properties

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Sounds good

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Ok, I will change this

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Yeah, exactly. It's a private function to prevent inheritance and outside use. It's just an internal helper for other methods, so it doesn't make sense to inherit or use it directly from outside. May I leave this like that or still need to be converted to _ ?

ikelos: I'd prefer it as a `_` because if someone wants to extend this class, they will need that function to be inheritted, so fully private makes that much more difficult.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: This is still needed for `renderers.conversion.unixtime_to_datetime()`

gcmoreira: I will import from `volatility3.framework.renderers import conversion` instead

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Specifically for the `is_valid()` method, it's not used as property anywhere in the framework:
```shell
$ grep -r -B1 'def is_valid(' | grep property | wc -l
0

$ grep -r -B1 'def is_valid(' | grep -v property | wc -l
104
```

Do you still want me to make it a property?

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: We probably ought to move that out of renderers, if it's a general purpose function?

ikelos: Maybe, `framework.objects.utility`?

gcmoreira: sure, but can do it in another PR? `wintime_to_datetime()` is used in many Windows related files

ikelos: Hehehe, yeah, no problem (although good if you could include a FIXME line just so we remember please)

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: The problem with returning non-standard types for things like simple numbers is that it messes up serialization and makes deepcopying really really slow. It's hard to spot and fix unless you're a pretty advanced Python developer. IMO it's not good idea to do this but up to you.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Errr, no, that's ok.  I think that's defined as a part of the interface somewhere so it shouldn't need it.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: I try to ensure that all parameters are standard types rather than complex objects for similar thinking, but the point is people should be able to treat this as an int and not care (because its base class *is* `int`).  As to deepcopying, that's usually not a good solution, but one that was necessary when it was used, so I don't see it as a huge reason.  Serialization (particularly if we ever want to parallelize this somehow) is more important but again, the property is getting called, so the value isn't getting stashed and the receiving party can decide what to do with it, but I've seen a lot of example code littered with these stray casts, not because they were concerned about performance, but because they didn't know whether they could just use it or not.  As such, I try hard to weed them out, so it's probably best to be consistent.  If there's an actual use case where it causes a massive speed difference I'll definitely consider it, but without that it should just be returned (although really, at this point, they're just acting as variable renames)?

gcmoreira: Yeah, that's the point. Without the cast that method works only as a more informative/clear name to get the inode number. The whole point of it was to make sure, with the function name and with the typing, that the user wants the inode integer number. At the end of the day, if the user wants the original type for any reason, this doesn't prevent the user of directly calling `inode.i_ino` instead and obtain the objects.Integer.

> people should be able to treat this as an int and not care (because its base class is int).

Hm, inheriting from `int` doesn't necessarily mean you can use them interchangeably everywhere. `int` is an atomic, immutable type, while `objects.Integer` is not. Of course, for basic stuff like doing math operations, comparisons and so it will work like an `int`.
For instance, you will have issues using this type with the standard [dataclasses](https://docs.python.org/3/library/dataclasses.html) when calling [astuple](https://docs.python.org/3/library/dataclasses.html#dataclasses.astuple). It will need to do a deepcopy taking a long time doing that conversion.

Anyway, back to `get_inode_number()` ... Do you still find it odd to have a method that returns an int() cast like that?

volatility3/framework/symbols/linux/extensions/__init__.py

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: done

volatility3/framework/symbols/linux/extensions/__init__.py

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: So possibly I'm a little disappointed that I put all this work into making sure people could use them exactly like ints (because they are), and then users still don't know/trust that it's the case and end up littering the code with `int` or losing valuable information about where the object came from because they threw it away without good reason.

I tested it, you're right, it does take a long time doing `astuple`, but it does complete just fine.  I think it should be the consumer that optimizes rather than producer, because it only needs optimizing for certain very specific use cases and otherwise the origin information may be useful.

Volshell snippet for a linux image:
```
>>> import dataclasses
>>> @dataclasses.dataclass
... class Point:
...      x: int
...      y: int

>>> volpoint = Point(ps()[0].pid, ps()[1].pid)
>>> dataclasses.astuple(volpoint)
(1, 2)
```

So yeah, I still find it a bit odd using it just to cast, and whilst you say people can just go back to the original, if all our examples and code only use the properties, they likely won't go digging to find out the right name for something that's already supposed to be exposed.  If we're putting all those properties in place just to wrap them in `int`, on the off-chance they may one day get deep-copied, I don't think that's a strong enough justification and I'd prefer we just used the original names and scrap those properties (unless they're overly cumbersome to type/unintuitive), because staying close to the source ensures that when people discuss stuff they're not using volatility-decided pseudonyms, they can discuss it with kernel developers and be using the actual struct members so everyone's on the same page.

I think adding this will just muddy people's idea of what goes on, degrade the consistency of the framework and I think it will just ultimately lead to more people wrapping things as ints without knowing why and further diluting the object model.  Having said all that I don't want to fight about this too much, and my will to keep arguing it is diminishing so if you really think it's vital that we provide an int wrapped convenience function because it's a significant issue with the core object model of the framework which will impact a lot of people very often, then I'll relent.

ikelos: I think it may actually just be this one property, for some reason I thought it was a whole slew of them...  That's less trend setting, so I'm not quite as upset letting it slip through, in general though, I tell people to ditch `int` wrappers unless they can come up with a really good reason (and too often when I ask it's been "oh I just thought these wouldn't work" or "oh I copied it from vol 2 where you had to and thought it would be the same").

gcmoreira: Fair enough, I agree that keeping the original names is a good practice. Plus, it's not a big deal to leave it as is and handle the casting on the consumer side. It was good anyway having this chance to discuss and be aware of some side-effects of not using the built-in types. Thanks

gcmoreira: > I think it may actually just be this one property, for some reason I thought it was a whole slew of them... That's less trend setting, so I'm not quite as upset letting it slip through, in general though, I tell people to ditch `int` wrappers unless they can come up with a really good reason (and too often when I ask it's been "oh I just thought these wouldn't work" or "oh I copied it from vol 2 where you had to and thought it would be the same").

Ups, for some reason I missed your last comment when I replied above. 

You first comment was already quite convincing to me :) instead, it seems we convinced each other simultaneously :) LOL

> I think it may actually just be this one **property**

Sorry, I didn't catch that. Currently, the function isn't a property. When you say `property`, do you mean we should convert it to one, or are you just referring to the function?
Do you think it's better to keep the member original name (removing entirely this function) or;  returning an int() with either a function, or a property?

ikelos: Sorry, it starts with get so it can stay as a function (although that seems less convenient to me).  If we convinced each other, then I'd much prefer to lean towards ditching the int cast (and if necessary the whole function) please.

gcmoreira: Great, here we go! Hopefully, I didn't mess anything up. Let's see what the test cases say

ikelos:
Looks good, thanks for all your hard work on it!  5:D

PR #1232, opened 2024-08-01
added psxview plugin

MY7H404:
added psxview plugin for windows

### DIFF
--- a/volatility3/framework/plugins/windows/psxview.py
+++ b/volatility3/framework/plugins/windows/psxview.py
@@ -1,0 +1,158 @@
+import datetime
+from typing import List, Iterable, Set, Dict
+from volatility3.framework import renderers, interfaces
+from volatility3.framework.configuration import requirements
+from volatility3.framework.interfaces import plugins
+from volatility3.plugins.windows import pslist, psscan
+
+class PsXview(plugins.PluginInterface):
+    """Find hidden processes with various process listings"""
+    _required_framework_version = (2, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.TranslationLayerRequirement(name='primary',
+                            description='Memory layer for the kernel',
+                            architectures=["Intel32", "Intel64"]),
+            requirements.SymbolTableRequirement(name="nt_symbols",
+                            description="Windows kernel symbols"),
+        ]
+
+    @classmethod
+    def check_pslist(cls, context: interfaces.context.ContextInterface, layer_name: str, symbol_table: str) -> Set[int]:
+        """Enumerate processes from PsActiveProcessHead"""
+        return {proc.UniqueProcessId for proc in pslist.PsList.list_processes(context, layer_name, symbol_table)}
+
+    @classmethod
+    def check_psscan(cls, context: interfaces.context.ContextInterface, layer_name: str, symbol_table: str) -> Iterable[interfaces.objects.ObjectInterface]:
+        """Enumerate processes with pool tag scanning"""
+        return psscan.PsScan.scan_processes(context, layer_name, symbol_table)
+
+    @classmethod
+    def check_thrdproc(cls, context: interfaces.context.ContextInterface, layer_name: str, symbol_table: str) -> Dict[int, bool]:
+        """Enumerate processes indirectly by ETHREAD scanning"""
+        thrdproc_info = {}
+        processes = pslist.PsList.list_processes(context, layer_name, symbol_table)
+        for proc in processes:
+            try:
+                threads = cls.list_threads(proc)
+                thrdproc_info[proc.UniqueProcessId] = len(threads) > 0
+            except AttributeError:
+                continue
+        return thrdproc_info
+
+    @classmethod
+    def list_threads(cls, proc):
+        """Lists the threads of a process"""
+        threads = []
+        try:
+            thread_list_entry = proc.ThreadListHead.Flink
+            while thread_list_entry != 0 and thread_list_entry != proc.ThreadListHead.vol.offset:
+                thread = thread_list_entry.dereference().cast("_ETHREAD")
+                threads.append(thread)
+                thread_list_entry = thread.ThreadListEntry.Flink
+        except:
+            pass
+        return threads
+
+    def _generator(self, psscan_tasks, pslist_pids, thrdproc_info):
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,158 @@
...
+from volatility3.plugins.windows import pslist, psscan
+
+class PsXview(plugins.PluginInterface):
+    """Find hidden processes with various process listings"""
+    _required_framework_version = (2, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.TranslationLayerRequirement(name='primary',
```
ikelos, line 15: We tend to use a `ModuleRequirement` these days, instead of separate `TranslationLayerRequirement`s and `SymbolTableRequirement`s.

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,158 @@
...
+
+    @classmethod
+    def check_pslist(cls, context: interfaces.context.ContextInterface, layer_name: str, symbol_table: str) -> Set[int]:
+        """Enumerate processes from PsActiveProcessHead"""
+        return {proc.UniqueProcessId for proc in pslist.PsList.list_processes(context, layer_name, symbol_table)}
+
+    @classmethod
+    def check_psscan(cls, context: interfaces.context.ContextInterface, layer_name: str, symbol_table: str) -> Iterable[interfaces.objects.ObjectInterface]:
+        """Enumerate processes with pool tag scanning"""
+        return psscan.PsScan.scan_processes(context, layer_name, symbol_table)
```
ikelos, line 30: When using other plugins, you'd need to have a `PluginRequirement` so that if the other plugin changes this plugin will be able to tell and complain about it.

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,158 @@
...
+        for proc in processes:
+            try:
+                threads = cls.list_threads(proc)
+                thrdproc_info[proc.UniqueProcessId] = len(threads) > 0
+            except AttributeError:
+                continue
+        return thrdproc_info
+
+    @classmethod
+    def list_threads(cls, proc):
```
ikelos, line 46: Methods like this could do with type information to make help editors that can understand such type information and let people know what goes into and comes out of the function.

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,158 @@
...
+                session_id_str = "True"
+            elif session_id == 1:
+                session_id_str = "False"
+            else:
+                session_id_str = "-"
+
+            thrdproc_present = thrdproc_info.get(pid, False)
+
+            name_str = str(task.ImageFileName.cast("string", max_length=task.ImageFileName.vol.count)).strip()
+            name_str = name_str[:20].ljust(20)
```
ikelos, line 79: Please don't add formatting information to the output.  That should be for the UI to determine, and with the CLI, you can use `-r pretty` rather than the default of `-r quick` in order to get more nicely aligned responses.  The reason for this is that there may be a web interface written for volatility and it would now get random spaces in the return value.

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,158 @@
...
+
+            yield (
+                0,
+                [
+                    f"0x{task.vol.offset:08x}",
```
ikelos, line 84: This should be returning an int, as a `format_hints.Hex`.  Again, so that other programs can use the output from plugins without having to parse them.

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,158 @@
...
+                    name_str,
+                    f"{pid:>5}",
```
ikelos, line 86: That goes for all formatting output.  If the value is a number, you can hint to the UI that it's a number using `format_hints` but still allow it to decide how it's rendered (and allow tools that just use the data not to get random characters or strings they can't do calculations on).

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,158 @@
...
+                    pslist_present,
+                    True,
+                    session_id_str,
+                    thrdproc_present,
+                    task.get_exit_time() if task.get_exit_time() else None
+                ]
+            )
+
+        for task in pslist_tasks:
+            pid = int(task.UniqueProcessId)
```
ikelos, line 96: This should need casting to an int, please avoid unnecssary casting.  You probably cast it like this because you saw someone else doing it, which is why I try to strongly advise people to only cast where it's actually necessary.  Volatility numeric types inherit from `int` so it should be possible to treat them the same and do all the same operations with them directly.

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,158 @@
...
+            psscan_present = pid in {int(proc.UniqueProcessId) for proc in psscan_tasks}
+            session_id = task.get_session_id()
+
+            if session_id == 0:
+                session_id_str = "True"
```
ikelos, line 101: Please use a boolean field, rather than string outputs, again to allow other programs to use the results more easily.

volatility3/framework/plugins/windows/psxview.py
```diff
@@ -0,0 +1,158 @@
...
+            elif session_id == 1:
+                session_id_str = "False"
+            else:
+                session_id_str = "-"
```
ikelos, line 105: For values that aren't present, please use one of the classes descended from `BaseAbsentValue`.  This allows other tools that may use the output to tell the difference between a string `-` and a value that was actually missing.  It also makes it much easier to keep output consistent between plugins.

ikelos (commented):
Thanks very much for your submission.  #1219 is just about to get merged having been through the review process earlier, but I still reviewed your code in case you were interested in submitting more plugins in the future.  There's quite a few places where you try to force the output to look a particular way, which will actually cause problems for other programs that might want to use volatility plugins without displaying them on a command line.  There are ways to get the CLI to produce prettier output, but turning numbers into strings and padding them just makes it more difficult for any of those other tools to do what they need to.  Hopefully these comments till help with your next submission.  I'm afraid I'm going to close this pull request once #1219 lands.

MY7H404:
Thank you for the suggestions and for reviewing my code. It really helped me understand the Volatility framework in depth. I appreciate it!

PR #1233, opened 2024-08-03
Linux Page Cache and IDR

gcmoreira:
This PR includes abstractions for RadixTree, XArray, Page Cache and IDR to ensure support with both older and the newest kernel versions. The RadixTree and XArray enable the implementation of other subsystems like the Page Cache and IDR.

This is the final iteration of other PRs which are part of this feature. 
- https://github.com/volatilityfoundation/volatility3/pull/1222
- https://github.com/volatilityfoundation/volatility3/pull/1230

Besides the Page Cache plugins, we added two more plugins to this PR to be able to test the whole code with different kernel versions.

# Inode Page Cache
## linux.pagecache.Files plugin 
This plugin can list all inodes, filter by type and find filenames. Basically, it shows superblock, dentry and inode info.
It also supports the Timeliner plugin.

For instance, to list all inodes:
```shell
$ ./vol.py -f ./ubuntu-4.9.0-19-32bit.core linux.pagecache.Files 
Volatility 3 Framework 2.7.1
SuperblockAddr  MountPoint      Device  InodeNum        InodeAddr       FileType        InodePages      CachedPages     FileMode        AccessTime      ModificationTime        ChangeTime      FilePath

0xf7094800      /       0:1     1       0xf4801e60      DIR     0       0       drwxr-xr-x      2024-07-29 13:44:58.916000 UTC  2024-07-29 13:44:58.916000 UTC  2024-07-29 13:44:58.916000 UTC  /
0xf7094800      /       0:1     7626    0xf4b2a5c0      DIR     0       0       drwx------      2022-06-30 18:26:10.000000 UTC  2022-06-30 18:26:10.000000 UTC  2024-07-29 13:44:58.228996 UTC  /root
0xf7154800      /sys    0:16    1       0xf4b492e0      DIR     0       0       dr-xr-xr-x      2024-07-29 13:44:58.500000 UTC  2024-07-29 13:44:58.500000 UTC  2024-07-29 13:44:58.500000 UTC  /sys
...
```
Filtering by file type. For instance, list the top Regular files with more pages in cache:
```shell
$ ./vol.py -f ./ubuntu-4.9.0-19-32bit.core linux.pagecache.Files --type REG | sort -k8 -n -r | head -10
0xf6845800      /run    0:18    9964    0xf6fd30b0      REG     647     647     -rw-r-----      2024-07-29 13:45:01.348000 UTC  2024-07-29 13:45:02.044000 UTC  2024-07-29 13:45:02.044000 UTC  /run/log/journal/791e6a5cbadf47da9808b2fdd26a7bf0/system.journal
0xf6f7b800      /       254:3   5506492 0xf46d5780      REG     549     474     -rwxr-xr-x      2024-07-29 13:00:30.392000 UTC  2020-06-20 17:33:46.000000 UTC  2024-01-07 10:46:57.080673 UTC  /usr/bin/perl
0xf6f7b800      /       254:3   2359824 0xf4536a08      REG     517     440     -rw-r--r--      2024-07-29 13:00:29.524000 UTC  2022-06-29 12:41:17.000000 UTC  2024-01-07 10:56:29.475108 UTC  /lib/systemd/libsystemd-shared-232.so
...
```

Find a specific file name:
```shell
$ ./vol.py -f ./ubuntu-4.9.0-19-32bit.core linux.pagecache.Files  --find /etc/shadow
Volatility 3 Framework 2.7.1
SuperblockAddr  MountPoint      Device  InodeNum        InodeAddr       FileType        InodePages      CachedPages     FileMode        AccessTime      ModificationTime        ChangeTime      FilePath

0xf6f7b800      /       254:3   7078755 0xf46c83f8      REG     1       1       -rw-r-----      2024-07-29 13:00:30.552000 UTC  2024-01-07 11:19:42.597348 UTC  2024-01-07 11:19:42.597348 UTC  /etc/shadow

```
## linux.pagecache.InodePages plugin 

This plugin requires either an inode address (--inode) or a filename (--find) to search for the inode, and optionally, you can use (--dump) to dump the inode's page cache contents which is the primary and greatest goal of this effort.

Example 1, via inode address:
```shell
$ ./vol.py -f ./ubuntu-4.9.0-19-32bit.core linux.pagecache.InodePages --inode 0xf6fd30b0 --dump system.journal
Volatility 3 Framework 2.7.1
PageVAddr       PagePAddr       MappingAddr     Index   DumpSafe        Flags
 
0xf5e9be90      0x7bf84000      0xf6fd31a8      0       True    active,dirty,lru,referenced,savepinned,swapbacked,uptodate
0xf5e9beb4      0x7bf85000      0xf6fd31a8      1       True    active,dirty,lru,referenced,savepinned,swapbacked,uptodate
0xf5e9bed8      0x7bf86000      0xf6fd31a8      2       True    active,dirty,lru,referenced,savepinned,swapbacked,uptodate
...
0xf5e8f944      0x7ba09000      0xf6fd31a8      646     True    dirty,lru,savepinned,swapbacked
```

Same inode but finding it via its filename:
```shell
$ ./vol.py -f ./ubuntu-4.9.0-19-32bit.core linux.pagecache.InodePages --find /run/log/journal/791e6a5cbadf47da9808b2fdd26a7bf0/system.journal --dump system.journal
Volatility 3 Framework 2.7.1
PageVAddr       PagePAddr       MappingAddr     Index   DumpSafe        Flags
 
0xf5e9be90      0x7bf84000      0xf6fd31a8      0       True    active,dirty,lru,referenced,savepinned,swapbacked,uptodate
0xf5e9beb4      0x7bf85000      0xf6fd31a8      1       True    active,dirty,lru,referenced,savepinned,swapbacked,uptodate
0xf5e9bed8      0x7bf86000      0xf6fd31a8      2       True    active,dirty,lru,referenced,savepinned,swapbacked,uptodate
...
0xf5e8f944      0x7ba09000      0xf6fd31a8      646     True    dirty,lru,savepinned,swapbacked
```
Checking the results:
```shell
$ file system.journal 
system.journal: Journal file, Mon Jul 29 13:44:59 2024, online, compressed lz4, header size 0xf0, entries 0x23f

$ journalctl --file=system.journal --list-boots
IDX BOOT ID                          FIRST ENTRY                  LAST ENTRY                  
  0 4f2a7426dab64f4dafa29343d13c1483 Mon 2024-07-29 23:44:59 AEST Mon 2024-07-29 23:45:01 AEST

$ journalctl --file=system.journal 
Jul 29 23:44:59 debian9.localdomain kernel: Linux version 4.9.0-19-686 (debian-kernel@lists.debian.org) (gcc version 6.3.0 20170516 (Debian 6.3.0-18+deb9u1) ) #1 SMP Debian 4.9.320-2 (2022-06-30)
Jul 29 23:44:59 debian9.localdomain kernel: x86/fpu: Supporting XSAVE feature 0x001: 'x87 floating point registers'
Jul 29 23:44:59 debian9.localdomain kernel: x86/fpu: Supporting XSAVE feature 0x002: 'SSE registers'
Jul 29 23:44:59 debian9.localdomain kernel: x86/fpu: Supporting XSAVE feature 0x004: 'AVX registers'
...
```
Example 2:
```shell
$ ./vol.py ./ubuntu-4.9.0-19-686.core linux.pagecache.Files --find /etc/passwd
Volatility 3 Framework 2.7.1        
SuperblockAddr  MountPoint      Device  InodeNum        InodeAddr       FileType        InodePages      CachedPages     FileMode        AccessTime      ModificationTime        ChangeTime      FilePath

0xf6f7b800      /       254:3   7078938 0xf45116c0      REG     1       1       -rw-r--r--      2024-07-29 13:00:30.368000 UTC  2024-01-07 11:07:28.924695 UTC  2024-01-07 11:07:28.924695 UTC  /etc/passwd
```
Dump the inode:
```shell
$ ./vol.py -f ./ubuntu-4.9.0-19-686.core linux.pagecache.InodePages --find /etc/passwd --dump passwd
Volatility 3 Framework 2.7.1
PageVAddr       PagePAddr       MappingAddr     Index   DumpSafe        Flags

0xf5e87b68      0x7b68a000      0xf45117b8      0       True    active,lru,mappedtodisk,referenced,uptodate
```

Check the file:
```shell
$ cat passwd
root:x:0:0:root:/root:/bin/bash
daemon:x:1:1:daemon:/usr/sbin:/usr/sbin/nologin
bin:x:2:2:bin:/bin:/usr/sbin/nologin
sys:x:3:3:sys:/dev:/usr/sbin/nologin
sync:x:4:65534:sync:/bin:/bin/sync
games:x:5:60:games:/usr/games:/usr/sbin/nologin
man:x:6:12:man:/var/cache/man:/usr/sbin/nologi
...
```

# IDR
## linux.ebpf.EBPF
A basic plugin to enumerate eBPF programs using the prog IDR. Tip for devs: Other contributions using the maps, links, etc. IDRs and even extending this plugin are welcome ;)

Test the plugin on a Cilium installation.
```shell
$ ./vol.py -f ./ram-5.8.0-53-ubuntu20.04-64bit.lime_cilium linux.ebpf.EBPF 
Volatility 3 Framework 2.8.0
Address Name    Tag     Type

0xc9000002f000  -       3b185187f1855c4c        BPF_PROG_TYPE_XDP
0xc90000031000  BEGIN   f4b3d30cab782162        BPF_PROG_TYPE_KPROBE
0xc90000067000  sys_enter_execv e4476dee689a817e        BPF_PROG_TYPE_TRACEPOINT
0xc90000063000  -       3b185187f1855c4c        BPF_PROG_TYPE_XDP
0xc900005d5000  xdp_prog_simple 3b185187f1855c4c        BPF_PROG_TYPE_XDP
0xc90000141000  -       59904229c5a1f55d        BPF_PROG_TYPE_SCHED_CLS
0xc90000559000  -       6deef7357e7b4530        BPF_PROG_TYPE_CGROUP_SKB
0xc90000409000  -       6deef7357e7b4530        BPF_PROG_TYPE_CGROUP_SKB
0xc9000079d000  -       6deef7357e7b4530        BPF_PROG_TYPE_CGROUP_SKB
0xc9000079b000  -       6deef7357e7b4530        BPF_PROG_TYPE_CGROUP_SKB
...
```
## linux.pidhashtable.PIDHashTable
This is based on the vol2 plugin, removing ancient kernel support, curating code and enhancing comments, while using the new IDR abstraction included also in this effort.

```shell
$ ./vol.py -f ./ram-5.8.0-53-ubuntu20.04-64bit.lime_cilium linux.pidhashtable.PIDHashTable --decorate-comm 
Volatility 3 Framework 2.8.0
OFFSET  PID     TID     PPID    COMM

0x8881396117c0  1       1       0       systemd
0x888139615f00  2       2       0       [kthreadd]
0x888139614740  3       3       2       [rcu_gp]
0x888139610000  4       4       2       [rcu_par_gp]
0x8881396317c0  6       6       2       [kworker/0:0H]
0x888139630000  9       9       2       [mm_percpu_wq]
0x888139632f80  10      10      2       [ksoftirqd/0]
0x8881396397c0  11      11      2       [rcu_sched]
...
```

### DIFF
--- a/volatility3/framework/plugins/linux/ebpf.py
+++ b/volatility3/framework/plugins/linux/ebpf.py
@@ -29,15 +29,21 @@ class EBPF(plugins.PluginInterface):
             ),
         ]
 
-    def get_ebpf_programs(self, vmlinux) -> interfaces.objects.ObjectInterface:
+    def get_ebpf_programs(
+        self,
+        context: interfaces.context.ContextInterface,
+        vmlinux_module_name: str,
+    ) -> interfaces.objects.ObjectInterface:
         """Enumerate eBPF programs walking its IDR.
 
         Args:
-            vmlinux: The kernel symbols object
-
+            context: The context to retrieve required elements (layers, symbol tables) from
+            vmlinux_module_name: The name of the kernel module on which to operate
         Yields:
             eBPF program objects
         """
+        vmlinux = context.modules[vmlinux_module_name]
+
         if not vmlinux.has_symbol("prog_idr"):
             raise exceptions.VolatilityException(
                 "Cannot find the eBPF prog idr. Unsupported kernel"
@@ -49,8 +55,7 @@ class EBPF(plugins.PluginInterface):
             yield bpf_prog
 
     def _generator(self):
-        vmlinux = self.context.modules[self.config["kernel"]]
-        for prog in self.get_ebpf_programs(vmlinux):
+        for prog in self.get_ebpf_programs(self.context, self.config["kernel"]):
             prog_addr = prog.vol.offset
             prog_type = prog.get_type() or renderers.NotAvailableValue()
             prog_tag = prog.get_tag() or renderers.NotAvailableValue()
--- a/volatility3/framework/plugins/linux/ebpf.py
+++ b/volatility3/framework/plugins/linux/ebpf.py
@@ -1,0 +1,78 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import binascii
+import logging
+from typing import List
+
+from volatility3.framework import renderers, interfaces, exceptions
+from volatility3.framework.objects import utility
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.interfaces import plugins
+from volatility3.framework.configuration import requirements
+
+vollog = logging.getLogger(__name__)
+
+
+class EBPF(plugins.PluginInterface):
+    """Enumerate eBPF programs"""
+
+    _required_framework_version = (2, 0, 0)
+
+    _version = (1, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Linux kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+        ]
+
+    def get_ebpf_programs(self, vmlinux) -> interfaces.objects.ObjectInterface:
+        """Enumerate eBPF programs walking its IDR.
+
+        Args:
+            vmlinux: The kernel symbols object
+
+        Yields:
+            eBPF program objects
+        """
+        if not vmlinux.has_symbol("prog_idr"):
+            raise exceptions.VolatilityException(
+                "Cannot find the eBPF prog idr. Unsupported kernel"
+            )
+
+        prog_idr_addr = vmlinux.get_symbol("prog_idr").address
+        prog_idr = vmlinux.object("idr", offset=prog_idr_addr)
+        for page_addr in prog_idr.get_page_addresses():
+            bpf_prog = vmlinux.object("bpf_prog", offset=page_addr, absolute=True)
+            yield bpf_prog
+
+    def _generator(self):
+        vmlinux = self.context.modules[self.config["kernel"]]
+        vmlinux_layer = vmlinux.context.layers[vmlinux.layer_name]
+        bpf_prog_types = vmlinux.get_enumeration("bpf_prog_type")
+        for prog in self.get_ebpf_programs(vmlinux):
+            prog_addr = prog.vol.offset
+            prog_type = bpf_prog_types.lookup(prog.type)
+            prog_tag_addr = prog.tag.vol.offset
+            prog_tag_size = prog.tag.count
+            prog_tag_bytes = vmlinux_layer.read(prog_tag_addr, prog_tag_size)
+            prog_tag = binascii.hexlify(prog_tag_bytes).decode()
+            prog_name = (
+                utility.array_to_string(prog.aux.name) or renderers.NotAvailableValue()
+            )
+            fields = (format_hints.Hex(prog_addr), prog_name, prog_tag, prog_type)
+            yield (0, fields)
+
+    def run(self):
+        headers = [
+            ("Address", format_hints.Hex),
+            ("Name", str),
+            ("Tag", str),
+            ("Type", str),
+        ]
+        return renderers.TreeGrid(headers, self._generator())
--- a/volatility3/framework/plugins/linux/ebpf.py
+++ b/volatility3/framework/plugins/linux/ebpf.py
@@ -44,7 +44,7 @@ class EBPF(plugins.PluginInterface):
             )
 
         prog_idr = vmlinux.object_from_symbol("prog_idr")
-        for page_addr in prog_idr.get_page_addresses():
+        for page_addr in prog_idr.get_entries():
             bpf_prog = vmlinux.object("bpf_prog", offset=page_addr, absolute=True)
             yield bpf_prog
 
--- a/volatility3/framework/plugins/linux/ebpf.py
+++ b/volatility3/framework/plugins/linux/ebpf.py
@@ -43,8 +43,7 @@ class EBPF(plugins.PluginInterface):
                 "Cannot find the eBPF prog idr. Unsupported kernel"
             )
 
-        prog_idr_addr = vmlinux.get_symbol("prog_idr").address
-        prog_idr = vmlinux.object("idr", offset=prog_idr_addr)
+        prog_idr = vmlinux.object_from_symbol("prog_idr")
         for page_addr in prog_idr.get_page_addresses():
             bpf_prog = vmlinux.object("bpf_prog", offset=page_addr, absolute=True)
             yield bpf_prog
--- a/volatility3/framework/plugins/linux/ebpf.py
+++ b/volatility3/framework/plugins/linux/ebpf.py
@@ -1,12 +1,10 @@
 # This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
-import binascii
 import logging
 from typing import List
 
 from volatility3.framework import renderers, interfaces, exceptions
-from volatility3.framework.objects import utility
 from volatility3.framework.renderers import format_hints
 from volatility3.framework.interfaces import plugins
 from volatility3.framework.configuration import requirements
@@ -53,18 +51,11 @@ class EBPF(plugins.PluginInterface):
 
     def _generator(self):
         vmlinux = self.context.modules[self.config["kernel"]]
-        vmlinux_layer = vmlinux.context.layers[vmlinux.layer_name]
-        bpf_prog_types = vmlinux.get_enumeration("bpf_prog_type")
         for prog in self.get_ebpf_programs(vmlinux):
             prog_addr = prog.vol.offset
-            prog_type = bpf_prog_types.lookup(prog.type)
-            prog_tag_addr = prog.tag.vol.offset
-            prog_tag_size = prog.tag.count
-            prog_tag_bytes = vmlinux_layer.read(prog_tag_addr, prog_tag_size)
-            prog_tag = binascii.hexlify(prog_tag_bytes).decode()
-            prog_name = (
-                utility.array_to_string(prog.aux.name) or renderers.NotAvailableValue()
-            )
+            prog_type = prog.get_type() or renderers.NotAvailableValue()
+            prog_tag = prog.get_tag() or renderers.NotAvailableValue()
+            prog_name = prog.get_name() or renderers.NotAvailableValue()
             fields = (format_hints.Hex(prog_addr), prog_name, prog_tag, prog_type)
             yield (0, fields)
 
--- a/volatility3/framework/plugins/linux/mountinfo.py
+++ b/volatility3/framework/plugins/linux/mountinfo.py
@@ -37,7 +37,7 @@ class MountInfo(plugins.PluginInterface):
 
     _required_framework_version = (2, 2, 0)
 
-    _version = (1, 1, 0)
+    _version = (1, 2, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
@@ -143,8 +143,8 @@ class MountInfo(plugins.PluginInterface):
             sb_opts,
         )
 
+    @staticmethod
     def _get_tasks_mountpoints(
-        self,
         tasks: Iterable[interfaces.objects.ObjectInterface],
         filtered_by_pids: bool = False,
     ):
@@ -247,17 +247,26 @@ class MountInfo(plugins.PluginInterface):
                 "Could not filter by mount namespace id. This field is not available in this kernel."
             )
 
-    def get_superblocks(self):
+    @classmethod
+    def get_superblocks(
+        cls,
+        context: interfaces.context.ContextInterface,
+        vmlinux_module_name: str,
+    ) -> Iterable[interfaces.objects.ObjectInterface]:
         """Yield file system superblocks based on the task's mounted filesystems.
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            vmlinux_module_name: The name of the kernel module on which to operate
 
         Yields:
             super_block: Kernel's struct super_block object
         """
         # No filter so that we get all the mount namespaces from all tasks
-        tasks = pslist.PsList.list_tasks(self.context, self.config["kernel"])
+        tasks = pslist.PsList.list_tasks(context, vmlinux_module_name)
 
         seen_sb_ptr = set()
-        for task, mnt, _mnt_ns_id in self._get_tasks_mountpoints(tasks):
+        for task, mnt, _mnt_ns_id in cls._get_tasks_mountpoints(tasks):
             path_root = linux.LinuxUtilities.get_path_mnt(task, mnt)
             if not path_root:
                 continue
--- a/volatility3/framework/plugins/linux/mountinfo.py
+++ b/volatility3/framework/plugins/linux/mountinfo.py
@@ -254,10 +254,7 @@ class MountInfo(plugins.PluginInterface):
             super_block: Kernel's struct super_block object
         """
         # No filter so that we get all the mount namespaces from all tasks
-        pid_filter = pslist.PsList.create_pid_filter()
-        tasks = pslist.PsList.list_tasks(
-            self.context, self.config["kernel"], filter_func=pid_filter
-        )
+        tasks = pslist.PsList.list_tasks(self.context, self.config["kernel"])
 
         seen_sb_ptr = set()
         for task, mnt, _mnt_ns_id in self._get_tasks_mountpoints(tasks):
--- a/volatility3/framework/plugins/linux/mountinfo.py
+++ b/volatility3/framework/plugins/linux/mountinfo.py
@@ -37,7 +37,7 @@ class MountInfo(plugins.PluginInterface):
 
     _required_framework_version = (2, 2, 0)
 
-    _version = (1, 0, 0)
+    _version = (1, 1, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
@@ -146,7 +146,7 @@ class MountInfo(plugins.PluginInterface):
     def _get_tasks_mountpoints(
         self,
         tasks: Iterable[interfaces.objects.ObjectInterface],
-        filtered_by_pids: bool,
+        filtered_by_pids: bool = False,
     ):
         seen_mountpoints = set()
         for task in tasks:
@@ -184,8 +184,8 @@ class MountInfo(plugins.PluginInterface):
         self,
         tasks: Iterable[interfaces.objects.ObjectInterface],
         mnt_ns_ids: List[int],
-        mount_format: bool,
-        filtered_by_pids: bool,
+        mount_format: bool = False,
+        filtered_by_pids: bool = False,
     ) -> Iterable[Tuple[int, Tuple]]:
         show_filter_warning = False
         for task, mnt, mnt_ns_id in self._get_tasks_mountpoints(
@@ -246,6 +246,31 @@ class MountInfo(plugins.PluginInterface):
             vollog.warning(
                 "Could not filter by mount namespace id. This field is not available in this kernel."
             )
+
+    def get_superblocks(self):
+        """Yield file system superblocks based on the task's mounted filesystems.
+
+        Yields:
+            super_block: Kernel's struct super_block object
+        """
+        # No filter so that we get all the mount namespaces from all tasks
+        pid_filter = pslist.PsList.create_pid_filter()
+        tasks = pslist.PsList.list_tasks(
+            self.context, self.config["kernel"], filter_func=pid_filter
+        )
+
+        seen_sb_ptr = set()
+        for task, mnt, _mnt_ns_id in self._get_tasks_mountpoints(tasks):
+            path_root = linux.LinuxUtilities.get_path_mnt(task, mnt)
+            if not path_root:
+                continue
+
+            sb_ptr = mnt.get_mnt_sb()
+            if not sb_ptr or sb_ptr in seen_sb_ptr:
+                continue
+            seen_sb_ptr.add(sb_ptr)
+
+            yield sb_ptr.dereference(), path_root
 
     def run(self):
         pids = self.config.get("pids")
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -486,7 +486,7 @@ class InodePages(plugins.PluginInterface):
             page_index = int(page_obj.index)
             page_file_offset = page_index * vmlinux_layer.page_size
             dump_safe = page_file_offset < inode_size
-            page_flags_list = page_obj.get_flags()
+            page_flags_list = page_obj.get_flags_list()
             page_flags = ",".join([x.replace("PG_", "") for x in page_flags_list])
             fields = (
                 page_vaddr,
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -131,7 +131,10 @@ class Files(plugins.PluginInterface, timeliner.TimeLinerInterface):
         ]
 
     @staticmethod
-    def _follow_symlink(inode, symlink_path) -> str:
+    def _follow_symlink(
+        inode: interfaces.objects.ObjectInterface,
+        symlink_path: str,
+    ) -> str:
         """Follows (fast) symlinks (kernels >= 4.2.x).
         Fast symlinks are filesystem agnostic.
 
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -1,0 +1,504 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import math
+import logging
+import datetime
+from dataclasses import dataclass, astuple
+from typing import List
+
+from volatility3.framework import renderers, interfaces
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.interfaces import plugins
+from volatility3.framework.configuration import requirements
+from volatility3.plugins import timeliner
+from volatility3.plugins.linux import mountinfo
+
+vollog = logging.getLogger(__name__)
+
+
+@dataclass
+class InodeUser:
+    """Inode user representation, featuring augmented information and formatted fields.
+    This is the data the plugin will eventually display.
+    """
+
+    superblock_addr: int
+    mountpoint: str
+    device: str
+    inode_num: int
+    inode_addr: int
+    type: str
+    inode_pages: int
+    cached_pages: int
+    file_mode: str
+    access_time: str
+    modification_time: str
+    change_time: str
+    path: str
+
+
+@dataclass
+class InodeInternal:
+    """Inode internal representation containing only the core objects
+
+    Fields:
+        superblock: 'super_block' struct
+        mountpoint: Superblock mountpoint path
+        inode: 'inode' struct
+        path: Dentry full path
+    """
+
+    superblock: interfaces.objects.ObjectInterface
+    mountpoint: str
+    inode: interfaces.objects.ObjectInterface
+    path: str
+
+    def to_user(
+        self, kernel_layer: interfaces.layers.TranslationLayerInterface
+    ) -> InodeUser:
+        """Augment the inode information to be presented to the user
+
+        Args:
+            kernel_layer: The kernel layer to obtain the page size
+
+        Returns:
+            An InodeUser dataclass
+        """
+        # Ensure all types are atomic immutable. Otherwise, astuple() will take a long
+        # time doing a deepcopy of the Volatility objects.
+        superblock_addr = self.superblock.vol.offset
+        device = f"{self.superblock.major}:{self.superblock.minor}"
+        inode_num = int(self.inode.i_ino)
+        inode_addr = self.inode.vol.offset
+        inode_type = renderers.UnparsableValue()
+        # Round up the number of pages to fit the inode's size
+        inode_pages = int(math.ceil(self.inode.i_size / float(kernel_layer.page_size)))
+        cached_pages = int(self.inode.i_mapping.nrpages)
+        file_mode = self.inode.get_file_mode()
+        access_time_dt = self.inode.get_access_time()
+        modification_time_str = self.inode.get_modification_time()
+        change_time_str = self.inode.get_change_time()
+
+        inode_user = InodeUser(
+            superblock_addr=superblock_addr,
+            mountpoint=self.mountpoint,
+            device=device,
+            inode_num=inode_num,
+            inode_addr=inode_addr,
+            type=inode_type,
+            inode_pages=inode_pages,
+            cached_pages=cached_pages,
+            file_mode=file_mode,
+            access_time=access_time_dt,
+            modification_time=modification_time_str,
+            change_time=change_time_str,
+            path=self.path,
+        )
+        return inode_user
+
+
+class Files(plugins.PluginInterface, timeliner.TimeLinerInterface):
+    """Lists files from memory"""
+
+    _required_framework_version = (2, 0, 0)
+
+    _version = (1, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Linux kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.PluginRequirement(
+                name="mountinfo", plugin=mountinfo.MountInfo, version=(1, 1, 0)
+            ),
+            requirements.ListRequirement(
+                name="type",
+                description="List of space-separated file type filters i.e. --type REG DIR",
+                element_type=str,
+                optional=True,
+            ),
+            requirements.StringRequirement(
+                name="find",
+                description="Filename (full path) to find",
+                optional=True,
+            ),
+        ]
+
+    @staticmethod
+    def _follow_symlink(inode, symlink_path) -> str:
+        """Follows (fast) symlinks (kernels >= 4.2.x).
+        Fast symlinks are filesystem agnostic.
+
+        Args:
+            inode: The inode (or pointer) to dump
+            symlink_path: The symlink name
+
+        Returns:
+            If it can resolve the symlink, it returns a string "symlink_path -> target_path"
+            Otherwise, it returns the same symlink_path
+        """
+        # i_link (fast symlinks) were introduced in 4.2
+        if inode and inode.is_link and inode.has_member("i_link") and inode.i_link:
+            i_link_str = inode.i_link.dereference().cast(
+                "string", max_length=255, encoding="utf-8", errors="replace"
+            )
+            symlink_path = f"{symlink_path} -> {i_link_str}"
+
+        return symlink_path
+
+    @classmethod
+    def _walk_dentry(cls, seen_dentries, root_dentry, parent):
+
+        for dentry in root_dentry.get_subdirs():
+            dentry_addr = dentry.vol.offset
+
+            # corruption
+            if dentry_addr == root_dentry.vol.offset:
+                continue
+
+            if dentry_addr in seen_dentries:
+                continue
+
+            seen_dentries.add(dentry_addr)
+
+            inode = dentry.d_inode
+            if not (inode and inode.is_valid()):
+                continue
+
+            # This allows us to have consistent paths
+            if dentry.d_name.name:
+                name = dentry.d_name.name_as_str()
+                # Do NOT use os.path.join() below
+                new_file = parent + "/" + name
+            else:
+                continue
+
+            yield new_file, dentry, dentry.d_parent.vol.offset
+
+            if inode.is_dir:
+                for new_file, dentry, parent_address in cls._walk_dentry(
+                    seen_dentries, dentry, new_file
+                ):
+                    yield new_file, dentry, parent_address
+
+    @classmethod
+    def get_inodes(
+        cls,
+        context: interfaces.context.ContextInterface,
+        config_path: str,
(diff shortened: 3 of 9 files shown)


gcmoreira:
Hey @ikelos, not sure why this cannot merge those changes. It looks pretty simple to me. Anyway, I'm not allowed to resolve those conflicts, it seems I will need your help here. Thanks

volatility3/framework/plugins/linux/ebpf.py
```diff
@@ -0,0 +1,78 @@
...
+            bpf_prog = vmlinux.object("bpf_prog", offset=page_addr, absolute=True)
+            yield bpf_prog
+
+    def _generator(self):
+        vmlinux = self.context.modules[self.config["kernel"]]
+        vmlinux_layer = vmlinux.context.layers[vmlinux.layer_name]
+        bpf_prog_types = vmlinux.get_enumeration("bpf_prog_type")
+        for prog in self.get_ebpf_programs(vmlinux):
+            prog_addr = prog.vol.offset
+            prog_type = bpf_prog_types.lookup(prog.type)
```
ikelos: I feel like you *should* just be able to ask `prog.type.lookup()` if prog of the enumeration type?  You could also do `prog.type.description` although it's technically more characters...

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -247,6 +247,31 @@ def _generator(
...
+    def get_superblocks(self):
+        """Yield file system superblocks based on the task's mounted filesystems.
+
+        Yields:
+            super_block: Kernel's struct super_block object
+        """
+        # No filter so that we get all the mount namespaces from all tasks
+        pid_filter = pslist.PsList.create_pid_filter()
+        tasks = pslist.PsList.list_tasks(
+            self.context, self.config["kernel"], filter_func=pid_filter
```
ikelos: `filter_func` can be unset, and it should default to the same thing?

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -247,6 +247,31 @@ def _generator(
...
+        )
+
+        seen_sb_ptr = set()
+        for task, mnt, _mnt_ns_id in self._get_tasks_mountpoints(tasks):
```
ikelos: Thanks, good to know where to get the namespace if we later want to use it!  5:D

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -0,0 +1,504 @@
...
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.interfaces import plugins
+from volatility3.framework.configuration import requirements
+from volatility3.plugins import timeliner
+from volatility3.plugins.linux import mountinfo
+
+vollog = logging.getLogger(__name__)
+
+
+@dataclass
```
ikelos, line 21: Do these belong here, or should they live in LinuxUtilities somewhere (ie, are they always likely to be used with this plugin, or should they be more general), given that there are already elf and bash extesions under `framework/symbols/linux/extensions` so support for adding new files and common types already exists (albeit, it does require tinkering with `__init__.py` to import them be default, I think).

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -0,0 +1,504 @@
...
+        if inode and inode.is_link and inode.has_member("i_link") and inode.i_link:
+            i_link_str = inode.i_link.dereference().cast(
+                "string", max_length=255, encoding="utf-8", errors="replace"
+            )
+            symlink_path = f"{symlink_path} -> {i_link_str}"
+
+        return symlink_path
+
+    @classmethod
+    def _walk_dentry(cls, seen_dentries, root_dentry, parent):
```
ikelos: Could do with a docstring and typing please, given you've been so good about the others...

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -0,0 +1,504 @@
...
+            ),
+            requirements.StringRequirement(
+                name="find",
+                description="Filename (full path) to find",
+                optional=True,
+            ),
+        ]
+
+    @staticmethod
+    def _follow_symlink(inode, symlink_path) -> str:
```
ikelos: Some typing isn't here, could we at least try to include simple types (which I'm assuming symlink_path is) please?

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -0,0 +1,504 @@
...
+                    seen_dentries, dentry, new_file
+                ):
+                    yield new_file, dentry, parent_address
+
+    @classmethod
+    def get_inodes(
+        cls,
+        context: interfaces.context.ContextInterface,
+        config_path: str,
+    ):
```
ikelos: If this yields a generator of inodes, could we capture that in the return type please?  Even if it's just `Generator[interfaces.object.ObjectInterface, None, None]`

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -0,0 +1,504 @@
...
+
+    @staticmethod
+    def format_fields_with_headers(headers, generator):
+        """Uses the headers type to cast the fields obtained from the generator"""
+        for level, fields in generator:
+            formatted_fields = []
+            for header, field in zip(headers, fields):
+                header_type = header[1]
+
+                if isinstance(
```
ikelos, line 348: There was a recent push to allow you to cast the response values as `format_hints.HexOrAbsent` to take care of the absent values, but that still looks like it might be a little cumbersome in this instance.  I'll look again and see what I can do about easing the type tests for these...

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -0,0 +1,504 @@
...
+            ),
+            requirements.ListRequirement(
+                name="type",
+                description="List of space-separated file type filters i.e. --type REG DIR",
+                element_type=str,
+                optional=True,
+            ),
+            requirements.StringRequirement(
+                name="find",
+                description="Filename (full path) to find",
```
ikelos, line 128: Is there a reason this isn't a substring match, but a full path match instead?  Substring presumably encompasses both (although you wouldn't get just the first hit).

volatility3/framework/plugins/linux/pidhashtable.py
```diff
@@ -0,0 +1,249 @@
...
+        super().__init__(*args, **kwargs)
+        self.vmlinux = None
+        self.vmlinux_layer = None
+
+    def _is_valid_task(self, task):
+        return task and task.pid > 0 and self.vmlinux_layer.is_valid(task.parent)
+
+    def _get_pidtype_pid(self):
+        # The pid_type enumeration is present since 2.5.37, just in case
+        pid_type_enum = self.vmlinux.get_enumeration("pid_type")
```
ikelos: I'd much prefer this parameterized than maintaining state.  I can't fortell if someone's going to reuse plugins in a different UI somehow, so maintaining it by passing it through parameters rather than state seems better and would allow these functions to be exposed later on if they prove useful for others.

volatility3/framework/plugins/linux/pidhashtable.py
```diff
@@ -0,0 +1,249 @@
...
+        # Typically PIDTYPE_PID = 0
+        return pidtype_pid
+
+    def _get_pidhash_array(self):
+        pidhash_shift = self.vmlinux.object_from_symbol("pidhash_shift")
+        pidhash_size = 1 << pidhash_shift
+
+        array_type_name = self.vmlinux.symbol_table_name + constants.BANG + "array"
+
+        pidhash_ptr = self.vmlinux.object_from_symbol("pid_hash")
```
ikelos: Hopefully these methods don't get called too often, otherwise passing through simple parameters and having to do look up would be quite inefficient, but I think I'd still prefer a complex parameter to a self one in a case like this

volatility3/framework/renderers/conversion.py
```diff
@@ -11,6 +11,7 @@
 from volatility3.framework import interfaces, renderers
 
 
+# FIXME: Move wintime_to_datetime() and unixtime_to_datetime() out of renderers, possibly framework.objects.utility
```
ikelos, line 14: Thanks!  5:)

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -419,3 +430,342 @@ def get_module_from_volobj_type(
...
+            constants.BANG + "radix_tree_root"
+        ) and vmlinux.get_type("radix_tree_root").has_member("xa_head")
+
+        if i_pages_is_xarray or i_pages_is_radix_tree_root:
+            return XArray(vmlinux)
+        else:
+            return RadixTree(vmlinux)
+
+
+class Tree(ABC):
```
ikelos: The name's a little generic.  The module helps identify it, but if there's a slightly more descriptive name than Tree, it would be good to just differentiate it a little.  There are other operating systems in here too, and we don't want all of them to have their own tree (plus, I'm quite sure linux will have different tree structures too).  5;P

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -419,3 +430,342 @@ def get_module_from_volobj_type(
...
+    """Abstraction to support both XArray and RadixTree"""
+
+    # Dynamic values, these will be initialized later
+    CHUNK_SHIFT = None
+    CHUNK_SIZE = None
+    CHUNK_MASK = None
+
+    def __init__(self, vmlinux: interfaces.context.ModuleInterface):
```
ikelos: For something like this, I'd prefer to pass in the context and a name lookup for `context.modules` please.  It ensures that the state is front and center, and that the rest of the parameters are simple in case they need to be pickled or anything for multithreading down the line.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -419,3 +430,342 @@ def get_module_from_volobj_type(
...
+
+    @property
+    @abstractmethod
+    def node_type_name(self) -> str:
+        """Returns the Tree implementation node type name
+
+        Returns:
+            A string with the node type name
+        """
+        raise NotImplementedError()
```
ikelos, line 504: Ok with these assuming they'll be filled in at some point?

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -419,3 +430,342 @@ def get_module_from_volobj_type(
...
+        self,
+        page_cache: interfaces.objects.ObjectInterface,
+        vmlinux: interfaces.context.ModuleInterface,
+    ):
+        """
+        Args:
+            page_cache: Page cache address space
+            vmlinux: Kernel module object
+        """
+        self.vmlinux = vmlinux
```
ikelos: Same as above, ideally.  First parameter would be `context`, then I'd go `vmlinux_name: str` and then `page_cache: interfaces.objects.ObjectInterface` or even `page_cache: 'PageCache'`

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -419,3 +430,342 @@ def get_module_from_volobj_type(
...
+        self._page_cache = page_cache
+        self._tree = LinuxUtilities.choose_kernel_tree(self.vmlinux)
```
ikelos: In fact, if this is all it's used for, you can just pass the `context` and `vmlinux_name` right through.

ikelos (commented):
Well, quite a long chunk of code and if this weren't from someone that already knows most of the code style points I'd raise, I'd have asked them to come back and submit them as several standalone chunks, but you've been submitting long enough you don't fall for most of the common traps so it didn't take too long to review.  5;P

Some docstrings and types in places (some are so short it's easier just to read the method, but any over a few lines should get their own docstring please, it'll help when people aren't looking at the source but just the documentation).

Towards the end I started noticing several functions that just passed an object through and in general we try hard to pass a context and then simple parameter types (like strings) that are used as a look key instead.  I've mentioned the reasons in the comments (multithreading and ease of understanding given types are hard to type further than `ObjectInterface`), but at this point it's practically code style for the project.

Generally though, it's all very nice,and I'm  very happy with some of the touches in here.  Really good work, thanks!

ikelos:
> Hey @ikelos, not sure why this cannot merge those changes. It looks pretty simple to me. Anyway, I'm not allowed to resolve those conflicts, it seems I will need your help here. Thanks

Yeah, that was a bit weird, they weren't really conflict from what I could see, but I guess some of the recent merges touched the files in those places.  Anyway, should be all resolved, lemme know if there's an issue...

volatility3/framework/plugins/linux/ebpf.py
gcmoreira: Done. I realized I had already added a `bpf_prog` extension object that returned the prog type when I was working on `Sockstat`. So, I updated the eBPF extension objects. This change will improve support for older kernels and help avoid code duplication in future eBPF plugins.

volatility3/framework/plugins/linux/pagecache.py
gcmoreira: Hm I don't think so. These dataclasses are specific for this plugin and its output. The idea of these two is similar to the MVC design pattern. Being InodeInternal and InodeUser the Model and View representations respectively. 
They work great! The code becomes a lot simpler and clearer.

volatility3/framework/symbols/linux/__init__.py
gcmoreira: Yeah, sorry I know. Actually, 'Tree' isn't quite right since XArray is not a tree. I had a hard time coming up with a good name. Basically, it's an abstract base class which allows selecting and executing the Linux kernel RadixTree and XArray data structures. Earlier kernel versions used RadixTree in various subsystems, but many of these have been replaced with XArrays.  Since these serve various functions in the kernel, it seems best to keep the name generic either way. What do you think of something like DSAbstract or DSBase (DS => Data Structure)?, any better?

volatility3/framework/symbols/linux/__init__.py
gcmoreira: Yes, see the RadixTree and XArray classes

volatility3/framework/symbols/linux/__init__.py
gcmoreira: Ok, renamed to IDStorage as it represents what all of them store, which is identifiers.

volatility3/framework/symbols/linux/__init__.py
gcmoreira: done

volatility3/framework/symbols/linux/__init__.py
gcmoreira: done

volatility3/framework/symbols/linux/__init__.py
gcmoreira: done

gcmoreira:
@ikelos I changed the state to `draft`. It's almost ready but I want to add some testcases for completeness. 

In the meantime, if you can please have a look at the **LinuxUtilities** version. I bumped the minor number when I added the ex **choose_kernel_tree()** helper, but, I later moved it to the **IDStorage** class. However, I made a code improvement,  which could also be considered a **patch**, since the functionality itself hasn't changed. Not sure the exact criteria we should take into account for this. If we need to update the version again, the **pidhashtable** plugin requirements also needs to be adjusted.

ikelos:
No problem moving this to draft.  If the change you've made doesn't add to the LinuxUtilities interface (or remove/change it in any way) but just alters *how* it does the stuff it does (without changing the results) then it can just be bumped on the PATCH version rather than the MINOR version.  Depends on the patch, but should be ok (otherwise we tend to leave the old functionality around with a deprecation warning, because bumping a MAJOR version on LinuxUtilities will be... painful.  5:P

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -67,7 +74,7 @@ def __init__(self, *args, **kwargs) -> None:
 class LinuxUtilities(interfaces.configuration.VersionableInterface):
     """Class with multiple useful linux functions."""
 
-    _version = (2, 1, 0)
+    _version = (2, 2, 0)
```
ikelos: This is right, given that `IDStorage` is being added. :+1:

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1584,20 +1608,59 @@ def get_state(self):
 
 
 class bpf_prog(objects.StructType):
-    def get_type(self):
+    def get_type(self) -> Union[str, None]:
+        """Returns a string with the eBPF program type"""
```
ikelos, line 1626: Awesome, thanks!  The typing and docstrings will help a lot  5:)

ikelos (commented):
There's getting to be a *lot* of changes in this one PR.  I realize they're all related/linked, but it might be better to add in distinct classes as their own PRs, and then tie them all together with a plugin that uses them later on, just so the reviewer doesn't have pages and pages of code to have to check over.  It might be easier to spot smaller gotchas (although it may also be harder to see how things interact).  This is also why having the typing information helps a lot, because it means that checking cross links to things should be easier knowing what they take in/return...

gcmoreira:
Hey @ikelos, IMO it's ready to merge. Again, my apologies. I know it's quite extensive chuck of code. I tried to break it down in some PRs, but I know I could have done it even better.  I'll work on improving this for the next PR. 
On a positive note, I'm pretty sure our DFIR users will find these features very useful. Cheers

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -247,6 +247,28 @@ def _generator(
...
+
+        Yields:
+            super_block: Kernel's struct super_block object
+        """
+        # No filter so that we get all the mount namespaces from all tasks
+        tasks = pslist.PsList.list_tasks(self.context, self.config["kernel"])
+
+        seen_sb_ptr = set()
+        for task, mnt, _mnt_ns_id in self._get_tasks_mountpoints(tasks):
+            path_root = linux.LinuxUtilities.get_path_mnt(task, mnt)
```
ikelos, line 270: Does this now need LinuxUtilities `(2,2,0)` or will it work with `(2,1,0)` as the plugin says?

volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -247,6 +247,28 @@ def _generator(
                 "Could not filter by mount namespace id. This field is not available in this kernel."
             )
 
+    def get_superblocks(self):
```
ikelos: Since this *isn't* being exposed as part of the plugin's API, it might be better to make that explicit by starting the function with `_get_superblocks` so that people don't try calling it from outside the plugin.  Either that, or you could expose it (although I'd probably parameterized it on `tasks` if you're going to do that.

volatility3/framework/plugins/linux/ebpf.py
```diff
@@ -0,0 +1,68 @@
...
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Linux kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+        ]
+
+    def get_ebpf_programs(self, vmlinux) -> interfaces.objects.ObjectInterface:
```
ikelos: This isn't being expose, so it's *technically* not a problem, but we tend to prefer call signatures to be `context` and then `module_name` and the lookup happen inside the function.  Your call whether to update it, but if do you ever intend to expose `get_ebpf_programs` (which seem like they'd be handy) then I'd prefer you did make this change please.

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -0,0 +1,519 @@
...
+        if not inode.is_reg:
+            vollog.error("The inode is not a regular file")
+            return
+
+        # By using truncate/seek, provided the filesystem supports it, a sparse file will be
+        # created, saving both disk space and I/O time.
+        # Additionally, using the page index will guarantee that each page is written at the
+        # appropriate file position.
+        try:
+            with open(filename, "wb") as f:
```
ikelos: This breaks if this is actually being used through a web interface or similar.  All file operations need to be conducted using the [plugin's `open` method](https://github.com/volatilityfoundation/volatility3/blob/4578434bbd1587a8ac6c56fcc2757d3dcbe2edb7/volatility3/framework/interfaces/plugins.py#L146).  See any of the `dump` methods for an example, but essentially make a file handle using `open` and then set the preferred filename attribute on the file handle.

volatility3/framework/plugins/linux/pidhashtable.py
```diff
@@ -0,0 +1,249 @@
...
+                name="decorate_comm",
+                description="Show `user threads` comm in curly brackets, and `kernel threads` comm in square brackets",
+                optional=True,
+                default=False,
+            ),
+        ]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.vmlinux = None
```
ikelos: Saving state is extremely uncommon in plugins and may prevent us doing this with plugins in the future (certainly it prevents us exposing methods to be used by other plugins).  Ideally every method that makes use of these would be parameterized to accept them (or more accurately, the context and the name to lookup the module).

volatility3/framework/plugins/linux/pidhashtable.py
```diff
@@ -0,0 +1,249 @@
...
+
+        return None
+
+    def get_tasks(self) -> interfaces.objects.ObjectInterface:
+        """Enumerates processes through the PID hash table
+
+        Yields:
+            task_struct objects
+        """
+        self.vmlinux = self.context.modules[self.config["kernel"]]
```
ikelos: Again, this might be better parameterizing the `context` and `kernel` value, and then make it a classmethod?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1897,3 +1960,250 @@ def get_file_mode(self) -> str:
...
+    @property
+    @functools.lru_cache()
+    def pageflags_enum(self) -> Dict:
+        """Returns 'pageflags' enumeration key/values
+
+        Returns:
+            A dictionary with the pageflags enumeration key/values
+        """
+        # FIXME: It would be even better to use @functools.cached_property instead,
+        # however, this requires Python +3.8
```
ikelos, line 2033: We've got a PR ( #1208 ) that may push the minimum version to 3.8, so keep an eye out.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1897,3 +1960,250 @@ def get_file_mode(self) -> str:
...
+
+        return flags
+
+    def to_paddr(self) -> int:
+        """Converts a page's virtual address to its physical address using the current physical memory model.
+
+        Returns:
+            int: page physical address
+        """
+        vmlinux = linux.LinuxUtilities.get_module_from_volobj_type(self._context, self)
```
ikelos: This doesn't check the version of `LinuxUtilities`, so we need to be a little careful using it.  Since these are both core components it should be ok, but it might be better to just check the version and throw an exception is it's not right (you can check the `VersionRequirement` code to see how to do that).

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1897,3 +1960,250 @@ def get_file_mode(self) -> str:
...
+        if vmlinux.has_symbol("mem_section"):
+            # SPARSEMEM_VMEMMAP physical memory model: memmap is virtually contiguous
+            if vmlinux.has_symbol("vmemmap_base"):
+                # CONFIG_DYNAMIC_MEMORY_LAYOUT - KASLR kernels >= 4.9
+                vmemmap_start = vmlinux.object_from_symbol("vmemmap_base")
+            else:
+                # !CONFIG_DYNAMIC_MEMORY_LAYOUT
+                if vmlinux_layer._maxvirtaddr < 57:
+                    # 4-Level paging -> VMEMMAP_START = __VMEMMAP_BASE_L4
+                    vmemmap_base_l4 = 0xFFFFEA0000000000
```
ikelos, line 2092: Mmmmm, not keen on hard coded values in here.  I assume this can't be calculated from any other values you can get?

ikelos (commented):
Ok, well.  There's a *lot* to go through there, so get that fixed up, I'll give it another quick pass and then we'll need to keep an eye on it and see if anything breaks...  5:S  Ideally smaller chunks in future please...

volatility3/framework/plugins/linux/mountinfo.py
gcmoreira: It's exposed and used by the **pagecache.Files** plugin, it's called from **get_inodes()**. This gathers all the system superblocks. To obtain that list, it needs to walk all the tasks. However, it doesn't make sense to call this with a subset of tasks.

volatility3/framework/plugins/linux/mountinfo.py
gcmoreira: linux.LinuxUtilities.get_path_mnt() wasn't modified by this PR. This can keep the (2,1,0) requirement

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Yeah, I agree, but yeah it's a kernel constant, so I kept the kernel constant names for clarity. It also shouldn't be used anywhere outside of this context. Otherwise, I would have put it in the Linux constants file.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: hm done, but I'm not sure we want this, have a look at the last commit

volatility3/framework/plugins/linux/pidhashtable.py
gcmoreira: resolved

volatility3/framework/plugins/linux/pidhashtable.py
gcmoreira: resolved

volatility3/framework/plugins/linux/pagecache.py
gcmoreira: done

volatility3/framework/plugins/linux/ebpf.py
gcmoreira: done

gcmoreira:
@ikelos if you are happy with the last changes, it should be ready to go

volatility3/framework/plugins/linux/mountinfo.py
ikelos: Ok, if it's supposed to be called from other plugins, it needs to be a classmethod or staticmethod then.  Instantiating this plugin within another plugin just for this functionality is a bit overkill and doesn't match the rest of the framework for how plugins expose functionality.

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -0,0 +1,521 @@
...
+
+        Args:
+            context: The context that the plugin will operate within
+            config_path: The path to configuration data within the context configuration data
+
+        Yields:
+            An InodeInternal object
+        """
+
+        superblocks_iter = mountinfo.MountInfo(
```
ikelos: This is a lot of machinery to call a function in another plugin.  Ideally the other plugin exposes it as a classmethod, and then there's no construction of the plugin required at all.

volatility3/framework/plugins/linux/pagecache.py
ikelos: Looks like comments in the pagecache file were hidden because the changes were too big, but do please review them.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -820,6 +823,26 @@ def d_ancestor(self, ancestor_dentry):
             current_dentry = current_dentry.d_parent
         return None
 
+    def get_subdirs(self) -> interfaces.objects.ObjectInterface:
```
ikelos, line 826: I feel as though the addition of this method should require a bump to the PATCH number at least, if not the MINOR number.  The only question is whether it's the whole framework or if this is part of the LinuxUtilities which is already versioned separately...

ikelos (commented):
Github keeps making this look smaller than it is by hiding files that changed a lot by default!  It's an enourmous patch and in future I'd prefer the groundwork/linux core stuff goes in first, then the plugins going in one by one (or at least file by file).  Having said that, this now looks just about done, but do please check *all* the comments (even those that github hides) to make sure they don't get missed.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Yeah, I see what you mean, but... I would like the check in there?  It looks like it could be reduced to a static method somewhere else that then appears as a oneliner at call sites, if the length/duplicative nature is what feels ugly?  I imagine a check like that will be useful for other code (or maybe even some that's already snuck in).

gcmoreira: Yeah, but it's becoming a checking versions infinite loop hehe. The get_module_from_volobj_type() was in LinuxUtilities to allow being reused, which need a version. Now, if we make another method to allow reusing the code to check the version of LinuxUtilities when calling get_module_from_volobj_type(), it won't also need a version :) ?

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: This is in the `dentry` extension object, but I don't think we've ever changed any versions when there were changes in this area. Hmm, it is not related to `LinuxUtilities`.

volatility3/framework/plugins/linux/pagecache.py
gcmoreira: done

volatility3/framework/plugins/linux/pagecache.py
gcmoreira: Done

volatility3/framework/plugins/linux/pagecache.py
gcmoreira: yeah, typically we only want to match one file. Otherwise, it has to find a way to find a name for each of them avoiding name collisions/overwriting the same file, escaping slashes, etc. I mean, not a big deal but let's keep it simple for now.

volatility3/framework/plugins/linux/pagecache.py
gcmoreira: Hmm... that's interesting! I'm not sure how useful it will be here, but feel free to give it a try!

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: @ikelos If I didn't miss anything, this should be the only open discussion remaining. Let me know if you're happy with how I fixed it

gcmoreira:
@ikelos I believe everything is done on my end. Please take a look when you have time. Cheers

volatility3/framework/plugins/linux/pidhashtable.py
```diff
@@ -44,17 +44,16 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.vmlinux = None
-        self.vmlinux_layer = None
-
     def _is_valid_task(self, task) -> bool:
-        return bool(task and task.pid > 0 and self.vmlinux_layer.is_valid(task.parent))
+        vmlinux = self.context.modules[self.config["kernel"]]
+        vmlinux_layer = self.context.layers[vmlinux.layer_name]
+        return bool(task and task.pid > 0 and vmlinux_layer.is_valid(task.parent))
```
ikelos, line 50: Why not just use the task's layer rather than the vmlinux layer (or has the task been made in a physical or process layer)?  `task.vol.layer_name` should get you back to the layer the task was created on?

volatility3/framework/plugins/linux/pidhashtable.py
```diff
@@ -44,17 +44,16 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
 
     def _get_pidtype_pid(self):
```
ikelos, line 52: I have to admit, I thought these would end up being parameterized classmethods, but as long as you're happy that they're only for use in this plugin this is fine.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: I might've missed where this code was defined.  If it's defined *in* LinuxUtilities then it can be certain of its own version number, but otherwise it should be checking to see that it's available.  I didn't mean add another method to LinuxUtilities, I meant add it to a shareable base class, but it's fine if not.  I can sometimes try to reduce co-reuse too much...  5;P

ikelos:
Yep, looks good, going to merge now, thanks for your patience getting it all sorted!  5:)

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Ok, I think I see what's going on here (and the LinuxUtilities checks I asked for earlier).  We've got bits of the core framework (extensions) that aren't versioned other than under the umbrella of the main core, that are using LinuxUtilities which we gave its own version number to.  So what *should* be/have been happening is the main core bumps (at least PATCH but probably minor) whenever a linux extension was added, only I'm not sure we always did that because people hate screwing with version numbers...  5:S  Probably we can fix this by going up by making sure all the extensions we want to add are in place, and then going to (2,9,0) and all plugins that need/make use of it require that as a minimum framework version.  Hopefully that'll clear everything up and even if it's a little broken in the middle, it'll give people a new version number they can then rely on to have the functionality they want/need.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: See my comment below, since these are changes to the linux extensions, they're rolled into the main framework version number, whose MINOR number should be bumped whenever an addition is made...

ikelos: (That also means we can rip out those LinuxUtilities checks I asked for and just rely on the framework and LinuxUtilities versions agreeing (sorry for the faffing, but getting these right so that people's plugins don't mysteriously not work on a unknown configurations is getting to be complex).

gcmoreira:
> Yep, looks good, going to merge now, thanks for your patience getting it all sorted! 5:)

Thank you, and again... very sorry for the long PR, I will submit smaller PRs next time. 
Very happy that Volatility3 now includes these features

PR #1234, opened 2024-08-06
add args and kwargs to threads.py init

dadokkio:
Without args and kwargs there were an issue with timeliner plugin that tried to pass additional parameters like progress_callback raising TypeError

```Traceback (most recent call last): File "/src/volatility3/volatility3/framework/plugins/timeliner.py", line 261, in run plugin = plugins.construct_plugin( ^^^^^^^^^^^^^^^^^^^^^^^^^ File "/src/volatility3/volatility3/framework/plugins/_init.py", line 62, in construct_plugin constructed = plugin( ^^^^^^^ TypeError: Threads.init_() got an unexpected keyword argument 'progress_callback'```

After a little debug I discovered that threads was the incriminated subplugin and it was missing additional parameters in __init__

### DIFF
--- a/volatility3/framework/plugins/windows/threads.py
+++ b/volatility3/framework/plugins/windows/threads.py
@@ -18,9 +18,9 @@ class Threads(thrdscan.ThrdScan):
     _required_framework_version = (2, 4, 0)
     _version = (1, 0, 0)
 
-    def __init__(self):
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


ikelos (approved):
Quite right, `interfaces.plugins.PluginInterface` takes parameters, but because it's inheritted, no error was raised, another reason I'm not keen on inheritance...  5:P

Thanks for the quick spot!  5:)

PR #1238, opened 2024-08-15
Linux: Add support for the mount namespace in kernels 6.8+

### LINKED ISSUE #1187: linux.mountinfo.MountInfo: broken on kernel 6.8+
**Describe the bug**
The `mnt_namespace.list` field got removed in kernel version 6.8, replaced with an rb-tree at `mnt_namespace.mounts`

**Context**
Volatility Version: 2.7.1 ( ac5769cf )
Operating System: Linux
Python Version: 3.12.2
Suspected Operating System: Ubuntu 24.10
Command: `vol linux.mountinfo`

**To Reproduce**

1. Acquire a memory dump of Ubuntu 24.10, kernel 6.8 or newer
2. Extract symbols from the kernel, applying patch from https://github.com/volatilityfoundation/dwarf2json/issues/57#issuecomment-2198205637 to dwarf2json
3. Try to use `linux.mountinfo`

**Expected behavior**
Mounts are getting printed

**Example output**
```
  File "/home/patrycja/Downloads/volatility3/vol.py", line 10, in <module>
    volatility3.cli.main()
  File "/home/patrycja/Downloads/volatility3/volatility3/cli/__init__.py", line 877, in main
    CommandLine().run()
  File "/home/patrycja/Downloads/volatility3/volatility3/cli/__init__.py", line 469, in run
    renderer.render(grid)
  File "/home/patrycja/Downloads/volatility3/volatility3/cli/text_renderer.py", line 198, in render
    grid.populate(visitor, outfd)
  File "/home/patrycja/Downloads/volatility3/volatility3/framework/renderers/__init__.py", line 245, in populate
  File "/home/patrycja/Downloads/volatility3/volatility3/framework/plugins/linux/mountinfo.py", line 191, in _generator
    for task, mnt, mnt_ns_id in self._get_tasks_mountpoints(
  File "/home/patrycja/Downloads/volatility3/volatility3/framework/plugins/linux/mountinfo.py", line 170, in _get_tasks_mountpoints
    for mount in mnt_namespace.get_mount_points():
  File "/home/patrycja/Downloads/volatility3/volatility3/framework/symbols/linux/extensions/__init__.py", line 1305, in get_mount_points
    for mount in self.list.to_list(mnt_type, "mnt_list"):
                 ^^^^^^^^^
  File "/home/patrycja/Downloads/volatility3/volatility3/framework/objects/__init__.py", line 971, in __getattr__
    raise AttributeError(
AttributeError: StructType has no attribute: symbol_table_name1!mnt_namespace.list
```

gcmoreira:
In the context of volatilityfoundation/volatility3#1187, this PR adds support for the mount namespace in kernels 6.8+. A basic Red-Black tree abstraction was also introduced to enable reuse in other parts of the framework if necessary.

```shell
./vol.py -f ./ubuntu_2410_x86-64_6.8.0-31.core linux.mountinfo.MountInfo 
Volatility 3 Framework 2.8.0
MNT_NS_ID       MOUNT ID        PARENT_ID       MAJOR:MINOR     ROOT    MOUNT_POINT     MOUNT_OPTIONS   FIELDS  FSTYPE  MOUNT_SRC       SB_OPTIONS

4026531841      55      25      0:7     /       /sys/kernel/debug       rw,nosuid,nodev,noexec,relatime shared:15       debugfs debugfs rw
4026531841      31      25      0:6     /       /sys/kernel/security    rw,nosuid,nodev,noexec,relatime shared:7        securityfs      securityfs  rw
4026531841      27      30      0:5     /       /dev    rw,nosuid,relatime      shared:2        devtmpfs        udev    rw
4026531841      25      30      0:23    /       /sys    rw,nosuid,nodev,noexec,relatime shared:6        sysfs   sysfs   rw
4026531841      1       1       0:2     /       /       rw              rootfs  rootfs  rw
...
```

### DIFF
--- a/volatility3/framework/symbols/linux/__init__.py
+++ b/volatility3/framework/symbols/linux/__init__.py
@@ -425,3 +425,36 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
         kernel = context.modules[kernel_module_name]
 
         return kernel
+
+
+class RBTree(object):
+    """Simple Red-Black tree abstraction"""
+
+    def __init__(self, root):
+        self.root = root
+
+    def _walk_nodes(self, root_node) -> Iterator[int]:
+        """Traverses the Red-Black tree from the root node and yields a pointer to each
+        node in this tree.
+
+        Args:
+            root_node: A Red-Black tree node from which to start descending
+
+        Yields:
+            A pointer to every node descending from the specified root node
+        """
+        if not root_node:
+            return
+
+        yield root_node
+        yield from self._walk_nodes(root_node.rb_left)
+        yield from self._walk_nodes(root_node.rb_right)
+
+    def get_nodes(self) -> Iterator[int]:
+        """Yields a pointer to each node in the Red-Black tree
+
+        Yields:
+            A pointer to every node in the Red-Black tree
+        """
+
+        yield from self._walk_nodes(root_node=self.root.rb_node)
--- a/volatility3/framework/symbols/linux/__init__.py
+++ b/volatility3/framework/symbols/linux/__init__.py
@@ -46,6 +46,7 @@ class LinuxKernelIntermedSymbols(intermed.IntermediateSymbolTable):
         # Might not exist in older kernels or the current symbols
         self.optional_set_type_class("mount", extensions.mount)
         self.optional_set_type_class("mnt_namespace", extensions.mnt_namespace)
+        self.optional_set_type_class("rb_root", extensions.rb_root)
 
         # Network
         self.set_type_class("net", extensions.net)
@@ -425,36 +426,3 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
         kernel = context.modules[kernel_module_name]
 
         return kernel
-
-
-class RBTree(object):
-    """Simple Red-Black tree abstraction"""
-
-    def __init__(self, root):
-        self.root = root
-
-    def _walk_nodes(self, root_node) -> Iterator[int]:
-        """Traverses the Red-Black tree from the root node and yields a pointer to each
-        node in this tree.
-
-        Args:
-            root_node: A Red-Black tree node from which to start descending
-
-        Yields:
-            A pointer to every node descending from the specified root node
-        """
-        if not root_node:
-            return
-
-        yield root_node
-        yield from self._walk_nodes(root_node.rb_left)
-        yield from self._walk_nodes(root_node.rb_right)
-
-    def get_nodes(self) -> Iterator[int]:
-        """Yields a pointer to each node in the Red-Black tree
-
-        Yields:
-            A pointer to every node in the Red-Black tree
-        """
-
-        yield from self._walk_nodes(root_node=self.root.rb_node)
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -1326,7 +1326,7 @@ class mnt_namespace(objects.StructType):
             vmlinux = linux.LinuxUtilities.get_module_from_volobj_type(
                 self._context, self
             )
-            for node in linux.RBTree(self.mounts).get_nodes():
+            for node in self.mounts.get_nodes():
                 mnt = linux.LinuxUtilities.container_of(
                     node, "mount", "mnt_list", vmlinux
                 )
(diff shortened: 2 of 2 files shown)


gcmoreira:
Sorry for the last minute improvement. I realized that an extension object was more appropriate here, since it was already attached to the **rb_root** type via **rb_node**, **rb_left**, and **rb_right**. 

It is now ready for review

ikelos (approved):
Looks good, thanks.

ikelos:
Reading #1187, does this PR require symbols generated with a patched `dwarf2json`?  I think we have ways of checking the version of a symbol table, but I can't tell whether we need to for this PR or not?

gcmoreira:
@ikelos There's a bug in dwarf2json that prevents generating valid ISFs. Based on [#63](https://github.com/volatilityfoundation/dwarf2json/issues/63) and [#57](https://github.com/volatilityfoundation/dwarf2json/issues/57), it affects Linux kernel version 6.5+, which is pretty bad. Since this ticket addresses changes in the mount namespace for kernels 6.8+, it's indirectly affected.

[One](https://github.com/volatilityfoundation/dwarf2json/issues/57#issuecomment-2198205637) of the comments suggests a workaround that’s currently awaiting review. I'm not sure if it's the best way to fix it, but it worked for me.

Do you want to block this until they fix that bug in dwarf2json and make it dependent on the new version?

ikelos:
Well, what I'm trying to avoid is users using an old symbol table, hitting whatever the bug is and then filing a new bug for a known problem.  The question is how best to warn about the bad table.  We can either do it on load of the bad table ("warning, you need to regenerate this using dwarf2json at least x.x.x") or whether we only do the check in known fields.  I don't think our cache database keeps the metadata (which perhaps it should, since it'll have read all the symbols at least once) so that we could do full sweeps of their stores (the warnings may get annoying if we do that on every run and they don't care though?).  The other options are to only flag on particular plugins that are affected, or to just do nothing and have vol failed on bad symbol tables without much notice.  I'd prefer to avoid that last one unless it's a vanishingly small percentage change of affecting users (and we're willing to handle bugs people bring up about it).

I think we've already got some code somewhere that checked for a symbol table's version, I think that was a per plugin check, lemme see if I can find it...  Yeah, it's the `kmsg` plugin, so at the moment that's our template for dealing with it, I guess?

https://github.com/volatilityfoundation/volatility3/blob/9bba6287a22bf93a99834834e57edb772cfac9ab/volatility3/framework/plugins/linux/kmsg.py#L520

So we've got a `self.context.symbol_space.verify_table_versions` method for checking both creator and version of the symbol table.  I guess we stick that on plugins that may be affected by it?  Does that sound a reasonable solution @gcmoreira ?

gcmoreira:
Yeah, the dwarf2json issue hasn't been fixed yet, so there's no specific version to use for that check. There's no guarantee that it will be fixed in the next version, so using something like `version > current dwarf2json version`  might not work either.
Also, this issue isn't limited to this plugin or the filesystem subsystem, so checking that here won't prevent users from filing new bug reports. As a proof of this, see for instance this note in Abyss-W4tcher's [volatility3-symbols](https://github.com/Abyss-W4tcher/volatility3-symbols?tab=readme-ov-file#190724)

gcmoreira:
Resolved conflict following the recent merge of [Linux Page Cache and IDR](https://github.com/volatilityfoundation/volatility3/pull/1233)

gcmoreira:
@ikelos https://github.com/volatilityfoundation/dwarf2json/pull/66 and https://github.com/volatilityfoundation/dwarf2json/issues/57 are done. Regarding our last comments, you want me to lock this to the dwarf2json 0.9.0 version?

ikelos:
Yeah, just wondering if it needs to be across all plugins or just specific ones?  I made PR #1305 without remembering about the function I already added, but I dunno whether it should be applied in a plug-in but phone basis (plus my PR breaks tests so I need to figure that out first...)

gcmoreira:
The dwarf2json issue is not limited to this plugin, it could happen with any type that has a binding in Rust.  If we want to add this check, it needs to be across all plugins.

For instance:

```c
$ pahole ./vmlinux-6.8.0-41-generic -C sockaddr
struct sockaddr {
	sa_family_t                sa_family;            /*     0     2 */
	union {
		char               sa_data_min[14];      /*     2    14 */
		struct {
			struct {
			} __empty_sa_data;               /*     2     0 */
			char       sa_data[0];           /*     2     0 */
		};                                       /*     2     0 */
	};                                               /*     2    14 */

	/* size: 16, cachelines: 1, members: 2 */
	/* last cacheline: 16 bytes */
};
```

```c
$ pahole ./vmlinux-6.8.0-41-generic --lang rust -C sockaddr
struct sockaddr {
	u16                        sa_family __attribute__((__aligned__(2))); /*     0     2 */
	struct sockaddr__bindgen_ty_1 __bindgen_anon_1 __attribute__((__aligned__(1))); /*     2    14 */

	/* size: 16, cachelines: 1, members: 2 */
	/* forced alignments: 2 */
	/* last cacheline: 16 bytes */
} __attribute__((__aligned__(2)));
```

As you can see above, the Rust bindings don't have all members. Another example:
```shell
$ pahole ./vmlinux-6.8.0-41-generic -C vm_area_struct | grep vm_start | wc -l
1

$ pahole ./vmlinux-6.8.0-41-generic --lang rust -C vm_area_struct | grep vm_start | wc -l
0
```

If `dwarf2json` ends up using one of those bindings instead of the original kernel type to generate the ISF, Volatility3 will eventually fail at some point.

ikelos:
This is dependent on #1305 then.  Looks like we've got a pretty good workable solution, we just need to get it developed and merged in...

gcmoreira:
I agree it could be better if #1305 goes first.. but technically it doesn't depend on that one. 
I wonder if it's possible to merge this as is? It’s been around for a while now, pretty much every distro currently has kernels >= 6.8 and there's another bug report on the way that depends on this merge to be triggered

ikelos (approved):
Looks good, sorry about the delay in reviewing.

PR #1239, opened 2024-08-16
Linux: Add lsblk plugin

dalonjon:
This pull request adds a plugin that mimics the linux lsblk utility. It loops through the block devices on the machine and prints out the device name, major, minor, removability, if the device is read only, type and mountpoint.

Tested on linux versions 3.13-6.8 with Ubuntu, Rhel, Centos, Debian, sles, and Oracle Linux distros. Worked with the latest version for each distro.

Here is an example invocation of this plugin:
`% ./vol.py -f data6.5-RaidTest.lime lsblk
Volatility 3 Framework 2.6.1
NAME    MAJOR   MINOR   RM      RO      SIZE    TYPE    MOUNTPOINT

loop0   7       0       0       1       4.0K    loop    /snap/bare/5
loop1   7       1       0       1       74.2M   loop    /snap/core22/1122
loop2   7       2       0       1       266.6M  loop    /snap/firefox/3836
loop3   7       3       0       1       497.0M  loop    /snap/gnome-42-2204/141
loop4   7       4       0       1       74.2M   loop    /snap/core22/1439
loop5   7       5       0       1       268.4M  loop    /snap/firefox/4650
loop6   7       6       0       1       12.3M   loop    /snap/snap-store/959
loop7   7       7       0       1       40.4M   loop    /snap/snapd/20671
sdb     8       16      0       0       32.0G   disk
sda     8       0       0       0       32.0G   disk
sdc     8       32      0       0       25.0G   disk
sdb1    8       17      0       0       32.0G   part
sdc1    8       33      0       0       25.0G   part
sda1    8       1       0       0       1.0M    part
sda2    8       2       0       0       513.0M  part    /boot/efi
sda3    8       3       0       0       31.5G   part    /
md0     9       0       0       0       25.0G   raid1
sr0     11      0       1       0       1024.0M rom
loop8   7       8       0       1       91.7M   loop    /snap/gtk-common-themes/1535
loop10  7       10      0       1       452.0K  loop    /snap/snapd-desktop-integration/83
loop9   7       9       0       1       38.8M   loop    /snap/snapd/21759`


volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
+from typing import List
```
ikelos, line 4: Please add the standard header (and ensure you're happy with the copyright) so that we can accept this.

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+        subsys_off = vmlinux.get_type("subsys_private").relative_child_offset("subsys")
+        kobj_off = vmlinux.get_type("kset").relative_child_offset("kobj")
+
+        if not vmlinux.has_type("hd_struct"):
+            block_device_offset = vmlinux.get_type(
+                "block_device"
+            ).relative_child_offset("bd_device")
+        else:
+            hd_struct_offset = vmlinux.get_type("hd_struct").relative_child_offset(
+                "__dev"
```
ikelos: Is this fragile or does `__dev` exist in all versions of `hd_struct`?

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+            )
+            gendisk_offset = vmlinux.get_type("gendisk").relative_child_offset("part0")
+
+        for kobj in class_kset.list.to_list(
+            vmlinux.symbol_table_name + constants.BANG + "kobject", "entry"
+        ):
+            subsys_private = vmlinux.object(
+                object_type="subsys_private",
+                offset=kobj.vol.offset - kobj_off - subsys_off,
```
ikelos: I'm having a little bit of a time getting my head around the maths here.  So we've got a kobj which lives within a `subsys` (whose first member is a `kset`) and we want to get back to the containing `subsys_private`?  Just a comment saying something like:

```
We want to go from `kobj` to `subsys_private` and the structures are laid out as:

subsys_private
|  ...
+--> subsys
  +--> kset
    +--> kobj
  |  ...
```
  
I'd probably also be tempted to calculate the `kset` location from `subsys` even if that always equates to 0, just so we're not making any assumptions here.  I might also have gotten what's going on completely wrong, in which case we still need documentation...  5;)

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+                # lsblk by default skips over ram devices
+                try:
+                    if utility.array_to_string(gendisk.disk_name).startswith("ram"):
+                        continue
+                except exceptions.InvalidAddressException:
+                    continue
+            else:
+                hd_struct = vmlinux.object(
+                    object_type="hd_struct",
+                    offset=device.vol.offset - hd_struct_offset,
```
ikelos: Again, this seems like we're working backwards from a `device` to an `hd_struct`?  A quick comment to explain that's what we're doing would be handy (it'll also help everyone make sure that what we're trying to do is what we're actually doing).

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+        # This the hash functions used to get the inode number https://elixir.bootlin.com/linux/v5.4/source/fs/block_dev.c#L867
+        val = self._major(devt) + self._minor(devt)
+
+        # Need these next 4 values for this hash function: https://elixir.bootlin.com/linux/v5.4/source/fs/inode.c#L474
+        i_hash_mask = vmlinux.object_from_symbol("i_hash_mask")
+        i_hash_shift = vmlinux.object_from_symbol("i_hash_shift")
+
+        # in linux 4.7, the GOLDEN_RATIO_PRIME constant changes. This struct gains a member from 4.6 to 4.7
+        if vmlinux.get_type("file_operations").has_member("iterate_shared"):
+            GOLDEN_RATIO_PRIME = 0x61C8864680B583EB
```
ikelos: Feels like these should be stored in the symbol table itself as a constant.  Are these not symbol values?  Such specific hard coded values are really bad for maintenance and things changing in the future...

ikelos: At the very least, this should live in `constants.linux` and have different names that can then be chosen by a plugin depending on which version the plugin decides to use.

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+            )
+            # These checks are from the find_inode function https://elixir.bootlin.com/linux/v5.4/source/fs/inode.c#L805
+            # Trying to find the inode with the correct device number and calculated inode number from the inode cache
+            if inode.i_sb == blockdev_superblock and inode.i_rdev == devt:
+                break
+            hlist_node = hlist_node.next
+
+        return inode.i_bdev
+
+    def _device_iterator(self, vmlinux, klist_devices):
```
ikelos: As I've said on other PRs (but I think this is one of the first times I've said it here), we much prefer `context` being the first parameter so that this can be made into a classmethod if necessary, and then simple types (str, int, etc) where possible for the other parameters (such as a module_name, which can then be looked up in the context).  Complex types (like ObjectStructures and so on) can be passed, but we're trying to keep them to a minimum.

ikelos: Also, type hints would be much appreciated!

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+                    offset=mapped_device.interface_ptr,
+                    absolute=True,
+                )
+
+                if hash_cell.name:
+                    return utility.pointer_to_string(hash_cell.name, 32)
+
+        return utility.array_to_string(gendisk.disk_name)
+
+    def _format_bytes(self, size):
```
ikelos: Feels like this could be used for more than just the `linux.lsblk` function, consider adding it as a staticmethod somewhere more common across plugins.  As mentioned above, this is really a UI function rather than a plugin function.  The plugin should return the raw data, and let the UI decide what needs to happen to display it properly (using a type hint to say how you'd expect it to be displayed).

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+
+            mountpoint = self._get_mountpoint(vmlinux, block_device)
+
+            yield 0, (
+                name,
+                self._major(device.devt),
+                self._minor(device.devt),
+                removable,
+                read_only,
+                self._format_bytes(size),
```
ikelos: Please don't format raw data that's returned from the plugin.  This might be being used to another plugin or some other code, so turning it into a string for a human to read creates *more* work for other plugins.  The UI should figure out how such things should be formatted.  If you want something like this it should inherit from `int` in the `format_hints` file, that way it only affects the CLI and a web user interface might find a better way of being able to display the abbreviated value but also let the user see the full value (via a tooltip for example).

I think I'm going to say this one *is* show stopper and the plugin won't be accepted until this has been changed.  This *must* return the full data as an integer so that other tools can use the value.  The format hint can be added later as part of a separate PR, or as part of this one.  Either would be acceptable.

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+            return f"{size}B"
+        elif size < 1024**2:
+            return f"{size / 1024:.1f}K"
+        elif size < 1024**3:
+            return f"{size / 1024 ** 2:.1f}M"
+        elif size < 1024**4:
+            return f"{size / 1024 ** 3:.1f}G"
+        return f"{size / 1024 ** 4:.1f}T"
+
+    def _get_type(self, vmlinux, block_device, gendisk):
```
ikelos: Again, type information and ideally context first and simpler parameters where possible afterwards.

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+                return self._get_scsi_device_type(scsi_cd.device.type)
+
+        return "disk"
+
+    def _get_scsi_device_type(self, type_code):
+        """
+        SCSI device types.  Copied almost as-is from kernel header
+        (include/scsi/scsi_proto.h)
+        """
+        type_map = {
```
ikelos, line 315: This is kinda ok here, but if anything else ever would be able to make use of it, it should be lifted out into `constants.linux` somewhere (if it doesn't already exist as an enumeration in linux symbol tables).

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+            if result:
+                return result
+            return "/"
+
+        return result
+
+    def run(self):
+
+        columns = [
+            ("NAME", str),
```
ikelos: Why are we shouting all the column names?  5;P  These can just be normal capitalized values like other plugins.

ikelos (changes requested):
Thanks very much for your submission, there's quite a lot going on so I can't guarantee these are all the comments to be made, but it should be enough to be getting on with.  There is at least one show stopper, where some results are converted to strings before being returned from the plugin, which is bad from the perspective of allowing other tools to use the data this plugin returns.  That will need to change before this is accepted, but most everything else looks pretty good.  There's a lot of backtracking from a subtype to a parent type, which ok, but could do with a lot of documentation so people can see/verify what's going on.

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+import logging
+
+from volatility3.framework import interfaces, renderers, constants, exceptions
+from volatility3.framework.configuration import requirements
+from volatility3.framework.objects import utility
+
+vollog = logging.getLogger(__name__)
+
+
+class lsblk(interfaces.plugins.PluginInterface):
```
gcmoreira: CapWords please

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+                absolute=True,
+            )
+            if subsys_private.member("class") == block_class.vol.offset:
+                break
+
+        klist_devices = subsys_private.klist_devices
+        for device in self._device_iterator(vmlinux, klist_devices):
+            # Before v5.11, partitions are represented by hd_struct instead of block_device
+            if not vmlinux.has_type("hd_struct"):
+                block_device = vmlinux.object(
```
gcmoreira: I think you can use `LinuxUtilities.container_of()` here and in the others.

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+                1
+                if (
+                    gendisk.part0.bd_read_only
+                    or (gendisk.state & (1 << GD_READ_ONLY)) != 0
+                )
+                else 0
+            )
+        else:
+            read_only = gendisk.part0.policy
+        return read_only
```
gcmoreira: could you please use a bool here instead of 0 and 1?

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,447 @@
...
+        return result
+
+    def run(self):
+
+        columns = [
+            ("NAME", str),
+            ("MAJOR", int),
+            ("MINOR", int),
+            ("RM", int),
+            ("RO", int),
```
gcmoreira: it's better to use a bool for removable and read-only

volatility3/framework/plugins/linux/lsblk.py
gcmoreira: +1 for adding a `format_hints` for this, that will be nice

volatility3/framework/plugins/linux/lsblk.py
dalonjon: It is in all versions of hd_struct

volatility3/framework/plugins/linux/lsblk.py
dalonjon: I changed this part to use linux.LinuxUtilities.container_of
I think that makes it a lot easier to follow

volatility3/framework/plugins/linux/lsblk.py
dalonjon: I added it to constants.linux

volatility3/framework/plugins/linux/lsblk.py
dalonjon: I added it to format_hints

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,388 @@
...
+
+vollog = logging.getLogger(__name__)
+
+
+class Lsblk(interfaces.plugins.PluginInterface):
+    """Lists the block devices present in a particular linux memory image."""
+
+    _required_framework_version = (2, 0, 0)
+
+    _version = (2, 2, 1)
```
gcmoreira: This is a new plugin, you should use (1, 0 ,0)

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,388 @@
...
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Linux kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.VersionRequirement(
+                name="linuxutils", component=linux.LinuxUtilities, version=(2, 0, 0)
```
gcmoreira: linux.LinuxUtilities is already in version=(2, 1, 0)

volatility3/framework/plugins/linux/lsblk.py
atcuno: It is matching the lsblk command on the live system:

```
$ lsblk
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda      8:0    0   40G  0 disk 
├─sda1   8:1    0   39G  0 part /
├─sda2   8:2    0    1K  0 part 
└─sda5   8:5    0  975M  0 part [SWAP]
sdb      8:16   0   40G  0 disk 
└─sdb1   8:17   0   40G  0 part /mnt/samples
sr0     11:0    1 1024M  0 rom  
```

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,388 @@
...
+#
+from typing import List
+import logging
+
+from volatility3.framework import interfaces, renderers, constants, exceptions
+from volatility3.framework.configuration import requirements
+from volatility3.framework.objects import utility
+from volatility3.framework.symbols import linux
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.constants.linux import GOLDEN_RATIO_PRIME_BEFORE_4_7, GOLDEN_RATIO_PRIME_AFTER_4_7
```
ikelos, line 12: Generally we import modules, rather than specific values, so that if people find it in this code and are lazy, they can't import this plugin more easily than importing the module they need.  So please import either `constants` or `linux` and then use either `constants.linux.GOLDEN...` or `linux.GOLDEN...` respectively (I think I'd prefer constants).

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,388 @@
...
+                    absolute=True,
+                )
+
+                if hash_cell.name:
+                    return utility.pointer_to_string(hash_cell.name, 32)
+
+        return utility.array_to_string(gendisk.disk_name)
+
+    def _get_type(self, context, block_device, gendisk):
+        vmlinux = context.modules[self.config["kernel"]]
```
ikelos, line 239: Rather than limiting this function to being usable during a run of this plugin, could we add a parameter for `kernel_module_name` please?  This would then allow a lot of these to become classmethods, and make then usable by other plugins in the future.

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,388 @@
...
+            # These checks are from the find_inode function https://elixir.bootlin.com/linux/v5.4/source/fs/inode.c#L805
+            # Trying to find the inode with the correct device number and calculated inode number from the inode cache
+            if inode.i_sb == blockdev_superblock and inode.i_rdev == devt:
+                break
+            hlist_node = hlist_node.next
+
+        return inode.i_bdev
+
+    def _device_iterator(self, context, klist_devices):
+        vmlinux = context.modules[self.config["kernel"]]
```
ikelos, line 187: Please could we parameterize this on `kernel_module_name` rather than going to the config?  It'll make the function more usable outside of the bounds of just this plugin.

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,388 @@
...
+                device_private = linux.LinuxUtilities.container_of(klist_node.vol.offset, "device_private", "knode_class", vmlinux)
+                yield device_private.device.dereference()
+        else:
+            for klist_node in klist_devices.k_list.to_list(
+                vmlinux.symbol_table_name + constants.BANG + "klist_node", "n_node"
+            ):
+                device = linux.LinuxUtilities.container_of(klist_node.vol.offset, "device", "knode_class", vmlinux)
+                yield device
+
+    def _get_name(self, vmlinux, block_device, gendisk):
```
ikelos, line 206: Also, rather than passing in vmlinux, the convention for volatility plugins is to pass in the context as the first parameter and then the module name as the second, and then lookup the module in `context.modules`.  Sticking to the convention should make it easier if/when we come to parallelize things rather than passing around lots of objects, it's just the context and then a reference for something inside the context.

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,388 @@
...
+            inode = linux.LinuxUtilities.container_of(hlist_node, "inode", "i_hash", vmlinux)
+            # These checks are from the find_inode function https://elixir.bootlin.com/linux/v5.4/source/fs/inode.c#L805
+            # Trying to find the inode with the correct device number and calculated inode number from the inode cache
+            if inode.i_sb == blockdev_superblock and inode.i_rdev == devt:
+                break
+            hlist_node = hlist_node.next
+
+        return inode.i_bdev
+
+    def _device_iterator(self, context, klist_devices):
```
ikelos, line 186: Would it be possible to provide basic docstring describing what these functions do and what their input arguments are and what output they produce please?

ikelos (commented):
Looking much better and the code is much easier to read through now, thanks!  Just a few more points but looking really close to the finish line.  5:)

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,388 @@
...
+    def _generator(
+        self,
+    ):
+        """Generates the list of block devices."""
+        vmlinux = self.context.modules[self.config["kernel"]]
+        try:
+            class_kset = vmlinux.object_from_symbol("class_kset")
+        except exceptions.SymbolError:
+            class_kset = None
+        if not class_kset:
```
gcmoreira, line 46: Please use `vmlinux.has_symbol()` instead of all these

volatility3/framework/plugins/linux/lsblk.py
```diff
@@ -0,0 +1,388 @@
...
+                gendisk = linux.LinuxUtilities.container_of(hd_struct.vol.offset, "gendisk", "part0", vmlinux)
+                try:
+                    if utility.array_to_string(gendisk.disk_name).startswith("ram"):
+                        continue
+                except exceptions.InvalidAddressException:
+                    continue
+
+                block_device = self._get_block_device(vmlinux, device.devt)
+                if not block_device:
+                    continue
```
gcmoreira, line 85: It seems like you could refactor this into an internal helper function, such as `_get_block_device()`, which would return either the `block_device` or `None`. This would help make the large for statement more readable and maintainable.

gcmoreira: Oops, I just noticed you already have a `_get_block_device()` method, but it seems to only work for kernels 5.11 and up, which adds to the confusion. You'll need to rethink how to refactor this. 
Ideally, there should be a single `_get_block_device()` method, and based on the kernel version (the type checks you are doing in the above lines), it can either call the method for kernels >= 5.11 or the one for < 5.11. This would make the logic much clearer.

volatility3/framework/plugins/linux/lsblk.py
Abyss-W4tcher: Could this helper make it in the core API please ? This can definitely be useful for any kind of device parser plugin 👍

PR #1242, opened 2024-08-24
Remove get_vmlinux calls

ikelos:
Sorry, I know I asked for it, but I hadn't quite figured out what was going on.  Things in the symbols/linux code are considered part of the framework, and therefore it's the framework version that should have been bumped.  Since the framework comes packages with LinuxUtilities we can rely on the version numbers to be suitable.  This cleans up the mess I caused, sorry for the extra work!  5:S


gcmoreira:
Awesome! Yeah, it looked horrible to me too lol, but I just wanted to make the changes so you could see how it turned out. Thanks for doing this for me. It looks good!

PR #1249, opened 2024-09-02
Mft size smear fix

atcuno:


volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -27,6 +27,14 @@ class MFTAttribute(objects.StructType):
     """This represents an MFT ATTRIBUTE"""
 
     def get_resident_filename(self) -> str:
```
ikelos: This should probably have been returning `Optional[str]` all along.  Any chance we could fix that up please?

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -42,6 +50,14 @@ def get_resident_filename(self) -> str:
             return None
 
     def get_resident_filecontent(self) -> bytes:
```
ikelos: Same for this `-> Optional[bytes]:`

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -42,6 +50,14 @@ def get_resident_filename(self) -> str:
...
+        # smear observed in mass testing of samples
+        # 4MB chosen as cutoff instead of 4KB to allow for recovery from format /L created file systems
+        if (
+            self.Attr_Header.ContentOffset > 4194304
```
ikelos: I kinda prefer these as `4 * 0x1000 * 0x1000`, since the number you've got there looks like it's actually only 400KB?  Could you make the values match (and ideally write it in hex, so it's easier to check that it's right visually)?

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -27,6 +27,14 @@ class MFTAttribute(objects.StructType):
     """This represents an MFT ATTRIBUTE"""
 
     def get_resident_filename(self) -> str:
+        # 4MB chosen as cutoff instead of 4KB to allow for recovery from format /L created file systems
+        # Length as 512 as its 256*2, which is the maximum size for an entire file path, so this is even generous
+        if (
+            self.Attr_Header.ContentOffset > 4194304
```
ikelos: Same as below, the number you're using look like 0x400000...

ikelos (commented):
Sorry, I was hoping this would be trivial but just a couple minor points.  Happy to merge it after that though...

volatility3/framework/symbols/windows/extensions/mft.py
atcuno: I don't mind doing this one time unrelated thing, but going forward I think a better/good use of time would be integrating mypy into the set of PR checks. mypy reported this error and the other one auto when I checked this morning:

```
$ mypy --disallow-untyped-defs --disallow-incomplete-defs --check-untyped-defs --follow-imports=skip volatility3/framework/symbols/windows/extensions/mft.py
volatility3/framework/symbols/windows/extensions/mft.py:36: error: Incompatible return value type (got "None", expected "str")  [return-value]
volatility3/framework/symbols/windows/extensions/mft.py:50: error: Incompatible return value type (got "None", expected "str")  [return-value]
volatility3/framework/symbols/windows/extensions/mft.py:59: error: Incompatible return value type (got "None", expected "bytes")  [return-value]
volatility3/framework/symbols/windows/extensions/mft.py:72: error: Incompatible return value type (got "None", expected "bytes")  [return-value]
Found 4 errors in 1 file (checked 1 source file)
```

The only downside to mypy checking the type usage is almost every file in vol3 would fail now, which would break other PRs, so we need to think of a smart way to do it. @dgmcdona would have smart ideas on this.

atcuno: @ikelos would an issue on the tracker here be good to work towards the mypy goal?

atcuno:
All feedback addressed @ikelos

volatility3/framework/symbols/windows/extensions/mft.py
ikelos: I did try getting as much typed as possible several years ago.  The issue then was that `ObjectInterface` gives you attributes that can't easily be checked (because we use `__getattr__` to make the object match the types dynamically based on the ISF).  A such, a lot of the typing will quickly boil down to `something something something ObjectInterface something something` which isn't really helpful and is why I don't tend to run mypy over the codebase any more.  Last time I did it I think we were at about 30 lines that didn't mention ObjectInterface.  The is one of the reasons I prefer passing Context and then simple types.

If you've got a way to suitably get the typing so it's accurate *and* useful then I'd be happy to try and get it applied to the whole codebase, but I'm not sure it's feasible.  That's not a good reason or excuse not to try and improve the types when we spot them (which is why I pointed it out here), just that a lot of the codebase will end up un-typabale..  So it'd be a big project, but if you're up for it, I'm happy to review whatever changes you come up with.  It will likely also pepper the code with None checks, which will make it safer (and potentially one day compilable) but also messier to read.

Mypy currently gives at "Found 876 errors in 164 files (checked 234 source files)" across the whole codebase, for reference...

It also had issues handling some of our more interesting namespace games that we play to get `framework.plugins` effectively mounted as `plugins` so that user code could override core code (not that I'm aware of many people using it).

In short, it's a lot of work and I'd want to be a bit more sure it would be feasible, but I'm all for having as much good typing in the code as we can get...

PR #1253, opened 2024-09-08
CLI: Filter on rendered values

### LINKED ISSUE #1251: Make better name/constant in --filters for not available/applicable columns
@ikelos we talked about this, but need a ticket to have it actually happen.

The issue was that there is no direct mechanism, besides specifying `Value` to filter to a column's value that is NotAvailable or NotApplicable. This is problematic as many of our malware reporting plugins will report blank columns when malware is present, so users will often want to see only these rows in the output - especially for plugins that otherwise produce substantial output.

ldrmodules is a perfect example:

```
$ python3 vol.py -f /mnt/samples/Sample14.lime windows.ldrmodules | wc -l
4624         
```

So 4,624 lines of output versus filtering to what is just the 7 lines related to the malware when using 'Value':

```
$ python3 vol.py --filters MappedPath,Value -f /mnt/samples/Sample14.lime windows.ldrmodules
Volatility 3 Framework 2.9.0
Pid	Process	Base	InLoad	InInit	InMem	MappedPath

836	lsass.exe	0x14894ca0000	False	False	False	N/A
836	lsass.exe	0x14894970000	False	False	False	N/A
836	lsass.exe	0x148949b0000	False	False	False	N/A
836	lsass.exe	0x14894c00000	False	False	False	N/A
4684	mi_x64.exe	0x4f0000	False	False	False	N/A
4684	mi_x64.exe	0x5a0000	False	False	False	N/A
4684	mi_x64.exe	0x2db0000	False	False	False	N/A
```

To get this fixed, we just need to come up with a better constant than 'Value'

### LINKED ISSUE #1252: Support hex digits in --filters for integer columns
A big stumbling point for training attendees at Black Hat and DFRWS was that --filters doesn't understand hex addresses. All of our plugins report addresses in hex, but currently --filters only understands decimal integer input.

Example:

```
$ python3 vol.py --filters Base,0x148949b0000 -f /mnt/samples/Sample14.lime windows.ldrmodules
```

No output ^, but converting that number to decimal does get the right line:

```
$ python3 vol.py --filters Base,1411242459136 -f /mnt/samples/Sample14.lime windows.ldrmodules        
Pid	Process   Base                       InLoad	InInit	InMem	MappedPath
836	lsass.exe 0x148949b0000   False	       False         False
```

@ikelos how difficult would it be to fix this? Ideally, we could make it where "0x" prefixed values are treated as hex for 'int' columns or worst case just make all "0x" inputs be treated as hex numbers.

ikelos:
This now filters JSON, JSONL and CSV as well as quick and pretty rendering.  The filtering is now carried out on the rendered values, this will allow a rudimentary filtering on AbsentValues but will require knowing how the specific absent values will be rendered into text.  It will also *no longer* allow filtering of values that are displayed as hex by using the integer value.

This also kinda needs us to bump a version number so we can tell if people are using a different version with filtering or not.  It could probably have been a patch version, but I think people would better understand the CLI changing syntax/behaviour with a MINOR bump at least.

Fixes #1252

### DIFF
--- a/volatility3/cli/text_renderer.py
+++ b/volatility3/cli/text_renderer.py
@@ -180,8 +180,7 @@ class QuickTextRenderer(CLIRenderer):
 
         def visitor(node: interfaces.renderers.TreeNode, accumulator):
             line = []
-            for column_index in range(len(grid.columns)):
-                column = grid.columns[column_index]
+            for column_index, column in enumerate(grid.columns):
                 renderer = self._type_renderers.get(
                     column.type, self._type_renderers["default"]
                 )
@@ -261,8 +260,7 @@ class CSVRenderer(CLIRenderer):
             # Nodes always have a path value, giving them a path_depth of at least 1, we use max just in case
             row = {"TreeDepth": str(max(0, node.path_depth - 1))}
             line = []
-            for column_index in range(len(grid.columns)):
-                column = grid.columns[column_index]
+            for column_index, column in enumerate(grid.columns):
                 renderer = self._type_renderers.get(
                     column.type, self._type_renderers["default"]
                 )
@@ -326,8 +324,7 @@ class PrettyTextRenderer(CLIRenderer):
 
             line = {}
             rendered_line = []
-            for column_index in range(len(grid.columns)):
-                column = grid.columns[column_index]
+            for column_index, column in enumerate(grid.columns):
                 renderer = self._type_renderers.get(
                     column.type, self._type_renderers["default"]
                 )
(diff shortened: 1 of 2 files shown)


volatility3/cli/text_renderer.py
```diff
@@ -179,7 +179,15 @@ def render(self, grid: interfaces.renderers.TreeGrid) -> None:
         outfd.write("\n{}\n".format("\t".join(line)))
 
         def visitor(node: interfaces.renderers.TreeNode, accumulator):
-            if self.filter and self.filter.filter(node.values):
+            line = []
+            for column_index in range(len(grid.columns)):
```
gcmoreira: enumerate(grid.columns)?

ikelos: Thanks, added  5:)

PR #1254, opened 2024-09-08
Core: Verify plugin requirements of plugins

ikelos:
Make all VersionRequirements verify any sub-version requirements the versioned components may depend on.

Should be minimal impact other than improving errors that may happen down the line.  Likely to merge in a couple of days unless anyone shouts.

### DIFF
--- a/volatility3/framework/configuration/requirements.py
+++ b/volatility3/framework/configuration/requirements.py
@@ -585,6 +585,24 @@ class PluginRequirement(VersionRequirement):
             version=version,
         )
 
+    def unsatisfied(
+        self, context: interfaces.context.ContextInterface, config_path: str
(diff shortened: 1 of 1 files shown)


atcuno:
@ikelos the tests failed on this one with some terse errors. Is this the validation finding places its lacking now (the point of this PR)? Or some other issue being exposed?

ikelos:
No, it's hitting `maximum recursion depth exceeded while calling a Python object`.  I suspect we've got a circular dependency somewhere (which I didn't code against), so I'll add that to it in a bit...

ikelos:
Ok, interesting.  So linux.pslist requires elfs.Elfs (to dump processes) and elfs.Elfs requires linux.pslist to run the plugin (which isn't used by any of the calls).  So now the question is, do we allow circular dependencies, as long as they're resolvable?

PR #1258, opened 2024-09-12
Timeliner plugin `plugin-filter` improvements

gcmoreira:
The current `Timeliner` plugin implementation, for every class inheriting from `TimeLinerInterface` will do the steps below. Note that it will happen even if the plugin isn't intended for the operating system being analyzed. 
- Evaluates the requirements
- Runs automagic
- Constructs and instantiates the plugin

...before finally checking if that plugin name matches the `plugin-filter` argument.  This is a waste of time when the user just want to execute just one or two plugins.

This PR optimizes the main loop by filtering plugin classes at an earlier stage, significantly reducing execution time.

### DIFF
--- a/volatility3/framework/plugins/timeliner.py
+++ b/volatility3/framework/plugins/timeliner.py
@@ -199,9 +199,10 @@ class Timeliner(interfaces.plugins.PluginInterface):
                                     ),
                                 )
                             )
-            except Exception:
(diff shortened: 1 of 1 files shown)


ikelos (approved):
Looks good, thanks for spotting this!  5:)

PR #1259, opened 2024-09-12
Linux Page Cache smear memory protection improvements

gcmoreira:
In the context of https://github.com/volatilityfoundation/volatility3/pull/1243 this PR improves the smear memory protection for the page cache plugins (`Files` and `InodePages`) and related function in `mountinfo`.
It also fixes a typo in the respective timeliner function.

### DIFF
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -104,7 +104,7 @@ class Files(plugins.PluginInterface, timeliner.TimeLinerInterface):
 
     _required_framework_version = (2, 0, 0)
 
-    _version = (1, 0, 0)
+    _version = (1, 0, 1)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
@@ -186,8 +186,12 @@ class Files(plugins.PluginInterface, timeliner.TimeLinerInterface):
 
             seen_dentries.add(dentry_addr)
 
-            inode = dentry.d_inode
-            if not (inode and inode.is_valid()):
+            inode_ptr = dentry.d_inode
+            if not (inode_ptr and inode_ptr.is_readable()):
+                continue
+
+            inode = inode_ptr.dereference()
+            if not inode.is_valid():
                 continue
 
             # This allows us to have consistent paths
@@ -242,8 +246,9 @@ class Files(plugins.PluginInterface, timeliner.TimeLinerInterface):
 
             # More dentry/inode sanity checks
             root_inode_ptr = root_dentry.d_inode
-            if not root_inode_ptr:
+            if not (root_inode_ptr and root_inode_ptr.is_readable()):
                 continue
+
             root_inode = root_inode_ptr.dereference()
             if not root_inode.is_valid():
                 continue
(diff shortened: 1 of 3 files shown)


volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -186,8 +186,12 @@ def _walk_dentry(
 
             seen_dentries.add(dentry_addr)
 
-            inode = dentry.d_inode
-            if not (inode and inode.is_valid()):
+            inode_ptr = dentry.d_inode
+            if not (inode_ptr and inode_ptr.is_readable()):
+                continue
+
+            inode = inode_ptr.dereference()
```
ikelos, line 193: Is this not already covered by `inode_ptr.is_readable`?  How do the two differ?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -843,25 +855,55 @@ def get_subdirs(self) -> interfaces.objects.ObjectInterface:
...
+        return inode_ptr.dereference()
+
 
 class struct_file(objects.StructType):
     def get_dentry(self) -> interfaces.objects.ObjectInterface:
-        if self.has_member("f_dentry"):
-            return self.f_dentry
-        elif self.has_member("f_path"):
+        """Returns a pointer to the dentry associated with this file"""
+        if self.has_member("f_path"):
```
ikelos, line 871: Same as in other code I just reviewed, will the change in order have an impact on the results returned and if so, do we need to bump a version number somewhere so people know?

ikelos (commented):
Hiya all looks good, possibly some excess code somewhere and I just want to check we don't need to bump a version number but otherwise looks nice, thanks!

volatility3/framework/plugins/linux/pagecache.py
gcmoreira: hm not sure which line are you referring to, I see this comment on line 193 which is dereferencing the pointer. This will raise an exception on invalid pointers for which you need the previous checks.
- if `inode_ptr`: only checks for `inode_ptr == 0` or `None` as we discussed in https://github.com/volatilityfoundation/volatility3/pull/1243.
- inode_ptr.is_readable(): checks that the pointer is valid in its address space.
- inode.is_valid(): Checks that the content of the inode is valid.

All of them are needed.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: I moved the `if` matching the most recent kernels to be first, which it's likely to be the most often case now and in the future.

gcmoreira: re: version.. there's no version for this

volatility3/framework/plugins/linux/pagecache.py
ikelos: Ok, just figured `is_readable` would be enough, but if they're all needed that's fine.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Ok, cool, just wanted to check.  5:)

PR #1262, opened 2024-09-12
Add detection of direct and indirect system calls

atcuno:
@ikelos these are the priority after pe_symbols. Inheritance is used correctly (versioned, no overriding generator or run, basically the same as threads/thrdscan/etc. and others), so don't freak out when you see it and just enjoy the beautiful, reusable code. I also used namedtuples as a test run and it turned out pretty cool. I left many comments, including the args/returns, and both files pass black, mypy and pylint.

This implements two more of the DEFCON plugins for detecting malware that bypasses EDRs and are required for the cutover release.

Detection of the HellsGate direct system call technique:

```
$ python3 vol.py --config halo.json windows.direct_system_calls
Volatility 3 Framework 2.10.0
Process	PID	Range	Address	Disasm

HellsGate.exe	920	\exclude_test\volexity\volexity\Saved Files\HalosGate\HellsGate.exe	0x7ff627831591	0x7ff627831591: mov r10, rcx; 0x7ff627831594: mov eax, dword ptr [rip + 0x3a66]; 0x7ff62783159a: syscall ; 0x7ff62783159c: ret ; 
```

Detection of the HellsHall indirect system call technique:

```
$ python3 vol.py --config hell.json windows.indirect_system_calls
Volatility 3 Framework 2.10.0
Process	PID	Range	Address	Disasm
HellsHall.exe	2112	\exclude_test\volexity\volexity\Saved Files\HalosGate\HellsHall.exee	0x7ff6949f165e	0x7ff6949f165e: mov r10, rcx; 0x7ff6949f1661: mov eax, dword ptr [rip + 0x3999]; 0x7ff6949f1667: jmp qword ptr [rip + 0x3997]; 
```

### DIFF
--- a/volatility3/framework/plugins/windows/direct_system_calls.py
+++ b/volatility3/framework/plugins/windows/direct_system_calls.py
@@ -49,6 +49,7 @@ class DirectSystemCalls(interfaces.plugins.PluginInterface):
     _required_framework_version = (2, 4, 0)
     _version = (1, 0, 0)
 
+    # DLLs that are expected to host system call invocations
     valid_syscall_handlers = ("ntdll.dll", "win32u.dll")
 
     def __init__(self, *args, **kwargs):
@@ -266,7 +267,7 @@ class DirectSystemCalls(interfaces.plugins.PluginInterface):
                         continue
 
                     # we only care about calls to system call DLLs
-                    path = cls._get_range_path(vads, target_address)
+                    path = cls.get_range_path(vads, target_address)
                     if not isinstance(path, str) or not path.lower().endswith(
                         cls.valid_syscall_handlers
                     ):
@@ -278,7 +279,7 @@ class DirectSystemCalls(interfaces.plugins.PluginInterface):
         return None
 
     @staticmethod
-    def _get_vad_maps(
+    def get_vad_maps(
         task: interfaces.objects.ObjectInterface,
     ) -> List[Tuple[int, int, str]]:
         """Creates a map of start/end addresses within a virtual address
@@ -304,14 +305,14 @@ class DirectSystemCalls(interfaces.plugins.PluginInterface):
         return vads
 
     @staticmethod
-    def _get_range_path(
+    def get_range_path(
         ranges: List[Tuple[int, int, str]], address: int
     ) -> Optional[str]:
         """
         Returns the path for the range holding `address`, if found
 
         Args:
-            ranges: VADs collected from `_get_vad_maps`
+            ranges: VADs collected from `get_vad_maps`
             address: the address to find
         Returns:
             The path holding the address, if any
@@ -399,7 +400,7 @@ class DirectSystemCalls(interfaces.plugins.PluginInterface):
         ):
             address = hit[0]
 
-            path = cls._get_range_path(vads, address)
+            path = cls.get_range_path(vads, address)
 
             # ignore hits in the system call DLLs
             if isinstance(path, str) and path.lower().endswith(
@@ -419,7 +420,7 @@ class DirectSystemCalls(interfaces.plugins.PluginInterface):
         ):
             proc_layer = self.context.layers[proc_layer_name]
 
-            vads = self._get_vad_maps(proc)
+            vads = self.get_vad_maps(proc)
 
             # for each valid process, look for malicious syscall invocations
             for address, vad_path in self._get_rule_hits(
--- a/volatility3/framework/plugins/windows/direct_system_calls.py
+++ b/volatility3/framework/plugins/windows/direct_system_calls.py
@@ -3,7 +3,7 @@
 #
 
 import logging
-import capstone
+
 from collections import namedtuple
 from typing import List, Tuple, Optional, Generator, Callable
 
@@ -16,6 +16,12 @@ from volatility3.plugins.windows import pslist
 
 vollog = logging.getLogger(__name__)
 
+try:
+    import capstone
+
+    has_capstone = True
+except ImportError:
+    has_capstone = False
 
 # Full details on the techniques used in these plugins to detect EDR-evading malware
 # can be found in our 20 page whitepaper submitted to DEFCON along with the presentation
@@ -114,7 +120,7 @@ class DirectSystemCalls(interfaces.plugins.PluginInterface):
 
         1) update RAX to the system call number
         2) update R10 to the first parameter
-        3) hit the 'termination' instrunction set in `syscall_finder_type`
+        3) hit the 'termination' instruction set in `syscall_finder_type`
 
         We also track whether the 'syscall' instruction was encountered while parsing
 
@@ -413,6 +419,12 @@ class DirectSystemCalls(interfaces.plugins.PluginInterface):
     def _generator(
         self,
     ) -> Generator[Tuple[int, Tuple[str, int, Optional[str], int, str]], None, None]:
+        if not has_capstone:
+            vollog.warning(
+                "capstone is not installed. This plugin requires capstone to operate."
+            )
+            return
+
         kernel = self.context.modules[self.config["kernel"]]
 
         for proc, proc_name, proc_layer_name, architecture in self.get_tasks_to_scan(
--- a/volatility3/framework/plugins/windows/direct_system_calls.py
+++ b/volatility3/framework/plugins/windows/direct_system_calls.py
@@ -304,7 +304,9 @@ class DirectSystemCalls(interfaces.plugins.PluginInterface):
         return vads
 
     @staticmethod
-    def _get_range_path(ranges: List[Tuple[int, int, str]], address: int) -> Optional[str]:
+    def _get_range_path(
+        ranges: List[Tuple[int, int, str]], address: int
+    ) -> Optional[str]:
         """
         Returns the path for the range holding `address`, if found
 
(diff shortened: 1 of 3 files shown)


volatility3/framework/plugins/windows/direct_system_calls.py
```diff
@@ -0,0 +1,455 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
```
ikelos: These seem new this year?  They should say their creation date for the copyright, so 2024 please.

volatility3/framework/plugins/windows/direct_system_calls.py
```diff
@@ -0,0 +1,455 @@
...
+        "get_syscall_target_address",
+        "wants_syscall_inst",
+        "rule_str",
+        "invalid_ops",
+        "termination_ops",
+    ],
+)
+
+syscall_finder_type.__doc__ = """
+This type to used to specify how malicious system call invocations should be found.
```
ikelos, line 42: Minor typo, `is used to` rather than `to used to`

volatility3/framework/plugins/windows/direct_system_calls.py
```diff
@@ -0,0 +1,455 @@
...
+
+        1) update RAX to the system call number
+        2) update R10 to the first parameter
+        3) hit the 'termination' instrunction set in `syscall_finder_type`
+
+        We also track whether the 'syscall' instruction was encountered while parsing
+
+        This function is reusable for every technique we found and studied during the DEFCON research timeframe
+
+        Args:
```
ikelos, line 129: Awesome docstrings, thanks!  5:D

volatility3/framework/plugins/windows/direct_system_calls.py
atcuno: Fixed

volatility3/framework/plugins/windows/direct_system_calls.py
```diff
@@ -0,0 +1,455 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import logging
+import capstone
```
ikelos: Can we add import error protection around this (try except ImportError), so if they don't have it, it'll raise a nice error telling them they need to get it?  Also, do we want to add it to one of the requirements.txt files?

volatility3/framework/plugins/windows/direct_system_calls.py
```diff
@@ -0,0 +1,455 @@
...
+            proc_layer: the memory layer of the process being scanned
+            architecture: the name of the architecture for the process being disassembled
+            vads: the ranges of this process under 10MB
+            address: the starting address to check for malicious syscall code blocks
+
+        Returns:
+            Optional[Tuple[int, str]]: For valid code blocks, the starting address of the block and the disassembly string
+        """
+        # the number bytes behind the yara rule hit to scan
+        behind = 32
```
ikelos, line 244: Can this not be encoded in the yara rule, or are we happy hard coding it here?  I guess it's about the same either way, just floating specific numbers get the hairs on the back of my neck up...  5;P

volatility3/framework/plugins/windows/direct_system_calls.py
```diff
@@ -0,0 +1,455 @@
...
+        """Creates a map of start/end addresses within a virtual address
+        descriptor tree.
+
+        Args:
+            task: The EPROCESS object of which to traverse the vad tree
+
+        Returns:
+            An iterable of tuples containing start and end addresses for each descriptor
+        """
+        vads: List[Tuple[int, int, str]] = []
```
ikelos, line 300: Feels like this code will already have been done somewhere else?  I'm happy if you want to duplicate it.  Genuinely not sure which would be the better solution (ie, how specific/responsive to future changes this would be?)

volatility3/framework/plugins/windows/indirect_system_calls.py
```diff
@@ -0,0 +1,124 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import struct
+import logging
+from typing import List, Optional
+
+import capstone
```
ikelos: Again, `ImportError` protection would be handy here, so it fails gracefully letting people know it can't operate.  I believe the yara plugins already do something similar?

volatility3/framework/plugins/windows/indirect_system_calls.py
```diff
@@ -0,0 +1,124 @@
...
+            proc_layer: the layer of the potential syscall block
+            inst: the terminating instruction of the syscall block check
+        Returns:
+            The target address of the jump if it can be computed
+        """
+
+        try:
+            jmp_address_str = proc_layer.read(inst.address, 6)
+        except exceptions.InvalidAddressException:
+            return None
```
ikelos, line 99: I assume we'd hit these so often that even adding debugging messages here to say which outcome was hit, wouldn't be worthwhile.  I just worry there's lots of ways this method can terminate, and you might want to know whether it was the very last one rather than the very first?  I know the debug log spews out a lot already, so I'm not gonna push it, just a question as to whether it would be more help than hinderance?

ikelos (commented):
Ok, generally pretty good, all minor things that turned up but generally pretty happy.  The graceful import errors needs to go in but otherwise I could let everything else slide.  I'll be interested to know whether the get_vad_maps method is a duplicate and then whether we think it's better to have a duplicate or have a dependency...

volatility3/framework/plugins/windows/direct_system_calls.py
atcuno: There are too many variations of it too really keep one central place. Some plugins want start, size - others start, end - some want only vads with mapped files, others filter to size, etc. Given all the variations, and that `get_vad_root` and `traverse` do all the heavy lifting, it seems best to let each plugin gather whatever ranges it needs.

volatility3/framework/plugins/windows/direct_system_calls.py
atcuno: Yes I much prefer it this way.

volatility3/framework/plugins/windows/direct_system_calls.py
atcuno: done. We should add to requirements.txt for parity release given how many plugins need it.

volatility3/framework/plugins/windows/indirect_system_calls.py
atcuno: done. We should add to requirements.txt for parity release given how many plugins need it.

volatility3/framework/plugins/windows/indirect_system_calls.py
atcuno: It wouldn't help. It would essentially be telling people that a `\xff\x25` byte pair we found in memory wasn't actually a jump. Others would be telling that the integer after a `\xff\x25` don't correspond to a valid offset. Across all processes you would likely have tens of 1000s of these being reported.

atcuno:
This is ready @ikelos but the test failed as the test system doesn't have capstone, and its part of the type hint. We need to:

1) get capstone in requirements.txt
2) Update the test system to have capstone installed

ikelos:
Capstone *is* already in [requirements.txt](https://github.com/volatilityfoundation/volatility3/blob/62aa9c1e12841fd2c2aa7180a489f4b8bdab7e51/requirements.txt#L12), but needs to be added to `test/requirements-testing.txt` as part of the PR please.

It also indicates that the error checking isn't working, because volatility should not throw an exception when a module can't be found.   I think that's because you've got it in the typing information, so it's accessed without a `has_capstone` guard in front of it.  I'd try putting that in quotes to delay resolving, but otherwise you'll need to weaken the type constraints or have the whole plugin bail out as soon as capstone can't be found.  It must be able to import succesfully or fail gracefully or it'll take the whole framework with it when the plugins are initially imported.  Yarascan should be an example of a plugin that bails gracefully if a necessary module can't be found for it to be of any use...

volatility3/framework/plugins/windows/indirect_system_calls.py
```diff
@@ -0,0 +1,128 @@
...
+from volatility3.framework.configuration import requirements
+from volatility3.plugins import yarascan
+from volatility3.plugins.windows import pslist, direct_system_calls
+
+vollog = logging.getLogger(__name__)
+
+# The generator of DirectSystemCalls will bail with a warning if capstone is not installed
+try:
+    import capstone
+except ImportError:
```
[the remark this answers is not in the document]
ikelos: Please also fix this with a `has_capstone` guard or similar.

volatility3/framework/plugins/windows/indirect_system_calls.py
ikelos: Thanks, it is already in `requirements.txt`, see below.

ikelos:
@atcuno, can you please add capstone to the `test/requirements-testing.txt` as part of this PR so we can get this merged please?

ikelos (approved):
There's only one small typo left?  I'll let it go and commit this tomorrow evening, but if you could get it fixed up before then it'll save someone else having to do it another time...  5;)  Otherwise all looks good, thanks for adding the requirement...

volatility3/framework/plugins/windows/indirect_system_calls.py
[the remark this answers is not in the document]
ikelos: This could also do with either a `has_capstone` or the comment moving into the except clause, just to make the CodeQL error go away please.

PR #1263, opened 2024-09-13
Linux: Add support for threads in both lsof and sockstat plugins.

gcmoreira:
This PR includes:
- lsof plugin: source code refactored.
- lsof plugin: Added the `device` column to complete the inode information. An inode number is specific to the filesystem/device it belongs to.
- lsof/sockstat plugins: Add threads support. 
  - Threads may or may not share the file descriptor table with the thread group leader, depending on whether the `CLONE_FILES` flag is included in the `clone()` syscall. 
  - Also, once started, a thread can unshare the fd table with its parent. Refer to the unshare() libc syscall wrapper man page, `unshare(2)`. 
  - Additionally, note that the `Linux lsof command` in user space includes thread listings by default as well. 
  - Now, there are two columns to identify the thread group ID  (PID) and the task/thread ID (TID).
- Added inode getters from both, the dentry and file structs. From kernels +3.9 the file struct caches the inode pointer. So, when possible, we get this value.
- Improve smear protection in these plugins and various APIs (https://github.com/volatilityfoundation/volatility3/pull/1243)

```shell
pyhon3 ./vol.py -r pretty \
    ../linux-sample-1.bin \
    linux.lsof 
Volatility 3 Framework 2.10.0
  |  PID |  TID |         Process |  FD |                       Path | Device |      Inode | Type |       Mode |                        Changed |                       Modified |                       Accessed |    Size
* |  322 |  322 |           udevd |   0 |                  /dev/null |    0:5 |       1520 |  CHR | crw-rw-rw- | 2014-06-24 10:22:36.468000 UTC | 2014-06-24 10:22:36.468000 UTC | 2014-06-24 10:22:36.468000 UTC |       0
...
* | 3968 | 3973 |           gdbus |  17 |             /tmp/vteNVRRHX |    8:1 |       7132 |  REG | -rw------- | 2014-06-24 10:41:49.976800 UTC | 2014-06-24 10:41:49.976800 UTC | 2014-06-24 10:41:46.448800 UTC |   29281
* | 3968 | 3974 |    dconf worker |   1 | /home/vol/.xsession-errors |    8:1 |     928155 |  REG | -rw------- | 2014-06-24 13:00:20.836818 UTC | 2014-06-24 13:00:20.836818 UTC | 2014-06-24 10:29:34.176798 UTC |    5249
...
```

```shell
python3 ./vol.py -r pretty \
    ../linux-sample-1.bin \
    linux.sockstat.Sockstat 
Volatility 3 Framework 2.10.0     
  | NetNS |  PID |  TID | FD |    Sock Offset |       Family |      Type |                  Proto |                             Source Addr | Source Port |                       Destination Addr | Destination Port |       State |                                         Filter
* |     - |  322 |  322 |  3 | 0x88001b61b0c0 |      AF_UNIX | SEQPACKET |                      - |                       /run/udev/control |        3693 |                                      - |                - | UNCONNECTED |                                              -
...
* |     - | 2652 | 2657 |  7 | 0x88001adf0880 |      AF_UNIX |    STREAM |                      - |                                       - |        7183 |        /var/run/dbus/system_bus_socket |             7184 | ESTABLISHED |                                              -
* |     - | 2652 | 2657 |  9 | 0x88001adf0200 |      AF_UNIX |    STREAM |                      - |                                       - |        7216 |        /var/run/dbus/system_bus_socket |             7217 | ESTABLISHED |                                              -

### DIFF
--- a/volatility3/framework/plugins/linux/lsof.py
+++ b/volatility3/framework/plugins/linux/lsof.py
@@ -3,7 +3,7 @@
 #
 import logging
 import datetime
-from dataclasses import dataclass, astuple, field
+import dataclasses
 from typing import List, Callable, Tuple
 
 from volatility3.framework import renderers, interfaces, constants
@@ -17,7 +17,7 @@ from volatility3.plugins import timeliner
 vollog = logging.getLogger(__name__)
 
 
-@dataclass
+@dataclasses.dataclass
 class FDUser:
     """FD user representation, featuring augmented information and formatted fields.
     This is the data the plugin will eventually display.
@@ -28,17 +28,23 @@ class FDUser:
     task_comm: str
     fd_num: int
     full_path: str
-    device: str = field(default=renderers.NotAvailableValue())
-    inode_num: int = field(default=renderers.NotAvailableValue())
-    inode_type: str = field(default=renderers.NotAvailableValue())
-    file_mode: str = field(default=renderers.NotAvailableValue())
-    change_time: datetime.datetime = field(default=renderers.NotAvailableValue())
-    modification_time: datetime.datetime = field(default=renderers.NotAvailableValue())
-    access_time: datetime.datetime = field(default=renderers.NotAvailableValue())
-    inode_size: int = field(default=renderers.NotAvailableValue())
+    device: str = dataclasses.field(default=renderers.NotAvailableValue())
+    inode_num: int = dataclasses.field(default=renderers.NotAvailableValue())
+    inode_type: str = dataclasses.field(default=renderers.NotAvailableValue())
+    file_mode: str = dataclasses.field(default=renderers.NotAvailableValue())
+    change_time: datetime.datetime = dataclasses.field(
+        default=renderers.NotAvailableValue()
+    )
+    modification_time: datetime.datetime = dataclasses.field(
+        default=renderers.NotAvailableValue()
+    )
+    access_time: datetime.datetime = dataclasses.field(
+        default=renderers.NotAvailableValue()
+    )
+    inode_size: int = dataclasses.field(default=renderers.NotAvailableValue())
 
 
-@dataclass
+@dataclasses.dataclass
 class FDInternal:
     """FD internal representation containing only the core objects
 
@@ -168,7 +174,7 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
             self.context, vmlinux_module_name, filter_func=filter_func
         ):
             fd_user = fd_internal.to_user()
-            yield (0, astuple(fd_user))
+            yield (0, dataclasses.astuple(fd_user))
 
     def run(self):
         pids = self.config.get("pid", None)
--- a/volatility3/framework/plugins/linux/lsof.py
+++ b/volatility3/framework/plugins/linux/lsof.py
@@ -1,12 +1,12 @@
 # This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
-"""A module containing a collection of plugins that produce data typically
-found in Linux's /proc file system."""
-import logging, datetime
+import logging
+from datetime import datetime
+from dataclasses import dataclass, astuple, field
 from typing import List, Callable
 
-from volatility3.framework import renderers, interfaces, constants, exceptions
+from volatility3.framework import renderers, interfaces, constants
 from volatility3.framework.configuration import requirements
 from volatility3.framework.interfaces import plugins
 from volatility3.framework.objects import utility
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: volatility3/framework/plugins/linux/lsof.py [hunk 1]
```diff
@@ -1,12 +1,12 @@
 # This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
-"""A module containing a collection of plugins that produce data typically
-found in Linux's /proc file system."""
-import logging, datetime
-from typing import List, Callable
+import logging
+from datetime import datetime
+from dataclasses import dataclass, astuple, field
```
ikelos: I know these are built-in, but please can we stick to the style guide we use, which says all imports should be modules rather than methods, classes or objects please (`typing` gets a free pass)?  This prevents people importing `lsof.astuple` and thinking that's where it's defined.  See https://google.github.io/styleguide/pyguide.html#22-import

This also applies to datetime which was previously correctly imported.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -843,25 +843,55 @@ def get_subdirs(self) -> interfaces.objects.ObjectInterface:
...
+            return self.f_dentry
         else:
             raise AttributeError("Unable to find file -> dentry")
 
     def get_vfsmnt(self) -> interfaces.objects.ObjectInterface:
         """Returns the fs (vfsmount) where this file is mounted"""
-        if self.has_member("f_vfsmnt"):
-            return self.f_vfsmnt
-        elif self.has_member("f_path"):
+        if self.has_member("f_path"):
```
ikelos, line 868: Does the change in ordering of these tests affect the result of them?  Could there ever be structures with both `f_path` and `f_vfsmnt` defined and which would we want to prefer?

ikelos (commented):
Generally looks good, just needs the merge conflicts (around process name and the pid/tid return) fixing up please...

volatility3/framework/plugins/linux/lsof.py
gcmoreira: hm sorry but that's the Google style guide.. is it mentioned anywhere that we stick to Google's? why?

We should stick to the Python's styles guide which is the PEP8 ... [PEP 8 – Style Guide for Python Code](https://peps.python.org/pep-0008/#imports)  , for which that's perfectly fine.. actually there is an example that shows that's correct:
```
# Correct:
from subprocess import Popen, PIPE
```

Also for classes:
```
When importing a class from a class-containing module, it’s usually okay to spell this:

from myclass import MyClass
from foo.bar.yourclass import YourClass

If this spelling causes local name clashes, then spell them explicitly:

import myclass
import foo.bar.yourclass

and use myclass.MyClass and foo.bar.yourclass.YourClass.
```

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: I moved the **if** matching the most recent kernels to be first, which it's likely to be the most often case now and in the future. Cannot be both members.

gcmoreira:
Ok, after a very long day fighting with my friend GIT and the changes in #1271 + related commits.. I think I managed to resolve all the conflicts

volatility3/framework/plugins/linux/lsof.py
ikelos: It's actually from the [Volatility style guide](https://github.com/volatilityfoundation/volatility/wiki/Style-Guide) from 2014 which is mostly in alignment with PEP8 with some extra requests.  I thought that was a bit terse and difficult to read, so I pointed you at Google's style guide which is longer with more discussion, but here's the original text:

* Importing specific objects or functions is discouraged.
** It pollutes the namespace and causes confusion. Previous versions of volatility allowed people to import functions from files they weren't defined in.
* "from blah import *" is strongly discouraged.
** Again, namespace pollution and inappropriate imports.

The entire rest of the code base is formatted that way and it's better to have a style guide than not.  I honestly didn't expect this request to be a big sticking point?  Can you explain why allowing namespace pollution is a better stylistic option than our current one, or is your disagreement just that the documentation was provided by Google and that code comes out slightly longer?

The style guide needs updating (it still includes formatting that's handled by black now, rather than yapf at the time) and it could do with reformatting since some justifications are at the same level as the bullet points they're justifying.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Ok, cool, thanks  5:)

volatility3/framework/plugins/linux/lsof.py
gcmoreira: No worries. I will take care of this.

I'm not saying it's better, but it's unclear to me why this is a concern now. I've never heard it mentioned in any of my previous contributions or by others. 

We'll also need to fix this, if we want to make an exception with Typing, in about 50 other imports. Otherwise, there are 244 imports to address.

Also, Volatility2 is a separate project, so we can't expect people to refer to it for guidance on writing code for Volatility3. Volatility3 should have its own dedicated style guide page, or clear instructions of which style guide we use.

I wonder if there's a way to include that as a rule in Black. Any idea? By the way, Black follows Python's PEP8 style guide.

I also learned we have a [.style.yapf ](https://github.com/volatilityfoundation/volatility3/blob/develop/.style.yapf) file, which configures the [Yapf Python formatted from Google](https://github.com/google/yapf). 

This **.style.yapf** file doesn't specify any `based_on_style`, which means it's defaulting to **PEP8** as the style guide. However, it seems we can switch it to Google's style guide if preferred. See [formatting-style](https://github.com/google/yapf?tab=readme-ov-file#formatting-style)

```yaml
based_on_style = google
```

Maybe, if we are going to use Google's style guide we should use **Yapf** instead of **Black**.
Otherwise, it might be a good idea to remove that file, since it creates confusion about which formatter the project actually uses.

gcmoreira:
@ikelos all done

volatility3/framework/plugins/linux/lsof.py
ikelos: You can go back through previous pull requests.  Where I've spotted it I should have always asked for it to be corrected (and provided the rationale).  Most people stick with it, I'm disappointed I missed it for 244 imports!  5:(  It's possible I had quibbled a lot about the PRs and was feeling sorry for them having to fix so many things, but I feel it's saved us for lots of cross-imports between plugins and the core so I think it's a worthwhile guideline to have.

Yep, fair point, the style guide never got pulled across to vol 3 I don't think and I didn't want to sit down and rewrite it again (I didn't even tidy up the one we had), but we should.

I don't think black has that as a flag (I don't think black has many flags, to be fair).  If you find one, it would be great to apply it to our automatic black check, so that it's applied to all PRs evenly when people submit them!

Yeah, that's the style mechanism we used very early on, it was deprecated years ago but I guess I didn't remove the file.  I've taken care of that now.  Yapf was very bad, it would reweight something, and then because of the reweighting running it again would change it back, meaning it never reached a stable equilibrium.  It's the reason we moved to black.

We don't use the Google style guide exclusively, but there are a couple of points in it that make sense, such as only importing modules (in fact, that's the main one I can think of), which we've carried over.  As mentioned the main style guide is in general PEP 8.

ikelos (approved):
Looks good, thanks for making the changes!

Issue #1256, opened 2024-09-10
Two vad related issues need warning changes/additions


atcuno:
1) traverse should print a debug message if overlapping regions are founded: https://github.com/volatilityfoundation/volatility3/blob/d56cd83510e64f9f875ff1dad11d8c8cbf5995f5/volatility3/framework/symbols/windows/extensions/__init__.py#L76

2) This messages needs to be toned down to debug instead of warning:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/vadyarascan.py#L71

This warning makes the vad yarascan plugin unusable on many samples (including ones we had for Black Hat and DFRWS) as it prints 1000s of times across processes.

ikelos:
If you can get a PR in place for what you're after we can get those tweaked.

eve-mem:
@atcuno for the overlapping regions - can I check this is what you're looking for. 

In this faked output there is an overlap with the starts and ends of these two entries. For the `0xff94a1d0` it overlaps with `0x8220e848` because the end of the first one is `0x332fff`, which is larger than the start of the `0x320000` - an overlap of `0x12fff`. If so, do you think it's enough to check if children overlap with there parents - or is it necessary to traverse everything and then check for overlaps? 

```python
$ python3 vol.py -r pretty -f  win-xp-laptop-2005-06-25.img  windows.vadwalk --pid 4012
Volatility 3 Framework 2.10.0
Formatting...0.00               PDB scanning finished                        
  |  PID | Process |     Offset |     Parent |       Left |      Right |      Start |        End |  Tag
<snip>
* | 4012 |  dd.exe | 0xff94a1d0 | 0x81f35458 | 0x8220e848 | 0xffaa0a90 |   0x320000 |   0x332fff | Vad 
* | 4012 |  dd.exe | 0x8220e848 | 0xff94a1d0 |        0x0 |        0x0 |   0x320000 |   0x32ffff | VadS
<snip>
```


For point two I've made that tiny change here: https://github.com/volatilityfoundation/volatility3/pull/1270

atcuno:
@eve-mem sorry for the late reply. Nice work on warning->debug, it has made the plugin usable in trainings.

I think just checking children is okay given the complexity otherwise.

PR #1270, opened 2024-09-25
Windows: change warnings around large memory maps to debug level as per issue #1256

eve-mem:
Hello :wave: 

In issue https://github.com/volatilityfoundation/volatility3/issues/1256 @atcuno wanted to see the warnings in vadyarascan lowered to a debug message. This PR makes just that minor change. 

Note that issue https://github.com/volatilityfoundation/volatility3/issues/1256 also talks about overlapping vads. I'm worried that if this is merged it'll auto close the issue - which shouldn't happen as the other part of the issue is not yet resolved. 

Thanks as always
:fox_face:


ikelos (approved):
Looks good, thanks!

PR #1271, opened 2024-09-25
Linux: Update sockstat to render process names

dgmcdona:
Currently, process names are not displayed for sockets in the sockstat plugin, making analysis more painful than it needs to be. This updates the `list_sockets` classmethod and the `_generator` method to return the process name in addition to the PID.

Because this is changing the public interface, this commit includes a major version bump for `linux.sockstat.Sockstat`.

### DIFF
--- a/volatility3/framework/plugins/linux/sockstat.py
+++ b/volatility3/framework/plugins/linux/sockstat.py
@@ -22,7 +22,7 @@ class SockHandlers(interfaces.configuration.VersionableInterface):
 
     _required_framework_version = (2, 0, 0)
 
-    _version = (1, 0, 0)
+    _version = (2, 0, 0)
 
     def __init__(self, vmlinux, task):
         self._vmlinux = vmlinux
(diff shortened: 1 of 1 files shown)


atcuno:
@ikelos this is a high priority change for the October release. Also, please make sure the version bump is correct here.

gcmoreira:
@ikelos This will have conflicts with https://github.com/volatilityfoundation/volatility3/pull/1263 .. depending on the order in which you want to merge them, we'll need to adjust the other accordingly.

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -548,7 +548,7 @@ def list_sockets(
             except AttributeError:
                 netns_id = NotAvailableValue()
 
-            yield task, netns_id, fd_num, family, sock_type, protocol, sock_fields
```
ikelos, line 551: `list_sockets` is an expose plugin, so this API is changing in a non-backwards compatible way, meaning the plugin needs a MAJOR version bump, and please verify that any consumers of this plugin (within the core codebase) have their dependencies bumped and are tested to work with the new return value.  At the moment it doesn't look like any other plugins depend on it, so it should be fine to just bump it to `_version = (2, 0, 0)`...

ikelos (changes requested):
Pretty straightforward change, but altering the return value requires a MAJOR version bump (because it's not a simple additive change to the API).

ikelos:
#1263 seems much bigger, might be easier to just get this one merged and then fix up the other one if this one is urgent?  I'll try and get to #1263 next, but I'm falling asleep here I'm afraid...  5:S

dgmcdona:
I thought I had already bumped the version but apparently not, it's done now. There are no consumers of this plugin that require updating.

ikelos (approved):
Looks good now, thanks

PR #1272, opened 2024-09-27
#816  cmdscan and console

### LINKED ISSUE #816: Port "cmdscan" and "console" plugins from Vol2 to Vol3, please
I want to inspect which commands are executed in a terminal and what is its output but with Vol3 there's no "cmdscan" and "console" plugin equivalent, so I can't get this information.

A "cmdscan" and "console" plugin port from Vol2 to Vol3

Thanks a lot!

P.S: "Screenshot" plugin would be nice to have, too.

superponible:
This is the initial support for cmdscan and consoles. It adds support for modern versions of Windows (17763, released in 2018, and above).


volatility3/framework/plugins/windows/cmdscan.py
```diff
@@ -0,0 +1,380 @@
...
+        Returns:
+            A list of tuples of:
+            vad_base: the base address
+            vad_size: the size of the VAD
+        """
+        vads = []
+        for vad in conhost_proc.get_vad_root().traverse():
+            base = vad.get_start()
+            if vad.get_size() < size_filter:
+                vads.append((base, vad.get_size()))
```
ikelos: Feels like this could be a generator, but no issue with it just returning a list instead...

volatility3/framework/plugins/windows/cmdscan.py
```diff
@@ -0,0 +1,380 @@
...
+
+        return vads
+
+    @classmethod
+    def get_command_history(
+        cls,
+        context: interfaces.context.ContextInterface,
+        kernel_layer_name: str,
+        kernel_table_name: str,
```
ikelos: This should probably have the word symbol in it somewhere too?  Possibly instead of kernel, or as well as?

volatility3/framework/plugins/windows/cmdscan.py
```diff
@@ -0,0 +1,380 @@
...
+                    scanners.BytesScanner(max_history_bytes),
+                    sections=sections,
+                ):
+                    command_history_properties = []
+
+                    try:
+                        command_history = conhost_module.object(
+                            "_COMMAND_HISTORY",
+                            offset=address
+                            - conhost_module.get_type(
```
ikelos: I'd probably calculate this value once outside of the loop and then just keep referencing it?

volatility3/framework/plugins/windows/cmdscan.py
```diff
@@ -0,0 +1,380 @@
...
+
+                        vollog.debug(
+                            f"Getting Command History properties for {command_history}"
+                        )
+                        command_history_properties.append(
+                            {
+                                "level": 0,
+                                "name": "_COMMAND_HISTORY",
+                                "address": command_history.vol.offset,
+                                "data": "",
```
ikelos: `""` rather than `None`?  I haven't read how this is going to be consumed, but empty strings  are sometimes less confusing as `None`...

volatility3/framework/plugins/windows/cmdscan.py
```diff
@@ -0,0 +1,380 @@
...
+                            }
+                        )
+                        command_history_properties.append(
+                            {
+                                "level": 1,
+                                "name": f"_COMMAND_HISTORY.Application",
```
ikelos: Did this need to be an f-string or were the `{` brackets missed?

volatility3/framework/plugins/windows/cmdscan.py
```diff
@@ -0,0 +1,380 @@
...
+                            except Exception as e:
+                                vollog.debug(
+                                    f"reading {bucket_cmd} encountered exception {e}"
+                                )
+                    except Exception as e:
+                        vollog.debug(
+                            f"reading {command_history} encountered exception {e}"
+                        )
+
+                    if command_history and command_history_properties:
```
ikelos, line 261: `command_history` needs to be initialized to `None` outside of the try/except, otherwise this could fire a `NameError` Not Defined exception itself...

volatility3/framework/plugins/windows/cmdscan.py
```diff
@@ -0,0 +1,380 @@
...
+                        found_history_for_proc = True
+                        yield conhost_proc, command_history, command_history_properties
+
+            if not found_history_for_proc:
```
ikelos, line 268: This seems like it could repeat entries that have already been printed?  Could you please add comment explaining why this is here?

volatility3/framework/plugins/windows/cmdscan.py
```diff
@@ -0,0 +1,380 @@
...
+            requirements.PluginRequirement(
+                name="consoles", plugin=consoles.Consoles, version=(1, 0, 0)
+            ),
+            requirements.BooleanRequirement(
+                name="no_registry",
+                description="Don't search the registry for possible values of CommandHistorySize",
+                optional=True,
+                default=False,
+            ),
+            requirements.ListRequirement(
```
ikelos, line 50: Is a specific list more useful that just a less-than/greater-than divide value?

ikelos: Also, why is the default `[50]`?  Will that return responses in most cases?  Should we not have a default value that returns *all* `CommandHistory` sizes?

volatility3/framework/symbols/windows/extensions/consoles.py
```diff
@@ -0,0 +1,439 @@
...
+            0x98,
+            0xD8,
+            0xE0,
+            0xE8,
+            0xF8,
+            0xF0,
+            0xA0,
+        )
+        if n == 0 and not valid:
+            print("Bad Dbcs Attribute {}".format(hex(c)))
```
ikelos: Please do all output using `vollog` at a specified level, `print`s shouldn't be going in the codebase.

volatility3/framework/symbols/windows/extensions/consoles.py
```diff
@@ -0,0 +1,439 @@
...
+        return valid
+
+    def get_text(self, truncate=True):
```
ikelos: Presumably this returns a `str`?  Please could we typing information in the signature?

volatility3/framework/symbols/windows/extensions/consoles.py
```diff
@@ -0,0 +1,439 @@
...
+                line = "".join(
+                    (
+                        char_row[i : i + 2].decode("utf-16le", errors="replace")
+                        if self._valid_dbcs(char_row[i + 2], char_row[i + 1])
+                        else ""
+                    )
+                    for i in range(0, len(char_row), 3)
+                )
+        except Exception as e:
+            print(e)
```
ikelos: Again, `vollog`, not `print` please.

volatility3/framework/symbols/windows/extensions/consoles.py
```diff
@@ -0,0 +1,439 @@
...
+        if isinstance(exe_name, objects.Pointer):
+            exe_name = exe_name.dereference()
+            return exe_name.get_string()
+
+        if self.ExeName.Length < 8:
+            return self.ExeName.Chars.cast(
+                "string",
+                encoding="utf-16",
+                errors="replace",
+                max_length=self.ExeName.Length * 2,
```
ikelos: Here `Length` is the length of the string, but below it's the length in bytes (because you check `< 1024` and then max it out at 512).  So... which is it?

volatility3/framework/symbols/windows/extensions/consoles.py
```diff
@@ -0,0 +1,439 @@
...
+            )
+        elif self.ExeName.Length < 1024:
+            return self.ExeName.Pointer.dereference().cast(
+                "string", encoding="utf-16", errors="replace", max_length=512
```
ikelos: If you know the length, why wouldn't you use that as the max_length?

volatility3/framework/symbols/windows/extensions/consoles.py
```diff
@@ -0,0 +1,439 @@
...
+
+        @param truncate: True if the empty rows at the
+        end (i.e. bottom) of the screen buffer should be
+        supressed.
+        """
+        rows = []
+
+        capacity = self.TextBufferInfo.BufferCapacity
+        start = self.TextBufferInfo.BufferStart
+        buffer_rows = self.TextBufferInfo.BufferRows.dereference()
```
ikelos: It's not clear why so many pointers are being explicitly dereferenced?  It should be that Self.TextBufferInfo.BufferRows.Rows will do the dereferencing for you.  I'd rather not litter the code with unnecessary `dereference`s in case people come to use it as a teaching guide...

volatility3/framework/symbols/windows/extensions/consoles.py
```diff
@@ -0,0 +1,439 @@
...
+                text = row.get_text(truncate_lines)
+                rows.append(text)
+            except:
+                break
+
+        if truncate_rows:
+            rows = self._truncate_rows(rows)
+
+        if rows:
+            rows = ["=== START OF BUFFER ==="] + rows + ["=== END OF BUFFER ==="]
```
ikelos: I understand why you might want to clearly delimit this for command line output, but it's surely going to be inconvenient for programatic use of this plugin (or if the data were the put into a web interface where the start and end can easily be determined)?

volatility3/framework/symbols/windows/extensions/consoles.py
```diff
@@ -0,0 +1,439 @@
...
+        else:
+            rows = ["=== NO BUFFER DATA FOUND  ==="]
```
ikelos: In this case I'd expect rows to be empty and then the generator to yield back a `rows or BaseAbsentValue` or similar?

volatility3/framework/symbols/windows/extensions/consoles.py
```diff
@@ -0,0 +1,439 @@
...
+        # Last displayed must be between -1 and max
+        if self.HistoryBufferCount < 1 or self.HistoryBufferCount > max_buffers:
+            return False
+
+        if not self.get_title() and not self.get_original_title():
+            return False
+
+        return True
+
+    def get_screens(self):
```
ikelos: Typing information for the return value please.

volatility3/framework/symbols/windows/extensions/consoles.py
```diff
@@ -0,0 +1,439 @@
...
+            while cur and cur.vol.offset != 0 and cur.vol.offset not in seen:
+                cur.TextBufferInfo.BufferRows.Rows.count = (
+                    cur.TextBufferInfo.BufferCapacity
+                )
+                yield cur
+                seen.add(cur.vol.offset)
+                cur = cur.Next
+
+    def get_histories(self):
+        for cmd_hist in self.HistoryList.dereference().to_list(
```
ikelos: Again, this should dereference properly unless the Pointer class has a `to_list` method of its own.

volatility3/framework/symbols/windows/extensions/consoles.py
```diff
@@ -0,0 +1,439 @@
...
+            self.Length < 1
+            or self.Allocated < 1
+            or self.Length > 1024
+            or self.Allocated > 1024
+        ):
+            return False
+
+        return True
+
+    def get_command(self):
```
ikelos: This feels like the same method as `ALIAS.get_source()`?  Is there any way of making sure the two stay in sync?

volatility3/framework/symbols/windows/extensions/consoles.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+class COMMAND_HISTORY(objects.StructType):
+    """A Command History Structure."""
+
+    @property
+    def CommandCount(self):
+        command_type = self.get_symbol_table_name() + constants.BANG + "_COMMAND"
+        command_size = self._context.symbol_space.get_type(command_type).size
+        return int((self.CommandBucket.End - self.CommandBucket.Begin) / command_size)
```
ikelos, line 333: If `command_size` ever isn't a multiple of the total size, this could end up rounding down.  If we're sure this'll always divide properly to provide an integer value, that's fine but if not it'll need tweaking.

ikelos (commented):
Generally looks good, thanks!  Major points were unnecessary dereferencing, missing typing and mixed return values (found by the code scanner).  In most of those cases it's just a matter or `return None` at the end, but it's not keen on falling off the bottom of code apparently...  5:)

volatility3/framework/plugins/windows/cmdscan.py
superponible: It's attempting to construct the object at each address returned by the scan. Not all of them are valid and end up going to the next loop iteration.

volatility3/framework/plugins/windows/cmdscan.py
superponible: nope, over aggressive copy/pasting.

volatility3/framework/plugins/windows/cmdscan.py
superponible: making it None

volatility3/framework/plugins/windows/cmdscan.py
superponible: good catch

volatility3/framework/plugins/windows/cmdscan.py
superponible: I was trying to make it flexible. In probably 99% of the cases it will always be 50. I've added the ability, which wasn't in vol2, for the code to walk the Registry for other values. That value is stored in a per user hive in the Registry, so it could theoretically be anything, though in practice it's unlikely people change it.  On the off chance it was changed, and the registry scan found a value of, say, 100, then having to to 50 < x < 100 or something would cause the scanning for find a whole range of potential values, rather than the 2 we want to look for.

volatility3/framework/symbols/windows/extensions/consoles.py
superponible: done

volatility3/framework/symbols/windows/extensions/consoles.py
superponible: done

volatility3/framework/plugins/windows/cmdscan.py
superponible: done

volatility3/framework/plugins/windows/cmdscan.py
superponible: done

volatility3/framework/plugins/windows/cmdscan.py
superponible: shouldn't repeat. added comment.

volatility3/framework/symbols/windows/extensions/consoles.py
superponible: done

volatility3/framework/plugins/windows/cmdscan.py
ikelos: Sure, but `conhost_module` is defined before the loop begins, `get_type("_COMMAND_HISTORY").relative_child_offset("CommandCountMax")` should therefore always be the same value every time round the loop, so... wouldn't it make sense to just figure it out once at the start?

ikelos:
Thanks for fixing a bunch of those, there's still some outstanding comments on it though.  Sometimes github can hide them when there's a lot, but it would be good to cover them all off please?

superponible:
> Thanks for fixing a bunch of those, there's still some outstanding comments on it though. Sometimes github can hide them when there's a lot, but it would be good to cover them all off please?

hey @ikelos sorry for the confusion. i've been fixing things locally but haven't pushed any changes up yet, so that's why you haven't seen them.  hope to wrap up tomorrow, at least the first pass and let you know any outstanding questions I may have.

volatility3/framework/plugins/windows/cmdscan.py
superponible: oh, you meant get the offset outside the loop, yeah, that makes sense. I can do that. i thought you were saying create command_history outside the loop. but it makes more sense when i look at the line number this comment was on.

volatility3/framework/symbols/windows/extensions/consoles.py
superponible: done

volatility3/framework/symbols/windows/extensions/consoles.py
superponible: it should be a multiple.

volatility3/framework/symbols/windows/extensions/consoles.py
superponible: Honestly don't remember where the 1024 came from. It might just be to prevent something very large that's totally incorrect. I got rid of the 512 and used self.Length in both places now.

volatility3/framework/symbols/windows/extensions/consoles.py
superponible: used self.Length*2 like in the other place

volatility3/framework/symbols/windows/extensions/consoles.py
superponible: worked without it, so removed

volatility3/framework/symbols/windows/extensions/consoles.py
superponible: also removed

volatility3/framework/symbols/windows/extensions/consoles.py
superponible: combined this in one place

superponible:
@ikelos I think everything has been addressed and committed.

ikelos (approved):
Awesome, thanks very much for getting all those changes made!  5:)

PR #1273, opened 2024-09-27
Windows: psxview win10 bugfix

dgmcdona:
The current implementation of this plugin does not incorporate the cookie on Win10+ systems, causing it to fail. This PR fixes that issue, and also contributes a performance improvement by extracting the call to `handles_plugin.get_type_map()` from the loop. Also replaces another for-loop with a list comprehension for clarity.

I took the opportunity to also add type hints to this module, update the type hints in `pslist`, and do a couple of very minor refactorings that I think add clarity to the code. Those changes in their own commits in case we want to revert any of them.

The bugfix component of this has been tested by @atcuno , and I have done extensive tests as well to confirm no breaking changes across a large sample set.

### DIFF
--- a/volatility3/framework/plugins/windows/pslist.py
+++ b/volatility3/framework/plugins/windows/pslist.py
@@ -4,7 +4,7 @@
 
 import datetime
 import logging
-from typing import Callable, Iterable, List, Type
+from typing import Callable, Iterator, List, Type, TYPE_CHECKING
 
 from volatility3.framework import renderers, interfaces, layers, exceptions, constants
 from volatility3.framework.configuration import requirements
@@ -12,6 +12,7 @@ from volatility3.framework.objects import utility
 from volatility3.framework.renderers import format_hints
 from volatility3.framework.symbols import intermed
 from volatility3.framework.symbols.windows.extensions import pe
+from volatility3.framework.symbols.windows import extensions
 from volatility3.plugins import timeliner
 
 vollog = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/windows/pslist.py
```diff
@@ -197,7 +198,7 @@ def list_processes(
         filter_func: Callable[
             [interfaces.objects.ObjectInterface], bool
         ] = lambda _: False,
-    ) -> Iterable[interfaces.objects.ObjectInterface]:
+    ) -> Iterator["extensions.EPROCESS"]:
```
ikelos, line 201: Is there a reason to put this in quotes?  It would've been imported already by now, wouldn't it?

volatility3/framework/plugins/windows/pslist.py
```diff
@@ -4,14 +4,15 @@
 
 import datetime
 import logging
-from typing import Callable, Iterable, List, Type
+from typing import Callable, Iterator, List, Type, TYPE_CHECKING
```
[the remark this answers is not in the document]
ikelos: Please fix up any code scanning identified issues...

ikelos (commented):
Yep, all looks reasonable, and nice optimization.  The original PR had a lot new going on which is possibly how it slipped through first time.  Just needs the automatic scanning issues resolved before merging...

volatility3/framework/plugins/windows/pslist.py
dgmcdona: I thought it would have, but ran into this after making the initial change: `AttributeError: partially initialized module 'volatility3.framework.symbols.windows.extensions' has no attribute 'EPROCESS' (most likely due to a circular import)`

ikelos: Ok, quotes it is then...  5:)

PR #1276, opened 2024-09-30
Windows: Callbacks - fixes bad callback validity check

dgmcdona:
This fixes a bug in the callbacks plugin which causes it to miss `IoRegisterShutdownNotification` callbacks on x86 samples. The `header.NameInfo.Name` field was being incorrectly treated as the device type. This fixes the issue by updating the `is_valid` method on the `_SHUTDOWN_PACKET` extension class to take a `type_map` parameter, and updates the method to correctly validate the object type.

LMK if we need to do a version bump on the callbacks plugin, and if so, what kind. I was unsure because the change to `is_valid`'s method signature is in the callbacks extensions module and not the plugin itself.

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/callbacks.py
+++ b/volatility3/framework/symbols/windows/extensions/callbacks.py
@@ -15,7 +15,7 @@ class _SHUTDOWN_PACKET(objects.StructType, pool.ExecutiveObject):
     It exposes a function which sanity-checks structure members.
     """
 
-    def is_parseable(self, type_map: Dict[int, str]) -> bool:
+    def is_valid(self) -> bool:
         """
         Perform some checks.
         """
@@ -30,6 +30,25 @@ class _SHUTDOWN_PACKET(objects.StructType, pool.ExecutiveObject):
                 )
                 return False
 
+        except exceptions.InvalidAddressException:
+            vollog.debug(
+                f"callback obj 0x{self.vol.offset:x} invalid due to invalid address access"
+            )
+            return False
+
+        return True
+
+    def is_parseable(self, type_map: Dict[int, str]) -> bool:
+        """
+        Determines whether or not this `_SHUTDOWN_PACKET` callback can be reliably parsed.
+        Requires a `type_map` that maps NT executive object type indices to string representations.
+        This type map can be acquired via the `handles.Handles.get_type_map` classmethod.
+        """
+        if not self.is_valid():
+            return False
+
+        try:
+
             device = self.DeviceObject
             if not device or not (device.DriverObject.DriverStart % 0x1000 == 0):
                 vollog.debug(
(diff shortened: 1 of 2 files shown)


atcuno:
@ikelos this is another one for the parity push

volatility3/framework/symbols/windows/extensions/callbacks.py
```diff
@@ -14,7 +15,7 @@ class _SHUTDOWN_PACKET(objects.StructType, pool.ExecutiveObject):
     It exposes a function which sanity-checks structure members.
     """
 
-    def is_valid(self) -> bool:
+    def is_valid(self, type_map: Dict[int, str]) -> bool:
```
ikelos: This is a change in API and will require bumps somewhere.  Please could you consider a different name, or determining the type_map within the call please?

ikelos (commented):
Thanks, happy to get this fixed, but please don't go changing `is_valid`.  Changing it from every other `is_valid` method, and potentially for any other users of it is pretty messy and would require lots of version bumps.  Please implement a different method name, or determine the type_map from inside the call.

volatility3/framework/symbols/windows/extensions/callbacks.py
dgmcdona: Changed to `is_parseable(type_map)`

volatility3/framework/symbols/windows/extensions/callbacks.py
```diff
@@ -14,7 +15,7 @@ class _SHUTDOWN_PACKET(objects.StructType, pool.ExecutiveObject):
     It exposes a function which sanity-checks structure members.
     """
 
-    def is_valid(self) -> bool:
```
ikelos: Did you get rid of `is_valid`?  That's also in the API, so would need a version bump to remove them?  I'd recommend `is_parsable` calls `is_valid` to replicate whatever it already did, and then just adds in the check based on the result at the end...

ikelos (commented):
Still needing a version bump in this iteration, I'd leave `is_valid` as it is (and make use of it in the new `is_parsable` function)...

volatility3/framework/symbols/windows/extensions/callbacks.py
dgmcdona: Done - I was torn on whether or not to duplicate the device object check in both methods and opted not to in order to avoid doing the pointer dereference twice, even if it makes the `is_valid` check somewhat less effective.

volatility3/framework/symbols/windows/extensions/callbacks.py
```diff
@@ -24,27 +25,54 @@ def is_valid(self) -> bool:
...
+                )
+            return is_valid
         except exceptions.InvalidAddressException:
             vollog.debug(
                 f"callback obj 0x{self.vol.offset:x} invalid due to invalid address access"
             )
             return False
-
-        try:
-            header = device.get_object_header()
```
ikelos, line 43: Looks like this test got remove from `is_valid`, which *could* return different results, but I'm kinda willing to overlook it since I imagine it would only affect a very small number of results.

ikelos (commented):
Thanks for making the requested changes.  I think it does still potentially alter the results that are returned, but I think it's likely a very small number of affected results.

PR #1277, opened 2024-09-30
Windows: Netscan - fix missing TCP connections

dgmcdona:
This fixes some missing TCP connections in Windows 10 20348 by adding a scan constraint for `TTcb` pool tags when using the `netscan-win10-20348` layer.


atcuno:
@ikelos this is another critical bug fix

ikelos (approved):
Looks ok, thanks  5:)

PR #1278, opened 2024-09-30
Linux: Sockstat - fix incorrect version bump

dgmcdona:
I erred when I bumped the version as a part of #1277 - The `SockHandlers` class got the major version bump, not `Sockstat`. This reverts the `SockHandlers` version bump and applies it to `Sockstat` instead.

### DIFF
--- a/volatility3/framework/plugins/linux/sockstat.py
+++ b/volatility3/framework/plugins/linux/sockstat.py
@@ -438,7 +438,7 @@ class Sockstat(plugins.PluginInterface):
 
     _required_framework_version = (2, 0, 0)
 
-    _version = (1, 0, 0)
+    _version = (2, 0, 0)
 
     @classmethod
     def get_requirements(cls):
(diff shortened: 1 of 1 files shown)


ikelos:
It's generally bad practice to decrease version numbers, even if they were bumped by mistake.  Does anything rely on either of these (ie, is the version number used at all anywhere)?

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -22,7 +22,7 @@ class SockHandlers(interfaces.configuration.VersionableInterface):
 
     _required_framework_version = (2, 0, 0)
 
-    _version = (2, 0, 0)
+    _version = (1, 0, 0)
```
ikelos: It's not a good idea to lower a version number just to correct a mistake.  I'd live with the increased version number and just move on from there...

dgmcdona: Reverted back to 2.0.0 and updated version requirement in `Sockscan` plugin class to match.

dgmcdona:
> It's generally bad practice to decrease version numbers, even if they were bumped by mistake. Does anything rely on either of these (ie, is the version number used at all anywhere)?

There are no dependencies outside of this module.

ikelos (approved):

PR #1279, opened 2024-09-30
Linux: Sockstat - adds pytest

dgmcdona:
This adds a pytest for the `linux.sockstat.Sockstat` plugin.

### DIFF
--- a/test/test_volatility.py
+++ b/test/test_volatility.py
@@ -341,6 +341,17 @@ def test_linux_tty_check(image, volatility, python):
     assert out.count(b"\n") >= 5
     assert rc == 0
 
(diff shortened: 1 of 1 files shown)


dgmcdona:
This is going to fail until  #1278 is merged - should I leave it as its own PR or add this to the other as an additional commit?

ikelos:
Seems to break, but otherwise I'm in favour of more tests...

dgmcdona:
> Seems to break, but otherwise I'm in favour of more tests...

It's fixed now that #1278 is merged

ikelos (approved):
Looks good, thanks!

PR #1283, opened 2024-10-01
Add linux hidden_modules plugin

gcmoreira:
This PR introduces a new kernel module scanning technique that is significantly faster and more efficient than the traditional Volatility2 plugin. It uses less memory and I/O, while also catching advanced threats missed by the earlier method.

> From kernels 4.2 ([93c2e10](https://github.com/torvalds/linux/commit/93c2e105f6bcee231c951ba0e56e84505c4b0483)) struct module allocation are aligned to the L1 cache line size. For i386, amd64, and arm64 architectures, this is usually 64 bytes (or 128 bytes, which still works). However, in x86 this can be changed in the Linux kernel configuration via [CONFIG_X86_L1_CACHE_SHIFT](https://elixir.bootlin.com/linux/v6.11.1/source/arch/x86/include/asm/cache.h#L8). For ARM64 this valued is fixed to [64 bytes](https://elixir.bootlin.com/linux/v6.11.1/source/arch/arm64/include/asm/cache.h#L8). The alignment can also be obtained from the **DWARF** info i.e. `DW_AT_alignment<64>`. Unfortunately **dwarf2json** doesn't support this feature yet.
> In kernels < 4.2, alignment attributes are absent in the `struct module`, meaning alignment cannot be guaranteed. Therefore, for older kernels, it's better to use the traditional scan technique.
> 

It checks whether Linux kernel module addresses are aligned to 64 bytes. If not, it falls back to 1-byte alignment and notifies the user of the adjustment.

# Demos
The following is an example of a machine infected with the **Reptile** rootkit. The machine has 2048MB of RAM.
```shell
$ time python3 ./vol.py \
    -r pretty \
    -f ../ubuntu180464bit_4.15.0-213-generic_reptile_infected.core \
    linux.hidden_modules
Volatility 3 Framework 2.10.0      
  |        Address |           Name
* | 0xffffc094f400 | reptile_module

real    0m16.501s
user    0m15.378s
sys     0m0.851s
```

Here is a Kovid rootkit sample provided by @Abyss-W4tcher :
```shell

$ time python3 ./vol.py \
    -r pretty \
    -f ../Ubuntu-jammy_5.15.0-87-generic_kovid.lime \
    linux.hidden_modules
Volatility 3 Framework 2.10.0
  |        Address |  Name
* | 0xffffc09ed4c0 | kovid

real    0m16.932s
user    0m16.187s
sys     0m0.739s
```

We also observed that advanced malware can modify certain values that are invalid when the module is loaded but have no impact once it's running. This technique enables sophisticated threats to evade detection from existing memory forensics methods.

The following is a slightly modified version of the **Kovid rootkit**, detectable only through the `fast scan` method.
```shell
$ time python3 ./vol.py \
    -r pretty \
    -f ../dump_ubuntu20.04_5.15.0-87-generic_kovid_99.core \
    linux.hidden_modules
  |        Address |  Name
* | 0xffffc07f8600 | kovid

real    0m18.233s
user    0m17.281s
sys     0m0.676s
```


Abyss-W4tcher:
Hello, awesome submission, this feature allows to uncover deeply hidden rootkits which is extremely valuable.

Having worked on this problematic for the 2023 contest, I ran your plugin against an infected sample, but unfortunately got no results. Would you be interested in acquiring this sample ("kovid" rootkit), to check if this might come from a scanning or restrictive constraint issue ?

Thanks again !

gcmoreira:
@Abyss-W4tcher interesting.. sure, could you share that with me?

Abyss-W4tcher:
A first restricting point might be the `MODULE_STATE_LIVE` assertion, whereas some rootkits modify this attribute to act "unloaded" :

- https://github.com/carloslack/KoviD/blob/master/src/kovid.c#L260

gcmoreira:
Thanks @Abyss-W4tcher .. easy fix with no impact on performance.
```shell
$ time python3 ./vol.py \
	-f ../Ubuntu-jammy_5.15.0-87-generic_kovid.lime \
	linux.hidden_modules
Volatility 3 Framework 2.10.0
Address Name

0xffffc09ed4c0  kovid

real    1m12.489s
user    1m11.930s
sys     0m0.555s

$ time python3 ./vol.py \
	-f ../Ubuntu-jammy_5.15.0-87-generic_kovid.lime \
	linux.hidden_modules \
	--fast
Volatility 3 Framework 2.10.0
Address Name

0xffffc09ed4c0  kovid

real    0m15.385s
user    0m14.835s
sys     0m0.543s
```

gcmoreira:
@Abyss-W4tcher added your suggestion. It's running even faster now. Thanks

gcmoreira:
Since the `fast scan` method is performing exceptionally well and can detect threats that the traditional method misses, I think we should rename it, any ideas @Abyss-W4tcher @ikelos ?

Another option would be to move it to a new plugin to ensure users don't overlook it, as it might go unnoticed if it's just an argument.

Abyss-W4tcher:
On my take, adding multiple plugins "doing the same thing" might only confuse users.

Maybe removing the vol2 method and implicitely relying on fast might be better (forensic wise) :

- l1 alignement OK : do the "fast" method
- l1 alignement NOK : do the "fast" method but instead of aligning to 64 bytes, scan byte per byte / every 8 bytes (pointer size)

I have some older infected samples (~kovid like) on which I can test this idea, I'll keep you informed if it is reliable.

gcmoreira:
maybe better going backwards 64, 32, 16, 8, 1 ... and keeping the already tested address in a set

gcmoreira:
> maybe better going backwards 64, 32, 16, 8, 1 ... and keeping the already tested address in a set

hm actually, if we are going to test each byte, it doesn't make sense to test the others. If  `_validate_alignment_patterns()` fails with 64 bytes, we should fallback to scan with 1 byte alignment and that's it.

If you see the following metrics, it still performs really good.

### Forcing 8 bytes alignment
```shell
$ time ./vol.py -f ../dump_ubuntu20.04_5.15.0-87-generic_kovid_99.core  linux.hidden_modules --fast 
Volatility 3 Framework 2.10.0               
Address Name

0xffffc07f8600  kovid

real    0m31.045s
user    0m29.886s
sys     0m0.820s
```

### Forcing 1 byte alignment
```shell
$ time ./vol.py -f ../dump_ubuntu20.04_5.15.0-87-generic_kovid_99.core  linux.hidden_modules --fast 
Volatility 3 Framework 2.10.0
Address Name

0xffffc07f8600  kovid

real    2m34.798s
user    2m33.008s
sys     0m0.837s
```

### The Vol2 method takes 1m9s but it doesn't find it
```shell
$ time ./vol.py -f ../dump_ubuntu20.04_5.15.0-87-generic_kovid_99.core  linux.hidden_modules
Volatility 3 Framework 2.10.0
Address Name


real    1m8.985s
user    1m7.025s
sys     0m0.852s
```

Abyss-W4tcher:
Looks great, so it doesn't try `_validate_alignment_patterns()` with 64, 32, 16, 8 and 1, but only 64, 8 and 1 ?

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -0,0 +1,414 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import re
+import logging
+import contextlib
+from typing import List, Set, Tuple, Iterable
+from volatility3.framework import renderers, interfaces, exceptions, objects
+from volatility3.framework.constants.architectures import LINUX_ARCHS
```
ikelos: Sorry, please can we import architectures and then access `architectures.LINUX_ARCHS` when we want to use it?

gcmoreira:
@Abyss-W4tcher only 1 byte and 8 bytes alignments. It was just to see how this would impact performance.
```python
module_address_alignment = 1 # 8 # cls._get_module_address_alignment(context, vmlinux_module_name)
```

Abyss-W4tcher:
Ah yes, this looks like a good compromise for me, as it's always preferable to spend more time but still detect advanced threats in the end. Should the vol2 method be left as a reference ?

atcuno:
@gcmoreira @Abyss-W4tcher what is the status of this one? It seems high priority based on it blocking other PRs

gcmoreira:
Okay, I made the fast scan method the default. Removed the vol2 implementation and updated the plugin to fall back to 1-byte alignment scanning if addresses aren't aligned with the L1 cache size. Now it's much easier to use, with less code and significantly more powerful!

Thanks @Abyss-W4tcher and @atcuno for your help and suggestions.

@ikelos this is now ready for review

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -0,0 +1,248 @@
...
+            mod_tree = vmlinux.object_from_symbol("mod_tree")
+            modules_addr_min = mod_tree.addr_min
+            modules_addr_max = mod_tree.addr_max
+        elif vmlinux.has_symbol("module_addr_min"):
+            modules_addr_min = vmlinux.object_from_symbol("module_addr_min")
+            modules_addr_max = vmlinux.object_from_symbol("module_addr_max")
+
+            if isinstance(modules_addr_min, objects.Void):
+                # Crap ISF! Here's my best-effort workaround
+                vollog.warning(
```
ikelos: Do we have a specific version number now?  Is this the kind of thing we can use a filter to detect if it's bad and warn them once at the start of the plugin?

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -0,0 +1,248 @@
...
+                    "Your ISF symbols are missing type information. You may need to update "
+                    "the ISF using the latest version of dwarf2json"
+                )
+                # See issue #1041. In the Linux kernel these are "unsigned long"
```
ikelos: This was a clang issue as I recall?  It almost feels like it would be easier to inject the missing base_type into the symbol_table once we've found out which one was there?  That would then work anywhere else you wanted to use it, and could become a general function/pattern that could be pulled out into the framework if it was useful?

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -0,0 +1,248 @@
...
+                for type_name in ("long unsigned int", "unsigned long"):
+                    if vmlinux.has_type(type_name):
+                        modules_addr_min = modules_addr_min.cast(type_name)
+                        modules_addr_max = modules_addr_max.cast(type_name)
+                        break
+                else:
+                    raise exceptions.VolatilityException(
+                        "Bad ISF! Please update the ISF using the latest version of dwarf2json"
```
ikelos: Whilst I appreciate the sentiment, this makes me feel like we're telling off our faithful dog "ISF" for having made a mess of the kitchen...  5:P  Could we make this a little more formal/professional please, or at least just drop the exclamation mark?

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -0,0 +1,248 @@
...
+        will still work.
+
+        Args:
+            context: The context to retrieve required elements (layers, symbol tables) from
+            vmlinux_module_name: The name of the kernel module on which to operate
+
+        Returns:
+            The struct module alignment
+        """
+        # FIXME: When dwarf2json/ISF supports type alignments. Read it directly from the type metadata
```
ikelos, line 91: Thanks for the comment and the forward thinking parameters.  5:)

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -0,0 +1,248 @@
...
+            vollog.warning(
+                f"Module addresses aren't aligned to {module_address_alignment} bytes. "
+                "Switching to 1 byte aligment scan method."
+            )
+            module_address_alignment = 1
+
+        mkobj_offset = vmlinux.get_type("module").relative_child_offset("mkobj")
+        mod_offset = vmlinux.get_type("module_kobject").relative_child_offset("mod")
+        offset_to_mkobj_mod = mkobj_offset + mod_offset
+        mod_member_template = vmlinux.get_type("module_kobject").vol.members["mod"][1]
```
ikelos: This is a little bit nasty, I feel there should be something to get to `member["mod"]` in some way, and the `[1]` ?  I'm pretty sure that from an `ObjectTemplate` you can get a child template using `.child_template('mod')`?  I *really* want to find a cleaner way of expressing this before it lands.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -36,6 +35,30 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._mod_mem_type = None  # Initialize _mod_mem_type to None for memoization
 
+    def is_valid(self):
```
ikelos, line 29: Could we get a docstring here detailling exactly what `is_valid` means, since we've got `is_valid` in some places and `is_readable` in others and people are really gonna need to know exactly what this does from the documentation.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -36,6 +35,30 @@ def __init__(self, *args, **kwargs):
...
+        layer = self._context.layers[self.vol.layer_name]
+        # Make sure the entire module content is readable
+        if not layer.is_valid(self.vol.offset, self.vol.size):
+            return False
+
+        core_size = self.get_core_size()
+        if not (
+            1 <= core_size <= 20000000
```
ikelos: These hardcoded limits come about from...?  For arbitrary values like 20000000, I'd prefer that they were stashed in `constants.linux` with a brief comment to describe them, so that they're consistent and can be changed all at once if needed.

ikelos (commented):
Generally pretty good, but a couple of points that have to be fixed before it goes live (even if I have to write new core methods to achieve it!).  First up is using `child_template` rather than directly addressing the structure of the `vol.members` structure.  Secondly shifting the hardcoded limits off to the `constants` file so they're a bit more obvious than being buried away in the middle of code somewhere.

Otherwise just some documentation other little nitpicks and it should be good to go.  5:)

volatility3/framework/plugins/linux/hidden_modules.py
gcmoreira: Ups.. yeah.. my bad

volatility3/framework/plugins/linux/hidden_modules.py
gcmoreira: done

volatility3/framework/plugins/linux/hidden_modules.py
gcmoreira: Yes, I simplified the checks here and added a verify_table_version in run(), similar to how it’s done in the kmsg plugin.

volatility3/framework/plugins/linux/hidden_modules.py
gcmoreira: I removed this.

volatility3/framework/plugins/linux/hidden_modules.py
gcmoreira: I removed this

volatility3/framework/plugins/linux/hidden_modules.py
gcmoreira: done

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: yeah, see vol2

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: hm, they serve very different purposes.
On the one hand, `is_readable()` is for checking if a pointer is readable .. based on what I explained on https://github.com/volatilityfoundation/volatility3/pull/1243 to correctly check a pointer we should do, almost in all cases, the following:
```python
ptr and ptr.is_readable()
```
On the other hand, `module::is_valid()` is for checking that the content of the module is valid. 
Anwyay, I will add a docstring explaining what the `is_valid()` method does if that helps

ikelos: `is_valid` is also used in the framework to mean "layer has a readable address at this offset", which is why I was hoping to get it documented.  This new `is_valid` seems to do a third thing, which is why I wanted to ensure the naming is consistent before we start using `is_valid` for two different things.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Errr, let's keep our reasoning in our code, rather than referring people off to a no-longer supported tool please?  If we're hardcoding a value, we should say why and if possible, where it comes from, and we should keep it in a big list of such hardcoded randomness like the `constants.linux` module.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: done

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -13,12 +13,24 @@
 
 from volatility3.framework import constants, exceptions, objects, interfaces, symbols
 from volatility3.framework.renderers import conversion
-from volatility3.framework.constants.linux import SOCK_TYPES, SOCK_FAMILY
-from volatility3.framework.constants.linux import IP_PROTOCOLS, IPV6_PROTOCOLS
-from volatility3.framework.constants.linux import TCP_STATES, NETLINK_PROTOCOLS
-from volatility3.framework.constants.linux import ETH_PROTOCOLS, BLUETOOTH_STATES
-from volatility3.framework.constants.linux import BLUETOOTH_PROTOCOLS, SOCKET_STATES
-from volatility3.framework.constants.linux import CAPABILITIES, PT_FLAGS
+from volatility3.framework.constants.linux import (
```
ikelos: This might be easier to view as `from ...constants import linux` and then refer to `linux.SOCK_TYPES`, etc?  Seems like it might tidy the imports and keep with the coding style too...

ikelos (commented):
Looks good, I'd like the change on the constants import, but it's not a show stopper.  Just lemme know when you're happy for it to go in and we can put it in...

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: I agree, it was quite a mess. I attempted to improve it a bit with that change, but it really needed the changes you're suggesting. It's done now. 
I renamed the import to `linux_constants` since `linux` was already used for the `symbols` import. I believe this makes it clearer, as it identifies exactly what it is rather than just a bare `linux`.

ikelos (approved):
Sadly one of the other recent merges conflicts, but if you can get that sorted, this looks good to go straight in now.  5:)

ikelos:
Looks like it's just the `cached_property` change, so should be trivial to rebase/fix up.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Yep, makes perfect sense, good call!  5:)

gcmoreira:
Cool, it's ready to go. Thanks @ikelos

PR #1284, opened 2024-10-02
Fix misuse of all() and any() functions across the codebase

gcmoreira:
We are misusing the `all()` and `any()` Python's built-in functions. Wrapping the iterable in [] converts it into a list before the function is called, which unnecessarily increases the runtime to the worst-case which is O(n) in every case.

## What we are currently doing:
```python
>>> import timeit
>>> timeit.timeit('any([n > 0 for n in range(10000)])', number=10000)
4.729606243010494
```

## How it should be:
```python
>>> import timeit
>>> timeit.timeit('any(n > 0 for n in range(10000))', number=10000)
0.019788431003689766
```

### DIFF
--- a/volatility3/cli/text_filter.py
+++ b/volatility3/cli/text_filter.py
@@ -53,7 +53,7 @@ class CLIFilter:
         """Filters the row based on each of the column_filters"""
         if not self._filters:
             return False
-        found = any([column_filter.found(row) for column_filter in self._filters])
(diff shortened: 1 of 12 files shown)


gcmoreira:
@ikelos let me know if you want me to change any version number

ikelos (approved):
Nope, as long as the results are the same no versions need bumping.  Thanks for the efficiency boost, I've been doing it wrong all this time!

PR #1289, opened 2024-10-03
Linux: Properly get paths in simple_dname paths

dgmcdona:
Some dentry paths were not being recovered from `simple_dname` paths, this fixes the issue by properly dereferencing the string pointer.

### DIFF
--- a/volatility3/framework/symbols/linux/__init__.py
+++ b/volatility3/framework/symbols/linux/__init__.py
@@ -206,7 +206,14 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
                 pre_name = "pipe"
 
             elif sym == "simple_dname":
-                pre_name = cls._get_path_file(task, filp)
+                name = dentry.d_name.name
+                if name:
+                    pre_name = name.dereference().cast(
+                        "string", max_length=255, errors="replace"
+                    )
+                    return "/" + pre_name + " (deleted)"
+                else:
+                    pre_name = ""
 
             elif sym == "ns_dname":
                 # From Kernels 3.19


volatility3/framework/symbols/linux/__init__.py
```diff
@@ -206,7 +206,14 @@ def _get_new_sock_pipe_path(cls, context, task, filp) -> str:
                 pre_name = "pipe"
 
             elif sym == "simple_dname":
-                pre_name = cls._get_path_file(task, filp)
+                name = dentry.d_name.name
+                if name:
+                    pre_name = name.dereference().cast(
+                        "string", max_length=255, errors="replace"
+                    )
+                    return "/" + pre_name + " (deleted)"
```
ikelos, line 214: I'm concerned that (deleted) should be indicated in some other way?  Otherwise how does this differentiate between a file `thing` that was deleted, and a file literally called `thing (deleted)`?  I don't like embedding other data into a single field.  I'm just thinking of how other tools might try to process this data?  I realize this isn't in the middle of a plugin, so returning additional attributes is tricky, but I'd almost like to have a separate parameter on a filp or something which can be used to determine whether the file has been deleted or not, and then just keep the filename data as it's supposed to be.

ikelos (commented):
I'm worried about the return value of this.  I don't know if there'd be a way of adding an additional parameter to a filp or something, but this feels like its mixing two concerns that shouldn't be mixed.

volatility3/framework/symbols/linux/__init__.py
atcuno: 1) 
re: (deleted), this code is mirroring how the kernel does it. If you (as root/sudo) on any recent Linux system run this:

```
grep deleted /proc/*/maps
```

You will see the files end in (deleted) that were deleted since getting mapped.

If a user named a file as "(deleted)" then the output from this code would report "(deleted) (deleted)".

2)

The current code for mimicing `simple_dname` is completely broken (my fault not catching this earlier), and the current code that prepends "/" and appends " (deleted)" is actually exactly the correct implementation that fully mirrors the kernel code, like here:

https://elixir.bootlin.com/linux/v6.11.2/source/fs/d_path.c#L318

So with the patch we match the kernel, whereas what is in develop is not even close.

ikelos: Ok, if we're not concerned about mixing the state of being deleted with the data that makes up the name, then otherwise the patch looks fine.

PR #1290, opened 2024-10-03
Windows: unloadedmodules bugfix

dgmcdona:
This fixes a bug in the `windows.unloadedmodules` plugin. An `InvalidAddressException` can be raised when parsing the module name; this adds a try/except block to replace the missing module name with a `renderers.UnreadableValue` when the exception occurs.


ikelos (approved):
Ok, as long as you think the data it returns is still valuable, that's fine by me...  5:)

Issue #1265, opened 2024-09-13
Unable to Retrieve Results despite correct symbol table and memory dump


misogare:
**Describe the bug**
As stated, i have tried to run linux.pslist.PsList on several different distributions including kali,debian,ubuntu with precise symbols and configuration either manually grabbed or from repositories but still i get similar log as this :
`  ─$ python3 vol.py -vvv -f /mnt/hgfs/shared/memd.raw   linux.pslist.PsList                          
Volatility 3 Framework 2.10.0
INFO     volatility3.cli: Volatility plugins path: ['/home/kali/volatility3/volatility3/plugins', '/home/kali/volatility3/volatility3/framework/plugins']
INFO     volatility3.cli: Volatility symbols path: ['/home/kali/volatility3/volatility3/symbols', '/home/kali/volatility3/volatility3/framework/symbols']
DEBUG    volatility3.plugins.yarascan: Using yara-python module
INFO     volatility3.framework.automagic: Detected a linux category plugin
INFO     volatility3.framework.automagic: Running automagic: ConstructionMagic
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
[...]

Issue #965, opened 2023-06-11
volatility3 doesn't show anything with linux ubuntu .mem file


marseaplage:
**Describe the bug**
Volatility3 doesn't show anything with any plugin

**Context**
Volatility Version:  Volatility 3 Framework 2.4.2
Operating System: Ubuntu 20.04  
Python Version:  Python 3.8.10
Suspected Operating System:  
Command:  

**To Reproduce**
I follow the same steps as this tutorial from the section "Linux ISF" and my kernel is 5.4.0-150-generic

1. Use command 'sudo python3 ./vol.py -f /home/prueba/ubuntu-10062023.mem linux.pstree'
2. See error
Volatility 3 Framework 2.4.2
Progress:  100.00               Stacking attempts finished                 
OFFSET (V)      PID     TID     PPID    COMM

(empty)

**Expected behavior**
To show processes

**Example output**
Please copy and paste the text demonstrating the issue, ideally with verbose output turned on (`vol.py -vvv ...`).  

other output: 

INFO     volatility3.cli: Volatility plugins path: ['/home/prueba/volatility3/volatility3/plugins', '/home/prueba/volatility3/volatility3/framework/plugins']
INFO     volatility3.cli: Volatility symbols path: ['/home/prueba/volatility3/volatility3/symbols', '/home/prueba/volatility3/volatility3/framework/symbols']
INFO     volatility3.plugins.yarascan: Python Yara (>3.8.0) module not found, plugin (and dependent plugins) not available
DEBUG    volatility3.framework: Traceback (most recent call last):
  File "/home/prueba/volatility3/volatility3/framework/__init__.py", line 185, in import_file
    importlib.import_module(module)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 848, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/prueba/volatility3/volatility3/framework/plugins/yarascan.py", line 20, in <module>
    raise ImportError
ImportError

[...]

Issue #413, opened 2021-01-04
linux.pslist returns no results


RafaelAybar:
**Volatily 3  is not able to find plugins**
I made a memory dump with Lime in raw format, and i wanted to process it with Volatility3. I have installed all dependencies even the optional dependencies.

**Context**
Volatility Version:  Volatility 3 Framework 2.0.0 Beta
Operating System:  Debian Testign
Python Version:  3.9.1
Kernel Version: 5.9.0.5-amd64
Command:  `python3 vol.py -vvvv  -s symbols/ --file /linux.mem  linux.bash.Bash`

**To Reproduce**
1- Run a python3 setup.py install
2- Generate JSON file with `dwarf2json linux --elf /usr/lib/debug/boot/vmlinux-5.9.0.5-amd64 > output.json`

**Expected behavior**
Volatility3 shoud find the plugins

**Command output**

https://pastebin.com/vdW3s7Ar

ikelos:
So this has nothing to with the plugins.  It clearly says that it can't determine the information it needs to load the layer (and therefore can't find the symbols it might need).  When using dwarf2json you should also use the System.map, because the debugging kernel may not contain all the necessary symbols.  You can then also check that the symbols you've created (output.json) live in the `volatility/symbols/linux` directory, and show up in the output of the `isfinfo` plugin (this can take a long time to run if you have lots of symbol files).  Once it's in there, you should see which banner it's looking for, and you can then run the `banners` plugin against your image and see if any of the banners it finds match the ones in the ISF file.  Please report back the results of these steps, and if you're still experiencing difficulty after that...  5:)

RafaelAybar:
Hi, tanks a lot for your fast answer, i uncompressed the linux.zip file,  and commpresed the folder linux  with output.json in order to generate another linux.zip file. Also, i have installed jsonschema beacuse it's another dependency.

Here you are the logs: https://pastebin.com/681Jv8yW

ikelos:
[...]

PR #1295, opened 2024-10-07
Docs: Add warning around LiME raw format

eve-mem:
Hello :wave: 

I was looking at the docs and thought it might be useful to have a little warning against using the LiME raw format. The raw format doesn't include the padding needing for vol to be able to parse the memory as the different segments of raw memory get misaligned. 

It's caused a few issues for us:
- https://github.com/volatilityfoundation/volatility3/issues/965
- https://github.com/volatilityfoundation/volatility3/issues/413
- https://github.com/volatilityfoundation/volatility3/issues/1265

:fox_face:


ikelos:
Looks good, thanks!

Issue #1296, opened 2024-10-07
Sort volshell methods alphabetically


iMHLv2:
By default, the volshell methods (`hh()` output) are not sorted alphabetically. In this case, I knew `cp` was a method and wanted to read the description, but couldn't it find it anywhere near `cl` or `cs`. 

<img width="1091" alt="image (1)" src="https://github.com/user-attachments/assets/2031f28d-ee8d-4af7-be1b-2953b2238540">

ikelos:
Cool, thanks for logging a ticket again the change.  I'm not sure why I did them in such a weird order or why I didn't think to sort them, but thanks for getting it fixed!  5:)

PR #1297, opened 2024-10-07
refs #1296 sort volshell methods alphabetically in hh() output

iMHLv2:
This sorts the output of `hh()` in volshell alphabetically.


ikelos (commented):
Heheh, yep, sorry I didn't think to do it sooner...  5:S

PR #1300, opened 2024-10-08
[plugin] Add dirty state IoC detection to malfind

Abyss-W4tcher:
Hi,

The current `windows.malfind` implementation does not leverage the dirty state indicator to identify potential threats, and filters out non writable VADs. However, some evasion techniques use an elevated `WriteProcessMemory` call to inject data inside remote processes `PAGE_EXECUTE_READ` regions :

- https://www.covertswarm.com/post/exploiting-microsoft-windows-11-via-process-no-hollowing

> As expected, we invoked WriteProcessMemory, specifying 0x00007FF7519BOECO as
target address (address of entry point in our example), and successfully changed the
read-only region.

> When using WriteProcessMemory to write to Read-only memory addresses we have
observed that associated errors and exceptions are not raised within the OS if we have
sufficient privileges on the target process. The question is, why?

- https://devblogs.microsoft.com/oldnewthing/20181206-00/?p=100415

> If the page is read-only, Write­Process­Memory temporarily changes the permission to read-write, updates the memory, and then restores the original permission.

To identify this behaviour, I edited the malfind code to scan each VAD page with `(R)-X` protection for any dirty state marker. 

If necessary, I can develop a PoC + sample (which might come handy for unit tests).

Inspired from Linux https://github.com/volatilityfoundation/volatility3/pull/995.

### DIFF
--- a/volatility3/framework/plugins/windows/malfind.py
+++ b/volatility3/framework/plugins/windows/malfind.py
@@ -122,10 +122,30 @@ class Malfind(interfaces.plugins.PluginInterface):
                 vadinfo.winnt_protections,
             )
             write_exec = "EXECUTE" in protection_string and "WRITE" in protection_string
+            dirty_page_check = False
 
-            # the write/exec check applies to everything
             if not write_exec:
-                continue
+                """
+                # Inspect "PAGE_EXECUTE_READ" VAD pages to detect
+                # non writable memory regions having been injected
+                # using elevated WriteProcessMemory().
+                """
+                if "EXECUTE" in protection_string:
+                    for page in range(
+                        vad.get_start(), vad.get_end(), proc_layer.page_size
+                    ):
+                        try:
+                            # If we have a dirty page in a non writable "EXECUTE" region, it is suspicious.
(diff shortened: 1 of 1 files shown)


atcuno:
Do you have some Windows 10 samples with the default Defender enabled state to test this? If not, we can run it over some of them soon.

Abyss-W4tcher:
I don't possess any sample infected by this technique, or by malware performing an RWX allocation, shellcode injection and then changing protection back to RX.

Sure, it'd be great to have this feature ran over many samples !

atcuno:
Ok we will run it across our sample set (a couple hundred Windows samples) and report back before/after this patch.

Abyss-W4tcher:
Hi @atcuno, any news on the testing of this feature ?

ikelos:
I'm not sure why this is still failing on the black test.  Could you try pushing a trivial commit to it please?  I thought they'd resolved the black issues, but this one still seems to be failing the tests...

ikelos:
Awesome, thanks!  5:)

ikelos (approved):
Looks good.

PR #1308, opened 2024-10-11
Linux: Netfilter - fix traceback on missing module name

dgmcdona:
When the module name can't be found, `None` ends up in this field,
causing a traceback when rendering. This fixes it by replacing `None`
with `renderers.NotAvailableValue()`.

This occurred in the context of a netfilter hook from a hidden module.


ikelos:
Looks good, nice spot!

PR #1310, opened 2024-10-11
Framework: Add 'Data' type to base types

dgmcdona:
This adds a `Data` base type, which will provide the information
(layer_name and offset) required to view variable amounts of data
both before and after that offset, instead of handing fixed-sized
chunks of data back.

This will allow plugin consumers to flexibly determine how much
data to render/extract without having to re-run the plugin with
different parameter values.

This PR is the first step in the outline proposed as part of 
the discussion surrounding #1287 - subsequent PRs will follow
that implement rendering of this new data type in the CLI and 
configuration of default parameters, and the conversion of existing
plugins to return this new data type where appropriate (yarascan,
vadyarascan, vmayarascan).

### DIFF
--- a/volatility3/framework/interfaces/renderers.py
+++ b/volatility3/framework/interfaces/renderers.py
@@ -105,9 +105,10 @@ class BaseAbsentValue(object):
 class Data(object):
     """Class that represents a view of data in a particular layer"""
 
-    def __init__(self, offset: int, layer_name: str):
+    def __init__(self, offset: int, layer_name: str, length: int):
         self.layer_name = layer_name
         self.offset = offset
+        self.length = length
 
 
 class Disassembly(object):
--- a/volatility3/framework/interfaces/renderers.py
+++ b/volatility3/framework/interfaces/renderers.py
@@ -102,6 +102,14 @@ class BaseAbsentValue(object):
     """Class that represents values which are not present for some reason."""
 
 
+class Data(object):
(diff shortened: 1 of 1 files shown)


dgmcdona:
@ikelos this PR is very minimal, so please let me know if I'm missing something here - hopefully it's enough to start discussion and get the ball rolling on the next steps.

ikelos:
I think we probably also want to include an expected minimal length.  The
UI/user can override it, but the plugin will probably know about how much
data it expects people to get, and we should include that in the return
value.

I'm quite remote at the moment so I'll try to give it a look when I can but
it may be next week before I get a chance to look it over.

Thanks for mocking up the discussed solution!

ikelos:
Looks good, probably best if we included (at least a basic) display mechanism (so the render method that will display these).  It should be off in its own method (the existing ones are all kinda lambda functions except maybe the dissembly one) and (for now) it should just pull out the length bytes and display them as a hex dump, I think?  Then (in a separate PR) we can figure out how to weave the output options through to the renderers (that code was never developed but there should be a pathway to feed options through to the renderers and we really should develop it properly, even if just a basic version).

So that should allow plugins to start making use of the new type and we should get roughly identical results to what we have now where the actual data is returned, only handled by the UI rather than the plugin.  Then step two will be to let the UI tinker with how it's displayed...

ikelos:
Probably should be closed off in favour of #1622?

PR #1316, opened 2024-10-16
Enable Remote ISF server for Linux testcases

gcmoreira:
This PR introduces changes to enhance our testing capabilities by allowing Linux test cases to use the @Abyss-W4tcher ISF repository on GitHub. 
This not only provides greater flexibility for updating the ISF database and keeping it in sync with the latest `dwarf2json` version, but also establishes a fully open-source workflow. 
This will enable us to easily add more images later on, such as different Linux kernel versions and distributions, which will be immediately ready for testing our code.

If it's successful and accepted, we can also migrate the macOS test cases to use this repository.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -32,23 +32,13 @@ jobs:
         curl -sLO "https://downloads.volatilityfoundation.org/volatility3/images/win-xp-laptop-2005-06-25.img.gz"
         gunzip win-xp-laptop-2005-06-25.img.gz
 
-    - name: Download and Extract symbols
-      run: |
-        cd ./volatility3/symbols
-        curl -sLO https://downloads.volatilityfoundation.org/volatility3/symbols/linux.zip
-        unzip linux.zip
-        cd -
-
     - name: Testing...
       run: |
-        py.test ./test/test_volatility.py --volatility=vol.py --image win-xp-laptop-2005-06-25.img -k test_windows -v
-        py.test ./test/test_volatility.py --volatility=vol.py --image linux-sample-1.bin -k test_linux -v
+        pytest ./test/test_volatility.py --volatility=vol.py --image win-xp-laptop-2005-06-25.img -k test_windows -v
+        pytest ./test/test_volatility.py --volatility=vol.py --image linux-sample-1.bin --remote-isf-url=https://github.com/Abyss-W4tcher/volatility3-symbols/raw/master/banners/banners.json" -k test_linux -v
 
     - name: Clean up post-test
       run: |
         rm -rf *.bin
         rm -rf *.img
-        cd volatility3/symbols
-        rm -rf linux
-        rm -rf linux.zip
         cd -
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -35,7 +35,7 @@ jobs:
     - name: Testing...
       run: |
         pytest ./test/test_volatility.py --volatility=vol.py --image win-xp-laptop-2005-06-25.img -k test_windows -v
-        pytest ./test/test_volatility.py --volatility=vol.py --image linux-sample-1.bin --remote-isf-url=https://github.com/Abyss-W4tcher/volatility3-symbols/raw/master/banners/banners.json" -k test_linux -v
+        pytest ./test/test_volatility.py --volatility=vol.py --image linux-sample-1.bin --remote-isf-url="https://github.com/Abyss-W4tcher/volatility3-symbols/raw/master/banners/banners.json" -k test_linux -v
 
     - name: Clean up post-test
       run: |
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -41,4 +41,4 @@ jobs:
       run: |
         rm -rf *.bin
         rm -rf *.img
-        cd -
+
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,5 +1,11 @@
 name: Test Volatility3
+
 on: [push, pull_request]
+
+env:
+  REMOTE_ISF_URL: https://github.com/Abyss-W4tcher/volatility3-symbols/raw/master/banners/banners.json
+  IMAGES_URL: https://downloads.volatilityfoundation.org/volatility3/images
+
 jobs:
 
   build:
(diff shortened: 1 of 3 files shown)


gcmoreira:
Bloody brilliant! It’s working like a kangaroo on a trampoline! :)

.github/workflows/test.yaml
```diff
@@ -32,23 +32,13 @@ jobs:
...
-        curl -sLO https://downloads.volatilityfoundation.org/volatility3/symbols/linux.zip
-        unzip linux.zip
-        cd -
-
     - name: Testing...
       run: |
-        py.test ./test/test_volatility.py --volatility=vol.py --image win-xp-laptop-2005-06-25.img -k test_windows -v
-        py.test ./test/test_volatility.py --volatility=vol.py --image linux-sample-1.bin -k test_linux -v
+        pytest ./test/test_volatility.py --volatility=vol.py --image win-xp-laptop-2005-06-25.img -k test_windows -v
+        pytest ./test/test_volatility.py --volatility=vol.py --image linux-sample-1.bin --remote-isf-url="https://github.com/Abyss-W4tcher/volatility3-symbols/raw/master/banners/banners.json" -k test_linux -v
```
ikelos: This can also be coded into a JSON file if it needs to be reused?  Volatility reads `~/.config/volatility3/vol.json` for its defaults, if that's easier than passing it on the command line each time?  I guess it depends how often it needs to be passed?

test/test_volatility.py
```diff
@@ -38,19 +37,30 @@ def runvol(args, volatility, python):
...
+    plugin_args += pluginargs if pluginargs else []
+    global_args = globalargs or []
+
+    common_args = [
+        "--single-location",
+        img,
+        "-q",
+    ]
+    if remote_isf_url:
+        common_args += ["--remote-isf-url", remote_isf_url]
```
ikelos, line 59: Oh, looks like it gets added automatically, so no need for the `vol.config` I guess, that's not too cumbersome at least.

test/test_volatility.py
```diff
@@ -126,10 +136,7 @@ def test_windows_dumpfiles(image, volatility, python):
 
     failed_chksms = 0
 
-    if sys.platform == "win32":
-        file_name = ntpath.basename(image)
-    else:
-        file_name = os.path.basename(image)
+    file_name = os.path.basename(image)
```
ikelos, line 139: Was this unnecessary, or do we just no longer intend to run tests on windows systems?

ikelos (commented):
Thanks, I'm roughly happy with the concept, but I'd like to run it past @npetroni who I believe made the first testing harness, and also get his views on downloading additional files from a server we don't control, which my security hat is just making me wonder about.  Another option would be to create our own mini repo with symbols just for the test cases we've got but I can't tell if that's prudent or paranoid.  I don't *think* a bad JSON file could do anything bad to us, it *shouldn't* be able to execute anything so the worst case is it causes our tests to fail, and we do already download the ISF files from a server (albeit a server we own).  So probably ok, but I'd like @npetroni 's opinion on it first please...

Abyss-W4tcher:
Hi, just passing by as I was mentioned. Thanks for wanting to integrate my repo as a symbol source, it's much appreciated !

However, if I take a neutral point of view, this increases the risk of supply chain bugs. Even if I put efforts to prevent that from happening, I think each testcase should come with its own ISF attached, hosted by volatility foundation ?

I am not trying to sabotage your work, but I am not sure that the remote isf feature fits this use case ?

Thanks again 👍

ikelos:
Yeah, that was my thinking, it's just a text file on a webserver that points the banner to a downloadable file, so we could definitely look after that and keep it updated with our testcases, and then that saves @Abyss-W4tcher from become an integral service to out testcases.  5;P  It will mean that we'll have to get better/quicker at updating those JSON files and where they're stored, but that's definitely a solvable "us" problem rather than something insurmountable.  5:)

@npetroni would you be happy with that (and/or, could you help us get it all in place please?)  5:)

test/test_volatility.py
gcmoreira: No, it's because that's precisely what the standard Python module `os.path` is for ... you don't need to check the OS yourself and pick one unless you specifically need to force it, which it doesn't seem the case. 

From https://docs.python.org/3/library/os.path.html

> Note: Since different operating systems have different path name conventions, there are several versions of this module in the standard library. The `os.path` module is always the path module suitable for the operating system Python is running on, and therefore usable for local paths. However, you can also import and use the individual modules if you want to manipulate a path that is always in one of the different formats. They all have the same interface:
> - posixpath for UNIX-style paths
> - ntpath for Windows paths

ikelos:
Hiya, could we please get the remote URL turned into a variable so it's only defined once clearly at the top of the test cases?  We're looking at making our own little ISF github repo just for the test materials to start off with.  We'll let you know what the URL will change to once we've started making it...

gcmoreira:
You mean to move the URL from being a parameter in the YAML/pytest to hard-coding it in `test_volatility.py`. 
I can make that change, but it seems less flexible overall. Right now, it's already defined once in this implementation, and keeping it as a pytest parameter allows for easy command-line changes during development, without modifying the code itself. I could be overlooking something, but the suggested approach seems like a step backward to when it was a hard-coded variable in the source.

ikelos:
No, I just mean as a variable in the YAML/pytest file, rather than in the middle of the command statement.  I'm pretty sure the yaml allows parameterization? 

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-environment-variables-for-a-single-workflow

So setting an env, (ideally pulling out $IMAGES_URL or something too, so that the duplicated parts of the image URL are in a single place) and turning it into a $REMOTE_ISF_URL env variable, so it's easier to see and the command used in the test is easier to read.

gcmoreira:
@ikelos Cool, all done

ikelos:
Awesome, now we just need a little local volatilityfoundation repo to keep the test ISFs in, and we should be good to go.  I believe @npetroni is working on that, but I might be able to get that moving next week since I've got a week off so hopefully I'll do a bunch of volatility housekeeping stuff...  5:D

ikelos:
@Abyss-W4tcher, how about shifting your repo over as a new year's present for 2026?  5:D  We've got the repo (https://github.com/volatilityfoundation/volatility3-symbols), so this could be a straight up copy of your existing repo and then we could set it as the default and avoid supply chain attacks!  5:D  If you've got permission problems, lemme know and we'll get them sorted.  5:)

PR #1317, opened 2024-10-18
Linux - Boottime support

gcmoreira:
In this PR we introduce the boot time support to the Volatility3 framework.

This enables other plugins to provide UTC datetime information based on the system boot time. A prime example is the `linux.pslist` plugin, which has been enhanced as part of this effort. Additionally, we have introduced a new `linux.boottime` plugin to retrieve the system boot time date, supporting also time namespaces (kernels >= 5.6). Both plugins offer support for the Timeliner.

## linux.pslist

Note: The task creation time is seen from the root time namespece, which is what's expected.

### 6.8.0-41
```shell
$ ./vol.py \
    -r pretty \
    -f ./dump_ubuntu2404amd64_6.8.0-41-generic_clocks_ns.core \
    linux.pslist 
Volatility 3 Framework 2.11.0
  |     OFFSET (V) |  PID |  TID | PPID |            COMM |                  CREATION TIME | File output
* | 0x8a8881300000 |    1 |    1 |    0 |         systemd | 2024-10-17 09:17:39.056885 UTC |    Disabled
* | 0x8a8881305200 |    2 |    2 |    0 |        kthreadd | 2024-10-17 09:17:39.056885 UTC |    Disabled
* | 0x8a8881302900 |    3 |    3 |    2 | pool_workqueue_ | 2024-10-17 09:17:39.056885 UTC |    Disabled
...
```
### 4.4.0-142
```shell
$ ./vol.py \
    -r pretty \
    -f ./dump_ubuntu140464bit_4.4.0-142-generic_clocks.core \
    linux.pslist 
Volatility 3 Framework 2.11.0
  |     OFFSET (V) | PID | TID | PPID |            COMM |                  CREATION TIME | File output
* | 0x88001de40000 |   1 |   1 |    0 |            init | 2024-10-18 00:17:01.032000 UTC |    Disabled
* | 0x88001de40f00 |   2 |   2 |    0 |        kthreadd | 2024-10-18 00:17:01.032000 UTC |    Disabled
* | 0x88001de41e00 |   3 |   3 |    2 |     ksoftirqd/0 | 2024-10-18 00:17:01.036000 UTC |    Disabled
...
```
### 3.13.0-185
```shell
$ ./vol.py \
    -r pretty \
    -f ./dump_ubuntu120464bit_3.13.0-185-generic_clocks.core \
    linux.pslist 
Volatility 3 Framework 2.11.0
  |     OFFSET (V) |  PID |  TID | PPID |            COMM |                  CREATION TIME | File output
* | 0x88001d7b0000 |    1 |    1 |    0 |            init | 2024-10-17 08:54:37.044000 UTC |    Disabled
* | 0x88001d7b1800 |    2 |    2 |    0 |        kthreadd | 2024-10-17 08:54:37.044000 UTC |    Disabled
* | 0x88001d7b3000 |    3 |    3 |    2 |     ksoftirqd/0 | 2024-10-17 08:54:37.044000 UTC |    Disabled
* | 0x88001d7b4800 |    4 |    4 |    2 |     kworker/0:0 | 2024-10-17 08:54:37.044000 UTC |    Disabled
```

### 3.2.0-4
```shell
$ ./vol.py \
    -r pretty \
    -f ./dump_debian764bit_3.2.0-4-amd64_clocks2.core \
    linux.pslist 
Volatility 3 Framework 2.11.0
  |     OFFSET (V) |  PID |  TID | PPID |            COMM |                  CREATION TIME | File output
* | 0x88001f900740 |    1 |    1 |    0 |            init | 2024-10-16 23:18:24.016001 UTC |    Disabled
* | 0x88001f900040 |    2 |    2 |    0 |        kthreadd | 2024-10-16 23:18:24.016001 UTC |    Disabled
* | 0x88001f903780 |    3 |    3 |    2 |     ksoftirqd/0 | 2024-10-16 23:18:24.020001 UTC |    Disabled
* | 0x88001f9077c0 |    5 |    5 |    2 |     kworker/u:0 | 2024-10-16 23:18:24.020001 UTC |    Disabled
...
```

## linux.boottime
In this first example, we will examine the boot time on a system with multiple time namespaces. 

Let's start by inspecting the root time namespace.
```shell
$ date +"%z +%Z"
+0000 +UTC

$ readlink /proc/$$/ns/time
time:[4026531834]

$ uptime --since
2024-10-10 11:29:21
```
Next, let's create a new time namespace with a different boot time, -2days:
```shell
$ sudo unshare --time --mount-proc --pid --fork --boottime $((2*24*60*60)) bash

# readlink /proc/$$/ns/time
time:[4026532228]

# uptime --since
2024-10-08 11:29:21
```

It's time to examine a memory dump that includes the steps mentioned above.
```shell
$ ./vol.py \
    -r pretty \
    -f ./dump_ubuntu2404amd64_6.8.0-41-generic_clocks2.core  \
    linux.boottime 
Volatility 3 Framework 2.11.0
  |    TIME NS |                      Boot Time
* | 4026531834 | 2024-10-10 11:29:21.866469 UTC
* | 4026532228 | 2024-10-08 11:29:21.866469 UTC
```

Other kernels tests:

### 6.8.0-41
```shell
$ ./vol.py \
    -r pretty \
    -f ./dump_ubuntu2404amd64_6.8.0-41-generic_clocks_new_ns.core \
    linux.boottime 
Volatility 3 Framework 2.11.0
  |    TIME NS |                      Boot Time
* | 4026531834 | 2024-10-17 09:17:39.758709 UTC
* | 4026532227 | 2024-10-15 09:17:39.758709 UTC
```

### 4.4.0-142
```shell
$ ./vol.py \
    -r pretty \
    -f ./dump_ubuntu140464bit_4.4.0-142-generic_clocks.core \
    linux.boottime 
Volatility 3 Framework 2.11.0
  | TIME NS |                      Boot Time
* |       - | 2024-10-18 00:17:01.597735 UTC
```
### 3.13.0-185
```shell
$ ./vol.py \
    -r pretty \
    -f ./dump_ubuntu120464bit_3.13.0-185-generic_clocks.core \
    linux.boottime 
Volatility 3 Framework 2.11.0
  | TIME NS |                      Boot Time
* |       - | 2024-10-17 08:54:37.076344 UTC
```
### 3.2.0-4
```shell
$ ./vol.py \
    -r pretty \
    -f ./dump_debian764bit_3.2.0-4-amd64_clocks2.core \
    linux.boottime 
Volatility 3 Framework 2.11.0
  | TIME NS |                      Boot Time
* |       - | 2024-10-16 23:18:24.063932 UTC
```

## Timeliner
```shell
 $ ./vol.py \
    -r pretty \
    -f ./dump_ubuntu2404amd64_6.8.0-41-generic_clocks2.core \
    timeliner.Timeliner \
    --plugin-filter linux.boottime linux.pslist
Volatility 3 Framework 2.11.0
  |   Plugin |                                       Description |                   Created Date | Modified Date | Accessed Date | Changed Date
* | Boottime |    System boot time for time namespace 4026532228 | 2024-10-08 11:29:21.866469 UTC |           N/A |           N/A |          N/A
* |   PsList |             Process 1/1 systemd (167256783890688) | 2024-10-10 11:29:21.055809 UTC |           N/A |           N/A |          N/A
* |   PsList |            Process 2/2 kthreadd (167256783880192) | 2024-10-10 11:29:21.055809 UTC |           N/A |           N/A |          N/A
* |   PsList |     Process 3/3 pool_workqueue_ (167256783901184) | 2024-10-10 11:29:21.055809 UTC |           N/A |           N/A |          N/A
...
* | Boottime |    System boot time for time namespace 4026531834 | 2024-10-10 11:29:21.866469 UTC |           N/A |           N/A |          N/A
* |   PsList | Process 167/167 kworker/R-raid5 (167257237148160) | 2024-10-10 11:29:21.963241 UTC |           N/A |           N/A |          N/A
* |   PsList |     Process 206/206 jbd2/dm-0-8 (167257237367040) | 2024-10-10 11:29:22.145494 UTC |           N/A |           N/A |          N/A
* |   PsList | Process 207/207 kworker/R-ext4- (167257237356544) | 2024-10-10 11:29:22.145652 UTC |           N/A |           N/A |          N/A
```


gcmoreira:
Hey @ikelos, the tests are failing because the ISF in linux.zip (`linux-image-3.2.0-4-amd64-dbg_3.2.57-3+deb7u2_amd64.json.xz`) doesn't contain the symbol types. So in [this](https://github.com/volatilityfoundation/volatility3/pull/1317/files#diff-dc8047d2b21ebae7092368b7efe49ec11221e192735929c2be49998e31335956R512) line the call to `vmlinux.object_from_symbol()` returns a void type which then fails in `TimespecVol3.new_from_timespec()` because it's the wrong type.

The ISF is circa 2018 and needs to be updated.
```shell
$ unzip -l linux.zip linux/linux-image-3.2.0-4-amd64-dbg_3.2.57-3+deb7u2_amd64.json.xz
Archive:  linux.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
   776408  2018-07-19 03:04   linux/linux-image-3.2.0-4-amd64-dbg_3.2.57-3+deb7u2_amd64.json.xz
---------                     -------
   776408                     1 file
```

Alternatively, this will be fixed once #1316 is merged. See the following using the same sample:
```shell
$ ./vol.py \
    --remote-isf-url 'https://github.com/Abyss-W4tcher/volatility3-symbols/raw/master/banners/banners.json' \
    -f ./linux-sample-1.bin \
    linux.pslist
Volatility 3 Framework 2.11.0
OFFSET (V)      PID     TID     PPID    COMM    CREATION TIME   File output

0x88001f994740  1       1       0       init    2014-06-24 10:22:33.016001 UTC  Disabled
0x88001f994040  2       2       0       kthreadd        2014-06-24 10:22:33.016001 UTC  Disabled
0x88001f9a2780  3       3       2       ksoftirqd/0     2014-06-24 10:22:33.056003 UTC  Disabled
0x88001f9a67c0  5       5       2       kworker/u:0     2014-06-24 10:22:33.056003 UTC  Disabled
```

volatility3/framework/plugins/timeliner.py
```diff
@@ -23,6 +23,7 @@ class TimeLinerType(enum.IntEnum):
     MODIFIED = 2
     ACCESSED = 3
     CHANGED = 4
+    BOOTTIME = 5
```
ikelos: I'm slightly concerned about is just adding a new category of time.  This is presumably based on autopsy output.  Is there a standard for this format, or is it just a first-implementer-decides type thing?

ikelos: I recall a whole bunch of stuff getting shoehorned into different timestamps (I imagine this one would fall under created/born)?  What make boottime special over other system events that might happen?  Why isn't this a more generic "system event" or even just "event"?  Boottime seems like it'll have one use in amongst 1000+ records...

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -830,3 +834,97 @@ def get_cached_pages(self) -> interfaces.objects.ObjectInterface:
...
             if page:
                 yield page
+
+
+@dataclasses.dataclass
+class TimespecVol3(object):
+    """Internal helper class to handle all required timespec operations, convertions and
+    adjustments.
+
+    NOTE: This is intended for exclusive use with get_boottime() and its related functions.
```
ikelos: Why... then... isn't it in a boottime specific class?  None of these methods seem very specific to `boottime`, they seem specific to timespec64, which is a linux specific extension.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -830,3 +834,97 @@ def get_cached_pages(self) -> interfaces.objects.ObjectInterface:
             page = self.vmlinux.object("page", offset=page_addr, absolute=True)
             if page:
                 yield page
+
+
+@dataclasses.dataclass
+class TimespecVol3(object):
```
ikelos: Again, not sure what means it needs Vol3 in the title of the class?  Seems oddly specific and out of character for any other classes have no.  Not an overly descriptive class title if someone just saw it somewhere.  How/why is this class different to a standard datetime object, or a `timespec64` or any of a myriad other time holding data structures?

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -830,3 +834,97 @@ def get_cached_pages(self) -> interfaces.objects.ObjectInterface:
...
+
+        result.normalize()
+
+        return result
+
+    def __sub__(self, timespec) -> "TimespecVol3":
+        """Returns a new TimespecVol3 object that subtracts the values in the timespec
+        argument from the current object's values"""
+        if not isinstance(timespec, (TimespecVol3, extensions.timespec64)):
+            raise TypeError("Cannot add a TimespecVol3 to this object")
```
ikelos: subtract, not add in this method.

volatility3/framework/symbols/linux/__init__.py
ikelos: Why aren't all these helper methods of a `timespec64` class override?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1882,7 +2078,7 @@ def get_capabilities(self) -> int:
 
 
 class timespec64(objects.StructType):
-    def to_datetime(self) -> datetime:
+    def to_datetime(self) -> datetime.datetime:
```
ikelos, line 2231: Thanks, these changes fit much better with the rest of the code base.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -830,3 +834,97 @@ def get_cached_pages(self) -> interfaces.objects.ObjectInterface:
...
+
+        result = TimespecVol3(
```
ikelos: Feels odd to have a negate function and then a whole (mostly duplicate) function for subtraction and addition.  I get that negate would change the underlying object, but we wouldn't it return a new one with the values negated?  It would allow sub to just be negate/add and it would

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -830,3 +834,97 @@ def get_cached_pages(self) -> interfaces.objects.ObjectInterface:
...
+        """Converts this TimespecVol3 to a UTC aware datetime"""
+        return conversion.unixtime_to_datetime(
+            self.tv_sec + self.tv_nsec / NSEC_PER_SEC
+        )
+
+    def to_timedelta(self) -> datetime.timedelta:
+        """Converts this TimespecVol3 to timedelta"""
+        return datetime.timedelta(seconds=self.tv_sec + self.tv_nsec / NSEC_PER_SEC)
+
+    def __add__(self, timespec) -> "TimespecVol3":
```
ikelos: This should probably be called `other` to stick with standard conventions, and so it's clear the ordering when we come to sub...

volatility3/framework/plugins/timeliner.py
```diff
@@ -171,18 +172,22 @@ def _generator(
...
+                                ),
                             ],
                         )
                     )
 
                     # Write each entry because the body file doesn't need to be sorted
                     if fp:
                         times = self.timeline[(plugin_name, item)]
-                        # Body format is: MD5|name|inode|mode_as_string|UID|GID|size|atime|mtime|ctime|crtime
+                        # Body format is: MD5|name|inode|mode_as_string|UID|GID|size|atime|mtime|ctime|crtime|boottime
```
ikelos: Other fields are already defined here for the body format?  Doesn't seem like diverging from the standard is a good idea...  5:S
https://wiki.sleuthkit.org/index.php?title=Body_file

ikelos (commented):
Generally the plugin seems fine, we need to get the tests sorted as soon as possible.  That's one for @npetroni .  My worry with this is the deviation from a [well known standard](https://wiki.sleuthkit.org/index.php?title=Body_file) for seemingly little gain and the naming of one of the classes feels quite vague/rushed.  Otherwise generally happy with it, like the comments detailing exactly which case is for which kernel version (down to the git commit in some places, I'm assuming?).

volatility3/framework/plugins/timeliner.py
gcmoreira: Yes, I missed that Timeliner is solely for generating a Sleuthkit body file. I'll update it to match the CREATED category.

volatility3/framework/symbols/linux/__init__.py
gcmoreira: I've been trying to create a dummy object within the framework but ran into issues. I attempted to replicate what's done in [object()](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/contexts/__init__.py#L125) using clone, ObjectInformation, etc., but no luck. 
Could you explain how to create a timespace64 object that allows updating values without being attached to an offset/address/layer? If that's possible, I could relocate this code within the timespace64 class; Otherwise, it seems it will have to remain a separate class, unfortunately.
Keep in mind that it doesn't always start with a kernel `timespec64`; the kernel (my code mimics that) construct timespec64 objects from other values.

volatility3/framework/symbols/linux/__init__.py
gcmoreira: Hm, sorry, I'm not quite following what you mean by "this" here. Are you referring to the method argument named 'timespec'?

volatility3/framework/symbols/linux/__init__.py
gcmoreira: It depends on the other comment about including all this code in the `timespec64` object. Currently, the argument can be either a `TimespecVol3` or a `timepec64`  object, so if the argument is a `timepec64`,  `self + timespec.negate()` won't work.

volatility3/framework/plugins/timeliner.py
gcmoreira: done

volatility3/framework/symbols/linux/__init__.py
gcmoreira: I'd like to hear your thoughts on creating a timespec64 dummy object. For now, I prefer to keep this 'private', for the same reasons you've mentioned in your other comments. I wasn't particularly keen on creating a class that feels 'outside' the framework, but I couldn't find another way to resolve it

volatility3/framework/symbols/linux/__init__.py
gcmoreira: done

volatility3/framework/plugins/timeliner.py
gcmoreira: done

volatility3/framework/symbols/linux/__init__.py
ikelos: That's ok.  It feels like you could create a class override that implents a `timespec64` interface, which just contains the necessary data as attributes.  For volatility created objects, those would be derived from the layer/offset and data read from the layer.  For others it could just be the storage of the two integers.  So the interface would contain the code that operated just one the two values, the class override would inherit from the interface, and then a stub class that instantiates the interface could be used to make arbitrary ones?  So:

```
Timespec64Interface - all the methods you actually want to work with
Timespec64(Timespec64Interface, Structure) - Class override applied to Timespec64 ISF data
Timespec64Concrete(Timespec64Interface) - Concrete implementation instantiated by the two variables, rather than a structure)
```

The name would need some thinking about, perhaps it should be Timespec64Abstract because it's just the concept of a timespec64, but the idea's the same.  Something you can use to make local timespec64-ish things.  Since both would adhere to the interface, the interface is where all the methods would go.

volatility3/framework/symbols/linux/__init__.py
ikelos: Combining them into a single interface means you wouldn't have to do the differentiation here (just check whether it's an instance of the interface or not).

volatility3/framework/symbols/linux/__init__.py
ikelos: `__add__(self, timespec)` should be `__add__(self, other)`.  It's not a big deal, or a showstopper, just in keeping with the usual `__` function signatures: https://docs.python.org/3/reference/datamodel.html#object.__add__

gcmoreira: done

volatility3/framework/symbols/linux/__init__.py
gcmoreira: done

volatility3/framework/symbols/linux/__init__.py
gcmoreira: Brilliant, did the trick!. It looks a lot cleaner and simpler now. Hope it's just what you wanted! 
This also includes `__sub__()` implemented as `__add__()` + `negate()` for your enjoyment  ;) .. thanks

ikelos (approved):
Looks good, thanks!  5:)

volatility3/framework/symbols/linux/__init__.py
ikelos: Yep, looks good, thanks for clearing that up!  Happy if you want to change the names, but at least they should be descriptive...

ikelos:
Ok, this seems to be failing on `boottime = Timespec64Concrete.new_from_timespec(wall_to_monotonic)` which expects a Timespec64Abstract, but doesn't get one seemingly.  Happy to merge this once we get the test passing/figure out what's going wrong.

ikelos:
It looks like `wall_to_monotonic`, which is made from n `object_from_symbol` call, isn't returning a `timespec64` (or at least not something that looks like a `Timespec64Abstract`, which is what it needs)

gcmoreira:
@ikelos Alright, merge conflict resolved. It will be good to go once the ISF is fixed.

gcmoreira:
@ikelos the ISF is now fixed, I reran the tests and it passed. it's ready to go!!

volatility3/framework/plugins/linux/pslist.py
```diff
@@ -81,7 +83,7 @@ def filter_func(x):
     @classmethod
     def get_task_fields(
         cls, task: interfaces.objects.ObjectInterface, decorate_comm: bool = False
-    ) -> Tuple[int, int, int, str]:
+    ) -> Tuple[int, int, int, int, str, datetime.datetime]:
```
ikelos, line 86: This change to the return response of a classmethod (the exposed API) should've meant that the MAJOR version got bumped, sorry for missing it...  5:S

PR #1320, opened 2024-10-24
Linux: Extensions - removes abc.Iterable definition

dgmcdona:
`hlist_head` is missing an implementation for `collections.abc.Iterable`, resulting in a crash at runtime. This fixes the issue by providing the required `__iter__` implementation for `hlist_head`.

### DIFF
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -1004,7 +1004,7 @@ class list_head(objects.StructType, collections.abc.Iterable):
         return self.to_list(self.vol.parent.vol.type_name, self.vol.member_name)
 
 
-class hlist_head(objects.StructType, collections.abc.Iterable):
+class hlist_head(objects.StructType):
     def to_list(
         self,
         symbol_type: str,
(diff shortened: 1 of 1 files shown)


gcmoreira:
Hey @dgmcdona it was just removed :) ... good catch anyway but the proposed fix is not correct. It's actually the other way around, we have to remove `collections.abc.Iterable` from the class declaration,  I forgot to do this, my bad. 
`self.vol.member_name` cannot be used as the member name to iterate this type and that's why a` __iter__` doesn't make sense here.

gcmoreira:
By the way, `__iter__` doesn’t make sense for `list_head` either, so it also needs to be removed from that object as well. I'll handle that in a separate PR soon, but it will need more extensive testing since many plugins rely on that object.

dgmcdona:
Sounds good, thanks @gcmoreira ! I'll close this one, then.

gcmoreira:
Oh we can fix this and take advantage of your finding here if you want.  You should remove the hlist_head `__iter__` method  and `collections.abc.Iterable` from the class declaration and it will work again. Sorry for the misunderstanding, I didn't mean you should close it.

gcmoreira:
Great! thanks @dgmcdona, it looks good.

@ikelos this is ready for your review. @dgmcdona caught this bug right away :1st_place_medal: . I messed it up in the last commit.

ikelos (approved):
Happy if you guys are...

PR #1323, opened 2024-10-29
Linux: netfilter plugin: Fix hooked field to match Volatility2 output

gcmoreira:
This PR fixes how the hooked module is interpreted to match the output of its Volatility2 counterpart.

### DIFF
--- a/volatility3/framework/plugins/linux/netfilter.py
+++ b/volatility3/framework/plugins/linux/netfilter.py
@@ -675,7 +675,7 @@ class Netfilter(interfaces.plugins.PluginInterface):
 
     _required_framework_version = (2, 0, 0)
 
-    _version = (1, 0, 0)
+    _version = (1, 1, 0)
 
     _required_linuxutils_version = (2, 1, 0)
     _required_lsmod_version = (2, 0, 0)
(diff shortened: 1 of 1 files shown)


ikelos (commented):
Happy to make the change, but I'd like it to be right rather than just matching something done previously.  The docstring also doesn't really shed any light on what hooked is supposed to mean.  Lastly, it's converted to a string rather than being kept as a boolean, is there any reason for that?  I'm happy to push this through as is, but it feels like it'll need tidying up at some point or another, so if you'd like to do it now I can hold off until you're happy...

ikelos:
So... for clarity, this is waiting on the docstring being improved (from just `hooked?`) and for someone to tell me that the column title is unambiguous and that flipping this return value won't confuse anyone in any way.

atcuno:
Yes it won't be confusing. Hooked is saying the network stack has been hijacked by a non-legitimate source. The logic was just inversed previously for determining the boolean value.

ikelos:
Would `hijacked` be clearer or is hooked the more accurate technical term?  If you're happy no one will get confused then I'm happy to merge it...

ikelos:
Cool, looks good thanks for the fix!  5:)

PR #1326, opened 2024-10-30
Linux: PageCache API: Fix typing

gcmoreira:
This PR fixes the typing return type for several methods in the Page Cache API.


ikelos (approved):
Looks good, thanks.

PR #1327, opened 2024-10-30
Linux: vfsmount type: Fix type comparison

gcmoreira:
This PR fixes an incorrect type comparison in the vfsmount object extension, which i.e. caused Flake to flag it as an error.


ikelos (approved):
Looks good, sorry the initial review didn't catch it...

PR #1328, opened 2024-10-30
Linux refactory module extension object mod_mem_type cache

gcmoreira:
From Python 3.8, the built-in `functools` module provides a `cached_property` decorator. This allow us to simplify the custom made property cache in `module::mod_mem_type`. This PR replaces that code with the decorator.

_Note that both this implementation and the previous one perform caching on a per-instance basis. While this may help in some scenarios, it’s not the most efficient approach; for instance, the `lsmod plugin` repeatedly resolves the enumerator for each module individually, making the cache ineffective. 
A more effective solution would be to adopt a similar strategy to the previous implementation but implement caching at the class level using a class variable._ 

@ikelos Do you think caching per instance is a reasonable trade-off, or would you prefer to enhance efficiency by caching the enumerator at the class level?

### DIFF
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -32,7 +32,7 @@ vollog = logging.getLogger(__name__)
 class module(generic.GenericIntelProcess):
 
     @functools.cached_property
-    def mod_mem_type(self):
+    def mod_mem_type(self) -> Dict:
         """Return the mod_mem_type enum choices if available or an empty dict if not"""
         # mod_mem_type and module_memory were added in kernel 6.4 which replaces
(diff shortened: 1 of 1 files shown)


ikelos (approved):
I think there's a few other places where we have memoization of a property.  I don't think the value ever changes over the lifetime of the plugin, so this should be fine.

ikelos:
Errr, class level things make me a little cautious.  If the value will be the same for one instance of the kernel, I'd almost sooner find a way of caching it there somehow.  So it's calculated once and then stored as part of the state of the system (which the kernel module somewhat represents).  I'm not sure how feasible that would be to do.  I think the hit for now is ok, and I'd it's raised as a problem, then we can look into how to improve it.

Issue #1286, opened 2024-10-02
Add detection of malicious ftrace and tracepoints


atcuno:
@Abyss-W4tcher we need to get ftrace and tracepoints into vol3 for the parity release.

I know you have these here:

https://github.com/Abyss-W4tcher/volatility-scripts/blob/master/Volatility_contest_2023/plugins/check_ftrace.py

https://github.com/Abyss-W4tcher/volatility-scripts/blob/master/Volatility_contest_2023/plugins/check_tracepoints.py

Are you up for converting these over? Myself and Gus can work on it if not.

Abyss-W4tcher:
Hi @atcuno, sure I can port these two plugins :) 

Would you like them to be put in a common `kernel_tracing` directory under `plugins`, so that users can easily understand the context of these plugins :

- `linux.kernel_tracing.ftrace`
- `linux.kernel_tracing.tracepoints`
- [..] any additional technique

atcuno:
yes that would be nice as eventually there will be 5+

Abyss-W4tcher:
Plugins are ready, but actually depend on `hidden_modules`, and an additional plugin I developed named `modxview` (which is basically psxview but for modules). So, a few parts need to move before ftrace and tracepoints get to a PR :)

gcmoreira:
Cool, I can test these plugins if you need it. Thanks

atcuno:
@Abyss-W4tcher @gcmoreira what is the latest on this one?

Abyss-W4tcher:
I am still waiting on modxview plugin review, which embeds APIs for ftrace and tracepoints.

PR #1330, opened 2024-10-31
New linux plugin: modxview

Abyss-W4tcher:
Hello,

This new plugin centralizes the module detection capabilities, and allows to quickly spot and list all the modules. This way, it is easier to detect copycat malicious modules, trying to mimic kernel modules names (ex: https://attack.mitre.org/techniques/T1036/006/). To keep the name familiar for long-term users, it follows the `psxview` naming convention. 

In addition, this PR introduces the capability to parse taint flags, which embeds additional data to help debug a kernel or a module (as it is in the end, quite similar). 

Here are some sample (stripped) outputs : 

"/proc/modules" hidden module :
```bash
$ python3 vol.py -f sample1.bin linux.modxview
  |                Name |        Address | In /proc/modules | In /sys/module/ | Hidden |                      Taints
* |                lime | 0xffffc0bd2040 |             True |            True |  False | OOT_MODULE,UNSIGNED_MODULE
* |       rootkit_ports | 0xffffc0a9b180 |             True |            True |  False | OOT_MODULE,UNSIGNED_MODULE
* |        rootkit_char | 0xffffc0a8f2c0 |             True |            True |  False | OOT_MODULE,UNSIGNED_MODULE
* |              vboxsf | 0xffffc0bf8800 |             True |            True |  False | OOT_MODULE,UNSIGNED_MODULE
* |      rootkit_sysfsh | 0xffffc0aa0180 |            False |            True |  False | OOT_MODULE,UNSIGNED_MODULE
```

"/proc/modules" hidden module (plain taints string) :
```bash
$ python3 vol.py -f sample1.bin linux.modxview --plain-taints
  |                Name |        Address | In /proc/modules | In /sys/module/ | Hidden | Taints
* |                lime | 0xffffc0bd2040 |             True |            True |  False |   GOE
* |       rootkit_ports | 0xffffc0a9b180 |             True |            True |  False |   GOE
* |        rootkit_char | 0xffffc0a8f2c0 |             True |            True |  False |   GOE
* |              vboxsf | 0xffffc0bf8800 |             True |            True |  False |   GOE
* |      rootkit_sysfsh | 0xffffc0aa0180 |            False |            True |  False |   GOE
```

Hidden module :
```bash
$ python3 vol.py -f sample2.bin linux.modxview
  |                Name |        Address | In /proc/modules | In /sys/module/ | Hidden |                      Taints
* |                 drm | 0xffffc03809c0 |             True |            True |  False |                          -
* |               kovid | 0xffffc09ed4c0 |            False |           False |   True |            UNSIGNED_MODULE
```

*Note: kovid [cleans out the taints attributes](https://github.com/carloslack/KoviD/blob/master/src/sys.c#L1079), but the UNSIGNED_MODULE one will be there nonetheless. This is still a great piece of information to spot LKM modules.*

Happy to read your reviews about this !


volatility3/framework/constants/linux/__init__.py
```diff
@@ -347,3 +347,88 @@ def flags(self) -> str:
 MODULE_MAXIMUM_CORE_SIZE = 20000000
 MODULE_MAXIMUM_CORE_TEXT_SIZE = 20000000
 MODULE_MINIMUM_SIZE = 4096
+
+
+TAINT_FLAGS = {
```
gcmoreira, line 366: it would be better a dictionary of [dataclasses](https://docs.python.org/3/library/dataclasses.html)

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -0,0 +1,195 @@
...
+
+        Returns:
+            Iterator of modules objects
+        """
+        seen_addresses = set()
+        for modules in run_results.values():
+            for module in modules:
+                if deduplicate and module.vol.offset in seen_addresses:
+                    continue
+                yield module
```
gcmoreira, line 76: you forgot the 
```python
seen_addresses.add(module.vol.offset)
```

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -0,0 +1,195 @@
...
+                        taints or NotAvailableValue(),
+                    ),
+                )
+
+    def run(self):
+        columns = [
+            ("Name", str),
+            ("Address", format_hints.Hex),
+            ("In /proc/modules", bool),
+            ("In /sys/module/", bool),
```
gcmoreira: It should work fine, but IMO it's better to avoid using "/" in the column name to make filtering by column easier later on. Maybe `procfs` and `sysfs` or similar?

volatility3/framework/constants/linux/__init__.py
Abyss-W4tcher: Done !

volatility3/framework/plugins/linux/modxview.py
Abyss-W4tcher: Fixed, thanks for the catch :)

volatility3/framework/plugins/linux/modxview.py
Abyss-W4tcher: Yep, that's true. procfs and sysfs seems fine to me, and shorter.

volatility3/framework/constants/linux/__init__.py
gcmoreira: It looks much cleaner and more polished now.

gcmoreira:
@Abyss-W4tcher it looks good to me. 
There are some potential improvements that could be included in this or a future PR, but they would impact the current design.
- It would be amazing to also have a `Linux Kernel Tainted flags` plugin that checks the ([tainted_mask](https://elixir.bootlin.com/linux/v6.11.6/source/kernel/panic.c#L56)) flags when `TaintFlags.module == False`. That's what you get when read `/proc/sys/kernel/tainted` and it's implemented [here](https://elixir.bootlin.com/linux/v6.11.6/source/kernel/panic.c#L520). This indicates that the tainted flags parsing code should be relocated to a shared/common location. How about consolidating to a single function, i.e. `parse_tainted_mask(tainted_mask, is_module)`, that can be used for both the module tainted mask and the kernel tainted mask?
- It would be ideal if the module tainted flags were reported consistently across the three other plugins you are using here, and not just in this one.

Abyss-W4tcher:
> @Abyss-W4tcher it looks good to me. There are some potential improvements that could be included in this or a future PR, but they would impact the current design.
> 
> * It would be amazing to also have a `Linux Kernel Tainted flags` plugin that checks the ([tainted_mask](https://elixir.bootlin.com/linux/v6.11.6/source/kernel/panic.c#L56)) flags when `TaintFlags.module == False`. That's what you get when read `/proc/sys/kernel/tainted` and it's implemented [here](https://elixir.bootlin.com/linux/v6.11.6/source/kernel/panic.c#L520). This indicates that the tainted flags parsing code should be relocated to a shared/common location. How about consolidating to a single function, i.e. `parse_tainted_mask(tainted_mask, is_module)`, that can be used for both the module tainted mask and the kernel tainted mask?
> * It would be ideal if the module tainted flags were reported consistently across the three other plugins you are using here, and not just in this one.

Alright, I thought about a `linux.kernel_taints` plugin, but wasn't sure about the relevance of such a small plugin, thinking it would be better integrated in a `windows.info` equivalent for Linux ? However, I'm definitely ok for parsing the tainted_mask. 

Anyway, the code can indeed be shared somewhere (maybe LinuxUtilities), while keeping the convenient and self-contained APIs inside the `extensions.module` as a wrapper for them ?

Putting the taint flags in other plugins outputs should be en easy task, maybe in a 3-for-1 PR if that's ok for ikelos.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -307,6 +378,17 @@ def section_strtab(self):
             return self.strtab
         raise AttributeError("module -> strtab: Unable to get strtab")
 
+    @property
+    def taints_value(self) -> int:
+        return self.taints
```
gcmoreira: Not sure about having a property like this if the member was consistently named taints. Is there a particular reason for this?

Abyss-W4tcher: Mostly type hinting, which allows to clearly differentiates properties, to separate between the taints value/taints string/taints parsed.

gcmoreira:
> Alright, I thought about a `linux.kernel_taints` plugin, but wasn't sure about the relevance of such a small plugin, thinking it would be better integrated in a `windows.info` equivalent for Linux ? However, I'm definitely ok for parsing the tainted_mask.

Sure, this is ultimately up to @ikelos, but I personally lean toward the Unix philosophy: small, specialized tools, each dedicated to a single purpose. This approach doesn't mean each plugin must detect malicious behavior on its own. Then, we can have let's say 'macro' plugins, like the one in this PR, that aggregates various indicators to help identify anomalies. This way, the small plugins can be reused in other "macro" plugins.

The kernel tainted flags plugin code will be minimal since we will reuse the code you have already written for modules, but the insights it could provide would be highly valuable and unique. For instance:
- You can use the kernel tainted flags plugin output to complement the module tainted flags in lsmod/hidden_modules/check_modules/modxview plugins by verifying that the global kernel tainted flag aligns with the individual module tainted flags. I mean, if the global flag says there's a out-of-tree module loaded it should be a module with out-of-tree flag set. Here, it could be that it was unloaded too, but that's also suspicious. BTW, the kernel also keeps track of the last unloaded module and the unload tainted modules, that will also be super interesting ;) see [last_unloaded_module](https://elixir.bootlin.com/linux/v6.11.6/source/kernel/module/main.c#L537), [unloaded_tainted_modules](https://elixir.bootlin.com/linux/v6.11.6/source/kernel/module/tracking.c#L17) and [print_unloaded_tainted_modules](https://elixir.bootlin.com/linux/v6.11.6/source/kernel/module/tracking.c#L49)  .. plugin number 9999999? .. hehe .. sorry :) 
- Soft lockups or bad pages could indicate an attempted or actual kernel compromise. Or, of course, just a kernel bug i.e.: https://feedly.com/cve/CVE-2022-48937
- ACPI table modifications:
  - https://www.blackhat.com/presentations/bh-europe-06/bh-eu-06-Heasman.pdf
  - https://www.phoronix.com/news/Ubuntu-18.04-Lockdown-Bypass
- Kernel live patched https://docs.kernel.org/livepatch/livepatch.html
- and so on

The output could be just in a single line, or we could opt for a more detailed, explanatory format such as:
```
Flag  Description
----- -----------
A     desc A
B     desc B
C     desc C
```

gcmoreira:
> Anyway, the code can indeed be shared somewhere (maybe LinuxUtilities)

Yeah, this is one is for @ikelos. IMO if it's just one function it could be in **LinuxUtilities**. Alternatively, if there are more related functions, I think it could be more convenient to have a separated class containing that subsystem API .. like https://github.com/volatilityfoundation/volatility3/pull/1332/files#diff-8456da6d20fc84f0d63dedc5bc816ffde418ff41c83b1828da7c614252bda150R840 with its own version, but let's see if ikelos agrees with this idea, since that PR is still pending review.

On a related note, I think it would be a good idea to reorganize **LinuxUtilities** in the future, grouping functions by subsystems such as mount, modules, etc. and only keep in **LinuxUtilities** the general or framework helpers like `container_of()`, `get_module_from_volobj_type()`, etc. So, if everyone agree with this, it may be a good opportunity to start differentiating components from **LinuxUtilities**. This will also help minimize the impact of modifying a specific subsystem API and reduce the scope of its corresponding testing.

> while keeping the convenient and self-contained APIs inside the `extensions.module` as a wrapper for them ?

Correct, and call the **LinuxUtilities** or modules API function with `parse_tainted_mask(self.taints_value, is_module=True)`  and from the the Linux kernel tainted plugin `parse_tainted_mask(tainted_mask_value, is_module=False)`.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: I double-checked this and **taints** was added to struct module in 2.6.19 here -> https://github.com/torvalds/linux/commit/2bc2d61a9638dab670d8361e928d1a5a291173ef .. so we are good to go. 
Would you mind adding a comment with this information please?
```python
# kernels >=2.6.19 2bc2d61a9638dab670d8361e928d1a5a291173ef
```

Abyss-W4tcher:
Alright, let's wait for ikelos to give its point of view on all these comments, then I'll unify the API depending on where we want to put it. 

The `kernel_taints` plugin might end up calling `modxview` and then check for a LKM flag etc., so it might be worth opening a dedicated issue to centralize what this plugin should check and get a clear picture of what it leverages and gives insights about. 

Globally, it sounds really good to me :)

gcmoreira:
Hey @ikelos any chance you could take a look at this PR? We need it sorted before moving on to https://github.com/volatilityfoundation/volatility3/issues/1286. Thanks

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -279,6 +279,77 @@ def get_symbol_by_address(self, wanted_sym_address):
...
+            c_true = chr(taint_flag.c_true)
+            c_false = chr(taint_flag.c_false)
+            if taint_flag.module and (self.taints_value & (1 << i)):
+                taints_string += c_true
+            elif taint_flag.module and c_false != " ":
+                taints_string += c_false
+
+        return taints_string
+
+    def get_taints_as_plain_string(self) -> str:
```
ikelos: These are externally visible additions to the API, which means a MINOR version number somewhere, needs to go up.  I suspect it may be the framework itself because I don't think anything further down is versioned?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -279,6 +279,77 @@ def get_symbol_by_address(self, wanted_sym_address):
...
+        """Convert the module's taints string to a 1-1 descriptor mapping.
+
+        Returns:
+            A comprehensive (user-friendly) taint descriptor list.
+
+        Documentation:
+            - module_flags_taint kernel function
+        """
+        comprehensive_taints = []
+        for c in self.get_taints_as_plain_string():
```
ikelos: Please don't use single letters in loops.  You only have to type it once, but `character` will just make it a little easier for readers of the code for years to come...

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: I'm not sure polluting the namespace to get better typing information is useful?  It now offers people two options for the samething.  Either `taints` shouldn't be public (`_taints`) or this property shouldn't exist (untill it does something different than returning just a plain member).

ikelos (commented):
Generally looks really nice, next to no comments (which is great work, you're making it harder and harder for me to nit-pick!)  5:P  Sadly, the ugly need for version numbers raises its head, so just get that bumped appropriately and sort out the other couple of bits and it should be good to go...  5:)

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: Alright, if @gcmoreira and you think this is overkill, we can resort to the plain struct attribute 👍

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: As discussed, we want to include this API in a more generic place, but keep these convenient self-contained functions. So right now I'm not sure of how many bumps are needed.

Abyss-W4tcher:
Hi, this is currently stale, we are waiting for a review regarding the LinuxUtilities stuff. Do we try moving and re-thinking this bit now, or do I just unify the taints function there as discussed ?

gcmoreira:
> Hi, this is currently stale, we are waiting for a review regarding the LinuxUtilities stuff. Do we try moving and re-thinking this bit now, or do I just unify the taints function there as discussed ?

@ikelos: In LinuxUtilities or as a separate/versioned class?

I think using a separate class is a better approach, as adding more functionality to LinuxUtilities can lead to issues. Any version change in LinuxUtilities would require version updates in all dependent plugins and APIs. To improve maintainability, we should move to versioned classes dedicated to each subsystem or API. For reference, see the approach taken in [this PR](https://github.com/volatilityfoundation/volatility3/pull/1332/files#diff-8456da6d20fc84f0d63dedc5bc816ffde418ff41c83b1828da7c614252bda150R840).

ikelos:
That sounds like a reasonable plan @gcmoreira .  @Abyss-W4tcher, would you be able to pull the additions out of LinuxUtilities and put them into a separate object in the same vein as the VMCore example Gus provided please?

ikelos:
@gcmoreira On the more general note, I'm happy to have the submodules split out into versioning subcomponents, but we need to make it slick and fast for people to both check that the submodule *exists* (so we can add news ones) and that it's the right version.  I figure some kind of a check function in LinuxUtilities itself, but then how to do the requirement dependencies gets harder.  I suppose we could do some kind of magic where if a module *doesn't* exist, it instead imports a dummy class successfully but whose version will always fail all version requirements?  Starting to sound a little hacky, but the other option would be a specific `LinuxUtilityRequirement` where you specify the name of the module and the version, and then it does the import attempt and fails.  Except then how does the plugin get the imported module if it exists?  5:S  We can't populate the config with it because it's not a simple value and I don't want to go wading into the config system to butcher something in place.  Otherwise there's no way for the config system to report back success or failure of a requirement (and we'd have to start littering the config system with the requirements).  So you'd need a `LinuxUtilities` lookup function *and* a `LinuxUtilitiesRequirement` which would allow the requirement to fail if it couldn't be fulfilled and otherwise the lookup to return the module?  I think that could work?  We'd need to leave the old ones in place to avoid having to change *every* plugin.  But they could probably be turned into shims that just do the module check themselves and then proxy the call?  Big PR, but I'm up for reviewing it if someone wants to write it...  5;)

ikelos:
So just to recap, that'd be:

```
LinuxUtilities:
    get_module('module_name') -> Module:
    // Probably this is a generic wrapper that we can parameterize so just old_function = smart_wrapper('module', 'function_name')
    old_function = smart_wrapper('module', 'function_name')
    smart_wrapper(args):
        modobj = get_module(module)
        return getattr(module, function_name)(args)
    
LinuxUtilitiesModule(Versionable):
    _version
    <previous functions>

LinuxUtilitiesRequirement // with params (module, version)
```

Abyss-W4tcher:
I made the requested changes, which lays the ground work for further tainting parsing capabilities as well:

```python
(layer_name) >>> from volatility3.framework.symbols.linux import Tainting
(layer_name) >>> kernel = context.modules[self.kernel.name]
(layer_name) >>> tainted_mask = kernel.object_from_symbol("tainted_mask")
(layer_name) >>> 
(layer_name) >>> Tainting(self.context, kernel.name).get_taints_parsed(tainted_mask)
['OOT_MODULE', 'UNSIGNED_MODULE']
```

@gcmoreira the `is_module` flag seems to only strictly apply to modules, in the sense that `tainted_mask` parsing does not account for this flag: https://elixir.bootlin.com/linux/v6.5/source/kernel/panic.c#L492 vs https://elixir.bootlin.com/linux/v6.5/source/kernel/module/main.c#L879. 

Happy to enhance those capabilities in other plugins once this gets merged, as discussed I might open a dedicated issue to centralize the area of research and the benefits investigators could get from it.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -376,13 +329,6 @@ def section_strtab(self):
             return self.strtab
         raise AttributeError("Unable to get strtab")
 
-    @property
```
ikelos: This property needs to be exposed to avoid a bump too....

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -279,76 +279,29 @@ def get_symbol_by_address(self, wanted_sym_address):
...
             The raw taints string.
-
-        Documentation:
-            - module_flags_taint kernel function
         """
-
-        if self.taint_flags_list:
-            return self._module_flags_taints_post_4_10_rc1()
-        return self._module_flags_taints_pre_4_10_rc1()
+        return linux.Tainting(
```
ikelos: This isn't quite what I had in mind, because it requires `LinuxUtilities` to import each of the supported modules, and doesn't allow for plugins to determine whether their requested function will exist or not.  I'll have a go at mocking something up tomorrow that's more what I had in mind.  I think the separate class stuff is all good, we just need to tweak how we find the function we want to run...

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -830,3 +831,123 @@ def get_cached_pages(self) -> Iterator[interfaces.objects.ObjectInterface]:
             page = self.vmlinux.object("page", offset=page_addr, absolute=True)
             if page:
                 yield page
+
+
+class Tainting:
```
ikelos: This might want to go in a `tainting.py` module, so it would become `linux.tainting.Tainting` (and so that the init doesn't get too full).

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -830,3 +831,123 @@ def get_cached_pages(self) -> Iterator[interfaces.objects.ObjectInterface]:
...
+    """Tainted kernel and modules parsing capabilities.
+
+    Relevant kernel functions:
+        - modules: module_flags_taint
+        - kernel: print_tainted
+    """
+
+    def __init__(
```
ikelos: I'd also expected these just to be function containers, rather than carrying state themselves.  We can have them carry state I guess, I'm just not sure the extra complexity/separation of parameters is worth it?

volatility3/framework/symbols/linux/__init__.py
ikelos: This also would need its own version number, and a required_framework_version for anything it made use of (and the `__init__` would need to check those...

ikelos (commented):
Thanks for the changes you made.  It's close to what I was suggesting, but I'll try and come up with a more formed example tomorrow for you to go off.  This is close but it's not quite how I envisaged it...

Abyss-W4tcher:
Alright, no problem. I took reference from https://github.com/volatilityfoundation/volatility3/pull/1332/files#diff-8456da6d20fc84f0d63dedc5bc816ffde418ff41c83b1828da7c614252bda150R840 as indicated, but it might not have been exactly what you had in mind for this topic.

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: This property was removed from the `module` class, not sure if you meant to write the comment here ?

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: Ok, so we will have to tweak the `module` class, to include versioned requirements of the new `Tainting` object ?

https://github.com/volatilityfoundation/volatility3/blob/e7e4d76d950601258a84057779b9dd1fdff1b732/volatility3/framework/symbols/linux/extensions/__init__.py#L29

ikelos:
What I'd been thinking of was a little more along these lines?

https://github.com/volatilityfoundation/volatility3/tree/feature/linux-utilities-split

It allows for all the old code to keep working, but new code should be to import the separate modules.  If a new plugin is used with an old framework, the import will fail and the plugin won't load (kinda gracefully), if an old plugin is used with a new framework, it should raise a deprecation warning and eventually we can bump LinuxUtilities MAJOR version and drop all the proxied functions.  That will probably have to happen after a MAJOR bump to any of the modules (otherwise the proxied calls would change without anyway of telling from the deprecated perspective).  Slowly we should be able to empty it down to just general linux functions and have all the others off using separate modules.  Hopefully that all makes sense, naming of the imported modules may require some work (because we typically only keep the module name, not several modules below, so linux and utilities will get lost (hence my `as linux_paths` in the example).  I'm not intended to push the whole example, probably just the deprecation capability, just I wanted to show how I was envisaging it all unfolding....

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Removing it is a non-additive API change.  That's why I said it needs to still be exposed there for anything that had been using it, otherwise a caller is going to call it expecting it to work and they'll get an ugly `AttributeError`.  The point of the version numbers is to avoid situations like that, so happy for you to bump the version number, or put the property back, but you need to do one of them...

ikelos: Ugh, I see, `module` isn't versioned at all!  5:S  That gets trickier...  5:\

ikelos: It still needs to be exposed then, otherwise any code calling `module_instance.taint_flags_list` is going to throw an error.  This is why I like people to be careful about what they throw into classes like this.  Changing them in the future is tricky...  5:P

Abyss-W4tcher: This property was added in this PR :), it was not present before. Changes might be confusing as we are moving things.

ikelos: If it was only to be used internally, it needed to be called `_taint_flags_list`...

Abyss-W4tcher: That is a good point, however there is no need for bumping as it hasn't made it into the framework yet. I'll make sure to make it "private" though.

ikelos: Thanks, sorry, I know it's tricky to get right (and I'm getting lost looking at commit diffs *inside* the PR, so I thought it had already gone in).  Yeah, anything that doesn't need to be accessed from the outside would should be marked with a leading `_`, otherwise it becomes part of the API and you've seen the mess of versioning that comes from that...  5;)

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Errr, there isn't really a way of versioning `module`, so it reverts to the version of the framework.  If it's purely an additive change (which it looks like it is) then we can just bump the MINOR version (and nothing else needs to change), but if we ever want to change it in the future, then the MAJOR version of the whole framework needs to change.  That's  why I'm so protective of things like this (and splitting them out into distinct versioned modules is good).  I think I thought these were all applying to LinuxUtilities.  I hadn't realized they were all being tacked onto `module`.  I haven't thought about how we version that separately yet.  The problem with every thing you version is that it then needs checking before use (which is painful in itself).  The requirements let us do that slightly more gracefully and finer grained, but we need to be really careful about what we add to major components of the framework and overriding the linux `module` class is one (at least at the moment)...  5:S

volatility3/framework/symbols/linux/__init__.py
ikelos: Yep, see my branch for an example.

volatility3/framework/symbols/linux/__init__.py
Abyss-W4tcher: With the separated utilities inheriting from the new `LinuxUtilityInterface`,  `get_module_from_volobj_type` won't be able to trace back a `module` object. To fix it, it requires the caller to provide `context, kernel_module_name` to each exposed `classmethod` API, which themselves forward them to internal APIs. I find it a bit far-fetched, adding a lot of "non-functional" parameters everywhere.

The only drawback I see with an `__init__`, is the need for a `vmlinux` object at class instantiation, but this can be circumvented by setting non-kernel dependent APIs to `classmethod`, allowing devs to operate on non-kernel contexts `Tainting.non_kernel_req_api(params)` vs `Tainting(context, kernel_module_name).kernel_req_api(params)`.

Abyss-W4tcher:
Alright, I have checked out your work and re-arranged things. 

One small problem, is that `modxview` will require a specific version of ` volatility3.framework.symbols.linux.utilities.tainting.Tainting` even though it does not directly call it: 

modxview ->`volatility3.framework.symbols.linux.extensions.get_taints_as_plain_string()` -> `volatility3.framework.symbols.linux.utilities.tainting.Tainting.get_taints_as_plain_string()` 

This is a "blind" requirement, but as discussed there is no way currently to require a versioned `module`.

volatility3/framework/symbols/linux/utilities/tainting.py
```diff
@@ -0,0 +1,133 @@
...
+
+        Args:
+            taints: The taints value, represented by an integer
+            is_module: Indicates if the taints value is associated with a built-in/LKM module
+
+        Returns:
+            The raw taints string.
+        """
+        taints_string = ""
+        for i, taint_flag in enumerate(self._kernel_taint_flags_list):
```
gcmoreira: Using a more descriptive name like `taint_bit` instead of `i` would make the code clearer.

volatility3/framework/symbols/linux/utilities/tainting.py
```diff
@@ -0,0 +1,133 @@
...
+            if is_module and is_module != taint_flag.module:
```
gcmoreira: minor but this should be:
```python
if is_module and not taint_flag.module:
```

volatility3/framework/symbols/linux/utilities/tainting.py
```diff
@@ -0,0 +1,133 @@
...
+        Args:
+            taints: The taints value, represented by an integer
+            is_module: Indicates if the taints value is associated with a built-in/LKM module
+
+        Returns:
+            The raw taints string.
+        """
+        taints_string = ""
+        for char, taint_flag in linux_constants.TAINT_FLAGS.items():
+            if is_module and is_module != taint_flag.module:
```
gcmoreira: minor but this should be:
```python
if is_module and not taint_flag.module:
```

volatility3/framework/symbols/linux/utilities/tainting.py
```diff
@@ -0,0 +1,133 @@
...
+
+        return taints_string
+
+    def get_taints_as_plain_string(self, taints: int, is_module: bool = False) -> str:
+        """Convert the taints value to a 1-1 character mapping.
+
+        Args:
+            taints: The taints value, represented by an integer
+            is_module: Indicates if the taints value is associated with a built-in/LKM module
+            s
```
gcmoreira: s? :)

gcmoreira:
@Abyss-W4tcher amazing work! .. just some minor stuff and it will be ready to go

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -0,0 +1,200 @@
...
+
+vollog = logging.getLogger(__name__)
+
+
+class Modxview(interfaces.plugins.PluginInterface):
+    """Centralize lsmod, check_modules and hidden_modules results to efficiently
+    spot modules presence and taints."""
+
+    _version = (1, 0, 0)
+    _required_framework_version = (2, 14, 0)
```
ikelos: This needs updating to 16, sorry, I could only fix the merge  conflict easily.

gcmoreira: Oh, and It would be great if you could also add a test case for this plugin. At least a basic one to ensure the plugin consistently meets its requirements. Since this plugin relies heavily on several others, having tests will be even more critical to catch any potential issues. Thanks

Abyss-W4tcher: Sure, but it would be optimal to add new (infected) samples to the pool (`kovid` one for example), allowing for a malware testcase and not only a functional testcase. This will also be relevant for later `ftrace` and `tracepoints` plugins. 

This might require an additional infrastructure to allow testers to push new samples into a storage/CDN, as done here: 

https://github.com/volatilityfoundation/volatility3/blob/cedc0b074aeda804809b77c8f00c6dbf8565682c/.github/workflows/test.yaml#L30

By the way, what is the current plan to handle storage of new samples ? Downloading them all for each PR action might be time consuming after a certain amount. I've seen a few solutions:  

- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows (10 GB max.)
- git LFS (https://github.com/volatilityfoundation/volatility3-test-data ?)
- self hosted runner (VPS): costly

Abyss-W4tcher:
This should be all set now, apart from the testcases. I'd be happy to provide them after the infrastructure is up, or we can let this PR opened in the meantime (though it might prevent `ftrace` and `tracepoints` plugins to move on) 👍.

volatility3/framework/plugins/linux/modxview.py
ikelos: Yep, that's the intention behind [the test-data repo](https://github.com/volatilityfoundation/volatility3-test-data).

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -0,0 +1,200 @@
...
+            requirements.VersionRequirement(
+                name="linux-tainting", component=tainting.Tainting, version=(1, 0, 0)
+            ),
+            requirements.PluginRequirement(
+                name="lsmod", plugin=lsmod.Lsmod, version=(2, 0, 0)
+            ),
+            requirements.PluginRequirement(
+                name="check_modules",
+                plugin=check_modules.Check_modules,
+                version=(0, 0, 0),
```
ikelos: If you're relying on a version number, it should be at least `(1,0,0)`, `(0,0,0)` is for testing and before the API has stabilized.  If the plugin has a classmethod, it should already have a version of 1.0.0 or greater.  If not, then one should be added in this PR.  (Also, please check for any other plugins that rely on it with too low a version number).

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -0,0 +1,200 @@
...
+                default=False,
+            ),
+        ]
+
+    @classmethod
+    def run_lsmod(
+        cls, context: interfaces.context.ContextInterface, kernel_name: str
+    ) -> List[extensions.module]:
+        """Wrapper for the lsmod plugin."""
+        return list(lsmod.Lsmod.list_modules(context, kernel_name))
```
ikelos: If what's the point of this passthrough?  It feels as though any caller, could instead call list_modules themselves?

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @classmethod
+    def run_check_modules(
```
ikelos: This also doesn't feel like it's adding enough to warrant putting it into a separate classmethod?

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Visible additions like this require a MINOR version bump at the next container level up, so if these live in a separate module then that will have its own version, if it lives in LinuxUtilities, than that needs a bump.  If you add several things at once it only needs one bump.  It's so someone can say "I depend on this thing that I know was added in this version) and everything either works or fails gracefully.

If these are going in a separate module, and they're just passthroughs, there's no point putting them in LinuxUtilities at all, anything new, will know to pull them in from the tainting module.  We only need to support things  that had been in here but are to be moved out (so that old plugins still work the same way).

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -279,6 +279,30 @@ def get_symbol_by_address(self, wanted_sym_address):
...
+        ).get_taints_as_plain_string(self.taints, True)
+
+    def get_taints_parsed(self) -> List[str]:
+        """Convert the module's taints string to a 1-1 descriptor mapping.
+        Convenient wrapper around framework's Tainting capabilities.
+
+        Returns:
+            A comprehensive (user-friendly) taint descriptor list.
+        """
+        return tainting.Tainting(
```
ikelos: Straight pass throughs like this don't need adding, people should just depend directly on `tainting.Tainting`.  We only need them for older methods that used to live here which people may try calling because the plugin hasn't been updated.

volatility3/framework/symbols/linux/utilities/__init__.py
```diff
@@ -0,0 +1,11 @@
+from volatility3 import framework
+from volatility3.framework import interfaces
+
+
+class LinuxUtilityInterface(interfaces.configuration.VersionableInterface):
```
ikelos: I'm not sure this is useful?  It basically a wrapper around VersionableInterface but it provides a default version, which is generally a bad move (we really want to make sure each and every implementation defines its own version).  Hopefully `VersionableInteface` already does a `require_framework_version` check, but even so that probably needs to be replicated, since I've no idea whether the version checking code will be run under inheritance.

ikelos (commented):
Getting much closer, the version numbers all look right, but I noticed we were defining some passthroughs directly on `LinuxUtilities`, which we should only be doing where we're trying to maintain the API.  Adding new methods in that do the same as something else is just giving people two ways of doing one new thing (and that'll lead to confusion down the road).  Also the `LinuxUtilityInterface` is a nice idea, but I think things should just inherit directly from `VersionableInterface` since it's not doing anything new/different...

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: These (new) straight pass through are convenient functions to allow plugins devs to do: 

```python
my_module.get_taints_as_plain_string()
```

instead of 

```python
tainting.Tainting(context,vmlinux.name).get_taints_as_plain_string(my_module.taints, True)
```

It also manages the correct attribute and parameters to pass to fit the module use case. This also applies to volshell `module` inspection use cases. If you still think that it is overkill, please let me know.

Regarding the need for an API bump, even if it is a passthrough, it still is a non-private method addition to the API, no matter its content ?

volatility3/framework/symbols/linux/utilities/__init__.py
Abyss-W4tcher: Sure, I was inspired by https://github.com/volatilityfoundation/volatility3/compare/develop...feature/linux-utilities-split#diff-5da496bf5fddb208451d63f414e6f38a1e6f6f87f9ae14935a2ad06e1070f3db but if it was only for demonstration purpose I will cut this intermediate class.

volatility3/framework/plugins/linux/modxview.py
Abyss-W4tcher: It was to clearly identify which plugins were called, by creating dedications functions, but it is a bit far-fetched indeed.

volatility3/framework/plugins/linux/modxview.py
Abyss-W4tcher: It was to clearly identify which plugins were called, by creating dedications functions, but it is a bit far-fetched indeed.

volatility3/framework/plugins/linux/modxview.py
Abyss-W4tcher: We had this issue in the past, as some plugins aren't versioned at all. I will fix it.

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: For some reason I can't answer the comment about API versioning so I'll do it here. 

The point of these new APIs is not only for modules, but for later plugins. We plan on doing a kernel tainting plugin, so putting these in a sub-LinuxUtilities is not beneficial right now but will be in the future.

API minor bumps (re-counted after merge of develop into this branch): 

- 1 (new) : tainting.Tainting.get_taints_*
- 2 (new) : module.get_taints_*

Which gives : `2.16.0` -> `2.18.0`

ikelos: Ok, sorry, I thought these were on LinuxUtilities, rather than on `module`.

Having said that, `module` is already very heavy (I had to scroll quite a while to see which class they were attached to).  So now what happens if suddenly the method requires an extra parameter?  Typically we leave the old one around but with a deprecation method, and then implement a *new* method that takes the additional parameter.  We also don't really have a way of versioning modules, which means it's either the framework version goes up, or we find a way of versioning the module class, which then everyone has to check before they make a call.  So... as long as you're happy that the api encompasses everything you want to do now *and* will want to do in the future, then it's ok to go in...

`tainting.Tainting` will be version itself, so no need for a framework bump, the module's either there and versioned or it'll throw an import error (which will rule out the plugin that's trying to import it).

That leaves only the bump on module.get_trains_* which is needed so that a plugin being run under the framework fails early rather than trying to make calls to an API that doesn't exist.

volatility3/framework/symbols/linux/utilities/__init__.py
ikelos: Thanks, I like interfaces but that one at least defines some shared methods and has a reason for being, an empty class is just another layer of categorization.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Technically, that proxy call also needs to check that the `Tainting` module is the right version (although it would only prevent errors if the `Tainting` module MAJOR version went up and someone forgot to change this method.  Should be spotted quickly, but the question is, is all that worth the convenience of having it defined on the module.  (Also, why is the tainting object taking parameters for construction, is that so it gets cached?

volatility3/framework/symbols/linux/__init__.py
ikelos: This is only useful if people are going to hang onto the tainting class they've made and my concern is that people will do something strange in the future once these are allowed to carry state.  I don't have a more solid reason than that, but something's setting off my spidey sense about it.  One would then ask why we don't do that with LinuxUtilities, since many of them seem to use the current kernel or at least the symbol table.  It would mean that you couldn't really add a tainting method that *didn't* need a kernel, unless you started mixing classmethods and instance methods.  I dunno, I'm not keen on it, but I can't verbalize why, so I guess it's an irrational fear, but I reserve the right to say I told you so later on...  5;P

Abyss-W4tcher: I try to keep things compact when possible, mixing `classmethod` and `instance` methods looked fine to me, keeping the best of both worlds. I will revert it to a non-state dependent implementation, not a big deal of course .👍

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: It's noted for `module.get_taints_*`, but doesn't adding `Tainting` for the first time technically be considered a new framework API (still making it `2.18.0`), even if obviously it would raise an `ImportError` on failure.

Removing this passthrough might prevent users/devs exploring the `module` API and unfamiliar with tainting to apply it. Also, I think we still have to define what should be the procedure to apply, if someone adds a new versioned LinuxUtility and wants to call it inside a `linux.extensions` class like `inode` or `socket`.

Abyss-W4tcher: @ikelos, I made a PoC to set requirements on (unversionable internal) APIs with a wrapper:

```python
# Requirements can be stacked
# Raises an exception (at plugin runtime) on requirement failure
@requirements.VersionRequirement.version_requirement_wrapper(
    name="linux-tainting", component=tainting.Tainting, version=(1, 0, 0)
)
def get_taints_as_plain_string(self) -> str:
```

Disregarding the underlying machinery, do you think this design could be an appropriate solution to flexibly emit requirements ?

ikelos: Yeah, you're probably right, it should get at least an additive MINOR version bump.  I think we've managed ok before but for the sake of accuracy you're right.

I'm not sure that a module not being exposed as a method on a class is that good an excuse to just keep heaping stuff on the module.  Plus if we don't start now, we'll end up doing it for every new feature in existance, and that'll become unwieldy down the line (assuming it isn't already unwieldy with all the methods it has at the moment).

If people don't know about tainting, they're not going to notice it in the module documentation, and if they know that's what they're looking for, they'll find it quickly enough (particularly with existing plugins in the framework that make use of it).  I don't think that argument holds up as to why it needs to be part of the module interface.

So at the moment, still not convinced it should be part of module.  

I'm actually not against the `version_requirement_wrapper` idea, although I don't think it helps as much as we'd want.  It'll do the code to fail when a requirement isn't met, and it makes it visible at the start of a function, but what about a class that makes use of several different versions of a function?  There's no way of using that information in something calling the plugin, so it will likely just perpetuate an intricate web rather than doing anything all that useful to keep it managable.  So for example, let's say the module now requires "Tainting-1.0.0", does the plugin know that calling `module` will require that?  I think tagging the point of calling with a dependency doesn't help, because by the time you've called it, it's too late.  You want to know up front, before you've invested any processing power, whether something's going to fail later on or not.

So tagging individual subfunctions within `module` will just create an object you can't easily use without checking each of the individual components in it that you need.  That's why I'd like to keep the functionality separate, rather than tying all of their functionality to the highest requirement of any of the subcomponents...

Abyss-W4tcher: I can understand that the extra-complexity for the sake of convenience might not be worthy, and as its counterpart is not that much longer I think it is best to stop here and move on 👍. I'll remove the proxies.

Currently, internal APIs calling `LinuxUtilities` aren't version restrained:

https://github.com/volatilityfoundation/volatility3/blob/a6777146f6fd789c10686b9ed1f8a5ea3521f9a5/volatility3/framework/symbols/linux/extensions/__init__.py#L1273

To fix this, we would need to basically make all internal classes inherit from `VersionRequirement`, so that they get automatically checked at instantiation. The decorator is only here as a "panic/debug" safeguard to help us understand immediately where a bug originates if we forgot to update API callers. Indeed, it can get "messy" if we start versioning internal APIs one-by-one with it. 

I think this versioning issue is already in the framework unfortunately, but I do not have an immediate way of solving it properly without updating each LinuxUtilities call. I opened an issue (#1554) as it starts to diverge from the original `modxview` plugin.

Abyss-W4tcher:
Hopefully, all issues should have been addressed or correctly relocated to the appropriate place. In the end, only a minor bump is needed, for the new `Tainting` class.

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -0,0 +1,184 @@
...
+            run_results["hidden_modules"] = list(
+                hidden_modules.Hidden_modules.get_hidden_modules(
+                    context,
+                    kernel_name,
+                    known_modules_addresses,
+                    modules_memory_boundaries,
+                )
+            )
+
+        return run_results
```
ikelos, line 127: I think this is ok, and the interface seems extensible.  I worry slightly about how we'd make changes to the formatting, but I guess we'd just change the keys of the return dictionary if they wanted to change their format so... should be fine.

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def _generator(self):
+        kernel_name = self.config["kernel"]
+        run_results = self.run_modules_scanners(self.context, kernel_name)
+        modules_offsets = {}
+        for key in ["lsmod", "check_modules", "hidden_modules"]:
+            modules_offsets[key] = set(module.vol.offset for module in run_results[key])
```
ikelos: So this bit is building a lookup of module_offsets of each of the search types that'll get used later?  Feels like a lot of variables floating around to cope with the way the results were originally generated.

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -0,0 +1,184 @@
...
+
+        seen_addresses = set()
+        for modules_list in run_results.values():
```
ikelos: This all seems like a lot of juggling just because the data structure is:

```
run_results[scan]
  -> [module]
    -> taints, etc.
```

Presumably the module objects instantiated on the same offset are the same, so... why not when generating them build it as:
```
  run_results[module.vol.offset] 
    -> (module, [scan] -> true or false/None/not present)
```

Seems like it might be easier to follow what's going on, rather than shuffling and moving everything around...

Little more preprocessing, but I think it'd be easier to comprehend than the acrobatics going on here...

ikelos (commented):
Ok generally looks good, happy with the tainting code being pulled out.  I find the intermediate data structure easy to understand, but then we go through all kinds of loops to produce the output, which I think would be easier with a different intermediate data structure.  If you genuinely think the current method is the best way then I'll go along with it, but I'd probably then ask for better documentation/variable names to say what's going on in all the loops.  Otherwise I think we could improve `run_module_scanners` to process the results before returning them.

volatility3/framework/plugins/linux/modxview.py
Abyss-W4tcher: I made an update which follows your suggestion 👍.

volatility3/framework/plugins/linux/modxview.py
Abyss-W4tcher: I didn't want to post-process the results, when the caller might want to do it himself. `run_modules_scanners`'s goal it s to provide abstraction for module scanning plugins (as you should have noticed, it needs a small setup). Its purpose might be extended in the future, so I would prefer to keep it simple (plugin -> run_result).

Abyss-W4tcher:
Sorry for the force-push message, I had an awkward git moment on my local installation '/

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -0,0 +1,189 @@
...
+        aggregated_modules = {}
+        # We want to be explicit on the plugins results we are interested in
+        for plugin_name in ["lsmod", "check_modules", "hidden_modules"]:
+            # Iterate over each recovered module
+            for module in run_results[plugin_name]:
+                # Use offsets as unique keys, whether a module
+                # appears in many plugin runs or not
+                if aggregated_modules.get(module.vol.offset, None) is not None:
+                    # Append the plugin to the list of originating plugins
+                    aggregated_modules[module.vol.offset][1].append(plugin_name)
```
ikelos, line 141: Ugh, I really dislike indexing values rather than naming them, but I'll allow it in this instance given the line below details the data structure format reasonably well.

ikelos (approved):
Ok, I'm happier with this, it's easier to follow, thanks for making the updates.  5:)

volatility3/framework/plugins/linux/modxview.py
Abyss-W4tcher: I could have made an Enum indeed, however given the context and comments I thought it was enough.

Abyss-W4tcher:
Thank you for your time in reviewing this PR. 👍

Issue #134, opened 2019-11-12
Need to add support for pages with PROT_NONE permissions


gleeda:
The `_PAGE_PRESENT` bit is cleared when `mprotect(...PROT_NONE)` is called on a page, therefore it is missed.

See: https://volatility-labs.blogspot.com/2015/05/using-mprotect-protnone-on-linux.html

ikelos:
Just to keep everything in one place, I'd recommend a LinuxMixin (that follows the same procedures as the WindowsMixin, including subclasses with Intel and the LinuxMixin) and then update the LintelStacker class to use those classes instead of the stock Intel ones...

gleeda:
Just a note that I'm hoping we'll have a resolution to the `unnamed_field` issue soon.  It is currently a potential blocker for verifying this issue.

ikelos:
#151 has been resolved, so hopefully this can keep moving forward?

gleeda:
I'm still having issues getting dwarf2json profiles to work with this version of Kali for some reason :-/ 

```$ python3 vol.py -vvv -f ~/Documents/Virtual\ Machines/Kali.vmwarevm/Kali-e1e76f89.vmem linux.proc.Maps
Volatility 3 Framework 1.0.0-beta.1
INFO     root        : Volatility plugins path: ['/Users/gleeda/Work/DEV/volatility3/volatility/plugins', '/Users/gleeda/Work/DEV/volatility3/volatility/framework/plugins']
INFO     root        : Volatility symbols path: ['/Users/gleeda/Work/DEV/volatility3/volatility/symbols', '/Users/gleeda/Work/DEV/volatility3/volatility/framework/symbols']
INFO     volatility.plugins.yarascan: Python Yara module not found, plugin (and dependent plugins) not available
DEBUG    volatility.framework: No module named 'yara'
DEBUG    volatility.framework: Failed to import module volatility.plugins.yarascan based on file: yarascan
INFO     volatility.plugins.yarascan: Python Yara module not found, plugin (and dependent plugins) not available
DEBUG    volatility.framework: No module named 'yara'
DEBUG    volatility.framework: Failed to import module volatility.plugins.windows.svcscan based on file: windows/svcscan
INFO     volatility.plugins.yarascan: Python Yara module not found, plugin (and dependent plugins) not available
DEBUG    volatility.framework: No module named 'yara'
[...]

PR #1335, opened 2024-11-07
Linux - PROT_NONE, L1TF mitigation and layer address translation fix

gcmoreira:
This PR introduces improvements and fixes for address translation in the Linux Intel layers

# PROT_NONE pages 
First, this PR fixes one of the oldest issues in the queue https://github.com/volatilityfoundation/volatility3/issues/134, adding support for PROT_NONE protected pages.

Using the @gleeda PoC from their excellent post [Using mprotect PROT_NONE on Linux](https://volatility-labs.blogspot.com/2015/05/using-mprotect-protnone-on-linux.html). 
```shell
$ ./victim 
PID 2116
buffer at 0x76f63a132000       <= PROT_NONE vma
buffer2 at 0x76f63a133000
```

Before:
```python
>>> task.pid
2116

>>> proc_layer.read(0x76F63A133000, 10)
b'not here\x00\x00'

>>> proc_layer.read(0x76F63A132000, 10)
  File "/home/user/vol3/volatility3/framework/layers/linear.py", line 45, in read
  File "/home/user/vol3/volatility3/framework/layers/intel.py", line 295, in mapping
  File "/home/user/vol3/volatility3/framework/layers/intel.py", line 351, in _mapping
  File "/home/user/vol3/volatility3/framework/layers/intel.py", line 159, in _translate
volatility3.framework.exceptions.PagedInvalidAddressException: Page Fault at entry 0x40fffffed82d920 in page entry
```

After:
```python
>>> proc_layer.read(0x76F63A133000, 10)
b'not here\x00\x00'

>>> proc_layer.read(0x76F63A132000, 10)
b'find me\x00\x00\x00'
```

# Intel Side Channel Vulnerability L1TF mitigation.

In 2018, to mitigate Intel's L1TF (L1 Terminal Fault) vulnerability, that code in Linux kernel was again updated in [this commit](https://github.com/torvalds/linux/commit/6b28baca9b1f0d4a42b865da7a05b1c81424bd5c), incorporating PFN inversion to calculate the PTE. This PR includes support for handling these cases as well.

# __PHYSICAL_MASK_SHIFT and _maxphyaddr
While addressing the previous issues, I found that the changes didn't work for older kernels. After further investigation, it turned out the problem was the `physical_mask` and the reason was [this](https://github.com/torvalds/linux/commit/b83ce5ee91471d19c403ff91227204fb37c95fb2) commit. 

In the Linux kernel, the [__PHYSICAL_MASK_SHIFT](https://elixir.bootlin.com/linux/v6.11.6/source/arch/x86/include/asm/page_64_types.h#L54) is a mask used to extract the physical address from a PTE. In Volatility3, this is referred to as `_maxphyaddr`.

Until kernel version 4.17, Linux x86-64 used a 46-bit mask. With [this](https://github.com/torvalds/linux/commit/b83ce5ee91471d19c403ff91227204fb37c95fb2) commit, this was extended to 52 bits, applying to both 4 and 5-level page tables.

We previously used 52 bits for all Intel 64-bit systems, but this produced incorrect results for **PROT_NONE** pages. Since the mask value is defined by a preprocessor macro, it's difficult to detect the exact bit shift used in the current kernel. 
Using 46 bits has proven reliable for our use case, as seen in tools like crashtool. See [this](https://github.com/crash-utility/crash/blob/master/x86_64.c#L293) and [this](https://github.com/crash-utility/crash/blob/master/defs.h#L4073).

### DIFF
--- a/volatility3/framework/constants/linux/__init__.py
+++ b/volatility3/framework/constants/linux/__init__.py
@@ -11,14 +11,6 @@ KERNEL_NAME = "__kernel__"
 
 """The value hard coded from the Linux Kernel (hence not extracted from the layer itself)"""
 
-# Translation Layer constants
-PAGE_BIT_PRESENT = 0
-PAGE_BIT_PSE = 7  # Page Size Extension: 4 MB (or 2MB) page
-PAGE_BIT_PROTNONE = 8
-PAGE_BIT_PAT_LARGE = 12  # 2MB or 1GB pages
-PAGE_PRESENT = 1 << PAGE_BIT_PRESENT
-PAGE_PROTNONE = 1 << PAGE_BIT_PROTNONE
-
 # include/linux/sched.h
 PF_KTHREAD = 0x00200000  # I'm a kernel thread
 
--- a/volatility3/framework/constants/linux/__init__.py
+++ b/volatility3/framework/constants/linux/__init__.py
@@ -11,6 +11,14 @@ KERNEL_NAME = "__kernel__"
 
 """The value hard coded from the Linux Kernel (hence not extracted from the layer itself)"""
 
+# Translation Layer constants
+PAGE_BIT_PRESENT = 0
+PAGE_BIT_PSE = 7  # Page Size Extension: 4 MB (or 2MB) page
+PAGE_BIT_PROTNONE = 8
+PAGE_BIT_PAT_LARGE = 12  # 2MB or 1GB pages
+PAGE_PRESENT = 1 << PAGE_BIT_PRESENT
+PAGE_PROTNONE = 1 << PAGE_BIT_PROTNONE
+
 # include/linux/sched.h
 PF_KTHREAD = 0x00200000  # I'm a kernel thread
 
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -13,7 +13,6 @@ from volatility3 import classproperty
 from volatility3.framework import exceptions, interfaces, constants
 from volatility3.framework.configuration import requirements
 from volatility3.framework.layers import linear
-from volatility3.framework.constants import linux as linux_constants
 
 vollog = logging.getLogger(__name__)
 
@@ -22,6 +21,16 @@ INTEL_TRANSLATION_DEBUGGING = False
 
 class Intel(linear.LinearlyMappedLayer):
     """Translation Layer for the Intel IA32 memory mapping."""
+
+    _PAGE_BIT_PRESENT = 0
+    _PAGE_BIT_PSE = 7  # Page Size Extension: 4 MB (or 2MB) page
+    _PAGE_BIT_PROTNONE = 8
+    _PAGE_BIT_PAT_LARGE = 12  # 2MB or 1GB pages
+
+    _PAGE_PRESENT = 1 << _PAGE_BIT_PRESENT
+    _PAGE_PSE = 1 << _PAGE_BIT_PSE
+    _PAGE_PROTNONE = 1 << _PAGE_BIT_PROTNONE
+    _PAGE_PAT_LARGE = 1 << _PAGE_BIT_PAT_LARGE
 
     _entry_format = "<I"
     _page_size_in_bits = 12
@@ -209,10 +218,10 @@ class Intel(linear.LinearlyMappedLayer):
                     "Page Fault at entry " + hex(entry) + " in table " + name,
                 )
             # Check if we're a large page
-            if large_page and (entry & (1 << linux_constants.PAGE_BIT_PSE)):
+            if large_page and (entry & self._PAGE_PSE):
                 # Mask off the PAT bit
-                if entry & (1 << linux_constants.PAGE_BIT_PAT_LARGE):
-                    entry -= 1 << linux_constants.PAGE_BIT_PAT_LARGE
+                if entry & self._PAGE_PAT_LARGE:
+                    entry -= self._PAGE_PAT_LARGE
                 # We're a large page, the rest is finished below
                 # If we want to implement PSE-36, it would need to be done here
                 break
@@ -545,10 +554,7 @@ class LinuxMixin(Intel):
         return pte & self.pte_flags_mask
 
     def is_pte_present(self, entry: int) -> bool:
-        return (
-            self.pte_flags(entry)
-            & (linux_constants.PAGE_PRESENT | linux_constants.PAGE_PROTNONE)
-        ) != 0
+        return (self.pte_flags(entry) & (self._PAGE_PRESENT | self._PAGE_PROTNONE)) != 0
 
     def _page_is_valid(self, entry: int) -> bool:
         # Overrides the Intel static method with the Linux-specific implementation
@@ -556,7 +562,7 @@ class LinuxMixin(Intel):
 
     def pte_needs_invert(self, entry) -> bool:
         # Entries that were set to PROT_NONE (PAGE_PRESENT/PAGE_GLOBAL) are inverted
-        return not (entry & linux_constants.PAGE_PRESENT)
+        return not (entry & self._PAGE_PRESENT)
 
     def protnone_mask(self, entry: int) -> int:
         """Gets a mask to XOR with the page table entry to get the correct PFN"""
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -569,7 +569,7 @@ class LinuxMixin(Intel):
 
     def _protnone_mask(self, entry: int) -> int:
         """Gets a mask to XOR with the page table entry to get the correct PFN"""
-        return ~0 & self._register_mask if self._pte_needs_invert(entry) else 0
+        return self._register_mask if self._pte_needs_invert(entry) else 0
 
     def _pte_pfn(self, entry: int) -> int:
         """Extracts the page frame number from the page table entry"""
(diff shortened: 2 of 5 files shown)


volatility3/framework/layers/intel.py
```diff
@@ -203,10 +209,10 @@ def _translate_entry(self, offset: int) -> Tuple[int, int]:
                     "Page Fault at entry " + hex(entry) + " in table " + name,
                 )
             # Check if we're a large page
-            if large_page and (entry & (1 << 7)):
+            if large_page and (entry & (1 << linux_constants.PAGE_BIT_PSE)):
                 # Mask off the PAT bit
-                if entry & (1 << 12):
-                    entry -= 1 << 12
+                if entry & (1 << linux_constants.PAGE_BIT_PAT_LARGE):
```
ikelos: These shouldn't be in `linux_constants` if they're for all intel layers?  They're presumably defined as part of the architecture and shouldn't change.  If they are OS specific then we need to rethink how we access them.

volatility3/framework/layers/intel.py
```diff
@@ -163,12 +164,17 @@ def _translate(self, offset: int) -> Tuple[int, int, str]:
...
-            offset, position, 0
-        )
+
+        pfn = self.pte_pfn(entry)
+        page_offset = self._mask(offset, position, 0)
+        page = pfn << self.page_shift | page_offset
 
         return page, 1 << (position + 1), self._base_layer
 
+    def pte_pfn(self, entry: int) -> int:
```
ikelos: This is a bump to the MINOR version of the framework, at a minimum.

volatility3/framework/layers/intel.py
```diff
@@ -163,12 +164,17 @@ def _translate(self, offset: int) -> Tuple[int, int, str]:
                 entry,
                 f"Page Fault at entry {hex(entry)} in page entry",
             )
-        page = self._mask(entry, self._maxphyaddr - 1, position + 1) | self._mask(
-            offset, position, 0
-        )
+
+        pfn = self.pte_pfn(entry)
```
ikelos: What's the improvement made by these changes?  It feels like it just splits things and rejoins them, or is it just for clarity?

volatility3/framework/constants/linux/__init__.py
```diff
@@ -11,6 +11,14 @@
 
 """The value hard coded from the Linux Kernel (hence not extracted from the layer itself)"""
 
+# Translation Layer constants
+PAGE_BIT_PRESENT = 0
+PAGE_BIT_PSE = 7  # Page Size Extension: 4 MB (or 2MB) page
+PAGE_BIT_PROTNONE = 8
```
ikelos: These don't belong in linux unless they're linux specific constants (in which case we need to rethink how they're used based on which OS the layer is constructed on.

volatility3/framework/layers/intel.py
```diff
@@ -501,3 +507,85 @@ class WindowsIntel32e(WindowsMixin, Intel32e):
 
     def _translate(self, offset: int) -> Tuple[int, int, str]:
         return self._translate_swap(self, offset, self._bits_per_register // 2)
+
+
+class LinuxMixin(Intel):
+    @functools.cached_property
+    def register_mask(self) -> int:
```
ikelos: This is exposing a lot of additional fields and I just want to make sure the balance is right between things that the OS might actually need to know about, and things that really should only be used internally by the intel layer.  At the moment the guts feel like they're on the outside and I want to be sure every single method we expose is worthwhile.

I'd also change TODOs that we know aren't accurate to FIXMEs please, otherwise there'll be no urgency of any kind of improve them and this will all just be adding extra lines for no reason.

ikelos (commented):
Roughly happy with the concept, but needs to be more accurate on what's linux and what's general Intel.  I'm a little concern that we're exposing the guts of the layer and that not that many parameters need to be accessible by the OS/outside the layer (we did name method starting with `_` for a reason).  If we're happy it's all important and necessary then it's not far off getting merged.

volatility3/framework/layers/intel.py
gcmoreira: Right. Where do you think it's the best place to move them? As a class attribute or as a global constant at the top of the intel.py file?

ikelos: Hmmm, I feel they probably should live in the class?  I guess that'd only be an additive change, so just a bump of the MINOR version on the framework.  It would be the first referable constant we don't keep in the `constants` files though?  I'm torn, which do you prefer?

gcmoreira: Yeah, I suppose only the layers will use them, but I don't think it's worth try to hide them if it means making the code ugly

gcmoreira: I'll make them internal attributes. I would prefer to keep them uppercase to emphasize that they are constants, but if you want me to follow the lowercase style please let me know.

```python
class Intel(linear.LinearlyMappedLayer):
    _PAGE_BIT_PRESENT = 0
    _PAGE_PRESENT = 1 << _PAGE_BIT_PRESENT
...
    _entry_format = "<I"
    _page_size_in_bits = 12
    _bits_per_register = 32
```

volatility3/framework/layers/intel.py
gcmoreira: First, while the use of _mask() was convenient, it does not allow for overriding the behavior of extracting the PFN. This is why the pte_pfn() function was added. Additionally, I took the opportunity to name the individual components using terminology commonly found in the literature and, for example, in the Linux kernel code. 
In my opinion, this makes the code more maintainable, and I believe it is crucial for this part of the framework. Since we don't modify it frequently and maybe a few people may work on it, it is essential to document it thoroughly. This will make life easier for the "unfortunate" person who needs to understand these lines in the future.

volatility3/framework/layers/intel.py
gcmoreira: I made all the methods introduced by the LinuxMixin internal.

> I'd also change TODOs that we know aren't accurate to FIXMEs please, otherwise there'll be no urgency of any kind of improve them and this will all just be adding extra lines for no reason.

Sorry, I'm not sure which TODOs you're referring to here. Could you please point them out to me?

volatility3/framework/layers/intel.py
gcmoreira: Ok, done.. please have a look

volatility3/framework/layers/intel.py
gcmoreira: done

volatility3/framework/constants/linux/__init__.py
gcmoreira: done

gcmoreira:
Hey @ikelos, I believe everything has been addressed in this PR. Are you waiting on anything else from me, or it's just you haven't had a chance to review it yet?

volatility3/framework/layers/intel.py
```diff
@@ -501,3 +516,84 @@ class WindowsIntel32e(WindowsMixin, Intel32e):
...
+        physical_mask = (1 << self._maxphyaddr) - 1
+        # TODO: Come back once SME support is available in the framework
+        return physical_mask
+
+    @functools.cached_property
+    def page_mask(self) -> int:
+        # Note that within the Intel class it's a class method. However, since it uses
+        # complement operations and we are working in Python, it would be more careful to
+        # limit it to the architecture's pointer size.
+        return ~(self.page_size - 1) & self._register_mask
```
ikelos, line 539: Yeah, I'm not a big fan of `~` in python, just because python ints can be any size.  If you're happy this will work properly I'll accept it, but I'd prefer some maths using the known actual limit to form the complement...

volatility3/framework/layers/intel.py
```diff
@@ -501,3 +516,84 @@ class WindowsIntel32e(WindowsMixin, Intel32e):
...
+    def _physical_page_mask(self) -> int:
+        return self.page_mask & self._physical_mask
+
+    @functools.cached_property
+    def _pte_pfn_mask(self) -> int:
+        return self._physical_page_mask
+
+    @functools.cached_property
+    def _pte_flags_mask(self) -> int:
+        return ~self._pte_pfn_mask & self._register_mask
```
ikelos, line 551: Same as above regarding `~`

ikelos (commented):
Ok, still can't entirely see why some of those need pulling out into properties rather than just one off variables defined at the class level, but I accept the point about documenting clearly.  I'd prefer not to let `~` sneak in, but I'm pretty sure Andrew managed to sneak some in in the past, so I'll let it slide if you want.  Just shout when you want it merged.

Also, this is a deep change to the core Intel layer, so after merge please keep a close eye on general use for the first week or two after merge, and longer term for tickets that might be because of this for a couple months.  I've cut the release branch before this so that we get a few more months testing of it before it goes live.

volatility3/framework/layers/intel.py
ikelos: I thought there was one that said "we know this isn't right", but I can only find the revisit once SME is supported, which is fine as a TODO.

volatility3/framework/layers/intel.py
gcmoreira: Yep, that's exactly what I was trying to explain in that comment :) .. it's safe when there's an AND with a mask like that one.
Actually, that's a bug (or let's say, an unsafe usage) of the current Intel implementation that this PR addresses.
See https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/layers/intel.py#L89

volatility3/framework/layers/intel.py
gcmoreira: It's safe when there's an AND with a mask like that one

volatility3/framework/layers/intel.py
ikelos: Any chance you could submit a separate PR to fix that one up (or at least file a bug against it) please?

PR #1338, opened 2024-11-07
Introduce scatter-gather scatterlists

Abyss-W4tcher:
Hello, 

scatter-gather lists allow to scatter (store) or gather (retrieve) data, and is notably used for DMA data transfer. It is useful when allocating contiguous large chunks of data is not possible/easy. Here, obviously, we only need the gather mechanisms :). These capabilities will be needed for later/potential DMA buffers extraction capabilities.

A general doc. for the curious : https://jungo.com/windriver/scatter-gather-or-contiguous-dma-data-transfer/


volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -2410,3 +2410,107 @@ def get_nodes(self) -> Iterator[int]:
...
+    def _get_sg_max_single_alloc(self) -> int:
+        """Based on kernel's SG_MAX_SINGLE_ALLOC.
+
+        Doc. from kernel source :
+            * Maximum number of entries that will be allocated in one piece, if
+            * a list larger than this is required then chaining will be utilized.
+        """
+        return self._context.layers[self.vol.layer_name].page_size // self.vol.size
+
+    def _sg_next(self) -> interfaces.objects.ObjectInterface:
```
ikelos: This should technically be `Optional[interfaces.objects.ObjectInterface]`

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -2410,3 +2410,107 @@ def get_nodes(self) -> Iterator[int]:
...
+    def get_content(
+        self,
+    ) -> Iterator[bytes]:
+        """Traverse a scatterlist to gather content located at each
+        dma_address position.
+
+        Returns:
+            An iterator of bytes
+        """
+        physical_layer = self._context.layers["memory_layer"]
```
ikelos: So this is... fragile.  For Intel layers, the underlying layer will likely be `memory_layer`  (but not in the case of a virtualized intel system, for instance), and there's no telling that Arm's lower layer will be called `memory_layer`.  I don't immediately have a better way of doing this.  The layer's dependencies list would tell you all the layers it lives on top of, and they're *likely* to be in order, but short of mapping an address down a layer (and if you pick the wrong one, you might hit swap or something else) there's no guaranteed way of knowing how the layers all fit together...  Any thoughts how best to handle this?

ikelos (commented):
Generally looks good, comments and typing information are much appreciated.  The main sticking point is how to determine which layer is the physical layer for which there's no guaranteed method at the moment (and this make assumptions about the intel layer).  We should really solve that before merging this (and yes, we may already have done similar things in the code base, but that just adds to the case of needing to fix the situation).

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: What we want to achieve, is basically getting `kernel` OR `process_layer` - 1 (base) layer, which should in most cases be the RAM. 

Maybe something like `self._context.layers[self.vol.layer_name].base_layer` should do it ?

Abyss-W4tcher: Ok, so here it can be a bit tricky, as we are "blind" of the layer position in the stacking list. However, pdbscan, linux.py and mac.py will refer to the physical layer by the config key "memory_layer", which is a TranslationLayerRequirement for Intel. We can use this information to determine where we are. FYI, the AArch64 implementation has no reason to change this naming, which would only complicate things I think.

Consequently, this implementation should do the trick :

```python 
# Either physical is layer-1 because we are a module layer, either physical is the current layer
physical_layer_name = self.context.layers[self.vol.layer_name].config.get(
    "memory_layer", self.vol.layer_name
)
physical_layer = self.context.layers[physical_layer_name]
```

ikelos: Yep, definitely need to use the code above to get it from this object's layer.  At the moment all plugins only deal with one image, but that's absolutely not the intention long term.

Abyss-W4tcher: Change has been applied.

Abyss-W4tcher:
A minor bump has been applied, this should be good to go.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -2511,6 +2511,10 @@ def get_content(
         Returns:
             An iterator of bytes
         """
-        physical_layer = self._context.layers["memory_layer"]
+        # Either "physical" is layer-1 because this is a module layer, either "physical" is the current layer
```
ikelos: This comment doesn't make that much sense, it's two eithers and no or?

Abyss-W4tcher: Sorry, probably a copy/paste mistake. It's fixed.

ikelos (approved):

PR #1341, opened 2024-11-08
Fix typo in SuspiciousThreads

joren485:


ikelos:
Good spot, and you're right, it does need fixing.  It's only been in the codebase for 5 months and I think adding a second one with the typo (that just raises a warning and then runs the correct version) would just be confusing.  Let's hope this doesn't cause confusion...

PR #1344, opened 2024-11-08
Add method to read inode content into memory

joren485:
This PR adds a `read_inode_content` method to read the conent of an inode. This makes it possible to use the content of an inode inside other plugins.

### DIFF
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -6,12 +6,13 @@ import math
 import logging
 import datetime
 from dataclasses import dataclass, astuple
-from typing import List, Set, Type, Iterable, Optional
+from typing import List, Set, Type, Iterable, Optional, IO
 
 from volatility3.framework import renderers, interfaces
 from volatility3.framework.renderers import format_hints
 from volatility3.framework.interfaces import plugins
 from volatility3.framework.configuration import requirements
+from volatility3.framework.symbols.linux.extensions import inode as inodeClass
 from volatility3.plugins import timeliner
 from volatility3.plugins.linux import mountinfo
 
@@ -389,7 +390,7 @@ class InodePages(plugins.PluginInterface):
 
     _required_framework_version = (2, 0, 0)
 
-    _version = (1, 0, 1)
+    _version = (1, 1, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
@@ -421,7 +422,8 @@ class InodePages(plugins.PluginInterface):
 
     @staticmethod
     def read_inode_content(
-        inode: interfaces.objects.ObjectInterface,
+        inode: inodeClass,
+        buffer: IO,
         page_size: int,
     ) -> Optional[bytes]:
         """
@@ -430,44 +432,43 @@ class InodePages(plugins.PluginInterface):
         As the page indices might be out of order, we use seek and write to
         make sure each page is written to the correct position.
 
+        By using truncate/seek, provided the filesystem supports it, a sparse file will be
+        created, saving both disk space and I/O time.
+
         Args:
             inode: The inode to read
+            buffer: the file-like object to write the inode contents too
             page_size: The pages size of the memory layer to read the inode data from
         """
-
         if not inode.is_reg:
             vollog.error("The inode is not a regular file")
             return None
 
-        inode_size = inode.i_size
+        try:
+            buffer.truncate(inode.i_size)
 
-        with io.BytesIO() as buffer:
-            try:
-                for page_idx, page_content in inode.get_contents():
-                    current_fp = page_idx * page_size
-                    max_length = inode_size - current_fp
-                    page_bytes = page_content[:max_length]
-                    if current_fp + len(page_bytes) > inode_size:
-                        vollog.error(
-                            "Page out of file bounds: inode 0x%x, inode size %d, page index %d",
-                            inode.vol.object,
-                            inode_size,
-                            page_idx,
-                        )
-                    buffer.seek(current_fp)
-                    buffer.write(page_bytes)
-
-            except IOError as e:
-                vollog.error(f"Unable to read inode: {e!s}")
-                return None
-
+            for page_idx, page_content in inode.get_contents():
+                current_fp = page_idx * page_size
+                max_length = inode.i_size - current_fp
+                page_bytes = page_content[:max_length]
+                if current_fp + len(page_bytes) > inode.i_size:
+                    vollog.error(
+                        "Page out of file bounds: inode 0x%x, inode size %d, page index %d",
+                        inode.vol.object,
+                        inode.i_size,
+                        page_idx,
+                    )
+                buffer.seek(current_fp)
+                buffer.write(page_bytes)
             buffer.seek(0)
-            return buffer.read()
+        except IOError as e:
+            vollog.error(f"Unable to read inode: {e!s}")
+            return None
 
     @classmethod
     def write_inode_content_to_file(
         cls,
-        inode: interfaces.objects.ObjectInterface,
+        inode: inodeClass,
         filename: str,
         open_method: Type[interfaces.plugins.FileHandlerInterface],
         vmlinux_layer: interfaces.layers.TranslationLayerInterface,
@@ -480,11 +481,10 @@ class InodePages(plugins.PluginInterface):
             open_method: class for constructing output files
             vmlinux_layer: The kernel layer to obtain the page size
         """
-        buffer = cls.read_inode_content(inode=inode, page_size=vmlinux_layer.page_size)
-        if buffer:
-            with open_method(filename) as h_file:
-                h_file.write(buffer)
-        return None
+        with open_method(filename) as h_file:
+            cls.read_inode_content(
+                inode=inode, buffer=h_file, page_size=vmlinux_layer.page_size
+            )
 
     def _generator(self):
         vmlinux_module_name = self.config["kernel"]
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -1,7 +1,6 @@
 # This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
-import io
 import math
 import logging
 import datetime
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -458,11 +454,34 @@ def write_inode_content_to_file(
...
+                    buffer.write(page_bytes)
+
+            except IOError as e:
+                vollog.error(f"Unable to read inode: {e!s}")
+                return None
+
+            buffer.seek(0)
+            return buffer.read()
+
+    def write_inode_content_to_file(
```
ikelos, line 468: By not parameterizing `open` this method is only usable by this plugin.  We try to encourage methods that could be reusable to be made class or static methods, and parameterized with everything they need (such as an open method).  That also changes the API in a non-backwards compatible way, so if were to allow it (unlikely) it would require a major version bump of the plugin.  Better to just parameterize with the same parameters as the original version and then no version bumps required.

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -420,33 +420,29 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
-    def write_inode_content_to_file(
-        inode: interfaces.objects.ObjectInterface,
-        filename: str,
-        open_method: Type[interfaces.plugins.FileHandlerInterface],
-        vmlinux_layer: interfaces.layers.TranslationLayerInterface,
-    ) -> None:
-        """Extracts the inode's contents from the page cache and saves them to a file
+    def read_inode_content(
+            inode: interfaces.objects.ObjectInterface,
+            vmlinux_layer: interfaces.layers.TranslationLayerInterface
```
ikelos: In general, I much prefer that methods take a context and a simple string rather than a full module (which carries around it own copy of the context) to ensure that it's very clear which context is in use (and plugins don't suddenly go weird because an object was made in a different place).  At the moment, we don't do that for `write_inode_content_to_file` and I'm not sure why I let that slide, but given this is a new function, I'd much prefer to take a context and a layer name and then do the lookup, to make everything more explicit as to what's going on please.

ikelos (commented):
Cool, thanks.  Roughly happy with it, but you need to keep the old method the same (so static method, no changing the input parameters) and I'd prefer the function signature for `read_inode_content_from_file` matched the style of the rest of the codebase, which is `context` as the first parameter, and then simple types for everything else please...

volatility3/framework/plugins/linux/pagecache.py
joren485: Thanks for your review! I changed `write_inode_content_to_file` back to the original parameters. I did make `write_inode_content_to_file` a `classmethod` (which shouldn't impact callers of the method), for easier access to `read_inode_content`, but I can change it back to a `staticmethod` if you wish.

I also did not do a (minor) version bump, but I could set `InodePages._version` to `(1, 0, 2)` if you want.

volatility3/framework/plugins/linux/pagecache.py
joren485: As `read_inode_content` only used `vmlinux_layer` to access the page size, the method now just takes the page size as an argument.

gcmoreira:
Hey @joren485, thanks for this contribution. My 2 cents: Adding a method to read the inode contents into memory could be useful for some applications, but calling it within `write_inode_content_to_file()` removes the ability to produce a [sparse file](https://en.wikipedia.org/wiki/Sparse_file). 
Keep in mind that not all inode pages may be cached, so the resulting file on disk might contain many zeroed pages. For instance, if an inode's size is 1TB but only the one page is cached, the original method would report an apparent size of 1TB while the actual disk space used is only 4096 bytes. With your changes, both apparent and actual sizes would align at 1TB. This is important for large-scale file recovery.

volatility3/framework/plugins/linux/pagecache.py
ikelos: Nope, class method is fine, it just needs to have the same function signature.  You should bump the minor version number though, so that tools that require the new function can ensure they're using a version that includes it.  `1.0.2` would be a patch bump, not a minor bump.  It would need to be `1.1.0`

gcmoreira:
Also, returning the entire buffer/file like that is inefficient for large files. Locally, it can lead to heavy disk I/O due to swapping. For remote applications, it results in unnecessary data being sent over the network.

Lastly, the read_inode_content() implementation contains a bug, as it assumes all pages are present. This can result in the final buffer being a different size than the original inode, leading to inaccurate conclusions.

joren485:
>  It would need to be 1.1.0

Done

> removes the ability to produce a sparse file.

I changed `read_inode_content` to take a `typing.IO` object (e.g. a file or a BytesIO) as an argument and added a `buffer.truncate` call. The functionality should now be the same as the original. 

> Lastly, the read_inode_content() implementation contains a bug, as it assumes all pages are present. This can result in the final buffer being a different size than the original inode, leading to inaccurate conclusions.

Good catch! As I did not change the functionality of reading inode contents and do not have access to a memory image to reproduce the bug, I think it is best to create a separate issue for this.

gcmoreira:
EDIT: Sorry I was on the phone and I missed some of the details.

> Good catch! As I did not change the functionality of reading inode contents and do not have access to a memory image to reproduce the bug, I think it is best to create a separate issue for this.

By calling truncate(), you've now fixed that bug in your implementation. The original write_inode_content_to_file() method already handles those cases without issue. 

There's still something a bit off here, a method named read_inode_content() that writes to a file doesn't quite feel right. Also the docstring is inaccurate "Read the contents of an inode into memory."

@joren485 Is there a particular use case you had in mind for read_inode_content()? It feels a bit like we're stretching to find one, but I'll defer to @ikelos if he thinks it has potential.

ikelos:
I don't have a horse in this race.  The write function was put in specifically to write sparsely, so changing/removing that functionality seems bad.  Now that the read function is parameterized with an IO object, I believe they should operate with the same results (it would be good if someone has an example to make sure the file still is actually sparse when it's written).  It sounds as though @joren485 had another plugin that wanted to make use of it?  If so, and if it doesn't remove functionality, I'm ok with it going in, but it would be nice to know there's a specific reason for it rather than just making changes for the sake of it...

atcuno:
It is on the parity list to break this out into a real API so other plugins can get file contents, so I would like to see it merged in some form.

Some common examples:

- Reading /etc/ld.so.preload for malicious LD_PRELOAD libraries
- Reading /home/*/.*profile files to check for hijacked $PATH
- Auto parsing cronjobs
- Auto parsing system (persistence) files

Many more like this.. All of them will need the ability to dynamically gather file contents and parse it.

joren485:
Last week, I was part of a volatility training where a `read_inode_content`-like function would have been useful. So, there is no specific plugin that I would like to add when this is merged, but as Andrew points out, there are many plugins that would benefit from such functionality.

If you feel like this change needs a better justification (i.e. is added when there is an actual plugin that uses it), it is also fine by me to just close this MR.

> Also the docstring is inaccurate "Read the contents of an inode into memory."
Fixed

atcuno:
@ikelos @gcmoreira it seems like the main concern is reading too much into memory? If so, the API could take an optional parameter of maximum size and cut off after. The files that would be auto parsed by plugins are pretty small - a 10MB cap would be more than enough of a limit.

ikelos:
I think it's more that it's being read with padding, rather than being read sparsely.  I don't know whether BytesIO supports a sparse mode (so that the mostly empty bytes aren't stored in memory), but otherwise constructing a data structure that yields a generator of the offsets and the needed bytes might better?  That would be consumable and effectively stream the data rather than send it in one big chunk?  I'm not a fan of arbitrary cut-off limits because they're not very flexible once they're hit and they create a bunch of boundary conditions that people then run into and complain about.

atcuno:
The idea is to have a wrapper function so every plugin doesn't have to do the gap reconstruction on its own as that removes repeated code and provides a consistent interface. With the max size being controlled by an optional parameter then plugin writers don't have anything complain about. They can jump bump the size if they want. Does this work?

ikelos:
Oh, I see.  Well as I mentioned I don't really have a strong opinion on it so if you think it's a reasonable solution and Gus doesn't have an objection then I'm fine adding it.  I guess it's up to @gcmoreira and you to sort out the best route forward?

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -420,8 +420,54 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
         ]
 
     @staticmethod
+    def read_inode_content(
+        inode: inodeClass,
+        buffer: IO,
+        page_size: int,
```
gcmoreira, line 426: The page_size should not be a configurable argument. Using a value different to the layer page_size will lead to incorrect results.

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -420,8 +420,54 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
         ]
 
     @staticmethod
+    def read_inode_content(
+        inode: inodeClass,
+        buffer: IO,
```
gcmoreira, line 425: You might want to choose a more appropriate parameter name, as the only user so far of this function isn’t actually a buffer; maybe "stream", "output_stream" or similar? Also, the method name seems misleading since it reads from the page cache and writes to an I/O stream. In my opinion, "write_inode_content_to_stream" would be a more accurate description of its functionality than "read_inode_content."

gcmoreira:
@atcuno for text files it will be ok. We could add that extra parameter if you think it's needed. I see that more useful for automatic processing of files, maybe?

Ideally users should call `linux.pagecache.Files` first to assess the number of inode pages and determine how many are cached.  Then before dumping that inode they should also see if the pages are safe to dump using the linux.pagecache.InodePages (of course, without --dump) and check the (DumpSafe) field. 
This way, they can evaluate if the content will add any value rather than dumping indiscriminately an useless stream of bytes. After doing that the user should be already aware of the file size.

Again, it doesn't hurt, so we can include that as well.

joren485:
Given the significant discussion, I think it is best to close this MR for now. I think someone, with a better understanding of Volatility3 than I, should implement this feature (if you decide you want it). Please feel free to use this MR in the future (or to ignore it completely), when you have a better idea of what the `linux.pagecache` API should look like.

ikelos:
@joren485 I think you were doing fine.  Sometimes PRs are where we flesh out how we want something to work, and that can take some time (and some initial disagreement).  Public back and forth can be difficult to deal with, but essentially I'd just leave the interested parties to figure out their differences, until a general agreement is made, and then implement any changes that fall out of that.

This looks like a valuable plugin, and we'd definitely like you to be the contributing author of the patches for the work but now that the branch has been deleted, it's going to be very difficult for others to work on the code.  Please consider resubmitting it as a new PR (referencing this one), and then let it sit there whilst the discussion continues and everyone figures out their differences and we can get it in a shape everyone's happy with and get it merged...

PR #1345, opened 2024-11-11
Environment variable for the Microsoft Symbol Server URL

mischw:
I am running Volatility in an environment where I provide already downloaded symbols using a local caching mirror of the symbol server. To make volatility look for my custom remote I need to be able to overwrite the URL it contacts. I solved this using a environment variable `VOL3_MSSYMBOLSERVER` which is preferred if set but stays with the default value if not. I figured some might do the same or might want to the same in the future so I thought I could open a pull request and see if you are interested.
I didn't find documentation about environment variables which are already used by Volatility. The name of the variable could be changed of course.

### DIFF
--- a/development/pdbparse-to-json.py
+++ b/development/pdbparse-to-json.py
@@ -26,7 +26,7 @@ class PDBRetreiver:
     def retreive_pdb(self, guid: str, file_name: str) -> Optional[str]:
         logger.info("Download PDB file...")
         file_name = ".".join(file_name.split(".")[:-1] + ['pdb'])
-        for sym_url in ['http://msdl.microsoft.com/download/symbols']:
+        for sym_url in [os.environ.get("VOL3_MSSYMBOLSERVER", "http://msdl.microsoft.com/download/symbols")]:
             url = sym_url + f"/{file_name}/{guid}/"
 
             result = None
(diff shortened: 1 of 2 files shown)


ikelos:
Thanks very much for the pull request, we don't tend to use environment variables for setting parameters, but you can pass the `-u URL` or `--remote-isf-url URL` to set the URL to use, and you can also override default parameters using the `vol.json` file (it's not well documented since the cli documentation doesn't end up in the final docs, but you can read a little about it at https://github.com/volatilityfoundation/volatility3/blob/develop/doc/source/vol-cli.rst#overriding-options).  As such I don't think this is the best way of achieving your goal, so I'm going to close this off, but feel free to reopen it if you feel there's a strong need for it (or if you want to improve the documentation or something?)

mischw:
I cant find much documentation about it but I think the `--remote-isf-url` parameter is for an entirely different feature? Looks like it requires a server which hosts a banners.json with already available symbol files.
In my PR I allow to change the default base url for downloading symbols from the Microsoft symbol server (`http://msdl.microsoft.com/download/symbols`) to something else (the mentioned caching mirror). Doesn't have to be implemented as an environment variable of course.

ikelos:
Oh, I see, sorry, yes, it is slightly different, however if you preprocess
the pdb files into ISF, you could use the banners solution.  I don't think
there's been sufficient interest in allowing local PDB servers to be used,
so I still don't think we'll be adding this feature, but thank you for your
contribution and interested parties will be able to find your patch here
and if there's an increase in interest then we can consider how best to add
it...

PR #1346, opened 2024-11-12
Linux - miscellaneous fixes

gcmoreira:
This PR addresses two bugs: one in a Page Cache log message and another that causes call recursion in the module object extension. Additionally, it enhances type annotations and refines log messages.

### DIFF
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -454,7 +454,7 @@ class InodePages(plugins.PluginInterface):
                     if current_fp + len(page_bytes) > inode_size:
                         vollog.error(
                             "Page out of file bounds: inode 0x%x, inode size %d, page index %d",
-                            inode.vol.object,
+                            inode.vol.offset,
                             inode_size,
                             page_idx,
                         )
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -454,7 +454,7 @@ def write_inode_content_to_file(
                     if current_fp + len(page_bytes) > inode_size:
                         vollog.error(
                             "Page out of file bounds: inode 0x%x, inode size %d, page index %d",
-                            inode.vol.object,
```
ikelos, line 457: Would've been nice to turn it into an f-string, but that's really a problem.

ikelos (approved):
Yep, look good, thanks!

volatility3/framework/plugins/linux/pagecache.py
gcmoreira: I love f'strings hehe but it's actually better not to use f'strings with the logger. That's because Python will parse and format all the strings, only for the logger to likely discard them if they don't match the appropriate log level, resulting in wasted processing time and resources. Ok, minimal, but every bit counts.

Some of the linters warns this: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html

PR #1347, opened 2024-11-12
Volshell: add regex_scan

eve-mem:
Hello :wave:

This PR adds a quailty of life feature to volshell allowing easy use of the regex scanner. I often find myself needing to scan for some bytes or a pattern. I can use a scanner with a loop to display the results but I do it so frequently I thought a little wrapper would be nice to have.

An optional size can be given for the displayed results as with the other fuctions (`db`, `db`, `dq`, etc).

You can of course specify a different layer name as well. 

Here are some examples:
```python
(volatility3) eve@xps:~/Documents/volatility3$ python3 volshell.py -f linux-sample-1.dmp -l
Volshell (Volatility 3 Framework) 2.11.0
Readline imported successfully  Stacking attempts finished                 

    Call help() to see available functions

    Volshell mode        : Linux
    Current Layer        : layer_name
    Current Symbol Table : symbol_table_name1
    Current Kernel Name  : kernel

(layer_name) >>> re(rb"swapper(\/0|\x00\x00)\x00\x00\x00\x00\x00\x00", 32)
0x88000160d3b8    73 77 61 70 70 65 72 2f 30 00 00 00 00 00 00 00    swapper/0.......
0x88000160d3c8    00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00    ................

0x88001d80eb00    73 77 61 70 70 65 72 2f 30 00 00 00 00 00 00 00    swapper/0.......
0x88001d80eb10    01 00 00 00 00 00 00 00 02 00 00 00 00 00 00 00    ................

0xffff8160d3b8    73 77 61 70 70 65 72 2f 30 00 00 00 00 00 00 00    swapper/0.......
0xffff8160d3c8    00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00    ................

(layer_name) >>> re(rb"swapper(\/0|\x00\x00)\x00\x00\x00\x00\x00\x00", 32, 'memory_layer')
0x160d3b8    73 77 61 70 70 65 72 2f 30 00 00 00 00 00 00 00    swapper/0.......
0x160d3c8    00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00    ................

0x1d80eb00    73 77 61 70 70 65 72 2f 30 00 00 00 00 00 00 00    swapper/0.......
0x1d80eb10    01 00 00 00 00 00 00 00 02 00 00 00 00 00 00 00    ................
```

Running the linux version scan:
```python
[...]

PR #1350, opened 2024-11-14
Volshell: Make the default number of bytes returned by db, dw, dd, dq, rx, and dis use a constant

eve-mem:
Hello :wave:

In this recent PR https://github.com/volatilityfoundation/volatility3/pull/1347 (Volshell: add regex_scan) we saw that all of the display commands all use the number of 128 bytes as the default number to display. 

This PR simply makes that a constant so that they can all be kept in sync. Perhaps overkill but I thought it might be a good idea. 

No offense taken if you feel this is unnecessary. 

Thanks!
:fox_face:

### DIFF
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -29,6 +29,8 @@ class Volshell(interfaces.plugins.PluginInterface):
 
     _required_framework_version = (2, 0, 0)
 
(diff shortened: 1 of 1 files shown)


ikelos (approved):
Yep, looks good.  I contemplated putting it into `constants`, but it's not really part of the framework, it's part of volshell, so it's in the right place.  5:)

PR #1352, opened 2024-11-14
Fix vmware layer so that files without a vmss or vmsn are still constructed with the warning is shown

eve-mem:
Hello :wave: 

User Super_Snorlax on slack raised an issue where a vmware sample failed to work with a `.vmem` extension but did with a `.raw` extension.

They've yet to report having tested this change but I suspect it will fix the issue. 

I believe this return None incorrectly stops some samples for working. The warning says `may be required` but the `return None` is meaning it will never be stacked.

:fox_face:


eve-mem:
Actually the discussion on the where the warning was added https://github.com/volatilityfoundation/volatility3/pull/1016 

It seems like the return None is okay. I'm changing this to draft until I hear from Super_Snorlax if this fixes the problem. (Or if I get my hands on a a flat vmware image type to test against)

eve-mem:
Super_Snorlax provided a sample and this did not help in the slightest. I'd got completely the wrong end of the stick.

PR #1354, opened 2024-11-15
New Linux plugin: fbdev graphics API

Abyss-W4tcher:
Hi 👋,

The Linux fbdev API allows user space software to interact with the framebuffer (memory representation of a pixel bitmap). Present since 2.1.109, it was basically appreciated for its simplicity and its ability to create text consoles or graphics outputs without a video adapter or an X server. 

Nowadays, it has mostly been replaced by the DRM subsystem, but a lot of drivers still provide backward compatibility for it, meaning that frames will still be stored in RAM/iomem !

The framebuffers can contain the terminal view (CTRL+ALT+FX), or any frame currently displayed by a software (e.g. video streaming) application. Here is a sample plugin run : 

```sh
$ python3 vol.py -f sample.bin -o out_dir/ linux.graphics.fbdev --dump
  |        Address | Device |             ID |    Size | Virtual resolution | BPP |   State |                           Filename
* | 0x981240641000 |    fb0 | vboxvideodrmfb | 1920000 |            800x600 |  32 | RUNNING | vboxvideodrmfb_800x600_32bpp.png
```
With its associated image:

vboxvideodrmfb_800x600_32bpp-1

From a very high user abstraction view, this is basically a "screenshot" plugin, but it is not magical in any way. 

Running this plugin against your old samples can sometimes give suprisingly accurate markers.

## Relevance

Captures of terminals and displays can provide interesting insights about the activity at capture time, and UI settings details. 

--- 

Additional runs : 

```sh
  |        Address | Device |        ID |    Size | Virtual resolution | BPP |   State |                     Filename
* | 0xaaf803801000 |    fb0 | svgadrmfb | 4163040 |           1176x885 |  32 | RUNNING | svgadrmfb_1176x885_32bpp.png
```

```sh
  |        Address | Device |        ID |    Size | Virtual resolution | BPP |   State |                    Filename
* | 0xc90000162000 |    fb0 | vboxdrmfb | 1920000 |            800x600 |  32 | RUNNING | vboxdrmfb_800x600_32bpp.png
```

--- 

This plugin is the first part of the `graphics` release, with a `DRM` plugin parser coming (hopefully) soon. However, it is more complex and needs a few things (and possibly more, regarding the recent BTF work) to move before even getting to the PR stage.


volatility3/framework/plugins/linux/graphics/fbdev.py
```diff
@@ -0,0 +1,314 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import logging
+import io
+
+# Image manipulation functions are kept in the plugin,
+# to prevent a general exit on missing PIL (pillow) dependency.
+from PIL import Image
```
ikelos: If PIL isn't installed, this needs to error gracefully.  We don't want the entire framework to require it just for this plugin, but we do want this plugin to error gracefully if it's not supported.  This should be done in a [similar way](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/check_syscall.py#L17) to capstone...

volatility3/framework/plugins/linux/graphics/fbdev.py
```diff
@@ -0,0 +1,314 @@
...
+        Returns:
+            The filename of the dumped buffer.
+        """
+        kernel = context.modules[kernel_name]
+        kernel_layer = context.layers[kernel.layer_name]
+        base_filename = f"{fb.id}_{fb.xres_virtual}x{fb.yres_virtual}_{fb.bpp}bpp"
+        if convert_to_image:
+            image = cls.convert_fb_raw_buffer_to_image(context, kernel_name, fb)
+            output = io.BytesIO()
+            image.save(output, image_format)
```
ikelos: Do we need to worry about callers of this method setting an invalid file format, or other PIL exceptions happening here?

volatility3/framework/plugins/linux/graphics/fbdev.py
```diff
@@ -0,0 +1,314 @@
...
+            file_handle = open_method(f"{base_filename}.{image_format.lower()}")
```
ikelos: I'd kinda prefer the file handle lines be unified and pulled out (so putting the data and filename into variables and then calling it once at the end).  Could we also please use a context manager (so 
```
with open_method(filename) as fp:
    fp.write(data)
```
and that then includes the close automatically and (I hope) handles closing the file in the case of an exception, etc)?

volatility3/framework/plugins/linux/graphics/fbdev.py
```diff
@@ -0,0 +1,314 @@
...
+
+        Documentation:
+            https://docs.kernel.org/fb/api.html:
+                - struct fb_fix_screeninfo stores device independent unchangeable information about the frame buffer device and the current format.
+                Those information can't be directly modified by applications, but can be changed by the driver when an application modifies the format.
+                - struct fb_var_screeninfo stores device independent changeable information about a frame buffer device, its current format and video mode,
+                as well as other miscellaneous parameters.
+        """
+        # NotAvailableValue() messes with the filename output on disk
+        id = utility.array_to_string(fb_info.fix.id) or "N-A"
```
ikelos: Is this going to be displayed to the user as a column?  If so, it should still be `NotAvailableValue` and then the code that sets the filename should check and do the replacement there...

volatility3/framework/plugins/linux/graphics/fbdev.py
```diff
@@ -0,0 +1,314 @@
...
+            if self.config["dump"]:
+                try:
+                    file_output = self.dump_fb(
+                        self.context, kernel_name, self.open, fb, bool(fb.color_fields)
+                    )
+                except exceptions.InvalidAddressException as excp:
+                    vollog.error(
+                        f'Layer {excp.layer_name} failed to read address {hex(excp.invalid_address)} when dumping framebuffer "{fb.id}".'
+                    )
+                    file_output = "Error"
```
ikelos: Might want to use a `BaseAbsentValue` derivative here?  I feel there should be one that covers this situation, it's not a showstopper though, I don't have a good reason for why Error isn't suitable (other than maybe localization?).

ikelos (commented):
First off, I just wanna say this looks awesome, and was a joy to review.  It's well commented, follows the style guide, really nice work all round!  There's just a couple of bits I'd like to see altered, but in general good work and I'm looking forward to getting this in the tree, thanks!  5:D

volatility3/framework/plugins/linux/graphics/fbdev.py
Abyss-W4tcher: Done, I had to remove PIL type hinting because it would not resolve unfortunately.

volatility3/framework/plugins/linux/graphics/fbdev.py
Abyss-W4tcher: I restricted it to png, as I don't see a reason to use multiple output formats for an RGB(A) format in the end.

volatility3/framework/plugins/linux/graphics/fbdev.py
Abyss-W4tcher: This is the current (manual) way of handling errors in file outputting across the framework, but indeed a `OutputFailure()` exception could be added somewhere in the renderers. I made sure to generate an error beforehand to inform the user on the failure reason, and not overload the `Filename` column. 

I can open an issue to track it and eventually update related bits once and for all when a solution comes out ?

volatility3/framework/plugins/linux/graphics/fbdev.py
ikelos: I'm ok with that.  You could possibly use the delayed typing (where you put the type name in a string), but I've no idea how that would work on a system without it (which we found killed plugin import for another plugin that tried to use capstone typing).

volatility3/framework/plugins/linux/graphics/fbdev.py
ikelos: Well, I don't necessarily want to throw an exception (which then stops the rest of the run), but I think the string "Error" isn't how we should indicate this having been an error, I think it should be a BaseAbsentValue derived value.  So either `Unreadable`, `Unparseable`, `NotAvailable` or we could invent a new one?

Abyss-W4tcher: Yes, I meant a new renderer not an exception sorry.

`UnreadableValue()` outputs a single dash (`-`) as well, which is fine if we want to rely on the error message solely ?

ikelos: It does in the CLI, but the point is that a web interface may display that return type differently.  If the plugin chooses that it's "Error", then it's Error for everyone, if the plugin returns it as a `BaseAbsentValue` then it might be able to do something better/cleverer that we haven't though of (and if not it can always just put a `-`).  I think the column data isn't the place to report errors, the error framework we've got is how that should be done, so yeah, I'd be fine just returning a BaseAbsentValue (or making a new one that we could provide an explanation with, but the CLI representation could just be "Error")?  Essentially, we should be programmatically providing the information about the error, rather than in a text representation...  Throwing an error should stop the program output anyway, so it might be academic?  5:P

ikelos: (As in, you might want to make it `vollog.warn` rather than `vollog.error`, since `error` will terminate the program afterwards)...  5;)

Abyss-W4tcher: Understandable, I'll make it an `UnreadableValue()`.

`vollog.error` does not exit the program, as it is just logging ? See this:

```sh
vollog.error("error")
vollog.info("still here")
```

I think it is more adapted in this case, as it is truly an error.

ikelos: Oh, I thought any time you logged an error it also caused a `sys.exit`?  Fair enough, perhaps I was thinking of something else?

ikelos: Sure, if it doesn't kill the program I'm fine with logging it as an error.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -483,6 +483,22 @@ def get_module_from_volobj_type(
 
         return kernel
 
+    @classmethod
+    def convert_fourcc_code(cls, code: int) -> str:
```
ikelos, line 487: I think adding a new method to `LinuxUtilities` is a MINOR bump please (and then updating the requirements of the plugin that needs this function).

Abyss-W4tcher: I've bumped the required numbers, it should be good right now.

ikelos (approved):
Looks good, nice to get this in main (a nice Christmas present!)...  5;D

PR #1356, opened 2024-11-18
Add flag --nofile to prevent a file from being written in the ConfigWriter plugin

JSCU-CNI:
The configwriter plugin automatically writes configuration files (duh!) to disk. If one just needs the output of the configwriter plugin, a file is automatically written. Hence we added the flag `--nofile` to prevent the configuration file from being written.

A better approach would have been to not have the configwriter plugin write a file by default and have a `-o` flag determine where the file would be written to. This is not backwards compatible, and has hence not been added. If you would like us to do that instead, we'd be glad to.

N.B.: this is our first PR! Glad to be part of the Volatility community. Feel free to give us (extensive) feedback on the contributions we make :)

### DIFF
--- a/volatility3/framework/plugins/configwriter.py
+++ b/volatility3/framework/plugins/configwriter.py
@@ -33,6 +33,12 @@ class ConfigWriter(plugins.PluginInterface):
                 default=False,
                 optional=True,
             ),
+            requirements.BooleanRequirement(
+                name="nofile",
+                description="Do not write file to output directory, only print to stdout",
+                default=False,
+                optional=True,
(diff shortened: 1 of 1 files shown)


ikelos:
Thanks very much for your contribution.  It feels like it would be more consistent to change it to a `--dump` option, to be in keeping with the rest of the plugins?  Since that would be a major shift in how the plugin works, it would be best to also bump the major version number or possibly rename the plugin to just `imageconfig` or something?  Then it could output the text as provided and if given the `--dump` configuration it could output the file as well?

PR #1359, opened 2024-11-18
Windows: Remove dep on vadyarascan from svcscan plugin

eve-mem:
Hello :wave: 

In https://github.com/volatilityfoundation/volatility3/pull/1353 we saw that the proposed vadregexscan plugin imported from vadyarascan to get a tasks vads - it seemed a little over kill and wasn't used by another other plugins. 

svcscan did use vadyarascan - this PR simply removes that dependency from svcscan. It doesn't change how the plugin works or its output. 

It functions just as it did before.  

Hope this makes sense. Thanks as always!
:fox_face:


ikelos (approved):
Yep, looks reasonable, thanks!

PR #1362, opened 2024-11-22
Follow XDG Base Directory Specification for cache directory

c0rydoras:
currently vol3 doesn't follow the [XDG BASE Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/)

> `$XDG_CACHE_HOME` defines the base directory relative to which user-specific non-essential data files should be stored. If `$XDG_CACHE_HOME` is either not set or empty, a default equal to `$HOME/.cache` should be used. 

As it is right now, whenever i use vol3 it creates a `~/.cache/volatility3` which isn't really ideal.

```bash
$ echo $XDG_CACHE_HOME
/home/arthurd/.local/var/cache

$ ls -a
.cert  .kube  .local  .mozilla  .sage  .ssh  .thunderbird  __all__

$ vol --help
[...]

$ ls -a
.cache .cert .kube .local .mozilla .sage .ssh .thunderbird __all__

$ ls .cache
volatility3
```

This PR fixes this.

```py
>>> import os
>>> os.path.join(os.path.expanduser("~"), ".cache", "volatility3")
'/home/arthurd/.cache/volatility3'
>>> os.path.join(os.environ.get("XDG_CACHE_HOME") or os.path.expanduser("~/.cache"), "volatility3")
'/home/arthurd/.local/var/cache/volatility3'
>>> os.environ["XDG_CACHE_HOME"] = ""
>>> os.path.join(os.environ.get("XDG_CACHE_HOME") or os.path.expanduser("~/.cache"), "volatility3")
'/home/arthurd/.cache/volatility3'
```

### DIFF
--- a/volatility3/framework/constants/__init__.py
+++ b/volatility3/framework/constants/__init__.py
@@ -68,7 +68,8 @@ LOGLEVEL_VVVV = 6
 
 
 CACHE_PATH = os.path.join(
-    os.environ.get("XDG_CACHE_HOME") or os.path.expanduser("~/.cache"), "volatility3"
+    os.environ.get("XDG_CACHE_HOME") or os.path.join(os.path.expanduser("~"), ".cache"),
+    "volatility3",
 )
 """Default path to store cached data"""
 
--- a/volatility3/framework/constants/__init__.py
+++ b/volatility3/framework/constants/__init__.py
@@ -6,6 +6,7 @@
 Stores all the constant values that are generally fixed throughout
 volatility This includes default scanning block sizes, etc.
 """
+
 import enum
 import os.path
 import sys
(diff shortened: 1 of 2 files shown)


volatility3/framework/constants/__init__.py
```diff
@@ -65,7 +66,10 @@
 LOGLEVEL_VVVV = 6
 """Logging level for four levels of detail: -vvvvvv"""
 
-CACHE_PATH = os.path.join(os.path.expanduser("~"), ".cache", "volatility3")
+
+CACHE_PATH = os.path.join(
+    os.environ.get("XDG_CACHE_HOME") or os.path.expanduser("~/.cache"), "volatility3"
```
ikelos: Just want to verify this will work on windows?  Usually we don't include the slash, but use `os.path.join` to ensure it will work on all operating systems...

ikelos (commented):
Thanks very much, looks good and happy to follow the XDG spec, but just want to make sure this won't impact the ability to run volatility on any operating system...

volatility3/framework/constants/__init__.py
c0rydoras: should be fixed here: https://github.com/volatilityfoundation/volatility3/pull/1362/commits/a2b96c0dc0b2989e87e4eb4a9c2a274ef4ed6e7f

even tho i'm not sure that it would even matter because of this:
```py
if sys.platform == "win32":
    CACHE_PATH = os.path.realpath(
        os.path.join(os.environ.get("APPDATA", os.path.expanduser("~")), "volatility3")
    )
```

ikelos: Ah true, I'd forgotten about that.  Still think it's better to let python handle the slashes.  Thanks for the quick fix up!  5:)

PR #1363, opened 2024-11-22
Core: Add in generic cache-manager chooser function

ikelos:
We've got a nice interface to use a generic cache manager, better to use that than require bits of the framework to specifically pull in our cache manager.

### DIFF
--- a/volatility3/framework/automagic/symbol_cache.py
+++ b/volatility3/framework/automagic/symbol_cache.py
@@ -492,6 +492,21 @@ class SqliteCache(CacheManagerInterface):
         return output
 
 
+def load_cache_manager(cache_file: Optional[str] = None) -> CacheManagerInterface:
+    """Loads a cache manager based on a specific cache file"""
+    if cache_file is None:
+        cache_file = os.path.join(constants.CACHE_PATH, constants.IDENTIFIERS_FILENAME)
+    # Different implementations of cache
+    if not os.path.exists(cache_file):
+        raise ValueError("Non-existant cache file provided")
+    with open(cache_file, "rb") as fp:
+        header = fp.read(4)
+        if header not in [b"SQLi"]:
+            raise ValueError("Identifier file not in recognized format")
+    # Currently only one choice, so use that
+    return SqliteCache(cache_file)
+
+
 ### Automagic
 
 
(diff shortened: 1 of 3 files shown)


ikelos:
I don't think this is controversial.  If it passes all the tests, I'll merge it in a couple days.

ikelos:
Hmm, we actually want this to start a new cache if the file it's given doesn't exist.  I guess we just need to check whether the path provided could be valid?

volatility3/framework/automagic/symbol_cache.py
```diff
@@ -492,6 +492,21 @@ def get_identifiers(self, operating_system: Optional[str]) -> List[bytes]:
...
+def load_cache_manager(cache_file: Optional[str] = None) -> CacheManagerInterface:
+    """Loads a cache manager based on a specific cache file"""
+    if cache_file is None:
+        cache_file = os.path.join(constants.CACHE_PATH, constants.IDENTIFIERS_FILENAME)
+    # Different implementations of cache
+    if not os.path.exists(cache_file):
+        raise ValueError("Non-existant cache file provided")
+    with open(cache_file, "rb") as fp:
+        header = fp.read(4)
+        if header not in [b"SQLi"]:
```
gcmoreira, line 504: It's reading 4 bytes and comparing against 4 bytes, not sure why the `header not in [b"SQLi"]`? shouldn't it be `header != b"SQLi"`? or one-liner `open(cache_file, "rb").read(4) != b"SQLi"`?

gcmoreira:
Hey @ikelos sorry about this review request, it got merged so quickly

ikelos:
No problem, sorry I wanted to act on it while it was still fresh.  Hopefully it appears problem free?

gcmoreira:
yeah, only that sqli header check but anyway a minor stuff

volatility3/framework/automagic/symbol_cache.py
ikelos: This is all for future extensibility, when we get different file formats for the cache.  The 4 byte header check will likely be used for whatever format we end up with, so I coded it with essentially a case statement with only one clause in it...

ikelos:
Yeah, we can make it a bit longer, but I thought better in case we make a cache file that's of a different format in the future.

PR #1364, opened 2024-11-22
use ruff for linting 

c0rydoras:
right now the CI should fail because some rules can't safely be fixed automatically

before I start fixing/ignoring things to make the CI pass, I'd like to know if linting like this is even something you (the maintainers) would want

(if this were ever to land it would make sense to add the commits to a `.git-blame-ignore-revs`; see https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)

### DIFF
--- a/development/schema_validate.py
+++ b/development/schema_validate.py
@@ -1,12 +1,11 @@
 import argparse
 import json
+import logging
 import os
 import sys
 
 # TODO: Rather nasty hack, when volatility's actually installed this would be unnecessary
 sys.path += ".."
-
-import logging  # noqa: E402
 
 console = logging.StreamHandler()
 console.setLevel(logging.DEBUG)
--- a/development/schema_validate.py
+++ b/development/schema_validate.py
@@ -9,7 +9,7 @@ sys.path += ".."
 
 console = logging.StreamHandler()
 console.setLevel(logging.DEBUG)
-formatter = logging.Formatter('%(levelname)-8s %(name)-12s: %(message)s')
+formatter = logging.Formatter("%(levelname)-8s %(name)-12s: %(message)s")
 console.setFormatter(formatter)
 
 logger = logging.getLogger("")
@@ -18,10 +18,10 @@ logger.setLevel(logging.DEBUG)
 
 from volatility3 import schemas  # noqa: E402
 
-if __name__ == '__main__':
+if __name__ == "__main__":
     parser = argparse.ArgumentParser("Validates ")
-    parser.add_argument("-s", "--schema", dest = "schema", default = None)
-    parser.add_argument("filenames", metavar = "FILE", nargs = '+')
+    parser.add_argument("-s", "--schema", dest="schema", default=None)
+    parser.add_argument("filenames", metavar="FILE", nargs="+")
 
     args = parser.parse_args()
 
--- a/development/schema_validate.py
+++ b/development/schema_validate.py
@@ -28,7 +28,7 @@ if __name__ == '__main__':
 
     schema = None
     if args.schema:
-        with open(os.path.abspath(args.schema), 'r') as s:
+        with open(os.path.abspath(args.schema)) as s:
             schema = json.load(s)
 
     failures = []
@@ -36,7 +36,7 @@ if __name__ == '__main__':
         try:
             if os.path.exists(filename):
                 print(f"[?] Validating file: {filename}")
-                with open(filename, 'r') as t:
+                with open(filename) as t:
                     test = json.load(t)
 
                 if args.schema:
--- a/development/schema_validate.py
+++ b/development/schema_validate.py
@@ -6,7 +6,7 @@ import sys
 # TODO: Rather nasty hack, when volatility's actually installed this would be unnecessary
 sys.path += ".."
 
-import logging
+import logging  # noqa: E402
 
 console = logging.StreamHandler()
 console.setLevel(logging.DEBUG)
@@ -17,7 +17,7 @@ logger = logging.getLogger("")
 logger.addHandler(console)
 logger.setLevel(logging.DEBUG)
 
-from volatility3 import schemas
+from volatility3 import schemas  # noqa: E402
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser("Validates ")
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -19,6 +19,8 @@ import sys
 
 import sphinx.ext.apidoc
 
+from importlib.util import find_spec
+
 
 def setup(app):
     volatility_directory = os.path.abspath(
@@ -124,7 +126,7 @@ def setup(app):
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath("../.."))
 
-from volatility3.framework import constants
+from volatility3.framework import constants  # noqa: E402
 
 # -- General configuration ------------------------------------------------
 
@@ -147,13 +149,9 @@ extensions = [
 
 autosectionlabel_prefix_document = True
 
-try:
-    import sphinx_autodoc_typehints
-
+if find_spec("sphinx_autodoc_typehints") is not None:
     extensions.append("sphinx_autodoc_typehints")
-except ImportError:
-    # If the autodoc typehints extension isn't available, carry on regardless
-    pass
+# If the autodoc typehints extension isn't available, carry on regardless
 
 # Add any paths that contain templates here, relative to this directory.
 # templates_path = ['tools/templates']
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -5,15 +5,6 @@
 # Check the python version to ensure it's suitable
 import glob
 import sys
-import zipfile
-import importlib
-import inspect
-import logging
-import os
-import traceback
-from typing import Any, Dict, Generator, List, Tuple, Type, TypeVar
-
-from volatility3.framework import constants, interfaces
 
 required_python_version = (3, 8, 0)
 if (
@@ -25,10 +16,18 @@ if (
     )
 ):
     raise RuntimeError(
-        "Volatility framework requires python version {}.{}.{} or greater".format(
-            *required_python_version
-        )
+        f"Volatility framework requires python version {'.'.join(map(str, required_python_version))} or greater"
     )
+else:
+    import zipfile
+    import importlib
+    import inspect
+    import logging
+    import os
+    import traceback
+    from typing import Any, Dict, Generator, List, Tuple, Type, TypeVar
+
+    from volatility3.framework import constants, interfaces
 
 
 # ##
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -97,8 +97,7 @@ def class_subclasses(cls: Type[T]) -> Generator[Type[T], None, None]:
         # The typing system is not clever enough to realize that clazz has a hidden attr after the hasattr check
         if not hasattr(clazz, "hidden") or not clazz.hidden:  # type: ignore
             yield clazz
-        for return_value in class_subclasses(clazz):
-            yield return_value
+        yield from class_subclasses(clazz)
 
 
 def import_files(base_module, ignore_errors: bool = False) -> List[str]:
@@ -159,9 +158,9 @@ def import_files(base_module, ignore_errors: bool = False) -> List[str]:
 
 def _filter_files(filename: str):
     """Ensures that a filename traversed is an importable python file"""
-    return (
-        filename.endswith(".py") or filename.endswith(".pyc")
-    ) and not filename.startswith("__")
+    return (filename.endswith((".py", ".pyc"))) and not filename.startswith(
+        "__"
+    )
 
 
 def import_file(module: str, path: str, ignore_errors: bool = False) -> List[str]:
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -5,29 +5,16 @@
 # Check the python version to ensure it's suitable
 import glob
 import sys
+import volatility3.framework.check_python_version  # noqa: F401
+import zipfile
+import importlib
+import inspect
+import logging
+import os
+import traceback
+from typing import Any, Dict, Generator, List, Tuple, Type, TypeVar
 
-required_python_version = (3, 8, 0)
-if (
-    sys.version_info.major != required_python_version[0]
-    or sys.version_info.minor < required_python_version[1]
-    or (
-        sys.version_info.minor == required_python_version[1]
-        and sys.version_info.micro < required_python_version[2]
-    )
-):
-    raise RuntimeError(
-        f"Volatility framework requires python version {'.'.join(map(str, required_python_version))} or greater"
-    )
-else:
-    import zipfile
-    import importlib
-    import inspect
-    import logging
-    import os
-    import traceback
-    from typing import Any, Dict, Generator, List, Tuple, Type, TypeVar
-
-    from volatility3.framework import constants, interfaces
+from volatility3.framework import constants, interfaces
 
 
 # ##
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -56,9 +56,7 @@ def require_interface_version(*args) -> None:
     if len(args):
         if args[0] != interface_version()[0]:
             raise RuntimeError(
-                "Framework interface version {} is incompatible with required version {}".format(
-                    interface_version()[0], args[0]
-                )
+                f"Framework interface version {interface_version()[0]} is incompatible with required version {args[0]}"
             )
         if len(args) > 1:
             if args[1] > interface_version()[1]:
@@ -70,7 +68,7 @@ def require_interface_version(*args) -> None:
                 )
 
 
-class NonInheritable(object):
+class NonInheritable:
     def __init__(self, value: Any, cls: Type) -> None:
         self.default_value = value
         self.cls = cls
@@ -187,9 +185,7 @@ def import_file(module: str, path: str, ignore_errors: bool = False) -> List[str
                     traceback.TracebackException.from_exception(e).format(chain=True)
                 )
             )
-            vollog.debug(
-                "Failed to import module {} based on file: {}".format(module, path)
-            )
+            vollog.debug(f"Failed to import module {module} based on file: {path}")
             failures.append(module)
             if not ignore_errors:
                 raise
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -5,7 +5,7 @@
 # Check the python version to ensure it's suitable
 import glob
 import sys
-import volatility3.framework.check_python_version  # noqa: F401
+from volatility3.framework import check_python_version as check_python_version
 import zipfile
 import importlib
 import inspect
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -22,14 +22,14 @@ if (
         )
     )
 
-import importlib
-import inspect
-import logging
-import os
-import traceback
-from typing import Any, Dict, Generator, List, Tuple, Type, TypeVar
+import importlib  # noqa: E402
+import inspect  # noqa: E402
+import logging  # noqa: E402
+import os  # noqa: E402
+import traceback  # noqa: E402
+from typing import Any, Dict, Generator, List, Tuple, Type, TypeVar  # noqa: E402
 
-from volatility3.framework import constants, interfaces
+from volatility3.framework import constants, interfaces  # noqa: E402
 
 
 # ##
@@ -74,7 +74,7 @@ class NonInheritable:
         self.cls = cls
 
     def __get__(self, obj: Any, get_type: Type = None) -> Any:
-        if type == self.cls:
+        if type is self.cls:
             if hasattr(self.default_value, "__get__"):
                 return self.default_value.__get__(obj, get_type)
             return self.default_value
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -6,6 +6,14 @@
 import glob
 import sys
 import zipfile
+import importlib
+import inspect
+import logging
+import os
+import traceback
+from typing import Any, Dict, Generator, List, Tuple, Type, TypeVar
+
+from volatility3.framework import constants, interfaces
 
 required_python_version = (3, 8, 0)
 if (
@@ -21,15 +29,6 @@ if (
             *required_python_version
         )
     )
-
-import importlib  # noqa: E402
-import inspect  # noqa: E402
-import logging  # noqa: E402
-import os  # noqa: E402
-import traceback  # noqa: E402
-from typing import Any, Dict, Generator, List, Tuple, Type, TypeVar  # noqa: E402
-
-from volatility3.framework import constants, interfaces  # noqa: E402
 
 
 # ##
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -143,9 +143,7 @@ def import_files(base_module, ignore_errors: bool = False) -> List[str]:
 
 def _filter_files(filename: str):
     """Ensures that a filename traversed is an importable python file"""
-    return (filename.endswith((".py", ".pyc"))) and not filename.startswith(
-        "__"
-    )
+    return (filename.endswith((".py", ".pyc"))) and not filename.startswith("__")
 
 
 def import_file(module: str, path: str, ignore_errors: bool = False) -> List[str]:
--- a/volatility3/framework/configuration/__init__.py
+++ b/volatility3/framework/configuration/__init__.py
@@ -3,3 +3,5 @@
 #
 
 from volatility3.framework.configuration import requirements
+
+__all__ = ["requirements"]
--- a/volatility3/framework/configuration/__init__.py
+++ b/volatility3/framework/configuration/__init__.py
@@ -2,6 +2,4 @@
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
 
-from volatility3.framework.configuration import requirements
-
-__all__ = ["requirements"]
+from volatility3.framework.configuration import requirements as requirements
--- a/volatility3/framework/constants/__init__.py
+++ b/volatility3/framework/constants/__init__.py
@@ -14,7 +14,7 @@ import warnings
 from typing import Callable, Optional
 
 import volatility3.framework.constants.linux
-import volatility3.framework.constants.windows
+import volatility3.framework.constants.windows  # noqa: F401
 from volatility3.framework.constants._version import (
     PACKAGE_VERSION,
     VERSION_MAJOR,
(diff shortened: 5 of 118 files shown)


ikelos:
Can you explain the pros (and cons) of using ruff over black please?

c0rydoras:
[ruff](https://docs.astral.sh/ruff/) is more than black, it **has** a [formatter](https://docs.astral.sh/ruff/formatter/) but it started as a [linter](https://docs.astral.sh/ruff/linter/), it (the formatter) is faster than black and more [configurable](https://docs.astral.sh/ruff/formatter/#configuration), it can also format [docstrings](https://docs.astral.sh/ruff/formatter/#docstring-formatting)

its [linter](https://docs.astral.sh/ruff/linter/) is not only faster than other linting tools (e.g. flake8, pylint etc.) but it also supports a lot of those linters (e.g. https://docs.astral.sh/ruff/rules/#flake8-builtins-a, https://docs.astral.sh/ruff/rules/#pylint-pl) (and a lot more rules in general)

the only real change here is that there's faster formatting (it might deviate a bit from black, see https://docs.astral.sh/ruff/formatter/black/) and just one dependency (`ruff`) for linting and formatting instead of 2 (`ruff` for linting and `black` for formatting)

ikelos:
Thanks, I'm a little suspicious that there weren't any downsides to using ruff, that's why I asked for pros and cons.  I've run some tests and there's 9 files that black and ruff can't agree on how to format.  I also kinda liked black in that it *wasn't* configurable, and so you wouldn't get bickering about people wanting the formatting done a particular way.  Would it be possible to add in ruff just as a linter initially, get those issues fixed up if possible (I prefer to fix rather than ignore, because it's either a bad rule that can be ignored anywhere, or it's a good rule that should always be followed) and then consider shifting the formatter over to ruff from black once the dust has cleared?

c0rydoras:
ruff is now only used for linting

ikelos:
Thanks very much, I'm generally happy with the changes it's made.  There's a couple places where it's decided not including defaults is better than being explicit, which I can get used to.  Would it be possible to make the suggested changes so get rid of the errors that ruff is reporting in the code base, or would that need to be done in a second commit (just worried this'll be importing a new test that then immediately fails, and I'd prefer we knew it wouldn't impede new commits).

Deciding on the ignore list will be tricky...

I think the module imports can move up, and I understand the reasoning behind not using f-strings in argparse, so I think we could manually make the necessary change without breaking anything.  Would you be up for making those please?

development/schema_validate.py
```diff
@@ -6,7 +6,7 @@
 # TODO: Rather nasty hack, when volatility's actually installed this would be unnecessary
 sys.path += ".."
 
-import logging
+import logging  # noqa: E402
```
ikelos: I'd move this import up?  I don't *think* it need to be below the `sys.path` modification, that's just for importing volatility3 below (which will sadly still need the noqa).  5:S

doc/source/conf.py
```diff
@@ -147,13 +149,9 @@ def setup(app):
 
 autosectionlabel_prefix_document = True
 
-try:
-    import sphinx_autodoc_typehints
-
+if find_spec("sphinx_autodoc_typehints") is not None:
```
ikelos, line 152: Cool?  I thought try/import/except was ok, or was it complaining about the pass?  We'd usually do a vollog there to say that the import wasn't found but also wasn't critical...

volatility3/framework/__init__.py
```diff
@@ -22,14 +22,14 @@
...
     )
 
-import importlib
-import inspect
-import logging
-import os
-import traceback
-from typing import Any, Dict, Generator, List, Tuple, Type, TypeVar
+import importlib  # noqa: E402
+import inspect  # noqa: E402
```
ikelos: Really don't like noqaing that many things.  We could put that into a `check_python_version` module instead, and just import it in the same place?  I don't think it affects any of the other imports, other than firing first?

volatility3/framework/configuration/__init__.py
```diff
@@ -3,3 +3,5 @@
 #
 
 from volatility3.framework.configuration import requirements
+
+__all__ = ["requirements"]
```
ikelos: I'm wary about wanting to add a new file and then forgetting it need adding here too.  What was this specific rule that flagged this?

volatility3/framework/constants/__init__.py
```diff
@@ -141,3 +141,36 @@ def __getattr__(name):
         return globals()[f"{deprecated_tag}{name}"]
 
     return getattr(__import__(__name__), name)
+
+
+__all__ = [
```
ikelos: Yep, again, not happy doing this, it's a sure fire way to forget to add something in a hurry.

volatility3/framework/interfaces/__init__.py
```diff
@@ -22,3 +22,14 @@
     symbols,
     automagic,
 )
+
+__all__ = [
```
ikelos: Nor this

volatility3/framework/layers/resources.py
```diff
@@ -29,7 +29,7 @@
 
 try:
     # Import so that the handler is found by the framework.class_subclasses callc
-    import smb.SMBHandler  # lgtm [py/unused-import]
+    import smb.SMBHandler  # lgtm [py/unused-import]  # noqa: F401
```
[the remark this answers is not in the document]
ikelos: Ugh, gonna get lots of these exceptions...  5:\

c0rydoras:
> Deciding on the ignore list will be tricky...

right now its configured to only use a few rules, there might be other ones we want to enable: https://docs.astral.sh/ruff/rules/

volatility3/framework/plugins/mac/kauth_scopes.py
```diff
@@ -80,7 +80,7 @@ def _generator(self):
                 (
                     identifier,
                     format_hints.Hex(scope.ks_idata),
-                    len([l for l in scope.get_listeners()]),
+                    len([li for li in scope.get_listeners()]),
```
ikelos: Huh, good catch, I generally keep an eye out for single letter variables myself!  Might prefer listener, but you've fixed up so much I'm not gonna be that picky.  5;)

volatility3/framework/plugins/windows/netscan.py
```diff
@@ -161,7 +161,7 @@ def determine_tcpip_version(
             raise NotImplementedError(
                 "Kernel Debug Structure version format not supported!"
             )
-        except:
+        except:  # noqa: E722
```
ikelos: This is exactly the type of thing *not* to ignore.  It should be fixed ideally.  I can ask the guy who wrote this what he was expecting to catch here...

volatility3/framework/plugins/windows/shimcachemem.py
```diff
@@ -146,7 +146,7 @@ def find_shimcache_win_xp(
             context, layer_name, kernel_symbol_table
         ):
             pid = process.UniqueProcessId
-            vollog.debug("checking process %d" % pid)
+            vollog.debug("checking process %d", pid)
```
ikelos, line 149: Errr, will that fill in the `%d` or will it just have `%d <pid>` in the text?

ikelos (commented):
Thanks very much for making those recent changes, there's a few bits I'd prefer to tweak before we introduce them though (sorry to be picky, but this will set the way we're picky for all future commits, so best to get it right now if possible)...  5:)

ikelos:
Yeah, I'd prefer as few exceptions as possible.  I'd even be ok with the lambda assignment one coming back in (assuming it wasn't difficult to fix the places it was complaining about)?  Just trying to ensure we don't get time wasted on tweak or fixing than we do on agreeing and writing code that adheres to our standard...

volatility3/framework/configuration/__init__.py
c0rydoras: https://docs.astral.sh/ruff/rules/unused-import/

AFAIK specifying `__all__` is also a best practice

volatility3/framework/plugins/windows/shimcachemem.py
c0rydoras: AFAIK this works

```python
>>> import logging
>>> logger = logging.getLogger(__name__)
>>> logger.warning("example %d", 12345)
example 12345
```
pretty sure .debug works the same

c0rydoras:
ruff can fix the lambda one by itself, I just thought it looked uglier in the diff

volatility3/framework/configuration/__init__.py
ikelos: Sure, but there's a few places where we expect the resolver to pick them up based on the file structure.  If we include `__all__` then that resolver never gets involved and bits can fall apart (which is why I don't want to set it as a precedent everywhere).

volatility3/framework/plugins/mac/kauth_scopes.py
c0rydoras: its `listener` now :)

volatility3/framework/plugins/windows/shimcachemem.py
ikelos: Ok, cool, happy with it then, thanks for testing it.  5:)

volatility3/framework/configuration/__init__.py
c0rydoras: what's the resolver? :sweat_smile:

doc/source/conf.py
c0rydoras: it was complaining about an unused import :)

volatility3/framework/layers/resources.py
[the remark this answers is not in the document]
ikelos: Oooh, it looks like you can add `import smb.SMBHandler as smb.SMBHandler` as another way of telling it not to complain (according to https://docs.astral.sh/ruff/rules/unused-import/ that you sent me).  I'd definitely prefer that when importing windows (assuming as allows you to include `.`)...

volatility3/framework/plugins/windows/netscan.py
c0rydoras: i would've fixed it if i would know the error(s) its catching here :sweat_smile:

volatility3/framework/layers/resources.py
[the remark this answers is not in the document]
c0rydoras: where does it get used? AFAICT this import does nothing

doc/source/conf.py
ikelos: Oh, I think a fix for that is to include `import thing as thing` and then it won't complain?  I think I'd prefer that than this which is just a little more cognitive load for people that aren't familiar with it...

volatility3/framework/layers/resources.py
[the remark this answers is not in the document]
c0rydoras: also `import something as some_module.SomeName` isn't possible, closest thing would be `import smb.SMBHandler as SMBHandler` :sweat_smile:

development/schema_validate.py
c0rydoras: fixed here: https://github.com/volatilityfoundation/volatility3/pull/1364/commits/436cc30fbc9dba31f988a0c786764cbb369a2d70

volatility3/framework/configuration/__init__.py
ikelos: Just if you ask python to `import blah.foo` it'll look up foo as a file under `blah` and try to load it, unless blah has an `__all__` that doesn't include it?

volatility3/framework/plugins/isfinfo.py
```diff
@@ -96,16 +97,12 @@ def _generator(self):
                     if filter_item in isf_file:
                         filtered_list.append(isf_file)
 
-        try:
-            import jsonschema
-
-            if not self.config["validate"]:
```
ikelos, line 102: This is actually an important bit of code, not sure why it took that out there...

ikelos: Oh, I see, it moved it up into the if statement

volatility3/framework/plugins/mac/kauth_scopes.py
```diff
@@ -80,7 +80,7 @@ def _generator(self):
                 (
                     identifier,
                     format_hints.Hex(scope.ks_idata),
-                    len([l for l in scope.get_listeners()]),
+                    len([listener for listener in scope.get_listeners()]),
```
ikelos, line 83: Hehehe, thanks!  5:D

volatility3/framework/plugins/isfinfo.py
c0rydoras: its not taken out, because the import isn't used it uses  `find_spec` from `importlib.util`

volatility3/framework/configuration/__init__.py
c0rydoras: i don't really know, AFAIK any `import foo.bar` is an anti-pattern and should be rewritten to `import foo.bar as bar` (or something like that) 

what I know is that what you specify in `__all__` is what gets exported when using `*` imports

volatility3/framework/constants/__init__.py
```diff
@@ -14,7 +14,7 @@
 from typing import Callable, Optional
 
 import volatility3.framework.constants.linux
-import volatility3.framework.constants.windows
+import volatility3.framework.constants.windows  # noqa: F401
```
ikelos: Hmmm, I'd really like that `import foo as foo` trick to work here instead of the noqa line.  Do we think it will?

doc/source/conf.py
c0rydoras: you'd still have an unused import tho and (imo) try/catch around imports is ugly

volatility3/framework/plugins/isfinfo.py
ikelos: Yeah, it was the if statement I was objecting to losing, but actually it moved it into the if statement and it all works out the same.  5:)

volatility3/framework/__init__.py
c0rydoras: I moved them up, AFAICT these would only throw/raise if you were to use something older than python 3.1

https://github.com/volatilityfoundation/volatility3/pull/1364/commits/39c6aa56511deeded3057a25ec0f733964e32721

volatility3/framework/constants/__init__.py
c0rydoras: `import as ...` might even break it? (never tried it but when importing it `as something` i'd expect the constants to then be on `something` e.g. `something.CONSTANT` and not in the scope of the file)

--edit: doesn't break it, and it is then under `something`, i misunderstood how this was used :sweat_smile:

c0rydoras: AFAICT and imo the cleanest way would probably be adding a `__all__` to those 2 files and then doing an `import * from ...`

but that also doesn't really seem ideal

volatility3/framework/__init__.py
c0rydoras: its now in its own module: https://github.com/volatilityfoundation/volatility3/pull/1364/commits/a2f71bf49161992bdfed0bdbd7c2d00affd83e6b

volatility3/framework/__init__.py
```diff
@@ -5,23 +5,8 @@
 # Check the python version to ensure it's suitable
 import glob
 import sys
+import volatility3.framework.check_python_version  # noqa: F401
```
ikelos: Would the `from volatility3.framework import check_python_version as check_python_versdion` trick work here to avoid the noqa directive?

volatility3/framework/configuration/__init__.py
ikelos: Hmmm, I'm happy with always using `as` to reduce module name length (although we have some modules named the same, so `context` and `interfaces.context`, but I can live with `interfaces_context` for clarity.  We also work hard to prevent people doing `import *` of anything, but adding a new file to the `__all__` list feels like an extra step we might forget.  I'd go so far as to ok that rule being added to the ignore list across the whole codebase, as long as it wouldn't allow anything else through?

volatility3/framework/constants/__init__.py
ikelos: Yeah, I've worked hard to keep `import *` out of the codebase, I'm not going to give people an example.  I assume then that changing this to `from volatility3.framework.constants import windows as windows` to the end, didn't resolve the issue?  The page you'd link earlier seemed to imply it would?

volatility3/framework/layers/resources.py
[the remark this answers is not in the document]
ikelos: The import registers a handler gloablly which then means `open('smb://blah')` magically just works...  5:S
How about `from smb import SMBHandler as SMBHandler`?  Does that work?

volatility3/framework/plugins/windows/netscan.py
ikelos: Yeah, sadly I don't know either (and that's why I wish the original author had put the list in here).  I guess either `Exception` (which the linter may complain about separately) or Just `NotImplementedError` and the classic `InvalidAddressException` since those the the most common one?  Then add a `FIXME` to tell devs to add any additional exceptions that are hit?

doc/source/conf.py
ikelos: Yeah, ok, it's not a showstopper and I guess we'll get used to it...  5;P

ikelos (commented):
Cool, this is looking good!  Thanks for all the effort you've put in!  I think the only last little bits to try and resolve are the imports with side-effects (`from thing import blah as blah`) and disabling the `__all__` complaint (as long as it doesn't disable anything else useful).

Lastly, I just need to run this by other members of the core team, since it may trip them up more than me these days.  So @atcuno @iMHLv2 @gcmoreira @npetroni any views or complaints about the changes in this PR (including linting all new code with ruff).  We might move to ruff for the formatting too, but for now I just want to catch coding/style errors where we can...

volatility3/framework/__init__.py
c0rydoras: it would probably work but imo the noqa is more readable, if i just saw that i'd ask myself, why import that as that

volatility3/framework/constants/__init__.py
c0rydoras: i never actually tried it

ikelos:
The CodeQL error we can fix up by moving the skip check up inside the exception handler just above it (ie, if there's no `basename`, it should still keep going, which seemed to be the intention of the original code.

Thank you so much for putting all the effort in to get this where it is!  I realize there's been a lot of nitpicking, but I really appreciate you having taken the time to work through and fix them all, it's much appreciated!

volatility3/framework/configuration/__init__.py
c0rydoras: the problem i see is that catching unused imports is well an important thing, ingoring it in all `__init__.py`s is what i'd suggest if the `__all__` stuff isn't possible/wanted

c0rydoras:
what does this check do?

```py
            if self.config["name"] and self.config["name"] not in BaseDllName:
                continue
```

AFAICT this just always continues if `self.config["name"]`

volatility3/framework/configuration/__init__.py
ikelos: I don't think that's a significant problem?  The style guide ensures that most imports are specific (down to naming the classes to import) and where we do import a full module without specifying the individual submodules we want, it's probably because we want everything (or we don't know yet what we're going to want).  In either care, we'll likely end up asking for everything, and if we then *don't* get it because someone forgot to include the new file in a variable, it causes friction.  It feels almost like duplication, in that you're explicitly stating the modules, as well as putting them in the directory?  Shouldn't putting them in the directory be enough to have them included?

volatility3/framework/__init__.py
ikelos: Yeah, I take your point.  Perhaps (in either case) we should add a comment explaining that the import has side-effects, either to run some code directly, or to ensure classes are registered so they show up as part of the plugin mechanism?  I just hate seeing code littered with exceptions because it makes people think the error checker can just be turned off rather than the problem fixed.  The fewer of those in the codebase, the better in my view (which is why a legit, but odd looking line is better than one we explicitly turn the error checker off for).

ikelos:
Because of the and, it only skips it if there's a name in the config that isn't the name mentioned in `BaseDllName`, so it's a filtration to only select unnamed things or the one that's specifically named.

ikelos:
If there's no name, everything gets output, if there's no BaseDLLName, it should get included (I believe) and then only when both the filter is set and the filter condition is met, are the items filtered.

volatility3/framework/__init__.py
c0rydoras: usually i'd opt for no side effects at all (and enabling all linting rules and ignoring single ones) but this doesn't really seem possible here (size of the codebase, the way its built right now etc.) i'm not quite sure what to do here.

additionally the `F401` isn't fixed by a
```py
import volatility3.framework.check_python_version as check_python_version
```

if i understand the rule correctly (and based on some testing) its `from .. import foo as foo` that counts as "redundant import"

c0rydoras: (seems like i was wrong)

volatility3/framework/constants/__init__.py
c0rydoras: `from volatility3.framework.constants import windows as windows` seems to work, applied here:

https://github.com/volatilityfoundation/volatility3/pull/1364/commits/f2771bce87e80c7a988247f855e5afa117753654

volatility3/framework/__init__.py
c0rydoras: fixed (using redundant alias name) here: https://github.com/volatilityfoundation/volatility3/pull/1364/commits/f2771bce87e80c7a988247f855e5afa117753654

volatility3/framework/constants/__init__.py
c0rydoras: removed here: https://github.com/volatilityfoundation/volatility3/pull/1364/commits/f2771bce87e80c7a988247f855e5afa117753654

volatility3/framework/configuration/__init__.py
c0rydoras: removed here: https://github.com/volatilityfoundation/volatility3/pull/1364/commits/f2771bce87e80c7a988247f855e5afa117753654

volatility3/framework/layers/resources.py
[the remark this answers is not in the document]
c0rydoras: i've played around with this a bit and can't get it to act as global helper without something like this:

```python
opener = urllib.request.build_opener(SMBHandler) # not global but smb can be opened using that opener
opener.open("smb://blah") // works
```

volatility3/framework/interfaces/__init__.py
c0rydoras: removed here: https://github.com/volatilityfoundation/volatility3/pull/1364/commits/f2771bce87e80c7a988247f855e5afa117753654

volatility3/framework/layers/resources.py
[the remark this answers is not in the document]
ikelos: Just importing it already worked?  We've had automatic SMB support for a while.  The only issue is that the import looks like it doesn't do anything (but it does)...

ikelos:
Awesome, looking good, tests are passing, just need some of the guys to comment on whether they're happy with us linting with it...

ikelos:
Sorry, some changes got merged that affect the `pyproject.toml` and `setup.py`, please could you rebase this?

volatility3/framework/plugins/windows/netscan.py
ikelos: I guess we'll keep the ignore in here, but could you add `FIXME: ` to one of the comments below please, just so we don't lose it again?  That should be it save for the reviews from the rest of the team...

ikelos:
Awesome, thanks for all that!  I've pinged everybody as well as mentioning them here, so hopefully they'll see it at least by Wednesday (the US is going to have Thanksgiving this week, so if they don't see it by Wednesday it'll be the week after at the earliest.  For now, I guess we just sit tight and hope the code doesn't diverge too much between now and when we want to merge it...

volatility3/framework/layers/resources.py
```diff
@@ -29,7 +29,7 @@
 
 try:
     # Import so that the handler is found by the framework.class_subclasses callc
-    import smb.SMBHandler  # lgtm [py/unused-import]
+    from smb import SMBHandler as SMBHandler
```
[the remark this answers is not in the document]
ikelos: Not sure why this one didn't work but the other instances did?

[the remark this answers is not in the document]
ikelos: Oh, this is codeQL not Ruff.  That's ok, it's not fatal.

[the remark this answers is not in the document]
ikelos: You could add back in the `lgtm` imperative to ignore it if you wanted, seems CodeQL never got the message that renamed imports really *really* should be imported...  5:P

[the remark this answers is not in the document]
c0rydoras: done: https://github.com/volatilityfoundation/volatility3/pull/1364/commits/dd4e783d18c9c72b9376e6cdd0aa2d9a93817de0

c0rydoras:
any updates on this?

ikelos:
I'm afraid not yet, but the next Volatility meeting will be on the 11th, so I'll raise it then...

c0rydoras:
> I'm afraid not yet, but the next Volatility meeting will be on the 11th, so I'll raise it then...

How did it go?

ikelos:
Unfortunately the meeting was postponed until next Wednesday, but it's
still first on the agenda...

ikelos:
@atcuno, this may also catch a lot of the typing errors you've spotted.  We're trying to keep this in check, and get it merged before the tree changes a lot.  Could you please look it over so we can get it (and its fixes) in place before the tree gets lots of little bug fixes that'll get in the way of this?  It's been open since last month and I mentioned it on the core members slack...

atcuno:
The changes made here are really nice, especially all the "yield from" and == -> is. I say we keep using this way forward.

ikelos:
Cool, sounds like we got enough of a go ahead @c0rydoras , just one more rebase and I'll apply it...  5:)

ikelos:
@c0rydoras Any idea why [this](https://github.com/volatilityfoundation/volatility3/pull/1455#discussion_r1893327145) didn't fail the ruff test we put in place?  Just want to make sure it's operating as expected...

c0rydoras:
AFAICT ruff doesn't recognize `vollog` as a logger: 
- https://docs.astral.sh/ruff/rules/logging-f-string/#known-problems
- https://github.com/astral-sh/ruff/blob/1feb3cf41a40bf382472495289675d0ac0a70e6a/crates/ruff_python_semantic/src/analyze/logging.rs#L62-L74

changing the name to e.g. `logger`/`logvol` etc. makes it fail as expected

ikelos:
Ok, thanks for the links!  Considering I thought that ruff could determine object types, it seems pretty odd to me that they use such a simple heuristic, but perhaps I'm expecting it to be like mypy.

Doing a sed from `vollog` to `vollogger` shows up 366 errors.  I don't intend to up-end the entire codebase for this (vollog renames *and* error message fixes), whilst it is a small optimization, I don't think we include values that take time to calculate in many of the error messages, so I intend to leave this as is.

Issue #1366, opened 2024-11-27
Linux.pstree bug


tuttimann:
**Describe the bug**
I wanted to investigate the following memory image: 
[13cubed Linux Memory Forensics Challenge](https://www.youtube.com/redirect?event=video_description&redir_token=QUFFLUhqbHdEdmRLcGVvUVdGLVZCYXgxbU03ekRWOFQ4Z3xBQ3Jtc0ttalBQMmQzcHB2ZUNvVGpwV1RTWVBsRjI2TDIyVWRUSjdtbG9ia0lTbGRMdnpRVVBxWTVNMTk0c3gxd1YtTHViVV9WYjBlQjlLSzRsR2YzNC1VbkZUbXRhSDRReW9zSGlXaU5YU3ZLdjJ4TWdKZG5Jaw&q=https%3A%2F%2F13cubed.s3.amazonaws.com%2Fdownloads%2Flinux_challenge.zip&v=IHd85h6T57E)

I downloaded the right symbols for the memory image and afterwards ran linux.pslist without any problems, but when I am running linux.pstree I get an error message, which I will paste below.

**Context**
Volatility Version:  2.11.0
Operating System:  Ubuntu 24.04.1 LTS
Python Version:  Python 3.12.3
Suspected Operating System:  Ubuntu 22.04 6.5.0-41
Command:  `vol3 -f memory.vmem linux.pstree`

**To Reproduce**

1. Use command on the given memory image
2. See error

**Expected behavior**
Output of pstree

**Example output**
```Volatility 3 Framework 2.11.0
Progress:  100.00		Stacking attempts finished                   
OFFSET (V)	PID	TID	PPID	COMM
Traceback (most recent call last):
  File "/home/fabian/Tools/volatility3/vol.py", line 11, in <module>
    volatility3.cli.main()
  File "/home/fabian/Tools/volatility3/volatility3/cli/__init__.py", line 917, in main
    CommandLine().run()
  File "/home/fabian/Tools/volatility3/volatility3/cli/__init__.py", line 505, in run
    renderer.render(grid)
  File "/home/fabian/Tools/volatility3/volatility3/cli/text_renderer.py", line 230, in render
    grid.populate(visitor, outfd)
  File "/home/fabian/Tools/volatility3/volatility3/framework/renderers/__init__.py", line 248, in populate
    treenode = self._append(parent, item)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #1367, opened 2024-11-28
Linux: vmayarascan --yara-string is broken


gcmoreira:
**Describe the bug**
`linux.vmayarascan --yara-string` crashes

**Context**
Volatility Version:  2.12.0
Operating System:  Linux
Python Version:  3.8.20
Suspected Operating System:  Linux
Command:  ```$ ./vol.py -f../linux-sample-1.bin linux.vmayarascan.VmaYaraScan --pid 1 --yara-string "ELF"```

**To Reproduce**

1. Use command:
```$ ./vol.py -f../linux-sample-1.bin linux.vmayarascan.VmaYaraScan --pid 1 --yara-string "ELF"```
3. See error
```python
$ ./vol.py -f../linux-sample-1.bin linux.vmayarascan.VmaYaraScan --pid 1 --yara-string "ELF"
Volatility 3 Framework 2.12.0
Progress:  100.00               Stacking attempts finished                 
Offset  PID     Rule    Component       Value
Traceback (most recent call last):
  File "./vol.py", line 11, in <module>
    volatility3.cli.main()
  File "/home/gmoreira/vol3/volatility3/cli/__init__.py", line 917, in main
    CommandLine().run()
  File "/home/gmoreira/vol3/volatility3/cli/__init__.py", line 505, in run
    renderer.render(grid)
  File "/home/gmoreira/vol3/volatility3/cli/text_renderer.py", line 230, in render
    grid.populate(visitor, outfd)
  File "/home/gmoreira/vol3/volatility3/framework/renderers/__init__.py", line 245, in populate
    for level, item in self._generator:
  File "/home/gmoreira/vol3/volatility3/framework/plugins/linux/vmayarascan.py", line 70, in _generator
    for match in rules.match(
AttributeError: 'builtins.Rules' object has no attribute 'match'
```
**Expected behavior**
```python
$ ./vol.py -f../linux-sample-1.bin linux.vmayarascan.VmaYaraScan --pid 1 --yara-string "ELF"
Volatility 3 Framework 2.9.0
Progress:  100.00               Stacking attempts finished                 
Offset  PID     Rule    Component       Value

0x7f489741c001  1       default.r1      $a      45 4c 46
0x7f4897620001  1       default.r1      $a      45 4c 46
0x7f48979ab001  1       default.r1      $a      45 4c 46
0x7f4897e0b001  1       default.r1      $a      45 4c 46
[...]

PR #1368, opened 2024-11-28
Linux pslist dependencies fix 1366

gcmoreira:
In the context of #1366, this PR updates pslist major version and its 18 dependent plugins.

Additionally, it adds 15 new test cases for the remaining plugins that didn't have a test case before. This will help maintain compatibility and ensure we don't face issues with the plugin requirements in the future. 

The `linux_library_list` test case was taking too long to execute, which was delaying the entire testing task. With the increasing number of testcases, it's become crucial to address this issue. This PR reduces the scope of the plugin to handle just one process, making it faster.

Please note that the **linux.vmayarascan plugin test** will fail until the issue discussed in [Volatility3 issue #1367](https://github.com/volatilityfoundation/volatility3/issues/1367) is addressed.

### DIFF
--- a/volatility3/framework/plugins/linux/pidhashtable.py
+++ b/volatility3/framework/plugins/linux/pidhashtable.py
@@ -19,8 +19,7 @@ class PIDHashTable(plugins.PluginInterface):
     """Enumerates processes through the PID hash table"""
 
     _required_framework_version = (2, 0, 0)
-
-    _version = (1, 0, 1)
+    _version = (1, 0, 2)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
@@ -31,7 +30,7 @@ class PIDHashTable(plugins.PluginInterface):
                 architectures=["Intel32", "Intel64"],
             ),
             requirements.PluginRequirement(
-                name="pslist", plugin=pslist.PsList, version=(2, 0, 0)
+                name="pslist", plugin=pslist.PsList, version=(3, 0, 0)
             ),
             requirements.VersionRequirement(
                 name="linuxutils", component=linux.LinuxUtilities, version=(2, 1, 0)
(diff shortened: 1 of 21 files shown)


gcmoreira:
Hey @ikelos, as I anticipated above, the `linux.vmayarascan` test case failed. Once https://github.com/volatilityfoundation/volatility3/issues/1367 is resolved, this ticket will pass all the tests.

ikelos:
Depends on #1370 landing.

gcmoreira:
@ikelos it passed all the tests. Please review to ensure that all version requirements are as expected.
This fixes https://github.com/volatilityfoundation/volatility3/issues/1366

volatility3/framework/plugins/linux/pstree.py
```diff
@@ -100,13 +101,11 @@ def _generator(
...
             task = self._tasks[pid]
 
-            row = pslist.PsList.get_task_fields(task, decorate_comm)
-            # update the first element, the offset, in the row tuple to use format_hints.Hex
-            # as a simple int is returned from get_task_fields.
-            row = (format_hints.Hex(row[0]),) + row[1:]
-
-            tid = task.pid
-            yield (self._levels[tid] - 1, row)
+            offset, pid, tid, ppid, name, _creation_time = (
```
ikelos, line 104: I much prefer this, thanks!

volatility3/framework/plugins/linux/pidhashtable.py
```diff
@@ -236,8 +235,8 @@ def _generator(
         self, decorate_comm: bool = False
     ) -> interfaces.objects.ObjectInterface:
         for task in self.get_tasks():
-            offset, pid, tid, ppid, name = pslist.PsList.get_task_fields(
-                task, decorate_comm
+            offset, pid, tid, ppid, name, _creation_time = (
```
ikelos, line 238: Really appreciate you naming the unused vars, think it'll make it easier if/when we come back to make use of it in the future.  5:)

volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -591,7 +593,7 @@ def _generator(self, pids: List[int], netns_id_arg: int, symbol_table: str):
             tasks: String with a list of tasks and FDs using a socket. It can also have
                    extended information such as socket filters, bpf info, etc.
         """
-        filter_func = lsof.pslist.PsList.create_pid_filter(pids)
```
ikelos, line 594: Yep, this was bad and I think I prefer the direct requirements (even though it would technically come bundled with lsof).  Good spot!

ikelos (approved):
I think that catches everything.  I'd probably have done it as two PRs (one tidying up pslist, and the other adding the new tests) because there's kinda more than one thing being done here, but certainly not a shopstopper.  Thanks!

PR #1377, opened 2024-12-01
Testing: Enable automatic OS image selection by matching test name and image filename prefixes

gcmoreira:
There is an issue with `pytest` in the development environment. Additionally, the current implementation does not scale well for handling multiple images per OS.

When `pytest` is executed it applies its discovery patterns, triggering the `pytest_generate_tests()` function. This function blindly adds all images to each test case, leading to incorrect behavior. For example, Linux tests end up being executed with Windows and Mac images, and vice versa, which makes many testcase-image combination to fail and it's a waste if time. 
In VSCode, for instance, the `pytest` discovery is executed during [test autodiscovery](https://code.visualstudio.com/docs/python/testing#_test-discovery). 

This PR enables `pytest` to automatically select the appropriate image to execute for each operating system.
For this functionality to work, two conditions must be respected:
1. Test case names must include the `test_[OS]_` prefix.
2. Images must have the `[OS]-` prefix.

**Both conditions are already satisfied in our current setup.**

Additionally, with this implementation, it becomes possible to execute a `pytest` command easily using a command like:
```
pytest ./test/test_volatility.py --volatility=vol.py --image-dir=./test_images -v
```
The only drawback with this is that if test case execution is parallelized, they will finish at different times and won't be grouped or sorted by operating system. 
However, if this grouping isn't important, we could consider extending these changes (@ikelos?)

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -35,16 +35,39 @@ def pytest_addoption(parser):
 def pytest_generate_tests(metafunc):
     """Parameterize tests based on image names"""
 
-    images = metafunc.config.getoption("image")
+    images = metafunc.config.getoption("image").copy()
     for image_dir in metafunc.config.getoption("image_dir"):
-        images = images + [
-            os.path.join(image_dir, dir) for dir in os.listdir(image_dir)
+        images += [
+            os.path.join(image_dir, dir_name) for dir_name in os.listdir(image_dir)
         ]
 
-    # tests with "image" parameter are run against images
+    # tests with "image" parameter are run against image
     if "image" in metafunc.fixturenames:
+        filtered_images = []
+        ids = []
+        for image in images:
+            image_base = os.path.basename(image)
+            test_name = metafunc.definition.originalname
+            if test_name.startswith("test_windows_") and not image_base.startswith(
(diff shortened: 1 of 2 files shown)


test/conftest.py
```diff
@@ -35,16 +35,39 @@ def pytest_addoption(parser):
...
-    # tests with "image" parameter are run against images
+    # tests with "image" parameter are run against image
     if "image" in metafunc.fixturenames:
+        filtered_images = []
+        ids = []
+        for image in images:
+            image_base = os.path.basename(image)
+            test_name = metafunc.definition.originalname
+            if test_name.startswith("test_windows_") and not image_base.startswith(
+                "win-"
```
ikelos, line 52: The discrepancy between windows and win is a bit jarring (and also makes me think it's the only thing stopping this being put in tight loop for the OSes, rather than a switch statement.

ikelos (approved):
The changes look good.  I think I'm ok parallelization messing with the order, as long as any failed tests can easily be identified, I don't see the ordering as important, other than for timing, but I don't think anyone makes use of the tests for that so...

PR #1381, opened 2024-12-05
Volshell: Add Dedicated Method to retrieve EPROCESS/Task object

the-rectifier:
This PR introduces a `gp()` (`get_process()/get_task()`) method to the Windows and Linux Volshell respectively.

I couldn't find a way for one to (quickly) grab an arbitrary `EPROCESS`/`Task` object from the shell and I thought it'd be quicker, when interactively exploring structures, to have a dedicated method rather than repeat a one-liner each time.

Also, update `linux.pslist` version requirements, since they changed in https://github.com/volatilityfoundation/volatility3/commit/fa060646352d41cf6d818e6cb8c40deaa91e7e7f:

```
Unsatisfied requirement plugins.Volshell.pslist: Version 2.0.0 dependency on volatility3.plugins.linux.pslist.PsList unmet
Unable to validate the plugin requirements: ['plugins.Volshell.pslist']
```

### DIFF
--- a/volatility3/cli/volshell/linux.py
+++ b/volatility3/cli/volshell/linux.py
@@ -51,42 +51,61 @@ class Volshell(generic.Volshell):
                 return None
         print(f"No task with task ID {pid} found")
 
-    def get_process(self, pid=None, offset=None):
-        """Get Task based on a process ID. Does not retrieve the layer, to change layer use the .pid attribute. The offset argument can be used both for physical or virtual offsets"""
+    def get_process(self, pid=None, virtaddr=None, physaddr=None):
+        """Return the task_struct object that matches the pid. If a physical or a virtual address is provided, construct the task_struct object at said address. Only one parameter is allowed.
 
-        if pid is not None and offset is not None:
+        Args:
+            pid (int, optional): PID to search for
+            virtaddr (int, optional): Virtual address to construct object at
+            physaddr (int, optional): Physical address to construct object at
+
+        Returns:
+            ObjectInterface: task_struct Object
+        """
+
+        if sum(1 if x is not None else 0 for x in [pid, virtaddr, physaddr]) != 1:
             print("Only one parameter is accepted")
             return None
 
-        if offset is not None:
-            vmlinux_module_name = self.config["kernel"]
-            vmlinux = self.context.modules[vmlinux_module_name]
+        vmlinux_module_name = self.config["kernel"]
+        vmlinux = self.context.modules[vmlinux_module_name]
 
-            kernel_layer_name = vmlinux.layer_name
-            kernel_layer = self.context.layers[kernel_layer_name]
+        kernel_layer_name = vmlinux.layer_name
+        kernel_layer = self.context.layers[kernel_layer_name]
 
-            memory_layer_name = kernel_layer.dependencies[0]
+        memory_layer_name = kernel_layer.dependencies[0]
 
-            ptask = self.context.object(
-                vmlinux.symbol_table_name + constants.BANG + "task_struct",
+        task_struct_symbol = vmlinux.symbol_table_name + constants.BANG + "task_struct"
+
+        if virtaddr is not None:
+            task = self.context.object(
+                task_struct_symbol,
+                layer_name=kernel_layer_name,
+                offset=virtaddr,
+            )
+
+        if physaddr is not None:
+            task = self.context.object(
+                task_struct_symbol,
                 layer_name=memory_layer_name,
-                offset=offset,
+                offset=physaddr,
                 native_layer_name=kernel_layer_name,
             )
 
+        if physaddr is not None or virtaddr is not None:
             try:
-                DescExitStateEnum(ptask.exit_state)
+                DescExitStateEnum(task.exit_state)
             except ValueError:
                 print(
-                    f"task_struct @ {hex(ptask.vol.offset)} as exit_state {ptask.exit_state} is likely not valid"
+                    f"task_struct @ {hex(task.vol.offset)} as exit_state {task.exit_state} is likely not valid"
                 )
 
-            if not (0 < ptask.pid < 65535):
+            if not (0 < task.pid < 65535):
                 print(
-                    f"task_struct @ {hex(ptask.vol.offset)} as pid {ptask.pid} is likely not valid"
+                    f"task_struct @ {hex(task.vol.offset)} as pid {task.pid} is likely not valid"
                 )
 
-            return ptask
+            return task
 
         if pid is not None:
             tasks = self.list_tasks()
@@ -107,7 +126,7 @@ class Volshell(generic.Volshell):
         result += [
             (["ct", "change_task", "cp"], self.change_task),
             (["lt", "list_tasks", "ps"], self.list_tasks),
-            (["gp", "get_process"], self.get_process),
+            (["gp", "get_process", "get_task"], self.get_process),
             (["symbols"], self.context.symbol_space[self.current_symbol_table]),
         ]
         if self.config.get("pid", None) is not None:
--- a/volatility3/cli/volshell/linux.py
+++ b/volatility3/cli/volshell/linux.py
@@ -3,11 +3,22 @@
 #
 
 from typing import Any, List, Optional, Tuple, Union
+from enum import Enum
 
 from volatility3.cli.volshell import generic
 from volatility3.framework import constants, interfaces
 from volatility3.framework.configuration import requirements
 from volatility3.plugins.linux import pslist
+
+
+# Could import the enum from psscan.py to avoid code duplication
+class DescExitStateEnum(Enum):
+    """Enum for linux task exit_state as defined in include/linux/sched.h"""
+
+    TASK_RUNNING = 0x00000000
+    EXIT_DEAD = 0x00000010
+    EXIT_ZOMBIE = 0x00000020
+    EXIT_TRACE = EXIT_ZOMBIE | EXIT_DEAD
 
 
 class Volshell(generic.Volshell):
@@ -40,6 +51,52 @@ class Volshell(generic.Volshell):
                 return None
         print(f"No task with task ID {pid} found")
 
+    def get_process(self, pid=None, offset=None):
+        """Get Task based on a process ID. Does not retrieve the layer, to change layer use the .pid attribute. The offset argument can be used both for physical or virtual offsets"""
+
+        if pid is not None and offset is not None:
+            print("Only one parameter is accepted")
+            return None
+
+        if offset is not None:
+            vmlinux_module_name = self.config["kernel"]
+            vmlinux = self.context.modules[vmlinux_module_name]
+
+            kernel_layer_name = vmlinux.layer_name
+            kernel_layer = self.context.layers[kernel_layer_name]
+
+            memory_layer_name = kernel_layer.dependencies[0]
+
+            ptask = self.context.object(
+                vmlinux.symbol_table_name + constants.BANG + "task_struct",
+                layer_name=memory_layer_name,
+                offset=offset,
+                native_layer_name=kernel_layer_name,
+            )
+
+            try:
+                DescExitStateEnum(ptask.exit_state)
+            except ValueError:
+                print(
+                    f"task_struct @ {hex(ptask.vol.offset)} as exit_state {ptask.exit_state} is likely not valid"
+                )
+
+            if not (0 < ptask.pid < 65535):
+                print(
+                    f"task_struct @ {hex(ptask.vol.offset)} as pid {ptask.pid} is likely not valid"
+                )
+
+            return ptask
+
+        if pid is not None:
+            tasks = self.list_tasks()
+            for task in tasks:
+                if task.pid == pid:
+                    return task
+            print(f"No task with task ID {pid} found")
+
+        return None
+
     def list_tasks(self):
         """Returns a list of task objects from the primary layer"""
         # We always use the main kernel memory and associated symbols
@@ -50,6 +107,7 @@ class Volshell(generic.Volshell):
         result += [
             (["ct", "change_task", "cp"], self.change_task),
             (["lt", "list_tasks", "ps"], self.list_tasks),
+            (["gp", "get_process"], self.get_process),
             (["symbols"], self.context.symbol_space[self.current_symbol_table]),
         ]
         if self.config.get("pid", None) is not None:
--- a/volatility3/cli/volshell/windows.py
+++ b/volatility3/cli/volshell/windows.py
@@ -44,10 +44,19 @@ class Volshell(generic.Volshell):
             )
         )
 
-    def get_process(self, pid=None, v_offset=None, p_offset=None):
-        """Returns the EPROCESS object that matches the pid. If v_offset/p_offset is provided, construct the EPROCESS object at the provided address. Only one parameter is allowed."""
+    def get_process(self, pid=None, virtaddr=None, physaddr=None):
+        """Returns the _EPROCESS object that matches the pid. If a physical or a virtual address is provided, construct the _EPROCESS object at said address. Only one parameter is allowed.
 
-        if sum(1 if x is not None else 0 for x in [pid, v_offset, p_offset]) != 1:
+        Args:
+            pid (int, optional): PID / UniqueProcessId to search for.
+            virtaddr (int, optional): Virtual address to construct object at
+            physaddr (int, optional): Physical address to construct object at
+
+        Returns:
+            ObjectInterface: _EPROCESS Object
+        """
+
+        if sum(1 if x is not None else 0 for x in [pid, virtaddr, physaddr]) != 1:
             print("Only one parameter is accepted")
             return None
 
@@ -61,20 +70,20 @@ class Volshell(generic.Volshell):
 
         eprocess_symbol = kernel.symbol_table_name + constants.BANG + "_EPROCESS"
 
-        if v_offset is not None:
+        if virtaddr is not None:
             eproc = self.context.object(
                 eprocess_symbol,
                 layer_name=kernel_layer_name,
-                offset=v_offset,
+                offset=virtaddr,
             )
 
             return eproc
 
-        if p_offset is not None:
+        if physaddr is not None:
             eproc = self.context.object(
                 eprocess_symbol,
                 layer_name=memory_layer_name,
-                offset=p_offset,
+                offset=physaddr,
                 native_layer_name=kernel_layer_name,
             )
 
--- a/volatility3/cli/volshell/windows.py
+++ b/volatility3/cli/volshell/windows.py
@@ -44,11 +44,58 @@ class Volshell(generic.Volshell):
             )
         )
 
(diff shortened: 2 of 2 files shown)


atcuno:
Hello @the-rectifier - nice timing on this as I was recently working on a similar feature as forcing people to do the list comprehension is pretty miserable as you noted.

With that said, accepting only a `pid` is not robust enough for this feature to be fully usable. There are a few reasons for this:

1) Smear can make where multiple processes (one active and one or more terminated) have the same PID, making it non-deterministic.

2) When specifying the pid, the only thing that can be done to find the process is walk the list (as your code does). This makes it impossible to specify processes hidden from the list by a rootkit and/or missing from the list due to smear.

To work around this, the offset (virtual or physical) needs to be allowed to be specified as this is 1) 100% precise on which process to analyze 2) supports analysis of processes found from psscan / psxview. The cleanest way in my view would be optional parameters like pid=None,v_offset=None,p_offset=None and then the code branch based on which is set.

If you are willing to work these offset parameters into your code then I can drop my code as I would rather encourage more external developers to contribute to the framework.

volatility3/cli/volshell/linux.py
```diff
@@ -50,6 +58,7 @@ def construct_locals(self) -> List[Tuple[List[str], Any]]:
         result += [
             (["ct", "change_task", "cp"], self.change_task),
             (["lt", "list_tasks", "ps"], self.list_tasks),
+            (["gp", "get_task"], self.get_task),
```
ikelos: How do we feel about the naming?  Should it be gt in linux, or will the difference be confusing?  Should we call it `get_process` instead to keep with windows?  We can allow alternatives (see `list_tasks` as both `lt` and `ps` in the line above).  Perhaps we should have it as `gt` and then with `gp` as a secondary?

ikelos (commented):
Generally looks fine, thanks!  Just wondering about an alias to keep the naming more consistent...

volatility3/cli/volshell/linux.py
the-rectifier: `gt` is reserved for `generate_tree` afaik

the-rectifier:
Hello @atcuno, thanks for the reply!

I wasn't aware of the issues with smearing. I just hacked up a quick way to interact with structures now that I started working on my thesis. 

That being said, I am not that familiar with volatility internals yet, to promise a timeline. I might look into a proper version sometime in the future but I think it'd be better to continue with your codebase

volatility3/cli/volshell/linux.py
ikelos: Ah, you're right... shame...  5:S  Would it be worth aliasing it to `get_process` as well, or do we just live with the inconsistency?

volatility3/cli/volshell/linux.py
```diff
@@ -40,6 +40,14 @@
                 return None
         print(f"No task with task ID {pid} found")
 
+    def get_task(self, pid):
```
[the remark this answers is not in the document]
ikelos: Please make sure this ends in an explicit `return None` if the pid isn't found or there's no tasks.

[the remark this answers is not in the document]
ikelos: It also looks like there's two bits of whitespace that are causing black to complain

atcuno:
@the-rectifier don't give up yet. The basic idea would be to allow either the virtual address or physical offset to be specified and then create an _EPROCESS object at it.

For physical offsets, there is a helper function:

https://github.com/volatilityfoundation/volatility3/blob/3ff304ceed2501df6d9291259f9bc6b3c587c303/volatility3/framework/plugins/windows/psscan.py#L164

Let me know if you have any questions.

ikelos:
I'd also be wary of putting too much into the generic volshell plugin.  If a plugin can do it better I think we should leave the plugin to do it (even if that includes working on a physical process offset).  Linking volshell functionality against a specific plugin (even a stock one) is a bad idea, as would duplicating the functionality be.   I'm sure there's a solution, but I don't think it's necessary for the immediate term...

atcuno:
I would be fine with code duplication in this case, even if copy/paste from the plugin, as being limited to `pid=` is very limiting and makes it impossible to use volshell on processes hidden by rootkits.

ikelos:
> ... and makes it impossible to use volshell on processes hidden by rootkits.

```
(layer_name) >>> import volatility3.plugins.linux.pslist
(layer_name) >>> dpo(volatility3.plugins.linux.pslist.PsList, kernel=self.kernel)
```

I think you mean inconvenient.  I've no problem loading up specific kernel offsets physical or virtual, but making volshell depend on psscan to find them rather than having to put a couple extra lines isn't worth the complexity it would bring to the command line tool in my opinion.

It's really not difficult to make use of a plugin and if that's too many lines to type, a snippet would reduce that.  We're currently in a state where the dependency on simple `pslist` was missed when the main pslist plugin got version bumped, meaning that stock volshell on the current commit doesn't load because of that dependency.  I'm not keen to add more when the plugin output is so readily accessible...

atcuno:
The psscan link was to the `virtual_process_from_physical` helper function.

I was saying that code can be copy/pasted into the volshell side to avoid the dependency. Then passing in a physical offset to get the _EPROCESS would work inside of volshell. That offset would come from the user running psscan or psxview first and noticing the hidden process.

ikelos:
My bad, that function doesn't look overly complex so I think I'd be ok duplicating it too...

the-rectifier:
Thanks for the replies, all. I played with the code a bit, and I added this to the windows side of things:

```python
kvo = self.config['kernel.offset']
ntkrnlmp = self.context.module(self.current_symbol_table, layer_name='layer_name', offset=kvo)

eproc = ntkrnlmp.object(
    object_type='_EPROCESS',
    offset=v_offset,
    absolute=True
)
```

As @atcuno suggested, much like the `psscan` plugin, since we have the virtual address we can just grab an `_EPROCESS` object at that address. 

However, I still can't wrap my head around what to do when a physical address is provided. Is there a way to get from the physical address to the corresponding virtual one?

eve-mem:
@the-rectifier if you look at the linux.psscan as an example you'll see that the object is made with offsets on the physical layer but its native layer is set to the virtual one.

It's a sort of magical 'it just works'TM magic part of vol. It allows you to make an object on one layer, but then follow any pointers etc in another. 

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3%2Fframework%2Fplugins%2Flinux%2Fpsscan.py#L148-L154

eve-mem:
Also, this discussion has made me think it might be useful to share scripts that people could import by `rs` that makes some customisation for specific tasks. 

Similar to how people make use of gdb configs. I need to think about sharing all the volshell snippets I use.

ikelos:
> ```python
> kvo = self.config['kernel.offset']
> ```

Also, we need to be slightly careful when putting in code that accessesd sub-config values using `kernel.offset`.  The `.` is reasonable but technically I believe it's technically `CONFIG_SEPARATOR` which means it can change and shouldn't really be coded in as part of a string.  I think using `interfaces.configuration.path_join('kernel', 'offset')` would be ok...

the-rectifier:
Hey all, happy new year!

I've been working on my thesis and I figured I'll work on this, now that I have a slightly better understanding of the framework and its inner workings. 

I updated `windows.py` to accept a virtual offset and to create the `_EPROCESS` object at that offset. The only issue for me is that on the windows side, I didn't find any way to create the object at the physical offset. From my understanding, `psscan` uses `poolscanner` to search for the `_POOL_HEADER` (and its tag) and creates the `_EPROCESS` from there...

thank you @eve-mem for suggesting that linux virtual/physical offsets can be used interchangeably when creating the object 

also implemented some extra sanity checks if we know what PID are we expecting

ikelos:
Happy New Year to you too!  5:D

So basically, if you've got `physical_layer_name`, `physical_offset` and `virtual_layer_name`, then you simply do `create_object(layer_name=physical_layer_name, offset=physical_offset, native_layer_name=virtual_layer_name)`.  This will read the data for the object out of the physical layer, but ensure that any pointers that are accessed within create their target objects on the virtual layer (at the offset mentioned in the pointer).  This has the effect of providing all the right data from the object and its children, but allows it to have been created on the physical layer.  Does that help explain the process?  You still have to have found the physical offset somehow yourself, but once you've got it, making the actual object is no problem...  5:)

volatility3/cli/volshell/windows.py
```diff
@@ -44,11 +44,63 @@
...
+            if pid is not None and eproc.UniqueProcessId != pid:
+                print(
+                    f"_EPROCESS @ {hex(v_offset)} UniqueProcessId mismatch with provided PID: {pid}"
+                )
+
+            if p_offset is not None:
+                (_, _, ph_offset, _, _) = list(
+                    memory.mapping(offset=eproc.vol.offset, length=0)
+                )[0]
+                if ph_offset != ph_offset:
```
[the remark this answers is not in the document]
ikelos: I assume one of these is supposed to be p_offset?  I'm not sure they'd ever match though?  Mapping should always return one layer down...

volatility3/cli/volshell/windows.py
```diff
@@ -44,11 +44,63 @@ def list_processes(self):
...
+        memory = self.context.layers[layer_name]
+
+        if p_offset is not None and v_offset is None:
+            # TODO
+            raise NotImplementedError
+
+        if v_offset is not None:
+            # v_off takes precedence, if pid is supplied do a sanity check and warn if not matched
+            ntkrnlmp = self.context.module(
+                symbol_table, layer_name=layer_name, offset=kvo
```
ikelos: The layer is being encoded in module you're creating here, but you can just call `context.object` providing the `symbol_table + constants.BANG + "_EPROCESS"`...

the-rectifier:
So I played with the code a bit, and before committing, I'd like to ask a few more questions:

- Is it correct to say that the `memory_layer` (Lime, Elf) is the physical layer and the `layer_name` and the layers created using `add_process_layer()` are the virtual layers (Intel32) (and is this always the case)?

- On the Linux side, at least from my testing, the offset can be either a virtual or a physical one, and the object is always created on the `memory_layer` with the `layer_name` as the native layer. On Windows this doesn't work so I would have to put different parameters for each type of offset (as currently implemented). So I wanted to ask if we should consider a type of address more important (concrete) or some precedence, ie physical > virtual > pid, or make it exclusively a single non-None parameter or leave it up to the user?

volatility3/cli/volshell/windows.py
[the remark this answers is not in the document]
the-rectifier: I checked it, you're right, they dont match. But why? The (user-supplied) `p_offset` comes from the psscan plugin, aren't I just creating the `_EPROCESS` at the virtual location just like plugin does?

ikelos:
No, that's not guaranteed, although it is convention.  Add process layer will typically create a virtual layer, the layer below an intel layer is typically called `memory_layerX`, but it's better to ask the virtual layer for `layer.config.get('memory_layer')` to determine the actual name of the lower layer.  Also note, that's only because intel's base layer is called `memory_layer` and for arm it may end up being different, but we'll cross that bridge later.

The native layer machinery is independent of the OS, but it may also be unnecessary.  If you've got the virtual address for it, that's best.  If you've got a physical address for it and you know the virtual layer it lives in that comes a close second, but there isn't really a precedent.  If you're intending to use a single offset parameter, don't it'll just confusing things, and other wise I'd say if someone provides more than one of (pid, virtual, physical) then you throw an error and say they need to provide only one...

the-rectifier:
Thanks for explaining the layers @ikelos, everything is much clearer now! I've made the changes you suggested, and also now I'm sourcing the layer names from the `kernel` module (like the linux psscan) as opposed to `self.current_layer` which can change. Since for linux the same code can be used for both offset types I have only one offset parameter, shall I change that?

the-rectifier:
Also, the force-push removed the documentation commit, will do that when I finalize the code

ikelos:
> Thanks for explaining the layers @ikelos, everything is much clearer now! I've made the changes you suggested, and also now I'm sourcing the layer names from the `kernel` module (like the linux psscan) as opposed to `self.current_layer` which can change. Since for linux the same code can be used for both offset types I have only one offset parameter, shall I change that?

I probably would, how do you know whether they provided a virtual or a physical offset otherwise?  If they both happen to be valid, you've no way of telling which one they meant...

volatility3/cli/volshell/windows.py
```diff
@@ -44,11 +44,58 @@ def list_processes(self):
...
+            print("Only one parameter is accepted")
+            return None
+
+        kernel_name = self.config["kernel"]
+        kernel = self.context.modules[kernel_name]
+
+        kernel_layer_name = kernel.layer_name
+
+        kernel_layer = self.context.layers[kernel_layer_name]
+        memory_layer_name = kernel_layer.dependencies[0]
```
ikelos, line 69: Again, likely to be right and until we get a proper API for what people generally mean when they say "physical layer" this'll do, but it's possibly it could be a swap layer or something entirely different one day.  Sadly, the question's not an easy one to answer, so for now this is ok (and much better than hardcoding it with a specific name!).  5:)

volatility3/cli/volshell/linux.py
```diff
@@ -40,6 +51,52 @@ def change_task(self, pid=None):
...
+
+            ptask = self.context.object(
+                vmlinux.symbol_table_name + constants.BANG + "task_struct",
+                layer_name=memory_layer_name,
+                offset=offset,
+                native_layer_name=kernel_layer_name,
+            )
+
+            try:
+                DescExitStateEnum(ptask.exit_state)
```
ikelos: Is this statement to test for offset validity?  If so I think something like `is_valid` would work better, so `ptask.exit_state.is_valid()`?

ikelos (commented):
Looking much better, thanks!  I think there's still a few improvements to be had, but to be honest I'd forgotten we used print rather than `vollog` within volshell, but that seems to be how the rest of it's coded...

volatility3/cli/volshell/windows.py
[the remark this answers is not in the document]
ikelos: I'm not sure, and now the other commit's not there I can't easily go back and check what it was doing.  It looks like it should've been doing a straight translation, but it depends what layer the eproc was actually constructed on (whether it was memory, or a different virtual layer or something).

volatility3/cli/volshell/linux.py
the-rectifier: I yanked this from the psscan where it checks for a valid exit state [psscan.py](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/psscan.py#L143). Also it seems that `is_valid()` is not a method of the members of the `task_struct` but rather the struct's itself [pslist.py](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/pslist.py#L262)

volatility3/cli/volshell/windows.py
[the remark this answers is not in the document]
the-rectifier: hmmm, I tested it again, most probably I was mapping from the physical layer, so yeah it returned a layer down as you said. Now I tried mapping from a virtual layer and the resulting address matches the physical address provided by psscan.

```
(layer_name) >>> _, _, pho, _, _ = list(self.context.layers["layer_name"].mapping(offset=vproc.vol.offset, length=0))[0]
(layer_name) >>> hex(pho)
'0x731082c0'
```

volatility3/cli/volshell/linux.py
ikelos: Oh, yeah, fair enough.  It will work, just sticks out a little to me...

volatility3/cli/volshell/linux.py
```diff
@@ -50,6 +107,7 @@ def construct_locals(self) -> List[Tuple[List[str], Any]]:
         result += [
             (["ct", "change_task", "cp"], self.change_task),
             (["lt", "list_tasks", "ps"], self.list_tasks),
+            (["gp", "get_process"], self.get_process),
```
ikelos: Can you add `"get_task"` to the end of that list too please?  `gt` won't work, but `get_task` then would...  Can't tell if it's worth adding that alias to windows too, I'll leave that one up to you...

volatility3/cli/volshell/linux.py
```diff
@@ -40,6 +51,52 @@ def change_task(self, pid=None):
                 return None
         print(f"No task with task ID {pid} found")
 
+    def get_process(self, pid=None, offset=None):
```
ikelos: This implies offset is always a physical offset.  Were you going to split that into `physical_offset` and `virtual_offset` or something?

volatility3/cli/volshell/linux.py
```diff
@@ -40,6 +51,52 @@ def change_task(self, pid=None):
...
+        """Get Task based on a process ID. Does not retrieve the layer, to change layer use the .pid attribute. The offset argument can be used both for physical or virtual offsets"""
```
ikelos: Also could you include an `Args:` and `Returns:` section here to specify the inputs and outputs please?

volatility3/cli/volshell/windows.py
```diff
@@ -44,11 +44,58 @@ def list_processes(self):
             )
         )
 
+    def get_process(self, pid=None, v_offset=None, p_offset=None):
+        """Returns the EPROCESS object that matches the pid. If v_offset/p_offset is provided, construct the EPROCESS object at the provided address. Only one parameter is allowed."""
```
ikelos: Same here with regards to `Args:` and `Returns:`.  Also I kinda prefer variables have more descriptive names (although I'll allow it if you really want, since it is likely to be typed be developers quite often).

volatility3/cli/volshell/linux.py
the-rectifier: shall I use `is_valid()` then on all proc/tasks? On windows since `dt()` breaks if the object is not valid, returning `None` makes more sense, than just printing a warning

volatility3/cli/volshell/linux.py
the-rectifier: I don't particularly like that we needed to mix process/task, so for the windows side I think i'd leave it just with the process suffix

volatility3/cli/volshell/linux.py
the-rectifier: I did split the code, basically the same as the windows side, but wanted to get your input on the `is_valid()` thing before committing

volatility3/cli/volshell/linux.py
ikelos: No, it's alright, it just stands out to me as a method of testing (and getting a second example in the code base will make it harder to prevent if more people add it that way).  There's currently a bit of a shift in the linux codebase where they're implementing check validity first rather than throwing exceptions.  If you'd like to and can get it to work, it might be better in keeping with the rest of the codebase, but if you don't want to or it's tricky to achieve then that's fine...

volatility3/cli/volshell/windows.py
the-rectifier: shall I use `virtaddr` and `physaddr` on both?

ikelos:
Ok, looks good, thanks!

PR #1382, opened 2024-12-05
PEP 488 – Elimination of PYO files

j-t-1:
Python 3.5 implememted PEP 488, eliminating .pyo files.


ikelos:
This doesn't pass the linter, but otherwise seems reasonable.  If you fix that up we can get it merged.  Thanks!  5:)

PR #1386, opened 2024-12-07
Cosmetic changes to documentation

j-t-1:


eve-mem:
@j-t-1 if you're already looking at fixing up docs, might also be worth adding in some explanations for the newer rx command in volshell. Added here: https://github.com/volatilityfoundation/volatility3/pull/1347

Something i didn't even think of doing :facepalm:

doc/source/symbol-tables.rst
```diff
@@ -25,17 +25,17 @@ as long as the symbol files stay in the same location.
...
+which matches the PDB name and GUID/age (or any compressed variant) will be used.  If such a symbol table cannot be found, then
 the associated PDB file will be downloaded from Microsoft's Symbol Server and converted into the appropriate JSON
 format, and will be saved in the correct location.
 
 Windows symbol tables can be manually constructed from an appropriate PDB file.  The primary tool for doing this
 is built into Volatility 3, called :file:`pdbconv.py`.  It can be run from the top-level Volatility path, using the
 following command:
 
-:command:`PYTHONPATH="." python volatility3/framework/symbols/windows/pdbconv.py`
+:command:`PYTHONPATH="."; python volatility3/framework/symbols/windows/pdbconv.py`
```
ikelos: I don't think that semi-colon belongs there.  It would run as two separate commands, rather than applying the environment variable `PYTHONPATH` to the command `python`.

ikelos (commented):
Thanks very much for reviewing this!  They all look good apart from the semi-colon in the command, which changes the meaning of the command.  If you could back that change out, the rest can go through just fine.  5:)

ikelos (approved):

j-t-1:
> @j-t-1 if you're already looking at fixing up docs, might also be worth adding in some explanations for the newer rx command in volshell. Added here: #1347

Will have a read. I am new to Volatility, so fixing some docs may be a good start.

PR #1388, opened 2024-12-09
Interfaces: change allow list for filenames to ensure they work safely on windows. Fixes issue #1387

### LINKED ISSUE #1387: File rename error with  windows.pslist.PsList --dump plugin on Windows (illegal character use)
**Describe the bug**
When using the windows.pslist.PsList --pid  --dump plugin on Windows, the .dmp file is not generated. It appears that a .tmp file of the dump is created, but then the program tries to rename it with an invalid character (?), which fails on Windows. 

**Context**
Volatility Version:  Volatility 3 Framework 2.12.0 (and latest stable)
Operating System:  Windows 10
Python Version:  3.8
Suspected Operating System:  
Command:  python vol.py   -f c:\Users\Administrator\Downloads\Win10-MalTest-ae85e0f6\Win10-MalTest-ae85e0f6.vmem windows.pslist.PsList --pid 3188 --dump

**To Reproduce**

1. Use command  use the pslist plugin to attempt to dump a process memory with the --pid and --dump switches
2. See error

**Expected behavior**
.dmp file created. However, I have a .tmp file because the rename of the file is failing because of the ? character not being valid on Windows.  The .tmp file is the process memory, it just is not being renamed as expected. 

**Example output**
python vol.py   -f c:\Users\Administrator\Downloads\Win10-MalTest-ae85e0f6\Win10-MalTest-ae85e0f6.vmem windows.pslist.PsList --pid 3188 --dump
Volatility 3 Framework 2.12.0
PID     PPID    ImageFileName   Offset(V)       Threads Handles SessionId       Wow64   CreateTime      ExitTime                File output
    volatility3.cli.main()
  File "C:\Users\Administrator\volatility3\volatility3\cli\__init__.py", line 917, in main
    CommandLine().run()
  File "C:\Users\Administrator\volatility3\volatility3\cli\__init__.py", line 505, in run
    renderer.render(grid)
  File "C:\Users\Administrator\volatility3\volatility3\cli\text_renderer.py", line 232, in render
    grid.populate(visitor, outfd)
  File "C:\Users\Administrator\volatility3\volatility3\framework\renderers\__init__.py", line 245, in populate
  File "C:\Users\Administrator\volatility3\volatility3\framework\plugins\windows\pslist.py", line 283, in _generator
    file_handle.close()
  File "C:\Users\Administrator\volatility3\volatility3\cli\__init__.py", line 842, in close
    os.rename(self._name, output_filename)
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'C:\\Users\\Administrator\\volatility3\\tmp_c4yb_gz6.vol3' -> 'C:\\Users\\Administrator\\volatility3\\3188.?WanaDecryptor.0x400000.dmp'

**Additional information**

Tried this on the latest stable and latest dev versions with same result. This worked in previous versions.

eve-mem:
Hello :wave: 

This PR changes the allowed characters in filenames to ensure that they work on windows. This fixes issue #1387 

I've verified this on a windows 11 computer to check that the chars I've removed were indeed blocked by windows. 

I have had to change the replacement character from `?` to `_`. This is because `?` is not allowed on windows. 

I've not bumped any version numbers - It feels like something should be updated though? What do you think? 

Microsoft were kind enough to provide a list already of chars that are not allowed:
https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file

> Use any character in the current code page for a name, including Unicode characters and characters in the extended character set (128–255), except for the following:
> 
>     The following reserved characters:
>         < (less than)
>         > (greater than)
>         : (colon)
>         " (double quote)
>         / (forward slash)
>         \ (backslash)
>         | (vertical bar or pipe)
>         ? (question mark)
>         * (asterisk)

### DIFF
--- a/volatility3/framework/interfaces/plugins.py
+++ b/volatility3/framework/interfaces/plugins.py
@@ -59,14 +59,14 @@ class FileHandlerInterface(io.RawIOBase):
 
     @staticmethod
     def sanitize_filename(filename: str) -> str:
-        """Sanititizes the filename to ensure only a specific whitelist of characters is allowed through"""
-        allowed = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.- ()[]{}!$%^:#~?<>,|"
+        """Sanititizes the filename to ensure only a specific allow list of characters is allowed through"""
+        allowed = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.- ()[]{}!$%^#~,"
         result = ""
         for char in filename:
             if char in allowed:
                 result += char
             else:
-                result += "?"
+                result += "_"  # change unwanted chars to an underscore
         return result
 
     def __enter__(self):


ikelos (approved):
Looks good, thanks!

PR #1389, opened 2024-12-10
fix(Windows: Handles): Unreliable SAR value on 24H2

dgmcdona:
Handles are not being decoded in 24H2+ samples. This is because the
`Handles._decode_pointer` method grabs the SAR shift value from the
disassembled function, but in these samples this value (`0x11`) is
incorrect. Adding a fallback to the default SAR value of `0x10` if the
obtained pointer is not valid in the kernel address space  resolves the
issue.

This is similar to #1214, but in this case the _correct_ constant pulled from the
disassembly via capstone is _incorrect_ for pointer decoding, so after decoding the
value, we need to check the validity of the pointer, and if it's
invalid, fall back to a pointer decoding with a default SAR value of
0x10. I've confirmed that this fixes the issue on 24H2 while not
breaking things on older samples.

### DIFF
--- a/volatility3/framework/plugins/windows/handles.py
+++ b/volatility3/framework/plugins/windows/handles.py
@@ -18,7 +18,7 @@ class Handles(interfaces.plugins.PluginInterface):
     """Lists process open handles."""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 0, 3)
+    _version = (2, 0, 0)
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/volatility3/framework/plugins/windows/handles.py
+++ b/volatility3/framework/plugins/windows/handles.py
@@ -21,11 +21,14 @@ except ImportError:
     has_capstone = False
 
 
+DEFAULT_SAR_VALUE = 0x10  # to be used only when decoding fails
+
+
 class Handles(interfaces.plugins.PluginInterface):
     """Lists process open handles."""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 0, 2)
+    _version = (1, 0, 3)
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -118,6 +121,10 @@ class Handles(interfaces.plugins.PluginInterface):
                         )
 
                 offset = self._decode_pointer(handle_table_entry.LowValue, magic)
+                if not self.context.layers[virtual].is_valid(offset):
+                    offset = self._decode_pointer(
+                        handle_table_entry.LowValue, DEFAULT_SAR_VALUE
+                    )
             else:
                 if handle_table_entry.InfoTable == 0:
                     return None
@@ -142,7 +149,6 @@ class Handles(interfaces.plugins.PluginInterface):
         pointers in the _HANDLE_TABLE_ENTRY which allows us to find the
         associated _OBJECT_HEADER.
         """
-        DEFAULT_SAR_VALUE = 0x10  # to be used only when decoding fails
 
         if self._sar_value is None:
             if not has_capstone:
--- a/volatility3/framework/plugins/windows/handles.py
+++ b/volatility3/framework/plugins/windows/handles.py
@@ -3,25 +3,15 @@
 #
 
 import logging
-from typing import List, Optional, Dict
+from typing import Dict, List, Optional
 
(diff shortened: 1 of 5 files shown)


eve-mem:
Hey @dgmcdona - really nice find. 

It's off by one, do you think that maybe some kind of flag is being encoded into that smallest bit? 

Or have you seen others that are wildly off?

No worries if you're not sure, I'm just interested.

Thanks!

ikelos:
Nice find, my only concern is that the solution is a little... finger in the air?  Do we know this is reliable and not brittle through anything other than empirical testing?  My concern is that this works, but we don't know why and we don't know if it works in every situation.  Otherwise I've no concerns about the patch.

ikelos:
Could we get a comment in there explaining the situation, possibly pointing at this bug and just a little more discussion to explain why we've implemented that behaviour?

dgmcdona:
@ikelos I agree on the nature of this fix - I don't necessarily like patching with things that just 'appear to work'. This was meant to be a quick fix to get analysis working on these samples again until I could find the time to do a proper RE effort on that kernel function that we've been relying on to get the offset. I'll do some testing this weekend to see if I can pin down a better solution.

dgmcdona:
I took a look at the structure definition going back across many versions of Windows 10 and Windows 8 and came up with what seems to be a very good solution for this: instead of using the `LowValue` member and detecting the shift amount through disassembly, we can use the `ObjectPointerBits` member of that union, shift the value left by 4 bits (for 16-byte aligned pointers), and get the same result.

I tested this across a large collection of windows samples spanning Windows 7 - 11, and had only positive results - in fact, thousands of new valid handle entries appeared in the output.

@eve-mem does this seem like a good solution? Am I missing anything?

eve-mem:
@dgmcdona i can safely say i don't have enough experience in windows to give you a useful opinion either way. I'm quite confident you know much more about it than I do!

dgmcdona:
I took a look into one of the cases where a new handle appeared - the offset produced by the old version of this plugin was off by 4. I think this is because of the mask value in `_decode_pointer`. If we're shifting right by `0x10` after applying a mask of `0xFFFFFFFFFFFFFFF8`, then we end up capturing some of the bits from another part of the bitfield, and the pointer doesn't end up 16-byte aligned, leading to the discrepancy. Shifting left by four instead guarantees that the pointer will be correctly aligned.

If we used a mask of `0xFFFFFFFFFFF00000`, then the lower 4 bits would be correctly zeroed, but I think the new approach is better anyway, since it accesses the pointer bits directly, and we don't have to rely on disassembly parsing with capstone, hoping that the compiler doesn't happen to use another `sar` instruction before the one we're looking for (and I've seen cases where this function does contain multiple `sar` instructions). And hopefully this will help us catch breaking changes in the future if the `ObjectPointerBits` member is removed but `LowValue` stays.

I think this is good to go.

volatility3/framework/plugins/windows/handles.py
```diff
@@ -135,78 +99,6 @@ def _get_item(self, handle_table_entry, handle_value):
         object_header.HandleValue = handle_value
         return object_header
 
-    def find_sar_value(self):
```
ikelos, line 138: This method doesn't start with an `_`.  It is an instance method, so there shouldn't be anyone calling this externally (and this is why I'd prefer things that are supposed to be internal only to be prepended with an `_`), but I really want to try and do things correctly rather than messing things up and getting a reputation as being unreliable because we made corner cases for our own rules.

As such, can this please be turned into a MAJOR version bump (rather than just a PATCH bump) due to the removal of this method, and all plugins that depend on it be updated to depend on the newer version (I think that's just `poolscanner`, `psxview`, `dumpfiles` and `callbacks` plugins).  Either that, or restore this deleted method (perhaps with a `DeprecationWarning` at the start?

ikelos (commented):
Really nice find, good work!  This simplifies so much stuff, I'm not sure we can sneak it in as a PATCH version bump though because it technically features the removal of a method that had been public (even though it wasn't a classmethod and should never have been called externally).  If we fix that (either bump it and consumers of it, or restore the removed function) then it can go in.  5:)

volatility3/framework/plugins/windows/handles.py
dgmcdona: Major version bumped on plugin + dependents 👍

ikelos:
Thanks very much for the speedy resolution!  5:)

PR #1390, opened 2024-12-10
Remove redundant part of if statement

j-t-1:
Also reorder imports.

### DIFF
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -555,8 +555,8 @@ class Volshell(interfaces.plugins.PluginInterface):
 
         for keyword, val in kwargs.items():
             if not isinstance(
-                val, interfaces.configuration.BasicTypes
-            ) and not isinstance(val, list):
+                val, (interfaces.configuration.BasicTypes, list)
+            ):
                 if all(
                     isinstance(x, interfaces.configuration.BasicTypes) for x in val
                 ):
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -554,11 +554,15 @@ class Volshell(interfaces.plugins.PluginInterface):
                 del kwargs[argname]
 
         for keyword, val in kwargs.items():
-            if not isinstance(val, (interfaces.configuration.BasicTypes, list)):
-                if all(isinstance(x, interfaces.configuration.BasicTypes) for x in val):
-                    raise TypeError(
-                        "Configurable values must be simple types (int, bool, str, bytes)"
-                    )
+            BasicType_or_list_of_BasicType = False  # excludes list of lists
+            if isinstance(val, interfaces.configuration.BasicTypes):
+                BasicType_or_list_of_BasicType = True
+            if all(isinstance(x, interfaces.configuration.BasicTypes) for x in val):
+                BasicType_or_list_of_BasicType = True
+            if not BasicType_or_list_of_BasicType:
+                raise TypeError(
+                    "Configurable values must be simple types (int, bool, str, bytes)"
+                )
             self.context.config[config_path + "." + keyword] = val
 
         constructed = clazz(self.context, config_path, **constructor_args)
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -11,17 +11,17 @@ import sys
 from typing import Any, Dict, Iterable, List, Optional, Tuple, Type, Union
 from urllib import parse, request
 
-from volatility3.cli import text_renderer, volshell
-from volatility3.framework import exceptions, interfaces, objects, plugins, renderers
-from volatility3.framework.configuration import requirements
-from volatility3.framework.layers import intel, physical, resources, scanners
-
 try:
     import capstone
 
     has_capstone = True
 except ImportError:
     has_capstone = False
+
+from volatility3.cli import text_renderer, volshell
+from volatility3.framework import exceptions, interfaces, objects, plugins, renderers
+from volatility3.framework.configuration import requirements
+from volatility3.framework.layers import intel, physical, resources, scanners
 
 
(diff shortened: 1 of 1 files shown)


j-t-1:
Please could you check the if statement re-write?

The original was:
```
        for keyword in kwargs:
            val = kwargs[keyword]
            if not isinstance(
                val, interfaces.configuration.BasicTypes
            ) and not isinstance(val, list):
                if not isinstance(val, list) or all(
                    isinstance(x, interfaces.configuration.BasicTypes) for x in val
                ):
                    raise TypeError(
                        "Configurable values must be simple types (int, bool, str, bytes)"
                    )
```

But maybe it should be:
 ```
        for keyword in kwargs:
            val = kwargs[keyword]
            if not isinstance(
                val, interfaces.configuration.BasicTypes
            ):
                if isinstance(val, list) and not all(
                    isinstance(x, interfaces.configuration.BasicTypes) for x in val
                ):
                    raise TypeError(
                        "Configurable values must be simple types (int, bool, str, bytes)"
                    )
```

**Comment by j-t-1** [comment]
1. Can this line be removed: # TODO: provide help, consider generic functions (pslist?) and/or providing windows/linux functions
2. Is the readline and rlcompleter imports needed post Python 3.4? Probably needed, but my computer is down so cannot currently check this.

File: volatility3/cli/volshell/generic.py [hunk 1]
```diff
@@ -553,14 +553,9 @@ def create_configurable(
...
-            val = kwargs[keyword]
-            if not isinstance(
-                val, interfaces.configuration.BasicTypes
-            ) and not isinstance(val, list):
-                if not isinstance(val, list) or all(
-                    isinstance(x, interfaces.configuration.BasicTypes) for x in val
-                ):
+        for keyword, val in kwargs.items():
+            if not isinstance(val, (interfaces.configuration.BasicTypes, list)):
+                if all(isinstance(x, interfaces.configuration.BasicTypes) for x in val):
```
gcmoreira: hm.. it seems the original implementation doesn't support `tuple`s, but yours does, the `if not isinstance(val, list) or` might be intentional. Additionally, if val is of type `float`, this might crash since it's neither in `BasicTypes`, a list, and is definitely not iterable. 

@ikelos On the other hand, we need to fix [this](https://github.com/volatilityfoundation/volatility3/blob/b8298a345fc5a3260ca2156d1e81a292530bde6e/volatility3/framework/interfaces/configuration.py#L788), where the values don't align with those in `BasicTypes`. Maybe, that line should use `BasicTypes` and `BasicTypes` needs to be updated too to also include `float` or ... ideally, we should have a dedicated function to centralize type validation for configuration values.

ikelos: So I don't believe floats were supposed to be allowed in config items (because SimpleTypes was supposed to be single point of truth for this), so they can *probably* be removed without problem from the config type verification.  I agree we should unify that list with `BasicTypes`.  I don't think there's necessarily point in a function for verification, but a shared list definitely makes sense (and the code you pointed out should be using it, the only reason I can think why it might legitimately not be using it is if it causes import loops).

Yeah, I've no idea why the ifs seemed to be run twice, but I did write it 7 years ago.  5;P  I'm not sure the replacement works correctly.  It's basically checking that everything's a basic type, or is a list of basic types (so no lists of lists, and no other complex types involved although vol types would be allowed because they all inherit from standard base types).

ikelos: The one you've got (and to be fair I think also the original) allows a list of anything to get through, which wasn't the intention).  The intention was either a `BasicType` or a list where each element is only a `BasicType` (and not a list).

j-t-1: This PR became a bit messy (the first commit allows a list of anything to get through), is it better if you close it and I do a new PR with just the change in the last commit?

From the idea of @gcmoreira above, may better to place in volatility3/framework/objects/utility.py.

volatility3/cli/volshell/generic.py
```diff
@@ -11,18 +11,18 @@
...
-from volatility3.framework.layers import intel, physical, resources, scanners
-
 try:
     import capstone
 
     has_capstone = True
 except ImportError:
     has_capstone = False
 
+from volatility3.cli import text_renderer, volshell
```
ikelos, line 21: I don't *think* the position of these affects anything?  The attempt to import capstone just sets a variable (which other modules shouldn't be using.  Was there a specififc reason for this shift, or just tidying up?  Generally I try to put all the code below all the imports unless there's a good reason?

volatility3/cli/volshell/generic.py
```diff
@@ -553,17 +553,16 @@ def create_configurable(
...
-            ) and not isinstance(val, list):
-                if not isinstance(val, list) or all(
-                    isinstance(x, interfaces.configuration.BasicTypes) for x in val
-                ):
-                    raise TypeError(
-                        "Configurable values must be simple types (int, bool, str, bytes)"
-                    )
+        for keyword, val in kwargs.items():
+            BasicType_or_list_of_BasicType = False  # excludes list of lists
+            if isinstance(val, interfaces.configuration.BasicTypes):
```
ikelos, line 558: Feels like his could be compressed into an or without need for an additional variable, but I'm fine if you prefer doing it explicitly.

ikelos (commented):
Thanks, this looks ok now if you're happy with it?

volatility3/cli/volshell/generic.py
j-t-1: It was to put in the order of standard libraries, 3rd party libraries, then ours. But I do not mind, and am fine with it remaining as it was. Could run the codebase through isort in the future.

j-t-1:
Is it easy for you to revert the imports change, and then commit? Or let me know and can do a completely completely new PR, or an additional commit to revert the imports change. Or feel free to commit as is.

PR #1391, opened 2024-12-10
Slightly modify documentation

j-t-1:
Include regex_scan, new functionality of volshell.

### DIFF
--- a/doc/source/glossary.rst
+++ b/doc/source/glossary.rst
@@ -61,9 +61,7 @@ Map, mapping
     of the :ref:`Range<range>`).  Mappings can be seen as a mathematical function, and therefore volatility 3
     attempts to use mathematical functional notation where possible.  Within volatility a mapping is most often
     used to refer to the function for translating addresses from a higher layer (domain) to a lower layer (range).
-    For further information, please see
-    `https://en.wikipedia.org/wiki/Function_(mathematics)`.
-
+    For further information, please see `[Function (mathematics) in Wikipedia](https://en.wikipedia.org/wiki/Function_(mathematics))`.
 
 .. _Member:
 
--- a/doc/source/glossary.rst
+++ b/doc/source/glossary.rst
@@ -123,6 +123,11 @@ Page Table
     possible to use them as a way to map a particular address within a (potentially larger, but sparsely populated)
     virtual space to a concrete (and usually contiguous) physical space, through the process of :ref:`mapping<map>`.
 
+.. _Plugin:
+
+Plugin
+    Plugins are the "functions" of the volatility framework. They carry out algorithms on data stored in layers using objects constructed from symbols.  Broadly, plugins take in a number of TranslationLayers (the data, which is a representation of part of an image, in a specified type described by templates) and outputs a TreeGrid.
+
 .. _Pointer:
 
 Pointer
--- a/doc/source/glossary.rst
+++ b/doc/source/glossary.rst
@@ -61,7 +61,7 @@ Map, mapping
     of the :ref:`Range<range>`).  Mappings can be seen as a mathematical function, and therefore volatility 3
     attempts to use mathematical functional notation where possible.  Within volatility a mapping is most often
     used to refer to the function for translating addresses from a higher layer (domain) to a lower layer (range).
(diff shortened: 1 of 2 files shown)


eve-mem:
Hey @j-t-1

Nice one! 

It might be a good idea to include an example use and output of rx as well if you can. What do you think?

j-t-1:
I think yes, and it's a brave move trying to do so with reStructuredText.

We could take your [example](https://github.com/volatilityfoundation/volatility3/pull/1347#issue-2651288363
):

re(rb"(Linux version|Darwin Kernel Version) [0-9]+\.[0-9]+\.[0-9]+")
 
amend it (does this work, or revert to the three digits?):

rx(rb"(Linux version) [0-9.]+")

and just take the first of the four blocks of output?

eve-mem:
I think just including one example should be okay personally. E.g. keep the full regex but only include one example really so that it doesn't take up too much space?

doc/source/volshell.rst
```diff
@@ -180,15 +180,29 @@ used:
...
 to `context.layers` and can be accessed by the name returned by `lf`.
+
+Regex
+^^^^^
+
+It is easy to scan for some bytes or a pattern using `regex_scan` or `rx`.
+
+An optional size can be given for the displayed results as with the other fuctions (db, dw, dd, dq, etc).
+
+You can of course specify a different layer name as well.
```
ikelos: This should probably be `You can, of course, specify...` if we're happy with the colloquialism.

ikelos (approved):
Yep, looks good!  Thanks for poring through our documentation, some of those typos have been there for years!  I'll apply this in a day in case you want to change it before it goes in.  I'm happy either way, my comment is such a minor point it's hardly worth it.  5:)

j-t-1:
Will update this commit and would like to add this to the glossary.rst:

Intermediate Symbol File (ISF).  They contain kernel debugging information formatted as JSON. For macOS and Linux analysis, the kernel needs to be located or added as an ISF file to the volatility 3 symbols directory. For Windows, the required ISF file can often be generated from PDB files automatically downloaded from Microsoft servers, and therefore does not require manual intervention.

Is this okay? Please amend in any way.

doc/source/volshell.rst
j-t-1: Have changed it (better), maybe would be best without the "of course". Is commit squashing for pull requests implemented? Then can make small increments like this without undue credit.

ikelos:
I'd prefer: 

`...They contain kernel structures and specific offsets formatted as JSON. For macOS and Linux analysis, the kernel needs to be added as an ISF file to the volatility 3 symbols directory...`

Otherwise yep, looks good.

j-t-1:
Add Plugin to glossary:

Plugins are the "functions" of the volatility framework. They carry out algorithms on data stored in layers using objects constructed from symbols. Broadly, plugins take in a number of TranslationLayers (the data, which is a representation of part of an image, in a specified type described by templates) and outputs a TreeGrid.

Same again check please. This is a precis from volatility3/framework/interfaces/plugins.py.

doc/source/glossary.rst
```diff
@@ -55,7 +62,7 @@ Map, mapping
     attempts to use mathematical functional notation where possible.  Within volatility a mapping is most often
     used to refer to the function for translating addresses from a higher layer (domain) to a lower layer (range).
     For further information, please see
-    `Function (mathematics) in wikipedia https://en.wikipedia.org/wiki/Function_(mathematics)`
+    `https://en.wikipedia.org/wiki/Function_(mathematics)`.
```
ikelos: I think this should be `[Function (mathematics) in wikipedia](https://en.wikipedia.org/wiki/Function_(mathematics))`, although I'm not sure how that'll work with brackets in the URL...

ikelos (approved):
Yep, the new section looks good, just want to get the link to Function on wikipedia working as intended (as opposed to how it originally was, which I think wasn't working).  5;)

j-t-1:
> Yep, the new section looks good, just want to get the link to Function on wikipedia working as intended (as opposed to how it originally was, which I think wasn't working). 5;)

This has a quick reference [devguide](https://devguide.python.org/documentation/markup/), and will change this to the external links format.

ikelos (approved):

Issue #1392, opened 2024-12-11
linux.pagecache plugins breaks dumping convention, needs to be converted


atcuno:
This plugin uses its own method to extract a file which really breaks the flow of people trying to learn Volatility 3 as well as trying to teach it to them. It also breaks the nice feature `-o` for an output directory.

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/pagecache.py#L521

```
Volatility 3 Framework 2.11.0
usage: volatility linux.pagecache.InodePages [-h] [--find FIND] [--inode INODE] [--dump DUMP]

optional arguments:
  -h, --help     show this help message and exit
  --find FIND    Filename (full path) to find
  --inode INODE  Inode address
  --dump DUMP    Output file path
```

So people send in a string as a `--dump` argument.

The plugin needs to use `--dump` like all the other plugins for consistency.

atcuno:
@ikelos the code for this looks good but I am not sure what, if any, version bumps are required. Otherwise its ready

gcmoreira:
@ikelos this can be closed, thanks

PR #1393, opened 2024-12-12
Linux: Update inode pagecache plugin to conform to framework dumping convention

gcmoreira:
This PR addresses #1392 issue

```shell
$ mkdir output_dir
$ ./vol.py \
    --output output_dir \
    -f ./data.lime \
    linux.pagecache.InodePages \
    --inode 0x9d9a457753e0 \
    --dump 
Volatility 3 Framework 2.12.0
PageVAddr       PagePAddr       MappingAddr     Index   DumpSafe        Flags

0xdd8747f885c0  0x1fe217000     0x9d9a45775558  0       True    dirty,lru,referenced,savepinned,swapbacked,uptodate
0xdd8747f87e80  0x1fe1fa000     0x9d9a45775558  1       True    dirty,lru,referenced,savepinned,swapbacked,uptodate
0xdd8747f87f40  0x1fe1fd000     0x9d9a45775558  2       True    dirty,lru,referenced,savepinned,swapbacked,uptodate

$ sha1sum output_dir/inode_0x9d9a457753e0.dmp 
f17f82631f0ace010884ff58cbd8c90ddd89a782  output_dir/inode_0x9d9a457753e0.dmp
```

### DIFF
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -389,7 +389,7 @@ class InodePages(plugins.PluginInterface):
 
     _required_framework_version = (2, 0, 0)
 
-    _version = (1, 0, 1)
+    _version = (2, 0, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
@@ -412,9 +412,10 @@ class InodePages(plugins.PluginInterface):
                 description="Inode address",
                 optional=True,
             ),
-            requirements.StringRequirement(
+            requirements.BooleanRequirement(
                 name="dump",
-                description="Output file path",
+                description="Extract inode content",
+                default=False,
                 optional=True,
             ),
         ]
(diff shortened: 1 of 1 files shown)


ikelos:
@atucno all these are likely to miss the next release (due to be rolled this week) because they haven't gone through code freeze or suitable enough testing.  If there's any critical changes required, please make a case and we can review them on a one-by-one basis, but otherwise these will be for the March/April release.

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -520,8 +521,13 @@ def _generator(self):
 
         if self.config["dump"]:
             filename = self.config["dump"]
```
ikelos: This line can be removed since the `dump` option is now a boolean.

ikelos (commented):
One minor point, otherwise all looks good.

atcuno:
That is fine. My main time concern is to have the full parity release cut by mid/late March (free freeze Feb), so we are pushing hard on that.

volatility3/framework/plugins/linux/pagecache.py
gcmoreira: thanks.. good catch! I just removed it. It should be all set once the test cases complete successfully.

ikelos (approved):
Looks good.  When a MAJOR bump happens we need to get in the habit of checking whether any plugins depend on it, but this one seem clear.

PR #1397, opened 2024-12-12
Modify using as a library documentation

j-t-1:
Tiny changes.

### DIFF
--- a/doc/source/using-as-a-library.rst
+++ b/doc/source/using-as-a-library.rst
@@ -3,7 +3,7 @@ Using Volatility 3 as a Library
 
 This portion of the documentation discusses how to access the Volatility 3 framework from an external application.
 
-The general process of using volatility as a library is to as follows:
+The general process of using volatility as a library is as follows:
 
 1. :ref:`create_context`
 2. (Optional) :ref:`available_plugins`
@@ -21,7 +21,7 @@ Creating a context
 First we make sure the volatility framework works the way we expect it (and is the version we expect).  The
 versioning used is semantic versioning, meaning any version with the same major number and a higher or equal
 minor number will satisfy the requirement.  An example is below since the CLI doesn't need any of the features
-from versions 1.1 or 1.2:
+from version 1.1:
 
 ::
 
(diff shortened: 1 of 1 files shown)


doc/source/using-as-a-library.rst
```diff
@@ -21,7 +21,7 @@ Creating a context
 First we make sure the volatility framework works the way we expect it (and is the version we expect).  The
 versioning used is semantic versioning, meaning any version with the same major number and a higher or equal
 minor number will satisfy the requirement.  An example is below since the CLI doesn't need any of the features
-from versions 1.1 or 1.2:
+from version 1.1:
```
ikelos: Perhaps, `version 1.1 or later`?

doc/source/using-as-a-library.rst
```diff
@@ -139,7 +139,7 @@ A suitable list of automagics for a particular plugin (based on operating system
 
 This will take the plugin module, extract the operating system (first level of the hierarchy) and then return just
 the automagics which apply to the operating system.  Each automagic can exclude itself from being used for specific
-operating systems, so that an automagic designed for linux is not used for windows or mac plugins.
+operating systems, so that for example an automagic designed for linux is not used for windows or mac plugins.
```
ikelos: I'm not sure this is suitably placed?  It should also be between parenthetical commas really?  Perhaps `such that an automagic` rather than `so that an automagic`?

doc/source/using-as-a-library.rst
```diff
@@ -157,16 +157,16 @@ Any exceptions that occur during the execution of the automagic will be returned
...
+have been successfully set).  We do this as follows, where plugin_config_path is the base_config_path (which defaults
 to `plugins` and then the name of the class itself):
 
 ::
 
     unsatisfied = plugin.unsatisfied(context, plugin_config_path)
 
 If unsatisfied is an empty list, then the plugin has been given everything it requires.  If not, it will be a
-Dictionary of the hierarchy paths and their associated requirements that weren't satisfied.
+dictionary of the hierarchy paths and their associated requirements that weren't satisfied.
```
ikelos: dictionary is a standard python class, so to be specific it should be `dict` rather than dictionary?

ikelos (commented):
Thanks very much for all this error checking!  Just a few comments on this one, but it's all much appreciated stuff!  Once we're happy with the final changes I'll get it merged in...  5:)

doc/source/using-as-a-library.rst
j-t-1: Changed, is kinda like explicit is better than implicit.

PR #1398, opened 2024-12-12
Update the CLI manual documentation

j-t-1:


j-t-1:
Please check this modification:
> --offline
>     Run offline mode (defaults to false).  Do not search online for additional JSON files, remote windows symbol tables, nor linux/mac banner repositories.

doc/source/vol-cli.rst
```diff
@@ -67,12 +67,12 @@ Options
...
     files or symbol zip packs.  Symbols must be within a particular directory
-    structure if they depending on the operating system of the symbols,
+    structure if they depend on the operating system of the symbols,
     whilst symbol packs must be in the root of the directory and named after
-    the after the operating system to which they apply.
+    the operating system to which they apply.
 
 -v, --verbose
-    A flag which can be used multiple times, each time increasing the level of
+    A flag which can be used multiple times (up to six, -vvvvvv), each time increasing the level of
```
ikelos, line 75: Well, whaddya know.  I always thought all this time it was 7, but I did the maths and you're right, six is the most!  5:D

doc/source/vol-cli.rst
```diff
@@ -120,9 +120,7 @@ Options
     Change the default path used to store the cache.
 
 --offline
-    Do not search online for additional JSON files.
-    Run offline mode (defaults to false) and for
-    remote windows symbol tables, linux/mac banner repositories. 
+    Run offline mode (defaults to false).  Do not search online for additional JSON files, remote windows symbol tables, nor linux/mac banner repositories.
```
ikelos, line 123: Yep, I think this is right.  The goal is actually to shut down any use of the internet at all (for running in offline mode) and I *think* that's all the instances where it may be relevant, but I guess listing them explicitly is safer (because an user could include a URL in some way and the opener would try to go out to the URL automatically, whether `--offline` is set or not), so this is good.  5:)

ikelos (approved):
Yep, all looks good.  I had to check my maths, but you were right, thanks very much for keeping me on my toes!  5:D

Issue #1385, opened 2024-12-07
Error analyzing Windows 11 VB dump when looking up `_MM_SESSION_SPACE`


aleguma:
**Bug**
I am trying to get the information from a Windows 11 VirtualBox VM dump (saved as .raw using the method described here https://d33pdiv3r.com/2024/05/22/virtualbox-vm-memory-dump-easy-step-by-step-tutorial/). When I try to use **windows.pslist** or **windows.pstree** plugins, the output is incomplete and the analysis crashes.

**Context**
Volatility Version:  3 Framework 2.12.0 (latest version available as of 7/12/24, same day as this post)
Operating System:  Windows 10
Python Version:  3.13
Suspected Operating System: Windows 11 
Command:  **python vol.py -f TEST.raw windows.pslist**

**To Reproduce**

1. I create a VM in VB, install Windows 11, and get the memory dump as described in https://d33pdiv3r.com/2024/05/22/virtualbox-vm-memory-dump-easy-step-by-step-tutorial/. 
2. I installed Volatility as instructed, including placing the zipped windows symbols in the path volatility3/symbols.
3. Use command 'python vol.py -vvvvv -f TEST.raw windows.pslist'

**Expected behavior**
I expected to get the full list of running processes in the machine.

**Example output**
> INFO     volatility3.cli: Volatility plugins path: ['C:\\Users\\Administrator\\Desktop\\volatility3-develop\\volatility3\\plugins', 'C:\\Users\\Administrator\\Desktop\\volatility3-develop\\volatility3\\framework\\plugins']
> INFO     volatility3.cli: Volatility symbols path: ['C:\\Users\\Administrator\\Desktop\\volatility3-develop\\volatility3\\symbols', 'C:\\Users\\Administrator\\Desktop\\volatility3-develop\\volatility3\\framework\\symbols']
> DETAIL 3 volatility3.cli: Cache directory used: C:\Users\Administrator\AppData\Roaming\volatility3
> INFO     volatility3.framework.automagic: Detected a windows category plugin
> INFO     volatility3.framework.automagic: Running automagic: ConstructionMagic
> DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
> DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
[...]

PR #1399, opened 2024-12-13
Windows: Handle missing _MM_SESSION_SPACE

dgmcdona:
refs #1385

As of Windows 11 24H2, the `_MM_SESSION_SPACE` type no longer appears in the kernel PDB. Instead, the `_EPROCESS.Session` member refers to a new, opaque type, `_PSP_SESSION_SPACE`. However, experimentation has shown that this new structure is functionally identical to the old structure - the `ProcessList` and `SessionId` members still appear to be at their old offsets. In order to account for this when analyzing these newer Windows versions, this catches the `SymbolError` and instantiates an `unsigned long` at the offset (8) where the `SessionId` member would normally be defined within an `_MM_SESSION_SPACE` structure.

This fixes plugins that rely on this member, including `windows.pslist` and `windows.verinfo`

@ikelos I considered adding a new symbol table with this `_PSP_SESSION_SPACE` structure, but this was the quick and easy solution - let me know if you'd prefer an approach like that instead.

### DIFF
--- a/volatility3/framework/plugins/windows/modules.py
+++ b/volatility3/framework/plugins/windows/modules.py
@@ -165,13 +165,33 @@ class Modules(interfaces.plugins.PluginInterface):
 
                 # create the session space object in the process' own layer.
                 # not all processes have a valid session pointer.
-                session_space = context.object(
-                    symbol_table + constants.BANG + "_MM_SESSION_SPACE",
-                    layer_name=layer_name,
-                    offset=proc.Session,
-                )
+                try:
+                    session_space = context.object(
+                        symbol_table + constants.BANG + "_MM_SESSION_SPACE",
+                        layer_name=layer_name,
+                        offset=proc.Session,
+                    )
+                    session_id = session_space.SessionId
 
-                if session_space.SessionId in seen_ids:
+                except exceptions.SymbolError:
+                    # In Windows 11 24H2, the _MM_SESSION_SPACE type was
+                    # replaced with _PSP_SESSION_SPACE, and the kernel PDB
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/windows/modules.py
```diff
@@ -165,13 +165,33 @@ def get_session_layers(
...
+                    # In Windows 11 24H2, the _MM_SESSION_SPACE type was
+                    # replaced with _PSP_SESSION_SPACE, and the kernel PDB
+                    # doesn't contain information about its members (otherwise,
+                    # we would just fall back to the new type). However, it
+                    # appears to be, for our purposes, functionally identical
+                    # to the _MM_SESSION_SPACE. Because _MM_SESSION_SPACE
+                    # stores its session ID at offset 8 as an unsigned long, we
+                    # create an unsigned long at that offset and use that
+                    # instead.
+                    session_id = int(
```
ikelos: This shouldn't need coercing to an int, is there a reason for doing this?  Since the base types should be based off of normal ints, comparison and equality (for example, for set membership) *should* all work normally (but please check, if it doesn't it might be better to fix that for the whole base type rather than doing casts).

ikelos (commented):
Looks reasonable, just want to make sure we're casting to `int` for a reason, otherwise it'll set a bad precedent (and perpetuate people's distrust of the volatility object system).

volatility3/framework/plugins/windows/modules.py
dgmcdona: Fixed, added a type hint to the method signature as well.

PR #1402, opened 2024-12-15
Refactor: move version `None` check to top

TheMythologist:
Otherwise, this will cause an error when generating the default description in the `if description is None` branch:
```
TypeError: 'NoneType' object is not iterable
```


ikelos (approved):
Good spot, thanks!  5:)

PR #1406, opened 2024-12-16
Remove use of int function after math.ceil

j-t-1:
Return type of math.ceil is already an int.


ikelos (approved):
Good spot, I think I was worried it might return a float, but `ceil` does have to return an int.  5:)

PR #1288, opened 2024-10-03
Add Linux ptrace plugin

gcmoreira:
This PR adds the `linux.ptrace` plugin to enumerate tracer and tracee tasks. Ptrace is often leveraged by attackers to gather credentials and other sensitive information.

## Example 1: strace
Let's take a random process thread:
```shell
$ ps -eLf
UID        PID  PPID   LWP  C NLWP STIME TTY          TIME CMD
root       955     1  1047  0    3 05:34 ?        00:00:00 /usr/lib/accountsservice/accounts-daemon
...
```

Let's ptrace the thread/task 1047
```shell
$ sudo strace -p 1047
...
```

```shell
$ ps -eLf
UID        PID  PPID   LWP  C NLWP STIME TTY          TIME CMD
root      2963  2961  2963  0    1 05:38 pts/1    00:00:00 strace -p 1047
...
```

```shell
$ sudo cat /proc/1047/status
Name:	gmain
Umask:	0022
State:	S (sleeping)
Tgid:	955
Ngid:	0
Pid:	1047
PPid:	1
TracerPid:	2963
...
```

```shell
 $ ./vol.py -r pretty \
    -f ../dump_ubuntu180464bit_5.4.0-117-generic_ptrace.core \
    linux.ptrace 
Volatility 3 Framework 2.10.0        
  | Process |  PID |  TID | Tracer TID | Tracee TID |                                                            Flags
* |   gmain |  955 | 1047 |       2963 |          - | PT_SEIZED|PT_TRACE_EXIT|PT_TRACE_EXEC|PT_TRACESYSGOOD|PT_PTRACED
* |  strace | 2963 | 2963 |          - |       1047 |                                                                -
```

# Example 2: gdb
```shell
$ sudo gdb -p 1047
```

```
$ ps -eLf | grep gdb
UID        PID  PPID   LWP  C NLWP STIME TTY          TIME CMD
...
root      4278  4277  4278  1    1 09:42 pts/1    00:00:00 gdb -p 1047
```

```shell
$ sudo cat /proc/1047/status
Name:	gmain
Umask:	0022
State:	t (tracing stop)
Tgid:	955
Ngid:	0
Pid:	1047
PPid:	1
TracerPid:	4278
...
```

```shell
$ ./vol.py -r pretty \
    -f ./dump_ubuntu180464bit_5.4.0-117-generic_ptrace_gdb.core \
    linux.ptrace 
Volatility 3 Framework 2.10.0
[...]

PR #1412, opened 2024-12-17
Linux: Fix task parent pid in several plugins

gcmoreira:
Linux plugins are using the wrong value for the parent process identifier (PPID).
Using `task.parent` can result in inaccurate results, especially in cases involving ptraced processes (refer to [PR #1288](https://github.com/volatilityfoundation/volatility3/pull/1288) for more details).

This PR introduces a new method in the task object extension that mimics the Linux kernel's [getppid](https://elixir.bootlin.com/linux/v6.12.5/source/kernel/sys.c#L984) syscall, ensuring correct and consistent retrieval of the parent PID.

### DIFF
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -635,6 +635,20 @@ class task_struct(generic.GenericIntelProcess):
         # root time namespace, not within the task's own time namespace
         return boottime + task_start_time_timedelta
 
+    def get_parent_pid(self) -> int:
+        """Returns the parent process ID (PPID)
+
+        This method replicates the Linux kernel's `getppid` syscall behavior.
+        Avoid using `task.parent`; instead, use this function for accurate results.
+        """
+
+        if self.real_parent and self.real_parent.is_readable():
+            ppid = self.real_parent.pid
+        else:
+            ppid = 0
+
+        return ppid
+
 
 class fs_struct(objects.StructType):
     def get_root_dentry(self):
--- a/volatility3/framework/constants/_version.py
+++ b/volatility3/framework/constants/_version.py
@@ -1,6 +1,6 @@
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 2  # Number of releases of the library with a breaking change
-VERSION_MINOR = 12  # Number of changes that only add to the interface
+VERSION_MINOR = 13  # Number of changes that only add to the interface
 VERSION_PATCH = 0  # Number of changes that do not change the interface
 VERSION_SUFFIX = ""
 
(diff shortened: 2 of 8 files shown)


volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -635,6 +635,20 @@ def get_create_time(self) -> datetime.datetime:
...
+        """Returns the parent process ID (PPID)
+
+        This method replicates the Linux kernel's `getppid` syscall behavior.
+        Avoid using `task.parent`; instead, use this function for accurate results.
+        """
+
+        if self.real_parent and self.real_parent.is_readable():
+            ppid = self.real_parent.pid
+        else:
+            ppid = 0
```
ikelos, line 648: Is this a suitable value to return if the `real_parent` can't be read?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -635,6 +635,20 @@ def get_create_time(self) -> datetime.datetime:
         # root time namespace, not within the task's own time namespace
         return boottime + task_start_time_timedelta
 
+    def get_parent_pid(self) -> int:
```
ikelos, line 638: I think technically this is an additive change, and there's no versioning on structs, so it's probably a MINOR bump on the framework version please?  That'll also mean all the required_framework_versions will need upping to whatever this ends up being (2.13.0 mebbe?)

ikelos (commented):
Thanks for fixing all the consumer plugins and bumping their versions.  I think there's still some version work to be done, and I'm a little concerned about the return pid if there's an error (rather than just throwing an exception).  Really don't like papering over problems and particularly when the caller of the function can't differentiate the error condition from an actual valid response.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Yes, that's correct, and it aligns with how `InvalidAddressException` was handled in certain cases before these changes. However, in other scenarios, such as pstree, the exception wasn't handled, causing crashes.

As noted in [PR #1243](https://github.com/volatilityfoundation/volatility3/pull/1243), simply checking the `proc.parent` boolean is insufficient for validating a pointer. This fix ensures plugins can continue processing remaining data even when a task's parent pointer is invalid.

It's important to clarify that determining whether a task itself is valid should be done through other methods like `is_valid()`. The purpose of this function is not to handle task validity checks.

ikelos: Certainly, it was more whether 0 could be an *actual* value, and we'd want an error condition to be something like -1?

gcmoreira: I don't think so. The original code always defaulted to 0 in such cases, so changing it would alter the existing behavior of these plugins. If future scenarios require a different value, those cases should be explicitly identified and handled within the code. Alternatively, we could add a parameter to allow setting this as a default value, if preferred.

ikelos: Ok, as long as you're happy with it.  I guess since init is pid 1, you shouldn't ever get a legit ppid of 0.  I don't think it needs a parameter, just wanted to check that it couldn't be confused with a valid value somehow...

PR #1369, opened 2024-11-28
Linux/Mac: Log producer metadata information

gcmoreira:
This PR adds ISF metadata to log lines, including the producer name, version, types, and symbol sources for Mac and Linux. This makes it easier to understand the user's setup when they report an issue, saving time in troubleshooting. 
The information will be available at the `debug` log level, so `-vv`, `-vvv` and so on.

## Example 1 - Current GitHub ISF
This is the current `linux-sample-1.bin` ISF we are using for testing on GitHub, generated by `dwarf2json 0.9.0`. The types and symbols were extracted from `vmlinux-3.2.0-4-amd64`.

```shell
$ ./vol.py -vvv \
    -f ./linux-sample-1.bin \
    linux.pslist
Volatility 3 Framework 2.12.0
...
DEBUG    volatility3.framework.automagic.symbol_finder: Identified banner: b'Linux version 3.2.0-4-amd64 (debian-kernel@lists.debian.org) (gcc version 4.6.3 (Debian 4.6.3-14) ) #1 SMP Debian 3.2.57-3+deb7u2\n\x00'
DEBUG    volatility3.framework.automagic.symbol_finder: Using symbol library: file:///home/gmoreira/vol3_profiles/linux-image-3.2.0-4-amd64_3.2.57-3_deb7u2.json.xz
DEBUG    volatility3.framework.automagic.symbol_finder: producer_name: dwarf2json, producer_version: 0.9.0
DEBUG    volatility3.framework.automagic.symbol_finder: Types:
DEBUG    volatility3.framework.automagic.symbol_finder:         {'kind': 'dwarf', 'name': 'vmlinux-3.2.0-4-amd64', 'hash_type': 'sha256', 'hash_value': '67f9060253c801bffd3c99615c41d0ec266f8284c755f5368dad95b021481962'}
DEBUG    volatility3.framework.automagic.symbol_finder: Symbols:
DEBUG    volatility3.framework.automagic.symbol_finder:         {'kind': 'dwarf', 'name': 'vmlinux-3.2.0-4-amd64', 'hash_type': 'sha256', 'hash_value': '67f9060253c801bffd3c99615c41d0ec266f8284c755f5368dad95b021481962'}
DEBUG    volatility3.framework.automagic.symbol_finder:         {'kind': 'symtab', 'name': 'vmlinux-3.2.0-4-amd64', 'hash_type': 'sha256', 'hash_value': '67f9060253c801bffd3c99615c41d0ec266f8284c755f5368dad95b021481962'}
[...]

PR #1414, opened 2024-12-17
Fix ProducerMetadata class bug introduced in #1369

gcmoreira:
Since only the `hidden_modules` and `kmesg` plugins use the affected method, and there are no test cases for either, the bug wasn't triggered on my end. I'll be adding test cases for these, as well as for the remaining Linux plugins, to prevent similar issues from occurring in the future.


ikelos (approved):
Good catch, I missed it too...

PR #1415, opened 2024-12-17
Fix: Error early if no inodes are found in linux.pagecache.InodePages plugin

TheMythologist:
If not, this will cause an error 5 lines down when running `if not inode.is_valid()`:
```
UnboundLocalError: cannot access local variable 'inode' where it is not associated with a value
```

### DIFF
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -483,6 +483,9 @@ class InodePages(plugins.PluginInterface):
                 if inode_in.path == self.config["find"]:
                     inode = inode_in.inode
                     break  # Only the first match
+            else:
+                vollog.error("Unable to find inode with path %s", self.config["find"])
+                return None
 
         elif self.config["inode"]:
             inode = vmlinux.object("inode", self.config["inode"], absolute=True)


gcmoreira:
yes, good catch! thanks for reporting this

atcuno:
I wanted to leave a note here that we are looking into a way for CI to catch bugs like these automatically.

As shown here, PyCharm detects it auto while coding:

pycharm

And @dgmcdona suggested pyright from the command line as it detects it too:

```
/home/rk/volatility3/volatility3/framework/plugins/linux/pagecache.py:492:16 - error: "inode" is possibly unbound (reportPossiblyUnboundVariable)
  /home/rk/volatility3/volatility3/framework/plugins/linux/pagecache.py:493:51 - error: "inode" is possibly unbound (reportPossiblyUnboundVariable)
  /home/rk/volatility3/volatility3/framework/plugins/linux/pagecache.py:496:16 - error: "inode" is possibly unbound (reportPossiblyUnboundVariable)
  /home/rk/volatility3/volatility3/framework/plugins/linux/pagecache.py:500:22 - error: "inode" is possibly unbound (reportPossiblyUnboundVariable)
  /home/rk/volatility3/volatility3/framework/plugins/linux/pagecache.py:501:25 - error: "inode" is possibly unbound (reportPossiblyUnboundVariable)
  /home/rk/volatility3/volatility3/framework/plugins/linux/pagecache.py:523:62 - error: "inode" is possibly unbound (reportPossiblyUnboundVariable)
  /home/rk/volatility3/volatility3/framework/plugins/linux/pagecache.py:524:46 - error: "inode" is possibly unbound (reportPossiblyUnboundVariable)
```

So by parity release time, hopefully CI will help us detect this issues before merge.

ikelos (approved):
Yep, looks good, thanks!

TheMythologist:
I use mypy for my personal projects, perhaps we can use that?

ikelos:
I don't think our typing information is strong enough yet to support that.  A lot of our functions can only say they'll return a volatility object and we use that quite a bit to return objects of all different kinds.  It would be fine to run it over the codebase, but not as something that could stop the build process (as other linting runs do).

PR #1416, opened 2024-12-17
Feature: Multiple REMOTE_ISF_URLs

TheMythologist:
Also removes duplicates of word "the"

### DIFF
--- a/doc/source/getting-started-linux-tutorial.rst
+++ b/doc/source/getting-started-linux-tutorial.rst
@@ -40,7 +40,7 @@ For plugin requests, please create an issue with a description of the requested 
         linux.check_creds.Check_creds
         linux.check_idt.Check_idt
 
-.. note:: Here the the command is piped to grep and head to provide the start of the list of linux plugins.
+.. note:: Here the command is piped to grep and head to provide the start of the list of linux plugins.
 
 
 Using plugins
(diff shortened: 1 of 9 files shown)


ikelos (commented):
Thanks, could you provide a rationale for why having multiple remote ISF urls would be useful?  My concern is the order of application, which I can see you've reversed to prioritize the first one.  This then also requires any users that have specified it in their default options to change their configurations (and know that the first one in the list takes precedence, which might be confusing to people).  I'd just like to talk/think this through a bit more before going ahead with it please...

TheMythologist:
My rationale for this was that we could provide multiple ISF URLs to reference, especially if you're using online ones that are curated by the community (not by yourself). I could also edit the code to be backwards compatible if that is your concern. Alternatively, if you don't see the need for this feature, feel free to close this PR entirely 😅

ikelos:
Just as an update to this, we did include a means to chain ISF tables, together, so in version 1 of the RemoteIdentifierFormat, we included an `additional` field, that could be used to provide other ISF locations to check.  This means that a custom server can be provided that then eventually includes the main table (or include the main table and then includes another table with overriding symbols).

https://github.com/volatilityfoundation/volatility3/blob/674996ffcddf808ff73707886ec4a5c8645801a3/volatility3/framework/automagic/symbol_cache.py#L579

Hopefully this would help you achieve the original goal you had?

PR #1422, opened 2024-12-17
Windows: PeDump - use contextmanager

dgmcdona:
Several tools, including pyright and PyCharm, report that `file_handle`
may be an unbound local. Regardless of whether or not this is likely to
happen in practice, it makes sense to just use a `ContextManager` here
anyway, since `FileHandlerInterface` implements it.


ikelos (approved):
Yeah, I'm ok with that.  Might be interesting to see if there's other places/plugins where an `open_method` could be used as a `ContextManager`?

PR #1423, opened 2024-12-17
Windows Extensions: Fixes type-hint on list_timers

dgmcdona:
This method is incorrectly type-hinted as returning a `Tuple` when it
should be returning an instance of the `KTIMER` extension class. This
leaves the version number as is since it only updates the type-hint, but
let me know if that's incorrect and we need to bump it.

### DIFF
--- a/volatility3/framework/plugins/windows/timers.py
+++ b/volatility3/framework/plugins/windows/timers.py
@@ -14,7 +14,7 @@ from volatility3.framework import (
 )
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints
-from volatility3.framework.symbols.windows import versions
+from volatility3.framework.symbols.windows import versions, extensions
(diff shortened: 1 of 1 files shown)


ikelos:
I've just had a case myself where the typing for an interface was wrong, but since the interface is never directly instantiated, I changed that and without a version bump.  I think that's ok, but it should probably be context dependent.  Since the call couldn't have worked otherwise we're probably fine, but might be worth bumping the PATCH (third) number, just so we can tell which version someone is using if we need to figure out whether it relates to a problem we encounter...

ikelos (approved):
Should be fine (I can't see it making any import loops).  Looks good, thanks!  Lemme know if you want to bump the PATCH version of the plugin or not...

dgmcdona:
> Should be fine (I can't see it making any import loops). Looks good, thanks! Lemme know if you want to bump the PATCH version of the plugin or not...

Sounds good, I'll go ahead and do that.

PR #1424, opened 2024-12-17
Use generator expressions

j-t-1:
PEP 289 – Generator Expressions.


ikelos (approved):
Thanks, I think these all come with a slight efficiency boost over creating a list object.  I'm going to pause applying new patches until after #1364 lands, but after that if this is still needed I'll apply it.  5:)

PR #1427, opened 2024-12-18
Linux - Envars plugin: Add function to retrieve environment variables for a specific task

### LINKED ISSUE #1411: There is no API for getting a process' environment variables on Linux
The envar plugin on Linux does all the 'work' directly in generator, which makes it impossible for other plugins to re-use:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/envars.py#L74

We broke this out during a training so students could write + use a plugin for auto detecting maliciously set environment variables.

@gcmoreira we threw this together based on the code in generator in the plugin:

https://github.com/volatilityfoundation/volatility3/commit/2c430c9d1b94c99b71862a99deba7419ddaa7c59

Could you make this properly PR ready (typing, doc strings, etc.) and then make a PR after testing?

gcmoreira:
This PR addresses #1411.
It also includes some code improvements.

### DIFF
--- a/volatility3/framework/plugins/linux/envars.py
+++ b/volatility3/framework/plugins/linux/envars.py
@@ -3,8 +3,9 @@
 #
 
 import logging
+from typing import Iterable, Tuple
 
-from volatility3.framework import exceptions, renderers
+from volatility3.framework import renderers, interfaces
 from volatility3.framework.configuration import requirements
 from volatility3.framework.interfaces import plugins
 from volatility3.framework.objects import utility
@@ -17,7 +18,7 @@ class Envars(plugins.PluginInterface):
     """Lists processes with their environment variables"""
 
     _required_framework_version = (2, 13, 0)
-    _version = (1, 1, 0)
+    _version = (2, 0, 0)
 
     @classmethod
     def get_requirements(cls):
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/linux/envars.py
```diff
@@ -39,76 +40,98 @@ def get_requirements(cls):
...
-            # kernel threads never have an mm as they do not have userland mappings
-            try:
-                mm = task.mm
-            except exceptions.InvalidAddressException:
-                # no mm so cannot get envars
-                vollog.debug(
-                    f"Unable to access mm for task {pid} {name} it is likely a kernel thread, will not extract any envars."
-                )
-                mm = None
+            if task.is_kernel_thread:
```
ikelos, line 113: Do we want to log the kernel thread like we did in the previous version?

volatility3/framework/plugins/linux/envars.py
```diff
@@ -39,76 +40,98 @@ def get_requirements(cls):
...
+        # Read the full task environment variable buffer.
+        envar_data = proc_layer.read(env_start, env_area_size)
+
+        # Parse envar data, envars are null terminated, keys and values are separated by '='
+        envar_data = envar_data.rstrip(b"\x00")
+        for envar_pair in envar_data.split(b"\x00"):
+            try:
+                env_key, env_value = envar_pair.decode().split("=", 1)
+            except ValueError:
+                # Some legitimate programs, like 'avahi-daemon', avoid reallocating the args
```
ikelos, line 99: Good comment, thanks!

ikelos (approved):
Looks good, just wondering about silently dropping kernel tasks, but otherwise fine with it.

volatility3/framework/plugins/linux/envars.py
gcmoreira: I don't think so. kernel threads don’t have their own memory context (mm/mm_struct), so logging that is redundant

PR #1428, opened 2024-12-18
Linux: task.get_parent_pid() fix to mimic getppid() syscall

gcmoreira:
In the context of #1412, I missed another bug when retrieving the task parent in the initial fix. The correct value should be `task.real_parent.tgid` rather than the initial `task.parent.pid`. See https://elixir.bootlin.com/linux/v6.12.5/source/kernel/sys.c#L989


ikelos (approved):
Ok, as long as people in Linux land are getting what they expect...

PR #1431, opened 2024-12-18
Use floor division

j-t-1:


ikelos (approved):
I kinda like the explicitness, but it's probably slightly less efficient...  5:)

PR #1432, opened 2024-12-18
Minor typing improvements for `mypy`

TheMythologist:


pyproject.toml
```diff
@@ -68,7 +69,7 @@ include = ["volatility3*"]
 mypy_path = "./stubs"
 show_traceback = true
 
-[tool.mypy.overrides]
```
ikelos, line 71: What are the double brackets here for?

volatility3/framework/__init__.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, value: Any, cls: Type) -> None:
         self.default_value = value
         self.cls = cls
 
-    def __get__(self, obj: Any, get_type: Type = None) -> Any:
+    def __get__(self, obj: Any, get_type: Type = Optional[None]) -> Any:
```
ikelos: This is the wrong place for the `Optional`?  It should be part of the type not the return value, shouldn't it?

ikelos (commented):
These all look reasonable (and time consuming, sorry we didn't do them right first time over).  I think the `abstractmethod` decorators may not be all that valuable unless the classes inherit from `ABC.Meta`, but I don't think they particularly hurt.  There's one place where I think the `Optional` was applied to the value rather than the type, so that'll need fixing, but otherwise look good, thanks.  5:)

volatility3/framework/__init__.py
TheMythologist: You're right, my bad 😅

pyproject.toml
TheMythologist: `ignore_missing_imports` for mypy is actually supposed to be only for specific modules (and it needs to be wrapped in double brackets). There's a warning if it isn't an array:
```
pyproject.toml: tool.mypy.overrides sections must be an array. Please make sure you are using double brackets like so: [[tool.mypy.overrides]]
```

TheMythologist: But seeing as how there's no specific module specified in `[[tool.mypy.overrides]]`, should I just remove it entirely since it's not serving any purpose right now?

TheMythologist:
> I think the `abstractmethod` decorators may not be all that valuable unless the classes inherit from `ABC.Meta`, but I don't think they particularly hurt.

Should I ensure all the classes that I added `abstractmethod` decorators for all inherit from `ABC.Meta`?

ikelos:
No, I don't think that's necessary, I was more just pointing out that this is more of an academic change than a practical one.

pyproject.toml
ikelos: Errr, I think I'd prefer it.  I might come along and think the double brackets are a mistake otherwise...  Either remove it or find some way to comment it with the documentation that says why double bracketing makes sense (I'm still not sure I understand why the section itself needs double bracketing, I think it's being interpreted strangely in some way, I'd just as soon deal with it later if it's not adding anything/necessary).

ikelos (approved):
Looks good, thanks!

PR #1440, opened 2024-12-18
Windows Extensions: Fix potential AttributeErrors

### LINKED ISSUE #1437: `get_owner` calls in network extension can lead to backtraces
These two functions call `get_owner` then in the same line `is_valid()` which can backtrace as `get_owner` returns None in a failure case:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/symbols/windows/extensions/network.py#L90-L97

Going to assign this to David

dgmcdona:
If `self.get_owner()` returns `None`, and the chained call to
`is_valid()` is executed, an `AttributeError` will occur. This fixes two
instances of this bug by intializing a local variable with the result of
the `get_owner()` call, checking for `None`, and returning if that's the
case.

Also adds type-hints for these methods.

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/network.py
+++ b/volatility3/framework/symbols/windows/extensions/network.py
@@ -4,7 +4,7 @@
 
 import logging
 import socket
-from typing import Dict, Tuple, List, Union
+from typing import Dict, Tuple, List, Union, Optional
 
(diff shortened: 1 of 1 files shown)


ikelos (approved):
Looks good, thanks for the thorough fix!

PR #1448, opened 2024-12-18
Windows PEDump: Revert overwritten changes

dgmcdona:
When #1364 was merged, it may not have been rebased onto the changes
introduced in #1422, and they ended up overwritten to the old version.
This reverts those changes.


ikelos:
Ah, sorry about that.  The ruff changes were significant and the review took quite a while so some changes went in before it got looked at.  Thanks for fixing it up (again).  5:)

ikelos (approved):
:+1:

PR #1453, opened 2024-12-19
Swap two letters in a typo

j-t-1:


ikelos (approved):
Heheheh, I'm likely to do a lot of that, I type quite quickly and now from my condition my left hand fingers aren't quite as certain where they are as they used to be.  Thanks for spotting all these!

j-t-1:
We could try using something like codespell on the codebase.

ikelos:
I'm up for it (although if there's not too may of them, it might be nice to do them all in one big PR, if it's just typos)?

PR #1454, opened 2024-12-19
Reformat how to write a simple plugin

j-t-1:


j-t-1:
Have not changed this beyond formatting (still a rookie!). I was going to update it to the current version of `DllList`, but the older version is simpler so kept with it.

ikelos (approved):
Sure, looks more like black/ruff's formatting and hopefully slightly easier to read.  5:)

PR #1455, opened 2024-12-19
Reorder requirements by type

j-t-1:


volatility3/framework/plugins/windows/dlllist.py
```diff
@@ -114,7 +117,7 @@ def _generator(self, procs):
                         mod_re = re.compile(self.config["name"], flags)
                     except re.error:
                         vollog.debug(
-                            "Error parsing regular expression: %s", self.config["name"]
+                            f'Error parsing regular expression: {self.config["name"]}'
```
ikelos: Thanks, but as part of the big ruff linting we did recently all vollog messages should have explicitly been reverted *back* to format string style.  Please see https://docs.astral.sh/ruff/rules/logging-f-string/ for the more information.  If you could strip this change out, everything else looks good.  5:)

ikelos (commented):
Just one change that we don't want including (logging messages using f-strings), but otherwise looks good, thanks!  5:)

PR #1457, opened 2024-12-20
testcases: Add basic volshell testcases for each OS image

### LINKED ISSUE #1456: Volshell doesn't have a testcase
**Describe the bug**
Volshell can be missed when plugin versions are bumped

We should add a simple test case that runs a script with `--script` off one of the existing samples, and verifies that various standard function works.  The test should verify that no exceptions (exit condition of the run is 0, all successful).

gcmoreira:
In the context of #1456, this PR adds a basic volshell testcases for each OS image.

@ikelos please verify this fails before the changes mentioned in #1456 

It runs `ps()` on each of them and verifies it gets enough processes for both Linux and Windows.
If you are interested in any other functions let me know and can add them as well. So far, it verifies:
- requirements
- ps() works.. and 
- basically, ensures that volshell runs and exits without issues

Which is already quite a bit from what we got.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -42,6 +42,11 @@ jobs:
 
     - name: Testing...
       run: |
+        # VolShell
+        pytest ./test/test_volatility.py --volatility=volshell.py --image-dir=./test_images -k test_windows_volshell -v
+        pytest ./test/test_volatility.py --volatility=volshell.py --image-dir=./test_images -k test_linux_volshell -v
+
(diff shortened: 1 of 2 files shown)


ikelos:
Yep, having clocked back to just before the volshell fix (and applying this PR), the test cases fail, so looks good, thanks very much for getting it put together so quickly!  5:D

https://github.com/volatilityfoundation/volatility3/actions/runs/12434282711/job/34717598554

PR #1459, opened 2024-12-20
Windows malfind: fix unbound page variable access

### LINKED ISSUE #1444: malfind has incorrect variable references
1) `proc_id` might not be set here if it is what causes the exception:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/malfind.py#L109

2) The page variable is assigned way away from this error message and it might not even hold the value from the current loop iteration:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/malfind.py#L158C105-L158C109

Abyss-W4tcher:
Hi,

This small fix cleans out a functional but non-clean variable access, from outside a for loop. Logic was simplified as well (unnecessary "check" variable removed).

Fixes #1444


ikelos (approved):
Looks good, thanks!  5:)

PR #1463, opened 2024-12-20
Add a required framework version

j-t-1:


j-t-1:
There is not a `_required_framework_version` attribute in SvcList, is this okay?

j-t-1:
I put a superfluous _the_ at the end of the amended line...

ikelos:
> There is not a `_required_framework_version` attribute in SvcList, is this okay?

It should have one, otherwise it says it works will all versions of the framework way back to pre-release versions.  That's quite possibly true, but better if it had one.  Not sure what the minimum would be, but I guess check back to the version when the plugin was added?

ikelos:
This looks good, but I'll leave it open a bit for you to make the `_required_framework_version` if you wanted.  If not, it's been that way a long time, it probably will never cause a problem (and get added as soon as it would)...

ikelos (approved):
Looks fine, thanks.  5:)

PR #1464, opened 2024-12-21
Add the suspended threads plugin from DEF CON 2024

atcuno:
This is one of the last remaining plugins from DEF CON to be merged.

Its output against a sample with process hollowing present:

```
$ python3 vol.py -r pretty -f /mnt/samples/Injection* windows.suspended
Volatility 3 Framework 2.13.0
  |        Process | PID |  TID |                   StartFile |        StartSymbol |   StartAddress |                                                            Win32StartFile | Win32StartSymbol | Win32StartAddress
* | HelloWorldCpp. | 240 | 4192 | \Windows\System32\ntdll.dll | RtlUserThreadStart | 0x7ffd8195cea0 | \Users\Administrator\Desktop\Injection\ProcessHollowing\HelloWorldCpp.exe |                - |    0x7ff67a2b12b4
```

### DIFF
--- a/volatility3/framework/plugins/windows/suspended_threads.py
+++ b/volatility3/framework/plugins/windows/suspended_threads.py
@@ -86,7 +86,9 @@ class SuspendedThreads(interfaces.plugins.PluginInterface):
                     continue
 
                 # Nothing useful to report if a process doesn't have VADs.. Also a sign of smear/terminated
-                vads = pe_symbols.PESymbols.get_vads_for_process_cache(vads_cache, owner_proc)
+                vads = pe_symbols.PESymbols.get_vads_for_process_cache(
+                    vads_cache, owner_proc
+                )
                 if not vads:
                     continue
 
@@ -144,4 +146,3 @@ class SuspendedThreads(interfaces.plugins.PluginInterface):
             ],
             self._generator(),
         )
-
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/suspended_threads.py
```diff
@@ -0,0 +1,148 @@
...
+                    continue
+
+                # Only compute this if needed as its expensive and 99.9% of samples
+                # will not have suspended threads
+                if not proc_modules:
+                    proc_modules = pe_symbols.PESymbols.get_process_modules(
+                        self.context, kernel.layer_name, kernel.symbol_table_name, None
+                    )
+
+                    path_and_symbol = partial(
```
ikelos: Would prefer this to be `functools.partial` and just `import partial` above.  The only exception we've made so far is for types, which are used *everywhere*.  I'd prefer not to let this open a gap for other functions to claim asylum please...

ikelos (commented):
Looks good, just wondering if the functionality would be useful broken out into a classmethod so other plugins could use it?  Feels like the functionality could be handy in other plugins?

volatility3/framework/plugins/windows/suspended_threads.py
atcuno: Fixed the import method.

I want to leave the plugin as is for the generator as the real fix will be coming soon where `gather_thread_info` in thrdscan returns the information about the paths and symbols so that all the thread related plugins get it. Right now our threads output is pretty bare bone and not very usable in investigations. Once the update goes in then threads/thrdscan/<all threads plugins> will produce actionable output.

PR #1465, opened 2024-12-23
Change variable names

j-t-1:
Makes it slightly more obvious that they are generics.

### DIFF
--- a/volatility3/__init__.py
+++ b/volatility3/__init__.py
@@ -8,7 +8,7 @@ from importlib import abc
 from typing import List, TypeVar, Callable, Any, Optional
 
 _T = TypeVar("_T")
-_S = TypeVar("_S")
+_U = TypeVar("_U")
 
 
 class classproperty(property):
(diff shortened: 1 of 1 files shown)


ikelos:
Sorry, I'm not sure of the value of this, it seems as though it's just changing the letters, changing T (common for Type) and S (for Source) and making them U and T (swapping the purpose of T in the process).  I don't understand how this makes it any clearer that these are generics?

j-t-1:
Understand. I did not realise S was for Source. The idea (before knowing this!) was when reading the code quickly you see the T first and then the S and it is not obvious why they are in that order or whether the S means something. I think they could also stand for Source and Target, which adds weight to keeping them.

ikelos:
Happy for you to swap the order of their definitions around, but otherwise I'd prefer to keep the code as is...  Marking this as closed, but feel free to reopen if you want to tinker on it at all...

j-t-1:
Yes, I think the only tiny improvement would be swapping the definition order, S before T. But is too trivial for a commit, and cannot find anything else to change in the file.

PR #1466, opened 2024-12-24
Tweak configuration.py

j-t-1:
Create a tuple directly and replace random.choice by random.choices.


j-t-1:
Should line 784 have `float`?

ikelos (approved):
Looks good, thanks!

j-t-1:
Should line 784 include a `float` as a simple type? Differs from line 49.

ikelos:
`float`s weren't supposed to be part of the `BasicTypes` and I don't think they're used anywhere?  That should probably use the variable though...

PR #1467, opened 2024-12-26
Use BasicTypes variable

j-t-1:
This removes a "float", which should be excluded.


ikelos (approved):
Thanks for fixing that up...  5:)

PR #1485, opened 2024-12-28
Make ETHREAD year check dynamic

j-t-1:
Change upper bound year check of ETHREAD to be a decade from now. Makes consistent with EPROCESS.


ikelos (approved):
Seems reasonable, I guess we're within 6 years of 2030 already!  5:S

PR #1488, opened 2024-12-28
Do not yield DRIVER_OBJECT instances found in scanning that are not a…

### LINKED ISSUE #1481: `MajorFunction` access crashes driverirp
I need to look into this more to see if its the enumerate() crashing somewhere in the array or `driver` instance being bad, but it crashed out on several samples:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/driverirp.py#L80

Sample: Win10x64_22H2_rawsocket2.lime

atcuno:
…ctual instances #1481

### DIFF
--- a/volatility3/framework/plugins/windows/driverscan.py
+++ b/volatility3/framework/plugins/windows/driverscan.py
@@ -53,7 +53,9 @@ class DriverScan(interfaces.plugins.PluginInterface):
 
         layer = context.layers[layer_name]
         module = context.module(symbol_table, layer_name, 0)
-        driver_start_offset = module.get_type("_DRIVER_OBJECT").relative_child_offset("DriverStart")
+        driver_start_offset = module.get_type("_DRIVER_OBJECT").relative_child_offset(
+            "DriverStart"
+        )
 
         for result in poolscanner.PoolScanner.generate_pool_scan(
             context, layer_name, symbol_table, constraints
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/driverscan.py
```diff
@@ -55,6 +55,18 @@ def scan_drivers(
...
             _constraint, mem_object, _header = result
+
+            # *Many* _DRIVER_OBJECT instances were found at the end of a page
+            # leading to member access causing backtraces across several plugins
+            # when members were accessed as the next page was paged out.
+            # `DriverStart` is the first member from the beginning of the structure
+            #  of interest to plugins, so if it is not accessible then this instance
+            # is not useful or usable during analysis
+            try:
+                mem_object.DriverStart
```
ikelos: Is just listing a variable enough to trip the output?  I'd think that could get optimized out?  Also, don't we care if the whole structure is valid, or would having just `DriverStart` be enough?

ikelos (commented):
Never seen a variable just *accessed* without an assignment or anything, but if it works?  Just want to check that getting back a structure with only one of those is still worthwhile...

atcuno:
@ikelos this is ready but not sure if a version bump is needed due to the error paths in `get_names_for_driver` changing to meet how we code things now. The function didn't type the return value before so that part didn't change. The only place in the current code base that externally calls it is drivermodule, which is going to change for other fixes right after this is merged anyway, so its an easy fix if needed.

ikelos:
Looks like some tests need updating here?  I dunno whether we lost legit values or something else went wrong, but it needs someone to look into it.  Technically it should get a bump given the return type is changed.  If it's not too hard to put it, it would probably be worthwhile, just in case it does affect someone...

ikelos:
Why did this get closed?  The associated ticket was also closed, did the problem just go away or did it get solved a different way?

atcuno:
This issue plus several others that popped up in testing are covered here: https://github.com/volatilityfoundation/volatility3/pull/1621

PR #1490, opened 2024-12-28
Prevent infinite looping and out of memory errors #1482

### LINKED ISSUE #1482: `get_full_key_name` infinite loop causes out of memory errors on analyst systems
I haven't been able to complete a full testing run as this function caused my testing VM (8 cores, 16GB RAM) to OOM and crash:

https://github.com/volatilityfoundation/volatility3/blob/b1a42d9510046377a426d62f7cbde1ce1a4ad528/volatility3/framework/symbols/windows/extensions/registry.py#L134

I tested one of the problematic samples (Win10Browsers_Mem.zip.lime) and it was going in an infinite loop due to smear and the string building was causing the VM to crash from such rapid memory usage.

The fix needs to include the maximum count + the usual `seen` usage.

atcuno:


atcuno:
This is ready @ikelos

volatility3/framework/symbols/windows/extensions/registry.py
```diff
@@ -133,8 +133,17 @@ def _skip_key_hive_entry_path(self, kcb_flags):
 
     def get_full_key_name(self) -> str:
         output = []
+        seen = set()
+
         kcb = self.KeyControlBlock
         while kcb.ParentKcb:
+            if kcb.ParentKcb.vol.offset in seen:
+                return ""
```
ikelos: Shouldn't these be `None` or something else?  I don't want the `BaseAbsentValues` sneaking in here, but it feels like we should be alerting people that things didn't work?  The downside with that is then you have to do error checking whenever you try to pull the full key name?  Just wondering of the consequences of it returning a value but blank string in case of an error?

ikelos (commented):
Happy with the error checking, just wondering how best to indicate the error to the user?  Possibly we should have a vollog warning that prints as much of the key name as could be recovered?  I'm not sure...

Also, if you want github to auto tag the issues these are supposed to fix, you can't just mention it in the title, you need to put "Fixes #blah" or "Closes #blah" in the bug body, please...

volatility3/framework/symbols/windows/extensions/registry.py
atcuno: That function is called from `handles` on this line:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/handles.py#L313

Which, in the `else` case of that same block, the empty string is used to indicate an error:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/handles.py#L318

If it makes more sense, I can change both places (the registry extension + line 318 in handles) to send back None instead of "", then have the yield() handler do the "variable or renderers....." setup. Is that preferred? That avoids the extension API having to send back a renderer instance.

atcuno: Also, thanks for this:

"Also, if you want github to auto tag the issues these are supposed to fix, you can't just mention it in the title, you need to put "Fixes #blah" or "Closes #blah" in the bug body, please..."

I am used to GitLab where tagging the number accomplishes it.

ikelos: Yeah, I think it would be better to return None as a way of indicating an error, and then let the caller figure out what to do with it.  Also, no problem.  5;)

atcuno:
@ikelos this is ready again

ikelos (approved):
Ok, looks good, thanks.

PR #1492, opened 2024-12-29
Significantly improve the smear/error handling in the netstat plugin

atcuno:
The netstat plugin initially backtraced on 100+ samples in the testing set from bugs in 10+ code locations.

After seeing this, I went through the plugin and revamped it all to bring the error handling up to our current way of doing it.

The plugin now makes it through every sample in the regression set without any backtraces or memory issues.

### DIFF
--- a/volatility3/framework/plugins/windows/netstat.py
+++ b/volatility3/framework/plugins/windows/netstat.py
@@ -340,7 +340,7 @@ class NetStat(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                 offset=tcpip_module_offset + part_table_symbol,
             )
         except exceptions.InvalidAddressException:
-            vollog.debug(f"`PartitionTable` not present in memory.")
+            vollog.debug("`PartitionTable` not present in memory.")
             return
 
         # part_table is the actual partition table offset and consists out of a dynamic amount of _PARTITION objects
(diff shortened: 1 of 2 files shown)


atcuno:
This is ready @ikelos

volatility3/framework/plugins/windows/netstat.py
```diff
@@ -188,16 +206,30 @@ def enumerate_structures_by_port(
...
+                        layer_name=layer_name,
+                        offset=next_obj_address - ptr_offset,
+                    )
+                except exceptions.InvalidAddressException:
+                    return
+
                 yield curr_obj
 
+                try:
+                    next_obj_address = cls._decode_pointer(curr_obj.Next)
```
ikelos, line 229: Subtle difference between these.  In the old one we checked whether the `curr_obj.Next` was false, in this one we check if `cls._decode_pointer(curr_obj.Next)` is false.  Are these both likely to trip on the same values?

ikelos (approved):
The decoding just masks off the lowest bits, so actually this might hit more cases than before...

volatility3/framework/plugins/windows/netstat.py
atcuno: It would work the same. If `Next` points to a value of 0 then the `decode_ptr` would do:

```
value = value & 0xFFFFFFFFFFFFFFFC

return value
```

With 0 as the starting input for `value`.

atcuno: versus if `Next` is an invalid address then the exception handler would trigger

PR #1494, opened 2024-12-29
Use enumerate for readability

j-t-1:


ikelos (approved):
Yep, that's nice.  Doesn't tend to occur to me when I'm writing code but I'll try to remember it in future, thanks!  5:)

PR #1495, opened 2024-12-30
Use rsplit instead of split

j-t-1:
Since we want to split rightmost only.


ikelos (approved):
Another one I didn't really know about, this is much clearer code, thanks!  5:D

PR #1497, opened 2024-12-30
Fix uncheck read() call and remove variable that would not be definie…

### LINKED ISSUE #1477: Crashes in skeleton key check from missing try/except on read
Another bug found in round 1 of mass testing:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/skeleton_key_check.py#L436

atcuno:
…d if exception triggers

Fixes #1477


atcuno:
This is ready @ikelos

ikelos (approved):
Looks reasonable.

PR #1498, opened 2024-12-30
Sort imports and swap two assignments

j-t-1:


ikelos (approved):
Not sure of the value of reordering the assignments, but no objections?

j-t-1:
Just to put them in the same order as the function signature.

Issue #1493, opened 2024-12-29
Multiple Bugs in handles enumeration


atcuno:
@dgmcdona I saw you did this work recently:

https://github.com/volatilityfoundation/volatility3/pull/1389/

But I am still getting the backtrace at the end of this writeup, which was discussed in another ticket as being related to getting the SAR value:

https://github.com/volatilityfoundation/volatility3/issues/1146

The sample is: Win10x64_22H2_showmypc.lime

```
DEBUG    volatility3.cli: Traceback (most recent call last):
  File "/home/ub/volatility3/volatility3/framework/plugins/windows/handles.py", line 68, in _get_item
    if not self.context.layers[virtual].is_valid(handle_table_entry.Object):
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ub/volatility3/volatility3/framework/objects/__init__.py", line 969, in __getattr__
    raise AttributeError(
AttributeError: UnionType has no attribute: symbol_table_name1!_HANDLE_TABLE_ENTRY.Object

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ub/volatility3/volatility3/cli/__init__.py", line 501, in run
    renderer.render(grid)
  File "/home/ub/volatility3/volatility3/cli/text_renderer.py", line 232, in render
    grid.populate(visitor, outfd)
  File "/home/ub/volatility3/volatility3/framework/renderers/__init__.py", line 241, in populate
    for level, item in self._generator:
  File "/home/ub/volatility3/volatility3/framework/plugins/windows/handles.py", line 295, in _generator
    for entry in self.handles(object_table):
  File "/home/ub/volatility3/volatility3/framework/plugins/windows/handles.py", line 266, in handles
    yield from self._make_handle_array(TableCode, table_levels)
  File "/home/ub/volatility3/volatility3/framework/plugins/windows/handles.py", line 241, in _make_handle_array
    item = self._get_item(entry, handle_value)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ub/volatility3/volatility3/framework/plugins/windows/handles.py", line 80, in _get_item
    if handle_table_entry.ObjectPointerBits == 0:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ub/volatility3/volatility3/framework/objects/__init__.py", line 961, in __getattr__
[...]

PR #1499, opened 2024-12-30
Windows Handles: Fix remaining tracebacks

dgmcdona:
refs #1493 

This fixes the remaining bugs in handle enumeration that can occur when accessing struct members. Apparently, just checking the base address of a `_HANDLE_TABLE_ENTRY` is not sufficient to prevent `InvalidAddressException`s, so `try/except` blocks have been added around the problematic member accesses and pointer dereferences as well.

This also includes a patch as requested by @atcuno  that cleans up some problematic object header object type name parsing.

### DIFF
--- a/volatility3/framework/plugins/windows/handles.py
+++ b/volatility3/framework/plugins/windows/handles.py
@@ -68,7 +68,12 @@ class Handles(interfaces.plugins.PluginInterface):
             if not self.context.layers[virtual].is_valid(handle_table_entry.Object):
                 return None
             fast_ref = handle_table_entry.Object.cast("_EX_FAST_REF")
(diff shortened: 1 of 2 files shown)


ikelos (approved):
This looks good, thanks.  Just musing over the best way to minimize the number of try/catches we need in the code.  I think it's right to return the appropriate error/exception when it happens, so I guess that just leaves setting as large a catch-all as appropriate.  Generally when exceptions occur I prefer them logged (since they're supposed to be exceptional), but here it looks like they're command and expected.  I dunno if there's a way of codifying that though...

ikelos:
Does this close off #1493 or only partially?

dgmcdona:
As far as I know, this should close #1493

PR #1500, opened 2024-12-31
Exception handling in registry-related code

### LINKED ISSUE #1469: cachedump backtrace from missing try/except around pointer access
This triggered in the first run of mass testing across many samples:

```
24-12-24 21:09:20 volatility3.cli DEBUG    Traceback (most recent call last):
    renderer.render(grid)
    grid.populate(visitor, outfd)
  File "/home/ub/volatility3/volatility3/framework/plugins/windows/cachedump.py", line 143, in _generator
    data = sechive.read(cache_item.Data + 4, cache_item.DataLength)
  File "/home/ub/volatility3/volatility3/framework/layers/linear.py", line 63, in read
    self._context.layers.read(layer, mapped_offset, mapped_length, pad)
                                          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ub/volatility3/volatility3/framework/layers/intel.py", line 510, in _translate
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ub/volatility3/volatility3/framework/layers/intel.py", line 457, in _translate_swap
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ub/volatility3/volatility3/framework/layers/intel.py", line 166, in _translate
volatility3.framework.exceptions.PagedInvalidAddressException: Page Fault at entry 0xffffffff00000430 in page entry```

### PR DESCRIPTION
**Pull request by dgmcdona**
This adds exception handling in a number of places where `InvalidAddressException`s were not being caught before.

Notably, `CM_KEY_VALUE.decode_data()` was missing exception handling for calls to `read` that can raise these exceptions. The new behavior is that a debug message will be logged, and null bytes of the same length will be returned in place of the actual data. This will allow partial data to be returned when reading large binary blobs from binary registry values.

It also adds exception handling in the amcache and hashdump plugins where the exceptions raise by `get_key` were not being caught exhaustively.

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/registry.py
+++ b/volatility3/framework/symbols/windows/extensions/registry.py
@@ -159,6 +159,11 @@ class CM_KEY_NODE(objects.StructType):
     """Extension to allow traversal of registry keys."""
 
     def get_volatile(self) -> bool:
+        """
+        Returns a bool indicating whether or not the key is volatile.
+
+        Raises ValueError if the key was not instantiated on a RegistryHive layer
+        """
         if not isinstance(self._context.layers[self.vol.layer_name], RegistryHive):
             raise ValueError(
                 "Cannot determine volatility of registry key without an offset in a RegistryHive layer"
@@ -166,7 +171,10 @@ class CM_KEY_NODE(objects.StructType):
         return bool(self.vol.offset & 0x80000000)
 
     def get_subkeys(self) -> Iterator["CM_KEY_NODE"]:
-        """Returns a list of the key nodes."""
+        """Returns a list of the key nodes.
+
+        Raises TypeError if the key was not instantiated on a RegistryHive layer
+        """
         hive = self._context.layers[self.vol.layer_name]
         if not isinstance(hive, RegistryHive):
             raise TypeError("CM_KEY_NODE was not instantiated on a RegistryHive layer")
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: volatility3/framework/symbols/windows/extensions/registry.py [hunk 1]
```diff
@@ -159,14 +159,22 @@ class CM_KEY_NODE(objects.StructType):
     """Extension to allow traversal of registry keys."""
 
     def get_volatile(self) -> bool:
+        """
+        Returns a bool indicating whether or not the key is volatile.
+
+        Raises ValueError if the key was not instantiated on a RegistryHive layer
```
dgmcdona: Unfortunately this raises `ValueError` when the other methods raise `TypeError` for the same reason, but I left it as is to avoid the major version bump that I suspect would be required if we changed this.

ikelos: Errr, I'm actually willing to go for gray area.  Yes, technically we probably should do a major bump, but equally I'm not sure how many things would break because we changed an exception (the process would already have been throwing an exception).   We might want to check anywhere in our code (at least) where we explicitly catch a ValueError to check that it isn't important.  If so, I'd be willing to let it slide with a MINOR bump rather than a full MAJOR one.  (Not relaxing the rule in general, just making this specific single exception)...

dgmcdona: Ok cool, I did the conversion + version bump, with an explanation for the version bump in the commit message.

dgmcdona:
closes #1469

ikelos:
Maybe it only does it if it's in the first comment?  Make it's case sensitive?  5:S  It does work usually, honest!

ikelos:
Ok, thanks, let's get this in...  5:)

PR #1503, opened 2025-01-01
Linux: Fix kmsg unguarded read of msg.len

ikelos:
Spotted an issue in kmsg on the `surge-output-debian9.0-4.9.0-3-amd64-x86_64.lime` image, whereby kmsg failed because msg.len couldn't be read.  Hopefully this is a reasonble solution?

### DIFF
--- a/volatility3/framework/plugins/linux/kmsg.py
+++ b/volatility3/framework/plugins/linux/kmsg.py
@@ -317,23 +317,27 @@ class Kmsg_3_5_to_3_11(ABCKmsg):
         while cur_idx < end_idx:
             msg_offset = log_buf_ptr + cur_idx  # type: ignore
             msg = self.vmlinux.object(object_type=log_struct_name, offset=msg_offset)
-            if msg.len == 0:
-                # As per kernel/printk.c:
-                # A length == 0 for the next message indicates a wrap-around to
-                # the beginning of the buffer.
-                cur_idx = 0
-                end_idx = log_next_idx
-            else:
-                facility, level, timestamp, caller = self.get_prefix(msg)
-                level_txt = self.get_level_text(level)
-                facility_txt = self.get_facility_text(facility)
+            try:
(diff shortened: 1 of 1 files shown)


gcmoreira:
Hey @ikelos, good catch! That will prevent the issue you found, however, I'm not a big fan of the recent extensive try/except `InvalidAddressException` additions here and in the other PRs. When possible, I think we should identify the root cause and i.e. validate the pointer with a straightforward `if` statement:
```python
    if not (pointer and pointer.is_readable()):
        # error
```
Or if it's an offset/address, we should check it using the respective layer:
```python
    if not (layer.is_valid(offset, size)):
        # error
```
Adding try/except blocks everywhere will slow down the framework and increase memory consumption, especially when included in large loops and/or recursive calls.

atcuno:
@gcmoreira we can take into account `is_readable` just for pointers, but most of the `try/except` blocks that David and I have added lately would not be covered just from `is_readable`.

The general flow of the backtraces from round 1 of mass testing were from two cases:

1) Accessing a pointer only (`is_readable` could cover this)

2) Accessing a pointer (object instance) where the first page of the instance is valid, but then accessing members on later page(s) triggers a backtrace. `is_readable` would not fix this. 

Also, most of the code paths are complex and mix 1) and 2) so you would end up with a bunch of `is_readable` inside of the `try/except` blocks, which seems redundant. 

Vol3 was designed specifically to let plugins deal with partial structure instances (1st page valid, pages 2->N maybe valid) so the careful exception handling is the designed way to handle this. We didn't document this well enough though so we have (had) many code paths that did not follow the convention. Given how much stuff broke in mass testing though, I feel like we have uncovered many of them that are being patched/patched in the last week. Code review should also catch them better going forward.

ikelos:
If there's no objections, I'll merge this in then.  Happy to continue the conversation here or elsewhere if needed...

PR #1506, opened 2025-01-02
Linux: ground work for multi-arch vmemmap_start calculation 

Abyss-W4tcher:
Hi, 

calculation for the `struct page` array `vmemmap_start` address  differs from one architecture to another (Intel, AArch64). This PR splits the current computation into internal APIs in the `page` class, allowing to easily insert new architectures in the future.

- `get_flags_list` was moved to the bottom to group related `paddr` APIs
- `to_paddr` input and output was not modified 
- as python3.7 is now unsupported, I fixed the `@functools.cached_property` bits

Only private APIs were added, which shouldn't require a minor bump. Please tell me otherwise.

### DIFF
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -15,7 +15,7 @@ from typing import Generator, Iterable, Iterator, Optional, Tuple, List, Union, 
 from volatility3.framework import constants, exceptions, objects, interfaces, symbols
 from volatility3.framework.renderers import conversion
 from volatility3.framework.constants import linux as linux_constants
-from volatility3.framework.layers import linear
+from volatility3.framework.layers import linear, intel
 from volatility3.framework.objects import utility
 from volatility3.framework.symbols import generic, linux, intermed
 from volatility3.framework.symbols.linux.extensions import elf
@@ -2525,16 +2525,13 @@ class address_space(objects.StructType):
 
 
 class page(objects.StructType):
-    @property
-    @functools.lru_cache
+    @functools.cached_property
     def pageflags_enum(self) -> Dict:
         """Returns 'pageflags' enumeration key/values
 
         Returns:
             A dictionary with the pageflags enumeration key/values
         """
-        # FIXME: It would be even better to use @functools.cached_property instead,
-        # however, this requires Python +3.8
         try:
             pageflags_enum = self._context.symbol_space.get_enumeration(
                 self.get_symbol_table_name() + constants.BANG + "pageflags"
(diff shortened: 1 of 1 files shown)


volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -2620,14 +2631,30 @@ def get_content(self) -> Union[str, None]:
         """
         vmlinux = linux.LinuxUtilities.get_module_from_volobj_type(self._context, self)
         vmlinux_layer = vmlinux.context.layers[vmlinux.layer_name]
-        physical_layer = vmlinux.context.layers["memory_layer"]
+        physical_layer_name = self._context.layers[self.vol.layer_name].config.get(
+            "memory_layer", self.vol.layer_name
```
ikelos, line 2635: Need to be really careful about this.  I feel at some point soon we're going to have to go back and find all these instances that get the physical layer that way (this ignores swap space and hardwires the expectation of a `memory_layer` into the code, which is only on intel and could theoretically change in the future).  Technically it should go through the `.dependencies` of the layer and choose one of them (somehow).  I won't push that at the moment, and I think we could find all instances of this by looking for the string `memory_layer`, just putting in a heads-up here...

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -2620,14 +2631,30 @@ def get_content(self) -> Union[str, None]:
...
 
+    def get_flags_list(self) -> List[str]:
+        """Returns a list of page flags
+
+        Returns:
+            List of page flags
+        """
+        flags = []
+        for name, value in self.pageflags_enum.items():
+            if self.flags & (1 << value) != 0:
```
ikelos, line 2653: Is this a feature we should add to enums (a way of listing all flags or something)?  Just want to avoid having the same little code snippet all over the place.

ikelos (approved):
Happy with this, one note to look out for in the future, and one that seems like it could make enums better in the future, but no show stoppers...

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: Automating it might be a good idea, but restraining to an automated method could complicate things when custom processing is needed (`if self.flags & (1 << value) != 0 and some_additional_checks`).

ikelos: Yeah, fair enough.  I know python has `IntFlag` so it can apply some constraints, but I guess for now we'll leave it unless we find that code happening over and over again...

PR #1509, opened 2025-01-03
linux: module symbols: Fixes and code improvements

gcmoreira:
- Fixed an issue causing the generation of an invalid, extra symbol.
- Reuse ELF sym API instead of reimplemented it
- Updated the function to return the symbol index, enabling the use of additional module tables.
- Ensured the ELF symbol object has `cached_strtab` set, allowing retrieval of critical symbol information like names.
- Added typing hints


gcmoreira:
Hey @ikelos, this is part of a larger pull request that I'm breaking into smaller, more manageable changes. While these updates might seem disconnected at the moment, they will align once the full implementation is in place.

The `module::get_symbols()` method interface has been updated. Since it was only used internally, I've adjusted all relevant references to match the new implementation.

Given the scope of these changes, at least a framework minor version update might be required. Let me know your thoughts!

volatility3/framework/symbols/linux/extensions/elf.py
```diff
@@ -328,22 +329,29 @@ def cached_strtab(self):
...
-            if idx != -1:
-                name_bytes = name_bytes[:idx]
-            return name_bytes.decode("utf-8", errors="ignore")
-        else:
-            # If we cannot read the name from the address space,
-            # we return None.
+        addr = self._cached_strtab + self.st_name
+
+        layer = self._context.layers[self.vol.layer_name]
+        name_bytes = layer.read(addr, max_size, pad=True)
```
Abyss-W4tcher: Do you think a try/except should be wrapped around the layer read to prevent from memory smear ?

gcmoreira: With pad=True it shouldn't raise any exceptions

Abyss-W4tcher: Ok, I had concerns about this early part, if `self.st_name` is an absurd value due to smear ?

https://github.com/volatilityfoundation/volatility3/blob/cedc0b074aeda804809b77c8f00c6dbf8565682c/volatility3/framework/layers/physical.py#L155-L163

gcmoreira: When pad=True, this `for` because `ignore_errors=pad`, will return immediately:
https://github.com/volatilityfoundation/volatility3/blob/cedc0b074aeda804809b77c8f00c6dbf8565682c/volatility3/framework/layers/linear.py#L45-L47

Then this line will pad everything with zeros:
https://github.com/volatilityfoundation/volatility3/blob/cedc0b074aeda804809b77c8f00c6dbf8565682c/volatility3/framework/layers/linear.py#L67

Abyss-W4tcher: Thanks for clarifying it !

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -213,65 +213,63 @@ def get_elf_table_name(self):
...
 
-    def get_symbols(self):
+    def get_symbols(
+        self,
+    ) -> Iterable[Tuple[int, interfaces.objects.ObjectInterface]]:
         """Get symbols of the module
 
         Yields:
-                A symbol object
+            A tuple containing the ELF symbol index and the corresponding ELF symbol object
```
ikelos: This makes assumptions about the ordering of the symbols returned (and it's also technically a change in the API since it'll break anything the currently doesn't expect a tuple) so if you want to make this change it'll need a MAJOR version bump somewhere, but I can't see the use in it?  What's the reason for needing to return the index?

The only existing call to this that I saw get updated throws the value of the index away?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -213,65 +213,63 @@ def get_elf_table_name(self):
...
+        elf_syms = self._context.object(
+            symbol_table_name + constants.BANG + "array",
             layer_name=self.vol.layer_name,
             offset=self.section_symtab,
-            subtype=self._context.symbol_space.get_type(
-                self._elf_table_name + constants.BANG + prefix + "Sym"
-            ),
-            count=self.num_symtab + 1,
+            subtype=sym_type,
+            count=self.num_symtab,
```
ikelos, line 234: Did the previous version have a bug (and set the count too high) or was missing the `+ 1` a typo?  Happy with either outcome, just checking which...

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -213,65 +213,63 @@ def get_elf_table_name(self):
...
         )
-        if self.section_strtab:
-            yield from syms
+        for elf_sym_num, elf_sym_obj in enumerate(elf_syms):
+            # Prepare the symbol object for methods like get_name()
+            elf_sym_obj.cached_strtab = self.section_strtab
```
ikelos, line 238: I'm not a massive fan of the `cached_strtab` attribute, since the implication is that it's set on creation, but it isn't part of the constructor (meaning an `elf_sym` could be constructed but not have it set) but this isn't the place to resolve that.   I see there's a similar design pattern for `elf_phdr` so probably one we want to figure out some time but the method used is the only one I've got off the top of my head...

volatility3/framework/symbols/linux/extensions/elf.py
```diff
@@ -328,22 +329,29 @@ def cached_strtab(self):
     def cached_strtab(self, cached_strtab):
         self._cached_strtab = cached_strtab
 
-    def get_name(self):
-        addr = self._cached_strtab + self.st_name
+    def get_name(self, max_size=KSYM_NAME_LEN) -> Optional[str]:
```
ikelos: Technically this is an additive change to the API.  I'm not sure what the first versioned component above this would be though?  Can we avoid this being a parameter, but make it a constant instead?  I suspect the maximum doesn't change often, and it would keep the API the same but pull the value out to make it more destinctly something people can see and set when needed?

ikelos (commented):
A few questions around some of the API changes, some of which seem unnecessary and would require bumps, but otherwise looks ok...

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: > but I can't see the use in it? 

yes, I tried to explain this  [here](https://github.com/volatilityfoundation/volatility3/pull/1509#issuecomment-2568880107):

> Hey @ikelos, this is part of a larger pull request that I'm breaking into smaller, more manageable changes. While these updates might seem disconnected at the moment, they will align once the full implementation is in place.

Yes, symbols in the symbol table are ordered, which is important because it is not the only table in a Linux kernel module. The module also contains a types table, where the indices correspond to those in the symbol table. This means that the type for the symbol at index (n) in the symbol table is located at the same index in the types table. 

We could handle this in the caller using enumerate, but that would mean `get_symbols()` must always preserve the original order. It cannot skip any symbol, whether it's invalid, null, empty, or whatever. If skipped, the symbols could shift in relation to the other tables, breaking the framework.

I believe this approach ensures the symbols stay correctly aligned with their indices.
Should this be considered a major version bump for the whole framework? Unfortunately, object extensions don't have versioning in place.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Yeap, it was a bug. That's the first bullet:
> Fixed an issue causing the generation of an invalid, extra symbol.

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Yeah, I'm not a fan of that either, but that's how it was implemented :). I try not to be too tempted to rewrite everything I don't like, hehe. Instead, I focus on reusing and keeping things as they are while working towards the main goal of the PR. Although, many times I fail and end up refactoring more than I should... I know.. and I guess you know that by now, haha .. sorry about that. But this time I didn't go down that route. Happy we can explore a better implementation for that in a future PR.

volatility3/framework/symbols/linux/extensions/elf.py
gcmoreira: Yes, we can remove it from the parameters if you prefer. Since the ELF API could be used for other purposes than just kernel symbols, and for some of those use cases, a shorter maximum symbol name length could make more sense there, I assumed that a general maximum-maximum length would be fine for all cases but allowing to customize a shorter one if for some reason it's needed. I also made it a parameter with default value to minimize the impact of this change.

The Linux kernel has it's maximum length, but for general/user-space ELF files the maximum size can be different. The kernel used to have a maximum length of 128 bytes until Rust happened. In kernels 6.1, it was increased to 512 bytes in [this](https://github.com/torvalds/linux/commit/b8a94bfb33952bb17fbc65f8903d242a721c533d) commit.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: I feel there's a better way of marrying the two up?  The symbols in an ISF can have a type parameter, telling you which type they use?  Is that not coming out of the dwarf2json?  Enumerating them here seems really fragile and we'll have to remember it if we ever tinker with any of the way the symbol tables work.  See [here](https://github.com/volatilityfoundation/volatility3/blob/a98a23fa41395b9bbd961f6e50d0a0f19201fcc7/volatility3/framework/interfaces/symbols.py#L23).

Recalculating it each time just opens up the chance something will go wrong later, the mapping is static data, it should be encoded in the ISF itself, surely?

volatility3/framework/symbols/linux/extensions/elf.py
ikelos: Hmmmm, ok.  I think still best to make it a constant, and then we can make a new class that overrides that class local constant down the line (and the class local constant can be set from constants if you want)...

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Hm, it's not related to those symbols. These are kernel module symbols, and this information is not part of the ISF. It is gathered directly from memory rather than from d2j. The symbol types are not included in the elf_sym structure and never will be, as this data is embedded within the module, not within the symbol itself

volatility3/framework/symbols/linux/extensions/elf.py
gcmoreira: So, do you mean something like this (internal or not)?
```python
class elf_sym(objects.StructType):
    _MAX_NAME_LENGTH = KSYM_NAME_LEN
```

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: @ikelos If we want to avoid a version major bump, I can make an intermediate function like:
`get_symbols_ex()` .. which yield index and symbol .. and get_symbols() calls it but ignoring the index. Then, who needs both can still call directly to `get_symbols_ex()`.

volatility3/framework/symbols/linux/extensions/elf.py
ikelos: `constants.KSYM_NAME_LEN`, but yeah, so there's constants (user overridable in one clear place, class constant so that inheritors could override it, and then the actual use point).  Does that make sense or is it just over complicating things?  I don't mind so much, so long as it's pulled out somewhere sensible...  5:)

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Ok, I think I missed where they came from then (there's a lot of special things in the linux extensions that I'm not up to speed on).  If you're happy it makes sense, then that's ok.  Yes, please put it in a different function and make the old one a shell that just returns the same type and has the same signature and name as the original.  Probably wise to include a deprecation warning in that call too, so anyone that is using it knows to shift themselves over...

ikelos: Couldn't someone just call enumerate on the generator before first use though?  If you're changing the function already I guess it makes it less brittle to ensure it's always count from the start, just trying to keep API changes to a minimum.  As I say, if really want it I'm ok with it, just want to make sure it doesn't need changing again down the line...  5:)

volatility3/framework/symbols/linux/extensions/elf.py
gcmoreira: done. Also, I fixed the imports.. have a look

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: sounds good, let's go with that

gcmoreira:
@ikelos it's ready. Thanks for the review

ikelos (approved):
Looks good, thanks!

PR #1512, opened 2025-01-03
Remove comment of find library for windows if needed

j-t-1:
The Windows library has previously been added to the try block.

### DIFF
--- a/volatility3/framework/layers/avml.py
+++ b/volatility3/framework/layers/avml.py
@@ -17,7 +17,6 @@ from volatility3.framework.layers import segmented
 vollog = logging.getLogger(__name__)
 
 try:
-    # TODO: Find library for windows if needed
     try:
         # Linux
         lib_snappy = ctypes.cdll.LoadLibrary("libsnappy.so.1")


ikelos:
I don't know that this works under windows, I haven't tried it?  The FIXME is there to get me to check it?  If you've checked that it works under windows, then I'll take it out happily, but otherwise we need to check it at somepoint (and having a FIXME there is a useful reminder).

j-t-1:
Have not run this under Windows. Was thinking (probably wrongly) that the comment was referring to lines 37 and 43. I think the comment could be changed to make it more self-explanatory.

ikelos:
No, you're right, those are the lines it's referring to, I just... never actually tested them.  5:P  As soon as someone's confirmed that they've checked it all works right on windows, I'm happy to ditch the FIXME...  5:D

j-t-1:
> No, you're right, those are the lines it's referring to, I just... never actually tested them. 5:P As soon as someone's confirmed that they've checked it all works right on windows, I'm happy to ditch the FIXME... 5:D

Installing python-snappy 0.7.3 on Windows and then trying `ctypes.cdll.LoadLibrary("snappy32")` and `ctypes.cdll.LoadLibrary("snappy64")` both give `FileNotFoundError`. On PyPI, the operating systems supported are MacOS and POSIX. This partially explains the TODO.

So for Snappy to work on Windows you need to compile from source? I do not know if this is correct, so gonna close this.

PR #1513, opened 2025-01-03
Add missing exception handling in env var recovery. Prevent backtraces

atcuno:


ikelos (approved):
Seems reasonable if those are the only attributes we need.

gcmoreira:
@ikelos @atcuno this was already fixed as part of https://github.com/volatilityfoundation/volatility3/pull/1542. The `env_start` and `env_end` members are not pointers, so an `InvalidAddressException` can only occur if `mm` is an invalid pointer or the comm cannot be read.  Since this plugin obtains the tasks via `list_tasks` [here](https://github.com/volatilityfoundation/volatility3/blob/03049f789559af5c4cdb56f343460178b52220f9/volatility3/framework/plugins/linux/envars.py#L131) and after #1542 `list_tasks` only yields valid tasks, this try/except block has effectively become redundant. Is there any chance to revert this?

ikelos:
Ok, you'll need to review and check the reversion, but if it's good then I'll merge it in.

PR #1514, opened 2025-01-03
Add proper exception handling in file descriptor enumeration

atcuno:
The lack of exception handling caused backtraces in multiple plugins over 50+ samples. These cleans up those missing checks.

### DIFF
--- a/volatility3/framework/symbols/linux/__init__.py
+++ b/volatility3/framework/symbols/linux/__init__.py
@@ -342,13 +342,12 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
         try:
             files = task.files
             fd_table = files.get_fds()
+            if fd_table == 0:
+                return None
+
+            max_fds = files.get_max_fds()
         except exceptions.InvalidAddressException:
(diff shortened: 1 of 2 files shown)


volatility3/framework/symbols/linux/__init__.py
```diff
@@ -307,15 +307,23 @@ def files_descriptors_for_process(
         symbol_table: str,
         task: interfaces.objects.ObjectInterface,
     ):
-        # task.files can be null
-        if not (task.files and task.files.is_readable()):
+        try:
```
ikelos, line 342: This doesn't need the exception handler setting up twice, all the calls can happen with the try/except and any failure within that block will result in a return `None`.

ikelos (commented):
There only needs to be a single try/except.  Pull them together and this is fine...

atcuno:
feedback addressed now @ikelos

ikelos:
I meant more that the single exception handler should span to encompass both methods (and anything in between).  Do we know that the second files call won't throw an exception?

PR #1515, opened 2025-01-03
Add exception handling in mountinfo enumeration path

atcuno:
The lack of exception handling caused backtraces in many samples in mass testing.

### DIFF
--- a/volatility3/framework/plugins/linux/mountinfo.py
+++ b/volatility3/framework/plugins/linux/mountinfo.py
@@ -6,7 +6,7 @@ import logging
 from collections import namedtuple
 from typing import Tuple, List, Iterable, Union
 
-from volatility3.framework import renderers, interfaces
+from volatility3.framework import renderers, interfaces, exceptions
 from volatility3.framework.configuration import requirements
 from volatility3.framework.interfaces import plugins
 from volatility3.framework.symbols import linux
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -96,9 +96,16 @@ def get_mountinfo(
...
 
         mnt_id: int = mnt.mnt_id
-        parent_id: int = mnt.mnt_parent.mnt_id
 
-        st_dev = f"{superblock.major}:{superblock.minor}"
+        # Many samples in mass testing backtraced from
+        # invalid superblocks being return
+        # mnt_parent can also smear so we check both
+        try:
+            parent_id: int = mnt.mnt_parent.mnt_id
```
gcmoreira, line 104: `mnt_parent` is a pointer.. this should be:
```python
if not (mnt.mnt_parent and mnt.mnt_parent.is_readable()):
    return None
```

Same thing for superblock:
```python
    def get_mnt_sb(self) -> int:
        """Returns a pointer to the super_block"""
```

ikelos: I would generally err towards testing before accepting (especially if there's gonna be extra code required either way).  I've tried in this project to be a little happier about exception handling, although ideally they'd only occur in exceptional circumstances not expected ones.  The good part about exception handling is that the try/catch can cover much larger chunks of code regardless of what exactly the issue was, so I'm trying to encourage try/except blocks that cover as much as could go wrong if something were to go wrong (but no more than that) so usually the body of a loop for instance, or something of that nature.

I'm gonna leave you two to figure out what's best ask permission/seek forgiveness, once you come to a consensus let me know, at the moment either one seems ok to me unless someone presents an argument as to why one is better or actively detrimental...

ikelos (commented):
This looks ok to me, but I'll leave @atcuno and @gcmoreira to come to a compromise they're both happy with and then go with that...

atcuno:
Closing this in favor of @gcmoreira 's work

PR #1517, opened 2025-01-03
Vma enumeration issues

atcuno:
These enumeration functions threw many backtraces and returned bad instances.

Thankfully, they are on a per-process basis anyway so the current code will cleanly bail out when smear/broken instances are found.

I plan to re-use this pattern in some other problematic places in Windows and Linux.

### DIFF
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -872,9 +872,10 @@ class mm_struct(objects.StructType):
 
         with contextlib.suppress(exceptions.InvalidAddressException):
             for vma in self._do_get_vma_iter():
-                    # this catches invalid instances returned by the internal interator functions
-                    vma.vm_start
-                    yield vma
+                # this catches invalid instances returned by the internal interator functions
+                vma.vm_start
+                yield vma
+
 
 class super_block(objects.StructType):
     # include/linux/kdev_t.h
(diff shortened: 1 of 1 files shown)


volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -866,6 +867,15 @@ def get_vma_iter(self) -> Iterable[interfaces.objects.ObjectInterface]:
         else:
             raise AttributeError("Unable to find mmap or mm_mt in mm_struct")
 
+    def get_vma_iter(self) -> Iterable[interfaces.objects.ObjectInterface]:
+        """Returns an iterator for the VMAs in an mm_struct. Automatically choosing the mmap or mm_mt as required."""
+
+        with contextlib.suppress(exceptions.InvalidAddressException):
```
ikelos, line 873: Presumably you want this inside the for loop, so it'll keep trying to return values?  I'm happy for it to bail out as soon as it hits a bad one, I'm just thinking it could potentially mask off a lot of good results as soon as one goes bad.  It also doesn't log, so the user has no way of knowing why or even that something's not quite right.  Smear is an issue, but ideally we should tell people when something's smear, rather than paper over it...

ikelos (commented):
I'm happy with the concept, but I'd like users (or at least developers) to know when something's gone wrong, and I wanted to check if you wanted to simply skip the bad entry (if this isn't a linked list for example) or bomb out at the first bad entry (which is what this code will do).

If it's got logging to say what's happening then I'm also not sure it needs a whole separate method, rather than just wrapping the old method in a well places `contextlib.suppress`?

atcuno:
Closing this in favor of @gcmoreira 's work

PR #1518, opened 2025-01-04
Prevent Linux's list_tasks from yielding completely broken process in…

atcuno:
…stances

### DIFF
--- a/volatility3/framework/plugins/linux/pslist.py
+++ b/volatility3/framework/plugins/linux/pslist.py
@@ -1,6 +1,7 @@
 # This file is Copyright 2021 Volatility Foundation and licensed under the Volatility Software License 1.0
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
+import logging
 import datetime
 import dataclasses
 import contextlib
@@ -14,6 +15,8 @@ from volatility3.framework.symbols import intermed
 from volatility3.framework.symbols.linux.extensions import elf
 from volatility3.plugins import timeliner
 from volatility3.plugins.linux import elfs
+
+vollog = logging.getLogger(__name__)
 
 
 @dataclasses.dataclass
@@ -112,6 +115,7 @@ class PsList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
             A TaskFields object with the fields to show in the plugin output.
         """
         name = utility.array_to_string(task.comm)
+
         if decorate_comm:
             if task.is_kernel_thread:
                 name = f"[{name}]"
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/linux/pslist.py
```diff
@@ -250,6 +254,14 @@ def list_tasks(
 
         # Note that the init_task itself is not yielded, since "ps" also never shows it.
         for task in init_task.tasks:
+            # the task list is often smeared in samples, espeically towards the end
+            # this stops the processing
+            if not task.has_valid_name_and_pid():
+                vollog.debug(
+                    f"Found an smeared/invalid task at offset {task.vol.offset:#x}"
```
ikelos, line 261: I'd drop the `smeared/` from this because otherwise the an sticks out.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -307,6 +307,22 @@ def section_strtab(self):
...
+        """
+        try:
+            pid = self.pid
+            name = utility.array_to_string(self.comm)
+        except exceptions.InvalidAddressException:
+            return False
+
+        # ensure the pid within the bounds of a signed int
+        # and that we have something of a name
+        return (0 < pid < 2147483647) and len(name) > 0
```
ikelos, line 324: If the type of pid is signed int, can it ever be outside these bounds?

ikelos (approved):
Yep, looks good, minor text alteration and a question but I'm happy otherwise.   5:)

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Aren't we opening the door to task hiding in memory forensics with this? Writing \x00 to the comm field could even be done from user space with just a syscall. Also, we definitely shouldn't use `break` there. Doing so not only allows a single task to hide but also exposes all subsequent tasks to the same risk.

gcmoreira: Minor but technically speaking it has to be  (0 < pid <= 2147483647) 
```python
>>> struct.unpack('i', b"\xff\xff\xff\x7f")
(2147483647,)
>>> struct.unpack('i', b"\xff\xff\xff\x80")
(-2130706433,)
```

volatility3/framework/plugins/linux/pslist.py
```diff
@@ -250,6 +254,14 @@ def list_tasks(
...
+                )
+                break
```
gcmoreira, line 263: `continue` please

ikelos:
Don't think this was supposed to be closed, was it?

ikelos:
Oh, maybe it was?

PR #1529, opened 2025-01-05
Tiny comment changes

j-t-1:


volatility3/framework/plugins/windows/cmdscan.py
```diff
@@ -100,7 +101,7 @@ def get_command_history(
             kernel_symbol_table_name: The name of the table containing the kernel symbols
             config_path: The config path where to find symbol files
             procs: list of process objects
```
ikelos: Seems odd to leave list capitalized if we're changing `an` to `An` just below it?

volatility3/framework/plugins/windows/cmdscan.py
```diff
@@ -67,6 +67,7 @@ def get_filtered_vads(
 
         Args:
             conhost_proc: the process object for conhost.exe
+            size_filter: filter (keep) vads less than this size (bytes)
```
ikelos: I understand the reason for reminding people how `filter` works in the volatility world (filter down to, not filter out), but `(keep)` doesn't seem like the clearest way of doing this.  This should work the way that [python's `filter` function](https://docs.python.org/3/library/functions.html#filter) works...

Perhaps we could go with `size above which vads will not be returned`?

ikelos (commented):
Just a couple minor points, but thanks for going through all the documentation for us!

PR #1530, opened 2025-01-05
Update case insensitive check

j-t-1:
Update link and use casefold() instead of lower().


ikelos (approved):
I was worried this might be overkill, but since it's done on both sides, it should be fine.  Thanks...

PR #1532, opened 2025-01-07
Linux: Fix copy-pasted module docstrings

dgmcdona:
This updates the module docstrings for five modules that duplicate the
docstring from the `proc` module. This was presumably the result of using
the `proc` module as a template for the others.


ikelos (approved):
Hehehe, good catch, thanks!  5:D

Issue #1535, opened 2025-01-08
Testing: Issues with `pyproject.toml` and `jsonschema` 


gcmoreira:
For the GitHub tests, we initially used [requirements-testing.txt](https://github.com/volatilityfoundation/volatility3/blob/v2.8.0/test/requirements-testing.txt) to manage the minimal installation of Python modules specifically for testing. With the introduction of the latest changes in `pyproject.toml`, as part of [Update pyproject.toml and remove requirements and setup files](https://github.com/volatilityfoundation/volatility3/pull/1342) an issue was introduced. 

This occurs when the `jsonschema` module was added to the `test` target indirectly through the `dev` dependency. 
```
dev = [
    "volatility3[full,cloud]",
    "jsonschema>=4.23.0,<5",
...
]
test = [
    "volatility3[dev]",
...
]
```

With this module present, as shown [here](https://github.com/volatilityfoundation/volatility3/blob/a98a23fa41395b9bbd961f6e50d0a0f19201fcc7/volatility3/schemas/__init__.py#L95) and [here](https://github.com/volatilityfoundation/volatility3/blob/a98a23fa41395b9bbd961f6e50d0a0f19201fcc7/volatility3/schemas/__init__.py#L102), Volatility3 will perform JSON schema validation of each ISF. This validation introduces substantial overhead, significantly impacting the execution time of each test.

To validate the ISF JSON schema should not be part of the GitHub tests, and it's better this module to be excluded during that process. We need to find a way to decouple `jsonschema` from the `test` target.

ikelos:
That's entirely doable, but just before we do that I'd like to look into making jsonschema much, much quicker than it is at the moment by compiling the schema at the start.  Gimme a few days to fix that up and get it in and then let's revisit this please...

ikelos:
Can you check if this was enough of a speedup please?  If not, we could maybe still move to `fastjsonschema`, but I'll need to implement special checking to relax JSON timestamp formats, since a lot of our existing ones (and previously created files, such as all the windows one and potentially those made by dwarf2json) fail due to not including a timezone at the end...  5:S

`Fastjsonschema` also doesn't really tell you where the failure occurred so is less useful for users and devs alike.  I'd probably sooner investigate another method of speed-up, such as a configuration option to disable validation when requested (not just when `jsonschema` isn't available).

atcuno:
I think we are good here. I haven't had the 15 minute+ startup times from validation in a while.

[...]

PR #1537, opened 2025-01-09
Issues/improve jsonschema speed

ikelos:
This precompiles the schema file for jsonschema, meaning that (for some reason) it's already faster than the old combined method, but also if many json files are verified against the same schema there'll be a several second win each additional time (due to not having to recompile the schema).

I tried using fastjsonschema but the error messages are less descriptive and it fails validation for most of our existing handcoded and generated files, because the timestamps they generate don't include the timezone offset at the end (+XX:XX) finding all datetimes within a json file would be a pig (and slow) so better to change the validator somehow (still investigating) and/or figure out why jsonschema allows the old times and fastjsonschema does not...

Anyway, lemme know what you think and if the changes speed-up your testing workflow much?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,10 +33,9 @@ cloud = [
 
 dev = [
     "volatility3[full,cloud]",
-    "jsonschema>=4.23.0,<5",
+    "fastjsonschema>=2.21.1,<3",
     "pyinstaller>=6.11.0,<7",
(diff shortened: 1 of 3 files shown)


ikelos:
Possibly enough to resolve #1535 , but I won't close it off in case this speed up isn't enough...

@atcuno could you give this a test please?

ikelos:
Sorry, just trying to verify whether this sped things up enough.

PR #1538, opened 2025-01-10
Introduce customizable plugin arparse epilog

Abyss-W4tcher:
Hi 👋, 

This PR introduces a feature to specify an argparse epilog for plugins, allowing to extend a plugin description easily, without modifying the short description.  Sample from https://docs.python.org/3/library/argparse.html#epilog:

```python
parser = argparse.ArgumentParser(
    description='A foo that bars',
    epilog="And that's how you'd foo a bar")
parser.print_help()

usage: argparse.py [-h]

A foo that bars

options:
 -h, --help  show this help message and exit

And that's how you'd foo a bar
```

### DIFF
--- a/volatility3/framework/interfaces/plugins.py
+++ b/volatility3/framework/interfaces/plugins.py
@@ -112,7 +112,7 @@ class PluginInterface(
     # Be careful with inheritance around this (We default to requiring a version which doesn't exist, so it must be set)
     _required_framework_version: Tuple[int, int, int] = (0, 0, 0)
     """The _version variable is a quick way for plugins to define their current interface, it should follow SemVer rules"""
-    _argparse_epilog: str = None
+    additional_description: str = None
     """Display additional description of the plugin after the description of the arguments. See: https://docs.python.org/3/library/argparse.html#epilog"""
 
     def __init__(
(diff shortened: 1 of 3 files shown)


volatility3/framework/interfaces/plugins.py
```diff
@@ -112,6 +112,8 @@ class PluginInterface(
     # Be careful with inheritance around this (We default to requiring a version which doesn't exist, so it must be set)
     _required_framework_version: Tuple[int, int, int] = (0, 0, 0)
     """The _version variable is a quick way for plugins to define their current interface, it should follow SemVer rules"""
+    _argparse_epilog: str = None
```
ikelos: This technically shouldn't be a private variable, since it's being access outside of the plugin (although I'm not certain whether we adhere to that with the _version or _required_framework_version).

Also, if it's defined in the interface, then `getattr` will always return the value, so you can just access the attribute directly.  I don't think I'd include argparse in the variable name, that's quite a specific yet unnecessary piece of information.  Particularly when the information may be used by a Web UI or something...  I assume it could be `additional_help` or `additional_description` or `extra_` something?

ikelos (commented):
Thanks, I'm happy with the idea, just want to get the interfaces changes nailed down properly...

volatility3/framework/interfaces/plugins.py
Abyss-W4tcher: Sure, happy to make it more understandable :)

ikelos (commented):
Much happier with this, but it's still an additional change to the framework, so please include a MINOR version bump (it'll need that before it can be merged)...

Abyss-W4tcher:
This should be all good, I keep a reminder here to add this feature to the plugin development doc. later.

ikelos (approved):
Thanks  5:)

PR #1539, opened 2025-01-10
Small readability improvements

j-t-1:


j-t-1:
@ikelos i wrongly deleted the `break`. Can you modify this so the `break` is put back and then if okay commit it?

ikelos:
It's tricky for me to change other people's PRs, but you can just add a commit and push that up and it'll update the PR?  I won't merge it until then...  5:)

j-t-1:
Modified again so that the `break` is still removed. Please check this new refactor.

ikelos (approved):
I don't overly have a problem using a variable as a temporary holder as we used to, but I can don't have an issue spinning up another variable for it, I think the performance hit is minimal.

Abyss-W4tcher:
Hi, sorry I might be mistaken, but why was the `break` statement removed ?

atcuno:
This seems like a really bad change. 

Previously, this line was used to only set the `aslr_shift` that gets returned if `tmp_aslr_shift` had a validate value:

https://github.com/volatilityfoundation/volatility3/pull/1539/files#diff-df85c3307d0f0fe9b1e87c22c85b85e63d2693fc0df9d89e36d537953dda3196L210

Now, even if we don't find a validated value, the loop will bleed through whatever bad value we found last.

ikelos:
Sorry, my bad.  The github interfaces seems to be really good at not showing me all of the changes a commit will make in the same place that it would show me all the changes if I'd clicked on a slightly different thing.  I'm reverting it.

ikelos:
(Interesting that our test runs didn't pick up the issue though, perhaps I rely on them too much).  5:S

atcuno:
To auto detect this, you would need a sample that picks up bad shifts but doesn't actually have ASLR enabled. This is where the regression set (more samples per OS) will come in handy.

ikelos:
Ok, all reverted.  I'll probably push back more on seemingly cosmetic changes in the future.

j-t-1:
@Abyss-W4tcher I wrongly convinced myself that the `break` could be removed. Thank you for quickly highlighting: going by the above conversation (verified by @atcuno) this helped minimize the time window it was wrong. @ikelos apologies for this erroneous PR.

Abyss-W4tcher:
@j-t-1 I guess you could make a new PR without removing the `break` statement ?

PR #1540, opened 2025-01-10
Windows: Certificates - handle uncaught RegistryFormatException

dgmcdona:
Changes variable import to module import, and catches an unhandled
`RegistryFormatException` in certificates.py


ikelos (approved):
Yep, looks good (and thanks for fixing the non-standard import that slipped through there!).  5:)

PR #1543, opened 2025-01-13
Linux and Windows: Ensure linked list object extensions consistently yield valid entries

gcmoreira:
In the context of #1516 and related issues, we noticed that core functions responsible for yielding entries from linked list sometimes include entries that are invalid. 

This pull request ensures that each stage of the linked list traversal process is validated, guaranteeing that the yielded objects are always valid.


ikelos (approved):
Ok, looks good.  The two methods seem quite similar, but give they're for different operating systems, I'm not sure there's much value in combining them at the moment.

PR #1544, opened 2025-01-13
Linux: Ensure VMA enumration functions yield only valid objects consistently

gcmoreira:
In the context of #1516 and related issues, we noticed some issues with Linux VMA enumeration APIs.
- When working with kernels prior to 6.1, the `_get_mmap_iter()` function yielded `vm_area_struct*`. However, with the introduction of the **maple tree** implementation and the `get_vma_iter()` function, the behavior changed. For kernels < 6.1, it continued yielding `vm_area_struct*`, whereas for kernels >= 6.1, it started yielding `vm_area_struct`, making the API inconsistent.
- The `_get_mmap_iter()` method failed to properly validate pointers, which could lead to processing invalid VMAs.
- The `_get_mmap_iter()` method incorrectly checked for seen pointers. `mm_struct.mmap` is already a pointer, so performing `self.mmap.vol.offset` actually returns the address of the pointer itself `&mm_struct.mmap`, not the value it points to.

This PR addresses all of the issues mentioned above. Additionally, it enhances the docstrings for better clarity and documentation.


ikelos (approved):
Yep, looks reasonable, thanks.

PR #1545, opened 2025-01-13
Linux - Harden mountinfo API and several related fixes

gcmoreira:
In the context of #1516 and related issues, this PR addresses several issues in the mountinfo API code and its associated functions.

This PR includes:
- Fix container_of(). Since object() can create and return an invalid object, we ensure this method will return a valid one checking it address using the layer. 
- Ensure _get_tasks_mountpoints() will always return valid mountpoint checkin all the pointers involved.
- Ensure do_get_path() will always return a valid string.
- Fix struct_file by removing invalid members. The `f_dentry` and `f_vfsmnt` members were incorrectly included, though they were never part of struct file. These were preprocessor macros used as shortcuts within kernel code but should not be present in the structure itself.
- Docstring improvements


volatility3/framework/plugins/linux/mountinfo.py
```diff
@@ -36,7 +36,7 @@ class MountInfo(plugins.PluginInterface):
     """Lists mount points on processes mount namespaces"""
 
     _required_framework_version = (2, 2, 0)
-    _version = (1, 2, 3)
```
ikelos, line 39: I don't think this is an externally visible change (the inputs stay the same and I expect the outputs to be similar too), so should be fine to leave it as it is, but also happy to have it bumped.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -121,22 +121,34 @@ def get_path_mnt(cls, task, mnt) -> str:
...
-            vfsmnt (vfsmount *): A pointer to the vfsmount
+            vfsmnt (vfsmount/vfsmount *): A vfsmount object (kernels >= 3.3) or a
+                vfsmount pointer (kernels < 3.3)
 
         Returns:
             str: Pathname of the mount point or file
         """
 
+        if not (rdentry and rdentry.is_readable() and rmnt and rmnt.is_readable()):
+            return ""
```
ikelos, line 152: Is a blank string the right return value?  I'd have though (given the API already allowed it to be `None`) that None would be more descriptive rather just  a blank string?  It makes it harder for people to distinguish an error from just an empty string...

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -121,22 +121,34 @@ def get_path_mnt(cls, task, mnt) -> str:
...
+
+        if isinstance(vfsmnt, objects.Pointer) and not (rmnt and rmnt.is_readable()):
```
ikelos: I'd probably rewrite this as:

```
if (rmt and rmnt.is_readable()):
  if not rdentry...:
    return None/""
else:
  if ininstance(vfsmnt...:
    return None/""
  ```
  
  It's fine as is, just reduces the complexity of the if clauses which should make it easier to read/parse.

File: volatility3/framework/symbols/linux/extensions/__init__.py [hunk 4]
```diff
@@ -1504,39 +1500,40 @@ def is_valid(self):
...
+        'mount' struct, versus later versions that include it.
+        See 7d6fec45a5131918b51dcd76da52f2ec86a85be6.
 
-        The 'mnt_parent' member was moved from struct 'vfsmount' to struct
-        'mount' when the latter was introduced.
+        # Following that commit, also in kernel version 3.3 (3376f34fff5be9954fd9a9c4fd68f4a0a36d480e),
+        # the 'mnt_parent' member was relocated from the 'vfsmount' struct to the newly
+        # introduced 'mount' struct.
 
         Alternatively, vmlinux.has_type('mount') can be used here but it is faster.
```
ikelos: Just spotted this no "alternative blah can be used *BUT* it is faster"?  Why wouldn't we use it if it's faster?  Either it needs to be this, or something about this needs changing...

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1504,39 +1500,40 @@ def is_valid(self):
...
-        Depending on the kernel version, the calling object (self) could be
-        a 'vfsmount \\*' (<3.3.8) or a 'vfsmount' (>=3.3.8). This way we trust
-        in the framework "auto" dereferencing ability to assure that when we
-        reach this point 'self' will be a 'vfsmount' already and self.vol.offset
-        a 'vfsmount \\*' and not a 'vfsmount \\*\\*'. The argument must be a 'vfsmount \\*'.
+        Depending on the kernel version, see 3376f34fff5be9954fd9a9c4fd68f4a0a36d480e,
+        the calling object (self) could be a 'vfsmount *' (<3.3) or a 'vfsmount' (>=3.3).
+        This way we trust in the framework "auto" dereferencing ability to assure that
+        when we reach this point 'self' will be a 'vfsmount' already and self.vol.offset
+        a 'vfsmount *' and not a 'vfsmount **'. The argument must be a 'vfsmount *'.
```
ikelos: The reason for the `\\` should be to escape the * and make sure it prints properly.  If that was overkill, that's fine, but I'd like confirmation someone has tested this and that the stars do actually come out in the compiled documentation as stars and not start-bold-emphasis or something strange...

ikelos (commented):
Generally happy with all of this, although not sure why the function signature for one bit changed (and without that change, there's no need for version bumps that I can see).  Needs a response to some of the points raised but no showstoppers at least...

volatility3/framework/symbols/linux/__init__.py
gcmoreira: ups, that was actually a bug. The second `if` was meant to check the vfsmount pointer,  I just fixed it.

volatility3/framework/symbols/linux/__init__.py
gcmoreira: I know, actually the original typing was wrong, it never returned `None` and has always returned a `str`. 
Also, the callers expect a `str` and are not prepared to receive a `None`. I guess it was just a bug in the typing. 
Anyway, while I agree that returning None would be a better solution, making this change would involve extensive modifications, which I thought were outside the scope of this PR. For example, the following callers and their related code would also need to be updated.
LinuxUtilities._get_path_file() -> str:
LinuxUtilities.get_path_mnt() -> str:
LinuxUtilities.path_for_file() -> str:

volatility3/framework/plugins/linux/mountinfo.py
gcmoreira: done, updated only the patch version

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: No worries, the escaping is back. That said, I'm not a big fan of massacre the Python docstring just to make it render correctly in HTML. Ideally, it should work the other way around, the module converting the docstring to whatever documentation format should handle escaping as needed for the final format. For example, a \\* might render correctly in HTML but could be the wrong escaping for other formats

gcmoreira: I also tried to test it, but the documentation is currently broken. To fix this, we need the excellent work from https://github.com/volatilityfoundation/volatility3/pull/1548 by @dgmcdona to restore functionality.

gcmoreira:
@ikelos thanks for the review. All the observations were addressed

volatility3/framework/symbols/linux/__init__.py
ikelos: Ugh, this feels like a choice between "do some work to fix things properly or let something we know is wrong through to the codebase".  5:S  I really don't like letting something we know is wrong go through.  I think we should choose the right option for the method to return (which I think is to allow `None` and keep the old typing, but I'm not against a change as long as it's suitable) and then fix up anything that makes bad assumptions.  You might be able to figure out what the callers *would* have done in the case of an error and cope with it whilst still returning a string, or we have to do a MAJOR bump, and it becomes a pig but we figure out how best to do it so we could do it again in the future if needed.

I don't really want to be held hostage to fixes because of old decisions, so knowing how to bump this or how to keep everything better laid out/separated would be good.  I think splitting things out of LinuxUtilities into subitems would be a good opportunity to a) improve the granularity of versioning and b) bump the version at that time anyway.  The trick will be to get as much abstracted out of LinuxUtilities in one go as possible, so that the dependency only needs to go up once.

ikelos: Any thoughts on this piece, it's still unresolved as best as I can see?

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: The typo's still here too...  5;P

volatility3/framework/symbols/linux/__init__.py
gcmoreira: Not much to add here. As I mentioned above, everything in the call chain returns and expects a string, so at least now it's consistent with the typing. We can update it to return None in these cases in a separate PR

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: Yeah, it looks like I was thinking about getting vmlinux to check that. I probably wasn't aware of `context.symbol_space` at the time, or it just didn't come to mind. Since `get_module_from_volobj_type()` is more expensive than `self.has_member("mnt_parent")` but technically produces the same result, I chose the latter.

I will update this here to unblock it, but I still feel it's not directly related to the goal or changes in this PR

ikelos: Oh, it was more just the text, "but it was faster, but..."

volatility3/framework/symbols/linux/__init__.py
ikelos: I'm concerned about making the change and then forgetting to change it back, rather than just not making the change in the first place.  `""` isn't a suitable way to identify error conditions, so I'd like to get a suitable way in place (or be satisfied that returning `""` won't cause confusion down the road)...

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: hah sorry, I changed it anyway... now it doesn't need the clarification so it also fixed the typo :)

volatility3/framework/symbols/linux/__init__.py
gcmoreira: ok, I can rollback these changes to:
```
def do_get_path(cls, rdentry, rmnt, dentry, vfsmnt) -> Union[None, str]:
```
if you want. Even if the current implementation will never return None

ikelos: Yes please.  If you're happy that returning an empty string is the right thing to do here then I'll go with it, you know the code way better than me, I just wanted to make sure it was thought about...

gcmoreira: Sorry, don't get me wrong. I do think distinguishing the error is a good idea. I do understand the differences and benefits of returning "" vs None, but I don't think this PR is the right place to change that. It would require updates in multiple places, and the focus here is on stabilizing pointer-related issues.

```
Anyway, while I agree that returning None would be a better solution, making this change would involve extensive modifications, which I thought were outside the scope of this PR.
```

ikelos: Ok, I'm fine with that, but changing the signature would be a major bump so that still needs to go back to allowing None, even if we never return it, unfortunately...

gcmoreira:
@ikelos hold on with this as it seems there's a commit missing

gcmoreira:
@ikelos no worries, it's all good! Turns out it was just how GitHub was displaying the changes, pretty confusing at first

volatility3/framework/symbols/linux/__init__.py
gcmoreira: Done, so with that typing back, it looks like we probably won't need to update the LinuxUtilities version to 2.3.0, right?

ikelos: Err, I think the change from taking *vfsmnt to vfsmnt/*vfsmnt is an additive change, so I'd leave it in there...

ikelos (approved):
Ok, looks good, let's go for it.  5:)

PR #1548, opened 2025-01-14
Fix dependency conflicts in pyproject.toml

### LINKED ISSUE #1546: pyproject.toml: dependency conflicts
I've been playing with using [uv](https://astral.sh/uv) to automatically create virtual environments, install dependencies, and run the `vol.py` script in a single command with `uv run --python 3.8.10 vol.py` and it has produced some errors when attempting to resolve volatility3's dependencies. 

The first issue seems to be a clear bug, which is that the sphinx `doc` [dependency](https://github.com/volatilityfoundation/volatility3/blob/a6777146f6fd789c10686b9ed1f8a5ea3521f9a5/pyproject.toml#L51) is impossible to resolve - the `<` and `>` need to be flipped since I'm assuming that we want the version to be within that range.

The second one that I know of so far is that pyinstaller 6.11.0 (our minimum pyinstaller required version) pinned `pefile != 2024.8.26` due to regressions introduced in `pefile`: https://github.com/pyinstaller/pyinstaller/blob/ee05de897aa82abad2efe7ce2e8d45555930d5a4/doc/CHANGES.rst#pyinstaller-core

However, we have `pefile` as a required dependency pinned to `pefile >= 2024.8.26`,  and 2024.8.26 is the highest available version on PyPi, which produces an unresolvable conflict.

I'll see if any others come up while attempting to resolve these.

dgmcdona:
Closes #1546 

This makes three changes to fix dependency resolution
- Fixes a bug where the sphinx version is unresolvable, and pins it to
  8.0.2 (the minimum version required by the minimum version of
  sphinx-autodoc-typehints)
- Lowers the required version of `pefile` in order to allow us to use
  the `pyinstaller` version specified in the `dev` dependencies
- Bumps the python version to 3.10 in order to support sphinx 8.0.2

I know bumping the minimum Python version is not desirable, but I wasn't sure which `sphinx`/`sphinx-autodoc-typehints` versions would be acceptable due to my unfamiliarity with our doc requirements. We can lower those if an acceptable version can be found that supports python 3.8. There isn't a way to specify different required Python versions for different sets of dependencies within `pyproject.toml`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ readme = "README.md"
 authors = [
     { name = "Volatility Foundation", email = "volatility@volatilityfoundation.org" },
 ]
-requires-python = ">=3.10.0"
+requires-python = ">=3.8.0"
 license = { text = "VSL" }
 dynamic = ["version"]
 
@@ -48,8 +48,8 @@ test = [
 
 docs = [
     "volatility3[dev]",
-    "sphinx==8.0.2",
-    "sphinx-autodoc-typehints>=2.5.0,<3",
+    "sphinx>=7,<=7.1.2",
+    "sphinx-autodoc-typehints==2.0.1",
     "sphinx-rtd-theme>=3.0.1,<4",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,12 +6,12 @@ readme = "README.md"
 authors = [
     { name = "Volatility Foundation", email = "volatility@volatilityfoundation.org" },
 ]
-requires-python = ">=3.8.0"
+requires-python = ">=3.10.0"
 license = { text = "VSL" }
 dynamic = ["version"]
 
 dependencies = [
-    "pefile>=2024.8.26",
+    "pefile>=2023.2.7",
 ]
 
 [project.optional-dependencies]
@@ -48,7 +48,7 @@ test = [
 
 docs = [
     "volatility3[dev]",
-    "sphinx>=8.0.0,<7",
+    "sphinx==8.0.2",
     "sphinx-autodoc-typehints>=2.5.0,<3",
     "sphinx-rtd-theme>=3.0.1,<4",
 ]
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ more details.
 
 ## Installing
 
-Volatility 3 requires Python 3.10.0 or later and is published on the [PyPi registry](https://pypi.org/project/volatility3).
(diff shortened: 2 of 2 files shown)


dgmcdona:
I reverted the Python version bumps and found the highest versions of `sphinx-autodoc-typehints` and `sphinx` that still support Python 3.8. Installing those versions and building the docs with `cd doc; make html` seemed to work fine, although there are some existing syntax errors within the documentation that don't cause the build to fail.

I also looked through the git history to see if there was a reason that we require such a high version of `pefile` and couldn't find one, it seems that it was bumped in [this commit](https://github.com/volatilityfoundation/volatility3/commit/1fd51772600dc14ad6893e76362c9b8f599be180)

pyproject.toml
```diff
@@ -11,7 +11,7 @@ license = { text = "VSL" }
 dynamic = ["version"]
 
 dependencies = [
-    "pefile>=2024.8.26",
```
ikelos, line 14: This is the wrong way to fix the issue.  I don't think pefile is at fault, I believe it's the way that pyinstaller used it, and pyinstaller is only used by devs to construct an exe (and it's only a performance degradation), so I think we bump pyinstaller down (it used to start at 6.5, but during the toml update it got bump to the latest version at that time).

pyproject.toml
```diff
@@ -48,8 +48,8 @@ test = [
 
 docs = [
     "volatility3[dev]",
-    "sphinx>=8.0.0,<7",
```
ikelos, line 51: I've tested and it appears to builds with version 8 ok, so I'm not sure this needs to drop to 7, I think the higher version should bump to 9.

pyproject.toml
```diff
@@ -48,8 +48,8 @@ test = [
...
-    "sphinx-autodoc-typehints>=2.5.0,<3",
```
ikelos, line 52: I assume this change is to work with the version of sphinx available?  Again, I believe the lower bound was pulled up artificially when the `pyproject.toml` inhereited the values from the `requirements.txt`.

ikelos (commented):
Thanks, I think this messes with a few too many things.  Many of the requirements were artificially bumped up by the person pushing up to move to pyproject.toml, but I'm not sure there's good cause for blocking the older versions.

I also don't really want to fix on an old version of pefile because pyinstaller (used once and only by devs) has a performance issue for the one time we run it.

ikelos:
Could you please review #1558, and if you've got the ability to easily check the dependencies and outputs, check that everything's ok?  I've left my comments for the reasons I took a different route in this PR, but feel free to point out problems in the new one, or reply here if you're not happy with any of the reasoning...  Thanks!

PR #1550, opened 2025-01-15
Linux: Add support for Intel 32bit with PAE

gcmoreira:
In the [vmcoreinfo PR](https://github.com/volatilityfoundation/volatility3/pull/1332), we added support for Intel 32-bit PAE, however, this feature was still missing in the original `LinuxIntelStacker` layer.

The changes in this PR are based on the size of the `pkmap_count` variable, which varies depending on whether PAE is enabled. From [2.3.27](https://elixir.bootlin.com/linux/2.3.27/source/mm/highmem.c#L95) to [2.3.28](https://elixir.bootlin.com/linux/2.3.28/source/mm/highmem.c#L96) the `LAST_PKMAP` macro value for PAE was **2048**. Starting from [2.3.29](https://elixir.bootlin.com/linux/2.3.29/source/include/asm-i386/highmem.h#L47), it was reduced to **512**. See also the latest kernel [6.13-rc3](https://elixir.bootlin.com/linux/v6.13-rc3/source/arch/x86/include/asm/pgtable_32_areas.h#L22)
```c
#ifdef CONFIG_X86_PAE
#define LAST_PKMAP 512
#else
#define LAST_PKMAP 1024
#endif

static int pkmap_count[LAST_PKMAP];
```

### DIFF
--- a/volatility3/framework/automagic/linux.py
+++ b/volatility3/framework/automagic/linux.py
@@ -71,10 +71,9 @@ class LinuxIntelStacker(interfaces.automagic.StackerLayerInterface):
                 elif "init_level4_pgt" in table.symbols:
                     layer_class = intel.LinuxIntel32e
                     dtb_symbol_name = "init_level4_pgt"
-                elif (
-                    "pkmap_count" in table.symbols
-                    and table.get_symbol("pkmap_count").type.count == 512
-                ):
+                elif "pkmap_count" in table.symbols and table.get_symbol(
+                    "pkmap_count"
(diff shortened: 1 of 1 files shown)


atcuno:
@ikelos these needs priority as its preventing analysis of basically every 32bit Linux sample in the testing set

ikelos (approved):
Ok, if this is a reliable distinguisher (and doesn't throw symbol not found exceptions) then looks ok to me...

PR #1556, opened 2025-01-16
Stable

ikelos:
Merge stable updates back into develop


ikelos:
Abandoning until vmayarascan and vadyarascan get sotred out.

PR #1557, opened 2025-01-16
Linux - Parity release - Harden pointer validation in several plugins and APIs

### LINKED ISSUE #1522: linux sockstat broken from missing pointer check
Sample: broken_mount_enumeration_ubuntu_22_04_x64.zip

Plugin: linux.sockstat.Sockstat

@gcmoreira this code path has numerous issues and is a good one for you to debug the member / pointer access.

1) This should become `get_inode` regardless so we remove raw accesses to `d_inode`:

 https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/sockstat.py#L524

2) This dereference() is called without a check after, which is what causes the backtrace below:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/sockstat.py#L536

3) This does `instance.<pointer member>.member`, which will cause a backtrace if `nsproxy` is invalid:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/sockstat.py#L549

This is the same root cause as https://github.com/volatilityfoundation/volatility3/issues/1523, where accessing through `nsproxy` does cause a backtrace. My current view (mention in the ticket before this one) is that we should make as many `get_<member>` accessors as possible for pointer members to avoid the try/except all over the plugins while still respecting the invalid address exception.

```
25-01-03 18:34:41 volatility3.cli DEBUG    Traceback (most recent call last):
    renderer.render(grid)
    grid.populate(visitor, outfd)
  File "/home/ub/volatility3/volatility3/framework/renderers/__init__.py", line 240, in populate
  File "/home/ub/volatility3/volatility3/framework/plugins/linux/sockstat.py", line 601, in _generator
    for (
  File "/home/ub/volatility3/volatility3/framework/plugins/linux/sockstat.py", line 538, in list_sockets
    sock_type = sock.get_type()
                ^^^^^^^^^^^^^^^
  File "/home/ub/volatility3/volatility3/framework/symbols/linux/extensions/__init__.py", line 1781, in get_type
    return linux_constants.SOCK_TYPES.get(self.sk_type, "")
                                          ^^^^^^^^^^^^
  File "/home/ub/volatility3/volatility3/framework/objects/__init__.py", line 961, in __getattr__
    member = template(context=self._context, object_info=object_info)
  File "/home/ub/volatility3/volatility3/framework/objects/templates.py", line 96, in __call__
    return self.vol.object_class(
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ub/volatility3/volatility3/framework/objects/__init__.py", line 168, in __new__
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ub/volatility3/volatility3/framework/objects/__init__.py", line 202, in _unmarshall
    data = context.layers.read(
           ^^^^^^^^^^^^^^^^^^^^
                                          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ub/volatility3/volatility3/framework/layers/intel.py", line 166, in _translate
volatility3.framework.exceptions.PagedInvalidAddressException: Page Fault at entry 0xcccccccccccccccc in page entry
```

gcmoreira:
This PR is the last in the series related to fix the issues mentioned in https://github.com/volatilityfoundation/volatility3/pull/1516 and related tickets. Additionally, I fixed other potential issues I found with pointers.

fix: #1517, #1522

### DIFF
--- a/volatility3/framework/plugins/linux/kmsg.py
+++ b/volatility3/framework/plugins/linux/kmsg.py
@@ -136,7 +136,12 @@ class ABCKmsg(ABC):
         """
 
     def get_string(self, addr: int, length: int) -> str:
-        txt = self._context.layers[self.layer_name].read(addr, length)  # type: ignore
+        layer = self._context.layers[self.layer_name]
+        if not layer.is_valid(addr, length):
+            return ""
+
+        txt = layer.read(addr, length)
+
         return txt.decode(encoding="utf8", errors="replace")
 
     def nsec_to_sec_str(self, nsec: int) -> str:
@@ -281,9 +286,13 @@ class Kmsg_3_5_to_3_11(ABCKmsg):
         log_struct_name = self._get_log_struct_name()
         log_struct_size = self.vmlinux.get_type(log_struct_name).size
         dict_offset = msg.vol.offset + log_struct_size + msg.text_len
-        dict_data = self._context.layers[self.layer_name].read(
-            dict_offset, msg.dict_len
-        )
+        layer = self._context.layers[self.layer_name]
+        try:
+            dict_data = layer.read(dict_offset, msg.dict_len)
+        except exceptions.InvalidAddressException:
+            vollog.debug("Unable to read kmsg dict from 0x%x", dict_offset)
+            return None
+
         for chunk in dict_data.split(b"\x00"):
             yield " " + chunk.decode()
 
--- a/volatility3/framework/plugins/linux/kmsg.py
+++ b/volatility3/framework/plugins/linux/kmsg.py
@@ -5,7 +5,7 @@ import re
 import logging
 from abc import ABC, abstractmethod
 from enum import Enum
-from typing import Generator, Iterator, List, Tuple
+from typing import Generator, Iterator, List, Tuple, Union
 
 from volatility3.framework import (
     class_subclasses,
@@ -135,10 +135,11 @@ class ABCKmsg(ABC):
             bool: True if the kernel being analyzed fulfill the class requirements.
         """
 
-    def get_string(self, addr: int, length: int) -> str:
+    def get_string(self, addr: int, length: int) -> Union[str, None]:
         layer = self._context.layers[self.layer_name]
         if not layer.is_valid(addr, length):
-            return ""
+            vollog.error("Failed to read log record at address 0x%x", addr)
+            return None
 
         txt = layer.read(addr, length)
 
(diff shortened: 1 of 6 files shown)


volatility3/framework/plugins/linux/kmsg.py
```diff
@@ -136,7 +136,12 @@ def symtab_checks(cls, vmlinux: interfaces.context.ModuleInterface) -> bool:
         """
 
     def get_string(self, addr: int, length: int) -> str:
-        txt = self._context.layers[self.layer_name].read(addr, length)  # type: ignore
+        layer = self._context.layers[self.layer_name]
+        if not layer.is_valid(addr, length):
+            return ""
```
ikelos: Would an exception or `None` not be better than returning a blank string?  My concern is the string having a length of 0 or similar, and no indication that it failed otherwise...

ikelos (commented):
All looks good except returning a blank string as an error indication.  How would someone differentiate between a blank string, and the data not being present, or would it not matter?

volatility3/framework/plugins/linux/kmsg.py
gcmoreira: have a look at the latest changes, thanks

volatility3/framework/plugins/linux/kmsg.py
```diff
@@ -135,8 +135,14 @@ def symtab_checks(cls, vmlinux: interfaces.context.ModuleInterface) -> bool:
             bool: True if the kernel being analyzed fulfill the class requirements.
         """
 
-    def get_string(self, addr: int, length: int) -> str:
-        txt = self._context.layers[self.layer_name].read(addr, length)  # type: ignore
+    def get_string(self, addr: int, length: int) -> Union[str, None]:
+        layer = self._context.layers[self.layer_name]
+        if not layer.is_valid(addr, length):
+            vollog.error("Failed to read log record at address 0x%x", addr)
```
ikelos: This throws a pretty big error, is it not a warning, or even just an info (not displayed by default)?  How big an issue is the string being invalid?

gcmoreira: Hehe, not that big, which is why it wasn't considered initially. But your other comment made me think you might want this to be visible. This change was only to report the error, the code already handled this because an empty string results in an empty list when `"".splitlines()` is called.
So, a `warning` is fine?

ikelos: Ok, I just can't tell how often it might flash up for the user, but if you think it's worth their seeing during a normal run then warning is fine...

gcmoreira: done

PR #1558, opened 2025-01-17
Issues/issue1546

### LINKED ISSUE #1546: pyproject.toml: dependency conflicts

ikelos:
This is designed to supercede #1548.  It should allow older versions of pyinstaller without the requirement on an older pefile.  It also fixes the sphinx requirements to relax the older version numbers, given we haven't found reason to disallow the older versions.

Fixes #1546.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ test = [
 
 docs = [
     "volatility3[dev]",
-    "sphinx>=8.0.0,<9",
+    "sphinx>=4.0.0,<9",
     "sphinx-autodoc-typehints>=2.0.0,<3",
     "sphinx-rtd-theme>=3.0.1,<4",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ test = [
 docs = [
     "volatility3[dev]",
     "sphinx>=8.0.0,<9",
-    "sphinx-autodoc-typehints>=2.5.0,<3",
+    "sphinx-autodoc-typehints>=2.0.0,<3",
     "sphinx-rtd-theme>=3.0.1,<4",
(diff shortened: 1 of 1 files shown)


ikelos:
Sorry, realized I'd already asked on the other bug and said all this over the weekend without actually giving you time to check!  Just trying to get everything tidied up, ignore me (or ignore me here, but not on #1548)...  5;)

dgmcdona:
@ikelos I think all of your comments in #1548 made sense, and I was able to install the full set of dependencies now, so I think this is good to go 👍

PR #1560, opened 2025-01-18
Intermed: leverage iterable hashmaps instead of lists

### LINKED ISSUE #1552: Internal ISF APIs use lists instead of hashmaps
Hi,

I identified a few issues with the following core APIs: 

https://github.com/volatilityfoundation/volatility3/blob/a6777146f6fd789c10686b9ed1f8a5ea3521f9a5/volatility3/framework/symbols/intermed.py#L412-L425

- Description indicates that the return type is an iterator, wheras it is a list;
- Each dict keys are iterated over and converted to a list;
- Lookups are performed in `O(n)` space, which is really slow, especially when a target doesn't exist in the list.

List or even iterator conversion is not optimal, as the data source is already a json/dict object in the form of a hashmap. What this means, is that we should leverage this property to do lookups over a `KeysView` object in `O(1)` space: 

```python
@property
def symbols(self) -> KeysView[str]:
    """Returns a dict KeysView of the available symbol names."""
    return self._json_object.get("symbols", {}).keys()
```

`KeysView` objects can still be queried with the `in` operator, be applied `list()` or `len()` over. I did a benchmark with `viztracer`: 

- `list` conversion, current implementation, ~11ms (will vary depending on the target position in the list):

list_version

- `hashmap` implementation, ~50µs: 

keys_version

Even if it is a ~200x performance improvement, it might not make a user perceptible difference at this scale. Still, it directly speeds up each framework run while making memory management way more efficient. When applied to plugins relying indirectly on these APIs thousands of times per run, there is a direct benefit to observe. 

Now, I understand that this require careful planning, even if I did researches on where/how it might break things. Technically, even if the return type changes, the current use of these APIs does not explicitely relies on a `list` type, but simply an object to look `in`. 

I have a PR ready to fire (or not), but I think it is best to discuss about the relevance of such a change before, even if I believe that it is better on many points.

Abyss-W4tcher:
Hi, 

To perform iterable lookups in `O(1)` time instead of `O(n)` time, this PR makes use of the `.keys()` property of internal ISF APIs (symbols, types, enumerations). The required `Iterable` return type does not change, while providing a memory management and (framework/plugins) runtime improvement. 

For example, I observed a three seconds gain on a `linux.pstree` run (~13 seconds -> ~10 seconds). This is even more pronounced on heavier plugins.

I improved a few comments here and there too, and fixed a missing abstract class (NativeTableInterface.types).

Closes #1552

### Versioning

Patch bumps (**superseeded by minor bumps**): 
	- indicate the correct return type for Module.symbols (iterable, not list)
Minor bumps:
	- add missing NativeTableInterface.types abstract class

### DIFF
--- a/volatility3/framework/constants/_version.py
+++ b/volatility3/framework/constants/_version.py
@@ -1,6 +1,6 @@
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 2  # Number of releases of the library with a breaking change
-VERSION_MINOR = 17  # Number of changes that only add to the interface
+VERSION_MINOR = 18  # Number of changes that only add to the interface
 VERSION_PATCH = 0  # Number of changes that do not change the interface
 VERSION_SUFFIX = ""
 
--- a/volatility3/framework/constants/_version.py
+++ b/volatility3/framework/constants/_version.py
@@ -1,7 +1,7 @@
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 2  # Number of releases of the library with a breaking change
-VERSION_MINOR = 18  # Number of changes that only add to the interface
-VERSION_PATCH = 0  # Number of changes that do not change the interface
+VERSION_MINOR = 17  # Number of changes that only add to the interface
+VERSION_PATCH = 1  # Number of changes that do not change the interface
 VERSION_SUFFIX = ""
 
 PACKAGE_VERSION = (
--- a/volatility3/framework/symbols/intermed.py
+++ b/volatility3/framework/symbols/intermed.py
@@ -411,18 +411,23 @@ class Version1Format(ISFormatTable):
 
     @property
     def symbols(self) -> Iterable[str]:
-        """Returns an iterator of the symbol names."""
-        return list(self._json_object.get("symbols", {}))
+        """Returns an iterable (KeysView) of the available symbol names."""
+        return self._json_object.get("symbols", {}).keys()
 
     @property
-    def enumerations(self) -> Iterable[str]:
-        """Returns an iterator of the available enumerations."""
-        return list(self._json_object.get("enums", {}))
+    def enumerations(self) -> Iterable[Any]:
+        """Returns an iterable (KeysView) of the available enumerations."""
+        return self._json_object.get("enums", {}).keys()
 
     @property
-    def types(self) -> Iterable[str]:
-        """Returns an iterator of the symbol type names."""
-        return list(self._json_object.get("user_types", {})) + list(self.natives.types)
+    def types(self):
+        """Returns an iterable (KeysView) of the available symbol type names."""
+        # self.natives.types (set) is generally very small compared to user_types,
+        # so the dict conversion overhead can be neglected
+        return {
+            **self._json_object.get("user_types", {}),
+            **dict.fromkeys(self.natives.types),
+        }.keys()
 
     def get_type_class(self, name: str) -> Type[interfaces.objects.ObjectInterface]:
         return self._overrides.get(name, objects.AggregateType)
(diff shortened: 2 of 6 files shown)


volatility3/framework/symbols/intermed.py
```diff
@@ -411,18 +411,23 @@ def get_symbol(self, name: str) -> interfaces.symbols.SymbolInterface:
...
-        return list(self._json_object.get("enums", {}))
+    def enumerations(self) -> Iterable[Any]:
+        """Returns an iterable (KeysView) of the available enumerations."""
+        return self._json_object.get("enums", {}).keys()
 
     @property
-    def types(self) -> Iterable[str]:
-        """Returns an iterator of the symbol type names."""
-        return list(self._json_object.get("user_types", {})) + list(self.natives.types)
+    def types(self):
```
ikelos: This doesn't have a type, could we type this properly so it's obvious whether it's changd?

volatility3/framework/constants/_version.py
```diff
@@ -1,6 +1,6 @@
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 2  # Number of releases of the library with a breaking change
-VERSION_MINOR = 17  # Number of changes that only add to the interface
+VERSION_MINOR = 18  # Number of changes that only add to the interface
```
ikelos: I'm not sure this is required given it's an internal change of the API?  I think we can get away with just a PATCH bump?

volatility3/framework/symbols/intermed.py
```diff
@@ -411,18 +411,23 @@ def get_symbol(self, name: str) -> interfaces.symbols.SymbolInterface:
...
+        """Returns an iterable (KeysView) of the available symbol type names."""
+        # self.natives.types (set) is generally very small compared to user_types,
+        # so the dict conversion overhead can be neglected
+        return {
+            **self._json_object.get("user_types", {}),
```
ikelos, line 432: This *should* be fine, but rather than using `**`, would it not be faster to do `set(self._json_object.get("user_types", {}).keys()).union(self.natives.types)

Seems like calling `dict.fromkeys` shouldn't be necessary?

ikelos (commented):
Thanks, excited to get this in, but I'm not sure it requires a MINOR bump, more than just a PATCH bump.  Also, I think using the `**` magic isn't as good as just unioning sets...

volatility3/framework/symbols/intermed.py
Abyss-W4tcher: I actually tested this way, but converting the "user_types" dict to a set is less efficient as it still needs to iterate over all the keys to construct it. On the other hand, converting the (generally very small) "self.natives.types" set to a dict can be neglected.

ikelos: Ok, still not keen on the `**` syntax, but if it's more efficient could you throw a comment in there to say so, so that I don't come along one day and try to tidy it please...  5;)

volatility3/framework/constants/_version.py
Abyss-W4tcher: Alright, I wanted to be careful with the NativeTableInterface.types add but I will simply bump the PATCH version.

volatility3/framework/symbols/intermed.py
Abyss-W4tcher: Sure, in fact there is a clean `|` syntax for merging dicts but python3.8 doesn't support it. I'll add a FIXME about it too.

See https://blog.teclado.com/python-dictionary-merge-update-operators/

volatility3/framework/constants/_version.py
ikelos: Thanks, it would've been ok, but as I say I think in the end no signatures actually changed.

PR #1553, opened 2025-01-16
Linux - Parity release harden Page Cache

gcmoreira:
In the context of https://github.com/volatilityfoundation/volatility3/pull/1516 and related issues, this PR addresses several bugs in the Page Cache API code and its associated functions. 

### NOTE1: This requires the fixes from #1543 (linked list object extensions) and #1545 (mountinfo). Please DO NOT test the samples below using only the changes from this branch!

**NOTE2: From now on, the `linux.pagecache.InodePages` plugin will no longer display inode page details when using the `--dump` argument. This change helps prevent redundant scanning of the inode page cache, significantly improving processing time, especially when dealing with large files, such as those mentioned in the related issues.**

fix: #1526 ,  #1527

# Issue 1526

## Sample: broken_rhel_load_as_2.lime 
```shell
$ time ./vol.py -f ./broken_rhel_load_as_2.lime linux.pagecache.Files 2>/tmp/broken_rhel_load_as_2.err | wc -l
23920

real    1m19.729s
...
$ wc -l /tmp/broken_rhel_load_as_2.err 
0 /tmp/broken_rhel_load_as_2.err
```

## Sample: Ubuntu_Server_x64_22.04_Auth.zip.lime

```shell
$ time ./vol.py -f ./Ubuntu_Server_x64_22.04_Auth.lime linux.pagecache.Files 2>/tmp/Ubuntu_Server_x64_22.04_Auth.err | wc -l
82978

real    4m31.273s

$ wc -l /tmp/Ubuntu_Server_x64_22.04_Auth.err
0 /tmp/Ubuntu_Server_x64_22.04_Auth.err
```

# Issue 1527
## 1 - Page out of the bounds issue
Fixes the issue reported by @atcuno in #1527 related to handling page out-of-bounds scenarios.

## 2 - Sample: unshared_fds_network_connection.zip.lime

I'm not sure how the `/proc/kcore` file ended up being written in that case. It's likely due to all these recent changes, however, in my tests the inode for the `/proc/kcore` file in that sample shows zero cached pages, meaning that using `--dump` doesn't result in any file being written.
```shell
$ ./vol.py -f ../unshared_fds_network_connection.lime linux.pagecache.Files | grep 9ec482a94e30
SuperblockAddr  MountPoint      Device  InodeNum        InodeAddr       FileType        InodePages      CachedPages     FileMode        AccessTime      ModificationTime      ChangeTime       FilePath
...
[...]

PR #1561, opened 2025-01-19
Pre linux.pagecache.recoverfs support

Abyss-W4tcher:
Hi, 

To prepare for the new `linux.pagecache.recoverfs` plugin, I figured it would be better to split changes non directly related to it in a separate PR. 

A few readability tweaks were operated, as well as splitting the `inode` content extraction and writing process to allow for better flexibility. 
The `InodeSize` column was added to `Files` output, as I think it is a valuable piece of information. It will also complement very well `linux.pagecache.RecoverFs` `Recovered FileSize` column, as the following snippet demonstrates (scroll right to the last columns) :

```bash
$ python3 vol.py -f sample.bin -r pretty linux.pagecache.RecoverFs
  | SuperblockAddr |               MountPoint | Device |   InodeNum |      InodeAddr | FileType |  InodePages | CachedPages |   FileMode |                     AccessTime |               ModificationTime |                     ChangeTime |                                                                                                                                                                         FilePath |       InodeSize | Recovered FileSize
* | 0x89388c64c800 |                        / |    8:1 |      72981 | 0x89388cd19378 |      REG |          51 |          24 | -rw-r----- | 2024-11-15 12:40:53.900001 UTC | 2024-11-15 14:59:14.121585 UTC | 2024-11-15 14:59:14.121585 UTC |                                                                                                                                                                  /var/log/syslog |          207810 |             207810
* | 0x89388c64c800 |                        / |    8:1 |      72605 | 0x89388cd185b0 |      REG |           2 |           2 | -rw-rw-r-- | 2024-10-02 14:31:24.629028 UTC | 2024-11-15 14:58:43.880001 UTC | 2024-11-15 14:58:43.880001 UTC |                                                                                                                                                                    /var/log/wtmp |            6144 |               6144
* | 0x89388c64c800 |                        / |    8:1 |      72606 | 0x89388cd18118 |      REG |           0 |           0 | -rw-rw---- | 2024-10-02 14:31:24.629028 UTC | 2024-10-02 14:33:48.019670 UTC | 2024-10-02 14:36:11.362212 UTC |                                                                                                                                                                    /var/log/btmp |               0 |                  0
* | 0x89388c64c800 |                        / |    8:1 |      80281 | 0x89388cdeeac0 |      DIR |           1 |           0 | drwx------ | 2024-11-15 12:40:45.328000 UTC | 2024-11-15 12:40:45.328000 UTC | 2024-11-15 12:40:45.328000 UTC |                                                                                                                                                                 /var/log/private |            4096 |                N/A
* | 0x89388c64c800 |                        / |    8:1 |      80249 | 0x89388cd04a98 |      DIR |           1 |           0 | drwxr-sr-x | 2024-11-15 12:40:58.716001 UTC | 2024-11-15 12:40:44.876000 UTC | 2024-11-15 12:40:44.876000 UTC |                                                                                                                                                                 /var/log/journal |            4096 |                N/A
* | 0x89388c64c800 |                        / |    8:1 |      80275 | 0x89388cd033a0 |      DIR |           1 |           0 | drwxr-sr-x | 2024-11-15 12:40:58.716001 UTC | 2024-11-15 12:40:55.952001 UTC | 2024-11-15 12:40:55.952001 UTC |                                                                                                                                /var/log/journal/c602156d161745d29b98f977eb66a96d |            4096 |                N/A
* | 0x89388c64c800 |                        / |    8:1 |      72930 | 0x89388cd073f0 |      REG |        2048 |         292 | -rw-r----- | 2024-11-15 14:58:37.228001 UTC | 2024-11-15 14:58:42.764001 UTC | 2024-11-15 14:58:42.764001 UTC |                                                                                                              /var/log/journal/c602156d161745d29b98f977eb66a96d/user-1000.journal |         8388608 |            1196032
* | 0x89388c64c800 |                        / |    8:1 |      72437 | 0x89388cd00118 |      REG |        2048 |         732 | -rw-r----- | 2024-11-15 14:58:37.228001 UTC | 2024-11-15 14:59:14.401585 UTC | 2024-11-15 14:59:14.401585 UTC |                                                                                                                 /var/log/journal/c602156d161745d29b98f977eb66a96d/system.journal |         8388608 |            2998272
```

As you can see, this will provide easy lookups to compare the recovered file size to the announced size. It also arguments the choice of placing `InodeSize` at the very end. Happy to discuss it however !

#### linux.pagecache.Files

- Minor bump:
	- get_inodes: add "follow_symlinks" argument
	- rendering: add "InodeSize" column

#### linux.pagecache.InodePages

- Minor bump:
	- add "write_inode_content_to_stream" method

### DIFF
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -473,7 +473,7 @@ class InodePages(plugins.PluginInterface):
             context: The context on which to operate
             layer_name: The name of the layer on which to operate
             inode: The inode to dump
-            stream: An IO steam to write to, typically FileHandlerInterface or BytesIO
+            stream: An IO stream to write to, typically FileHandlerInterface or BytesIO
         """
         layer = context.layers[layer_name]
         # By using truncate/seek, provided the filesystem supports it, and the
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -8,6 +8,7 @@ import datetime
 from dataclasses import dataclass, astuple
 from typing import List, Set, Type, Iterable
 
+from volatility3.framework.constants import architectures
 from volatility3.framework import renderers, interfaces
 from volatility3.framework.renderers import format_hints
 from volatility3.framework.interfaces import plugins
@@ -112,7 +113,7 @@ class Files(plugins.PluginInterface, timeliner.TimeLinerInterface):
             requirements.ModuleRequirement(
                 name="kernel",
                 description="Linux kernel",
-                architectures=["Intel32", "Intel64"],
+                architectures=architectures.LINUX_ARCHS,
             ),
             requirements.PluginRequirement(
                 name="mountinfo", plugin=mountinfo.MountInfo, version=(1, 2, 0)
@@ -397,7 +398,7 @@ class InodePages(plugins.PluginInterface):
             requirements.ModuleRequirement(
                 name="kernel",
                 description="Linux kernel",
-                architectures=["Intel32", "Intel64"],
+                architectures=architectures.LINUX_ARCHS,
             ),
             requirements.PluginRequirement(
                 name="files", plugin=Files, version=(1, 0, 0)
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -112,7 +112,7 @@ class Files(plugins.PluginInterface, timeliner.TimeLinerInterface):
 
     _required_framework_version = (2, 0, 0)
 
-    _version = (1, 2, 0)
+    _version = (1, 1, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -38,6 +38,11 @@ class InodeUser:
     modification_time: str
     change_time: str
     path: str
+    inode_size: int
+
+    @staticmethod
+    def format_symlink(symlink_source: str, symlink_dest: str):
+        return f"{symlink_source} -> {symlink_dest}"
 
 
 @dataclass
@@ -81,6 +86,7 @@ class InodeInternal:
         access_time_dt = self.inode.get_access_time()
         modification_time_dt = self.inode.get_modification_time()
         change_time_dt = self.inode.get_change_time()
+        inode_size = int(self.inode.i_size)
 
         inode_user = InodeUser(
             superblock_addr=superblock_addr,
@@ -96,6 +102,7 @@ class InodeInternal:
             modification_time=modification_time_dt,
             change_time=change_time_dt,
             path=self.path,
+            inode_size=inode_size,
         )
         return inode_user
 
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -220,12 +220,14 @@ class Files(plugins.PluginInterface, timeliner.TimeLinerInterface):
         cls,
         context: interfaces.context.ContextInterface,
         vmlinux_module_name: str,
+        follow_symlinks: bool = True,
     ) -> Iterable[InodeInternal]:
         """Retrieves the inodes from the superblocks
 
         Args:
             context: The context that the plugin will operate within
             vmlinux_module_name: The name of the kernel module on which to operate
+            follow_symlinks: Whether to follow symlinks or not
 
         Yields:
             An InodeInternal object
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -80,6 +86,7 @@ def to_user(
         access_time_dt = self.inode.get_access_time()
         modification_time_dt = self.inode.get_modification_time()
         change_time_dt = self.inode.get_change_time()
+        inode_size = int(self.inode.i_size)
```
ikelos, line 89: Why is this being cast to an `int`?  Is it not already an `int`?  Please check, I'm really try to avoid people including unnecessary casts because they don't expect something to be an int, because then other people see it and they think they need to cast to an int, then some people do it out of caution, and before you know it the whole codebase is bursting full of pointless casts...

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -439,32 +452,48 @@ def write_inode_content_to_file(
...
-                    f.seek(current_fp)
-                    f.write(page_bytes)
-
+                InodePages.write_inode_content_to_stream(inode, f, vmlinux_layer)
         except OSError as e:
             vollog.error("Unable to write to file (%s): %s", filename, e)
 
+    @staticmethod
+    def write_inode_content_to_stream(
+        inode: interfaces.objects.ObjectInterface,
```
ikelos, line 467: This is starting to make use of more complex data types (not too bad) and break away from the standard format for methods (first param `context`).  I'd really prefer this take a context and a layer_name rather than just a layer.

I don't have very good reason for sticking with convention, originally it was because everything you'd do needed a context (and if you wanted to extend this in the future, you'll likely want the context there to do it).  It was also the thought in the back of my mind of one day serializing such calls for some reason, although I no longer remember what (perhaps parallelization).  Regardless, I think it'd be a good convention to maintain unless you have strong objections...

ikelos (commented):
Thanks, generally looks good, just the broken-out function doesn't stick to the general convention we've got going for plugin API methods and I'm trying to stick to it for as long as possible...

Also, there's an `int` cast that I think is unnecessary...  5;)

volatility3/framework/plugins/linux/pagecache.py
Abyss-W4tcher: This was explained by gcmoreira, in the comment right before the added line:

> Ensure all types are atomic immutable. Otherwise, astuple() will take a long time doing a deepcopy of the Volatility objects.

I was also able to verify that not casting it to `int`, like the other properties, results in nested deepcopies.

volatility3/framework/plugins/linux/pagecache.py
Abyss-W4tcher: I basically followed the  `write_inode_content_to_file` design, as it was the accepted way of doing it.

I also prefer to pass a context and a layer name, happy to revert it that way !

ikelos: Yes please, I try not to be too nitpicky but it's hard keeping everything in order as it comes in so quickly.  So it's be MINOR bump if we're just adding `write_inode_content_to_stream` and MAJOR bump if we're changing the signature of `write_inode_content_to_file`.

Abyss-W4tcher: I proceeded to update `write_inode_content_to_file` as well.

ikelos:
Ok, looks good, thanks!

gcmoreira:
Hey @ikelos, the changes in this PR are good. However, the timing is challenging, as @atcuno and I are currently focused on stabilizing several core APIs: [Page Cache](https://github.com/volatilityfoundation/volatility3/pull/1553), [Mountinfo](https://github.com/volatilityfoundation/volatility3/pull/1545), VMA, tasks, etc with related PRs still under review. 

Could we consider temporarily reverting this PR and revisiting it once the pending parity release  PRs are merged and @atcuno confirms that the changes have passed the full regression tests? Additionally, I would greatly appreciate it, and it seems prudent, to delay the merging of any other pending or future PRs that impact this code until the parity release PRs have been finalized. Thanks

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -37,6 +38,11 @@ class InodeUser:
     modification_time: str
     change_time: str
     path: str
+    inode_size: int
+
+    @staticmethod
+    def format_symlink(symlink_source: str, symlink_dest: str) -> str:
+        return f"{symlink_source} -> {symlink_dest}"
```
gcmoreira, line 45: Not too fan of this `format_symlink` staticmethod. I'm not sure about the benefit of including a method in the inode data class that simply formats two external strings to 's1 -> s2'?
I think it would make more sense, and it's my mistake, but `_follow_symlink`  should ideally be a public method in the inode object extension

Abyss-W4tcher: A future implementation will require to format symlinks given an inode as well. Unifying it to a function allows to prevent code duplication, even if it might seem futile for one line indeed.

I put it under InodeUser as it is in fact only needed to make it more user readable after rendering 👍.

PR #1562, opened 2025-01-19
Revert "Add missing exception handling in env var recovery. Prevent backtraces"

ikelos:
Reverts volatilityfoundation/volatility3#1513


gcmoreira (approved):
Thanks, it looks good.

PR #1564, opened 2025-01-21
New Linux plugin: linux.tracing.ftrace

Abyss-W4tcher:
Hi :),

This new plugin adds detection of malicious kernel functions hooking (see #1286), by inspecting the `Ftrace` (**F**unction **T**racer) infrastructure. Attackers can leverage its capabilities to hook critical low level calls, and perform man-in-the middle or output manipulation. 

It also relies on the recent `linux.hidden_modules` and `linux.modxview` to extend the uncovering of callbacks parents/modules.  

Documentation:

- https://i.blackhat.com/USA21/Wednesday-Handouts/us-21-Fixing-A-Memory-Forensics-Blind-Spot-Linux-Kernel-Tracing-wp.pdf
- https://xcellerator.github.io/posts/linux_rootkits_02/

Superseeds https://github.com/Abyss-W4tcher/volatility-scripts/blob/master/Volatility_contest_2023/plugins/check_ftrace.py

Hopefully, this does not interfere with the parity release ?

### Versioning

Minor framework bumps:
	- add framework.symbols.linux.utilities.modules.Modules API


volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -0,0 +1,289 @@
...
+    FTRACE_OPS_FL_PERMANENT = auto()
+    FTRACE_OPS_FL_DIRECT = auto()
+    FTRACE_OPS_FL_SUBOP = auto()
+
+
+class Check_ftrace(interfaces.plugins.PluginInterface):
+    """Detect ftrace hooking"""
+
+    _version = (1, 0, 0)
+    _required_framework_version = (2, 17, 0)
```
ikelos: This is probably `(2, 19, 0)`, no?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -0,0 +1,41 @@
+from volatility3 import framework
+from volatility3.framework import interfaces
+from volatility3.framework.symbols.linux import extensions, LinuxUtilities
+from typing import Iterable, Optional
+
+
+class Modules(interfaces.configuration.VersionableInterface):
```
ikelos, line 10: Is there any other functionality you can move over to here (with deprecation warnings), given it would be best to keep things together as we move them?  Just feels weird have a whole module for a single helper function...

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -0,0 +1,289 @@
...
+                description="Show ftrace flags associated with an ftrace_ops struct",
+                optional=True,
+                default=False,
+            ),
+        ]
+
+    @classmethod
+    def _set_hidden_modules_run(cls) -> None:
+        """Use a self-contained setter, to prevent running hidden_modules multiple times."""
+        cls._hidden_modules_run = True
```
ikelos: Errr, this will affect *all* instances of this plugin, which probably isn't what you want to do.  I feel there'll be a better way of achieving the outcome you're after...

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -0,0 +1,289 @@
...
+        """
+        kernel = context.modules[kernel_name]
+        callback = ftrace_ops.func
+
+        # Try to lookup within the known modules if the callback address fits
+        module = modules_utilities.Modules.module_lookup_by_address(
+            context, kernel.layer_name, known_modules, callback
+        )
+        # Run hidden_modules plugin if a callback origin couldn't be determined (only done once, results are re-used afterwards)
+        if module is None and not cls._hidden_modules_run:
```
ikelos: So, this would be better returned from the function and optionally handed in, so it's in the hands of the caller whether to rerun it (hand in `None`) or use some existing results (which they could then do from the output of this method).

volatility3/framework/plugins/linux/tracing/ftrace.py
Abyss-W4tcher: Yes, thanks 👍

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -0,0 +1,289 @@
...
+        """Parse an ftrace_ops struct to highlight ftrace kernel hooking.
+        Iterates over embedded ftrace_func_entry entries, which point to hooked memory areas.
+
+        Args:
+            known_modules: A set of known modules to iterate over, used to locate callbacks origin
+            ftrace_ops: The ftrace_ops struct to parse
+            parse_flags: Whether to parse ftrace_ops flags or not
+
+        Yields:
+            A tuple containing a selection of useful fields (callback, hook, module) related to an ftrace_func_entry struct
```
ikelos: Would it be better to return a `NamedTuple` or a `Dataclass` here?

ikelos (commented):
Looking pretty reasonable.  Just a couple possible improvements and if we're adding in a new modules utility, it might be good to start shifting associated functionality out of `LinuxUtilities` if possible.  I haven't figured out how to move stuff out of the module class, so that still needs some thinking about, but maybe just duplicating functionality in a versionable module, and then a deprecation warning in the module method?  That can be for another time...

volatility3/framework/symbols/linux/utilities/modules.py
Abyss-W4tcher: Sure, I was planning on it. I think it is best to split it in a dedicated PR, it will be easier to trace back.

volatility3/framework/plugins/linux/tracing/ftrace.py
Abyss-W4tcher: I see, I noticed that in late plugins (fbdev, pagecache), we were sliding to this design. I'll extract the needed fields on the generator side 👍.

volatility3/framework/plugins/linux/tracing/ftrace.py
Abyss-W4tcher: This should only affect the `my_plugin = ftrace.Check_ftrace()` instance launched by automagic ? But sure, I can switch to this design as well.

volatility3/framework/symbols/linux/utilities/modules.py
Abyss-W4tcher: I'll leave the existing LinuxUtilites, by making them sole proxies and adding a depreciation warning. Then, I'll make another PR to update the existing calls across the framework, which might take a bit more time.

volatility3/framework/plugins/linux/tracing/ftrace.py
Abyss-W4tcher: Done.

volatility3/framework/plugins/linux/tracing/ftrace.py
Abyss-W4tcher: Done, I leveraged the keys returned by modxview ("lsmod", "check_modules", "hidden_modules") associated to each plugin run result. This way, we can know if `hidden_modules` was run depending if the key is present in the results.

Abyss-W4tcher:
I made all the changes in a single commit, as it should be easier for you to review the requested changes. There were a few tweaks too, like an `Enum` property that works well in 3.12 but errors silently (renders to None) on earlier versions :). 

I'll definitely look into writing unit tests for the modules plugins once the infrastructure is up, to prevent missing these tricky errors in the future.

ikelos:
Thanks for making all the requested changes.  I'm still a little concerned about the Module standing out with just that one function.  I know you said you'd include them in a different PR, but I'd almost like to get that PR in first so I can see what it'll look like and if it's the best way of doing it.  I think the class name and the module name should probably match, because it'll be a real pain to include different classes in the same file (the file will import, but an exception will get thrown when the non-existant class is accessed as part of the requirement, so if anything it'll break less cleanly).

I'm also not sure the main project version bump is needed because plugins that can't import the file will just fail during the import phase, and they'll do that before they can check the main project version number anyway, so depending on the higher version number doesn't gain anything.

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -0,0 +1,310 @@
...
+                constants.LOGLEVEL_VV,
+                f"ftrace_func_entry list of ftrace_ops@{ftrace_ops.vol.offset:#x} is empty/invalid. Skipping it...",
+            )
+            return []
+
+        while current_bucket_ptr.is_readable():
+            yield current_bucket_ptr.dereference().cast("ftrace_func_entry")
+            current_bucket_ptr = current_bucket_ptr.next
+
+        return None
```
atcuno, line 130: this isn't needed in a generator and seems out of place?

Abyss-W4tcher: codeql complained about it, because of the first explicit return. This fixed it.

Abyss-W4tcher:
This is ready for a review after the merge of #1567.

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -0,0 +1,311 @@
...
+vollog = logging.getLogger(__name__)
+
+
+# https://docs.python.org/3.13/library/enum.html#enum.IntFlag
+class FtraceOpsFlags(IntFlag):
+    """Denote the state of an ftrace_ops struct.
+    Based on https://elixir.bootlin.com/linux/v6.13-rc3/source/include/linux/ftrace.h#L255.
+    """
+
+    FTRACE_OPS_FL_ENABLED = auto()
```
ikelos: Would this not be better as a volatility enum, then the values could be stored in JSON as data rather than in code?  It would mean adding an Ftrace symbol table, but then it should just be possible to cast ftrace_ops.flags as the enum?  That might also then get around the python <= 3.10 issue you were seeing?

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -0,0 +1,311 @@
...
+    module_address: int
+    flags: str
+
+
+class CheckFtrace(interfaces.plugins.PluginInterface):
+    """Detect ftrace hooking"""
+
+    _version = (1, 0, 0)
+    _required_framework_version = (2, 19, 0)
+    additional_description = """Investigate the ftrace infrastructure to uncover kernel attached callbacks, which can be leveraged
```
ikelos: Shouldn't this be part of the docstring?  So something like:

```
"""Detect ftrace hooking

Investigate the ftrace infrastructure to uncover kernel attached callbacks...
"""
```

ikelos: Just want to get it right before we start doing it everywhere, and I figure the little first line/rest formatting should be easy to turn into appropriate rules for the CLI help?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -1,18 +1,45 @@
...
+        Args:
+            context: The context on which to operate
+            layer_name: The name of the layer on which to operate
+            modules: An iterable containing the modules to match the address against
+            target_address: The address to check for a match
+        """
+
+        for module in modules:
+            _, start, end = cls.mask_mods_list(context, layer_name, [module])[0]
+            if start <= target_address <= end:
```
ikelos: I assume modules can't overlap, so returning the first one is fine?

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -0,0 +1,311 @@
...
+            )
+
+        # Iterate over ftrace_func_entry list
+        for ftrace_func_entry in cls.extract_hash_table_filters(ftrace_ops):
+            hook_address = ftrace_func_entry.ip.cast("pointer")
+
+            # Determine the symbols associated with a hook
+            hooked_symbols = kernel.get_symbols_by_absolute_location(hook_address)
+            hooked_symbols = ",".join(
+                [s.split(constants.BANG)[-1] for s in hooked_symbols]
```
ikelos: Any chance of a slightly longer name?  Even `hs` or ideally `hooked_symbol` would be better?

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -0,0 +1,311 @@
...
+        if (
+            module is None
+            and run_hidden_modules
+            and "hidden_modules" not in known_modules
+        ):
+            vollog.info(
+                "A callback module origin could not be determined. hidden_modules plugin will be run to detect additional modules.",
+            )
+            known_modules_addresses = set(
+                context.layers[kernel.layer_name].canonicalize(module.vol.offset)
```
ikelos: This will keep looking up the layer to get to the same function (which might even be defined on the class?).  The layer should be put in a variable much earlier on...

ikelos (commented):
Looks good, a small optimization, and also I'd like to consider whether using a volatility enumeration would be better/easier?  If there isn't support for multiple flags (I don't remember) it's the kind of thing we can add because it would be useful, but given we've got a big and powerful parser that integrates with everything else, I'm keen to make use of it rather than rolling our own in python...

volatility3/framework/plugins/linux/tracing/ftrace.py
Abyss-W4tcher: I'm not against the idea, both should work well. However `IntEnum` provides really useful APIs out of the box, and I don't have to do a manual list comprehension and mask the value. The little string manipulation I did was to prettify the output because we already use "|" as a column separator.

I think JSON is more interesting when we have a lot of constants or nested structs to declare.

volatility3/framework/plugins/linux/tracing/ftrace.py
Abyss-W4tcher: I didn't want to clutter `python3 vol.py -h` even more, and make it harder to navigate/scroll, especially if we do this "long" description for every plugin ? Both ways have pros and cons, happy with either of them. :)

volatility3/framework/symbols/linux/utilities/modules.py
Abyss-W4tcher: Modules memory space allocated in kernel space should never overlap indeed.

Abyss-W4tcher: I added an extra check for this scenario,  and it would most likely be the result of a false positive in the modules scanning capabilities. Extremely unlikely, but at least we cover it ? I chose `warnings.warn` as it will be printed only once.

volatility3/framework/plugins/linux/tracing/ftrace.py
ikelos: The idea wouldn't be to make the `--help` longer, it would be to split the docstring and take only the first line for `--help` and then use the rest as the `additional_description`?  It reduces the clutter of another variable while (hopefully) having little impact on the existing output...

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: I'd probably have made it a `vollog.debug` since I don't know that the user will be that interested in it, but it should be pretty exceptional, so however you want to handle it should be fine...

volatility3/framework/plugins/linux/tracing/ftrace.py
ikelos: Well, I'd like to improve our enum support to the point it becomes as easy to use as IntEnum?  If you could tell me what APIs it's missing, we could add them and give people more reason to use it...

volatility3/framework/plugins/linux/tracing/ftrace.py
Abyss-W4tcher: If I understood well, you would like to modify the current CLI help handler to split `__doc__` at "\n" and show only the first line in `python3 vol.py -h` ?

Then, basically, `additional_description` would be the lines following that ?

ikelos: Yep, that's exactly what I was thinking?  Do you reckon it'll work, I didn't think it all the way through but I couldn't immediately see any issues with it?

Abyss-W4tcher: I will need to update the current (small) descriptions that embed newlines to make them fit on one. Also, this makes the `additional_description` variable deprecated quite early. No big deal, should I add a deprecation message under it, that points to where we prefer devs to put the additional help ?

ikelos: Well, let's try it first and see what it's like, but no, it hasn't been included in on stable yet, and it's very recent so I think we can just silently ditch it and class it as part of development.

volatility3/framework/plugins/linux/tracing/ftrace.py
Abyss-W4tcher: The use of an `IntEnum` does not require to specifically make a JSON dict with hardcoded powers of 2, and I think the readability offered by `auto()` is quite nice.

---

Nevertheless, I still looked on how we could add an easy int masking API (or leverage the one in https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/symbols/wrappers.py) to enums.

I noticed that objects returned by `my_symbol_space.get_enumeration("XX")` aren't of type `Enumeration`, but `volatility3.framework.objects.templates.ObjectTemplate`

I think I was able to understand the calling tree:

- `get_enumeration` instantiates an `ObjectTemplate`:
https://github.com/volatilityfoundation/volatility3/blob/ad90804af04c6c1aead311c08fc24006a969bf51/volatility3/framework/symbols/intermed.py#L543-L548

- No inherited properties from `Enumeration`, but its `VolTemplateProxy`:

https://github.com/volatilityfoundation/volatility3/blob/ad90804af04c6c1aead311c08fc24006a969bf51/volatility3/framework/objects/templates.py#L25-L45

https://github.com/volatilityfoundation/volatility3/blob/ad90804af04c6c1aead311c08fc24006a969bf51/volatility3/framework/objects/__init__.py#L640-L642

```python
>>> dir(kernel.get_enumeration("XX"))
['__call__', '__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattr__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_vol', 'child_template', 'children', 'clone', 'has_member', 'lookup', 'relative_child_offset', 'replace_child', 'size', 'update_vol', 'vol']
```

No `description`, `is_valid_choice` etc. available, which also makes it difficult to add new methods. As opposed to a symbol for example, which value can change depending on the memory sample, isn't the enum a simple static definition from the ISF ? Why is the constructed enum limited ?

I may have missed something obvious, but right now I am not sure if this is intended or if there is another way ?

ikelos: So `get_enumeration` would be the equivalent of `get_type`.  If you wanted to apply it to some data, you'd use `context.Object` to create, just as you would a type.  Then you'll get a bunch of functions that should let you interrogate it.  If there's mirror methods you think we can expose on the template because they don't require the data, I'm happy to add those in too.  The point is, the enumeration type won't improve unless we have cases where we try to use it, so even if we do end up going with a normal IntEnum, I'd like to improve the enumeration API so that it's similarly easy to use to do the things we want to be able to do with it.

`Auto()` actually scares me, because an accidental addition (or a change of the upstream code) and all the autos fall out of place.  That's one of the reasons I'd prefer having the actual data encoded into it, because then you can see exactly what you're getting, and other people can use the data themselves. The fact that it numbers them for you helps for precisely one time, so not as important as the thousands of time the enum is going to be used by everybody else...

ikelos: It looks as though I have mirrored lookup, which I think allows you to pass in a value and get the identifier for it back...

Abyss-W4tcher: `Enumeration` objects are designed to operate in a `context`, technically we could do `ftrace_ops.flags.cast("ftrace_symbol_table!ftrace_flags")`. Still, for this use case I'm not sure it is worth the overhead of a custom ISF loading.

I think we'll have to do live enumeration parsing for later `tainting` plugins, I'll work on a way to parse the flags more intelligibly, if that is ok for you ? It should be easier to design APIs with a concrete use case.

---

Happy to change `auto()` values to explicit `1 << n` values, thanks for raising this concern !

ikelos: Yeah, fair enough.  I am interested in getting our enum implementation up to the same ease-of-use as the built-in ones (or better!), but it might be a bit more overhead for this example now.

I would prefer the `auto()` values be changed to explicit values though, thanks.

Abyss-W4tcher:
Hopefully, all issues have been addressed now. 👍

Abyss-W4tcher:
Hello, I think it is ready to go in ? No hurries of course if this is not on your schedule.

ikelos:
Can you prod me again next week please?  I think we've done all the jiggery-pokery we'd intended, but I just want to get to next week to let things settle if that's ok...

Abyss-W4tcher:
Hi @ikelos, would you have time to give a final look over this plugin this week ? I was planning on pushing the additional `tracepoints` plugin soon in an effort to close #1286.

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -0,0 +1,316 @@
...
+                constants.LOGLEVEL_VV,
+                f"ftrace_func_entry list of ftrace_ops@{ftrace_ops.vol.offset:#x} is empty/invalid. Skipping it...",
+            )
+            return []
+
+        while current_bucket_ptr.is_readable():
+            yield current_bucket_ptr.dereference().cast("ftrace_func_entry")
+            current_bucket_ptr = current_bucket_ptr.next
+
+        return None
```
ikelos, line 130: This looks a little weird to me, but seems fine (and I guess it's to deal with black/ruff complaints).

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -0,0 +1,316 @@
...
+                callback_symbol,
+                callback,
+                hooked_symbols,
+                module_name,
+                module_address,
+                # FtraceOpsFlags(ftrace_ops.flags).name is valid in > Python3.10, but
+                # returns None <= Python 3.10. We need to manipulate it like so to ensure compatibility:
+                # FtraceOpsFlags.FTRACE_OPS_FL_IPMODIFY|FTRACE_OPS_FL_ALLOC_TRAMP
+                # -> FTRACE_OPS_FL_IPMODIFY,FTRACE_OPS_FL_ALLOC_TRAMP
+                str(FtraceOpsFlags(ftrace_ops.flags)).split(".")[-1].replace("|", ","),
```
ikelos: This... seems nasty.  Any chance you could do it in more reable pieces before the yield statement please?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -1,18 +1,63 @@
...
+
+        Returns:
+            The first memory module in which the address fits
+        """
+        matches = []
+        seen_addresses = set()
+        for module in modules:
+            _, start, end = cls.mask_mods_list(context, layer_name, [module])[0]
+            if (
+                start <= target_address <= end
```
ikelos: Should this be < end rather than <= end?  Just want to check that end includes the last byte of the thing, rather than being the marker for the last byte...

ikelos (commented):
Ok, generally looks pretty good.  It could go in as is, but there's a just a couple bits I wanted to check/clean-up if that's ok.  Should be good to go in after that!  5:)

volatility3/framework/plugins/linux/tracing/ftrace.py
Abyss-W4tcher: It is indeed to deal with codeql complaints.

volatility3/framework/symbols/linux/utilities/modules.py
Abyss-W4tcher: It is indeed non inclusive, as specified here: https://elixir.bootlin.com/linux/v6.4.13/source/include/linux/module.h#L613

I added the relevant kernel functions in the docstring 👍

Abyss-W4tcher:
I think linking the issue now is a bit early, it will be closed when `linux.tracepoints` will be merged 👍.

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -0,0 +1,318 @@
...
+            hooked_symbols = ",".join(
+                [
+                    hooked_symbol.split(constants.BANG)[-1]
+                    for hooked_symbol in hooked_symbols
+                ]
+            )
+            # Manipulate FtraceOpsFlags(ftrace_ops.flags) like so:
+            # "FtraceOpsFlags.FTRACE_OPS_FL_IPMODIFY|FTRACE_OPS_FL_ALLOC_TRAMP"
+            # -> "FTRACE_OPS_FL_IPMODIFY,FTRACE_OPS_FL_ALLOC_TRAMP"
+            formatted_ftrace_flags = (
```
ikelos: That's just kind of... moving the issue, but not actually clarifying what's going on.  It *looks* like the following might work and should be alot easier to understand?

```
','.join([flag.name for flag in <NameOfEnum> if flag in <ValueToCheck>])
```
Also, after 3.11, you should just be able to do :
```
','.join([flag.name for flag in <ValueToCheck>])
```

It's not the fastest of operations, but it's a *lot* easier to read on sight than all the splits, indexes and replacements.  Tested and works in python 3.8 and 3.12.

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -0,0 +1,318 @@
...
+                str(FtraceOpsFlags(ftrace_ops.flags)).split(".")[-1].replace("|", ",")
```
ikelos: I *think* the following should work...

```suggestion
                # FIXME: Once python 3.11 is the minimum, this can become ','.join([flag.name for flag in FtraceOpsFlags(ftrace_ops.flags)])
                ','.join([flag.name for flag in FtraceOpsFlags if flag in FtraceOpsFlags(ftrace_ops.flags)])
```

volatility3/framework/plugins/linux/tracing/ftrace.py
Abyss-W4tcher: Both ways look good to me, we loose the `IntEnum` benefits but it is a bit more understandable in the end. I updated the code.

volatility3/framework/plugins/linux/tracing/ftrace.py
Abyss-W4tcher: Providing a value is only possible for `IntFlag`, I made a similar change that achieves what you had in mind.

ikelos: I was pretty sure I ran my tests with `IntFlag` but as long as you're happy, I'm happy.  Thanks for fixing it up!  5:)

Abyss-W4tcher: Ah yes, it works as well ! I accidentally didn't switch back to `IntFlag` to try your code after making the last commit...

PR #1566, opened 2025-01-22
Add low stub offset kernel detection

Danking555:
### Summary
This PR introduces a Low-Stub based method as an additional approach to kernel offset discovery in Volatility.
When available, this method resolves the kernel offset in ~1 second, a significant improvement compared to the ~15 seconds required by the current KDBG scan for large memory dumps (e.g., 32GB).
If the Low Stub method fails (e.g., for Hyper-V .vmem files), Volatility gracefully falls back to the existing KDBG scan.

### Motivation
The KDBG scan is computationally intensive, particularly for large memory dumps, and can delay forensic investigations and incident response.
By leveraging the Low Stub, which contains the DTB and kernel base address hint, the kernel offset can be determined far more efficiently, improving overall runtime for all plugins dependent on this step.

### Testing
This hybrid approach was tested on:

- Physical memory dumps from Windows 11
- Edge cases where the Low Stub is unavailable, with successful fallback to the KDBG scan (e.g., VMs memory)

### References

- Memprocfs: Inspired the Low Stub based approach to kernel offset discovery.
- Alex Ionescu: [Recon 2017 Talk on the Low Stub][0] (starting at 43:36).

[0]: <https://www.youtube.com/watch?v=_ShCSth6dWM>

Special thanks to Ulf Frisk and Alex Ionescu for their pioneering research, which informed and inspired this contribution.

### Acknowledgment

Thank you for maintaining Volatility as an indispensable tool in the forensic community. I hope this contribution enhances its usability and performance. I welcome any feedback or suggestions for further improvement.

### DIFF
--- a/volatility3/framework/automagic/pdbscan.py
+++ b/volatility3/framework/automagic/pdbscan.py
@@ -11,6 +11,7 @@ import contextlib
 import logging
 import math
 import os
+import struct
 from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, Union
 
 from volatility3.framework import constants, exceptions, interfaces, layers
@@ -376,8 +377,43 @@ class KernelPDBScanner(interfaces.automagic.AutomagicInterface):
                     valid_kernel = (virtual_layer_name, address, res[0])
         return valid_kernel
 
+    def method_low_stub_offset(self,
+        context: interfaces.context.ContextInterface,
+        vlayer: layers.intel.Intel,
+        progress_callback: constants.ProgressCallback = None,
+    ) -> Optional[ValidKernelType]:
+        kernel_hint = 0
+        kernel_base = 0
+        physical_layer = context.layers.get('memory_layer')
+
+        # try locating kernel base via x64 Low Stub in lower 1MB starting from second page (4KB)
+        # if "Discard Low Memory" setting is disabled in BIOS, the Low Stub may be at the third/fourth or further pages
+        for offset in range(0x1000,0x100000, 0x1000):
+            if 0xffffffffffff00ff & int.from_bytes(physical_layer.read(offset, 0x8), "little") != 0x00000001000600E9:
+                continue  # not _PROCESSOR_START_BLOCK->Jmp
+            potential_kernel_hint = int.from_bytes(physical_layer.read(offset + 0x70, 0x8), "little")
+            if (0xfffff80000000003 & potential_kernel_hint) != 0xfffff80000000000:
+                continue  # not _PROCESSOR_START_BLOCK->LmTarget
+            kernel_hint = potential_kernel_hint & 0xffffffffffff
+            kernel_base = kernel_hint & (~0x1fffff) & 0xffffffffffff
+            break
+
+        if kernel_base:
+            # Scanning 32mb in 2mb chunks for the 'ntoskrnl' base address
+            while (kernel_base + 0x2000000) > kernel_hint:
+                for i in range(0, 0x200000, 0x1000):
+                    valid_kernel = self.check_kernel_offset(
+                        context, vlayer, kernel_base, progress_callback
+                    )
+                    if valid_kernel:
+                        return valid_kernel
+                kernel_base -= 0x200000
+
+        return None
+
     # List of methods to be run, in order, to determine the valid kernels
     methods = [
+        method_low_stub_offset,
         method_kdbg_offset,
         method_module_offset,
         method_fixed_mapping,
--- a/volatility3/framework/automagic/pdbscan.py
+++ b/volatility3/framework/automagic/pdbscan.py
@@ -11,7 +11,6 @@ import contextlib
 import logging
 import math
 import os
-import struct
 from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


volatility3/framework/automagic/pdbscan.py
```diff
@@ -376,8 +377,43 @@ def check_kernel_offset(
...
+        kernel_base = 0
+        physical_layer = context.layers.get('memory_layer')
+
+        # try locating kernel base via x64 Low Stub in lower 1MB starting from second page (4KB)
+        # if "Discard Low Memory" setting is disabled in BIOS, the Low Stub may be at the third/fourth or further pages
+        for offset in range(0x1000,0x100000, 0x1000):
+            if 0xffffffffffff00ff & int.from_bytes(physical_layer.read(offset, 0x8), "little") != 0x00000001000600E9:
+                continue  # not _PROCESSOR_START_BLOCK->Jmp
+            potential_kernel_hint = int.from_bytes(physical_layer.read(offset + 0x70, 0x8), "little")
+            if (0xfffff80000000003 & potential_kernel_hint) != 0xfffff80000000000:
```
ikelos: Will this work with address spaces once the 64-bit space is extended past the currently reserved bits?  Will this work on a 32-bit system?

volatility3/framework/automagic/pdbscan.py
```diff
@@ -376,8 +377,43 @@ def check_kernel_offset(
...
+        kernel_hint = 0
+        kernel_base = 0
+        physical_layer = context.layers.get('memory_layer')
+
+        # try locating kernel base via x64 Low Stub in lower 1MB starting from second page (4KB)
+        # if "Discard Low Memory" setting is disabled in BIOS, the Low Stub may be at the third/fourth or further pages
+        for offset in range(0x1000,0x100000, 0x1000):
+            if 0xffffffffffff00ff & int.from_bytes(physical_layer.read(offset, 0x8), "little") != 0x00000001000600E9:
+                continue  # not _PROCESSOR_START_BLOCK->Jmp
+            potential_kernel_hint = int.from_bytes(physical_layer.read(offset + 0x70, 0x8), "little")
```
ikelos: Documenting these offsets, why +0x70, etc, would be really useful.

ikelos (commented):
Thanks very much, this looks really interesting!  Have you managed to test it against older systems (x86, PAE, etc)?  My concern is that it has a lot of hard coded hex bytes which may not be applicable to all x86-based architectures or memory layouts so I'd want to be careful about that.  It would also be good to get a few more comments (although I appreciate the continue comments explaining which bit the guard statements were looking for)...

Could you please also apply ruff (which just wants the unnecessary `struct` import removing) and black to format the file appropriately please?  Having it pass all the tests will give us a bit of confidence that it's working as intended...

volatility3/framework/automagic/pdbscan.py
atcuno: This method is only for 64-bit kernels, but it looks like the methods list is populated correctly so that this is method is checked then the current ones are used if it fails:

https://github.com/volatilityfoundation/volatility3/blob/c10572905f3f3760594db07575534a5805a10fe3/volatility3/framework/automagic/pdbscan.py#L415

ikelos: That wasn't my issue... The question was if it's 64-bit only, why not bomb out immediately if `not isinstance(vlayer, intel.Intel32e)`?  Seems like a waste to bother scanning systems you know it can't succeed for...

ikelos: (Also, the concern about support for when the higher bits are no longer reserved still stands).

Danking555:
Thanks for your feedback! I'm checking and working on everything you mentioned.

Danking555:
Hi @ikelos, I've addressed your feedback and made the following updates:

* Removed unnecessary ```struct``` import.
* Added the ```LowStubLayout``` class to centralize all relevant constants of the ```_PROCESSOR_START_BLOCK``` structure for better organization and maintainability.
* Added documentation for the ```LowStubLayout``` class and ```method_low_stub_offset```.
* Updated the guard statement to address the concern about reservation bits.
* Added a new guard check statement that compares the previously detected page table offset stored in ```vlayer._initial_entry``` with the value of ```cr3``` which can be deferred from the structure for additional validation. Let me know if you'd prefer it removed or adjusted.
* Confirmed the method works on three physical machines.
* After verifying functionality, applied ```ruff``` and ```black``` for code cleanup and formatting.

Please let me know if there's anything else you'd like me to refine or modify!

ikelos:
Thanks very much for clearing all those up!  It's looking really good now.  Rather than a separate class in the middle of the code, would it be possible to move the constants and their comments to the `constants.windows` module please?  All of our other windows constants live there so it just keeps things in a single place to find them all.  I guess move the docstring as an enlarged comment just before the section where you include those constants.  Then it looks like it's good to go in!  5:D

Danking555:
I've moved the constants into ```constants.windows```, added the docstring comment, and ran ruff and black again.
Let me know what you think-thanks!

ikelos (approved):
Yep, very happy with that, thanks!  5:D  I'll merge it as soon as the tests finish...

Issue #1547, opened 2025-01-13
Multiple plugins missing version/plugin requirements on imported plugins


atcuno:
I am putting this here to store the information until we are very close to the parity cut.

The base issue was found when I wrote a quick script to search for plugins that import another plugin, but do not have a Version or Plugin requirement on it. This found quite a few instances throughout the codebase. This will all be fixed at once after the artifact testing is done for Windows and Linux to avoid silly merge conflicts.

The second item that I am working with David is to have our GitHub actions check for these missing requirements during PRs to avoid them getting into the codebase again.

skalvino:
i can't see this 2 files

requirements-minimal.txt
requirements.txt

ikelos:
Plugins don't need to specify a version themselves if they don't have any public interface (classmethods) that could require version information by others, but yep, it would be good to ensure all calls to external modules have the correct version requirements behind them...

ikelos:
Any progress on the github actions?  We can fix up the deps once we know what they are...

atcuno:
Yes, its ready to go we are just waiting until last to not cause unnecessary merge conflicts.

atcuno:
This is finished given https://github.com/volatilityfoundation/volatility3/issues/1644 and other related PRs being merged.

PR #1567, opened 2025-01-23
Add deprecation decorator and split Linux modules utilities

Abyss-W4tcher:
Hi,

This PR adds a dedicated deprecation decorator to centralize the process. Sample:

```
WARNING  volatility3.framework.configuration.requirements: Method "volatility3.framework.symbols.linux.lookup_module_address" is deprecated, use "volatility3.framework.symbols.linux.utilities.modules.Modules.lookup_module_address" instead.
```

Happy to discuss on the verbosity too (WARNING, INFO, DEBUG ?).

A couple of modules related LinuxUtilities were deprecated and moved to a dedicated module. The "kernel" related utilities (even if it is a module) weren't moved out. The framework was updated to use the new APIs paths, so the deprecation messages will be output only for external plugins.

Note: Some plugins weren't bumped due to the lack of versioning, but there is already on-going work on this subject in #1547.

### Versioning 

A minor bump should be needed for the new decorator, but I'll wait for a review on the general idea first.

### DIFF
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -500,6 +500,16 @@ class CommandLine:
                 renderer.filter = text_filter.CLIFilter(grid, args.filters)
                 renderer.column_hide_list = args.hide_columns
                 renderer.render(grid)
+        except exceptions.UnsatisfiedException as excp:
+            output = sys.stderr
+            output.write(
+                "An unsatisfied framework exception was encountered post plugin construction:\n"
+            )
+            self.process_unsatisfied_exceptions(excp)
+            output.write(
+                f"Unable to validate the requirements: {[x for x in excp.unsatisfied]}\n",
+            )
+            sys.exit(1)
         except exceptions.VolatilityException as excp:
             self.process_exceptions(excp)
 
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -500,16 +500,6 @@ class CommandLine:
                 renderer.filter = text_filter.CLIFilter(grid, args.filters)
                 renderer.column_hide_list = args.hide_columns
                 renderer.render(grid)
-        except exceptions.UnsatisfiedException as excp:
-            output = sys.stderr
-            output.write(
-                "An unsatisfied framework exception was encountered post plugin construction:\n"
-            )
-            self.process_unsatisfied_exceptions(excp)
-            output.write(
-                f"Unable to validate the requirements: {[x for x in excp.unsatisfied]}\n",
-            )
-            sys.exit(1)
         except exceptions.VolatilityException as excp:
             self.process_exceptions(excp)
 
@@ -583,6 +573,8 @@ class CommandLine:
         fulltrace = traceback.TracebackException.from_exception(excp).format(chain=True)
         vollog.debug("".join(fulltrace))
 
+        file_a_bug_msg = f"Please re-run with -vvv and file a bug with the output at {constants.BUG_URL}"
+
         if isinstance(excp, exceptions.InvalidAddressException):
             general = "Volatility was unable to read a requested page:"
             if isinstance(excp, exceptions.SwappedInvalidAddressException):
@@ -627,9 +619,7 @@ class CommandLine:
         elif isinstance(excp, exceptions.LayerException):
             general = f"Volatility experienced a layer-related issue: {excp.layer_name}"
             detail = f"{excp}"
-            caused_by = [
-                "A faulty layer implementation (re-run with -vvv and file a bug)"
-            ]
+            caused_by = [f"A faulty layer implementation. {file_a_bug_msg}"]
         elif isinstance(excp, exceptions.MissingModuleException):
             general = f"Volatility could not import a necessary module: {excp.module}"
             detail = f"{excp}"
@@ -640,13 +630,17 @@ class CommandLine:
             general = "Volatility experienced an issue when rendering the output:"
             detail = f"{excp}"
             caused_by = ["An invalid renderer option, such as no visible columns"]
+        elif isinstance(excp, exceptions.VersionMismatchException):
+            general = "A version mismatch was detected between two components:"
+            detail = f"{excp}"
+            caused_by = [
+                excp.failure_reason or "An outdated API caller, such as a method.",
+                file_a_bug_msg,
+            ]
         else:
             general = "Volatility encountered an unexpected situation."
             detail = ""
-            caused_by = [
-                "Please re-run using with -vvv and file a bug with the output",
-                f"at {constants.BUG_URL}",
-            ]
+            caused_by = [file_a_bug_msg]
 
         # Code that actually renders the exception
         output = sys.stderr
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -77,14 +77,12 @@ class Deprecation:
             def wrapper(*args, **kwargs):
                 nonlocal replacement, replacement_version, additional_information
                 # Prevent version mismatches between deprecated (proxy) methods and the ones they proxy
-                if replacement_version is not None and callable(replacement):
-                    # example: replacement = volatility3.MyClass.my_dummy_function
-                    # "MyClass.my_dummy_function" -> "MyClass"
-                    replacement_base_class_name = replacement.__qualname__.split(".")[0]
-                    # replacement.__globals__ example: {'MyClass': <class 'volatility3.MyClass'>}
-                    replacement_base_class = replacement.__globals__.get(
-                        replacement_base_class_name
-                    )
+                if (
+                    replacement_version is not None
+                    and callable(replacement)
+                    and hasattr(replacement, "__self__")
+                ):
+                    replacement_base_class = replacement.__self__
 
                     # Verify that the base class inherits from VersionableInterface
                     if inspect.isclass(replacement_base_class) and issubclass(
@@ -99,7 +97,7 @@ class Deprecation:
                                 deprecated_func,
                                 replacement_base_class,
                                 replacement_version,
-                                "A deprecated method was unable to proxy the call to its replacement",
+                                "This is a bug, the deprecated call needs to be removed and the caller needs to update their code to use the new method.",
                             )
 
                 deprecation_msg = f"Method \"{deprecated_func.__module__ + '.' + deprecated_func.__qualname__}\" is deprecated, use \"{replacement.__module__ + '.' + replacement.__qualname__}\" instead. {additional_information}"
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -12,9 +12,11 @@ import logging
 import os
 import traceback
 import functools
+import warnings
 from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, Type, TypeVar
 
-from volatility3.framework import constants, interfaces
+from volatility3.framework import constants, exceptions, interfaces
+from volatility3.framework.configuration import requirements
 
 
 # ##
@@ -57,20 +59,64 @@ class Deprecation:
     """Deprecation related methods."""
 
     @staticmethod
-    def deprecated_method(replacement: Callable, additional_information: str = ""):
+    def deprecated_method(
+        replacement: Callable,
+        replacement_base_class_required_version: Tuple[int, int, int] = None,
+        additional_information: str = "",
+    ):
         """A decorator for marking functions as deprecated.
 
         Args:
             replacement: The replacement function overriding the deprecated API, in the form of a Callable (typically a method)
+            replacement_base_class_required_version: The "replacement" base class version that the deprecated method expects before proxying to it. This implies that "replacement" is a method from a class that inherits from VersionableInterface.
             additional_information: Information appended at the end of the deprecation message
         """
 
         def decorator(deprecated_func):
             @functools.wraps(deprecated_func)
             def wrapper(*args, **kwargs):
-                nonlocal replacement, additional_information
-                deprecation_msg = f"Method \"{deprecated_func.__module__ + '.' + deprecated_func.__name__}\" is deprecated, use \"{replacement.__module__ + '.' + replacement.__name__}\" instead. {additional_information}"
-                vollog.warning(deprecation_msg)
+                nonlocal replacement, replacement_base_class_required_version, additional_information
+                # Prevent version mismatches between deprecated (proxy) methods and the ones they proxy
+                if replacement_base_class_required_version is not None and callable(
+                    replacement
+                ):
+                    # example: replacement = volatility3.MyClass.my_dummy_function
+                    # "MyClass.my_dummy_function" -> "MyClass"
+                    replacement_base_class_name = replacement.__qualname__.split(".")[0]
+                    # replacement.__globals__ example: {'MyClass': <class 'volatility3.MyClass'>}
+                    replacement_base_class = replacement.__globals__.get(
+                        replacement_base_class_name
+                    )
+
+                    # Verify that the base class inherits from VersionableInterface
+                    if inspect.isclass(replacement_base_class) and issubclass(
+                        replacement_base_class,
+                        interfaces.configuration.VersionableInterface,
+                    ):
+                        # Construct a requirement
+                        req = requirements.VersionRequirement(
+                            name=replacement_base_class.__name__,
+                            component=replacement_base_class,
+                            version=replacement_base_class_required_version,
+                        )
+                        # Verify the requirement
+                        if not req.matches_required(
+                            req._version, req._component.version
+                        ):
+                            full_unsat_req_path = (
+                                deprecated_func.__module__
+                                + "."
+                                + deprecated_func.__qualname__
+                                + "."
+                                + req.name
+                            )
+                            # Catched by the cli and redirected to process_unsatisfied_exceptions
+                            raise exceptions.UnsatisfiedException(
+                                {full_unsat_req_path: req}
+                            )
+
+                deprecation_msg = f"Method \"{deprecated_func.__module__ + '.' + deprecated_func.__qualname__}\" is deprecated, use \"{replacement.__module__ + '.' + replacement.__qualname__}\" instead. {additional_information}"
+                warnings.warn(deprecation_msg, FutureWarning)
                 # Return the wrapped function with its original arguments
                 return deprecated_func(*args, **kwargs)
 
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -61,25 +61,23 @@ class Deprecation:
     @staticmethod
     def deprecated_method(
         replacement: Callable,
-        replacement_base_class_required_version: Tuple[int, int, int] = None,
+        replacement_version: Tuple[int, int, int] = None,
         additional_information: str = "",
     ):
         """A decorator for marking functions as deprecated.
 
         Args:
             replacement: The replacement function overriding the deprecated API, in the form of a Callable (typically a method)
-            replacement_base_class_required_version: The "replacement" base class version that the deprecated method expects before proxying to it. This implies that "replacement" is a method from a class that inherits from VersionableInterface.
+            replacement_version: The "replacement" base class version that the deprecated method expects before proxying to it. This implies that "replacement" is a method from a class that inherits from VersionableInterface.
             additional_information: Information appended at the end of the deprecation message
         """
 
         def decorator(deprecated_func):
             @functools.wraps(deprecated_func)
             def wrapper(*args, **kwargs):
-                nonlocal replacement, replacement_base_class_required_version, additional_information
+                nonlocal replacement, replacement_version, additional_information
                 # Prevent version mismatches between deprecated (proxy) methods and the ones they proxy
-                if replacement_base_class_required_version is not None and callable(
-                    replacement
-                ):
+                if replacement_version is not None and callable(replacement):
                     # example: replacement = volatility3.MyClass.my_dummy_function
                     # "MyClass.my_dummy_function" -> "MyClass"
                     replacement_base_class_name = replacement.__qualname__.split(".")[0]
@@ -93,26 +91,15 @@ class Deprecation:
                         replacement_base_class,
                         interfaces.configuration.VersionableInterface,
                     ):
-                        # Construct a requirement
-                        req = requirements.VersionRequirement(
-                            name=replacement_base_class.__name__,
-                            component=replacement_base_class,
-                            version=replacement_base_class_required_version,
-                        )
-                        # Verify the requirement
-                        if not req.matches_required(
-                            req._version, req._component.version
+                        # SemVer check
+                        if not requirements.VersionRequirement.matches_required(
+                            replacement_version, replacement_base_class.version
                         ):
-                            full_unsat_req_path = (
-                                deprecated_func.__module__
-                                + "."
-                                + deprecated_func.__qualname__
-                                + "."
-                                + req.name
-                            )
-                            # Catched by the cli and redirected to process_unsatisfied_exceptions
-                            raise exceptions.UnsatisfiedException(
-                                {full_unsat_req_path: req}
+                            raise exceptions.VersionMismatchException(
+                                deprecated_func,
+                                replacement_base_class,
+                                replacement_version,
+                                "A deprecated method was unable to proxy the call to its replacement",
                             )
 
                 deprecation_msg = f"Method \"{deprecated_func.__module__ + '.' + deprecated_func.__qualname__}\" is deprecated, use \"{replacement.__module__ + '.' + replacement.__qualname__}\" instead. {additional_information}"
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -11,7 +11,8 @@ import inspect
 import logging
 import os
 import traceback
-from typing import Any, Dict, Generator, List, Optional, Tuple, Type, TypeVar
+import functools
+from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, Type, TypeVar
 
 from volatility3.framework import constants, interfaces
 
@@ -50,6 +51,32 @@ def require_interface_version(*args) -> None:
                         ".".join(str(x) for x in args[0:2]),
                     )
                 )
+
+
+class Deprecation:
+    """Deprecation related methods."""
+
+    @staticmethod
+    def deprecated_method(replacement: Callable, additional_information: str = ""):
+        """A decorator for marking functions as deprecated.
+
+        Args:
+            replacement: The replacement function overriding the deprecated API, in the form of a Callable (typically a method)
+            additional_information: Information appended at the end of the deprecation message
+        """
+
+        def decorator(deprecated_func):
+            @functools.wraps(deprecated_func)
+            def wrapper(*args, **kwargs):
+                nonlocal replacement, additional_information
+                deprecation_msg = f"Method \"{deprecated_func.__module__ + '.' + deprecated_func.__name__}\" is deprecated, use \"{replacement.__module__ + '.' + replacement.__name__}\" instead. {additional_information}"
+                vollog.warning(deprecation_msg)
+                # Return the wrapped function with its original arguments
+                return deprecated_func(*args, **kwargs)
+
+            return wrapper
+
+        return decorator
 
 
 class NonInheritable:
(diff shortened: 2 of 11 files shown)


volatility3/framework/configuration/requirements.py
```diff
@@ -723,3 +724,25 @@ def build_configuration(
         """Builds the appropriate configuration for the specified
         requirement."""
         return context.modules[value].build_configuration()
+
+
+def deprecated_method(replacement: str, additional_information: str = ""):
+    """A decorator for marking functions as deprecated.
+
+    Args:
+        replacement: The replacement function overriding the deprecated API (full path preferred, starting from "volatility3."). String was preferred, for convenience and to prevent import conflicts on caller side.
```
ikelos: Given that the deprecated functions have to call the new modules (and therefore import them), I'm not sure why using a string is more convenient (particularly since the full path is much longer than the imported name)?  If we pull in the specific class, we can include the required version, do a version check and then either forward the call or throw a fatal error (not sure how useful that is, just means if we forget that we major bumped the new helper, we'd spot it more gracefully if anyone tried to use the old one before we removed it?)

Also, is requirements the right place to put this?  Feels a strange place to jam it, it might be better at the top level of `framework` or a new `framework.deprecation` module?

volatility3/framework/plugins/linux/keyboard_notifiers.py
```diff
@@ -66,8 +67,10 @@ def _generator(self):
         ):
             call_addr = call_back.notifier_call
 
-            module_name, symbol_name = linux.LinuxUtilities.lookup_module_address(
-                vmlinux, handlers, call_addr
+            module_name, symbol_name = (
+                linux_utilities_modules.Modules.lookup_module_address(
```
ikelos, line 76: Don't forget to include the new requirements in the plugins please.

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -263,8 +264,10 @@ def get_module_name_for_address(self, addr) -> str:
         """Helper to obtain the module and symbol name in the format needed for the
         output of this plugin.
         """
-        module_name, symbol_name = linux.LinuxUtilities.lookup_module_address(
-            self.vmlinux, self.handlers, addr
+        module_name, symbol_name = (
+            linux_utilities_modules.Modules.lookup_module_address(
```
ikelos, line 282: Needs a version requirement.

volatility3/framework/plugins/linux/tty_check.py
```diff
@@ -79,8 +80,10 @@ def _generator(self):
 
                 recv_buf = tty_dev.ldisc.ops.receive_buf
 
-                module_name, symbol_name = linux.LinuxUtilities.lookup_module_address(
-                    vmlinux, handlers, recv_buf
+                module_name, symbol_name = (
+                    linux_utilities_modules.Modules.lookup_module_address(
```
ikelos, line 89: Also needs a version requirement.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -382,41 +363,9 @@ def generate_kernel_handler_info(
...
-                        symbol_name = (
-                            symbols[0].split(constants.BANG)[1]
-                            if constants.BANG in symbols[0]
-                            else symbols[0]
-                        )
-
-                break
-
-        return mod_name, symbol_name
+        ] + linux_utilities_modules.Modules.mask_mods_list(
```
ikelos, line 389: Any reason not to shift generate_kernel_handler_info over into the same module, since they seem related?

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -504,6 +453,44 @@ def convert_fourcc_code(cls, code: int) -> str:
             [chr((code >> (i * 8)) & 0xFF) for i in range(code_bytes_length)]
         )
 
+    ## Deprecated APIs ##
```
ikelos: Any reason to move these in the source code?  It makes the diff much harder to follow because the various chunks of code get disassociated from each other?  I'd prefer to just add the changes, that move the whole function signature please...

volatility3/framework/plugins/linux/kthreads.py
```diff
@@ -88,8 +89,10 @@ def _generator(self):
                 if kthread.has_member("full_name")
                 else task_name
             )
-            module_name, symbol_name = linux.LinuxUtilities.lookup_module_address(
-                vmlinux, handlers, threadfn
+            module_name, symbol_name = (
+                linux_utilities_modules.Modules.lookup_module_address(
```
ikelos, line 98: Needs to be in the plugin requirements.

ikelos (commented):
Looks good, but a few comments.  Where we put the generic deprecation header is still an open question.  I don't think `configuration.requirements` is the right place for it.  It's more of a framework-wide helper...

volatility3/framework/symbols/linux/__init__.py
Abyss-W4tcher: I thought it would be easier to group deprecated APIs that are now pure proxies, for later readability ?

volatility3/framework/symbols/linux/__init__.py
Abyss-W4tcher: As specified in the description, this is more of a `Kernel` related utility, not exactly LKM's or built-ins modules. I think it would fit better in a `Kernel` LinuxUtility class.

volatility3/framework/plugins/linux/keyboard_notifiers.py
Abyss-W4tcher: Yes, sorry I'll update all the bits.

volatility3/framework/configuration/requirements.py
Abyss-W4tcher: This would be a good fit for a new class in `framework` indeed, right under `require_interface_version`.

A string allowed for more flexibility, I had a circular import situation but it turns out the fix is easy. I think I can revert to a `Callable`parameter.

Abyss-W4tcher: If we start versioning the new helper, won't it circle back to the "internal framework requirements issue" ? Do you plan on doing in-line versioning like `Deprecated(required_version=(1,0,0)).deprecated_method()` ?

Again, this would raise an exception at runtime, like the dynamic decorator I submitted a few weeks ago, which had the same "flaw".

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -688,6 +691,11 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                 description="Linux kernel",
                 architectures=["Intel32", "Intel64"],
             ),
+            requirements.VersionRequirement(
+                name="linux_utilities_modules",
```
ikelos, line 710: Hmmmm, I'm not super keen on the requirement needing to be remembered by anyone using the `AbstractNetfilter` class, but I can't see a clear way of move the requirement over to where it should live (without someone still having to remember there's a requirement that needs adding)...  5:S  Ah well, if you can think of something, shout, otherwise I'll let this one slide (since I already let the AbstractNetfilter class in long ago).

volatility3/framework/__init__.py
```diff
@@ -52,6 +53,32 @@ def require_interface_version(*args) -> None:
...
+        Args:
+            replacement: The replacement function overriding the deprecated API, in the form of a Callable (typically a method)
+            additional_information: Information appended at the end of the deprecation message
+        """
+
+        def decorator(deprecated_func):
+            @functools.wraps(deprecated_func)
+            def wrapper(*args, **kwargs):
+                nonlocal replacement, additional_information
+                deprecation_msg = f"Method \"{deprecated_func.__module__ + '.' + deprecated_func.__name__}\" is deprecated, use \"{replacement.__module__ + '.' + replacement.__name__}\" instead. {additional_information}"
```
ikelos: Can we include a test here of the callable's parent class, verify it's versionable and then check the version (if a version's been provided to the decorator)?  If that's two hard then passing in the class and the method name could work.  That way, it can throw an exception rather than just a warning if the version is suitably wrong...

Also, this shouldn't use vollog, it should use `warnings.warn(f"{name} is deprecated", FutureWarning)` or similar, since Deprecation warnings only show up once whereas normal log warnings may show up 100 times if called 100 times in the code (in a loop for example)...

volatility3/framework/symbols/linux/__init__.py
ikelos: I think the easily grepable decorator would be better.  There's pros and cons to both styles, but eventually when we do a deprecation pass, they'll just get removed, and anyone reading the code to find out about with will clearly see the decorator and the docstring.

ikelos (commented):
Thanks, I prefer the improvements, but still some tweaks I think we could do.  Lemme know what you think?

volatility3/framework/symbols/linux/__init__.py
ikelos: Hmmm, how are we going to have cross-module dependencies?  Probably ok, because they're all in the same codebase so released at the same time (should be difficult to have a module from one version, called by a class from another...)  Ok, that seems reasonable.

volatility3/framework/configuration/requirements.py
ikelos: Well, I don't see it as versioning, it's just a catch for up if we forget something.  The idea in my head was, we move everything to modules, update the module, bump the version but forget that the forwarder is still there.  Someone calls the forwarder and it fails with a big back trace and spaghetti and we have to unpick it to find out why.  If we take a version number for the proxied call, then we can fail more gracefully if we forget to update the deprecated call?  As you say, it's all internal, if we're on the ball there should never be a problem.  I just can't guarantee we're (I'm) always that observant...  5;P

ikelos: Basically, if the other one ever goes MAJOR version bump, this method now fails with a clear "this was removed and is no longer supported".  The version on this should never need bumping.

volatility3/framework/__init__.py
Abyss-W4tcher: Thanks for pointing out `warnings.warn`, I was also concerned about the message going out a lot.

I managed to grab the base class from the method, to make sanity checks and finally verify the version requirement. `UnsatisfiedException` will accomplish the nice backtrace needed to debug and patch the issue.

volatility3/framework/plugins/linux/netfilter.py
Abyss-W4tcher: I think gcmoreira thought about it, but I didn't notice it at first. I added a check in `AbstractNetfilter.__init__`.

volatility3/framework/__init__.py
Abyss-W4tcher: Example:

```
An unsatisfied framework exception was encountered post plugin construction:

Unsatisfied requirement volatility3.framework.symbols.linux.lookup_module_address.Modules: Version 1.2.0 dependency on volatility3.framework.symbols.linux.utilities.modules.Modules unmet
Unable to validate the requirements: ['volatility3.framework.symbols.linux.lookup_module_address.Modules']
```

volatility3/cli/__init__.py
```diff
@@ -500,6 +500,16 @@ def run(self):
                 renderer.filter = text_filter.CLIFilter(grid, args.filters)
                 renderer.column_hide_list = args.hide_columns
                 renderer.render(grid)
+        except exceptions.UnsatisfiedException as excp:
+            output = sys.stderr
```
ikelos: Why not just put a new entry in the `process_exceptions` handler?  We really don't need to be doing the full requirement/unsatisfied stuff, this is only when there's a mismatch between the framework old method and the framework new method...

volatility3/framework/__init__.py
```diff
@@ -57,20 +59,64 @@ class Deprecation:
...
+                        interfaces.configuration.VersionableInterface,
+                    ):
+                        # Construct a requirement
+                        req = requirements.VersionRequirement(
+                            name=replacement_base_class.__name__,
+                            component=replacement_base_class,
+                            version=replacement_base_class_required_version,
+                        )
+                        # Verify the requirement
+                        if not req.matches_required(
```
ikelos: I thought this was a class/static method?  It shouldn't a whole instance instantiating to check the version...

ikelos: Also, accessing private members is bad, we should have both version numbers (replacement class's version, and the one this deprecation warning requires).

volatility3/framework/__init__.py
```diff
@@ -57,20 +59,64 @@ class Deprecation:
...
+                        ):
+                            full_unsat_req_path = (
+                                deprecated_func.__module__
+                                + "."
+                                + deprecated_func.__qualname__
+                                + "."
+                                + req.name
+                            )
+                            # Catched by the cli and redirected to process_unsatisfied_exceptions
+                            raise exceptions.UnsatisfiedException(
```
ikelos: I really wouldn't worry about making the req for this.  It's going to be an internal issue for the codebase, so it just has to alert the user that something went very wrong and they should file a bug about it (including the original function called, and the proxied class it tried to call, both of which we should know).

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -345,6 +345,24 @@ def files_descriptors_for_process(
 
                 yield fd_num, filp, full_path
 
+    @classmethod
+    @Deprecation.deprecated_method(
+        replacement=linux_utilities_modules.Modules.mask_mods_list,
+        replacement_base_class_required_version=(1, 0, 0),
```
ikelos: I'd probably just go with `replacement_version`, I'm all for descriptive names, but that's getting a little excessive, even if we'll only use them a few times.  If they need to know what the version is for, the decorator should document that.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -345,6 +345,24 @@ def files_descriptors_for_process(
...
+        context: interfaces.context.ContextInterface,
+        layer_name: str,
+        mods: Iterator[interfaces.objects.ObjectInterface],
+    ) -> List[Tuple[str, int, int]]:
+        """
+        DEPRECATED: use "volatility3.framework.symbols.linux.utilities.modules.Modules.mask_mods_list" instead.
+
+        A helper function to mask the starting and end address of kernel modules
+        """
+        return linux_utilities_modules.Modules.mask_mods_list(context, layer_name, mods)
```
ikelos, line 364: Now that we know replacement, we could just let the decorator handle it with *args, **kwargs?

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -372,6 +390,27 @@ def generate_kernel_handler_info(
...
+        target_address: int,
+    ) -> Tuple[str, str]:
+        """
+        DEPRECATED: use "volatility3.framework.symbols.linux.utilities.modules.Modules.lookup_module_address" instead.
+
+        Searches between the start and end address of the kernel module using target_address.
+        Returns the module and symbol name of the address provided.
+        """
+        return linux_utilities_modules.Modules.lookup_module_address(
+            kernel_module.context, kernel_module.name, handlers, target_address
```
ikelos, line 411: Works less well for when the parameters need converting, so I guess maybe not doing it in the wrapper is better...  5;P

ikelos (commented):
Looking better, and definitely very thorough, but probably overkill.  It doesn't need any of the requirements machinery, it just needs to semver check (which I think is a static/classmethod).  You've done the hard part of finding the class from the method, so then just grab the version number and test it and throw an exception if they're wrong.  We can put a standard exception handler to explain what they should do...  5:)

volatility3/framework/symbols/linux/__init__.py
Abyss-W4tcher: This could be a good idea, but unfortunately this won't work well if we have to do transformation (like `lookup_module_address`). 

It also freezes the deprecated function in time and won't allow the use of new parameters (minor bumps) in the replacement, encouraging callers to migrate.

volatility3/framework/symbols/linux/__init__.py
Abyss-W4tcher: I am not sure to understand the concern here ? The "new" `lookup_module_address` was updated to fit the vol3 calling convention, and it shouldn't be problematic to make the proxy a transformer between the old and new calling when available ?

When people will migrate to the new API, they'll have to adapt to the new format, which by the way shouldn't be problematic as they'll already have a `vmlinux` object to get the context from.

Abyss-W4tcher:
I made the requested changes, hopefully it matches more closely what you had in mind ? I think the result is more explicit now:

```
A version mismatch was detected between two components:
volatility3.framework.symbols.linux.LinuxUtilities.lookup_module_address: Version (1, 2, 0) dependency on volatility3.framework.symbols.linux.utilities.modules.Modules (1, 0, 0) unmet.

        * A deprecated method was unable to proxy the call to its replacement
        * Please re-run with -vvv and file a bug with the output at https://github.com/volatilityfoundation/volatility3/issues

```

I'll count the required framework bumps at the end. :)

volatility3/framework/symbols/linux/__init__.py
ikelos: It's not a concern, it was kinda walking back my last comment about *args and **kwargs with this line as an example of why it wouldn't work (an API change between the deprecated function and the new function).

volatility3/framework/symbols/linux/__init__.py
ikelos: Forcing people to migrate is fine, but I agree (and that's what my comment below was about) it doesn't work if we want to do transformations, which clearly we do...  5;P

volatility3/cli/__init__.py
```diff
@@ -630,13 +630,17 @@ def process_exceptions(self, excp):
             general = "Volatility experienced an issue when rendering the output:"
             detail = f"{excp}"
             caused_by = ["An invalid renderer option, such as no visible columns"]
+        elif isinstance(excp, exceptions.VersionMismatchException):
+            general = "A version mismatch was detected between two components:"
+            detail = f"{excp}"
+            caused_by = [
+                excp.failure_reason or "An outdated API caller, such as a method.",
```
ikelos, line 637: I'd make this a little more explicit ("This is a bug, the deprecated call needs to be removed and the caller needs to update their code to use the new method").  We really want them to know they found a bug in the code and to file a ticket, rather than just go "oh, it's not working" and ignore it...

volatility3/framework/__init__.py
```diff
@@ -52,6 +55,63 @@ def require_interface_version(*args) -> None:
...
+
+        def decorator(deprecated_func):
+            @functools.wraps(deprecated_func)
+            def wrapper(*args, **kwargs):
+                nonlocal replacement, replacement_version, additional_information
+                # Prevent version mismatches between deprecated (proxy) methods and the ones they proxy
+                if replacement_version is not None and callable(replacement):
+                    # example: replacement = volatility3.MyClass.my_dummy_function
+                    # "MyClass.my_dummy_function" -> "MyClass"
+                    replacement_base_class_name = replacement.__qualname__.split(".")[0]
```
ikelos: __qualname__ doesn't quite do what we want I think?  We want the class that was called, not the class where the method was defined...

For example:
```
>>> class A:
...   version = (1,0,0)
...   def thing(self):
...     print(self.version)
... 
>>> class B(A):
...   version = (2,0,0)
... 
>>> ver2 = B()
>>> ver2.thing()
(2, 0, 0)
>>> ver2.thing.__qualname__
'A.thing'
```
I think we want to be using `__self__`:
```
>>> ver2.thing.__self__
<__main__.B object at 0x7f3049a33ec0>
```

That also saves us needing up lookup the class using `__globals__`.

volatility3/framework/exceptions.py
```diff
@@ -134,3 +135,31 @@ class RenderException(VolatilityException):
 
 class LinuxPageCacheException(VolatilityException):
     """Thrown if there is an error during Linux Page Cache processing"""
+
+
+class VersionMismatchException(VolatilityException):
```
ikelos, line 140: Again, feels a little overkill, but we do need a new exception for `process_exceptions` to figure out what's going on, so why not...  5:)

volatility3/framework/exceptions.py
```diff
@@ -134,3 +135,31 @@ class RenderException(VolatilityException):
...
+            failure_reason: A detailed failure reason to enhance debugging and bug tracking
+        """
+        super().__init__(*args)
+        self.source_component = source_component
+        self.target_component = target_component
+        self.target_version = target_version
+        self.failure_reason = failure_reason
+
+    def __str__(self):
+        return f"{self.source_component.__module__+ '.' + self.source_component.__qualname__}: Version {self.target_version} dependency on {self.target_component.__module__+ '.' + self.target_component.__qualname__} {self.target_component.version} unmet."
```
ikelos: Again, I think probably `self.target_component.__self__.__module__ + '.' + self.target_component.__self__.__class__.__name__` rather than `self.target_component.__qualname__`.  An intermediate variable to hold __self__ can probably make it easier to read...  5;)

volatility3/framework/plugins/linux/netfilter.py
```diff
@@ -98,6 +99,20 @@ def __init__(
                 f"linux.LinuxUtilities version not suitable: required {linuxutils_required_version} found {linuxutils_current_version}"
             )
 
+        linux_utilities_modules_required_version = (
```
ikelos, line 102: Can we leave this chunk out for the moment please?  I need to consider what it's doing and how valuable it is given how long winded it is and I'm more keen to get the rest of the work you've done into the codebase...

ikelos: Rather than throwing an exception, it would be cleaner to do the test in `run_all` and catch the exception gracefully, and log a warning saying that the implementation couldn't be tested because of the version discrepancy, but as I say, that can be a separate PR.

ikelos (commented):
Keeps getting better, just a couple more points and then I think we can get it across the line!  5:)

volatility3/cli/__init__.py
Abyss-W4tcher: Shouldn't this message go in the "failure_reason" in `deprecated_method` ? `VersionMismatch` tends to be generic, and not specifically related to deprecation. I can make both default and tailored messages more explicit anyway.

volatility3/framework/__init__.py
Abyss-W4tcher: `__self__` works on an stateful instance, however we are referencing `linux_utilities_modules.Modules.lookup_module_address` not `linux_utilities_modules.Modules().lookup_module_address`. I am not sure that instantiating the class is exactly what we want, as it might imply non-optional arguments sometimes.

That's a good point however, thanks for catching it ! Python internals are making it quite difficult to get `B` on this specific scenario.

I think using an extra class argument in `deprecated_method` will save us a lot of headaches, or sketchy/magical processing ?

volatility3/cli/__init__.py
ikelos: Oh, good point.  Yeah, I think one explicitly for deprecation issues where it's a bug in our code and they just need to tell us, we should give a stronger message.  VersionMismatch can stay as it is if it's still used anywhere else...

volatility3/framework/exceptions.py
Abyss-W4tcher: We can get more expressive with an additional exception indeed, else we currently have the plain `PluginVersionException`...

volatility3/framework/__init__.py
ikelos: It still seems to work for class methods?

```
>>> class A:
...   version = (1,0,0)
...   def thing(self):
...     print(self.version)
...   @classmethod
...   def classthing(cls):
...     print(cls.version)
... 
>>> class B(A):
...   version = (2,0,0)
... 
>>> B.classthing
<bound method A.classthing of <class '__main__.B'>>
>>> B.classthing.__self__
<class '__main__.B'>
```

Abyss-W4tcher: Yes, nice catch as well. However it doesn't work with `staticmethod` (thanks Python ?), which we use a lot.

ikelos: I think that's a better reason to go back through versioned components and change them from `staticmethod` to `classmethod`?  Also, this only affects new things so that we can deprecate old ones.  We can just require new ones to be `classmethods` (will need documenting somewhere, and we should probably go back through versioned things and make the staticmethods into classmethods (in a different PR!)...

volatility3/framework/exceptions.py
Abyss-W4tcher: `target_component` is a class, so the issue you pointed out earlier shouldn't really apply here: 

```
>>> print(B.__name__)
B
>> print(B.__qualname__)
B
```

volatility3/framework/__init__.py
Abyss-W4tcher: Honestly, I am not sure that it is worth the hassle, and `staticmethod` has valid use cases. Sometimes, a class only purpose is basically to group functions under it without necessarily making them communicate or access its properties ?

We could simply do the following to resolve the issue explicitely: 

```python
    @Deprecation.deprecated_method(
        replacement=linux_utilities_modules.Modules.lookup_module_address,
        replacement_base_class=linux_utilities_modules.Modules,  # <-- added
        replacement_version=(1, 0, 0),
    )
```

This also prevents edge cases that we haven't anticipated when accessing objects internals ?

ikelos: The valid use case is a slight boost in efficiency because it doesn't need to lug its class around with it, but now we need the class to get to the version property and otherwise the two are essentially indistinguishable?  I'm ok with staticmethods being used in certain places, but generally not in versioned classes (which essentially means all plugins).  I remember being a little against going to staticmethod when someone first tried to put one in a year or two ago.

I think people will find it odd that they need to define a second variable that's *already contained in the first* in order to get deprecation warnings to work, but also, I'd like to be able to get to the version of any method defined on a versionable component...

volatility3/framework/exceptions.py
ikelos: Fair enough, just figured it might be better to keep everything consistent.

volatility3/framework/__init__.py
ikelos: One of the first times I mentioned it...  5:S  https://github.com/volatilityfoundation/volatility3/pull/115#discussion_r339367397

Abyss-W4tcher: Ok if this is a requirement for versioned methods it makes sense to apply this restriction. I will check this PR (and my other ones) for any `staticmethod`.

Spider-Sense kicked in 5 years ago for an issue we are encountering today 🕸️

volatility3/framework/plugins/linux/netfilter.py
Abyss-W4tcher: This strictly follows the current accepted ways of checking for dependencies (see the previous lines of `__init__`). This design should already have been reviewed ?

volatility3/framework/__init__.py
```diff
@@ -52,6 +55,61 @@ def require_interface_version(*args) -> None:
...
+                    # Verify that the base class inherits from VersionableInterface
+                    if inspect.isclass(replacement_base_class) and issubclass(
+                        replacement_base_class,
+                        interfaces.configuration.VersionableInterface,
+                    ):
+                        # SemVer check
+                        if not requirements.VersionRequirement.matches_required(
+                            replacement_version, replacement_base_class.version
+                        ):
+                            raise exceptions.VersionMismatchException(
```
ikelos, line 96: I thought we were going to do the `warnings.warn` here, rather than bomb out with an exception?

volatility3/framework/plugins/linux/netfilter.py
ikelos: Yeah, it has, you're right, I'm just liking it less now it's growing...  5;P

volatility3/framework/__init__.py
ikelos: Oh wait, ignore me, this is only if there's a mismatch!

ikelos (approved):
Ok, I think we're there.  Thanks for your patience!

volatility3/framework/__init__.py
Abyss-W4tcher: This won't get to the deprecation message if the proxy cannot even call the replacement. We want to throw an exception here.

`warnings.warn` will shout if no version mismatch occurs.

Abyss-W4tcher:
@ikelos I was planning on bumping the framework version first ?

Thanks for your patience as well :)

Abyss-W4tcher:
In the end, one minor bump would be needed I think.

ikelos:
Happy for you to throw it in a separate PR if you want.  I guess it is an additive change...

PR #1571, opened 2025-01-24
chore(ci): pin ruff

c0rydoras:
pin ruff in ci, so that new ruff releases don't cause previously passing checks to fail

### DIFF
--- a/.github/workflows/ruff.yaml
+++ b/.github/workflows/ruff.yaml
@@ -13,3 +13,4 @@ jobs:
         with:
           args: check
           src: "."
+          version: 0.9.3


ikelos:
I'm not convinced we want to pin ourselves, we'd have to remember to update at some point and then fix all the differences, rather than fix them as they crop up?  That feels like it'll be more work just to keep some tests from breaking and identifying the fixes at the point they'd happen?  We can apply any non-related changes to the main tree if people don't want to put them in a separate commit in their PR, either way it feels like less work to keep current than to bunch up issues and have to fix them all at once?  Also, we don't have stats on how often ruff changes rules?

ikelos:
Thanks for the suggestion, but we're going to live with the pain and make sure we test against the latest suggestions and issues that ruff identifies.  Marking this as closed.

PR #1572, opened 2025-01-24
Windows: Allow windows.dlllist to report back DLLs from wow64 processes

hsarkey:
Updates to allow windows.dlllist to report back DLLs from wow64 processes. Created a new get_peb32() function and created a new symbol table (framework/symbols/windows/wow64.json).

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -485,10 +485,8 @@ class FILE_OBJECT(objects.StructType, pool.ExecutiveObject):
         ].is_valid(self.FileName.Buffer)
 
     def file_name_with_device(self) -> Union[str, interfaces.renderers.BaseAbsentValue]:
-        name: Union[str, interfaces.renderers.BaseAbsentValue] = (
-            renderers.UnreadableValue()
-        )
-        
+        name: Union[str, interfaces.renderers.BaseAbsentValue] = renderers.UnreadableValue()
+
         # this pointer needs to be checked against native_layer_name because the object may
         # be instantiated from a primary (virtual) layer or a memory (physical) layer.
         if self._context.layers[self.vol.native_layer_name].is_valid(self.DeviceObject):
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -776,7 +776,7 @@ class EPROCESS(generic.GenericIntelProcess, pool.ExecutiveObject):
         )
         return peb
 
-    def get_peb32(self) -> interfaces.objects.ObjectInterface:
+    def get_peb32(self) -> Optional[interfaces.objects.ObjectInterface]:
         """Constructs a PEB32 object"""
         if constants.BANG not in self.vol.type_name:
             raise ValueError(
@@ -834,6 +834,14 @@ class EPROCESS(generic.GenericIntelProcess, pool.ExecutiveObject):
         )
         return peb32
 
+    def set_types(self, peb) -> str:
+        ldr_data = self._context.symbol_space.get_type(
+            self._32bit_table_name + constants.BANG + "_PEB_LDR_DATA"
+        )
+        peb.Ldr = peb.Ldr.cast("pointer", subtype=ldr_data)
+        sym_table = self._32bit_table_name
+        return sym_table
+
     def load_order_modules(self) -> Iterable[interfaces.objects.ObjectInterface]:
         """Generator for DLLs in the order that they were loaded."""
         try:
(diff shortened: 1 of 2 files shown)


volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -775,39 +776,131 @@ def get_peb(self) -> interfaces.objects.ObjectInterface:
         )
         return peb
 
+    def get_peb32(self) -> interfaces.objects.ObjectInterface:
+        """Constructs a PEB32 object"""
+        if constants.BANG not in self.vol.type_name:
+            raise ValueError(
+                f"Invalid symbol table name syntax (no {constants.BANG} found)"
```
ikelos, line 783: I can't recall if this is strictly true.  I suspect it's true but I think it's possible you could ask the symbol table directly for this type, and it wouldn't necessarily include the table name.  Were you running into errors or is this just defensive coding?

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -775,39 +776,131 @@ def get_peb(self) -> interfaces.objects.ObjectInterface:
...
+        # add_process_layer can raise InvalidAddressException.
+        # if that happens, we let the exception propagate upwards
+        proc_layer_name = self.add_process_layer()
+        proc_layer = self._context.layers[proc_layer_name]
+
+        # Determine if process is running under WOW64.
+        if self.get_is_wow64():
+            proc = self.get_wow_64_process()
+        else:
+            return None
```
ikelos, line 795: This means the function signature needs to be `Optional[interfaces.objects.ObjectInterface]`.

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -775,39 +776,131 @@ def get_peb(self) -> interfaces.objects.ObjectInterface:
...
-                "InInitializationOrderLinks",
-            )
+            pebs = [
+                self.get_peb(),
+                self.get_peb32(),
+            ]
+            for peb in pebs:
+                if peb:
+                    sym_table = self.get_symbol_table_name()
+                    if peb.Ldr.vol.type_name.endswith("unsigned long"):
```
ikelos: Hard coding specific types in may be troublesome in the future.  There was an instance in the linux world where the pointers went from `unsigned long` to `long unsigned int` because of a change in Clang.  This also just verifies the end, rather than everything after the `constants.BANG` so you might catch unwanted/unexpected types.  Probably better to do something like `constants.BANG.split(...vol.type_name)[-1] == 'unsigned long'` although that would need testing to make sure it works properly in all instances.

ikelos: Would there be a big problem in just always casting the value, regardless of whether it's already the right type, or just verify it's a point to the right subtype and otherwise cast it?

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -775,39 +776,131 @@ def get_peb(self) -> interfaces.objects.ObjectInterface:
...
     def mem_order_modules(self) -> Iterable[interfaces.objects.ObjectInterface]:
         """Generator for DLLs in the order that they appear in memory"""
-
         try:
-            peb = self.get_peb()
-            yield from peb.Ldr.InMemoryOrderModuleList.to_list(
-                f"{self.get_symbol_table_name()}{constants.BANG}_LDR_DATA_TABLE_ENTRY",
-                "InMemoryOrderLinks",
-            )
+            pebs = [
```
ikelos, line 891: This feels like duplicating quite a big more code than the previous version.  It might be worth pulling out the pointer check on these into a separate little private subfunction?

ikelos (commented):
Thanks, this is really nice!  I like the way you've handled dealing with both type of pebs irrespective of which what it actually is and you've handled all the corner cases I can think of, good work!  Also, awesome comments, just what was needed when it was needed!  5:D

Just a couple of little points, I think it's probably better to check if the pointer's right, than check if it's specifically an unsigned long and then cast it.  Really good though, thanks!

ikelos:
Ok, looks good now, thanks.  Still not sure the constants.BANG check is necessary, but it shouldn't hurt to leave it in...  5:)

hsarkey:
Thanks for the feedback and review! For what it's worth,  I did run into errors without the constants.BANG check, so felt it was best to keep it in.

ikelos:
> Thanks for the feedback and review! For what it's worth, I did run into errors without the constants.BANG check, so felt it was best to keep it in.

That's fair enough.  We probably want to iron those out at some point, but good to know it was necessary at least.  Thanks for reporting back!  5:)

PR #1578, opened 2025-01-26
Shift exposed staticmethods to classmethods

### LINKED ISSUE #1575: Versioned components should only ever use `classmethod`s not `staticmethod`s to be able to recover the class's version
When versioning a component, all of its methods should be defined as classmethods in order to be able to recover the class it's a member of and therefore check the version.

Essentially any `staticmethod` on a `Versionable` class needs to be turned into a `classmethod`.  See #1567 for more discussion around the need.

ikelos:
This shifts all exposed staticmethods to classmethods, so that the class and version can be determined from just the method.

Fixes #1575.


gcmoreira (approved):
It looks good

PR #1579, opened 2025-01-27
Feature/coding style

ikelos:
As part of #1578 I figured we should really get the coding style more front and center.  This is where we can work together on a branch to get it in shape and make sure everyone's happy.  We should include reasoning behind our decisions to avoid arguing or bikeshedding.  Even if it ends up in here, it doesn't mean anything's set in stone, just that that's how we do it for now, but if there's a good reason to change in the future, then we can revise it...

If you all could take a look at it and put in anything we tend to do that I've forgotten to note down, or things you don't agree with (or that we don't do or stuff you think we should do), at least it'll be recorded and then we can discuss it to reach a consensus.  5:)

### DIFF
--- a/CODING_STYLE.md
+++ b/CODING_STYLE.md
@@ -81,7 +81,7 @@ Okay for one-liners. Prefer generator expressions over map() or filter() with a 
 Default Arguments
 -----------------
 
-Default arguments are fine, but not with mutable types (because they're consructed once at module load time and can lead to confusion/errors.)
+Default arguments are fine, but not with mutable types (because they're constructed once at module load time and can lead to confusion/errors.)
 
 True/False Evaluations
 ----------------------
--- a/CODING_STYLE.md
+++ b/CODING_STYLE.md
@@ -32,6 +32,12 @@ Exemptions from this rule:
         * `collections.abc` module
         * `typing_extensions` module
 
+Function calls
+--------------
+
+For longer function calls, where line length is no longer an issue, favour using keyword arguments for clarity over unnamed positional arguments.
+This helps coders learning the code from examples to know what parameters to pass in and avoids ordering mistakes.
+
 Global Mutable State
 --------------------
 
--- a/CODING_STYLE.md
+++ b/CODING_STYLE.md
@@ -2,61 +2,91 @@ Coding Standards
 ================
 
 The coding standards for volatility are mostly by our linter and our code formatter.
-All code submissions will be vetted automatically through tests from both and the submission
-will not be accepted if either of these fail.
+All code submissions will be vetted automatically through tests from both and the submission will not be accepted if either of these fail.
 
 Code Linter: Ruff
 Code Formatter: Black
 
-In additio, there are some coding practices that we employ to prevent specific failure cases and
-ensure consistency across the codebase.  These are documented below along with the rationale for the decision.
+In additiom, there are some coding practices that we employ to prevent specific failure cases and ensure consistency across the codebase.  These are documented below along with the rationale for the decision.
+
+This is heavily based upon https://google.github.io/styleguide/pyguide.html with minor modifications for volatility use.
 
 Imports
 -------
 
-Import should be of a module (not a class or a method), and ideally just one module except where naming would cause confusion.
-This is to prevent people importing an imported method (which can lead to confusion and add in an unnecessary dependency in the import chain).
+Use import statements for packages and modules only, not for individual types, classes, or functions and ideally not aliased naming would cause confusion.  This is to prevent people importing an imported method (which can lead to confusion and add in an unnecessary dependency in the import chain).
 
-Good example:
+* Use `import x` for importing packages and modules.
+* Use `from x import y where x` is the package prefix and y is the module name with no prefix.
+* Use `from x import y as z` in any of the following circumstances:
+    * Two modules named `y` are to be imported.
+    * `y` conflicts with a top-level name defined in the current module.
+    * `y` conflicts with a common parameter name that is part of the public API (e.g., `features`).
+    * `y` is an inconveniently long name.
+    * `y` is too generic in the context of your code (e.g., `from storage.file_system import options as fs_options`).
 
-```
-from module import submodule
-from module.submodule import submodule as subsubmodule
+Exemptions from this rule:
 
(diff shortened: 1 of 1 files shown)


CODING_STYLE.md
```diff
@@ -0,0 +1,98 @@
...
+
+This is heavily based upon https://google.github.io/styleguide/pyguide.html with minor modifications for volatility use.
+
+Imports
+-------
+
+Use import statements for packages and modules only, not for individual types, classes, or functions and ideally not aliased naming would cause confusion.  This is to prevent people importing an imported method (which can lead to confusion and add in an unnecessary dependency in the import chain).
+
+* Use `import x` for importing packages and modules.
+* Use `from x import y where x` is the package prefix and y is the module name with no prefix.
```
dgmcdona: I think that second backtick needs to be shifted left by two words

CODING_STYLE.md
```diff
@@ -0,0 +1,98 @@
+Coding Standards
+================
+
+The coding standards for volatility are mostly by our linter and our code formatter.
+All code submissions will be vetted automatically through tests from both and the submission will not be accepted if either of these fail.
+
+Code Linter: Ruff
+Code Formatter: Black
+
+In additiom, there are some coding practices that we employ to prevent specific failure cases and ensure consistency across the codebase.  These are documented below along with the rationale for the decision.
```
digitalisx:
```suggestion
In addition, there are some coding practices that we employ to prevent specific failure cases and ensure consistency across the codebase.  These are documented below along with the rationale for the decision.
```

CODING_STYLE.md
```diff
@@ -0,0 +1,98 @@
...
+
+    * Symbols from the following modules are used to support static analysis and type checking:
+        * `typing` module
+        * `collections.abc` module
+        * `typing_extensions` module
+
+Function calls
+--------------
+
+For longer function calls, where line length is no longer an issue, favour using keyword arguments for clarity over unnamed positional arguments.
```
digitalisx, line 38:
```suggestion
For longer function calls, where line length is no longer an issue, favor using keyword arguments for clarity over unnamed positional arguments.
```

CODING_STYLE.md
```diff
@@ -0,0 +1,98 @@
...
+
+Lambda functions
+----------------
+
+Okay for one-liners. Prefer generator expressions over map() or filter() with a lambda.
+
+Default Arguments
+-----------------
+
+Default arguments are fine, but not with mutable types (because they're consructed once at module load time and can lead to confusion/errors.)
```
digitalisx:
```suggestion
Default arguments are fine, but not with mutable types (because they're constructed once at module load time and can lead to confusion/errors.)
```

CODING_STYLE.md
```diff
@@ -0,0 +1,98 @@
...
+
+Shared functionality
+--------------------
+
+Within a plugin, there may be functions that are useful to other plugins.  These are created as `classmethod`s so that the plugin can be depended upon by other plugins in their requirements section, without needing to instantiate a whole copy of the plugin.  It is not a staticmethod, because the caller may wish to determine information about the class the method is defined in, and this is not easily accessible for staticmethods.
+A classmethod usually takes a `context` for its first method (and if it requires one, a configuration string for it second).  All other parameters should generally be basic types (such as strings, numbers, etc) so that future work requiring paralellization does not have complex types to have to keep in sync.  In particular, the idea was to ensure only one context was used per method (and each object brings its own context with it, meaning the function signature should not include objects to avoid discrepancies).
```
digitalisx:
```suggestion
A classmethod usually takes a `context` for its first method (and if it requires one, a configuration string for it second).  All other parameters should generally be basic types (such as strings, numbers, etc) so that future work requiring parallelization does not have complex types to have to keep in sync.  In particular, the idea was to ensure only one context was used per method (and each object brings its own context with it, meaning the function signature should not include objects to avoid discrepancies).
```

CODING_STYLE.md
ikelos: That's the UK spelling so until we decide on a region for documentation, I'm going to continue writing it natively...

eve-mem:
I'm not sure if my comments are coding style...

What about suggestions for variables names for common things? e.g. a `task` for linux processes or `context` for the context. It's used almost everywhere in the framework like this - to not do so would be confusing to me. 

What about the order of variables for a function, e.g. context first, then layer/symbol names, then other bits. It's almost always in that kind of order I think? What about suggesting passing things like layers by name rather than the actual objects. You touch on these in the classmethods part, but maybe suggest it for everything?

CODING_STYLE.md
```diff
@@ -0,0 +1,98 @@
...
+Code Formatter: Black
+
+In addition, there are some coding practices that we employ to prevent specific failure cases and ensure consistency across the codebase.  These are documented below along with the rationale for the decision.
+
+This is heavily based upon https://google.github.io/styleguide/pyguide.html with minor modifications for volatility use.
+
+Imports
+-------
+
+Use import statements for packages and modules only, not for individual types, classes, or functions and ideally not aliased naming would cause confusion.  This is to prevent people importing an imported method (which can lead to confusion and add in an unnecessary dependency in the import chain).
```
dgmcdona: This sentence reads a little strangely - if my understanding of the intent is correct, maybe it should be "... and ideally not aliased unless the imported name could otherwise cause confusion."

dgmcdona: "This is to prevent people < from> importing an imported <item? object?>"

dgmcdona:
I don't know if this is the best place for this or not, but this scenario has cropped up a couple of times now and it's worth paying attention to during dev + code review. We'll often use lists/sets/dicts to store offsets that we've already seen, but what ends up getting added is not a simple primitive but an `objects.Pointer` or `objects.Integer` which are substantially larger than the Python primitive, and it ends up blowing up memory usage. Maybe worth a mention in the style guide?

ikelos:
Sure, feel free to add commits to the branch adding appropriate sections?  It's not supposed to be an edict from on high, it's supposed to be a collaborative effort we all add to to get right...  5:)

ikelos:
Doesn't feel like anyone's had massive issues with this, so I'm going to merge it in and then people can PR changes they may want...

PR #1582, opened 2025-01-27
Embed additional plugin description in docstring

Abyss-W4tcher:
Hi,

This PR introduces a new way to provide an additional plugin description, by relying exclusively on its docstring.

The first line of the docstring will be used as the short help (`vol.py -h`), while the following lines will be printed at the bottom of the plugin full help (`vol.py myplugin -h`). This only directly impacts `windows.psxview` for now, given its current description (see the diff for more info). New result:

```sh
$ python3 vol.py windows.psxview -h         
Volatility 3 Framework 2.19.0
usage: vol.py windows.psxview.PsXView [-h] [--physical-offsets]

Lists all processes found via four of the methods described in "The Art of Memory Forensics" which may help identify processes that are trying to hide themselves.

options:
  -h, --help          show this help message and exit
  --physical-offsets  List processes with physical offsets instead of virtual offsets.

We recommend using -r pretty if you are looking at this plugin's output in a terminal.
```

Superseeds and reverts #1538. This might be worth adding this information to the plugin development guidelines, or the general coding style ?

### DIFF
--- a/volatility3/framework/plugins/linux/modxview.py
+++ b/volatility3/framework/plugins/linux/modxview.py
@@ -15,8 +15,8 @@ vollog = logging.getLogger(__name__)
 
 
 class Modxview(interfaces.plugins.PluginInterface):
-    """Centralize lsmod, check_modules and hidden_modules results to efficiently
-    spot modules presence and taints."""
+    """Centralize lsmod, check_modules and hidden_modules results to efficiently \
+spot modules presence and taints."""
 
     _version = (1, 0, 0)
     _required_framework_version = (2, 17, 0)
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -363,12 +363,13 @@ class CommandLine:
             metavar="PLUGIN",
         )
         for plugin in sorted(plugin_list):
-            # First line of a plugin docstring will be the short description for -h
-            # Following lines will be the additional description (argparse epilog)
+            # First line of a plugin docstring will be the short description for -h.
+            # Text after the first two consecutive new lines will be
+            # the additional description (argparse epilog).
             short_help = additional_help = None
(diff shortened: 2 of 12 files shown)


volatility3/framework/plugins/linux/modxview.py
```diff
@@ -15,8 +15,8 @@
 
 
 class Modxview(interfaces.plugins.PluginInterface):
-    """Centralize lsmod, check_modules and hidden_modules results to efficiently
```
ikelos, line 18: Ugh, that's not ideal.  Perhaps we could look for the first split two new lines appear without any text between them?

ikelos: Word wrapping is useful to keep things in check...

ikelos (commented):
Actually, generally pretty happy with that, just wondering about the new lines and the wrapping, I'd prefer not to need to remember `\`, but instead have it keyed off blank lines, so:

```Something new that goes
on for too long won't get split.

But once there's a line the rest gets split?

I dunno if we want a third line to denote notes that don't show up in help, but that seems a bad idea, I'd prefer people kept the notes useful rather than informal.
```

What do you think?

volatility3/framework/plugins/linux/modxview.py
Abyss-W4tcher: Sure, this will be easier to review and prevent mistakes for future docstrings. I updated the code accordingly.

I left the edited descriptions with "\\" as it should nonetheless be the way to make non newline separated sentences.

ikelos: For triple quoted strings it shouldn't matter, but they also don't matter, so happy to leave them in there...

ikelos (approved):

PR #1585, opened 2025-01-27
Expose and use `get_pefile_obj` rather than having local copies

### LINKED ISSUE #1577: `skeleton_key_check._get_pefile_obj` looks very similar to `pe_symbols._get_pefile_obj`
These two methods share almost identical code.  The main blocker for why they don't directly share code appears to be that they're both private (starting with `_`).  Should we expose the version in `pe_symbols` and then just depend on and use it in `skeleton_key_check`?

ikelos:
Fixes #1577


atcuno:
This is good to go ikelos. I tested on it samples with skeleton key's present and it still finds them

PR #1586, opened 2025-01-31
Framework improve conversion to strings

gcmoreira:
The current `array_to_string()` and `pointer_to_string()` functions introduce significant overhead during string conversion. They read large buffers (typically 256 or 512 bytes), create Volatility objects and strip null-terminated characters to finally return a Python string.

This PR refactors and consolidates these functions into a single `address_to_string()` function. This change reduces overhead and simplifies the code. Moreover, `address_to_string()` is now also useful in cases where, for some reason, the string address is an integer and not a Volatility Pointer. The overhead improvement comes from the introduction of a block reader, allowing customization of the block size according to the specific context. The `array_to_string()` and `pointer_to_string()` functions have been updated to support this block size customization as well.

Note that `array_to_string()` was originally documented to return an `interfaces.objects.ObjectInterface`, however, it actually returned a `str`, see below. 
```python
>>> type(array.cast("string", max_length=count, errors=errors))
<class 'str'>
```
This PR does not change the return type; it corrects the typing to improve the documentation.

Last but not least, it fixes a bug in the `String` class where it searches for the null character twice.

### DIFF
--- a/volatility3/framework/objects/utility.py
+++ b/volatility3/framework/objects/utility.py
@@ -127,7 +127,7 @@ def address_to_string(
 
     layer = context.layers[layer_name]
     text = b""
-    while len(text) <= count:
+    while len(text) < count:
         current_block_size = min(count - len(text), block_size)
         temp_text = layer.read(address + len(text), current_block_size)
         idx = temp_text.find(b"\x00")
--- a/volatility3/framework/objects/utility.py
+++ b/volatility3/framework/objects/utility.py
@@ -29,9 +29,23 @@ def bswap_64(value: int) -> int:
 
 
 def array_to_string(
-    array: "objects.Array", count: Optional[int] = None, errors: str = "replace"
-) -> interfaces.objects.ObjectInterface:
-    """Takes a volatility Array of characters and returns a string."""
+    array: "objects.Array",
+    count: Optional[int] = None,
+    errors: str = "replace",
+    block_size=32,
+) -> str:
+    """Takes a Volatility 'Array' of characters and returns a Python string.
+
+    Args:
+        array: The Volatility `Array` object containing character elements.
+        count: Optional maximum number of characters to convert. If None, the function
+               processes the entire array.
+        errors: Specifies error handling behavior for decoding, defaulting to "replace".
+        block_size: Reading block size. Defaults to 32
+
+    Returns:
+        A decoded string representation of the character array.
+    """
     # TODO: Consider checking the Array's target is a native char
     if not isinstance(array, objects.Array):
         raise TypeError("Array_to_string takes an Array of char")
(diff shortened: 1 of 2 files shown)


ikelos:
Wow, yep, fair enough.  I thought casting would return a vol object, but nope, you're right, it's a string!  This should be a great speed up!  Just need to handle bad lengths (I think that's what's causing the failing test at the moment)...

volatility3/framework/objects/utility.py
```diff
@@ -29,29 +29,115 @@ def bswap_64(value: int) -> int:
...
+        address: The address where the string is located in memory
+        count: The number of bytes to read
+        errors: The error handling scheme to use for encoding errors. Defaults to "replace"
+        block_size: Reading block size. Defaults to 32
+
+    Returns:
+        The decoded string extracted from memory.
+    """
+    if not isinstance(address, int):
+        raise TypeError("It takes an int")
```
ikelos, line 123: A little more descriptive please...  Things like "it" don't really belong in error messages.  5;P

volatility3/framework/objects/utility.py
```diff
@@ -29,29 +29,115 @@ def bswap_64(value: int) -> int:
...
+
+    if count < 1:
+        raise ValueError("It requires a positive count")
```
ikelos, line 126: Same again

gcmoreira:
yeah, it was < not <= :)

volatility3/framework/objects/utility.py
```diff
@@ -29,29 +29,115 @@ def bswap_64(value: int) -> int:
...
+
+    layer = context.layers[layer_name]
+    text = b""
+    while len(text) <= count:
+        current_block_size = min(count - len(text), block_size)
```
ikelos, line 131: This needs to be positive (which I guess includes not 0).  Probably should improve the error checking in the physical layer too, but it would/should throw an exception so might as well fix it here too...

ikelos (commented):
Looks good, but currently dying on one of the tests.  I think it's because current_block_size may end up as 0, so the `<=` need to just become a `<`, otherwise looks good!

ikelos:
Hehehe, caught it whilst I was finding it myself!  Something something something great minds...  5;P

gcmoreira:
yeah, I quickly fixed that ;)

gcmoreira:
ups, I was about to fix the error messages :)

ikelos:
Doh!  Got too excited, I just realized I hit fire too soon!  You can push direct to main just to fix them if you want...  5:)  Either that or I can do it...

ikelos:
(Got engrossed in trying to figure out the optimization in your other PR).

PR #1587, opened 2025-01-31
intel: address translation: performance improvements caching by page address

gcmoreira:
The current Intel address translation implementation includes caching when retrieving the respective table from memory in the `_get_valid_table()` function. However, there's still room for optimization by performing the translation based on page addresses as a whole, rather than translating each individual address. This change reduce the overhead and improve the overall performance.

### Before
```
$ time ./vol.py -f ./dump_ubuntu2404amd64_6.8.0-41-generic_debug.core linux.hidden
Volatility 3 Framework 2.19.0
Address Name

real    0m11.870s
user    0m11.254s
sys     0m0.601s

$ time ./vol.py -f ./dump_ubuntu2404amd64_6.8.0-41-generic_debug.core linux.hidden
Volatility 3 Framework 2.19.0
Address Name

real    0m12.211s
user    0m11.537s
sys     0m0.664s

$ time ./vol.py -f ./dump_ubuntu2404amd64_6.8.0-41-generic_debug.core linux.hidden
Volatility 3 Framework 2.19.0
Address Name

real    0m12.114s
user    0m11.456s
sys     0m0.649s
```
### After
```
$ time ./vol.py -f ./dump_ubuntu2404amd64_6.8.0-41-generic_debug.core linux.hidden
Volatility 3 Framework 2.19.0
Address Name

real    0m10.337s
user    0m9.730s
sys     0m0.599s

$ time ./vol.py -f ./dump_ubuntu2404amd64_6.8.0-41-generic_debug.core linux.hidden
Volatility 3 Framework 2.19.0
Address Name

real    0m10.335s
user    0m9.731s
sys     0m0.604s

$ time ./vol.py -f ./dump_ubuntu2404amd64_6.8.0-41-generic_debug.core linux.hidden
Volatility 3 Framework 2.19.0
Address Name

real    0m10.361s
user    0m9.743s
sys     0m0.618s
```

Note that the `@functools.lru_cache(maxsize=1025)` decorator for `_get_valid_table()` was intentionally removed, as it is no longer necessary and would only contribute to increased memory usage for the process.

### DIFF
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -186,6 +186,31 @@ class Intel(linear.LinearlyMappedLayer):
 
         Returns the translated entry value
         """
+        offset &= self.address_mask
+
+        if not (self.minimum_address <= offset <= self.maximum_address):
+            raise exceptions.InvalidAddressException(
+                offset, f"Address {offset:#x} outside virtual address range"
+            )
+
+        page_address = offset & self.page_mask
+        return self._translate_page(page_address)
+
+    @functools.lru_cache(maxsize=1024)
+    def _translate_page(self, page_address: int) -> int:
+        """Translates a page address based on paging tables.
+
+        Args:
+            page_address: The page base address
+
+        Returns:
+            the translated entry value
+        """
+        if page_address & ~self.page_mask != 0:
+            raise exceptions.InvalidAddressException(
+                page_address,
+                f"Invalid page address {page_address:#x}. The address must be aligned to the page size",
+            )
         # Setup the entry and how far we are through the offset
         # Position maintains the number of bits left to process
         # We or with 0x1 to ensure our page_map_offset is always valid
@@ -193,11 +218,13 @@ class Intel(linear.LinearlyMappedLayer):
         entry = self._initial_entry
 
         if not (
-            self.minimum_address <= (offset & self.address_mask) <= self.maximum_address
+            self.minimum_address
+            <= (page_address & self.address_mask)
+            <= self.maximum_address
         ):
             raise exceptions.PagedInvalidAddressException(
                 self.name,
-                offset,
+                page_address,
                 position + 1,
                 entry,
                 "Entry outside virtual address range: " + hex(entry),
@@ -209,7 +236,7 @@ class Intel(linear.LinearlyMappedLayer):
             if not self._page_is_valid(entry):
                 raise exceptions.PagedInvalidAddressException(
                     self.name,
-                    offset,
+                    page_address,
                     position + 1,
                     entry,
                     "Page Fault at entry " + hex(entry) + " in table " + name,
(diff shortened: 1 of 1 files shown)


volatility3/framework/layers/intel.py
```diff
@@ -186,18 +186,45 @@ def _translate_entry(self, offset: int) -> Tuple[int, int]:
 
         Returns the translated entry value
         """
+        offset &= self.address_mask
+
+        if not (self.minimum_address <= offset <= self.maximum_address):
```
ikelos, line 191: Applying the address mask *should* ensure this I think (maximum_address and address_mask are usually based off one another, I believe), but it's not necessarily a cast iron guarantee and there might be something funky that happens in the future so happy for this to be left in...

volatility3/framework/layers/intel.py
```diff
@@ -236,17 +263,15 @@ def _translate_entry(self, offset: int) -> Tuple[int, int]:
...
                     "Page Fault at entry " + hex(entry) + " in table " + name,
                 )
 
             # Read the data for the next entry
-            entry_data = table[
-                (index << self._index_shift) : (index << self._index_shift)
-                + self._entry_size
-            ]
+            entry_data_start = index << self._index_shift
+            entry_data = table[entry_data_start : entry_data_start + self._entry_size]
```
ikelos, line 274: I guess variable assignment saves time over bit-shifting twice?

volatility3/framework/layers/intel.py
```diff
@@ -186,18 +186,45 @@ def _translate_entry(self, offset: int) -> Tuple[int, int]:
...
+    def _translate_page(self, page_address: int) -> int:
+        """Translates a page address based on paging tables.
+
+        Args:
+            page_address: The page base address
+
+        Returns:
+            the translated entry value
+        """
+        if page_address & ~self.page_mask != 0:
```
ikelos, line 209: I think I tried to studiously avoid the use of `~` because with unbounded integers (python integers), because I don't know/trust what it'll do when applied to a fixed bit number, although apparently I've let it sneak in three times already...  5:S  As long as one of the two number is bounded, it should be fine because it's just being applied as a mask... so no way to create a number bigger than the minimum of the two...  I just worry about it extending far enough to cover large numbers, but I suppose python's already figured that out...

ikelos (commented):
I see, so the optimization comes from filtering the translations down to just pages and then applying a cache on that once they're all uniform.  Fair enough, that seems reasonable.  I'm always a little twitchy about updates to the core intel layer, and I was really worried we'd trip the exception at the start of translate_page, but the only caller (at the moment) ensures it first.  I'm still slightly surprised that the offset doens't matter, but I guess for entries those will always be paging table information rather than an actual offset...

volatility3/framework/layers/intel.py
gcmoreira: Not drastically hehe, but it does enhance readability a bit ;)

volatility3/framework/layers/intel.py
gcmoreira: yeah, we have to be careful with that and I found some issues already.. but that's safe because it has the AND with the integer (page_address)

PR #1595, opened 2025-01-31
Add smear checks and missing absolute flag to walk_internal_list

atcuno:
`walk_internal_list` was completely broken before due to a missing `absolute=True` parameter. This API was also missing the smear hardening that the other enumeration APIs have.


ikelos (commented):
Is the count needed?  Can the count ever top 4096 either a) legitimately or b) if there's runaway random data?  Happy to include it if so, just wanted to make sure it was necessary (and possible document better why it's there, rather than just saying that it's breaking at the count).

atcuno:
Fixed it to work better + docs @ikelos

ikelos (approved):
Looks ok to me, thanks.  5:)

PR #1596, opened 2025-01-31
Fix get_name API, fix malfind

atcuno:
The `get_name` API of vm_area_struct instances could (and did in testing) throw invalid address exception, which doesn't match how other `get_*` APIs work and really defeats the purpose of them.  So I fixed the API to properly return None on these errors and updated the typing to reflect that (it was missing before). I also changed the callers of this API to properly display a renderer value when the name isn't found.

malfind for vol3 was also strange in that it didn't show the path of code injection regions, which is a key artifact. Given that malfind uses `get_name`, I added the path fixes for it in along with its API fixing.

Sample: ggmemday1

### DIFF
--- a/volatility3/framework/plugins/linux/malfind.py
+++ b/volatility3/framework/plugins/linux/malfind.py
@@ -2,7 +2,7 @@
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
 
-from typing import List
+from typing import List, Tuple, Optional
 import logging
 from volatility3.framework import interfaces
 from volatility3.framework import renderers, symbols
@@ -39,7 +39,7 @@ class Malfind(interfaces.plugins.PluginInterface):
             ),
         ]
 
-    def _list_injections(self, task):
+    def _list_injections(self, task) -> Tuple[interfaces.objects.ObjectInterface, Optional[str], bytes]:
         """Generate memory regions for a process that may contain injected
         code."""
 
--- a/volatility3/framework/plugins/linux/malfind.py
+++ b/volatility3/framework/plugins/linux/malfind.py
@@ -56,10 +56,10 @@ class Malfind(interfaces.plugins.PluginInterface):
             )
             if (
                 vma.is_suspicious(proc_layer)
-                and vma.get_name(self.context, task) != "[vdso]"
+                and vma_name != "[vdso]"
             ):
                 data = proc_layer.read(vma.vm_start, 64, pad=True)
-                yield vma, data
+                yield vma, vma_name, data
 
     def _generator(self, tasks):
         # determine if we're on a 32 or 64 bit kernel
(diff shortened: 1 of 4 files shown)


volatility3/framework/plugins/linux/malfind.py
```diff
@@ -54,12 +54,9 @@ def _list_injections(self, task):
...
                 f"Injections : processing PID {task.pid} : VMA {vma_name} : {hex(vma.vm_start)}-{hex(vma.vm_end)}"
             )
-            if (
-                vma.is_suspicious(proc_layer)
-                and vma.get_name(self.context, task) != "[vdso]"
-            ):
+            if vma.is_suspicious(proc_layer) and vma_name != "[vdso]":
                 data = proc_layer.read(vma.vm_start, 64, pad=True)
-                yield vma, data
+                yield vma, vma_name, data
```
ikelos, line 61: It's an internal/non-exposed function, so this change is fine, but could we get a return type in the function signature please?

volatility3/framework/plugins/linux/malfind.py
```diff
@@ -88,6 +85,7 @@ def _generator(self, tasks):
                         process_name,
                         format_hints.Hex(vma.vm_start),
                         format_hints.Hex(vma.vm_end),
+                        vma_name or renderers.NotAvailableValue(),
```
ikelos, line 90: Should be ok, since the generator is private and the TreeView that eventually gets returned from `Run` should be inspectable/self-describing, so I think this is fine.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1072,6 +1072,12 @@ def get_name(self, context, task):
             fname = "Anonymous Mapping"
         return fname
 
+    def get_name(self, context, task) -> Optional[str]:
```
ikelos, line 1075: Yeah, there wasn't a return type defined, so theoretically you could make a case for it being ok, but I'd at least bump the PATCH number, so we know that changes were made and be able to differentiate if strange things start being reported please.  Hopefully this is in line to be moved to its own separate module in the same style as the ftrace stuff at some point...

ikelos (commented):
Well, not sure how you managed it, but looks like you managed to scrape by without really needing a version bump on anything!  I would still request at least a PATCH version bump on `LinuxUtilities` before I merge it, so we can more quickly identify what might have changed if people start experiencing issues...

volatility3/framework/symbols/linux/extensions/__init__.py
atcuno: ok where do I bump the PATCH number for this for now?

volatility3/framework/plugins/linux/malfind.py
atcuno: Added

PR #1598, opened 2025-01-31
Windows: Handles - catch exception in handle iteration

### LINKED ISSUE #1573: Handle bug round 2
I triggered the backtrace at the bottom of the ticket on latest develop. 

I know a bunch of stuff was fixed [here](https://github.com/volatilityfoundation/volatility3/issues/1493), so this looks like something that maybe wasn't reachable before due to the other bugs. This code is so complicated I don't want to break it, but it looks like maybe individual elements of the array being bad (pointers?) could trigger it. If thats the case just make sure we continue past the bad ones in the fix.

Sample: Win2016x64_14393.lime.raw

```
25-01-23 22:14:24 volatility3.cli DEBUG    Traceback (most recent call last):
  File "/home/analyst/hashtest/volatility3/cli/__init__.py", line 502, in run
    renderer.render(grid)
  File "/home/analyst/hashtest/volatility3/cli/text_renderer.py", line 232, in render
    grid.populate(visitor, outfd)
  File "/home/analyst/hashtest/volatility3/framework/renderers/__init__.py", line 240, in populate
  File "/home/analyst/hashtest/volatility3/framework/plugins/windows/handles.py", line 321, in _generator
    for entry in self.handles(object_table):
  File "/home/analyst/hashtest/volatility3/framework/plugins/windows/handles.py", line 292, in handles
    yield from self._make_handle_array(TableCode, table_levels)
  File "/home/analyst/hashtest/volatility3/framework/plugins/windows/handles.py", line 256, in _make_handle_array
    yield from self._make_handle_array(entry, level - 1, depth)
  File "/home/analyst/hashtest/volatility3/framework/plugins/windows/handles.py", line 246, in _make_handle_array
    for entry in table:
  File "/usr/local/lib/python3.8/_collections_abc.py", line 874, in __iter__
    v = self[i]
  File "/home/analyst/hashtest/volatility3/framework/objects/__init__.py", line 794, in __getitem__
    result += [self.vol.subtype(context=self._context, object_info=object_info)]
  File "/home/analyst/hashtest/volatility3/framework/objects/templates.py", line 96, in __call__
    return self.vol.object_class(
  File "/home/analyst/hashtest/volatility3/framework/objects/__init__.py", line 168, in __new__
  File "/home/analyst/hashtest/volatility3/framework/objects/__init__.py", line 408, in _unmarshall
    data = context.layers.read(object_info.layer_name, object_info.offset, length)
  File "/home/analyst/hashtest/volatility3/framework/interfaces/layers.py", line 635, in read
  File "/home/analyst/hashtest/volatility3/framework/layers/linear.py", line 45, in read
  File "/home/analyst/hashtest/volatility3/framework/layers/intel.py", line 302, in mapping
  File "/home/analyst/hashtest/volatility3/framework/layers/intel.py", line 358, in _mapping
  File "/home/analyst/hashtest/volatility3/framework/layers/intel.py", line 510, in _translate
  File "/home/analyst/hashtest/volatility3/framework/layers/intel.py", line 457, in _translate_swap
  File "/home/analyst/hashtest/volatility3/framework/layers/intel.py", line 166, in _translate
volatility3.framework.exceptions.PagedInvalidAddressException: Page Fault at entry 0x153e5400000000 in page entry
```

dgmcdona:
An `InvalidAddressException` can occur inside of `__iter__` when
iterating over the handle table (the exact exception occurs when
creating the subtype in `objects.Array.__getitem__`. This changes the
handle code to do a manual iteration over the sequence using the array
length and indexes, catch the exception, log the index, and continue.

In the test sample that prompted this change, the exception occurred on
the access of the very last item in the array.

closes #1573

### DIFF
--- a/volatility3/framework/plugins/windows/handles.py
+++ b/volatility3/framework/plugins/windows/handles.py
@@ -243,7 +243,12 @@ class Handles(interfaces.plugins.PluginInterface):
         layer_object = self.context.layers[virtual]
         masked_offset = offset & layer_object.maximum_address
 
-        for entry in table:
+        for i in range(len(table)):
+            try:
+                entry = table[i]
+            except exceptions.InvalidAddressException:
+                vollog.debug(f"Failed to get handle table entry at index {i}")
+                continue
             # This triggered a backtrace in many testing samples
             # in the level == 0 path
             # The code above this calls `is_valid` on the `offset`


volatility3/framework/plugins/windows/handles.py
```diff
@@ -243,7 +243,12 @@ def _make_handle_array(self, offset, level, depth=0):
         layer_object = self.context.layers[virtual]
         masked_offset = offset & layer_object.maximum_address
 
-        for entry in table:
+        for i in range(len(table)):
```
gcmoreira, line 246: Should the following be better?
```python
for index in enumerate(table):
```

gcmoreira: not sure if that will raise the exception on the same enumerate(). It shouldn't but you may need to double-check it

ikelos: I thought `enumerate` returned `index, value` not just `index`?

dgmcdona: I think it will raise the same exception, so we need to leave it as-is here:

```
 |  The enumerate object yields pairs containing a count (from start, which
 |  defaults to zero) and a value yielded by the iterable argument.
 |  
 |  enumerate is useful for obtaining an indexed list:
 |      (0, seq[0]), (1, seq[1]), (2, seq[2]), ...
 ```

ikelos (approved):

PR #1607, opened 2025-02-07
Make a generic DLL enumeration function that ensures the base address…

### LINKED ISSUE #1475: ldrmodules needs fix after dlllist update
ldrmodules is crashing when attempting to access a LDR_DATA_TABLE_ENTRY that isn't paged in:

```
raceback (most recent call last):
    renderer.render(grid)
    grid.populate(visitor, outfd)
  File "/home/ub/volatility3/volatility3/framework/plugins/windows/ldrmodules.py", line 54, in _generator
    load_order_mod = dict(
                     ^^^^^
  File "/home/ub/volatility3/volatility3/framework/plugins/windows/ldrmodules.py", line 55, in <genexpr>
    (mod.DllBase, mod) for mod in proc.load_order_modules()
     ^^^^^^^^^^^
```

Which is really a bug in load_order_modules. Hannah has a big PR coming to add support for 32bit/wow64 DLLs, so after that is merged I will write a fix for this one.

atcuno:
… is set and we return as many entries as possible #1475

This PR performs two major updates:

1) The previous code widely caught an invalid address exception, stopping processing early. This patch makes the try/except more granular so we recover much more entries

2) The previous code returned entries right on the edge of pages, so accessing a required member (DllBase) would then cause a backtrace. This broke ldrmodules in mass testing. We no longer return broken/useless entries.

Closes #1475

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -491,9 +491,9 @@ class FILE_OBJECT(objects.StructType, pool.ExecutiveObject):
         ].is_valid(self.FileName.Buffer)
 
     def file_name_with_device(self) -> Union[str, interfaces.renderers.BaseAbsentValue]:
-        name: Union[str, interfaces.renderers.BaseAbsentValue] = (
-            renderers.UnreadableValue()
-        )
+        name: Union[
+            str, interfaces.renderers.BaseAbsentValue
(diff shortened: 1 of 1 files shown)


ikelos (approved):
Looks good, I'm a little bit squeamish at diving into the `vol.type_name`, but I see that's how it was done previously so not for cleaning up in this PR at least...

PR #1608, opened 2025-02-07
Switch virtual and physical addresses to lists to support dumping mul…

### LINKED ISSUE #1319: Support multiple offsets in dumpfiles
In many situations, an individual file is cached multiple times, leading filescan to report several addresses. Right now, Volatility takes in `--virtaddr` and `--physaddr` as an IntRequirement. We need to switch this to a ListRequirement like other plugins and then adjust the code to extract from each offset.

atcuno:
…tiple files at once #1319

Closes #1319 

Dumpfiles only accepted one addresses at a time, which is really painful from a usability perspective. This switches it to accepting lists.

I tested this on old and new Windows version samples to test the virtual and physical list support.

### DIFF
--- a/volatility3/framework/plugins/windows/dumpfiles.py
+++ b/volatility3/framework/plugins/windows/dumpfiles.py
@@ -322,17 +322,19 @@ class DumpFiles(interfaces.plugins.PluginInterface):
         elif offsets:
             virtual_layer_name = kernel.layer_name
 
-            #FIXME - change this after standard access to physical layer
+            # FIXME - change this after standard access to physical layer
             physical_layer_name = self.context.layers[virtual_layer_name].config[
                 "memory_layer"
             ]
 
-           # Now process any offsets explicitly requested by the user.
+            # Now process any offsets explicitly requested by the user.
             for offset, is_virtual in offsets:
                 try:
                     file_obj = self.context.object(
                         kernel.symbol_table_name + constants.BANG + "_FILE_OBJECT",
-                        layer_name=virtual_layer_name if is_virtual else physical_layer_name,
+                        layer_name=(
+                            virtual_layer_name if is_virtual else physical_layer_name
+                        ),
                         native_layer_name=virtual_layer_name,
                         offset=offset,
                     )
--- a/volatility3/framework/plugins/windows/dumpfiles.py
+++ b/volatility3/framework/plugins/windows/dumpfiles.py
@@ -44,13 +44,15 @@ class DumpFiles(interfaces.plugins.PluginInterface):
                 description="Process ID to include (all other processes are excluded)",
                 optional=True,
             ),
-            requirements.IntRequirement(
+            requirements.ListRequirement(
                 name="virtaddr",
+                element_type=int,
                 description="Dump a single _FILE_OBJECT at this virtual address",
                 optional=True,
             ),
-            requirements.IntRequirement(
+            requirements.ListRequirement(
                 name="physaddr",
+                element_type=int,
                 description="Dump a single _FILE_OBJECT at this physical address",
                 optional=True,
             ),
@@ -318,6 +320,7 @@ class DumpFiles(interfaces.plugins.PluginInterface):
                         )
 
         elif offsets:
+
             # Now process any offsets explicitly requested by the user.
             for offset, is_virtual in offsets:
                 try:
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/dumpfiles.py
```diff
@@ -44,13 +44,15 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                 description="Process ID to include (all other processes are excluded)",
                 optional=True,
             ),
-            requirements.IntRequirement(
+            requirements.ListRequirement(
                 name="virtaddr",
+                element_type=int,
                 description="Dump a single _FILE_OBJECT at this virtual address",
```
ikelos: This should now be these addresses, presumably?

volatility3/framework/plugins/windows/dumpfiles.py
```diff
@@ -44,13 +44,15 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
                 optional=True,
             ),
-            requirements.IntRequirement(
+            requirements.ListRequirement(
                 name="physaddr",
+                element_type=int,
                 description="Dump a single _FILE_OBJECT at this physical address",
```
ikelos: This also shouldn't say single

volatility3/framework/plugins/windows/dumpfiles.py
```diff
@@ -355,10 +358,14 @@ def run(self):
...
-            offsets.append((self.config["virtaddr"], True))
-        elif self.config.get("physaddr", None) is not None:
-            offsets.append((self.config["physaddr"], False))
+        if self.config.get("virtaddr"):
+            for virtaddr in self.config["virtaddr"]:
+                offsets.append((virtaddr, True))
+
+        elif self.config.get("physaddr"):
+            for physaddr in self.config["physaddr"]:
+                offsets.append((physaddr, False))
```
ikelos, line 368: If the physical and virtual offsets all go into the same list, how do we ensure the offset found is from the right layer?  Something seems... off... about this plugin?

ikelos (commented):
I mean, it looks ok, but I have concerns about the accuracy of the plugin...  Could you give it a once over to explain why offsets is just a single list, before I merge this please?

volatility3/framework/plugins/windows/dumpfiles.py
ikelos: Oh, I see, we're storing a list of tuples...  5:S

volatility3/framework/plugins/windows/dumpfiles.py
```diff
@@ -318,6 +320,7 @@ def _generator(self, procs: List, offsets: List):
                         )
 
         elif offsets:
+
             # Now process any offsets explicitly requested by the user.
             for offset, is_virtual in offsets:
                 try:
```
ikelos, line 332: The code below this will re-lookup the layers for each offset, far better to pick out the physical and virtual layers first (assuming we can identify what those are, there should be a `FIXME` next to the memory_layer bit to say "only works with intel and should be fixed once we choose a specific means for finding physical layers"...

volatility3/framework/plugins/windows/dumpfiles.py
ikelos: Also, this `elif` means you'll only ever get one or the other in the list...  5:S

volatility3/framework/plugins/windows/dumpfiles.py
atcuno: done

volatility3/framework/plugins/windows/dumpfiles.py
atcuno: done

volatility3/framework/plugins/windows/dumpfiles.py
atcuno: Accepted both and then get the PIDs otherwise.

ikelos: Well, it accepts virtual, then it accepts physical if no virtual, then it tries pids.  If you define both, virtual will take precedence (because it's `elif`, not just `if`)

ikelos: The `else` then porbably wants to be a guard on both configs not being specified...

atcuno: Hey sorry, I was making comments while prepping the push to make sure I covered everything

volatility3/framework/plugins/windows/dumpfiles.py
atcuno: Adding this into the upcoming push too. Final test running now

ikelos (approved):
Ok, looks good, thanks!

atcuno:
@ikelos feedback is addressed now. Let me know if you still have questions about the offset handling. All my changes were aimed at making virtual and physical into lists instead of one offset. The behaviour of PID handling stayed the same.

ikelos:
Yep, made sense and looked good, thanks  5:)

PR #1609, opened 2025-02-08
Prevent yielding smeared/broken modules from the unloaded module list

atcuno:
Dozens of memory samples crashed in testing from smeared/broken module entries being yielded from the list. We now ensure the three relevant members are valid before yielding a driver instance.


ikelos (approved):
Looks good (although that syntax still seems odd to me, but I guess it's our style now...)  5:P

PR #1610, opened 2025-02-08
New plugin: linux.tracing.tracepoints

### LINKED ISSUE #1286: Add detection of malicious ftrace and tracepoints

Abyss-W4tcher:
Hi,

This new plugin adds detection of malicious kernel functions hooking (see https://github.com/volatilityfoundation/volatility3/issues/1286), by inspecting the tracepoints subsystem. Attackers can leverage its capabilities to hook critical low level calls, and perform man-in-the middle or output manipulation.

It also relies on the recent linux.hidden_modules and linux.modxview to extend the uncovering of probe handlers parents/modules. Its structure is quite similar to `linux.tracing.ftrace`.

### Sample runs

```sh
$ python3 vol.py -r pretty -f Ubuntu-xenial_4.4.0-210-generic_tracepoints.lime linux.tracing.tracepoints
Volatility 3 Framework 2.20.0
  |   tracepoint | tracepoint address |              Probe |  Probe address | Probe priority |         Module | Module address
* | sched_switch |     0xffff81d83d50 | probe_sched_switch | 0xffffc0386000 |             10 | tracepoint_mod | 0xffffc0388040
* | sched_wakeup |     0xffff81d83d60 | probe_sched_wakeup | 0xffffc0386010 |             10 | tracepoint_mod | 0xffffc0388040
```

```sh
$ python3 vol.py -r pretty -f Ubuntu-jammy_5.15.0-87-generic_tracepoints.lime linux.tracing.tracepoints
Volatility 3 Framework 2.20.0
  |    tracepoint | tracepoint address |               Probe |  Probe address | Probe priority |         Module | Module address
* |  mm_page_free |     0xffffa7df8820 |  probe_mm_page_free | 0xffffc08a0010 |             10 | tracepoint_mod | 0xffffc08a20c0
* | mm_page_alloc |     0xffffa7df8760 | probe_mm_page_alloc | 0xffffc08a0000 |             10 | tracepoint_mod | 0xffffc08a20c0
```

```sh
$ python3 vol.py -r pretty -f Debian-bookworm_6.1.0-13-amd64_tracepoints.lime linux.tracing.tracepoints
Volatility 3 Framework 2.20.0
  |    tracepoint | tracepoint address |               Probe |  Probe address | Probe priority |         Module | Module address
* |  mm_page_free |     0xffff9da0bf80 |  probe_mm_page_free | 0xffffc0738010 |             10 | tracepoint_mod | 0xffffc073a080
* | mm_page_alloc |     0xffff9da0bec0 | probe_mm_page_alloc | 0xffffc0738000 |             10 | tracepoint_mod | 0xffffc073a080
```

Closes #1286.

### DIFF
--- a/volatility3/framework/plugins/linux/tracing/tracepoints.py
+++ b/volatility3/framework/plugins/linux/tracing/tracepoints.py
@@ -212,10 +212,10 @@ if the "hidden_modules" key is present in known_modules.
 
         tracepoints = []
         tracepoints_start = kernel.object_from_symbol("__start___tracepoints_ptrs")
-        tracepoints_end = kernel.object_from_symbol("__stop___tracepoints_ptrs")
-        tracepoints_array_size = (
-            tracepoints_end.vol.offset - tracepoints_start.vol.offset
+        tracepoints_end = kernel.get_absolute_symbol_address(
+            "__stop___tracepoints_ptrs"
         )
+        tracepoints_array_size = tracepoints_end - tracepoints_start.vol.offset
         # kernel's tracepoint_ptr_deref() and tracepoint_ptr_t
         # adjust depending on the use of PC-relative addressing
         # or not.
--- a/volatility3/framework/plugins/linux/tracing/tracepoints.py
+++ b/volatility3/framework/plugins/linux/tracing/tracepoints.py
@@ -217,8 +217,10 @@ if the "hidden_modules" key is present in known_modules.
             tracepoints_end.vol.offset - tracepoints_start.vol.offset
         )
         # kernel's tracepoint_ptr_deref() and tracepoint_ptr_t
-        # adjust depending on the use of relocated pointers
-        # or not
+        # adjust depending on the use of PC-relative addressing
+        # or not.
+        # Relocation is commonly used to store pointers as offsets
+        # relative to their own address rather than absolute addresses/pointers.
         config_have_arch_prel32_relocations = (
             tracepoints_start.vol.subtype.type_name
             == kernel.symbol_table_name + constants.BANG + "int"
@@ -230,9 +232,13 @@ if the "hidden_modules" key is present in known_modules.
                 subtype=kernel.get_type("int"),
             )
             for relative_offset in tracepoints_relative_offsets:
+                # relative_offset is the value stored at relative_offset.vol.offset
+                # See kernel's offset_to_ptr(). Example:
+                # 0xffff9da125e0 = 0x7af138        + 0xffff9d2634a8
+                absolute_address = relative_offset + relative_offset.vol.offset
                 tracepoint = kernel.object(
                     "tracepoint",
-                    relative_offset + relative_offset.vol.offset,
+                    absolute_address,
                     absolute=True,
                 )
                 tracepoints.append(tracepoint)
--- a/volatility3/framework/plugins/linux/tracing/tracepoints.py
+++ b/volatility3/framework/plugins/linux/tracing/tracepoints.py
@@ -1,0 +1,305 @@
+# This file is Copyright 2025 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+# Public researches: https://i.blackhat.com/USA21/Wednesday-Handouts/us-21-Fixing-A-Memory-Forensics-Blind-Spot-Linux-Kernel-Tracing-wp.pdf
+
+import logging
+from typing import Dict, Iterable, List, Optional
+from dataclasses import dataclass
+
+import volatility3.framework.symbols.linux.utilities.modules as linux_utilities_modules
+from volatility3.plugins.linux import hidden_modules, modxview
+from volatility3.framework import constants, exceptions, interfaces
+from volatility3.framework.configuration import requirements
+from volatility3.framework.renderers import format_hints, NotAvailableValue, TreeGrid
+from volatility3.framework.symbols.linux import extensions
+from volatility3.framework.objects import utility
+from volatility3.framework.constants import architectures
+
+vollog = logging.getLogger(__name__)
+
+
+@dataclass
+class ParsedTracepointFunc:
+    """Parsed tracepoint_func struct, containing a selection of forensics valuable
+    informations."""
+
+    tracepoint_name: str
+    tracepoint_address: int
+    probe_name: str
+    probe_address: int
+    probe_priority: int
+    module_name: str
+    module_address: int
+
+
+class CheckTracepoints(interfaces.plugins.PluginInterface):
+    """Detect tracepoints hooking
+
+    Investigate the tracepoints subsystem to uncover kernel attached probes, which can be leveraged
+    to hook kernel functions and modify their behaviour."""
+
+    _version = (1, 0, 0)
+    _required_framework_version = (2, 19, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Linux kernel",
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/linux/tracing/tracepoints.py
```diff
@@ -0,0 +1,305 @@
...
+        Handles CONFIG_HAVE_ARCH_PREL32_RELOCATIONS.
+
+        Returns:
+            A list of tracepoint structs
+        """
+
+        kernel = context.modules[kernel_name]
+
+        tracepoints = []
+        tracepoints_start = kernel.object_from_symbol("__start___tracepoints_ptrs")
```
ikelos, line 214: I'm not sure these need creating as objects?  It's fine, but you don't seem to necessarily use it for the object it is, you more just pull out the offset and type, which you could do by just looking up the symbol (without constructing the object)?  Might be slightly more efficient is all...

volatility3/framework/plugins/linux/tracing/tracepoints.py
```diff
@@ -0,0 +1,305 @@
...
+        tracepoints_end = kernel.object_from_symbol("__stop___tracepoints_ptrs")
+        tracepoints_array_size = (
+            tracepoints_end.vol.offset - tracepoints_start.vol.offset
+        )
+        # kernel's tracepoint_ptr_deref() and tracepoint_ptr_t
+        # adjust depending on the use of relocated pointers
+        # or not
+        config_have_arch_prel32_relocations = (
+            tracepoints_start.vol.subtype.type_name
+            == kernel.symbol_table_name + constants.BANG + "int"
```
ikelos, line 226: This seems a little brittle.  Can we be certain that it'll always have that type?

volatility3/framework/plugins/linux/tracing/tracepoints.py
```diff
@@ -0,0 +1,305 @@
...
+        )
+        if config_have_arch_prel32_relocations:
+            tracepoints_relative_offsets = tracepoints_start.cast(
```
ikelos, line 229: How is this different o the `utility.array_of_pointers` call below?  Feels like these should do the same thing?
At least document with a comment why these two cases exist.  It feels like it's making them integers to then use as pointers?  Could they all just be recast as pointers to the right type and then treated the same?

ikelos (commented):
Looks pretty good.  A couple of questions about efficiency so I'll give you a chance to look them over before committing, but not really any showstoppers.  Please write a response on here, one way or another, to tell me when you're happy with it...

volatility3/framework/plugins/linux/tracing/tracepoints.py
Abyss-W4tcher: I left a comment in the code, it is related to relocations (https://elixir.bootlin.com/linux/v6.5/source/include/linux/tracepoint.h#L113): 

```c
#ifdef CONFIG_HAVE_ARCH_PREL32_RELOCATIONS
static inline struct tracepoint *tracepoint_ptr_deref(tracepoint_ptr_t *p)
{
	return offset_to_ptr(p);
}

#else
static inline struct tracepoint *tracepoint_ptr_deref(tracepoint_ptr_t *p)
{
	return *p;
}
```

```c
/**
 * offset_to_ptr - convert a relative memory offset to an absolute pointer
 * @off:	the address of the 32-bit offset value
 */
static inline void *offset_to_ptr(const int *off)
{
	return (void *)((unsigned long)off + *off);
}
```

As you can see, the implementation tends to follow this mechanism. Relocations are used to save 4 bytes when storing the "pointer".

Happy to add more comments if needed.

volatility3/framework/plugins/linux/tracing/tracepoints.py
Abyss-W4tcher: Based on the kernel implementation (see the comment below), it will be an `int` when relocations are used.

ikelos: So, only 4 bytes, for what could be an 8-byte pointer?

volatility3/framework/plugins/linux/tracing/tracepoints.py
ikelos: Yeah, I think the 3 lines you left aren't really enough to describe what's going on (I still don't have a clear picture in my head).

Abyss-W4tcher: Relocation is commonly used to store pointers as offsets relative to their own address rather than absolute addresses.

The content **and** the position of a  `__start___tracepoints_ptrs` entry are required to determine the true location of the `tracepoint` object. 

Here is a direct example from the code, to illustrate:
```python
# relative_offset is the value stored at relative_offset.vol.offset
for relative_offset in tracepoints_relative_offsets:
    vollog.info(f"{hex(relative_offset)} + {hex(relative_offset.vol.offset)} = {hex(relative_offset + relative_offset.vol.offset)}")
```

> INFO     volatility3.plugins.linux.tracing.tracepoints: 0x7af138 + 0xffff9d2634a8 = 0xffff9da125e0
INFO     volatility3.plugins.linux.tracing.tracepoints: 0x7af0d4 + 0xffff9d2634ac = 0xffff9da12580
INFO     volatility3.plugins.linux.tracing.tracepoints: 0x7b0750 + 0xffff9d2634b0 = 0xffff9da13c00

*On a side note, using `PC-relative addressing (PREL32)` like this allows better security against GOT/ROP attacks, as it is harder to patch a relative pointer to hijack control flow.*

volatility3/framework/plugins/linux/tracing/tracepoints.py
Abyss-W4tcher: Yes, either an 8 byte "classic" pointer, or an integer indicating the need to resolve it into an absolute address.

volatility3/framework/plugins/linux/tracing/tracepoints.py
Abyss-W4tcher: In fact, the object creation is useful for easier array manipulation in `utility.array_of_pointers(` and `.cast("array")`. I thought it would be more convenient that way ? Happy to use `get_absolute_symbol_address` for `__stop___tracepoints_ptrs` however.

volatility3/framework/plugins/linux/tracing/tracepoints.py
```diff
@@ -0,0 +1,311 @@
...
+                "array",
+                count=tracepoints_array_size // kernel.get_type("int").size,
+                subtype=kernel.get_type("int"),
+            )
+            for relative_offset in tracepoints_relative_offsets:
+                # relative_offset is the value stored at relative_offset.vol.offset
```
ikelos, line 235: Thanks, this example makes it clearer.

volatility3/framework/plugins/linux/tracing/tracepoints.py
ikelos: Yeah, ok, as long as there's a good reason for making the whole object.  If you just needed the values, then a `get_symbol` would probably less work, but it's unlikely to have a significant impact on the runtime...

volatility3/framework/plugins/linux/tracing/tracepoints.py
ikelos: Ok, I see now.  I'm not sure just from the earlier read that I took away relative meant before its immediate use.  The background is useful.

PR #1611, opened 2025-02-09
Re-apply "Pre linux.pagecache.recoverfs support"

Abyss-W4tcher:
Hi,

This PR re-applies changes required for the upcoming `linux.pagecache.recoverfs` plugin, after they were reverted. Please see #1561 for the original description. 

As discussed, this should not raise any PR conflicts now.

### DIFF
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -417,7 +417,7 @@ class InodePages(plugins.PluginInterface):
 
     _required_framework_version = (2, 0, 0)
 
-    _version = (2, 0, 2)
+    _version = (3, 0, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -465,51 +465,68 @@ class InodePages(plugins.PluginInterface):
             inode: The inode to dump
             filename: Filename for writing the inode content
             open_method: class for constructing output files
-            vmlinux_layer: The kernel layer to obtain the page size
+        """
+        try:
+            with open_method(filename) as file_obj:
+                cls.write_inode_content_to_stream(context, layer_name, inode, file_obj)
+        except OSError as e:
+            vollog.error("Unable to write to file (%s): %s", filename, e)
+
+    @classmethod
+    def write_inode_content_to_stream(
+        cls,
+        context: interfaces.context.ContextInterface,
+        layer_name: str,
+        inode: interfaces.objects.ObjectInterface,
+        stream: IO,
+    ) -> None:
+        """Extracts the inode's contents from the page cache and saves them to a stream
+
+        Args:
+            context: The context on which to operate
+            layer_name: The name of the layer on which to operate
+            inode: The inode to dump
+            stream: An IO stream to write to, typically FileHandlerInterface or BytesIO
         """
         if not inode.is_reg:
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -37,6 +38,11 @@ class InodeUser:
     modification_time: str
     change_time: str
     path: str
+    inode_size: int
```
ikelos, line 41: How do we feel about changing a dataclass, given it's not versioned?  Hopefully no one will come to rely on inode_size, but we really need to be careful about just slotting in extra data and then making use of it.  Fine for us, not so good for external plugins trying to access the same data.  The plugin that uses it is bumped, but we need to be careful about this kind of thing...

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -80,6 +86,7 @@ def to_user(
         access_time_dt = self.inode.get_access_time()
         modification_time_dt = self.inode.get_modification_time()
         change_time_dt = self.inode.get_change_time()
+        inode_size = int(self.inode.i_size)
```
ikelos, line 89: Is this already an `int`?   Why are we casting this?

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -95,6 +102,7 @@ def to_user(
             modification_time=modification_time_dt,
             change_time=change_time_dt,
             path=self.path,
+            inode_size=inode_size,
```
ikelos, line 105: This value never seems to be referenced, so why are we changing this structure to include it, if it's not necessary?  It really should've been in the first version if it was important.

ikelos (commented):
So, there's an unnecessary cast (really trying to get people beyond casting these things when they're already `int`s) and I'm uncomfortable about trying to slide `inode_size` into the structure given it isn't used and wasn't deemed important in the first iteration of this plugin.   The plugin version has been bumped, so I'm willing to let it slide this time but it's really dangerous for maintaining compatibility with things and I'd prefer we spend longer thinking through dataclass wrappers when we first make then, than trying to fudge in additional data after the fact...

volatility3/framework/plugins/linux/pagecache.py
Abyss-W4tcher: `inode_size` is actually outputted now under the `InodeSize` column. It is a valuable piece of information to determine the size of an inode, and eventually later compare it to the size in memory.

volatility3/framework/plugins/linux/pagecache.py
Abyss-W4tcher: As specified by gcmoreira's comment beforehand, it is needed to avoid nested deepcopies. It slows down processing by a lot.

volatility3/framework/plugins/linux/pagecache.py
Abyss-W4tcher: If other plugins manipulate `InodeUser` live instances, it should be fine. If one needs to create those manually by hand, it could be an issue indeed.

In the case where `InodeUser` was versioned, the non optional `inode_size` parameter addition would still have required a major bump, thus already causing a version mismatch between external plugins and the class.

ikelos:
Ok, I'll let it through this time because it's essentially part of the plugin's API, and therefore versioned along with the plugin, however we should have a serious think about how we're defining structures (given the FDUser structure in `lsof` says that it's an `InodeUser` class in the return parameter, when it's actually an FDUser class.  There's a lot flying about at the moment and it feels a little disjointed.  This one can go in, but please can people align the various plugins and their data structures to make sure everything works across plugins no just in their own little world...

PR #1613, opened 2025-02-10
New plugin: linux.pagecache.recoverfs

### LINKED ISSUE #1468: Add Volatility2's `linux_recover_filesystem` capabilities to `linux.pagecache` plugin
It would be awesome to integrate capabilities from https://github.com/volatilityfoundation/volatility/blob/master/volatility/plugins/linux/recover_filesystem.py into the `pagecache` extraction and parsing system. 

The following user parameters could be added as well:
- Decide to replicate file attributes to the exported files **or not**
- Add a general file limit/truncate size, which prevents file outputs or adds a `.trunc` at the end of impacted files on disk 

Ideas:
- Think of a way to handle symlinks, or leave them as is (e.g. `/forensics/vol3_output/dumped_fs/symlink_to_etc_passwd` points to host's `/etc/passwd` instead of `/forensics/vol3_output/dumped_fs/etc/passwd`)

Abyss-W4tcher:
Hi,

This new plugin is a port and improvement over Volatility2's `linux_recover_filesystem` plugin, that allows to automatically extract and save the cached filesystem.

Here, it relies on a tarball instead of writing directly to the host filesystem. Doing so, Volatility3 is exempt from handling long filename errors, or "no disk space left" halfway to the end. It is also OS-agnostic, and more IO efficient as it isn't writing to disk constantly. Columns are the same as `linux.pagecache.files`, with the addition of `Recovered FileSize` at the very end:

```sh
$ python3 vol.py -f sample.bin -r pretty -o test_out/ linux.pagecache.RecoverFs > files.txt
$ cat files.txt
  | SuperblockAddr |               MountPoint | Device |   InodeNum |      InodeAddr | FileType |  InodePages | CachedPages |   FileMode |                     AccessTime |               ModificationTime |                     ChangeTime |                                                                                                                                                                         FilePath |       InodeSize | Recovered FileSize
* | 0x89388c64c800 |                        / |    8:1 |       1240 | 0x893884df0118 |      DIR |           1 |           0 | drwxr-xr-x | 2024-11-15 12:40:54.112001 UTC | 2021-07-12 10:56:03.000000 UTC | 2024-10-02 14:36:01.560177 UTC |                                                                                                                                               /etc/networkd-dispatcher/carrier.d |            4096 |                N/A
* | 0x89388c64c800 |                        / |    8:1 |        515 | 0x893884d38ee0 |      REG |           1 |           1 | -rw-r--r-- | 2024-11-15 12:40:54.668001 UTC | 2022-02-03 05:27:54.000000 UTC | 2024-10-02 14:36:01.521177 UTC |                                                                                                                                                                    /etc/gai.conf |            2584 |               2584
* | 0x89388c64c800 |                        / |    8:1 |       1304 | 0x89388ce78ee0 |      LNK |           1 |           0 | lrwxrwxrwx | 2024-11-15 12:40:54.348001 UTC | 2024-10-02 14:31:56.313182 UTC | 2024-10-02 14:36:01.564177 UTC |                                                                                                                                            /etc/vtrgb -> /etc/alternatives/vtrgb |              23 |                N/A
* | 0x89388c64c800 |                        / |    8:1 |        270 | 0x89388ce7a5d8 |      DIR |           1 |           0 | drwxr-xr-x | 2024-11-15 12:40:53.632001 UTC | 2024-10-02 14:32:21.095297 UTC | 2024-10-02 14:36:01.511177 UTC |                                                                                                                                                                      /etc/cron.d |            4096 |                N/A
* | 0x89388c64c800 |                        / |    8:1 |        271 | 0x893884d273f0 |      REG |           1 |           1 | -rw-r--r-- | 2024-11-15 12:40:53.636001 UTC | 2022-01-08 20:02:36.000000 UTC | 2024-10-02 14:36:01.511177 UTC |                                                                                                                                                          /etc/cron.d/e2scrub_all |             201 |                201
$ ls -l test_out/                
total 64992
-rw------- 1 forensics forensics 61621704 Feb 10 11:32 recovered_fs.tar.gz
```

The files don't even have to be extracted from the generated tarball, as one can leverage the `ratarmount` tool to mount the tarball locally:

```sh
$ ratarmount recovered_fs.tar.gz ./recovered_fs_mounted/
$ ls -l                     
total 64993
drwxrwxrwx 1 forensics forensics        0 Feb 10 11:39 recovered_fs_mounted
-rw------- 1 forensics forensics 61621704 Feb 10 11:32 recovered_fs.tar.gz
-rw-r--r-- 1 forensics forensics  4923392 Feb 10 11:37 recovered_fs.tar.gz.index.sqlite
$ ls -l recovered_fs_mounted/
total 10
# Notice that symlinks are patched:
lr--r--r-- 1 root root 0 Feb 10 11:30 bin -> usr/bin
drwxr-xr-x 1 root root 0 Feb 10 11:30 boot
drwxr-xr-x 1 root root 0 Feb 10 11:30 dev
drwxr-xr-x 1 root root 0 Feb 10 11:30 etc
# Inspect a recovered file content
$ cat recovered_fs_mounted/etc/issue   
Ubuntu 22.04.5 LTS \n \l

```

I hope that makes sense !

Closes #1468.

### DIFF
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -790,13 +790,34 @@ class RecoverFs(plugins.PluginInterface):
 
         visited_paths = seen_prefixes = set()
         for inode_in in inodes_iter:
+
+            # Code is slightly duplicated here with the if-block below.
+            # However this prevents unneeded tar manipulation if fifo
+            # or sock inodes comes through for example.
+            if not (
+                inode_in.inode.is_reg or inode_in.inode.is_dir or inode_in.inode.is_link
+            ):
+                continue
+
+            if not inode_in.path.startswith("/"):
+                vollog.debug(
+                    f'Skipping processing of potentially smeared "{inode_in.path}" inode name as it does not starts with a "/".'
+                )
+                continue
+
+            # Construct the output path
             if uuid_as_prefix:
                 prefix = f"/{inode_in.superblock.uuid}"
             else:
                 prefix = f"/{inode_in.superblock.major}:{inode_in.superblock.minor}"
             prefixed_path = prefix + inode_in.path
 
+            # Sanity check for already processed paths
             if prefixed_path in visited_paths:
+                vollog.log(
+                    constants.LOGLEVEL_VV,
+                    f'Already processed prefixed inode path: "{prefixed_path}".',
+                )
                 continue
             elif prefix not in seen_prefixes:
                 self._tar_add_dir(tar, prefix, mtime)
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -633,11 +633,8 @@ class InodePages(plugins.PluginInterface):
 class RecoverFs(plugins.PluginInterface):
     """Recovers the cached filesystem (directories, files, symlinks) into a compressed tarball.
 
-    Metadata aren't replicated to extracted objects and timestamps are set to the plugin run time. To prevent extraction errors related to long paths, please consider using https://github.com/mxmlnkn/ratarmount.
-    To mount:
-    "ratarmount recovered_fs.tar.gz ./recovered_fs_mounted/".
-    To unmount:
-    "umount ./recovered_fs_mounted/".
+    Details: level 0 directories are named after the UUID of the parent superblock; metadata aren't replicated to extracted objects; objects modification time is set to the plugin run time.
+    Troubleshooting: to fix extraction errors related to long paths, please consider using https://github.com/mxmlnkn/ratarmount.
     """
 
     _version = (1, 0, 0)
@@ -673,6 +670,7 @@ class RecoverFs(plugins.PluginInterface):
         layer_name: str,
         tar: tarfile.TarFile,
         reg_inode_in: InodeInternal,
+        path_prefix: str = "",
         mtime: float = None,
     ) -> int:
         """Extracts a REG inode content and writes it to a TarFile object.
@@ -682,6 +680,7 @@ class RecoverFs(plugins.PluginInterface):
             layer_name: The name of the layer on which to operate
             tar: The TarFile object to write to
             reg_inode_in: The inode to extract content from
+            path_prefix: A custom path prefix to prepend the inode path with
             mtime: The modification time to set the TarInfo object to
 
         Returns:
@@ -694,7 +693,7 @@ class RecoverFs(plugins.PluginInterface):
         inode_content_buffer.seek(0)
         handle_buffer_size = inode_content_buffer.getbuffer().nbytes
 
-        tar_info = tarfile.TarInfo(reg_inode_in.path)
+        tar_info = tarfile.TarInfo(path_prefix + reg_inode_in.path)
         # The tarfile module only has read support for sparse files:
         # https://docs.python.org/3.12/library/tarfile.html#tarfile.LNKTYPE:~:text=and%20longlink%20extensions%2C-,read%2Donly%20support,-for%20all%20variants
         tar_info.type = tarfile.REGTYPE
@@ -707,20 +706,20 @@ class RecoverFs(plugins.PluginInterface):
         return handle_buffer_size
 
     @classmethod
-    def _tar_add_dir_inode(
+    def _tar_add_dir(
         cls,
         tar: tarfile.TarFile,
-        reg_dir_in: InodeInternal,
+        directory_path: str,
         mtime: float = None,
     ) -> None:
         """Adds a directory path to a TarFile object, based on a DIR inode.
 
         Args:
             tar: The TarFile object to write to
-            reg_dir_in: The inode to base the new directory on
+            directory_path: The directory path to create
             mtime: The modification time to set the TarInfo object to
         """
-        tar_info = tarfile.TarInfo(reg_dir_in.path)
+        tar_info = tarfile.TarInfo(directory_path)
         tar_info.type = tarfile.DIRTYPE
         tar_info.mode = 0o755
         if mtime is not None:
@@ -774,11 +773,30 @@ class RecoverFs(plugins.PluginInterface):
             vmlinux_module_name=vmlinux_module_name,
             follow_symlinks=False,
         )
-        visited_paths = set()
+
+        # Prefix paths by the super_block uuid's to prevent overlaps.
+        # Switch to device major and device minor for older kernels (< 2.6.39-rc1).
+        uuid_as_prefix = vmlinux.get_type("super_block").has_member("s_uuid")
+        if not uuid_as_prefix:
+            vollog.warning(
+                "super_block struct does not support s_uuid attribute. Consequently, level 0 directories won't refer to the superblock uuid's, but to its device_major:device_minor numbers."
+            )
+
+        visited_paths = seen_prefixes = set()
         for inode_in in inodes_iter:
-            if inode_in.path in visited_paths:
+            if uuid_as_prefix:
+                prefix = f"/{inode_in.superblock.uuid}"
+            else:
+                prefix = f"/{inode_in.superblock.major}:{inode_in.superblock.minor}"
+            prefixed_path = prefix + inode_in.path
+
+            if prefixed_path in visited_paths:
                 continue
-            visited_paths.add(inode_in.path)
+            elif prefix not in seen_prefixes:
+                self._tar_add_dir(tar, prefix, mtime)
+                seen_prefixes.add(prefix)
+
+            visited_paths.add(prefixed_path)
             extracted_file_size = renderers.NotApplicableValue()
 
             # Inodes parent directory is yielded first, which
@@ -786,10 +804,15 @@ class RecoverFs(plugins.PluginInterface):
             # tarfile will take care of creating it anyway.
             if inode_in.inode.is_reg:
                 extracted_file_size = self._tar_add_reg_inode(
-                    self.context, vmlinux_layer.name, tar, inode_in, mtime
+                    self.context,
+                    vmlinux_layer.name,
+                    tar,
+                    inode_in,
+                    prefix,
+                    mtime,
                 )
             elif inode_in.inode.is_dir:
-                self._tar_add_dir_inode(tar, inode_in, mtime)
+                self._tar_add_dir(tar, prefixed_path, mtime)
             elif (
                 inode_in.inode.is_link
                 and inode_in.inode.has_member("i_link")
@@ -799,7 +822,7 @@ class RecoverFs(plugins.PluginInterface):
                 symlink_dest = inode_in.inode.i_link.dereference().cast(
                     "string", max_length=255, encoding="utf-8", errors="replace"
                 )
-                self._tar_add_lnk(tar, inode_in.path, symlink_dest, mtime)
+                self._tar_add_lnk(tar, prefixed_path, symlink_dest, mtime)
                 # Set path to a user friendly representation before yielding
                 inode_in.path = InodeUser.format_symlink(inode_in.path, symlink_dest)
             else:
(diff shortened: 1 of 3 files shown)


gcmoreira:
It looks really good. Some suggestions:
- In the page cache there are files from multiple devices where the dentry name could be repeated, overwriting the file. Also, the mount point is not warranty since it can be reused for different filesystems at different times. It would be good to be able to choose the device major:minor of the filesystem to dump or somehow differentiate that in the tarball name , i.e. using those ids. 
(Note that the page cache inodes plugin only dumps the first match when the --find is used). 
- I think tar and the tar module support sparse files, check if you need to do something extra to enable that.
Hope that helps

Abyss-W4tcher:
Thanks for pointing this out ! I will investigate to find the most convenient solution for users.

I left a comment in the code regarding sparse files, unfortunately the tarfile module only has read support for sparse files:
https://docs.python.org/3.12/library/tarfile.html#tarfile.LNKTYPE:~:text=and%20longlink%20extensions%2C-,read%2Donly%20support,-for%20all%20variants

Abyss-W4tcher:
@gcmoreira, do you think prefixing paths with device's major + minor be strong enough ?

```sh
$ ls -l recovered_fs_mounted
total 17
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:12
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:19
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:2
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:20
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:21
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:22
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:23
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:24
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:26
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:27
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:28
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:29
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:30
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:31
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:32
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:33
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:34
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:35
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:36
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:37
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:38
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:39
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:4
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:40
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:42
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:45
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:5
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:6
drwxr-xr-x 1 root root 0 Feb 10 20:31 0:7
drwxr-xr-x 1 root root 0 Feb 10 20:31 7:0
drwxr-xr-x 1 root root 0 Feb 10 20:31 7:1
drwxr-xr-x 1 root root 0 Feb 10 20:31 7:2
drwxr-xr-x 1 root root 0 Feb 10 20:31 8:1
```

I also made a version with superblock uuids, but we do not reference them anywhere currently:

```sh
$ ls -l recovered_fs_mounted  
total 5
drwxr-xr-x 1 root root 0 Feb 10 20:17 00000000-0000-0000-0000-000000000000
drwxr-xr-x 1 root root 0 Feb 10 20:17 12a0b472-174e-4504-a652-a2a6c7ff5576
drwxr-xr-x 1 root root 0 Feb 10 20:17 1c5c3546-1f04-42d6-a138-6fef6d520f73
drwxr-xr-x 1 root root 0 Feb 10 20:17 222039a1-8c63-47e6-adbd-22291375a115
drwxr-xr-x 1 root root 0 Feb 10 20:17 546e1837-8607-4619-9fbe-1a9ce6ca3e80
drwxr-xr-x 1 root root 0 Feb 10 20:17 7032fd83-0c63-4d36-9fce-f8caf8e380ad
drwxr-xr-x 1 root root 0 Feb 10 20:17 b0b6fcd7-4f40-4e3b-9266-1fc7aa4407c0
drwxr-xr-x 1 root root 0 Feb 10 20:17 bc350c4f-4c79-4039-98e1-2b6a103ce9fb
drwxr-xr-x 1 root root 0 Feb 10 20:17 c66b2216-6955-41f0-b8b8-5bb4b4feead2
```

gcmoreira:
I think major:minor should be safe but the block id will be safer

Abyss-W4tcher:
Thanks, I decided to go with the `super_block` uuid, with `device_major:device_minor` as a backup for older kernels.

This is ready for review 👍.

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -625,3 +628,234 @@ def run(self):
...
+        tar_info.size = handle_buffer_size
+        tar_info.mode = 0o444
+        if mtime is not None:
+            tar_info.mtime = mtime
+        tar.addfile(tar_info, inode_content_buffer)
+
+        return handle_buffer_size
+
+    @classmethod
+    def _tar_add_dir(
```
ikelos, line 709: These don't appear to be used by anything else?  Why bother exposing them as classmethods?  I've got no real issue with it, it just seems strange?  As private methods they shouldn't be accessed by anyone but the class itself...

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -625,3 +628,234 @@ def run(self):
...
+            tar: The TarFile object to write to
+            symlink_source: The symlink source path
+            symlink_dest: The symlink target/destination
+            mtime: The modification time to set the TarInfo object to
+        """
+        # Patch symlinks pointing to absolute paths,
+        # to prevent referencing the host filesystem.
+        if symlink_dest.startswith("/"):
+            inode_depth = symlink_source.strip("/").count("/")
+            symlink_dest = "../" * inode_depth + symlink_dest.lstrip("/")
```
ikelos: I know this will probably work, I just want to check that there isn't a more recommended way of doing this kind of thing?  Usually slashes and `..`s cause a world of hurt between operating systems.  I know this is all gonna go in a tarball, but I thought I'd ask if this is definitely the best way.  Manual path manipulation often has... unforeseen issues.

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -625,3 +628,234 @@ def run(self):
...
+        uuid_as_prefix = vmlinux.get_type("super_block").has_member("s_uuid")
+        if not uuid_as_prefix:
+            vollog.warning(
+                "super_block struct does not support s_uuid attribute. Consequently, level 0 directories won't refer to the superblock uuid's, but to its device_major:device_minor numbers."
+            )
+
+        visited_paths = seen_prefixes = set()
+        for inode_in in inodes_iter:
+            if uuid_as_prefix:
+                prefix = f"/{inode_in.superblock.uuid}"
```
ikelos, line 810: I'm guessing these don't need slashes at the end, because we can rely on the inode_in.path to *always* start with a slash?

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -625,3 +628,234 @@ def run(self):
...
+                and inode_in.inode.i_link.is_readable()
+            ):
+                symlink_dest = inode_in.inode.i_link.dereference().cast(
+                    "string", max_length=255, encoding="utf-8", errors="replace"
+                )
+                self._tar_add_lnk(tar, prefixed_path, symlink_dest, mtime)
+                # Set path to a user friendly representation before yielding
+                inode_in.path = InodeUser.format_symlink(inode_in.path, symlink_dest)
+            else:
+                continue
```
ikelos, line 856: Probably worth a debug message here saying something was found that didn't match the expected cases.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1057,14 +1058,28 @@ class super_block(objects.StructType):
...
     def major(self) -> int:
         return self.s_dev >> self.MINORBITS
 
-    @property
+    @functools.cached_property
     def minor(self) -> int:
         return self.s_dev & ((1 << self.MINORBITS) - 1)
 
+    @functools.cached_property
+    def uuid(self) -> str:
```
ikelos, line 1070: Ah, hence the need for the PATCH version bump, ok.  Still feel there should be a better way to achieve this, but good work on doing the bump correctly!  Thanks!  5:D

ikelos (commented):
Looks pretty nice, a couple minor points but should be easy to get fixed up.  Thanks!  5:)

volatility3/framework/plugins/linux/pagecache.py
Abyss-W4tcher: Habit I guess, didn't really thought about it. It should be more natural as a basic class indeed.

volatility3/framework/plugins/linux/pagecache.py
Abyss-W4tcher: Yes, this can be tricky to deal with, however here it is more of a pure string manipulation process with one OS-agnostic output format (tarball). I tried introducing `pathlib.PurePath` in the equation but it doesn't understand very well the extra `super_block_uuid` that prefixes the symlink source (like a chroot'ed environment would have implicitely).

This would be a different story if we were directly writing to the host filesystem of course (or writing a web application), but the logic here only calculates the depth and prefixes the absolute destination to make it relative. I don't think we can do something entirely automated, as it does not interact with the FS.

volatility3/framework/plugins/linux/pagecache.py
Abyss-W4tcher: This would go off a lot, as we are skipping sockets, blocks, chars etc. This else block is an implicit blacklist, because some inodes aren't relevant to output in the tarball.

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: gcmoreira and I settled on using the super_block UUID's, as it should prevent overlaps pretty consistently, while preserving this plugin's ease of use. Happy to hear about other approaches however 👍

volatility3/framework/plugins/linux/pagecache.py
Abyss-W4tcher: Each inode should have been sanitized previously, and it is not really the role of this plugin to do the sanitization.

I guess a smear happening right at the start of the inode name could be a thing... Nevertheless, if there is a lot of "slash smear" (broken sample), then a lot of outputted paths will look weirdly concatened, but we can't really detect that.

To be extra safe, I added a small check with a debug message for this edge case.

volatility3/framework/plugins/linux/pagecache.py
Abyss-W4tcher: I'll try to give it another look with `Purepath` to see if we can normalize it at least. 

edit: I was able to make a normalization through `PurePath`, which should be a little more clean.

volatility3/framework/plugins/linux/pagecache.py
```diff
@@ -625,3 +629,260 @@ def run(self):
...
+            tar: The TarFile object to write to
+            symlink_source: The symlink source path
+            symlink_dest: The symlink target/destination
+            symlink_source_prefix: A custom path prefix to prepend the symlink source with
+            mtime: The modification time to set the TarInfo object to
+        """
+        # Patch symlinks pointing to absolute paths,
+        # to prevent referencing the host filesystem.
+        if symlink_dest.startswith("/"):
+            relative_dest = PurePath(symlink_dest).relative_to(PurePath("/"))
```
ikelos, line 749: Since these are builtin types, I think people should realize to import them directly, rather than importing them from this class., so I'll let you off this time but strictly, these should all be `pathlib.PurePath`.

ikelos (approved):

PR #1615, opened 2025-02-12
Windows Envars: Fix unbound locals

dgmcdona:
Technically, these were protected against `UnboundLocalError` exceptions through the use of the `sys` and `ntuser` boolean variables, but this really isn't the best way to prevent that from happening, and type-checkers still warn about the potentially unbound locals. This fix instead uses the `sys` and `ntuser` variables to hold the registry keys, preinitializing them to `None` and checking their value before attempting to access instance methods.


ikelos (approved):
Yeah, that's a reasonable change.  No point using extra variables when it can tell us directly whether they're valid or not...

PR #1616, opened 2025-02-12
Windows Cmdline: Clean up output

dgmcdona:
The strings being used as return values here would (IMO) be better as debug statements, with the plugin returning `renderers.UnreadableValue()` for any of the `InvalidAddressException` code paths.


ikelos:
Sure, that seems reasonable, although if there are complaints I'm also happy to roll it back.  I can't see any issue with it, but as you say it's mostly a preference thing I think.  It's slightly straightforward for a user to see that something's gone wrong, but for those that just want the results, putting text into the output isn't a great idea, so we'll give it a go...  5:)

PR #1617, opened 2025-02-12
Windows Envars: Remove extra config check

dgmcdona:
Tiny bit of cleanup here - there's no need to re-check the config for `SILENT` for each variable. We can just initialize it once as either an empty list or the calculated list of silent vars depending on the config value.


ikelos (approved):
Yes, good spot, thanks  5:)

PR #1619, opened 2025-02-13
Windows FileScan: Add physical offset optional requirement

dgmcdona:
This adds an optional boolean requirement that will allow the plugin to return physical offsets as an alternative to virtual offsets.

### DIFF
--- a/volatility3/framework/plugins/windows/filescan.py
+++ b/volatility3/framework/plugins/windows/filescan.py
@@ -4,7 +4,7 @@
 
 from typing import Iterable
 
-from volatility3.framework import renderers, interfaces, exceptions
+from volatility3.framework import exceptions, interfaces, layers, renderers
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints
 from volatility3.plugins.windows import poolscanner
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/filescan.py
```diff
@@ -60,6 +66,13 @@ def scan_files(
     def _generator(self):
         kernel = self.context.modules[self.config["kernel"]]
 
+        if self.config["physical"]:
+            physical_layer = self.context.layers[kernel.layer_name]
```
ikelos, line 70: I doubt the kernel layer would ever be a physical layer?  I don't think this will ever get triggered properly?  Layer on you also try `physical_layer.mapping` which, if it's a DataLayer, won't work.  I think maybe the naming is off here?

ikelos (commented):
I also slightly wonder whether the physical offset is valuable, and if so why not just always return it?  It's also a point offset, which means that without the full mapping, any data beyond a certain point could come from a completely different place.  What's the drive behind adding this?

dgmcdona:
I found a need for the physical offsets when I was testing something outside of the framework, and figured it would make sense to add since the `windows.dumpfiles` plugin takes a `--physaddr` optional requirement. But I was also using vol3 as a library in that case, so I can just do the mappings myself via the context if we decide that this isn't a good addition.

ikelos:
Well, it'll need a bit more care around the name of the layer and making sure it's actually the physical layer you wanted.  Just keep in mind, it might be coming out of a swap layer or similar, so it'll end up more complicated than just what's in there now.  It's up to you whether you think it's worthwhile, I'll leave the PR open and I'm happy to work with you on improving it, but I'd probably want to see a bit more demand for it before merging it, if that's ok?

dgmcdona:
Closing since I don't see this being a desirable feature any time soon.

PR #1620, opened 2025-02-14
Enforce kernel boundaries correctly. Fix bugs. Closes #1474

### LINKED ISSUE #1474: Orphan kernel threads needs to check process validity before access
Another backtrace triggered in mass testing:

Sample: AdvancedSystemCare.lime.raw

```
    renderer.render(grid)
    grid.populate(visitor, outfd)
  File "/home/ub/volatility3/volatility3/framework/plugins/windows/thrdscan.py", line 100, in _generator
    for ethread in self.implementation(self.context, kernel_name):
  File "/home/ub/volatility3/volatility3/framework/plugins/windows/orphan_kernel_threads.py", line 84, in list_orphan_kernel_threads
    if proc.UniqueProcessId != 4 and proc.InheritedFromUniqueProcessId != 4:
       ^^^^^^^^^^^^^^^^^^^^
```

atcuno:


volatility3/framework/plugins/windows/modules.py
```diff
@@ -127,6 +127,32 @@ def _generator(self):
                 file_output,
             )
 
+    @classmethod
```
ikelos, line 130: This is a new exposed method (an additive change to the API), so that means the version should go to `(2, 1, 0)` rather than `(2, 0, 2)`...

volatility3/framework/plugins/windows/modules.py
```diff
@@ -127,6 +127,32 @@ def _generator(self):
...
+    def get_kernel_space_start(cls, context, layer_name: str, module_name: str) -> int:
```
ikelos: I might be tempted to swap these around, and also, I believe that modules come with an associated layer.  If there's reason to want a different layer to the module, then this is fine, but otherwise I'd just drop the `layer_name` parameter and uses the module's `layer_name` instead?

ikelos (commented):
Looks good save for a couple minor points.  5:)

atcuno:
Did both of those @ikelos

ikelos (approved):
Looks good now, thanks  5:)

PR #1621, opened 2025-02-15
Stop reporting terminated/smeared drivers. Enforce names present. Add…

### LINKED ISSUE #1487: drivermodule reports too many FPs from driverscan
I will make it enforce that at least 1 name is present before reporting a driver.

Sample: Win10x64_22H2_process_hollow_word_T1055.012_2.lime

atcuno:
… kernel start checking in all places

Closes #1487

### DIFF
--- a/volatility3/framework/plugins/windows/driverscan.py
+++ b/volatility3/framework/plugins/windows/driverscan.py
@@ -2,19 +2,19 @@
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
 
-from typing import Iterable
+from typing import Iterable, Optional, Tuple
 
 from volatility3.framework import renderers, interfaces, exceptions
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints
-from volatility3.plugins.windows import poolscanner
+from volatility3.plugins.windows import poolscanner, modules
 
 
 class DriverScan(interfaces.plugins.PluginInterface):
     """Scans for drivers present in a particular windows memory image."""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 0, 0)
+    _version = (2, 0, 0)
 
     @classmethod
     def get_requirements(cls):
@@ -33,8 +33,9 @@ class DriverScan(interfaces.plugins.PluginInterface):
     def scan_drivers(
         cls,
         context: interfaces.context.ContextInterface,
+        kernel_module_name: str,
         layer_name: str,
-        symbol_table: str,
+        symbol_table_name: str,
     ) -> Iterable[interfaces.objects.ObjectInterface]:
         """Scans for drivers using the poolscanner module and constraints.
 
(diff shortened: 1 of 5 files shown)


volatility3/framework/plugins/windows/driverscan.py
```diff
@@ -33,8 +33,9 @@ def get_requirements(cls):
     def scan_drivers(
         cls,
         context: interfaces.context.ContextInterface,
+        kernel_module_name: str,
```
ikelos, line 36: If you're handing in the name of the kernel module, why are you also handing in the layer name and the symbol_table_name?  Will these ever be *different* than `kernel.layer_name` and `kernel.symbol_table_name`?  I'm happy to accept there might be instances where these would differ, but if you're not aware of any or can't think of any in the future, it might be worth reducing this (it can always be changed again in the future if it turns out they do need to be different).

volatility3/framework/plugins/windows/driverscan.py
```diff
@@ -72,37 +101,44 @@ def get_names_for_driver(cls, driver):
...
             yield (
                 0,
                 (
                     format_hints.Hex(driver.vol.offset),
                     format_hints.Hex(driver.DriverStart),
                     format_hints.Hex(driver.DriverSize),
-                    service_key,
-                    driver_name,
-                    name,
+                    service_key or renderers.NotAvailableValue(),
```
ikelos, line 138: Thanks, I think this'll make the classmethods more usable...

volatility3/framework/plugins/windows/modules.py
```diff
@@ -138,16 +138,22 @@ def get_kernel_space_start(cls, context, module_name: str) -> int:
...
 
-        kernel_space_start = module.object(
-            object_type=object_type, offset=range_start_offset
-        )
+        try:
+            kernel_space_start = module.object(
+                object_type=object_type, offset=range_start_offset
+            )
+        except exceptions.InvalidAddressException:
+            kernel_space_start = default_start
```
ikelos, line 159: Could you add a debug log here to say that the actual address couldn't be found and the default is being used, please?

ikelos (commented):
The version number for modules should *probably* change to (2,1,1) to show that something internally is different, but it's not critical.  Otherwise from what I can see scan_drivers is *always* called against the kernel and so splitting out the kernel name, the layer name and the symbol table name is pretty academic (and people don't like to retype stuff) so if possible reduce that given this'll change the API anyway.

Other than that, all looks good.

volatility3/framework/plugins/windows/driverscan.py
atcuno: Great

volatility3/framework/plugins/windows/driverscan.py
atcuno: Fixed

volatility3/framework/plugins/windows/modules.py
atcuno: done

atcuno:
Feedback addressed @ikelos

ikelos (approved):
Ok, all looks good now, thanks!  5:)

PR #1622, opened 2025-02-15
Feature/data renderer

ikelos:
Initial code to get the LayerData renderer in place.  The options and configurability can be made layer, but this gets the important data down.

I threw in a `no_surrounding` flag to say whether it should ever bother printing additional context (sadly, context is confusing within volatility, but if someone has a better name that `no_surrounding` I'm all ears).

### DIFF
--- a/volatility3/framework/plugins/mac/malfind.py
+++ b/volatility3/framework/plugins/mac/malfind.py
@@ -13,7 +13,7 @@ from volatility3.plugins.mac import pslist
 class Malfind(interfaces.plugins.PluginInterface):
     """Lists process memory ranges that potentially contain injected code."""
 
-    _required_framework_version = (2, 0, 0)
+    _required_framework_version = (2, 0, 1)
 
     @classmethod
     def get_requirements(cls):
--- a/volatility3/framework/plugins/mac/malfind.py
+++ b/volatility3/framework/plugins/mac/malfind.py
@@ -68,9 +68,7 @@ class Malfind(interfaces.plugins.PluginInterface):
                 else:
                     architecture = "intel64"
 
-                disasm = interfaces.renderers.Disassembly(
-                    data, vma.links.start, architecture
-                )
+                disasm = renderers.Disassembly(data, vma.links.start, architecture)
 
                 yield (
                     0,
@@ -99,7 +97,7 @@ class Malfind(interfaces.plugins.PluginInterface):
                 ("End", format_hints.Hex),
                 ("Protection", str),
                 ("Hexdump", format_hints.HexBytes),
-                ("Disasm", interfaces.renderers.Disassembly),
+                ("Disasm", renderers.Disassembly),
             ],
             self._generator(
                 list_tasks(self.context, self.config["kernel"], filter_func=filter_func)
--- a/volatility3/framework/plugins/windows/malfind.py
+++ b/volatility3/framework/plugins/windows/malfind.py
@@ -17,7 +17,7 @@ vollog = logging.getLogger(__name__)
 class Malfind(interfaces.plugins.PluginInterface):
     """Lists process memory ranges that potentially contain injected code."""
 
-    _required_framework_version = (2, 4, 0)
+    _required_framework_version = (2, 22, 0)
 
     @classmethod
     def get_requirements(cls):
--- a/volatility3/framework/plugins/windows/malfind.py
+++ b/volatility3/framework/plugins/windows/malfind.py
@@ -182,6 +182,7 @@ class Malfind(interfaces.plugins.PluginInterface):
                     layer_name=proc_layer_name,
                     offset=start,
                     length=length,
+                    no_surrounding=True,
                 )
                 yield (vad, data)
 
--- a/volatility3/framework/plugins/windows/malfind.py
+++ b/volatility3/framework/plugins/windows/malfind.py
@@ -103,7 +103,7 @@ class Malfind(interfaces.plugins.PluginInterface):
         symbol_table: str,
         proc: interfaces.objects.ObjectInterface,
     ) -> Generator[
-        Tuple[interfaces.objects.ObjectInterface, interfaces.renderers.LayerData],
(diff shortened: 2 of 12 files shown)


volatility3/framework/plugins/windows/malfind.py
```diff
@@ -88,6 +88,25 @@ def list_injections(
...
     ) -> Iterable[Tuple[interfaces.objects.ObjectInterface, bytes]]:
+        for vad, data_object in cls.list_injection_sites(
+            context, kernel_layer_name, symbol_table, proc
+        ):
+            yield vad, data_object.context.layers[data_object.layer_name].read(
+                data_object.offset, data_object.length
+            )
+
+    @classmethod
+    def list_injection_sites(
```
dgmcdona, line 100: I think this plugin needs a minor version bump since we're adding a public method

volatility3/framework/plugins/mac/malfind.py
```diff
@@ -99,7 +97,7 @@ def run(self):
                 ("End", format_hints.Hex),
                 ("Protection", str),
                 ("Hexdump", format_hints.HexBytes),
-                ("Disasm", interfaces.renderers.Disassembly),
+                ("Disasm", renderers.Disassembly),
```
dgmcdona, line 100: This needs a patch version bump

volatility3/framework/plugins/windows/mbrscan.py
```diff
@@ -254,8 +256,8 @@ def run(self) -> renderers.TreeGrid:
                     ("EndingCHS", int),
                     ("EndingSector", int),
                     ("SectorInSize", format_hints.Hex),
-                    ("Disasm", interfaces.renderers.Disassembly),
-                    ("Bootcode", format_hints.HexBytes),
+                    ("Disasm", renderers.Disassembly),
+                    ("Bootcode", renderers.LayerData),
```
dgmcdona, line 262: Patch version bump for this?

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -453,7 +453,7 @@ def run(self):
                 ("Record Number", int),
                 ("MFT Type", str),
                 ("Filename", str),
-                ("Hexdump", format_hints.HexBytes),
+                ("Hexdump", renderers.LayerData),
```
dgmcdona, line 456: Patch version bump

dgmcdona (changes requested):
This looks really good! I gave a lot of thought to the text renderer implementation and stress-tested the `context_byte_length` value but couldn't find any issues with it. 

The only thing required in the current set of changes is a double-checking of the required framework version number on the updated plugins + patch version bumps for the plugins themselves. It also looks like `yarascan.YaraScan`, `windows.vadyarascan`, and `linux.vmayarascan` haven't been updated yet. Are we holding off on those for some reason?

ikelos:
Nope, just intending to get it reviewed before I started reworking much else.  I'll get the yara plugins added in and then we should be able to merge it...

ikelos:
Hmmmm, still debating whether to use `__str__` of `__format__` as the default method?  `__str__` technically gets called by the default format string, so it theoretically more useful, I also can't think what format specs we'd want to add to the thing, so I guess they can always come later if necessary.  Should be ready to go then?

dgmcdona:
@ikelos is this pending another review from me or is it good to go?

ikelos:
Well, it's probably good to go, unless you have a view of `__str__` or `__format__`?  I think `str` is more universal if people get used to calling `str` then we can't use format hints to control how it's output (although I guess we shouldn't, it should probably all go through render options anyway).  We still need to figure out how to externally pass it options, which at the moment I'm just passing in an opaque "RenderOption" type to leave my options open but, we really ought to nail that down before it goes live...

dgmcdona:
I was wondering earlier if we should implement `__bytes__` now that we have `gather_contiguous_bytes_at_address` from #1708 . If so, do we suppress the possible `InvalidAddressException` and return an empty byte string if no data can be read?

The question of `__format__` is interesting, there are tons of possibilities, but I don't know how deep we want to go with formatting binary data including possibly writing a custom format spec. It could be a convenient way to provide simple hexdump views of the available data in different endiannesses and widths, or the ability to display disassembly. I can easily see that getting out of hand, though. One upside could be that it would allow users to determine how they want the gap data to be rendered.

For render options, the ones I can think of off the top of my head are:
--before-context  (how many bytes in advance of the layer data to display)
--after-context (how many bytes after the layer data to display)
--placeholder-character (which character should be rendered in place of missing bytes, right now we're using `.` in the `LayerDataRenderer`)

I'm not sure the best way to pass render options. It's defined as a `List[Any]` right now, but does it need to be a list, or could it be something like a dataclass or `TypedDict`?

As you said before, we'll have to think of a different name for the first two since context already has significant meaning within the framework.

ikelos:
Yeah, the `List[Any]` was literally a placeholder, but it seemed reasonable that it would be a list of options.  Anything more complex comes with maintenance and management, but it depends what you had in mind?  It's *very* similar to our configuration system, but the requirements machinery is really complex and clunky and probably overkill for what we're after.  The slight difficulty we have is how to house these things in a tree:

The placeholder char almost feels like a UI choice, rather than something that should be exposed to the user (although the UI could then expose it to the user, so I guess it's the same outcome)?  We also need to think whether different formats and different options (perhaps for JSON you want four surrounding bytes, but for CLI you want none)?  The cell renderers are a level down from the renderers, but also shared between renderers, so the hierarchy we choose will need to be careful to allow as many combinations as possible without making setting them too onerous.

I agree format is probably too tricky and would allow a separate way of passing render options (essentially) because the UI could include things like the surrounding bytes in the format string.  Probably best to steer clear of it.

I don't know that all objects have an `__bytes__` method, but I believe they all have a `__str__` method?  `__bytes__` is very strictly for binary objects, so I don't think it would be useful as a generic override (this is supposed to be a fallback, so that if new cell-renderers are developed in the future, they can still just define `__str__` and they'll get rendered appropriately).

PR #1623, opened 2025-02-16
Layers: Reduce double bounds test and simplify intel translation

### LINKED ISSUE #1618: Issue: Performance Regression in linux.psaux Plugin After Commit c82ef0c
**Describe the bug**  
After updating to commit `c82ef0c` (`"intel: address translation: performance improvements caching by page address"`), the `linux.psaux` plugin has significantly slowed down when running on an Ubuntu machine. The performance degradation was not present in previous commits.  

**Context**  
- **Volatility Version:** c82ef0c
- **Operating System of mem dump:** Ubuntu 22.04
- **Python Version:** python3.10
- **Command:**  
  ```bash
  python3 vol.py -f memory.img linux.psaux
  ```

**To Reproduce**  

1. Use the following command to run `linux.psaux`:  
   ```bash
   python3 vol.py -f memory.img linux.psaux
   ```  
2. Observe the performance slowdown compared to an earlier commit.  

**Expected behavior**  
The `linux.psaux` plugin should perform at least as efficiently as before, with improved or stable processing times.  

**Example output**  
(Provide relevant output logs, preferably using verbose mode for detailed insights.)  

```bash
python3 vol.py -vvv -f memory.img linux.psaux
```

**Additional information**  
- Performance degradation was observed after commit `c82ef0c`.  
- The issue does not appear on earlier commits.  
- Could be related to changes in address translation caching.

ikelos:
This reduces the number of tests introduced by #1587.

Testing for the maximum entry and then maximum page was a waste, and that reduced _translate_entry to just masking by the `page_mask` which we can do at the call sites.  The function therefore continues as normal, but is only ever called with page addresses.  If someone somewhere, somehow manages to call it with a full entry address, that'll be slower, but for standard intel usage, it should give us the same boost that #1587 did, but without additional checks which might introduce slowdown on multiple page translations...

### DIFF
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -159,7 +159,7 @@ class Intel(linear.LinearlyMappedLayer):
         translated address lives in and the layer_name that the address
         lives in
         """
-        entry, position = self._translate_entry(offset)
+        entry, position = self._translate_entry(offset & self.page_mask)
 
         # Now we're done
         if not self._page_is_valid(entry):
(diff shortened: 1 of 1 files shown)


gcmoreira:
Hey @ikelos, I don't see how the changes in this PR would fix the supposed performance issue. Do you think these:
```python
if not (self.minimum_address <= offset <= self.maximum_address):
```
and
```python
if page_address & ~self.page_mask != 0:
```
are slowing down the framework?

ikelos:
If they're being done millions of times for translation, I think they could be a factor?  They also don't seem like they'd affect the output at all?

gcmoreira:
Yes, it should work. Thanks

gcmoreira (approved):

PR #1079, opened 2024-01-09
Linux ip.Addr and ip.Link plugins

gcmoreira:
This PR adds the `linux.ip.Addr` and `linux.ip.Link` plugins. 

### linux.ip.Addr:
 - It displays information about all network interfaces including network namespace, interface index, interface name, mac address, promiscuous status, IPv4/IPv6 address, prefix, scope type and state.

Example output:
```shell
$ python3 ./vol.py -r pretty \
  -f ram-6.2.0-26 \
  linux.ip.Addr
Volatility 3 Framework 2.5.2
Formatting...0.00               Stacking attempts finished                 
  |      NetNS | Index | Interface |               MAC | Promiscuous |                       IP | Prefix | Scope Type |   State
* | 4026531840 |     1 |        lo | 00:00:00:00:00:00 |       False |                127.0.0.1 |      8 |       host | UNKNOWN
* | 4026531840 |     1 |        lo | 00:00:00:00:00:00 |       False |                      ::1 |    128 |       host | UNKNOWN
* | 4026531840 |     2 |     ens32 | 00:0c:29:7f:8b:ab |       False |           172.16.141.130 |     24 |     global |      UP
* | 4026531840 |     2 |     ens32 | 00:0c:29:7f:8b:ab |       False | fe80::20c:29ff:fe7f:8bab |     64 |       link |      UP
* | 4026531840 |     3 |     ens33 | 00:0c:29:7f:8b:b5 |       False |          192.168.249.129 |     24 |     global |      UP
* | 4026531840 |     3 |     ens33 | 00:0c:29:7f:8b:b5 |       False | fe80::20c:29ff:fe7f:8bb5 |     64 |       link |      UP
``` 

### linux.ip.Link (by @eve-mem )
 - It displays information about all network devices configuration including network namespace,  interface name, mac address, state, MTU, Qdisc, Qlen and flags.
```shell
$ python3 ./vol.py -r pretty \
  -f ram-6.2.0-26 \
  linux.ip.Link
Volatility 3 Framework 2.5.2
Formatting...0.00               Stacking attempts finished                 
  |         NS | Interface |               MAC |   State |   MTU |    Qdisc | Qlen |                           Flags
* | 4026531840 |        lo | 00:00:00:00:00:00 | UNKNOWN | 65536 |  noqueue | 1000 |            LOOPBACK,LOWER_UP,UP
[...]

PR #1624, opened 2025-02-16
Feature/linux ifconfig plugin

ikelos:
A continuation of #1079, with fixes.


ikelos:
@Abyss-W4tcher , any chance you could give this a look with an eye to pulling out the IP stuff into its own versionable module please?

Abyss-W4tcher:
Hi, I'll give a look over it tomorrow 👍

ikelos:
Thanks, much appreciated!  5:)

eve-mem:
Thanks for taking the time on this one @ikelos. Really looking forward to this one.

ikelos:
It looked important, just want to make sure we keep LinuxUtilities modular enough to keep the whole thing spritely.  5:)

Abyss-W4tcher:
Quite frankly, I think each child function's context rightfully applies to the type they depend on. The only one that could be split as a generic helper is `net_device._format_as_mac_address`, and even then it is a bit far-fetched in my opinion given that it depends on two internal values implicitely.

Did you have some APIs in mind that would have been eligible to a split ?

ikelos:
Oh, no, I hadn't realized they were all in their own classes, but then why are they all in the `linux/__init__.py`.  It's several thousand lines long, and we should *definitely* be splitting those up into at least a `linux/net.py`...  Sorry, I thought it was similar to the tainting, but it's clearly not.  I'll have a crack at refactoring it at least.  I'll also have a think about how to version the modules so we can change stuff within them without messing up any other linux bits.

Thanks for taking a look into it, sorry for the false alarm (at least my misunderstanding confirmed to me that the file is way too big).  5;)

Abyss-W4tcher:
No problem ! Yeah, I think splitting it in a `extensions/net.py` file is an excellent idea indeed, we could even split everything else in other extensions files ?

ikelos:
Ok, well this got a bit uglier, but I think it's a reasonable way of splitting out (and versioning) the various types.  Unfortunately, to maintain backwards compatibility (and avoid bumping the version up to 3.0.0) I've had to maintain the moved types in the main table, but I've changed all the plugins that made use of it (and used the new types), to explicitly apply the symbol table as needed to the kernel symbol table.  Version checks should be implemented in all the right places, and I think this offers a blueprint for how new  types should be added.

It needs extensive testing, please give the tyres a kick, lemme know if you think the proposed method works and pros/cons of splitting common symbols out of the main linux symbol table, requiring them to be applied separately.  There is no real need at the moment, but updates to these classes will become impossible to version and will cause unexpected changes in the future if they ever change in a non-additive way...

ikelos:
This also highlights why *all* methods should stick to the standard of taking the context and then simple elements (in relation to the changes needed to `SockHandles`).  You never know when you're suddenly going to require extra information and the context should always allow you to get to *everything* you need.  Without it, you're either playing tricks to try to locate a possible context, or you can't make changes that might be necessary to it.

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -2673,7 +2402,9 @@ def get_inode_type(self) -> Union[str, None]:
         else:
             return None
 
-    def _time_member_to_datetime(self, member) -> datetime.datetime:
+    def _time_member_to_datetime(
+        self, member
+    ) -> Union[datetime.datetime, interfaces.renderers.BaseAbsentValue]:
```
gcmoreira, line 2407: I noticed that the renders are now being used in the core API, whereas previously, they were only used in the plugins final stage (mostly in the _generator or around it). Doesn't that seem a bit unusual when the framework is used as a library?

gcmoreira, line 2407: I noticed that renders are now being used in the core API, whereas previously, they were only used in the final stage of plugins (mostly in the _generator). Doesn't that seem a bit unusual when the framework is used as a library?

ikelos:
It absolutely does,

I feel I'll likely have had a conversation about it, but there's only so
much defense I can put up against such things before people think I'm a
tyrant.  5;P  I only added the typing, but I'd prefer not returning
renderer values except when they're to be handed to the UI.  If you can
help clean them up, I'd much appreciate it, but they're in the code base
and not the largest priority at the moment for me at the moment... 5:)

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -3028,7 +2759,9 @@ def get_entries(self) -> Iterable[int]:
 
 
 class rb_root(objects.StructType):
-    def _walk_nodes(self, root_node: int) -> Iterator[int]:
+    def _walk_nodes(
+        self, root_node: interfaces.objects.ObjectInterface
```
gcmoreira, line 2763: `root_node` is a *Pointer* whose base class is **Integer**, and **Integer** itself is a mixin of **int** and **PrimitiveObject -> ObjectInterface**. Is there a specific reason why **ObjectInterface** is preferred here?

I feel that using `int` is more restrictive, helping to catch cases where the method is called with a dereferenced object instead of a **Pointer**. Typically, the dereferenced object is also an **ObjectInterface**, so this restriction could be beneficial whereas making it that generic might go against the goal of type-checking.

gcmoreira (commented):
Hey! Thanks for updating and resurrecting this PR. 
The new additions look great, really excited to see this plugin get merged!

My observations:

The changes from (Linux: Fix ruff errors) eb21f89cf34323ce324b6df64ce0b00e6825a0b8 where s/net/network/ was applied might not be the best fit. I do understand where it's coming from, but `net` isn't just an abbreviation from network, it's the actual type name in the Linux kernel for the network namespace struct https://elixir.bootlin.com/linux/v6.13.3/source/include/net/net_namespace.h#L61 . If we want a (bit) more descriptive name, it's referred to as `net_ns` in some parts of the Linux kernel. However, I wouldn't use `network` since it has a broader connotation and might add confusion rather than clarity. 

Making variable names more expressive isn't always beneficial. For example, [sock](https://elixir.bootlin.com/linux/v6.13.3/source/include/net/sock.h#L342) and [socket](https://elixir.bootlin.com/linux/v6.13.3/source/include/linux/net.h#L117)  are two different types in the kernel, and changing **sock** to **socket** might initially seem like a brilliant idea, but it's actually the opposite.

That being said, I think it's best to keep variable and type names aligned with those in the Linux kernel to improve maintainability. Even if the names aren't perfect, they are still the most easily understood by those familiar with the kernel source. 

On the other hand, with the same first reasoning in mind, shouldn't the two new `net.py` files be named `network.py` instead? same with `NetSymbols`, shouldn't be `NetworkSymbols`?.. and so on

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: never mind, it needs to be a Pointer

volatility3/framework/symbols/linux/extensions/__init__.py
gcmoreira: > It absolutely does,
> 
> I feel I'll likely have had a conversation about it, but there's only so
> much defense I can put up against such things before people think I'm a
> tyrant.  5;P  I only added the typing, but I'd prefer not returning
> renderer values except when they're to be handed to the UI.  If you can
> help clean them up, I'd much appreciate it, but they're in the code base
> and not the largest priority at the moment for me at the moment... 5:)

It was answered there ^^. 
Great, I'm glad we are on the same page with this. Unfortunately, I don't have much time to handle this right now, so I'll leave it to the other contributors. Thanks @ikelos

gcmoreira:
Btw, great teamwork @eve-mem @Abyss-W4tcher @ikelos 👏🏻

ikelos:
I've left it as NetSymbols, but renamed the files to network (and corrected the originally renamed values back to net).

gcmoreira (approved):
It looks excellent, thanks

PR #1626, opened 2025-02-18
Unify variable names

b3nguang:


ikelos:
Thanks very much for your contribution!  The files in the development folder are actually standalone scripts, that aren't part of the main volatility tree.  As such, they have their own logger and don't need to refer to vollog at all.  If it weren't so deeply ingrained in people's minds, I'd probably change vollog to just logger.  5:P  Regardless, thank you for contributing, but to ensure people don't confuse these tools with some designed to be part of the main tree, I'm going to keep the variables name as-is.  I hope that's ok?  Do keep contributing, we'll review all pull requests, so just because this one didn't succeed, don't be afraid to submit other changes!  Thanks...  5:)

PR #1627, opened 2025-02-19
Unified library

b3nguang:


ikelos:
Thanks very much for your submission, we'd actually prefer to use libraries like hashlib (which are part of python) than external dependencies like the Crypto library (pycryptodome).  I've actually added commit 3ea563bd which resolves the discrepancy and now only uses `hashlib`, but thank you for noticing and pointing out the inconsistency!

PR #1628, opened 2025-02-19
windows.threads - fix pid argument

the-rectifier:
Hello,

I found a small issue in the `windows.threads` plugin where it would not correctly grab the `pid` argument.

In `list_process_threads()`, the `pid` argument is passed through the `context` parameter and exists under `plugins.Thread.pid` instead of just `pid` as usual.

Printing `self.config` from `__init__()` will reveal it under the `pid` key as normal, but due to the inheritance, in the `list_process_threads()` it appears under the key mentioned above:

```
[...]
"plugins.Threads.kernel.symbol_table_name.symbol_mask": 0,
"plugins.Threads.pid": [
    4988
],
"plugins.Threads.thrdscan": false,
[...]
```

I don't know if this is the correct way to solve this, but I figured I would post this for future reference if somebody comes across it.

### DIFF
--- a/volatility3/framework/plugins/windows/threads.py
+++ b/volatility3/framework/plugins/windows/threads.py
@@ -74,8 +74,9 @@ class Threads(thrdscan.ThrdScan):
         layer_name = module.layer_name
         symbol_table_name = module.symbol_table_name
 
-        filter_func = pslist.PsList.create_pid_filter(context.config.get("pid", None))
-
+        pid_arg = interfaces.configuration.path_join("plugins", "Threads", "pid")
+        filter_func = pslist.PsList.create_pid_filter(context.config.get(pid_arg), None)
+        
         for proc in pslist.PsList.list_processes(
             context=context,
             layer_name=layer_name,


ikelos:
Thanks for spotting this, the configuration is hierarchical, meaning the plugin can access anything under `plugin.Threads` (but shouldn't directly access the full tree, using `context.config`).  The correct fix is to use `self.config` which exposes the configuration for the threads plugin to the threads plugin, but this is an exported method (a classmethod) so there is no self, which means there's no way of reliably determining the config path to read it from (`plugin.Threads` only works when it's been invoked from the Threads plugin, and not if another plugin launches it with a different configuration tree).  This PR isn't the right way to solve this issue, but thank you for bringing it to our attention, we'll find a way to get it resolved.

@atcuno Sorry, I missed something in #1225 that's now been raised here.  You can't use the context's config to fulfill parameters you need, they need to be part of the function signature for an exported method (which this is).  (This is one of the complexities that makes me stay away from inheritance.)  It looks as though we'll need to alter both the `thrdscan` plugin and the `threads` plugin to support passing in a pid parameter (a parameter that has no bearing on the `thrdscan plugin`) or restore the filter parameter and find a way to pass it in from the instance plugin, rather than the class (which I think previously wasn't used/didn't function).  I dunno if you want to solve this a different way, but if we don't get this resolved soon, the best temporary fix will be to remove the pid parameter and have it run unfiltered (where people would have to rely on grep for shortened results)...

the-rectifier:
Thanks for the info! Feel free to close this whenever. Just a question about (any possible) next time, should I have opened an issue and a PR in parallel? to document the issue if the PR wasn't the solution we were looking for?

ikelos:
I'll convert this across to an issue, so you can follow along with its progress...  Thanks for understanding!

ikelos:
This has now been raised as #1629

Issue #1408, opened 2024-12-16
Volatility keeps broken/partial PDBs, preventing future analysis


atcuno:
I am putting this ticket so it gets the parity label and we deep dive it before the parity release.

We noticed during a training in a place with unstable wifi that:

1) Volatility will start to download a PDB upon plugin run/automagic
2) It seemingly cannot download the whole file due to the wifi cutting out
3) An exception is thrown related to an index error when parsing the PDB
4) It (vol3) keeps the broken PDB around for future analysis and breaks running plugins indefinitely against samples matching that PDB

We are going to simulate this by turning an interface off in a VM while the PDB is being downloaded, likely with sleep() calls to make sure it happens on time.

After diagnosing it further then we can figure out what the fix should look like.

This happened to almost every student on the first day across samples so we eventually distributed known-good sets of symbol tables so they could do the labs.

ikelos:
Yeah, we stash everything we receive directly into a cache file.  I guess we could either mark is as temp until we know it all downloaded (although that doesn't cope with Microsoft handing us bad files) or we could just not use the caching mechanism, since it's supposed to be turned into a JSON anyway?  I don't know what happens if you get the PDB fine, but then interrupt the JSON production whether it still writes a partial JSON?  I don't *think* it should, but that's something you should test too.  How we cope with a partial JSON I'm not sure...

gcmoreira:
It seems the PDB document table contains a hash that we can use to check integrity.

https://github.com/dotnet/runtime/blob/main/docs/design/specs/PortablePdb-Metadata.md#document-table-0x30

```
Document Table: 0x30

The Document table has the following columns:

    Name (Blob heap index of [document name blob](https://github.com/dotnet/runtime/blob/main/docs/design/specs/PortablePdb-Metadata.md#document-name-blob))
    HashAlgorithm (Guid heap index)
    Hash (Blob heap index)
    Language (Guid heap index)

The table is not required to be sorted.

There shall be no duplicate rows in the Document table, based upon document name.

Name shall not be nil. It can however encode an empty name string.

Hash is the file content hashed using the specified HashAlgorithm. It is used to validate that a source file matches the one used by the compiler when compiling the source code.
```

ikelos:
[...]

PR #1631, opened 2025-02-23
Issues/issue1408 - Delete partial cache downloads

ikelos:
This should ensure partial downloads are deleted rather than causing future failures.

Caveats:
 * We don't try to continue partial downloads
 * If the server being downloaded from doesn't respond with the content-length field, the file is always cached
 * This does nothing to prevent bad files living in the cache, since the framework is effectively unaware of the caching and the caching is unaware of the filetype, there isn't easily a way to delete the cached file, if it's found (by the framework) to be invalid.

So this is a stop-gap that should alleviate the primary symptoms, but still isn't ideal by a fair way.  Ideally someone would sit and work through the opener and rewrite it (so that automatic decompression and caching all still work correctly, but with better control over intermediate files).

### DIFF
--- a/volatility3/framework/constants/__init__.py
+++ b/volatility3/framework/constants/__init__.py
@@ -120,6 +120,9 @@ OFFLINE = False
 REMOTE_ISF_URL = None  # 'http://localhost:8000/banners.json'
 """Remote URL to query for a list of ISF addresses"""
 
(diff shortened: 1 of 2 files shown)


ikelos:
This improves, but probably doesn't fix #1408

PR #1632, opened 2025-02-23
Check Win10+ SlushSize member to support Windows 11 pool scanning. Re…

atcuno:
…strict the Pools that current scanners accept.

This PR fixes two major bugs in Volatility 3 that I ran into while trying to add the GUI/win32k.sys support.

1) Pool scanning on Windows 11 was completely broken as the `PoolType` member of big pools that the type check relied on is now gone (meaning it existed from Vista->later Windows 10), but does not exist on XP or Windows 11. To remedy this, I added an `or` check to also include `SlushSize` as this member was added in Windows 10 and is still present in 11. I verified scanners work again with this fix.

2) The first issue and potentially others (still investigating) were hidden by the fact that all the scanners listed that objects could come from every possible pool (paged, non paged, free) - which does not make any sense. The acceptable pool per type is now included for each type.

### DIFF
--- a/volatility3/framework/plugins/windows/poolscanner.py
+++ b/volatility3/framework/plugins/windows/poolscanner.py
@@ -205,6 +205,7 @@ class PoolScanner(plugins.PluginInterface):
                 b"AtmT",
                 type_name=symbol_table + constants.BANG + "_RTL_ATOM_TABLE",
                 size=(200, None),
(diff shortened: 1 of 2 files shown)


ikelos:
Seems reasonable, but I really don't know what caused the choice of `PAGED` or `NONPAGED` was made for most of the options.  If you're happy it's all correct, I'm fine with it too...

PR #1635, opened 2025-02-25
Linux: Fix typing is lsof

### LINKED ISSUE #1443: fd_fields bad typing breaks type checking
The second element is listed as `int`:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/lsof.py#L57

But it is really the `file` object for the file:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/linux/lsof.py#L70

@gcmoreira Can you double check this and fix as needed?

ikelos:
Fixes #1443


atcuno:
It looks good.

PR #1652, opened 2025-03-01
Add complete smear protection to linux.pstree

atcuno:
This plugin was infinitely looping in yesterday's mass test.

When reviewing the code to fix the infinite looping, I found the plugin to be very strange - it was reporting swapper and also it had a `seen` variable being checked but never added to... It also only had seen in one path.

I added into all the paths, added smeared protection for processes with a ppid of 0, and stopped swapper from showing up in the tree (its not a real process and doesn't show in any Linux interface).

### DIFF
--- a/volatility3/framework/plugins/linux/pstree.py
+++ b/volatility3/framework/plugins/linux/pstree.py
@@ -52,19 +52,38 @@ class PsTree(interfaces.plugins.PluginInterface):
         Args:
             pid: PID to find the level in the hierarchy
         """
-        seen = set([pid])
+        seen_ppids = set()
+        seen_offsets = set()
+
         level = 0
         proc = self._tasks.get(pid)
-        while proc and proc.get_parent_pid() not in seen:
+
+        while proc:
+            # we don't want swapper in the tree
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/linux/pstree.py
```diff
@@ -52,19 +52,38 @@ def find_level(self, pid: int) -> None:
...
             else:
                 parent_pid = proc.tgid
 
+            if parent_pid in seen_ppids or proc.vol.offset in seen_offsets:
+                break
+
+            # only pid 1 (init/systemd) or 2 (kthreadd) should have swapper as a parent
+            # any other process with a ppid of 0 is smeared or terminated
+            if parent_pid == 0 and proc.pid > 2:
+                break
```
ikelos, line 84: Worth a debugging (not normally visible) message for us to figure out why they're not output, or common enough to be bothersome?

ikelos (commented):
One small comment, otherwise looks fine.

atcuno:
Feedback addressed @ikelos

ikelos (approved):

PR #1653, opened 2025-03-01
Fix crash-causing bugs. Add typing where possible.

atcuno:
kmsg crashed in several place across samples. The plugin is very kernel-version dependent, so these occurred in the handling of the kernel-range specific issues.

I re-ran the plugins across all samples after these changes and no more crashes.


ikelos (approved):
Stictly we should be versioning ABCKmsg (although I don't think any of the changes made actually required a bump, but either way) but I'm willing to leave that for another time...

PR #1655, opened 2025-03-01
Convert get_net_namespace_id to the proper convention

atcuno:
This function was throwing an exception on smeared pointers even though its a `get_*`. It now returns None.


atcuno:
This was named poorly and had two commits instead of one

PR #1656, opened 2025-03-01
Fix broken code and APIs in Linux networking paths

atcuno:
The Linux networking paths that supported the two `ip` plugins had many issues, including a `get_*` throwing a paging exception, not catching exceptions for reading names, and other issues. These caused backtraces all over in testing. These are now fixed.

### DIFF
--- a/volatility3/framework/plugins/linux/ip.py
+++ b/volatility3/framework/plugins/linux/ip.py
@@ -39,7 +39,7 @@ class Addr(plugins.PluginInterface):
         try:
             net_ns_id = net_dev.get_net_namespace_id()
         except AttributeError:
-            net_ns_id = renderers.NotAvailableValue()
+            net_ns_id = None
 
         # Interface IPv4 Addresses
         in_device = net_dev.ip_ptr.dereference().cast("in_device")
@@ -47,7 +47,7 @@ class Addr(plugins.PluginInterface):
             prefix_len = in_ifaddr.get_prefix_len()
             scope_type = in_ifaddr.get_scope_type()
             ip_addr = in_ifaddr.get_address()
-            yield net_ns_id or renderers.NotAvailableValue(), iface_ifindex, iface_name, mac_addr, promisc, ip_addr, prefix_len, scope_type, operational_state
+            yield net_ns_id, iface_ifindex, iface_name, mac_addr, promisc, ip_addr, prefix_len, scope_type, operational_state
 
         # Interface IPv6 Addresses
         inet6_dev = net_dev.ip6_ptr.dereference().cast("inet6_dev")
@@ -55,9 +55,9 @@ class Addr(plugins.PluginInterface):
             prefix_len = inet6_ifaddr.get_prefix_len()
             scope_type = inet6_ifaddr.get_scope_type()
             ip6_addr = inet6_ifaddr.get_address()
-            yield net_ns_id or renderers.NotAvailableValue(), iface_ifindex, iface_name, mac_addr, promisc, ip6_addr, prefix_len, scope_type, operational_state
+            yield net_ns_id, iface_ifindex, iface_name, mac_addr, promisc, ip6_addr, prefix_len, scope_type, operational_state
 
-    def _generator(self):
+    def _enumerate_net_namespace_list(self):
         vmlinux = self.context.modules[self.config["kernel"]]
 
         net_type_symname = vmlinux.symbol_table_name + constants.BANG + "net"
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/linux/ip.py
```diff
@@ -47,15 +47,15 @@ def _gather_net_dev_info(self, net_dev: net_extensions.net_device):
...
+            yield net_ns_id or renderers.NotAvailableValue(), iface_ifindex, iface_name, mac_addr, promisc, ip_addr, prefix_len, scope_type, operational_state
 
         # Interface IPv6 Addresses
         inet6_dev = net_dev.ip6_ptr.dereference().cast("inet6_dev")
         for inet6_ifaddr in inet6_dev.get_addresses():
             prefix_len = inet6_ifaddr.get_prefix_len()
             scope_type = inet6_ifaddr.get_scope_type()
             ip6_addr = inet6_ifaddr.get_address()
-            yield net_ns_id, iface_ifindex, iface_name, mac_addr, promisc, ip6_addr, prefix_len, scope_type, operational_state
+            yield net_ns_id or renderers.NotAvailableValue(), iface_ifindex, iface_name, mac_addr, promisc, ip6_addr, prefix_len, scope_type, operational_state
```
ikelos: I thought we were favouring returning `None` and then at render time doing `or NotAvailableValue()`?

volatility3/framework/symbols/linux/extensions/network.py
```diff
@@ -199,7 +199,7 @@ def promisc(self) -> bool:
         """
         return self.flags & self._get_net_device_flag_value("IFF_PROMISC") != 0
 
-    def get_net_namespace_id(self) -> int:
+    def _do_get_net_namespace_id(self) -> int:
```
ikelos, line 202: Is it really worth keeping the old one around?  No one calls it directly, so might just as well as the exception check into the main body?

ikelos (commented):
Looks ok, just want to check on returning renderer values in code callable anywhere in the framework (convention is to leave it as None until the moment it's going to be rendered).  And also using a facade to keep an old function around for little reason...  The old code's never gonna get called, so it's just adding an extra hidden function with no value...

volatility3/framework/symbols/linux/extensions/network.py
atcuno: I like the flow of letting a `_do_*` function do the work then checking its return value (or exception handling usually in Vol3) in the wrapper. Its a convention the Linux kernel follows and keeps the code nice and clean. Its also nice as we aren't changing public APIs that exist already to do this.

volatility3/framework/plugins/linux/ip.py
atcuno: I left this as acceptable here b/c the function is just gathering the data directly sent through the generator. If we make it return None in the line you highlighted then this line in the generator (this is a link to the full file) would need to unpack all those fields (10 or so) versus just passing it through:

https://github.com/volatilityfoundation/volatility3/blob/6a9a7bc1c6b0b1c5b90c03aec5b71ab7ab39dd23/volatility3/framework/plugins/linux/ip.py#L71

atcuno:
@ikelos for this one I left you replies to both your comments.

volatility3/framework/plugins/linux/ip.py
ikelos: Mmmmm, not sure how I feel about that.  It's a good reason but given this is going to be reused a *lot*, I'd prefer we did the unpacking (once) rather than letting the RendererValues sneak further into the API.  It's much harder to convince people not to do this stuff, if they can point to a core example doing it...  5:S  It's a little more work now, but it'll keep everything future and dependable in the future.  Could you make the changes please?

volatility3/framework/symbols/linux/extensions/network.py
ikelos: Ok, I don't have any great stakes in the game so if you want to do it for linux that's fine but it'll expand the API file quite a bit when looking for methods (you'll have to bounce through this facade every time you want to debug it)...

volatility3/framework/plugins/linux/ip.py
atcuno: Updated.

atcuno:
This is ready again @ikelos

volatility3/framework/plugins/linux/ip.py
```diff
@@ -127,7 +150,7 @@ def _gather_net_dev_link_info(self, net_device):
         ]
         flags_str = ",".join(flags_list)
 
-        yield net_ns_id, iface_name, mac_addr, operational_state, mtu, qdisc_name, qlen, flags_str
+        yield net_ns_id or renderers.NotAvailableValue(), iface_name, mac_addr, operational_state, mtu, qdisc_name or renderers.NotAvailableValue(), qlen, flags_str
```
ikelos, line 153: I mean, this is technically an internal function (because it's not a `classmethod`) so I'll let you off, but if people try and copy this for classmethods, I will point them at this comment and probably shake my finger and tut...  5:P

volatility3/framework/symbols/linux/extensions/network.py
```diff
@@ -247,17 +265,34 @@ def get_queue_length(self) -> int:
 
 class in_device(objects.StructType):
     def get_addresses(
-        self,
+        self, max_devices=128
```
ikelos, line 268: Hmmmm, technically we should bump the minor version so people know when this parameter was introduced, but I'm willing to let it slide in this case.  It's unlikely anyone trying to use it will be having their plugin used with an older version of the framework...

ikelos (approved):
A couple little comments, but nothing preventing this from being merged...

PR #1659, opened 2025-03-03
chore: Pin GitHub Actions to immutable SHA hashes

wroersma:
# chore: Pin GitHub Actions to immutable SHA hashes

## Summary
This PR updates GitHub Actions to pin them to specific SHA hashes rather than version tags to mitigate supply chain attack risks.

## Security Rationale
Using version tags like `@v4` creates a security vulnerability as the content behind those tags can be modified at any time by action maintainers or by attackers who compromise their accounts. Recent supply chain attacks like the xz vulnerability (CVE-2024-3094) demonstrate how dependencies can be compromised through trusted distribution mechanisms.

By pinning to immutable SHA hashes, we ensure that the exact code run by our workflows is never silently changed. This follows GitHub's own security best practices: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

## Changes
This PR updates multiple GitHub Actions to their latest versions, including:

* actions/checkout from v4 to v4.2.2 (11bd71901bbe5b1630ceea73d27597364c9af683)
* actions/setup-python from v4/v5 to v5.4.0 (42375524e23c412d93fb67b49958b491fce71c38)
* actions/upload-artifact from v4 to v4.6.1 (4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1)
* astral-sh/ruff-action from v1 to v3.2.1 (39f75e526a505e26a302f8796977b50c13720edf)
* actions/stale from v5 to v9.1.0 (5bef64f19d7facfb25b37b414482c7164d639639)

## Notable Security Improvements
* All actions are now pinned to specific SHA hashes instead of version tags
* Updates include security patches from newer versions
* Eliminates the risk of a compromised action repository affecting CI/CD security

## Testing
All workflows have been validated to ensure they continue to function with the updated action versions.

🔒 This PR was created to enhance the security posture of the Volatility 3 GitHub workflow pipeline.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -6,7 +6,7 @@ jobs:
   lint:
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
       - uses: psf/black@stable
         with:
           options: "--check --diff --verbose"
--- a/.github/workflows/build-pyinstaller.yml
+++ b/.github/workflows/build-pyinstaller.yml
@@ -18,9 +18,9 @@ jobs:
       matrix:
         python-version: ["3.11"]
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38  # v5.4.0
       with:
         python-version: ${{ matrix.python-version }}
 
@@ -40,7 +40,7 @@ jobs:
         mv dist/volshell.exe volshell.exe
 
     - name: Archive 
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1  # v4.6.1
       with:
         name: volatility3-pyinstaller
         path: |
--- a/.github/workflows/build-pypi.yml
+++ b/.github/workflows/build-pypi.yml
@@ -20,9 +20,9 @@ jobs:
       matrix:
         python-version: ["3.8"]
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38  # v5.4.0
       with:
         python-version: ${{ matrix.python-version }}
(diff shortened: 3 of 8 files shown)


ikelos:
This also decreases the ability for us to benefit from fixes applied to tagged versions of the code by legitimate upstream when they're made?  The example of the xz vulnerability could still affect us by one of our python dependencies (such as pefile) or by someone ingratiating themselves with the team and eventually getting direct commit access.  The point is, the risk of that happening is much lower than a security bug being legitimately fixed and us missing it because we're using a SHA hash.  Blocking ourselves for all future updates because we're paranoid seems like worse security rather than better.

The guidance you pointed to is for *third-party* actions, which at the moment is only ruff-action.

If someone wanted in to our repository, I'm sure there are much easier ways than compromising github actions to achieve it, so I'm going to close this PR.  Thanks for your consideration but having thought about it, the detriments outweigh the benefits for our project.

wroersma:
Thanks for the thoughtful feedback, @ikelos. I understand your concerns about missing important security updates when pinning to specific SHA hashes.

I'd like to offer a potential middle-ground solution: I've developed a GitHub Action specifically to address this issue - [[ThreatFlux/githubWorkFlowChecker](https://github.com/ThreatFlux/githubWorkFlowChecker)]

This tool:
1. Can be scheduled to run periodically (like dependabot)
2. Automatically checks for updates to pinned GitHub Actions
3. Creates PRs when updates are available, allowing you to review and approve changes
4. Can also be used as a CLI tool for manual update checks

This way, you could get the security benefits of SHA pinning (protection from supply chain attacks) while still having a reliable mechanism to receive and review important security updates. You wouldn't miss fixes for vulnerabilities, but you'd retain control over when those updates are applied.

An important security consideration: While GitHub recommends SHA pinning for third-party actions, it's worth noting that even GitHub's first-party actions aren't immune to supply chain attacks. If a vulnerability like the xz package issue (or a similar compromise) affected any dependencies used by GitHub's core actions, pinning would protect you from automatically inheriting that vulnerability. The distinction between first-party and third-party actions becomes less relevant when considering that both ultimately depend on complex supply chains that could be compromised.

If this approach interests you, I'd be happy to submit a new PR that implements both the SHA pinning and this update automation. Just let me know!

ikelos:
Thanks for your quick response!  Notifications about updates does sound useful (and we should probably get something like that anyway, given ruff-action has already gone up to tag versions, as has stale.

I still think it's overkill for our project, because I think our individual developers would be much easier targets for someone that specifically wanted to take over our tool/development.  This would likely just add an extra layer of us being notified of an update, and then updating it (which would offer a chance of spotting a security issue, but at the cost of more manual maintenance and no guarantee our checking will be enough to spot an issue, after all the XZ vulnerability made it out into the world and was only spotted by a very specific timing issue).  As such, while I appreciate your offer and taking the time to examine the details with me, I think I'll still pass (although happy for others on the project to suggest otherwise if they'd like).

If you'd be happy to submit a PR for something that checks our actions and alerts us when they're out of date, I'd be happy to look at pulling that on board.  It feels like it would have a much more beneficial direct impact...  5:)

PR #1663, opened 2025-03-05
Fix several bugs in IDT verification for Linux

atcuno:
This PR fixes several bugs in the linux.check_idt plugin:

1) Old/broken symbol tables will be missing the required type. The code previously would fall through a default type that did not exist

2) There was a member check for `Address`, which is an extension method not even implemented in Volatility 3.

3) Incorrect handling to detect smeared table entries.

### DIFF
--- a/volatility3/framework/plugins/linux/check_idt.py
+++ b/volatility3/framework/plugins/linux/check_idt.py
@@ -3,7 +3,7 @@
 #
 
 import logging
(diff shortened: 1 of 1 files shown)


ikelos (approved):
Looks ok to me.  Wondering whether the `a`/`b` stuff would be better handled by a class, but that might be excessive....

PR #1665, opened 2025-03-05
Prevent backtraces in netfilter due to smear

atcuno:


ikelos (approved):
Looks good, I was worried it wasn't versioned but I don't think any of its changes require a bump, so should be ok.

PR #1671, opened 2025-03-06
Fix several bugs found in the tracing plugins during mass testing

atcuno:
This fixes several bugs found during mass testing:

1) a vollog.warning instead of .debug flooded the screen when analyzing infected samples even though the plugin reported - (dashes), which is the actual rootkit marker

2) Missing smear checks

3) Expanded version support by using member checks and then getting alternates

4) A huge backtrace on symbol missing instead of error() and cleanly bail.

### DIFF
--- a/volatility3/framework/plugins/linux/tracing/ftrace.py
+++ b/volatility3/framework/plugins/linux/tracing/ftrace.py
@@ -5,7 +5,7 @@
 # Public researches: https://i.blackhat.com/USA21/Wednesday-Handouts/us-21-Fixing-A-Memory-Forensics-Blind-Spot-Linux-Kernel-Tracing-wp.pdf
 
 import logging
-from typing import Dict, List, Iterable, Optional
+from typing import Dict, List, Generator
 from enum import Enum
 from dataclasses import dataclass
 
@@ -67,7 +67,7 @@ class CheckFtrace(interfaces.plugins.PluginInterface):
     Investigate the ftrace infrastructure to uncover kernel attached callbacks, which can be leveraged
     to hook kernel functions and modify their behaviour."""
 
-    _version = (1, 0, 0)
+    _version = (2, 0, 0)
     _required_framework_version = (2, 19, 0)
 
     @classmethod
@@ -103,14 +103,14 @@ class CheckFtrace(interfaces.plugins.PluginInterface):
     def extract_hash_table_filters(
         cls,
         ftrace_ops: interfaces.objects.ObjectInterface,
-    ) -> Optional[Iterable[interfaces.objects.ObjectInterface]]:
+    ) -> Generator[interfaces.objects.ObjectInterface, None, None]:
         """Wrap the process of walking to every ftrace_func_entry of an ftrace_ops.
         Those are stored in a hash table of filters that indicates the addresses hooked.
 
         Args:
             ftrace_ops: The ftrace_ops struct to walk through
 
-        Returns:
+        Return, None, None:
             An iterable of ftrace_func_entry structs
         """
 
@@ -140,7 +140,7 @@ class CheckFtrace(interfaces.plugins.PluginInterface):
         known_modules: Dict[str, List[extensions.module]],
         ftrace_ops: interfaces.objects.ObjectInterface,
         run_hidden_modules: bool = True,
-    ) -> Optional[Iterable[ParsedFtraceOps]]:
+    ) -> Generator[ParsedFtraceOps, None, None]:
         """Parse an ftrace_ops struct to highlight ftrace kernel hooking.
         Iterates over embedded ftrace_func_entry entries, which point to hooked memory areas.
 
@@ -237,12 +237,10 @@ if the "hidden_modules" key is present in known_modules.
                 formatted_ftrace_flags,
             )
 
-        return None
-
     @classmethod
     def iterate_ftrace_ops_list(
         cls, context: interfaces.context.ContextInterface, kernel_name: str
-    ) -> Optional[Iterable[interfaces.objects.ObjectInterface]]:
+    ) -> Generator[interfaces.objects.ObjectInterface, None, None]:
         """Iterate over (ftrace_ops *)ftrace_ops_list.
 
         Returns:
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/linux/tracing/tracepoints.py
```diff
@@ -183,16 +190,21 @@ def parse_tracepoint(
...
+                prio = renderers.NotAvailableValue()
+
             yield ParsedTracepointFunc(
-                utility.pointer_to_string(tracepoint.name, count=512),
+                tracepoint_name,
                 tracepoint.vol.offset,
                 probe_handler_symbol,
                 probe_handler_address,
-                tracepoint_func.prio,
+                prio,
```
ikelos, line 207: Should this be `None`, and then `NotAvailableValue` by used just before rendering (only issue would be if prio could be 0, although `prio if prio is not None else NotAvailableValue()` would work)?  APIs should generally work with standard python options if possible...

volatility3/framework/plugins/linux/tracing/tracepoints.py
```diff
@@ -258,11 +270,11 @@ def _generator(self):
...
         if not kernel.has_symbol("__start___tracepoints_ptrs"):
-            raise exceptions.SymbolError(
-                "__start___tracepoints_ptrs",
-                self.vmlinux.symbol_table_name,
-                'The provided symbol table does not include the "__start___tracepoints_ptrs" symbol. This means you are either analyzing an unsupported kernel version or that your symbol table is corrupted.',
+            vollog.error(
+                'The provided symbol table does not include the "__start___tracepoints_ptrs" symbol.'
+                "This means you are either analyzing an unsupported kernel version or that your symbol table is corrupted."
             )
+            return
```
ikelos, line 277: Should we not still raise a symbol error, or do you want to continue after this?

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -114,21 +114,24 @@ def extract_hash_table_filters(
...
         try:
-            current_bucket_ptr = ftrace_ops.func_hash.filter_hash.buckets.first
+            current_bucket_ptr = ftrace_hash.buckets.first
         except exceptions.InvalidAddressException:
             vollog.log(
                 constants.LOGLEVEL_VV,
                 f"ftrace_func_entry list of ftrace_ops@{ftrace_ops.vol.offset:#x} is empty/invalid. Skipping it...",
             )
-            return []
+            return
```
ikelos, line 129: Because of the yield statements, I think it's ok if the type signature for this isn't `Optional` (although I'm not sure how the interpreter would know that)...?

ikelos (commented):
Just a couple of style questsions, but generally all looks good.

volatility3/framework/plugins/linux/tracing/tracepoints.py
atcuno: This is in the generator, so we want to report hard errors to the user. The symbol errors being raised has been for when plugins can into APIs. Raising symbol error in generator causes a huge backtrace instead of an orderly error message.

ikelos: Generally the way to do that is to catch the exception in the caller and report it then?  This way the caller has to do a check for a `None` response?  I'm kinda happy either way, but strictly the python way should be to throw the exception and catch it (so that other callers know something's really gone wrong, rather than just being able to sweep it under the rug)...

volatility3/framework/plugins/linux/tracing/ftrace.py
atcuno: I overhauled the types here and did a version bump. We have other places in the framework that also have Iterable instead of Generator in the typing, but will leave that for later.

volatility3/framework/plugins/linux/tracing/tracepoints.py
atcuno: This is in the generator though, so there is nothing to catch it, which is why we get the huge backtrace if we raise a symbolerror.

volatility3/framework/plugins/linux/tracing/tracepoints.py
atcuno: good catch

volatility3/framework/plugins/linux/tracing/tracepoints.py
ikelos: Oh, I see, sorry, sometimes it's hard to spot what's what in the github UI.

volatility3/framework/plugins/linux/tracing/ftrace.py
ikelos: Sure, no problem, just wondered whether it needed to be optional, or would just return an empty iterable/generator?

atcuno: The push I just made fixes the typing, it shound't have optional or iterable on generators.

ikelos:
Cool, looks good thanks.  At some point we should decide whether we prefer `Iterable` or `Generator` to be returned?  I think in most cases just `Iterable` encompasses more but it depends whether they're roughly interchangeable or not...?

atcuno:
> Cool, looks good thanks. At some point we should decide whether we prefer `Iterable` or `Generator` to be returned? I think in most cases just `Iterable` encompasses more but it depends whether they're roughly interchangeable or not...?

Generators convey directly that the function uses `yield`, which tells you a few things, but for Vol3 purposes largely means that you can only walk the list once as its lazy evaluated versus a regular list [], where you can walk as many times as you want. Iterable also has a meaning that `__iter__` is supported plus a few other things, like next(), that we don't really use in Vol3 backend code.

ikelos:
No, but an `Iterable` *can* be a `list` or a `Generator`, whereas a `Generator` can only be a `Generator`, but both of them can only be walked once (I think)?  So figured it would give us room to make changes in the future, and we don't tend to use any property of Generators (feeding in values to the process, etc) that makes them different to an `Iterable`?  As I say, one we can decide in the future, but probably should.  Then we can throw David's tree-sitter stuff at it at some point I imagine...  5;P

PR #1672, opened 2025-03-06
Pdbconv: Make symbol server URL constant

dgmcdona:
This removes the hard-coded symbol server URL from `PdbRetreiver.retrieve_pdb` and defines it as a constant within the `constants` module.


ikelos (approved):
Yep, perfectly happy with this, good spot!  Thanks!  5:D

PR #1673, opened 2025-03-06
Merge two Linux APIs for mapping kernel modules to pointers. Fix bugs…

atcuno:
… in new API. Add deprecation warnings throughout old API.

This PR fixes a serious set of issues in the Linux APIs related to mapping kernel pointers to their module and symbol name (a critical aspect of rootkit detection). Previously, I had written an API in the `symbols/linux/__init__.py` that performed these tasks and were very easy for plugin writers to use - you called a function to get the handler information then you called a second function anytime you needed to map an address inside the handlers.

There was a recent push for covering the tracing APIs (ftrace and tracepoints) that added an entirely new API for mapping modules and symbols, but this work had three major issues that were not handled in attempting to replicate my old API, but with new features added:

1) This API did not incorporate the kernel .text space in its "modules" set, which meant that every pointer inside the kernel was reported as a dash (rootkit marker) in plugins. In the mass testing runs, this led to every single sample reporting rootkits all over the place.

2) This API did not directly incorporate the symbol lookups, leading to a bunch of copy/paste code between plugins that wanted to use it - see the diffs in this PR for ftrace and tracepoint - now they match the easy call setup from my old API, but also incorporates the hidden module searching.

3) The PRs for the new API left the old API in place, plus many existing plugins using the old API. This left Volatility with two public APIs for mapping Linux kernel pointers, neither of which were complete. The updated API is unified and gets the best features of each (hidden modules, mapping inside kernel .text), while leaving it very easy on plugin writers.

This PR fixes all the above issues, with the minimal amount of code possible. To rectify all the issues, and get it to where all the plugins will use new API, I did the following:

1) Deprecated all the old code paths

2) Bumped the Linux utilities version on the plugins using the old paths, but left them still using the deprecated paths to minimze the size of this PR. In a PR right after this one is merged, I will update them to use the new paths to avoid the deprecation warnings.

3) Volatility 3 has the really nice `deprecated_method` feature, but this does not cover when an entire function will be deprecated and going away without a direct replacement. I added this feature. This feature was necessary as the old API functions that worked with the tuples of (name, start, end) are just in no way compatible with the new API.

4) Moved all shared code to the utility class to avoid plugins importing from multiple plugins + the utility. I assume these functions were spread out through plugins before the versioned Modules utility, but that is the wrong way to approach it now. This also prevents and avoids the need for circular imports.

5) I fixed up typing where I noticed it in plugins already getting a major version bump.

### DIFF
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -11,12 +11,9 @@ import inspect
 import logging
 import os
 import traceback
-import functools
-import warnings
-from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, Type, TypeVar
+from typing import Any, Dict, Generator, List, Optional, Tuple, Type, TypeVar
 
-from volatility3.framework import constants, exceptions, interfaces
-from volatility3.framework.configuration import requirements
+from volatility3.framework import constants, interfaces
 
 if (
     sys.version_info.major != constants.REQUIRED_PYTHON_VERSION[0]
@@ -64,77 +61,6 @@ def require_interface_version(*args) -> None:
                         ".".join(str(x) for x in args[0:2]),
                     )
                 )
-
-
-class Deprecation:
-    """Deprecation related methods."""
-
-    @staticmethod
-    def method_being_removed(message: str):
-
-        def decorator(deprecated_func):
-            @functools.wraps(deprecated_func)
-            def wrapper(*args, **kwargs):
-                warnings.warn(
-                    f"This API ({deprecated_func.__module__}.{deprecated_func.__qualname__}) will be removed in a release very soon. {message}",
-                    FutureWarning,
-                )
-                return deprecated_func(*args, **kwargs)
-
-            return wrapper
-
-        return decorator
-
-    @staticmethod
-    def deprecated_method(
-        replacement: Callable,
-        replacement_version: Tuple[int, int, int] = None,
-        additional_information: str = "",
-    ):
-        """A decorator for marking functions as deprecated.
-
-        Args:
-            replacement: The replacement function overriding the deprecated API, in the form of a Callable (typically a method)
-            replacement_version: The "replacement" base class version that the deprecated method expects before proxying to it. This implies that "replacement" is a method from a class that inherits from VersionableInterface.
-            additional_information: Information appended at the end of the deprecation message
-        """
-
-        def decorator(deprecated_func):
-            @functools.wraps(deprecated_func)
-            def wrapper(*args, **kwargs):
-                nonlocal replacement, replacement_version, additional_information
-                # Prevent version mismatches between deprecated (proxy) methods and the ones they proxy
-                if (
-                    replacement_version is not None
-                    and callable(replacement)
-                    and hasattr(replacement, "__self__")
-                ):
-                    replacement_base_class = replacement.__self__
-
-                    # Verify that the base class inherits from VersionableInterface
-                    if inspect.isclass(replacement_base_class) and issubclass(
-                        replacement_base_class,
-                        interfaces.configuration.VersionableInterface,
-                    ):
-                        # SemVer check
-                        if not requirements.VersionRequirement.matches_required(
-                            replacement_version, replacement_base_class.version
-                        ):
-                            raise exceptions.VersionMismatchException(
-                                deprecated_func,
-                                replacement_base_class,
-                                replacement_version,
-                                "This is a bug, the deprecated call needs to be removed and the caller needs to update their code to use the new method.",
-                            )
-
-                deprecation_msg = f"Method \"{deprecated_func.__module__ + '.' + deprecated_func.__qualname__}\" is deprecated, use \"{replacement.__module__ + '.' + replacement.__qualname__}\" instead. {additional_information}"
-                warnings.warn(deprecation_msg, FutureWarning)
-                # Return the wrapped function with its original arguments
-                return deprecated_func(*args, **kwargs)
-
-            return wrapper
-
-        return decorator
 
 
 class NonInheritable:
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -70,6 +70,19 @@ class Deprecation:
     """Deprecation related methods."""
 
     @staticmethod
+    def method_being_removed(message: str):
+
+        def decorator(deprecated_func):
+            @functools.wraps(deprecated_func)
+            def wrapper(*args, **kwargs):
+                warnings.warn(f"This API ({deprecated_func.__module__}.{deprecated_func.__qualname__}) will be removed in a release very soon. {message}", FutureWarning)
+                return deprecated_func(*args, **kwargs)
+
+            return wrapper
+
+        return decorator
+
+    @staticmethod
     def deprecated_method(
         replacement: Callable,
         replacement_version: Tuple[int, int, int] = None,
--- a/volatility3/framework/__init__.py
+++ b/volatility3/framework/__init__.py
@@ -75,7 +75,10 @@ class Deprecation:
         def decorator(deprecated_func):
             @functools.wraps(deprecated_func)
             def wrapper(*args, **kwargs):
-                warnings.warn(f"This API ({deprecated_func.__module__}.{deprecated_func.__qualname__}) will be removed in a release very soon. {message}", FutureWarning)
+                warnings.warn(
+                    f"This API ({deprecated_func.__module__}.{deprecated_func.__qualname__}) will be removed in a release very soon. {message}",
+                    FutureWarning,
+                )
                 return deprecated_func(*args, **kwargs)
 
             return wrapper
--- a/volatility3/framework/deprecation.py
+++ b/volatility3/framework/deprecation.py
@@ -15,6 +15,14 @@ from volatility3.framework.configuration import requirements
 
 
 def method_being_removed(message: str, removal_date: str):
+    """A decorator for marking functions as being removed in the future and without a replacement.
+       Callers to this function should explicitly list the API paths that should be used instead.
+
+    Args:
+        message: A message added to the standard deprecation warning. Should include the replacement API paths
+        removal_date: A YYYY-MM-DD formatted date of when the function will be removed from the framework
+    """
+
     def decorator(deprecated_func):
         @functools.wraps(deprecated_func)
         def wrapper(*args, **kwargs):
@@ -39,6 +47,7 @@ def deprecated_method(
 
     Args:
         replacement: The replacement function overriding the deprecated API, in the form of a Callable (typically a method)
+        removal_date: A YYYY-MM-DD formatted date of when the function will be removed from the framework
         replacement_version: The "replacement" base class version that the deprecated method expects before proxying to it. This implies that "replacement" is a method from a class that inherits from VersionableInterface.
         additional_information: Information appended at the end of the deprecation message
     """
--- a/volatility3/framework/deprecation.py
+++ b/volatility3/framework/deprecation.py
@@ -1,0 +1,81 @@
+# This file is Copyright 2025 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+# This file contains the Deprecation class used to deprecate methods in an orderly manner
+
+import warnings
+import functools
+import inspect
+
+from typing import Callable, Tuple
+
+from volatility3.framework import interfaces, exceptions
+from volatility3.framework.configuration import requirements
+
+
+def method_being_removed(message: str):
+
+    def decorator(deprecated_func):
+        @functools.wraps(deprecated_func)
+        def wrapper(*args, **kwargs):
+            warnings.warn(
+                f"This API ({deprecated_func.__module__}.{deprecated_func.__qualname__}) will be removed in a release very soon. {message}",
+                FutureWarning,
+            )
+            return deprecated_func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+def deprecated_method(
+    replacement: Callable,
+    replacement_version: Tuple[int, int, int] = None,
+    additional_information: str = "",
+):
+    """A decorator for marking functions as deprecated.
+
+    Args:
+        replacement: The replacement function overriding the deprecated API, in the form of a Callable (typically a method)
+        replacement_version: The "replacement" base class version that the deprecated method expects before proxying to it. This implies that "replacement" is a method from a class that inherits from VersionableInterface.
+        additional_information: Information appended at the end of the deprecation message
+    """
+
+    def decorator(deprecated_func):
+        @functools.wraps(deprecated_func)
+        def wrapper(*args, **kwargs):
+            nonlocal replacement, replacement_version, additional_information
+            # Prevent version mismatches between deprecated (proxy) methods and the ones they proxy
+            if (
+                replacement_version is not None
+                and callable(replacement)
+                and hasattr(replacement, "__self__")
+            ):
+                replacement_base_class = replacement.__self__
+
+                # Verify that the base class inherits from VersionableInterface
+                if inspect.isclass(replacement_base_class) and issubclass(
+                    replacement_base_class,
+                    interfaces.configuration.VersionableInterface,
+                ):
+                    # SemVer check
+                    if not requirements.VersionRequirement.matches_required(
+                        replacement_version, replacement_base_class.version
+                    ):
+                        raise exceptions.VersionMismatchException(
+                            deprecated_func,
+                            replacement_base_class,
+                            replacement_version,
+                            "This is a bug, the deprecated call needs to be removed and the caller needs to update their code to use the new method.",
+                        )
+
+            deprecation_msg = f"Method \"{deprecated_func.__module__ + '.' + deprecated_func.__qualname__}\" is deprecated, use \"{replacement.__module__ + '.' + replacement.__qualname__}\" instead. {additional_information}"
+            warnings.warn(deprecation_msg, FutureWarning)
+            # Return the wrapped function with its original arguments
+            return deprecated_func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
--- a/volatility3/framework/deprecation.py
+++ b/volatility3/framework/deprecation.py
@@ -14,13 +14,12 @@ from volatility3.framework import interfaces, exceptions
 from volatility3.framework.configuration import requirements
 
 
-def method_being_removed(message: str):
-
+def method_being_removed(message: str, removal_date: str):
     def decorator(deprecated_func):
         @functools.wraps(deprecated_func)
         def wrapper(*args, **kwargs):
             warnings.warn(
-                f"This API ({deprecated_func.__module__}.{deprecated_func.__qualname__}) will be removed in a release very soon. {message}",
+                f"This API ({deprecated_func.__module__}.{deprecated_func.__qualname__}) will be removed in the first release after {removal_date}. {message}",
                 FutureWarning,
             )
             return deprecated_func(*args, **kwargs)
@@ -32,8 +31,9 @@ def method_being_removed(message: str):
 
 def deprecated_method(
     replacement: Callable,
+    removal_date: str,
     replacement_version: Tuple[int, int, int] = None,
-    additional_information: str = "",
+    additional_information: str = ""
 ):
     """A decorator for marking functions as deprecated.
 
@@ -71,7 +71,7 @@ def deprecated_method(
                             "This is a bug, the deprecated call needs to be removed and the caller needs to update their code to use the new method.",
                         )
 
-            deprecation_msg = f"Method \"{deprecated_func.__module__ + '.' + deprecated_func.__qualname__}\" is deprecated, use \"{replacement.__module__ + '.' + replacement.__qualname__}\" instead. {additional_information}"
+            deprecation_msg = f"Method \"{deprecated_func.__module__ + '.' + deprecated_func.__qualname__}\" is deprecated and will be removed in the first release after {removal_date}, use \"{replacement.__module__ + '.' + replacement.__qualname__}\" instead. {additional_information}"
             warnings.warn(deprecation_msg, FutureWarning)
             # Return the wrapped function with its original arguments
             return deprecated_func(*args, **kwargs)
--- a/volatility3/framework/deprecation.py
+++ b/volatility3/framework/deprecation.py
@@ -33,7 +33,7 @@ def deprecated_method(
     replacement: Callable,
     removal_date: str,
     replacement_version: Tuple[int, int, int] = None,
-    additional_information: str = ""
+    additional_information: str = "",
 ):
     """A decorator for marking functions as deprecated.
 
--- a/volatility3/framework/plugins/linux/check_modules.py
+++ b/volatility3/framework/plugins/linux/check_modules.py
@@ -40,6 +40,7 @@ class Check_modules(plugins.PluginInterface):
     @classmethod
     @deprecation.deprecated_method(
         replacement=linux_utilities_modules.Modules.get_kset_modules,
+        removal_date="2025-09-25",
         replacement_version=(2, 0, 0),
     )
     def get_kset_modules(
--- a/volatility3/framework/plugins/linux/check_modules.py
+++ b/volatility3/framework/plugins/linux/check_modules.py
@@ -6,7 +6,7 @@ import logging
 from typing import List, Dict
 
 import volatility3.framework.symbols.linux.utilities.modules as linux_utilities_modules
-from volatility3.framework import interfaces, renderers, Deprecation
+from volatility3.framework import interfaces, renderers, deprecation
 from volatility3.framework.configuration import requirements
 from volatility3.framework.interfaces import plugins
 from volatility3.framework.objects import utility
@@ -38,7 +38,7 @@ class Check_modules(plugins.PluginInterface):
         ]
 
     @classmethod
-    @Deprecation.deprecated_method(
+    @deprecation.deprecated_method(
         replacement=linux_utilities_modules.Modules.get_kset_modules,
         replacement_version=(2, 0, 0),
     )
--- a/volatility3/framework/plugins/linux/check_modules.py
+++ b/volatility3/framework/plugins/linux/check_modules.py
@@ -3,14 +3,15 @@
 #
 
 import logging
-from typing import List
+from typing import List, Dict
 
-from volatility3.framework import interfaces, renderers, exceptions, constants
+import volatility3.framework.symbols.linux.utilities.modules as linux_utilities_modules
+from volatility3.framework import interfaces, renderers, Deprecation
 from volatility3.framework.configuration import requirements
 from volatility3.framework.interfaces import plugins
 from volatility3.framework.objects import utility
 from volatility3.framework.renderers import format_hints
-from volatility3.plugins.linux import lsmod
+from volatility3.framework.symbols.linux import extensions
 
 vollog = logging.getLogger(__name__)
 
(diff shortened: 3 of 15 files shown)


volatility3/framework/__init__.py
```diff
@@ -69,6 +69,22 @@ def require_interface_version(*args) -> None:
...
     """Deprecation related methods."""
 
+    @staticmethod
+    def method_being_removed(message: str):
+
+        def decorator(deprecated_func):
+            @functools.wraps(deprecated_func)
+            def wrapper(*args, **kwargs):
+                warnings.warn(
+                    f"This API ({deprecated_func.__module__}.{deprecated_func.__qualname__}) will be removed in a release very soon. {message}",
```
ikelos: Might be worth adding an extra parameter with a version number to say when it'll definitely be deprecated by?  Problem is the versions bump at unexpected times, so perhaps a datetime for removal is better?  It would also make it easy for us to grep them all at release and remove the ones whose time has come?

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -367,9 +368,8 @@ def files_descriptors_for_process(
                 yield fd_num, filp, full_path
 
     @classmethod
-    @Deprecation.deprecated_method(
-        replacement=linux_utilities_modules.Modules.mask_mods_list,
-        replacement_version=(1, 0, 0),
+    @Deprecation.method_being_removed(
+        "This method is not needed. The correct API for mapping kernel pointers to modules is `linux_utilities_modules.Modules.run_module_scanners`"
```
ikelos: The message should be something a little softer but more direct.  "This method is no longer recommended and will be removed from the framework in a future release.  Code using this method should be adapted to make use of `linux_utilities_modules.Modules.run_module_scanners`"

Not needed implies it's fine to keep using it, and "correct API" sounds like you're calling this one wrong, which is wasn't right up until we deprecated it (and it'll actually still work afterwards just fine for a while).

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -3,7 +3,8 @@
 #
 import logging
 from typing import List, Set, Tuple, Iterable
-from volatility3.framework import renderers, interfaces, exceptions, objects
+import volatility3.framework.symbols.linux.utilities.modules as linux_utilities_modules
```
ikelos:
```suggestion
from volatility3.framework.symbols.linux.utilities import modules as linux_utilities_modules
```
Just a stylistic point and has no impact on the code, but I kinda think it's easier to stick with the "always import from" and then do the renaming after?

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -3,7 +3,8 @@
...
+from volatility3.framework import renderers, interfaces, exceptions, Deprecation
```
ikelos:
```suggestion
from volatility3 import framework
from volatility3.framework import renderers, interfaces, exceptions
```

Then use `framework.Deprecation`?  I'd expect David's future tool to flag this, so... might as well do it right first time round?

ikelos: To be fair, it's a little odd they're in a class anyway?  The class isn't versioned or anything, just seems to be used for grouping.  I'd almost prefer a `deprecation.py` and have them just live straight in there?

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -29,46 +30,30 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+            requirements.VersionRequirement(
+                name="linux_utilities_modules",
+                component=linux_utilities_modules.Modules,
+                version=(2, 0, 0),
+            ),
         ]
 
-    @classmethod
+    @staticmethod
+    @Deprecation.deprecated_method(
```
ikelos: Yep, shouldn't start with a capital D, definitely prefer these methods moving to a module of their own please.

ikelos: Happy for you to do it in this PR, but fine if you just want to put a bug number here and we can do it later, but we might as well do it before use becomes widespread.

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -29,46 +30,30 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+        return linux_utilities_modules.Modules.get_modules_memory_boundaries(
+            context, vmlinux_module_name
+        )
 
+    @Deprecation.deprecated_method(
+        replacement=linux_utilities_modules.Modules.get_module_address_alignment,
+        replacement_version=(2, 0, 0),
+    )
     @classmethod
     def _get_module_address_alignment(
```
ikelos, line 62: Surprised I let a private method be exposed as a classmethod, but just as well to deprecate it, thanks!

ikelos: (It probably had to be a classmethod to be called from within another classmethod, that makes sense).

volatility3/framework/plugins/linux/lsmod.py
```diff
@@ -6,7 +6,8 @@
 import logging
 from typing import List, Iterable
 
-from volatility3.framework import exceptions, renderers, constants, interfaces
+import volatility3.framework.symbols.linux.utilities.modules as linux_utilities_modules
+from volatility3.framework import exceptions, renderers, interfaces, Deprecation
```
ikelos:
```suggestion
from volatility3.framework.symbols.linux.utilities import modules as linux_utilities_modules
from volatility3 import framework
from volatility3.framework import exceptions, renderers, interfaces
```

volatility3/framework/plugins/linux/lsmod.py
```diff
@@ -29,43 +30,38 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
             ),
+            requirements.VersionRequirement(
+                name="linux_utilities_modules",
+                component=linux_utilities_modules.Modules,
+                version=(2, 0, 0),
+            ),
         ]
 
     @classmethod
+    @Deprecation.deprecated_method(
```
ikelos:
```suggestion
    @framework.Deprecation.deprecated_method(
```

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -88,27 +73,14 @@ def _get_module_address_alignment(
...
-
-        Returns:
-            True if all the addresses meet the alignment
-        """
-        return all(addr % address_alignment == 0 for addr in addresses)
+        return linux_utilities_modules.get_module_address_alignment(
+            context, vmlinux_module_name
+        )
 
+    @Deprecation.deprecated_method(
```
ikelos:
```suggestion
    @framework.Deprecation.deprecated_method(
```

volatility3/framework/plugins/linux/hidden_modules.py
ikelos:
```suggestion
    @framework.Deprecation.deprecated_method(
```

volatility3/framework/plugins/linux/check_modules.py
```diff
@@ -29,58 +30,33 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
-                name="lsmod", plugin=lsmod.Lsmod, version=(2, 0, 0)
+            requirements.VersionRequirement(
+                name="linux_utilities_modules",
+                component=linux_utilities_modules.Modules,
+                version=(2, 0, 0),
             ),
         ]
 
     @classmethod
+    @Deprecation.deprecated_method(
```
ikelos:
```suggestion
    @framework.Deprecation.deprecated_method(
```

volatility3/framework/plugins/linux/check_modules.py
```diff
@@ -3,22 +3,23 @@
 #
 
 import logging
-from typing import List
+from typing import List, Dict
 
-from volatility3.framework import interfaces, renderers, exceptions, constants
+import volatility3.framework.symbols.linux.utilities.modules as linux_utilities_modules
+from volatility3.framework import interfaces, renderers, Deprecation
```
ikelos:
```suggestion
from volatility3.framework.symbols.linux.utilities import modules as linux_utilities_modules
from volatility3 import framework
from volatility3.framework import interfaces, renderers
```

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -3,8 +3,10 @@
 #
 import logging
 from typing import List, Dict, Iterator
-from volatility3.plugins.linux import lsmod, check_modules, hidden_modules
-from volatility3.framework import interfaces
+
+import volatility3.framework.symbols.linux.utilities.modules as linux_utilities_modules
+
+from volatility3.framework import interfaces, Deprecation
```
ikelos:
```suggestion
from volatility3.framework.symbols.linux.utilities import modules as linux_utilities_modules
from volatility3 import framework
from volatility3.framework import interfaces
```

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -54,6 +48,10 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
         ]
 
     @classmethod
+    @Deprecation.deprecated_method(
```
ikelos:
```suggestion
    @framework.Deprecation.deprecated_method(
```

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -67,83 +65,53 @@ def flatten_run_modules_results(
...
-                if deduplicate and module.vol.offset in seen_addresses:
-                    continue
-                seen_addresses.add(module.vol.offset)
-                yield module
+        return linux_utilities_modules.Modules.flatten_run_modules_results(
+            run_results, deduplicate
+        )
 
     @classmethod
+    @Deprecation.deprecated_method(
```
ikelos:
```suggestion
    @framework.Deprecation.deprecated_method(
```

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -10,11 +10,9 @@
 from dataclasses import dataclass
 
 import volatility3.framework.symbols.linux.utilities.modules as linux_utilities_modules
```
ikelos, line 12:
```suggestion
from volatility3.framework.symbols.linux.utilities import modules as linux_utilities_modules
```

volatility3/framework/plugins/linux/tracing/ftrace.py
```diff
@@ -264,16 +217,15 @@ def _generator(self):
...
+            vollog.error(
+                'The provided symbol table does not include the "ftrace_ops_list" symbol. This means you are either analyzing an unsupported kernel version or that your symbol table is corrupted.'
             )
+            return
 
-        # Do not run hidden_modules by default, but only on failure to find a module
-        known_modules = modxview.Modxview.run_modules_scanners(
-            self.context, kernel_name, run_hidden_modules=False
+        known_modules = linux_utilities_modules.Modules.run_modules_scanners(
+            self.context, kernel_name, run_hidden_modules=True
```
ikelos, line 226: Is hidden_modules fast enough to include by default now?

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -367,9 +368,8 @@ def files_descriptors_for_process(
                 yield fd_num, filp, full_path
 
     @classmethod
-    @Deprecation.deprecated_method(
-        replacement=linux_utilities_modules.Modules.mask_mods_list,
-        replacement_version=(1, 0, 0),
+    @Deprecation.method_being_removed(
```
ikelos:
```suggestion
    @framework.Deprecation.method_being_removed(
```

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -385,13 +385,18 @@ def mask_mods_list(
         return linux_utilities_modules.Modules.mask_mods_list(context, layer_name, mods)
 
     @classmethod
+    @Deprecation.method_being_removed(
```
ikelos:
```suggestion
    @framework.Deprecation.method_being_removed(
```

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -1,77 +1,121 @@
...
 
 from volatility3 import framework
-from volatility3.framework import constants, interfaces
+from volatility3.framework import (
+    constants,
+    interfaces,
+    Deprecation,
+    exceptions,
+    objects,
+)
```
ikelos:
```suggestion
from volatility3.framework import (
    constants,
    interfaces,
    exceptions,
    objects,
)
```

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -1,77 +1,121 @@
...
+
+            if symbol_name:
+                symbol_name = symbol_name.split(constants.BANG)[1]
+
+            return match, symbol_name
+
+        return None, None
 
     @classmethod
+    @Deprecation.method_being_removed(
```
ikelos:
```suggestion
    @framework.Deprecation.method_being_removed(
```

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -1,77 +1,121 @@
...
+        "This method is not needed. The correct API for mapping kernel pointers to modules is `linux_utilities_modules.Modules.run_module_scanners`"
```
ikelos: Again, just soften the "correct" API, and we don't need to say it's not needed, just that it's being removed.

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -1,77 +1,121 @@
...
+    )
     def mask_mods_list(
         cls,
         context: interfaces.context.ContextInterface,
-        layer_name: str,
-        mods: Iterator[interfaces.objects.ObjectInterface],
-    ) -> List[Tuple[str, int, int]]:
+        kernel_layer_name: str,
+        mods: Iterator[extensions.module],
+    ) -> List[extensions.module]:
```
ikelos: Errr, the ocde looks like it still returns tuples?  It's fine because it's deprecated but... I think the original typing information was correct, at least for the output?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -116,3 +163,369 @@ def lookup_module_address(
...
+
+        end_addr = kernel.object_from_symbol("_etext")
+        end_addr = end_addr.vol.offset & mask
+
+        return Modules.ModuleInfo(
+            start_addr, constants.linux.KERNEL_NAME, start_addr, end_addr
+        )
+
+    @classmethod
+    def run_modules_scanners(
```
ikelos, line 218: Something to consider here is whether we want to do something like the linux pslist thing we did?  With different methods with the same signature that scanned different ways?  Then you just hand in the list of scanners you want rather than boolean flags to turn on and off specific sets of scanners?

ikelos (commented):
Cool, looks really good!  For a minute I was worried we were changing the `LinuxUtilities` MAJOR version, but then saw it was just the Modules extension.  5:)

There's a couple considerations about changing the API but they may be excessive, so feel free to reject them.  I really don't like the way there's a Deprecation class and we're importing it directly, I'd really like that cleaned up as part of this PR please but will accept an open bug if you want to put it off a bit.  Otherwise looks really good and simplifies/collects together a lot of the functionality, nice!  5:D

volatility3/framework/plugins/linux/hidden_modules.py
atcuno: Yea, the Deprecation is only previously used in `linux/symbols/__init__.py` so far, which is being changed in this PR anyway. I thought the capital D was odd so I can fix that. Also, can put into its own file under framework/. Your other feedback is good too. I like the date idea, so I will do that tomorrow as well. Brain is too burned for today, but will knock out first thing in the morning.

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: I was going to revisit the `implementation_method` for this after this PR goes on. I honestly don't even like the skipping hidden part, but left it here as it was there already. We should just get all the sources all the time.

Abyss-W4tcher: The issue with doing the hidden_modules scanning by default, is that if the caller never needs it (i.e. it is able to resolve all symbols to modules without it), then we will just have wasted time and CPU scanning the space. 

The previous design was triggering the scanning on a per-need basis, if a resolution issue was raised. I think we should leave the choice, depending on the use case ?

This is even more true on unaligned spaces, where hidden_modules can take quite a long time running.

atcuno: Its only slow when the path for 1 byte alignment is hit, which I am going to fix in a follow up PR. Otherwise, it didn't change speed much in my testing since it only scans inside the vmmemap bounds anyway. If there are samples where this is problematic let me know though as I am confident we can make the scanner quick vs having to conditionally run it.

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: Good catch.

volatility3/framework/symbols/linux/__init__.py
atcuno: ok, `method_being_removed` already adds the generic deprecation method so will shorten to just the replacement method

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -3,7 +3,8 @@
 #
 import logging
 from typing import List, Set, Tuple, Iterable
-from volatility3.framework import renderers, interfaces, exceptions, objects
+from volatility3.framework.symbols.linux.utilities import modules as linux_utilities_modules
```
ikelos:
```suggestion
from volatility3.framework.symbols.linux.utilities import (
    modules as linux_utilities_modules,
)
```
Black complains about it

atcuno:
@ikelos I addressed all your feedback except for adding a datetime to deprecation headings as I am not sure what the implementation would look like.

So, if the calls took a `datetime`, what would happen with them? The specific date printed in the warning message? Backtrace the calling plugin with a new deprecation exception? Or would they be a parameter just to be grepped before releases?

ikelos:
Yeah, didn't need to be a `datetime` per-se, even a date string would do (`YYYY-MM-DD` please), so that we could easily search for ones that should be removed, and we could tell everyone exactly when the method should no longer be used...

ikelos:
We could theoretically throw a full on exception if it tries to be used after that date, but I think we should just allow it until we get around to removing the method (hopefully it'll happen ever release, one for us to check during the release process).

ikelos:
We need to come up with a general policy, I'd suggest at least one year (although I'd possibly go as low as 6 months, because with our release cadence it's more likely to be 9 months) or when a MAJOR version bump happens (no point carry over cruft when everything changes)...

atcuno:
I did 6 months plus a few days @ikelos

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -1,77 +1,122 @@
...
+            The first memory module in which the address fits and the symbol name for `target_address`
 
         Kernel documentation:
             "within_module" and "within_module_mem_type" functions
         """
+        kernel = context.modules[kernel_module_name]
+
+        kernel_layer = context.layers[kernel.layer_name]
+
+        if modules[0].start != modules[0].start & kernel_layer.address_mask:
```
ikelos: This seems a *really* suspect test, it's checking that high bits aren't set?  Why is it only doing it for one entry?  What if the list is empty (for some reason)?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -1,77 +1,122 @@
...
-                f"Address {hex(target_address)} fits in modules at {[hex(module.vol.offset) for module in matches]}, indicating potential modules memory space overlap.",
-                UserWarning,
-            )
-            return matches[0]
-        elif len(matches) == 1:
-            return matches[0]
+        if len(matches) >= 1:
+            if len(matches) > 1:
+                warnings.warn(
+                    f"Address {hex(target_address)} fits in modules at {[hex(module.start) for module in matches]}, indicating potential modules memory space overlap.",
```
ikelos: Since this doesn't then bomb out, you should say which one will win.

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -116,3 +165,369 @@ def lookup_module_address(
...
+        """
+        kernel = context.modules[kernel_module_name]
+
+        mask = context.layers[kernel.layer_name].address_mask
+
+        start_addr = kernel.object_from_symbol("_text")
+        start_addr = start_addr.vol.offset & mask
+
+        end_addr = kernel.object_from_symbol("_etext")
+        end_addr = end_addr.vol.offset & mask
```
ikelos:
```suggestion
        address_mask = context.layers[kernel.layer_name].address_mask

        start_addr = kernel.object_from_symbol("_text")
        start_addr = start_addr.vol.offset & address_mask

        end_addr = kernel.object_from_symbol("_etext")
        end_addr = end_addr.vol.offset & address_mask
```
Could we call this address_mask to keep it consistent with other methods please?

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Can you raise a ticket and note it here to make sure this gets addressed in the future please?  (The redo to make the set of gatherers more generic/pluggable)

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -116,3 +165,369 @@ def lookup_module_address(
...
+            if modinfo:
+                run_results["check_modules"].append(modinfo)
+
+        # hidden_modules
+        if run_hidden_modules:
+            known_modules_addresses = set(
+                context.layers[kernel.layer_name].canonicalize(modinfo.start)
+                for modinfo in run_results["kernel"]
+                + run_results["lsmod"]
+                + run_results["check_modules"]
```
ikelos, line 268:
```suggestion
                for modinfo in cls.flatten_run_modules_results(run_results)
```
This should probably be driven by the contents of the list, rather than hard coding the available tests, which looks to be what flatten_run_modules_results does?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -116,3 +165,369 @@ def lookup_module_address(
...
+        Yields:
+            The modules present in the `layer_name` layer's modules list
+
+        This function will throw a SymbolError exception if kernel module support is not enabled.
+        """
+        vmlinux = context.modules[vmlinux_module_name]
+
+        modules = vmlinux.object_from_symbol(symbol_name="modules").cast("list_head")
+
+        table_name = modules.vol.type_name.split(constants.BANG)[0]
```
ikelos: This should just be `vmlinux.symbol_table_name`, no?

ikelos (commented):
Looks good (assuming a follow up to make the types of module scans more... modular... so to speak).  A couple minor points but no show stoppers...

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: I will add the check for an empty list (good catch), but I want to keep the index 0 check as it will enforce callers going through the right API flow to reach that code. All kernel modules will have the high bits set initially, so if the right API flow isn't used then non-masked ones will make it there and everything processes incorrectly.

atcuno:
@ikelos I managed to break the commit history while rebasing. I asked @dgmcdona to look at it. Will finish it up after.

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Shouldn't it be something like `any(modules[i].start =! modules[i].start & kernel_layer.address_mask for i in modules)` though?  Picking out the first one seems like a half hearted check?

ikelos:
@j-t-1, please avoid modifying other people's pull requests.  It can cause a huge amount of confusion, and in general, it's better to have any changes you want to make to go through  a proper PR so they can be reviewed...

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: If the caller comes through the right API then all will be masked the same, if they don't then none will be masked... if they somehow randomly enumerate on their own and only mask half.. then they are just trying to be broken.

ikelos: That's not a good argument for why not to check them all?  You're applying trust to something you could check, so why not just check it?  Otherwise I'm certain someone could write code that masks the first value and then makes this call.  There's little point in half a test (and it also requires that the list has elements, whereas any/all would not)...

j-t-1:
> @j-t-1, please avoid modifying other people's pull requests. It can cause a huge amount of confusion, and in general, it's better to have any changes you want to make to go through a proper PR so they can be reviewed...

Apologies for this. If I understand what has caused this, I deleted `pid_list = pid_list or []`, which I then realised broke the iteration in the line below, so I undeleted it in a subsequent commit, but only the first was used in a rebase. I will not deliberately modify other people's pull requests (if there is a way to remove access for me to do this, please do as one less potential error pathway).

ikelos:
Thanks, we will be improving permissions so that we don't have to spot when things go wrong, but can prevent them from being done in the first place.  I fully understand it was done intending to improve the commit, but we probably ought to prevent others from modifying someone's PR as you say to reduce potential error pathways.  Thanks for understanding...  5:)

volatility3/framework/deprecation.py
```diff
@@ -0,0 +1,81 @@
...
+import functools
+import inspect
+
+from typing import Callable, Tuple
+
+from volatility3.framework import interfaces, exceptions
+from volatility3.framework.configuration import requirements
+
+
+def method_being_removed(message: str, removal_date: str):
```
ikelos, line 17: Any chance of a docstring for this please?

ikelos (commented):
Looking really good, just the currently open comments to resolve please

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: The ticket is here now: https://github.com/volatilityfoundation/volatility3/issues/1699

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: Updated

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: added

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: changed

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: Yes this will be a part of: https://github.com/volatilityfoundation/volatility3/issues/1699

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: Good call. Much cleaner

volatility3/framework/deprecation.py
atcuno: added

atcuno:
all the feedback should be addressed now @ikelos

ikelos (approved):
Ok, looks good (although I'd still have preferred the `from ... import ... as ...` syntax, but it's not a showstopper)...  5;P

PR #1674, opened 2025-03-07
Change the type information reported for missing attributes

atcuno:
This triggered an issue today when I was testing Windows:

https://github.com/volatilityfoundation/volatility3/pull/1657

The problem is that Pointers and Void objects can end up through there, which do not have the vol.type_name set. From debugging with many print statements, including inside of Python itself, it seems many paths cross there in Volatility. At least getting the attribute name and type should help much more than before when it didn't report anything specifc.

This was very hard to debug and I had to add a bunch of print statements inside of the Python core libraries, which made my VMs quite unhappy :|

### DIFF
--- a/volatility3/framework/interfaces/objects.py
+++ b/volatility3/framework/interfaces/objects.py
@@ -133,7 +133,7 @@ class ObjectInterface(metaclass=abc.ABCMeta):
 
     def __getattr__(self, attr: str) -> Any:
         """Method for ensuring volatility members can be returned."""
-        raise AttributeError(f"Unable to find {attr} for type {self.vol.type_name}")
+        raise AttributeError(f"Unable to find {attr} for type {type(self)}")
 
     @property
     def vol(self) -> ReadOnlyMapping:


ikelos:
Surely it's better to give void and pointer a vol.type_name?  That should be possible to rely upon for all volatility objects (so I'm impressed they could be constructed without a value set?).  Pointer is derived from `Pointer -> Integer -> PrimitiveObject -> interfaces.objects.ObjectInterface`  and `Void -> interfaces.objects.ObjectInterface` , which always takes a type_name parameter?

In fact, volshell shows:
```
(layer_name) >>> ps()[0].SecurityPort.vol.type_name
'symbol_table_name1!pointer'
(layer_name) >>> thing = ps()[0].SecurityPort.cast("void")
(layer_name) >>> thing.vol.type_name
'symbol_table_name1!void'
```
so at least *some* pointer have a type_name.  Which means, somewhere pointers are being created that *don't* and that needs to be fixed rather than masked over.  Now that you know how to find/instances where this happens, could you either provide a sample or figure out why you're getting malformed VolatilityObjects (if that's what they are) please?

atcuno:
This was the code path causing it: https://github.com/volatilityfoundation/volatility3/pull/1691

I will further investigate the issue with pointers not getting a name separately here. It is easy to reproduce now.

atcuno:
also, @ikelos this code is only triggered deep inside deepcopy calls, so I am not sure if its worth investigating before parity or just ban calls? See the other PR again as deepcopy() without validation seems like a bad pattern anyway.

atcuno:
@ikelos formula for replicating (should work on any sample):

1) Run windows.thread and get the address (first column) for any thread
2) Go into volshell 
3) do the following, substituting the thread structure address for 12345

```
t = self.context.modules[self.config["kernel"]].object(offset=12345, object_type="_ETHREAD")
```

4) Print `t.StartAddress`, make sure its not 0. If its 0 go back to step 1 and pick a different thread
5) do:

```
import copy
x = {"ntdll.dll" : {"addresses" : t.StartAddress}}
copy.deepcopy(x)
```

This will break with an infinite recursion backtrace.

atcuno:
> @ikelos formula for replicating (should work on any sample):
> 

When doing these steps, `t` does have a name:

```
(layer_name) >>> t.StartAddress.vol.type_name
'symbol_table_name1!pointer'
```

So I just think its something crazy with deepcopy.

ikelos:
Hmmmm, might still give it a prod.  All I can think is that each object has a context, so if it's not smart about not duplicating that, then it could easily explode (but I'd expect a recursion error?).  Very strange, but thanks for debugging it!  Good work!

atcuno:
It does cause infinitite recursion then a stack backtrace about max depth exceeded.

ikelos:
So how did it manifest without an exception, but with a type_name not being set?

atcuno:
The suspended threads plugin was calling into the deepcopy() code path with a pointer (the thread start address) which triggered it. Then the deepcopy went off the rails, leading to a backtrace.

ikelos:
I have an even more root cause for your root cause.  Care to check out #1692 and let me know if it a) fixes this and then b) when you still want this one committing please?

ikelos:
Awww, you beat me to it!  5:P

ikelos:
I like yours better, let's go with that...

ikelos (approved):

PR #1675, opened 2025-03-07
Update the list_threads API to current standards and update current c…

### LINKED ISSUE #1648: Fix the list_threads API to just take the kernel name
This API needs to take the symbol name and layer name instead of the kernel object directly:

https://github.com/volatilityfoundation/volatility3/blob/plugin/windows_gui/volatility3/framework/plugins/windows/threads.py#L40

We then need to update all callers + the required version bumps.

atcuno:
…allers to new form

Closes #1648

### DIFF
--- a/volatility3/framework/plugins/windows/threads.py
+++ b/volatility3/framework/plugins/windows/threads.py
@@ -16,7 +16,7 @@ class Threads(thrdscan.ThrdScan):
     """Lists process threads"""
 
     _required_framework_version = (2, 4, 0)
-    _version = (2, 0, 0)
+    _version = (3, 0, 0)
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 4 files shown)


volatility3/framework/plugins/windows/threads.py
```diff
@@ -36,9 +36,11 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
             ),
         ]
 
-    @classmethod
+    @staticmethod
```
ikelos: Nope, staticmethods are bad.   It means you then can't get back to the calling class (for instance, to find version information if you need to).  So this is a red flags.

ikelos:
```suggestion
    @classmethod
```

volatility3/framework/plugins/windows/threads.py
```diff
@@ -36,9 +36,11 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
     def list_threads(
-        cls, kernel, proc: interfaces.objects.ObjectInterface
+        context: interfaces.context.ContextInterface,
```
ikelos, line 42:
```suggestion
        cls,
        context: interfaces.context.ContextInterface,
```

ikelos (changes requested):
Looks fine, but staticmethods are a no-no (makes it impossible to look up the owning class to check versioning information).

atcuno:
I fixed your static -> class feedback just by clicking buttons in the web interface. That was interesting.. this should be good now.

ikelos:
Yeah, I saw someone else using "suggestions" and figured out how to use them myself.  Seems a better way of explaining trivial stuff clearly than starting a back and forth...  5;)  Thanks for getting those fixed up, I'll merge it once the tests have finished.  5:)

ikelos (approved):
Looks good now, thanks!

PR #1676, opened 2025-03-07
Fix how the task name is reported in getsids error path

### LINKED ISSUE #1643: getsids does not used converted process name in error path
This line just calls str() on the array:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/getsids.py

Leading to output like this:

```
4700    <Array symbol_table_name1!array (.ImageFileName): layer_name @ 0xce83f8e4e628 #15>      Token unreadable
```

atcuno:
Closes https://github.com/volatilityfoundation/volatility3/issues/1643


ikelos (approved):
Looks good, thanks!

PR #1679, opened 2025-03-07
windows.lsadump: Improve secret readability

Abyss-W4tcher:
Hi,

The current `windows.ldsadump` plugin outputs secrets both as decoded and hex-encoded format. However, the output is generally very messy (source is `win-xp-laptop-2005-06-25.img`): 

image

This can be circumvented by using the "-r json" renderer, but I've encountered a lot of confusion amongst analysts that think the plugin is broken or outputting garbage data. Some even roll back to the hexdump-ed Vol2 version. 

I do believe keeping both columns is better, but the "printable" one purpose should only be to have a quick look over secrets and spot "human readable" ones. If someone wants to manipulate the secret, the "Hex" column will then come handy.

Here is the updated version, which improves readability while offering precise information about hex values: 

image

`b''` enclosing is commonly used to indicate mixed printable and non-printable characters.

Thanks :)

### DIFF
--- a/volatility3/framework/plugins/windows/lsadump.py
+++ b/volatility3/framework/plugins/windows/lsadump.py
@@ -204,7 +204,7 @@ class Lsadump(interfaces.plugins.PluginInterface):
             else:
                 secret = self.decrypt_aes(enc_secret, lsakey)
 
-            yield (0, (key.get_name(), secret.decode("latin1"), secret))
+            yield (0, (key.get_name(), str(secret), secret))
 
     def run(self):
         offset = self.config.get("offset", None)
--- a/volatility3/framework/plugins/windows/lsadump.py
+++ b/volatility3/framework/plugins/windows/lsadump.py
@@ -14,6 +14,7 @@ from volatility3.framework.layers import registry
 from volatility3.framework.symbols.windows import versions
 from volatility3.plugins.windows import hashdump
 from volatility3.plugins.windows.registry import hivelist
+from volatility3.framework.renderers import format_hints
 
 vollog = logging.getLogger(__name__)
 
@@ -204,7 +205,7 @@ class Lsadump(interfaces.plugins.PluginInterface):
             else:
                 secret = self.decrypt_aes(enc_secret, lsakey)
 
-            yield (0, (key.get_name(), str(secret), secret))
+            yield (0, (key.get_name(), format_hints.HexBytes(secret), secret))
 
     def run(self):
         offset = self.config.get("offset", None)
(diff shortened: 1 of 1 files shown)


ikelos:
I'm ok with the change (although I really don't like including the b'', it looks like we just let it bleed through from the python, which essentially we did) but wouldn't this be an awesome place to use the HexBytes or the new LayerData?  That should output both the hex data and the printable characters?  Failing that, we should try handling unicode with a "replace errors with ?"  perhaps?  I'm all for fixing the output, but not this way (also, this should be suitable for web interfaces, or others if needed)...

Abyss-W4tcher:
The HexBytes renderer is nice, although slightly lengthy and obviously redundant with "Hex". Happy with HexBytes, could you provide an example of what https://github.com/volatilityfoundation/volatility3/pull/1622 will look like (sorry if there is one I didn't catch it) ? Thanks !

ikelos:
At the moment, it looks like this (from malfind):
```
2240	MsMpEng.exe	0x1e020000	0x1e21ffff	VadS	PAGE_EXECUTE_READWRITE	512	1	Disabled	N/A	
09 00 38 00 09 00 01 05 8b 55 18 8b 4d 54 8d 54 ..8......U..MT.T
0a fc 89 d6 b9 04 00 1a 00 ff 95 64 37 00 00 8b ...........d7...
4d 1c 89 08 83 45 18 fc 8d 45 1c 8b 4d 18 89 08 M....E...E..M...
81 6d 18 98 02 00 00 9f 0f 90 c0 66 89 45 38 8d .m.........f.E8.	
0x1e020000:	or	dword ptr [eax], eax
0x1e020002:	cmp	byte ptr [eax], al
0x1e020004:	or	dword ptr [eax], eax
0x1e020006:	add	dword ptr [0x8b18558b], eax
0x1e02000c:	dec	ebp
0x1e02000d:	push	esp
0x1e02000e:	lea	edx, [edx + ecx - 4]
0x1e020012:	mov	esi, edx
0x1e020014:	mov	ecx, 0x1a0004
0x1e020019:	call	dword ptr [ebp + 0x3764]
0x1e02001f:	mov	ecx, dword ptr [ebp + 0x1c]
0x1e020022:	mov	dword ptr [eax], ecx
0x1e020024:	add	dword ptr [ebp + 0x18], -4
0x1e020028:	lea	eax, [ebp + 0x1c]
0x1e02002b:	mov	ecx, dword ptr [ebp + 0x18]
0x1e02002e:	mov	dword ptr [eax], ecx
0x1e020030:	sub	dword ptr [ebp + 0x18], 0x298
0x1e020037:	lahf	
0x1e020038:	seto	al
0x1e02003b:	mov	word ptr [ebp + 0x38], ax
```

ikelos:
It will eventually support providing the offset (a bit pointless in this instance) and allow providing prior and post context (also a bit pointless), so it might be overkill...

Abyss-W4tcher:
Looks awesome and offers a lot of details ! Might be a bit overkill as you said for this use case...

PR #1681, opened 2025-03-07
Fix error handling and reporting around PE reconstruction calls

### LINKED ISSUE #1637: dos_header.reconstruct() calls need to catch ValueError
This function throws ValueErrors in many places, but all the calling plugins all over the code base (likely copy/paste from a long time ago), only catch for invalidaddress exception. Once the GUI code and API overhaul PR is merged, then we need to hunt down these calls and catch the right exception.

atcuno:
Fixes #1637

### DIFF
--- a/volatility3/framework/plugins/windows/iat.py
+++ b/volatility3/framework/plugins/windows/iat.py
@@ -74,7 +74,7 @@ class IAT(interfaces.plugins.PluginInterface):
                         pe_data.seek(offset)
                         pe_data.write(data)
                 except (exceptions.InvalidAddressException, ValueError) as excp:
-                    vollog.debug(
+                    vollog.warning(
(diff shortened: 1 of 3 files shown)


volatility3/framework/plugins/windows/iat.py
```diff
@@ -69,11 +69,23 @@ def _generator(self, procs):
...
 
-                for offset, data in dos_header.reconstruct():
-                    pe_data.seek(offset)
-                    pe_data.write(data)
+                try:
+                    for offset, data in dos_header.reconstruct():
+                        pe_data.seek(offset)
+                        pe_data.write(data)
+                except (exceptions.InvalidAddressException, ValueError) as excp:
+                    vollog.debug(
```
ikelos: Should this not be more visible (although possibly shorter), if we know something bad has happened?

ikelos (commented):
Fine, but feels like we're just quashing errors with a not-user-visible warning?

volatility3/framework/plugins/windows/iat.py
atcuno: I made it warning now.

PR #1682, opened 2025-03-07
Bring hash dumping plugins up to current coding standards and error c…

### LINKED ISSUE #1638: Fix `get_bootkey`, update plugins to correctly handle failed calls
Cachedump, lsadump, and hashdump all backtrace on many samples due to using `get_bootkey` incorrectly. Also, `get_bootkey` doesn't have docs and it calls `read` outside try/except leading to a backtrace, which is what we have been changing `get_*` calls to return None for. All these related issues needs to get fixed.

atcuno:
…hecking patterns. Fix bugs and typing.

These plugins are several years old and due to some bad patterns from then, had missing typing, caused many backtraces, etc. 

I have updated them to the modern flow now.

Fixes #1638

### DIFF
--- a/volatility3/framework/plugins/windows/cachedump.py
+++ b/volatility3/framework/plugins/windows/cachedump.py
@@ -22,7 +22,7 @@ class Cachedump(interfaces.plugins.PluginInterface):
     """Dumps lsa secrets from memory"""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 0, 1)
+    _version = (1, 0, 2)
 
     @classmethod
     def get_requirements(cls):
--- a/volatility3/framework/plugins/windows/cachedump.py
+++ b/volatility3/framework/plugins/windows/cachedump.py
@@ -22,7 +22,7 @@ class Cachedump(interfaces.plugins.PluginInterface):
     """Dumps lsa secrets from memory"""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 1, 1)
+    _version = (1, 0, 1)
 
     @classmethod
     def get_requirements(cls):
--- a/volatility3/framework/plugins/windows/cachedump.py
+++ b/volatility3/framework/plugins/windows/cachedump.py
@@ -22,7 +22,7 @@ class Cachedump(interfaces.plugins.PluginInterface):
     """Dumps lsa secrets from memory"""
 
     _required_framework_version = (2, 0, 0)
-    _version = (2, 0, 0)
+    _version = (1, 0, 1)
 
     @classmethod
     def get_requirements(cls):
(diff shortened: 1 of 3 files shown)


volatility3/framework/plugins/windows/cachedump.py
```diff
@@ -22,7 +22,7 @@ class Cachedump(interfaces.plugins.PluginInterface):
     """Dumps lsa secrets from memory"""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 0, 1)
+    _version = (2, 0, 0)
```
ikelos: This doesn't need a bump if it's requirements are bumped.  The API hasn't changed, and the requirements are checked when necessary, so just a PATCH bump will do.

volatility3/framework/plugins/windows/lsadump.py
```diff
@@ -22,7 +22,7 @@ class Lsadump(interfaces.plugins.PluginInterface):
     """Dumps lsa secrets from memory"""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 0, 0)
+    _version = (2, 0, 0)
```
ikelos: Just needs a PATCH bump from what I can see...

ikelos (commented):
Thanks, looks fine, but all the version bumps only need to be PATCH bumps (so cachedump doesn't need changing at all)...

volatility3/framework/plugins/windows/cachedump.py
atcuno: Hmm ok. I added typing on classmethods, which in the past you said needed major bumps. Should it be only major if the parameters actually change then? I am fine with undoing the bumps, just want to understand.

atcuno: ahh nevermind I get it.

atcuno: ok I reverted the bad version changes and req change.

volatility3/framework/plugins/windows/hashdump.py
```diff
@@ -21,7 +21,7 @@ class Hashdump(interfaces.plugins.PluginInterface):
     """Dumps user hashes from memory"""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 1, 0)
```
ikelos, line 24: Why is this one getting bumped?  It should also be a PATCH bump (but please do bump the patch versions, for all the ones that change, so we can tell if the change we made goes wrong).

atcuno: Ok this is just patch bump now for hashdump.py

atcuno:
ok @ikelos i think I have it right now with just the PATCH bumps for each.

volatility3/framework/plugins/windows/hashdump.py
ikelos: I mean... you can but there's no need.  Dependencies changing should require a bump of the plugin version (because the plugin won't necessarily know it externals have changed), but as I say, not a problem, better to bump then all than none of them...  5:)

volatility3/framework/plugins/windows/cachedump.py
```diff
@@ -22,7 +22,7 @@ class Cachedump(interfaces.plugins.PluginInterface):
     """Dumps lsa secrets from memory"""
 
     _required_framework_version = (2, 0, 0)
-    _version = (1, 0, 1)
+    _version = (1, 0, 2)
```
ikelos, line 25: Sorry, meant this comment for this change...

I mean... you can but there's no need.  Dependencies changing shouldn't require a bump of the plugin version (because the plugin won't necessarily know it externals have changed), but as I say, not a problem, better to bump then all than none of them...  5:)

PR #1683, opened 2025-03-07
Windows: DeviceTree traversal/offset bugfixes

dgmcdona:
There are two issues with the current implementation of the `windows.devicetree.DeviceTree` plugin: 
- The traversal of the device tree is incomplete. The `DEVICE_OBJECT.get_attached_devices` extension method does not traverse the full breadth at each level of the device tree through the `NextDevice` linked-list.
- Every entry uses the driver's virtual address as the offset. Device entries should yield their own virtual addresses in the generator.

This PR fixes the issue by implementing a recursive, private staticmethod on the `DeviceTree` plugin class that does a complete depth-first traversal of the tree, yielding many more device objects (using their virtual address as the `Offset` member instead of that of the parent driver).

### DIFF
--- a/volatility3/framework/plugins/windows/devicetree.py
+++ b/volatility3/framework/plugins/windows/devicetree.py
@@ -3,12 +3,12 @@
 #
 
 import logging
+from typing import Iterator, List, Set, Tuple
 
-from typing import Iterator, List, Tuple
-
-from volatility3.framework import constants, renderers, exceptions, interfaces
+from volatility3.framework import constants, exceptions, interfaces, renderers
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints
+from volatility3.framework.symbols.windows.extensions import DEVICE_OBJECT
 from volatility3.plugins.windows import driverscan
 
 DEVICE_CODES = {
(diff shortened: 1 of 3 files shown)


volatility3/framework/plugins/windows/devicetree.py
```diff
@@ -186,6 +147,48 @@ def _generator(self) -> Iterator[Tuple]:
                 )
                 continue
 
+    @staticmethod
```
ikelos: Please don't use staticmethod inside versionable objects (even for private methods) because it prevents us being able to get back to the version from just the method.

ikelos (commented):
Looks ok, but please no static methods in classes that are versionable.  Also, I'd tend to agree with the scanner and suggest that at least a comment explaining why the exception can be ignored should be included...

dgmcdona:
@ikelos One consideration for this should be what to do about the `get_attached_devices` method on the `DEVICE_OBJECT` extension class. It's current implementation is flawed, which is why I moved away from it in this PR, but maybe we should deprecate + remove or fix it as part of this effort. I would have fixed it and used it here but it would have required a method signature change in order to also return the depth in the device tree that we need when constructing the `TreeGrid`.

ikelos:
I'd pick a different method name, and run them alongside each other (deprecating the old one)?  Eventually, if it's really worth it, you can shift the new one back to the old name.

If the code belongs centrally, it's much better to have it live there than in a plugin where someone may not know about it and try to reimplement (possibly incorrectly).  If it's only for the generator (ie, not to be used by other plugins) then it's fine.  You could also consider a visitor design pattern, that takes a function and runs it on each node in the tree in a particular order, calling it with the node and the depth at each point?

dgmcdona:
Thanks, marking as draft again while I revisit this.

ikelos:
I didn't think this was too far off prime time?  Could you dust it off and see what would be needed to get it flying again please?

PR #1684, opened 2025-03-08
Prevent truecrypt from throwing a backtrace when the module isn't fou…

atcuno:
…nd and print a warning.

This threw a huge backtrace on any sample that didn't have truecrypt loaded (99.999999% of them).


ikelos (approved):
Hehehe, good catch, thanks (I'll let it slide this time, but be good to get in the habit of bumping the PATCH version please)...

PR #1686, opened 2025-03-08
Update fbdev plugin to gracefully handle old/broken symbol tables

atcuno:
This plugin backtraced when 1) fbdev was compiled as a module and 2) an old version of dwarf2json was used to generate the symbol table.

This caused the existing check for the symbol to exist to succeed, but its only as a forward declaration. Then `object_from_symbol` would fail as the address was recorded as 0.

My patch here makes it gracefully warn the user + exit when an old symbol table is passed, and then the existing check will work for modern symbol tables and samples with fbdev as a kernel module.

### DIFF
--- a/volatility3/framework/plugins/linux/graphics/fbdev.py
+++ b/volatility3/framework/plugins/linux/graphics/fbdev.py
@@ -262,7 +262,7 @@ You can try using ffmpeg to decode the raw buffer. Example usage:
 
         if not kernel.has_symbol("num_registered_fb"):
             vollog.error(
-                '"num_registered_fb" symbol does not exist in the symbol table. This means you are either analyzing an unsupported kernel version,  your symbol table is corrupt, or the fbdev driver is compiled as a kernel module..'
+                '"num_registered_fb" symbol does not exist in the symbol table. This means you are either analyzing an unsupported kernel version,  your symbol table is corrupt, or the fbdev driver is compiled as a kernel module.'
             )
             return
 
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/linux/graphics/fbdev.py
```diff
@@ -255,22 +255,28 @@ def _generator(self):
...
-                "The provided symbol does not exist in the symbol table. This means you are either analyzing an unsupported kernel version or that your symbol table is corrupt.",
+            vollog.error(
+                '"num_registered_fb" symbol does not exist in the symbol table. This means you are either analyzing an unsupported kernel version,  your symbol table is corrupt, or the fbdev driver is compiled as a kernel module.'
+            )
+            return
+
+        try:
+            num_registered_fb = kernel.object_from_symbol("num_registered_fb")
+        except exceptions.SymbolError:
+            vollog.error(
```
ikelos, line 272: Do we know the exact lowest required version number of dwarf2json?  I think we've got code that can check and warn on such things? (see [here](https://github.com/volatilityfoundation/volatility3/blob/d13cf7ae2205584ec89dbc5088d0c5d38fc9acbc/volatility3/framework/plugins/linux/kmsg.py#L563) )

ikelos (commented):
Generally just looks fine, but may be better to provide a warning at the top before anything gets going just to save them some time?

volatility3/framework/plugins/linux/graphics/fbdev.py
atcuno: hmm the thing there is that, if the fbdev driver is compiled into the kernel, then the plugin will work with the old symbols tables. The `object_from_symbol` call only fails when both conditions are met (compiled as module + old symbol table).

ikelos: Ok, so then the check belongs in here?  The point is we should be able to tell whether their symbol table is too old or if that's not the problem?

Abyss-W4tcher:
Thanks for spotting this issue! Once the btf2json framework integration is implemented, I think we'll be able to obtain the required module symbols directly from memory and handle that case properly :)

PR #1689, opened 2025-03-08
Tweak complex plugin documentation

j-t-1:


j-t-1:
> At the moment, the main information that layers may populate are:
> 
> * `os` with values of `Windows`, `Linux`, `Mac` or `unknown`
> * `architecture` with values of `Intel32`, `Intel64` or `unknown`

Are there now more architectures?

ikelos:
> > At the moment, the main information that layers may populate are:
> > 
> > * `os` with values of `Windows`, `Linux`, `Mac` or `unknown`
> > * `architecture` with values of `Intel32`, `Intel64` or `unknown`
> 
> Are there now more architectures?

Not as of yet, but there's a pull request to get arm64 in, which unfortunately is a big task to review and test properly.

doc/source/complex-plugin.rst
```diff
@@ -76,8 +76,7 @@ Writing Scanners
 Scanners are objects that adhere to the :py:class:`~volatility3.framework.interfaces.layers.ScannerInterface`.  They are
 passed to the :py:meth:`~volatility3.framework.interfaces.layers.TranslationLayerInterface.scan` method on layers which will
 divide the provided range of sections (or the entire layer
-if none are provided) and call the :py:meth:`~volatility3.framework.interfaces.layers.ScannerInterface`'s call method
-method with each chunk as a parameter, ensuring a suitable amount of overlap (as defined by the scanner).
+if none are provided) and call the :py:meth:`~volatility3.framework.interfaces.layers.ScannerInterface`'s call method with each chunk as a parameter, ensuring a suitable amount of overlap (as defined by the scanner).
```
ikelos: I can't see what's changed here, it looks like you've just removed the line break?  They _shouldn't_ show up in the final document, but they are useful to keep the text visible within an editor.  If you didn't have a specific reason for removing the break here, could you add it back please?

ikelos (commented):
Thanks, generally all looks fine, but there's one change that appears to just be removing a line break, and that should have no effect on the output, but does affect readability within an editor.  If we can get the break added back in then happy for it to get merged, thanks!

doc/source/complex-plugin.rst
j-t-1: There is _method method_ in the original, and I have removed one of them. Restored the line break.

ikelos: Oh, cool, good catch (I even missed it a second time!)  Thanks for putting the line break back in!  5:D

ikelos (approved):

PR #1690, opened 2025-03-08
Convert poolscanning APIs to new coding standards

### LINKED ISSUE #1646: Convert pool scanner API to only send kernel name for kernel objects
`generate_pool_scan` currently takes the layer name and symbol name as parameters even though, until the GUI work, all scanners targeted the kernel.

After the GUI PR merge, we need to do the following:

1) Convert `generate_pool_scan` to just take in the kernel module name
2) Convert `generate_pool_scan_extended` to just take the kernel module name for the kernel parts
3) Update the existing callers to use the new method 
4) All the required version bumps (major poolscanner bump, minor plugins, major any APIs)

atcuno:
This PR converts the poolscanner APIs to the new coding standards and changes the minimum amount of code paths necessary to get existing callers up to date.

Part of these changes ended up being a large overhaul of the `info` plugin as its methods were called in these paths, and the `info` plugin has some really crazy design like exposing a method to essentially rebuild the kernel module using its layer and symbol table. The callbacks plugin also had the old methods in its enumeration functions, which required update so that `scan` could use the proper method. 

Overall, it was a pretty clean switch over though. I tested every plugin across ~60 Windows samples last night and no issues were found.

Fixes #1646

### DIFF
--- a/volatility3/framework/plugins/windows/psxview.py
+++ b/volatility3/framework/plugins/windows/psxview.py
@@ -11,7 +11,6 @@ from volatility3.framework.renderers import TreeGrid, format_hints
 from volatility3.framework.symbols.windows import extensions
 from volatility3.plugins.windows import (
     handles,
-    info,
     pslist,
     psscan,
     thrdscan,
@@ -50,13 +49,10 @@ We recommend using -r pretty if you are looking at this plugin's output in a ter
                 architectures=["Intel32", "Intel64"],
             ),
             requirements.VersionRequirement(
-                name="info", component=info.Info, version=(1, 0, 0)
-            ),
-            requirements.VersionRequirement(
                 name="pslist", component=pslist.PsList, version=(3, 0, 0)
             ),
             requirements.VersionRequirement(
-                name="psscan", component=psscan.PsScan, version=(1, 0, 0)
+                name="psscan", component=psscan.PsScan, version=(2, 0, 0)
             ),
             requirements.VersionRequirement(
                 name="thrdscan", component=thrdscan.ThrdScan, version=(1, 0, 0)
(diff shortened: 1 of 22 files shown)


volatility3/framework/plugins/windows/psxview.py
```diff
@@ -114,10 +110,10 @@ def _check_pslist(self, tasks):
         return self._proc_list_to_dict(tasks)
 
     def _check_psscan(
-        self, layer_name: str, symbol_table: str
+        self,
     ) -> Dict[int, extensions.EPROCESS]:
         res = psscan.PsScan.scan_processes(
-            context=self.context, layer_name=layer_name, symbol_table=symbol_table
+            context=self.context, kernel_module_name=self.config["kernel"]
```
ikelos, line 116: Hmmm, this is ok, but will make it harder to turn this into a classmethod if we ever want in the future.  If we were going to do that we'd need to take in context as well, so probably not worth changing, just something to keep in mind.  I think this is one of the very few plugins left that have actual non-classmethod methods...

ikelos (commented):
Welp, we're going to make any consumers of these plugins very unhappy, but it is in the effort of consistency, and the changes are clear at least.  I can't tell how I feel about maintaining some methods as not fully parameterized (using self instead of taking in the parameter), but I'll go with it if you think that's fine.  I'm trusting that all the cross-dependencies (even from plugins that haven't otherwise changed) have been updated and nothing should break when this gets merged?

Lemme know you're happy before I hit the button please (going back from this will be even more painful).

volatility3/framework/plugins/windows/psxview.py
atcuno: Yea, I didn't want to rewrite psxview so extensively as its essentially calling the proces enumeration functions anyway then doing whatever conversion is needed to get a EPROCESS from it. Making psxview have classmethods for that would duplicate what the plugins are providing anyway, so keeping it how it is seems best. And yes, I am happy with the other changes and they have been tested with running all the plugins over many samples.

PR #1691, opened 2025-03-08
Remove deepcopy call in pe_symbols. Validate wanted symbol informatio…

atcuno:
…n passed through resolving functions

The entry point to the issue in https://github.com/volatilityfoundation/volatility3/pull/1674

Was that the deepcopy() call was receiving Pointers and then the `self.vol.type_name` was going off the deep end, hence 1674.

This PR updates pe_symbols to avoid the deep copy call, so that instead of just accepting anything in the set of symbols to resolve, we actually validate each entry before resolving it.

This only required dependecy updates in the core plugins as they already call the API correctly, but will ensure that future plugins do not have mystery errors or missing data.

### DIFF
--- a/volatility3/framework/plugins/windows/pe_symbols.py
+++ b/volatility3/framework/plugins/windows/pe_symbols.py
@@ -671,8 +671,9 @@ class PESymbols(interfaces.plugins.PluginInterface):
                         else:
                             yield symbol_key, value_index, symbol_value, wanted_value  # type: ignore
 
-    @staticmethod
+    @classmethod
     def _validate_wanted_modules(
+        cls,
         wanted: PESymbolFinder.cached_module_lists,
     ) -> Optional[PESymbolFinder.cached_module_lists]:
         """
--- a/volatility3/framework/plugins/windows/pe_symbols.py
+++ b/volatility3/framework/plugins/windows/pe_symbols.py
@@ -1,7 +1,6 @@
 # This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 
-import copy
 import io
 import logging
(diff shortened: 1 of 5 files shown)


volatility3/framework/plugins/windows/pe_symbols.py
```diff
@@ -671,6 +671,52 @@ def _get_symbol_value(
                         else:
                             yield symbol_key, value_index, symbol_value, wanted_value  # type: ignore
 
+    @staticmethod
```
ikelos: No staticmethod in versioned classes please, not even this one...  5:)

volatility3/framework/plugins/windows/pe_symbols.py
```diff
@@ -671,6 +671,52 @@ def _get_symbol_value(
...
+    def _validate_wanted_modules(
+        wanted: PESymbolFinder.cached_module_lists,
```
ikelos:
```suggestion
    @classmethod
    def _validate_wanted_modules(
        cls,
        wanted: PESymbolFinder.cached_module_lists,
```

ikelos (approved):
Looks good, might be nice to document somewhere exactly what the issue was somewhere, but happy to get this in before that.  Just one staticmethod to swap out for a classmethod please.

volatility3/framework/plugins/windows/pe_symbols.py
atcuno: Just did the web interface autocommit for that. Check all my comments here for the base issue: https://github.com/volatilityfoundation/volatility3/pull/1674

ikelos:
Still want this merging even after #1674 ?

atcuno:
Yes still merge as otherwise callers will be missing data if they send in bad information, without a warning.. whereas this gives a very explicit ValueError thrown on bad data coming to the API.

PR #1692, opened 2025-03-08
Fix volatility object deepcopy

ikelos:
Root cause of several issues revolving around deepcopy...


ikelos:
Outdone by #1674  5:D

PR #1694, opened 2025-03-09
Remove comment

j-t-1:
These mypy issues are now closed.

### DIFF
--- a/volatility3/framework/plugins/linux/pslist.py
+++ b/volatility3/framework/plugins/linux/pslist.py
@@ -85,6 +85,7 @@ class PsList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
         Returns:
             Function which, when provided a process object, returns True if the process is to be filtered out of the list
         """
+        pid_list = pid_list or []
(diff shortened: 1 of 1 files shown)


j-t-1:
This errors with `filter_list = [x for x in pid_list if x is not None]\nTypeError: \'NoneType\' object is not iterable\n'`.

I will amend this to just remove the comment. Because the mypy issues are closed this seems okay (to remove the comment)?

ikelos:
Did you check whether mypy actually still complains about this, or just that the upstream tickets were marked as closed?

j-t-1:
> Did you check whether mypy actually still complains about this, or just that the upstream tickets were marked as closed?

I just checked the uptream tickets are closed...

ikelos:
Yeah, fair enough.  I think the code is reasonable and won't flag any "unnecessary code" warnings, so we can probably ditch the comment.  Thanks...  5:)

j-t-1:
The PR that closes them I think is what the Volatility code comment was referring. Will continue to get a fuller understanding before removing the equivalent comments in mac and windows.

PR #1697, opened 2025-03-10
Ensure that new symbol tables are only created when needed. Avoid nam…

atcuno:
…e collisions leading to inconsistent behaviour. Make API for symbol table acquisition sane.

This PR fixes a very nasty bug that @dgmcdona 's comparison harness detected between Vol2 and Vol3. In particular, it was reported that Vol2 was recovering names of drivers inside of devicetrees that Vol3 was reported as not available (paged out).

I debugged it today and the root cause is the API fixed here - `get_pool_header_table` - which had a really bad issue. 

The purpose of this function is to return the symbol table name that the pool scanners should use when scanning (really just one with POOL_HEADER), but the function was receiving the name of a symbol table and then checking the global symbol space to see if it had the type. For kernels with the POOL_HEADER inside the regular types or existing tables with it, this led to getting back the same table again. After that, the code would then make a module() with this name at offset 0. Having an offset at 0 completely breaks operations of getting names out of POOL_HEADERs as they use relative offsets.

Figuring out all this out was very complicated and very hard to debug as we tried git bisect, but we were pointed to a previous PR that broke it, but then realized that Vol3 has had on and off again support all along the way - a situation that the automated regression tests and harness will catch much quicker going forward.

The reason for the back and forth between working or not, was that, for some versions of Vol3, the symbol table named passed into the pool scanners (always `symbol_table_name1` in testing),  was already defined as a module (context.modules[] element) and already had the right kernel offset set. In these cases, the broken `get_pool_header_table` had no effect as the name was already used. In the versions of Vol3 where a previous `symbol_table_name1` was not set due to code changes over time, then the pool scanners broke when reconstructing names as they were constructing the module at kernel offset 0.

This PR fixes all the above - directly using the kernel when it has the symbol, and only going to json files when needed, plus always constructing the module at the kernel offset as the plugins expect. 

Devicetree matches Vol2 across samples with this code in place.

### DIFF
--- a/volatility3/framework/plugins/windows/callbacks.py
+++ b/volatility3/framework/plugins/windows/callbacks.py
@@ -42,7 +42,7 @@ class Callbacks(interfaces.plugins.PluginInterface):
                 name="ssdt", plugin=ssdt.SSDT, version=(2, 0, 0)
             ),
             requirements.PluginRequirement(
-                name="poolscanner", plugin=poolscanner.PoolScanner, version=(2, 0, 0)
+                name="poolscanner", plugin=poolscanner.PoolScanner, version=(3, 0, 0)
             ),
             requirements.PluginRequirement(
                 name="driverirp", plugin=driverirp.DriverIrp, version=(1, 0, 0)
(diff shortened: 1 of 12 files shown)


ikelos (approved):
Good work!  I guess it's a facet of using the moduling system to do all the symbol stuff.  Previously (and here) it was expected that you'd generally always be working in absolute offsets (and the symbol table was just a symbol table), but I guess we now need to be careful about modules.  It might be worth a scan through the code (or a developer warning message) to look for places that a module is constructed with an offset of 0.  It's still usable as a symbol table holder, but it may be confusing if people start trying to create objects from it (or maybe just something when creating an object in a module that has an offset of 0?)...

PR #1701, opened 2025-03-10
1476 all calls to get name on registry keys need auditing

### LINKED ISSUE #1476: All calls to `get_name` on registry keys need auditing
The scheduled task plugin crashed on many samples here:

```
24-12-26 13:29:01 volatility3.cli DEBUG    Traceback (most recent call last):
    renderer.render(grid)
    grid.populate(visitor, outfd)
  File "/home/ub/volatility3/volatility3/framework/plugins/windows/scheduled_tasks.py", line 1403, in <genexpr>
    (
  File "/home/ub/volatility3/volatility3/framework/plugins/windows/scheduled_tasks.py", line 1381, in _generator
    for task in self._parse_task_key(key, task_name_map):
  File "/home/ub/volatility3/volatility3/framework/plugins/windows/scheduled_tasks.py", line 1241, in _parse_task_key
    task_name = guid_mapping.get(str(key.get_name()), renderers.NotAvailableValue())
                                     ^^^^^^^^^^^^^^
  File "/home/ub/volatility3/volatility3/framework/symbols/windows/extensions/registry.py", line 251, in get_name
    return self.Name.cast("string", max_length=namelength, encoding="latin-1")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

We need to fix this one + audit all calls to `get_name` to ensure they are in a try/except

superponible:
@ikelos ready for review, mostly just adding a bunch of exceptions that weren't being caught.


volatility3/framework/plugins/windows/envars.py
```diff
@@ -71,13 +71,25 @@ def _get_silent_vars(self) -> List[str]:
                 sys = hive.get_key(
                     "CurrentControlSet\\Control\\Session Manager\\Environment"
                 )
-            except (KeyError, registry.RegistryFormatException):
-                with contextlib.suppress(KeyError, registry.RegistryFormatException):
+            except (
+                KeyError,
+                registry.RegistryFormatException,
```
ikelos: Should both of these exceptions inherit from a top level exception (`registry.RegistryException`) that then is made more specific?  Seems like it'll be useful in most instances?

volatility3/framework/plugins/windows/prefetch.py
```diff
@@ -0,0 +1,535 @@
+# References :
```
ikelos: This should really be a separate PR please...  5:S  I've let this kind of stuff slide, but we *really* need to get in the habit of separating these things out, even if one depends on the other.  This PR is about fixing up the Registry Exceptions, not about adding the prefetch plugin...

ikelos (changes requested):
So one comment on the registry key audits, but the whole prefetch plugin needs ripping out and putting in a different PR.

I haven't reviewed it because it doesn't belong in this PR (although I have some questions about why it doesn't use `dataclasses` and why there are just some random methods outside of the plugin and thus not versioned).

This needs some changes before it can be merged.

volatility3/framework/plugins/windows/prefetch.py
superponible: that's my bad, it's something I had started on a while ago and just had the changed code but added it by accident.  it shouldn't be here as it probably doesn't even work.

volatility3/framework/plugins/windows/envars.py
superponible: Yeah, I had wondered about that too when I started making the changes.  It'd definitely be cleaner, I can switch it over to RegistryException

ikelos: Sounds like a plan, thanks.  5:)

volatility3/framework/plugins/windows/prefetch.py
ikelos: All good, it was just github hides it as being a big change by default so easy to miss.  Pull it out and the rest is pretty much good to go once the registry exceptions have been combined.  5:)

volatility3/framework/plugins/windows/envars.py
superponible: ok, the latest has RegistryException

ikelos (approved):
Looks good, thanks!  5:)

PR #1702, opened 2025-03-10
#1645 - duplicate list_processes code

### LINKED ISSUE #1334: Registry layer relies on standard plugin
**Describe the bug**
A core part of the framework (registry layer) relies on pslist to figure out the process that the registry lives in.  This is an awkward introduces a lot of cross-dependency that shouldn't exist.  We should try to find a way to achieve the same goal without pulling in so much extra gubbins just to find the process list.  I'd almost be happy duplicating code, or storing it in a separate common module (perhaps the generic process work from #1000 could help)?

See [#1307](https://github.com/volatilityfoundation/volatility3/pull/1307#discussion_r1817447132) for more details.

### LINKED ISSUE #1645: Change registry processing to not walk the process list in the registry layer
@superponible this is right up your alley..

See the discussion here: https://github.com/volatilityfoundation/volatility3/pull/1640#discussion_r1972609170

superponible:
@ikelos I think what you wanted with duplicating the list_processes code is done.  Still not sure this does what you're asking with what was "kernel_module_name". Let me know if it still doesn't look right and if you have any thoughts on it.

### DIFF
--- a/volatility3/framework/layers/registry.py
+++ b/volatility3/framework/layers/registry.py
@@ -14,7 +14,7 @@ from volatility3.framework.configuration.requirements import (
 from volatility3.framework.exceptions import InvalidAddressException
 from volatility3.framework.layers import linear
 from volatility3.framework.symbols import intermed
-from volatility3.plugins.windows import pslist
+from volatility3.framework.symbols.windows import extensions
 
 vollog = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


volatility3/framework/layers/registry.py
```diff
@@ -96,6 +95,41 @@ def __init__(
...
 
+    def _find_registry_process(self) -> Optional["extensions.EPROCESS"]:
+        """Walk the active process list and return the Registry process if it exists. Duplicates
+        PsList.list_processes() since pulling in the plugin causes problems.
+
+        Returns:
+            The Registry EPROCESS object if it exists, or None
+        """
+
+        kernel = self.context.modules.get("kernel")
```
ikelos, line 106: This shouldn't be hard coded as `kernel`, there's no guarantee what it'll be called.  So this should be `kernel_module_name`, although I'm just looking for if there's a nice way to get that from the information you've already got access to...

ikelos:
```suggestion
        kernel = self.context.modules.get(self.config["kernel_module_name"])
```
This shouldn't be hard coded as `"kernel"`, there's no guarantee what it'll be called. So this should be possible to pull from the config (as that's how it used to work).  I don't _think_ there's a nicer way of getting to it, but at least this won't need to import pslist now...

volatility3/framework/layers/registry.py
```diff
@@ -96,6 +95,41 @@ def __init__(
                 f"Exception when setting hive {self.name} max address, using {hex(self._maxaddr)}",
             )
 
+    def _find_registry_process(self) -> Optional["extensions.EPROCESS"]:
```
ikelos: Since you actually import extensions, this probably doesn't need to be in `"`.  You can try removing the quotes, but if that fails, it's fine to leave them in.

ikelos (commented):
Yep, that's very much what I was after, it's no longer importing a plugin, but just something fixed in the main framework.  Thanks!  Just one small point about the hardcoded `kernel` name (and some quotes that probably don't have to be there), but otherwise looks good!  5:)

superponible:
> Yep, that's very much what I was after, it's no longer importing a plugin, but just something fixed in the main framework. Thanks! Just one small point about the hardcoded `kernel` name (and some quotes that probably don't have to be there), but otherwise looks good! 5:)

those are both fixed.

PR #1704, opened 2025-03-10
No longer allocate session modules in support of the GUI objects at o…

atcuno:
…ffset 0. Instantiate objects at the absolute address explicitly.

This removes an instance of the pattern we changed yesterday of where modules get created at offset 0. The GUI subsystem support used this to create scanned for objects with the gui symbol table and the session layer the object was found in. Since this all happened through one call, we just have to set absolute=True to get it correct.


ikelos (approved):
That seems good, thanks.

PR #1706, opened 2025-03-11
#1446 - pid is objects.Pointer, not int

### LINKED ISSUE #1446: pstree.py has broken types and `pid` handling
This is code is confusing as it expects a two element tuple even though the default is None:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/pstree.py#L64

I was trying to understand `pid` but its typing in the function parameters shows:

`pid: objects.Pointer,`

But then `self._processes` is declared as:

`self._processes: Dict[int, Tuple[interfaces.objects.ObjectInterface, int]] = {}`

And it definitely seems to store ints:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/pstree.py#L100

@superponible  can you take a look at this one?

superponible:


ikelos (commented):
I'd probably just change the signature of `find_level` to be `int`?  We don't use the "pointerish" ness of the pid anywhere, and in fact, it's a fake pointer (since it is actually just a number).  Also, would be good to get better typing information on `yield_processes` please (which should also have `pid` as an `int`...

superponible:
@ikelos I switched them all to int

PR #1707, opened 2025-03-12
1471 registry plugins spread across directories needs standardization

### LINKED ISSUE #1471: Registry plugins spread across directories. Needs standardization
We have a sub directory 'registry' under `framework/plugins/windows` that holds two registry related plugins.

We then have amcache (https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/amcache.py), cachedump (https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/cachedump.py),  lsadump (https://github.com/volatilityfoundation/volatility3/blob/b1a42d9510046377a426d62f7cbde1ce1a4ad528/volatility3/framework/plugins/windows/lsadump.py), hashdump (https://github.com/volatilityfoundation/volatility3/blob/b1a42d9510046377a426d62f7cbde1ce1a4ad528/volatility3/framework/plugins/windows/hashdump.py#L330), and maybe others which use the registry, outside the folder.

There are 20+ new and useful plugins that can be written that target the registry, and I was talking to some college students about them as an easy intro to learning the framework - but then I realized we have plugins in multiple places and need to standardize this before it gets worse and before parity.

@ikelos @superponible - I assume these should all be under registry? What needs to happen version wise when we move plugins? These plugins also triggered many bugs in mass testing so we need to get them moved rather quickly so the bugs can be then be patched also.

superponible:
@ikelos I kept the versions as they were like you mentioned in your comment https://github.com/volatilityfoundation/volatility3/issues/1471#issuecomment-2565763558, I also thought that made more sense.

If you just run with a plugin name like "hashdump" you get the multiple matches error:

vol.py: error: argument PLUGIN: plugin hashdump matches multiple plugins (windows.hashdump.Hashdump, windows.registry.hashdump.Hashdump)

So you can run the new plugin with "registry.hashdump" but then "windows.hashdump" would also work until we remove those deprecated one.

Only issue is I don't see the warning get printed, and I'm not sure how to make it happen.  Any thoughts?


volatility3/framework/plugins/windows/amcache.py
```diff
@@ -1,651 +1,23 @@
...
-            }
-
-            program_id = _get_string_value(values, val_enum.ProgramID.value)
-            path = _get_string_value(values, val_enum.Path.value)
-            company = _get_string_value(values, val_enum.Company.value)
-            last_mod_time = _get_datetime_filetime_value(
-                values, val_enum.LastModTime.value
+    def __getattr__(self, *args, **kwargs):
+        warnings.warn(
+            DeprecationWarning(
```
ikelos: `DeprecationWarning`s only get seen by developers who run python with a special flag, I think it needs to be `FutureWarning` (see [here](https://docs.python.org/3/library/exceptions.html#FutureWarning) for more information).

ikelos (commented):
Cool, glad it all works (save for the warning).  I think that just needs to be `FutureWawrning` and then it should all be good...  5:)

volatility3/framework/plugins/windows/registry/scheduled_tasks.py
```diff
@@ -0,0 +1,1431 @@
...
+        cls, reader: _ScheduledTasksReader, version: int
+    ) -> Optional["TaskAction"]:
+        command = reader.read_bstring()
+        args = reader.read_bstring()
+        if command is None or args is None:
+            return None
+
+        workdir = reader.read_bstring()
+        if version == 3:
+            _flags = reader.read_u2()
```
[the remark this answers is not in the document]
ikelos: Eh, this is fine if you want to, but I don't mind it the other way, your call...

volatility3/framework/plugins/windows/amcache.py
superponible: @ikelos So I checked again and FutureWarning does print the message if I change it to override `run` instead of `__getattr__`, but I think you had suggested getattr so people would see it if they imported the file and used a classmethod rather than calling the plugin directly.

superponible: using `__getattribute__` instead of getattr works for the command line, but still not when I import one in the windows directory, i.e. in registry.cachedump, do `from volatility3.plugins.windows import hashdump` instead of the one in windows.registry.

ikelos:
Ok, so `__getattr__` is only called when an attribute can't be found.  Since this inherits, all the attributes can be found, so it either needs to not inherit (and call the plugins version of the method), or `__getattr__` needs to become `__getattibute__` (both as the method name, and the call to super).

volatility3/framework/plugins/windows/cachedump.py
```diff
@@ -1,187 +1,23 @@
...
-            if uname_len == 0 or len(ch) == 0:
-                continue
-            dec_data = self.decrypt_hash(enc_data, nlkm, ch, not vista_or_later)
-
-            (username, domain, domain_name, hashh) = self.parse_decrypted_cache(
-                dec_data, uname_len, domain_len, domain_name_len
+    def __getattr__(self, *args, **kwargs):
+        warnings.warn(
+            FutureWarning(
+                "This plugin is now called windows.registry.cachedump.Cachedump"
```
ikelos: Might also be worth letting people know this will go away (so they have a reason to change) and including a deprecation date (six months plus a week seems to be what @atcuno 's been going with, so happy to standardize on that).

superponible: Yeah, I mentioned in another comment that I figured out if I changed it to `__getattribute__`, it printed on the command line if you called the deprecated version directly (`vol.py ... windows.hashdump`), but if I modify cachedump to import windows.hashdump instead of windows.registry.hashdump (to simulate somone writing code that imports the deprecated file), then when it calls the hashdump.get_bootkey() classmethod, no warning is printed, which is why you had originally suggested using getattr.

I pushed up a version with getattribute and the warning date. Not sure if you have any ideas on getting the message if the old code is imported. If not then I guess this is ready.

volatility3/framework/plugins/windows/amcache.py
```diff
@@ -14,8 +14,11 @@ class Amcache(amcache.Amcache):
     _required_framework_version = (2, 0, 0)
     _version = (2, 0, 0)
 
-    def __getattr__(self, *args, **kwargs):
+    def __getattribute__(self, *args, **kwargs):
         warnings.warn(
-            FutureWarning("This plugin is now called windows.registry.amcache.Amcache")
+            FutureWarning(
+                "The windows.amcache.Amcache plugin is deprecated and will be removed on "
+                "September 19, 2025. Use windows.registry.amcache.Amcache instead."
```
ikelos: Sorry, don't mean to be picky, but can we use 2025-09-15, just to keep it consistent with the other deprecation messages (and make it a bit easier to grep for)?

ikelos:
Looking really good.  Just a reformat of the date (if you're happy to do so) and then it's good to go in!  5:)

volatility3/framework/plugins/windows/cachedump.py
ikelos: Hmmmm, that's a little strange?  I guess you could implement both `getattribute` and `getattr`, but I dunno why it would need to be looking that up (and I thought that would pass through `getattribute` anyway?)

ikelos: I guess the classmethod decorator falls back to the class it's actually defined in?  I'd still expect the lookup to get called though?

ikelos:
Ok, found a solution, but can't tell how filthy it is...  5:S  Also can't tell with `__func__` is still a valid way of referring to a classmethod's method (it seems a `__wrapped__` was introduced in 3.9 but then deprecated again in 3.11 for removal in 3.13)...  5:S

```
def renamed_class(deprecated_class_name: str, message: str, removal_date: str):
    """A decorator for marking classes as being renamed and removed in the future.
       Callers to this function should explicitly update to use the other plugins instead.

    Args:
        deprecated_class_name: The name of the class being deprecated
        message: A message added to the standard deprecation warning. Should include the replacement API paths
        removal_date: A YYYY-MM-DD formatted date of when the function will be removed from the framework
    """

    def decorator(replacement_func):
        @functools.wraps(replacement_func)
        def wrapper(*args, **kwargs):
            warnings.warn(
                f"This plugin ({deprecated_class_name}) has been renamed and will be removed in the first release after {removal_date}. {message}",
                FutureWarning,
            )
            return replacement_func(*args, **kwargs)

        return wrapper

    return decorator


class PluginRenameClass:
    """Class to move all classmethod invocations (for when a plugin has been moved)"""

    def __init_subclass__(cls, replacement_class, **kwargs):
        deprecated_class_name = f"{cls.__module__}.{cls.__qualname__}"
        super().__init_subclass__(**kwargs)
        for attr, value in replacement_class.__dict__.items():
            if isinstance(value, classmethod):
                setattr(
                    cls,
                    attr,
                    classmethod(
                        renamed_class(
                            deprecated_class_name=deprecated_class_name,
                            removal_date="2025-09-14",
                            message=f"Please ensure all method calls to this plugin are replaced with calls to {replacement_class.__module__}.{replacement_class.__qualname__}",
                        )(value.__func__)
                    ),
                )
            else:
                setattr(cls, attr, value)
        return super(replacement_class).__init_subclass__(**kwargs)


class Hashdump(
    interfaces.plugins.PluginInterface,
    PluginRenameClass,
    replacement_class=hashdump.Hashdump,
):
    """Dumps user hashes from memory (deprecated)"""

    _version = (1, 1, 1)
```

Anyway, seems to achieve what we want, probably worth adding the inherited class and decorator to the deprecation module and then the deprecated plugin needs to inherit from the ancestor class of the target plugin's class (in most cases this'll be `interfaces.plugins.PluginInterface`).  This doesn't deprecate run or generator (non-class-methods), but it assumes those will call one of the classmethods too, so should still show the warning.  Seemed to trigger the warning twice in my tests, not quite sure why but also... still classing it as a win...  5;P

volatility3/framework/plugins/windows/amcache.py
superponible: No problem, that's done. One other question.  Is there a reason the certificates.py is in volatility3/plugins/windows/registry and not volatility3/framework/plugins/windows/registry? Should we move it into the framework directory as part of this?

ikelos: Yeah, it was supposed to be an example of a contributed/not supported plugin.  To show people where they're supposed to put plugins they write.  There's another plugin lurking under there too.  I dunno if anyone ever really realized, but I'd like to keep them there, since the intention was people would put their plugins in the main plugin directory, not the core plugins (they can overload core plugins that way too, since it takes precedence).

volatility3/framework/plugins/windows/cachedump.py
ikelos: See my comment below for something that works (although, hard to tell how hacky it is, probably needs checking on a bunch of versions of python)...

superponible:
> Ok, found a solution, but can't tell how filthy it is... 5:S Also can't tell with `__func__` is still a valid way of referring to a classmethod's method (it seems a `__wrapped__` was introduced in 3.9 but then deprecated again in 3.11 for removal in 3.13)... 5:S

@ikelos I tried working this in, but having problems. I started trying to add it to windows.amcache and had two issues. First, windows.amcache is no longer found, only windows.registry.amcache, and second, the warning was printing when I ran registry.amcache, and it also printed when I ran registry.cachedump, even though I hadn't modified that yet.

ikelos:
Ok, the machinery shouldn't be touching the path for the new classes, so if it's tripping when they're run it sounds like something wasn't setup properly.  Could you spin up a branch so I can take a look over it please?  The new class should have all the code, and the old class location should just be a husk inheritting from PluginRenameClass and the new plugin...

superponible:
@ikelos this https://github.com/volatilityfoundation/volatility3/pull/1707/commits/f431f519b682d886edcda9b0371f483ee9735c97 has the class and decorator and the changes for just amcache

If I run windows.amcache it can't find the plugin

If I run registry.amcache (or just amcache since there's no name conflict), it prints

<pre>
Volatility 3 Framework 2.23.0
/opt/github/vol.git/volatility3/volatility3/framework/deprecation.py:106: FutureWarning: This plugin (volatility3.plugins.windows.amcache.Amcache) has been renamed and will be removed in the first release after 2025-09-25. Please ensure all method calls to this plugin are replaced with calls to volatility3.plugins.windows.registry.amcache.Amcache
  warnings.warn(
</pre>

ikelos:
Ok, looks like I was a little overzealous copying over all the methods from the replacement class (we copied some that were used to identify the class, so it was showing up as registry.amcache twice!).  Should be all fixed in the following patch:

```patch
diff --git i/volatility3/framework/deprecation.py w/volatility3/framework/deprecation.py
index ba07743a..b9b84f00 100644
--- i/volatility3/framework/deprecation.py
+++ w/volatility3/framework/deprecation.py
@@ -121,7 +121,7 @@ class PluginRenameClass:
         deprecated_class_name = f"{cls.__module__}.{cls.__qualname__}"
         super().__init_subclass__(**kwargs)
         for attr, value in replacement_class.__dict__.items():
-            if isinstance(value, classmethod):
+            if isinstance(value, classmethod) and attr != "get_requirements":
                 setattr(
                     cls,
                     attr,
@@ -134,5 +134,6 @@ class PluginRenameClass:
                     ),
                 )
             else:
-                setattr(cls, attr, value)
+                if not attr.startswith("__"):
+                    setattr(cls, attr, value)
         return super(replacement_class).__init_subclass__(**kwargs)
```

ikelos:
It will overwrite the version and required_framework_version, so we might want to make it just check for `startswith('_')` rather than `startswith('__')`  Classmethods should get copied regardless, so all the methods should still work...

superponible:
Thanks, the main changes worked and I got the message when running a windows.* plugin from the CLI and when running a registry.* plugin that imported the windows.hashdump instead of registry.hashdump.

But if I tried changing the check to "_", it doesn't work:

<pre>
    volatility3.cli.main()
  File "/opt/github/vol.git/volatility3/volatility3/cli/__init__.py", line 924, in main
    CommandLine().run()
  File "/opt/github/vol.git/volatility3/volatility3/cli/__init__.py", line 508, in run
    grid = constructed.run()
  File "/opt/github/vol.git/volatility3/volatility3/framework/plugins/windows/registry/cachedump.py", line 186, in run
    self._generator(syshive, sechive),
AttributeError: 'Cachedump' object has no attribute '_generator'
</pre>

ikelos:
Oh, yeah, ok, we'd need to special case the both version numebers then.  Probably not worthwhile since it's duplicating the new plugin so it should match the new plugin, but just a gotcha, we might take the old plugin away without realizing (in which case we should get rid of it completely on a MAJOR version bump).  Ok, so stick with the original version posted and hopefully that all works as intended.  Thanks for testing it out!  5:)

superponible:
Ok, I fixed that and updated the other plugins. should be good now.

ikelos (approved):
Looks good, thanks!

ikelos:
Any chance you can resolve the conflicts please?  Looks like there may have been some changes to the plugins since this branch was started, so need to make sure the changes were copied across properly...

superponible:
conflicts fixed

volatility3/framework/deprecation.py
```diff
@@ -88,3 +88,52 @@
...
+                            deprecated_class_name=deprecated_class_name,
+                            removal_date=removal_date,
+                            message=f"Please ensure all method calls to this plugin are replaced with calls to {replacement_class.__module__}.{replacement_class.__qualname__}",
+                        )(value.__func__)
+                    ),
+                )
+            else:
+                if not attr.startswith("__"):
+                    setattr(cls, attr, value)
+        return super(replacement_class).__init_subclass__(**kwargs)
```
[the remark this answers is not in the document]
ikelos: Hmmm, so I *think* we should be using `super(PluginRenameClass)` because the __init_subclass__ is essentially just a hook (a way to make common metaclass coding a bit easier) and I don't think it will alter anything unless someone inherits from this _deprecated_ class, so best to just let the machinery continue as is.  I was worried that if the replacement_class inherits from a different class (which we've tried hard to avoid doing), then it might not pull in the right inheritance/subclass, but I don't this method handles that bit, I think it's literally just a hook, so...  I'm gonna commit the suggestion...

ikelos (approved):
Looks good, thanks for getting it over the line!

PR #1708, opened 2025-03-13
Properly reconstruct strings from memory buffers and allow for plugin…

atcuno:
…-specified encodings

When working on the final GUI support, I ran into an issue with how Vol3 handles reading strings from memory and then reconstructing them. I then asked @dgmcdona for help as he has extensive experience with parsing UTF craziness in forensics context, and he said the existing function was just broken and incorrect.

In particular, the code assumes all strings are 1 char in length even though most Windows strings are utf8. Windows also have utf16, 32, etc. strings in use.

This code removes the hardcoded behaviour and allows Python's comprehensive decode() function to determine the characters. Our code then removes the junk off the end, including the unicode replacement character and obviously null.

This patch also expands the API so that callers can set the encoding to anything supported by Python. Finally, there is a minor framework bump since this expands the API.

### DIFF
--- a/volatility3/framework/objects/utility.py
+++ b/volatility3/framework/objects/utility.py
@@ -2,6 +2,8 @@
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
 
+import re
+
 from typing import Optional, Union
 
 from volatility3.framework import interfaces, objects, constants, exceptions
@@ -101,7 +103,9 @@ def pointer_to_string(
     )
 
 
-def gather_contiguous_bytes_from_address(layer, address: int, count: int) -> bytes:
+def gather_contiguous_bytes_from_address(
+    context, data_layer, starting_address: int, count: int
+) -> bytes:
     """
     This method reconstructs a string from memory while also carefully examining each page
 
@@ -115,48 +119,63 @@ def gather_contiguous_bytes_from_address(layer, address: int, count: int) -> byt
 
     data = b""
 
-    left_to_read = count
+    last_address = None
 
-    # read as many pages as possible that are contiguous
-    # if the first page is missed, we re-raise the InvalidAddressException
-    # if we have at least 1 page that was read succesfully,
-    # then we try to construct a string from it
-    while left_to_read > 0:
-        # compute aligned address of current page and next the page
-        aligned = address & ~0xFFF
-        next_page = aligned + 0xFFF + 1
+    for address, length, _, _, _ in data_layer.mapping(
+        offset=starting_address, length=count, ignore_errors=True
+    ):
+        # Used to track when we hit a paged out page
+        if not last_address:
+            last_address = address + length
 
-        # all fits on the current page, last read
-        if address + left_to_read < next_page:
-            try:
-                data += layer.read(address, left_to_read)
-            except exceptions.InvalidAddressException:
-                # if we have data, just break the loop
-                if data:
-                    break
-                # Raise if no data was read as this means the first page was invalid
-                else:
-                    raise
+        # we hit a swapped out page
+        elif last_address and last_address != address:
+            break
 
-            left_to_read = 0
+        data += data_layer.read(address, length)
 
-        else:
-            # how many bytes are left on the current page
-            len_to_read = next_page - address
-
-            try:
-                data += layer.read(address, len_to_read)
-            except exceptions.InvalidAddressException:
-                if data:
-                    break
-                # Raise if no data was read as this means the first page was invalid
-                else:
-                    raise
-
-            address += len_to_read
-            left_to_read -= len_to_read
+    # if we were able to read from the first page, we want to try and construct the string
+    # if the first page fails -> throw exception
+    if data:
+        return data
+    else:
+        raise exceptions.InvalidAddressException(
+            layer_name=data_layer, invalid_address=starting_address
+        )
 
     return data
+
+
+def bytes_to_decoded_string(data: bytes, encoding: str, errors: str) -> bytes:
+    """
+    This function takes a bytes buffer that contains at a string of unknown
+    length starting at the first byte, and returns the properly decoded string
+
+    It starts by using Python's `bytes.decode` to attempt to decode the entire string
+    It then finds the termination character (\ufffd or \x00) and splices the string
+    Finally, it returns this spliced string after its been decoded with the
+        caller-specified encoding
+    """
+    # this is the standard byte used to replace bad unicode characters
+    unicode_replacement_char = "\ufffd"
+
+    # used to find the terminating byte
+    termination_re = re.compile(f"{unicode_replacement_char}|\x00")
+
+    # run over the entire string, letting Python replace invalid characters
+    full_decoded_string = data.decode(encoding=encoding, errors="replace")
+
+    # stop at the first terminating character or get the whole string if not found
+    try:
+        idx = termination_re.search(full_decoded_string).start()
+    except AttributeError:
+        idx = len(full_decoded_string)
+
+    # cut at terminating byte, if found
+    data = data[:idx]
+
+    # return with caller-specified encoding and errors
+    return data.decode(encoding=encoding, errors=errors)
 
 
 def address_to_string(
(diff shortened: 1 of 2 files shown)


ikelos (commented):
Thanks, I like the idea and it is much cleaner, but the previous code was designed to cope with the following situation...

```
  stringyou'reinterestin\x00rubbish<end of page><next page is paged out>...
```
Which in your code (and probably the code before this monstrosity) would prevent the string being read, even though all of it was present in memory.
You can't read it padded (because you may get erroneous nulls in it).  I guess you could read the mapping and then just use the first contiguous values?  Or you could make the user try harder to recover whatever's readable (although they probably wouldn't try).  Also, if you're trying to decode the entire thing (without taking the terminator into account) what comes after it may be invalid, and the chosen `errors` option may throw an exception, even though the string is perfectly valid (but the rubbish after it is not).  This really needs to find the terminator first (and throw address exceptions if it can't find it), and then attempt the decode...

atcuno:
ok @ikelos check it now. I added `gather_contiguous_bytes_from_address` which, meets the best-of-scenario you described in that it will read page by page (dynamically computing the page boundaries), and, if the first page is missing, it re-raises the exception, but, if at least one page was there then it returns whats currently been read.

This means strings that start on a page and do not cross it will be safely read and recovered.

Towards the end I fix the missing bytes caused by trying to naively look for nulls to terminate, which is something that came up in testing.

volatility3/framework/objects/utility.py
```diff
@@ -102,6 +101,64 @@ def pointer_to_string(
...
+
+    left_to_read = count
+
+    # read as many pages as possible that are contiguous
+    # if the first page is missed, we re-raise the InvalidAddressException
+    # if we have at least 1 page that was read succesfully,
+    # then we try to construct a string from it
+    while left_to_read > 0:
+        # compute aligned address of current page and next the page
+        aligned = address & ~0xFFF
```
ikelos: This is hard coding pages size, I think the layers have [functions](https://github.com/volatilityfoundation/volatility3/blob/d82200f9f747c38149a3f55f3453be0f88b51679/volatility3/framework/layers/intel.py#L95) to do this so that we don't need to hardcode them?  (Think it only works on intel pages).

Why not just get the mapping back (ignoring errors so it'll map with gaps) and then you can read as much as is needed and then barf if the next chunk isn't contiguous with the previous one?  Saves even needing to know about page values at all, and more efficient if you're reading from a large page for some reason?

volatility3/framework/objects/utility.py
```diff
@@ -131,18 +188,38 @@ def address_to_string(
     if count < 1:
         raise ValueError("Count must be greater than 0")
 
+    encodings = {"utf8": 1, "utf16": 2, "utf32": 4}
+    if encoding not in encodings:
+        raise ValueError(
+            f"Encoding ({encoding} is invalid. Must be one of {[e for e in encodings]}."
```
ikelos: I thought encodings (as used by python) could contain an endianess too?

volatility3/framework/objects/utility.py
```diff
@@ -131,18 +188,38 @@ def address_to_string(
...
     layer = context.layers[layer_name]
 
-    # Purposely do not catch exception
-    data = layer.read(address, count)
+    data = gather_contiguous_bytes_from_address(layer, address, count)
+
+    # we need to find the ending nulls, which the amount of nulls varies based on encoding
+    ending_nulls = b"\x00" * encodings[encoding]
+
+    end_idx = data.find(ending_nulls)
```
ikelos: Given we now handle data breaking across pages, we could convert the data (with an error handling method) and then find the terminator and do the conversion again after cutting the string at the right length?  The difficulty will be cutting the original string at the right length based on cutting the encoded version of the string...

volatility3/framework/objects/utility.py
```diff
@@ -125,19 +188,45 @@ def address_to_string(
     if count < 1:
         raise ValueError("Count must be greater than 0")
 
+    encodings = {
```
ikelos: This is... a bit ugly.  Isn't there a means of determining character length from the encode/decode framework?  There's a suggestion further down that will avoid us needing to know (by attempting the conversion to find the correct breaking point and then doing the conversion again up to that point)...

ikelos (commented):
Thanks, much improved!  I think changing a couple of the techniques used will help it flow a little more cleanly though...

atcuno:
ok @ikelos check now

volatility3/framework/objects/utility.py
```diff
@@ -115,50 +119,65 @@ def gather_contiguous_bytes_from_address(layer, address: int, count: int) -> byt
...
-            address += len_to_read
-            left_to_read -= len_to_read
+    last_address = None
+
+    for address, length, _, _, _ in data_layer.mapping(
+        offset=starting_address, length=count, ignore_errors=True
+    ):
+        # Used to track when we hit a paged out page
+        if not last_address:
+            last_address = address + length
```
ikelos: After the first time round the loop, `last_address` will have been set, so this won't get updated and will stay the same for every iteration of the loop after.  I think last_address needs to be update after the data addition has been done...

volatility3/framework/objects/utility.py
```diff
@@ -94,16 +99,93 @@
...
+    # if we were able to read from the first page, we want to try and construct the string
+    # if the first page fails -> throw exception
+    if data:
+        return data
+    else:
+        raise exceptions.InvalidAddressException(
+            layer_name=data_layer, invalid_address=starting_address
+        )
+
+    return data
```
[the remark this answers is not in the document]
ikelos: Either ditch this line, or change the guard to an `if not data: raise ...`

volatility3/framework/objects/utility.py
```diff
@@ -94,16 +99,93 @@ def pointer_to_string(
...
+    termination_re = re.compile(f"{unicode_replacement_char}|\x00")
+
+    # run over the entire string, letting Python replace invalid characters
+    full_decoded_string = data.decode(encoding=encoding, errors="replace")
+
+    # stop at the first terminating character or get the whole string if not found
+    try:
+        idx = termination_re.search(full_decoded_string).start()
+    except AttributeError:
+        idx = len(full_decoded_string)
```
ikelos: If no terminator is found, then we return the entire length of the string, and if we hit a badly encoded character, it'll return everything that was well encoded up until then?  The docstring should definitely document that (ideally in an 
```
Args:
  data: ...
  encoding: ...
  errors: ...
  
Returns:
  ...
```
-style docstring...

We may want it to throw an error if something didn't decode correctly (or the user should have *some* way to know that things didn't go as expected) rather than them complaining of truncated strings and us scratching our heads...

ikelos (commented):
This looks *much* cleaner, thanks!  Still a couple of questions around behaviour and documentation, but something like this was always going to be tricky to get the semantics right...  5:)

ikelos:
And from the tests, it looks like you'll need to check whether the later is a `TranslationLayer` or not.  If not, you can just check that the data finishes before the end of the layer's maximum address and assume that everything else is contiguous...

volatility3/framework/objects/utility.py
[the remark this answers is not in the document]
atcuno: deleted

volatility3/framework/objects/utility.py
atcuno: Yea, its there now.. not sure how it got missed

ikelos: Still doesn't look like it gets incremented except if it didn't exist previously?

volatility3/framework/objects/utility.py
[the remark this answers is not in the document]
ikelos: Not sure the commit's been pushed up yet?  I'll wait a bit before checking again...

volatility3/framework/objects/utility.py
atcuno: I added a new parameter for this to give the caller control and documented it all.

volatility3/framework/objects/utility.py
```diff
@@ -94,16 +99,97 @@
...
+
+    if isinstance(data_layer, interfaces.layers.TranslationLayerInterface):
+        last_address = None
+
+        for address, length, _, _, _ in data_layer.mapping(
+            offset=starting_address, length=count, ignore_errors=True
+        ):
+            # Used to track when we hit a paged out page
+            if not last_address:
+                last_address = address + length
```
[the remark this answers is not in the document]
ikelos: The above is probably because of the elif below.  I  You can either dismiss it, or just make it an if to help the AI out...  5:P

volatility3/framework/objects/utility.py
```diff
@@ -94,16 +99,113 @@ def pointer_to_string(
...
+                break
+
+            data += data_layer.read(address, length)
+
+            last_address = address + length
+
+    elif starting_address + count < data_layer.maximum_address:
+        data = data_layer.read(starting_address, count)
+
+    # if we were able to read from the first page, we want to try and construct the string
```
ikelos, line 139: I think there's *technically* the case where the starting address, doesn't begin where the search is supposed to...

```
|   |<---asked for--->|  |
|paged out|actual data|   |
```

volatility3/framework/objects/utility.py
```diff
@@ -94,16 +99,113 @@ def pointer_to_string(
...
+
+    If a page is missing, the code initially catches the exception.
+    If data is non-empty (meaning at least one read succeeded), then we return what was read
+    If the first page fails, then we re-raise the exception
+    """
+
+    data = b""
+
+    if isinstance(data_layer, interfaces.layers.TranslationLayerInterface):
+        last_address = None
```
ikelos: This should probably be initialized to `start_address` to make sure everything gets checked properly.  Also obviates the need for the `if no last_address` clause...  5:)

ikelos (commented):
Looks good, just that one last corner case and then it should be good to go!  5:)

volatility3/framework/objects/utility.py
atcuno: fixed

volatility3/framework/objects/utility.py
atcuno: Hmm you will have to explain more. If we got a later page first (actual data in your diagram), then the `last_address` check would fail and we wouldn't read.

atcuno:
I just pushed the fixed last_adderss fix, and I think your concern on the edge case is off so we should be good there too.

volatility3/framework/objects/utility.py
atcuno: overhauled

volatility3/framework/objects/utility.py
ikelos: Yeah, but before we set `last_address` at the start of the loop based on the first entry (where `address` may not have been `starting_address`).  I think we're all good now.  5:)

ikelos (approved):
Yep, looks good now, thanks!

Issue #1699, opened 2025-03-10
Update `run_modules_scanners` to take in a configurable list of scanners


atcuno:
The API does a bunch of work now, but only for a hardcoded set of scanners. We want to make this configurable so plugins can send in one scanner or any combo of scanners to compute results.

https://github.com/volatilityfoundation/volatility3/pull/1673#discussion_r1986308281

atcuno:
This was merged but didn't auto close.

PR #1710, opened 2025-03-13
Update nearly all callers of now deprecated Linux kernel APIs

atcuno:
This updates all callers except netfilter and kallsyms as netfilters needed an API change and kallsyms is having its own huge update after this.

### DIFF
--- a/volatility3/framework/plugins/linux/modxview.py
+++ b/volatility3/framework/plugins/linux/modxview.py
@@ -37,6 +37,11 @@ spot modules presence and taints."""
                 version=(3, 0, 0),
             ),
             requirements.VersionRequirement(
+                name="linux_utilities_module_gatherers",
+                component=linux_utilities_modules.ModuleGatherers,
+                version=(1, 0, 0),
+            ),
+            requirements.VersionRequirement(
                 name="linux-tainting", component=tainting.Tainting, version=(1, 0, 0)
             ),
             requirements.BooleanRequirement(
@@ -89,44 +94,42 @@ spot modules presence and taints."""
         )
 
     def _generator(self):
-        kernel_name = self.config["kernel"]
+        kernel = self.context.modules[self.config["kernel"]]
 
-        kernel = self.context.modules[kernel_name]
-
-        wanted_sources = [
-            linux_utilities_modules.Modules.source_lsmod_identifier,
-            linux_utilities_modules.Modules.source_sysfs_identifier,
-            linux_utilities_modules.Modules.source_hidden_identifier,
+        wanted_gatherers = [
+            linux_utilities_modules.ModuleGathererLsmod,
+            linux_utilities_modules.ModuleGathererSysFs,
+            linux_utilities_modules.ModuleGathererScanner,
         ]
 
         run_results = linux_utilities_modules.Modules.run_modules_scanners(
             context=self.context,
             kernel_module_name=self.config["kernel"],
-            caller_wanted_sources=wanted_sources,
+            caller_wanted_gatherers=wanted_gatherers,
             flatten=False,
         )
 
         aggregated_modules = {}
         # We want to be explicit on the plugins results we are interested in
-        for plugin_name in wanted_sources:
+        for gatherer in wanted_gatherers:
             # Iterate over each recovered module
-            for mod_info in run_results[plugin_name]:
+            for mod_info in run_results[gatherer]:
                 # Use offsets as unique keys, whether a module
                 # appears in many plugin runs or not
                 if aggregated_modules.get(mod_info.offset, None) is not None:
                     # Append the plugin to the list of originating plugins
-                    aggregated_modules[mod_info.offset].append(plugin_name)
+                    aggregated_modules[mod_info.offset].append(gatherer)
                 else:
-                    aggregated_modules[mod_info.offset] = [plugin_name]
+                    aggregated_modules[mod_info.offset] = [gatherer]
 
-        for module_offset, originating_plugins in aggregated_modules.items():
-            # Tainting parsing capabilities applied to the module
+        for module_offset, gatherers in aggregated_modules.items():
             module = kernel.object("module", offset=module_offset, absolute=True)
 
+            # Tainting parsing capabilities applied to the module
             if self.config.get("plain_taints"):
                 taints = tainting.Tainting.get_taints_as_plain_string(
                     self.context,
-                    kernel_name,
+                    self.config["kernel"],
                     module.taints,
                     True,
                 )
@@ -134,7 +137,7 @@ spot modules presence and taints."""
                 taints = ",".join(
                     tainting.Tainting.get_taints_parsed(
                         self.context,
-                        kernel_name,
+                        self.config["kernel"],
                         module.taints,
                         True,
                     )
@@ -145,9 +148,9 @@ spot modules presence and taints."""
                 (
                     module.get_name() or NotAvailableValue(),
                     format_hints.Hex(module_offset),
-                    "lsmod" in originating_plugins,
-                    "check_modules" in originating_plugins,
-                    "hidden_modules" in originating_plugins,
+                    linux_utilities_modules.ModuleGathererLsmod in gatherers,
+                    linux_utilities_modules.ModuleGathererSysFs in gatherers,
+                    linux_utilities_modules.ModuleGathererScanner in gatherers,
                     taints or NotAvailableValue(),
                 ),
             )
@@ -158,7 +161,7 @@ spot modules presence and taints."""
             ("Address", format_hints.Hex),
             ("In procfs", bool),
             ("In sysfs", bool),
-            ("Hidden", bool),
+            ("In scan", bool),
             ("Taints", str),
         ]
 
--- a/volatility3/framework/plugins/linux/modxview.py
+++ b/volatility3/framework/plugins/linux/modxview.py
@@ -38,7 +38,17 @@ spot modules presence and taints."""
             ),
             requirements.VersionRequirement(
                 name="linux_utilities_module_gatherers",
-                component=linux_utilities_modules.ModuleGatherers,
+                component=linux_utilities_modules.ModuleGathererLsmod,
+                version=(1, 0, 0),
+            ),
+            requirements.VersionRequirement(
+                name="linux_utilities_module_gatherers",
+                component=linux_utilities_modules.ModuleGathererSysFs,
+                version=(1, 0, 0),
+            ),
+            requirements.VersionRequirement(
+                name="linux_utilities_module_gatherers",
+                component=linux_utilities_modules.ModuleGathererScanner,
                 version=(1, 0, 0),
             ),
             requirements.VersionRequirement(
@@ -113,14 +123,14 @@ spot modules presence and taints."""
         # We want to be explicit on the plugins results we are interested in
         for gatherer in wanted_gatherers:
             # Iterate over each recovered module
-            for mod_info in run_results[gatherer]:
+            for mod_info in run_results[gatherer.name]:
                 # Use offsets as unique keys, whether a module
                 # appears in many plugin runs or not
                 if aggregated_modules.get(mod_info.offset, None) is not None:
                     # Append the plugin to the list of originating plugins
-                    aggregated_modules[mod_info.offset].append(gatherer)
+                    aggregated_modules[mod_info.offset].append(gatherer.name)
                 else:
-                    aggregated_modules[mod_info.offset] = [gatherer]
+                    aggregated_modules[mod_info.offset] = [gatherer.name]
 
         for module_offset, gatherers in aggregated_modules.items():
             module = kernel.object("module", offset=module_offset, absolute=True)
@@ -148,9 +158,9 @@ spot modules presence and taints."""
                 (
                     module.get_name() or NotAvailableValue(),
                     format_hints.Hex(module_offset),
-                    linux_utilities_modules.ModuleGathererLsmod in gatherers,
-                    linux_utilities_modules.ModuleGathererSysFs in gatherers,
-                    linux_utilities_modules.ModuleGathererScanner in gatherers,
+                    linux_utilities_modules.ModuleGathererLsmod.name in gatherers,
+                    linux_utilities_modules.ModuleGathererSysFs.name in gatherers,
+                    linux_utilities_modules.ModuleGathererScanner.name in gatherers,
                     taints or NotAvailableValue(),
                 ),
             )
--- a/volatility3/framework/plugins/linux/modxview.py
+++ b/volatility3/framework/plugins/linux/modxview.py
@@ -34,7 +34,7 @@ spot modules presence and taints."""
             requirements.VersionRequirement(
                 name="linux_utilities_modules",
                 component=linux_utilities_modules.Modules,
-                version=(2, 0, 0),
+                version=(3, 0, 0),
             ),
             requirements.VersionRequirement(
                 name="linux-tainting", component=tainting.Tainting, version=(1, 0, 0)
@@ -93,13 +93,22 @@ spot modules presence and taints."""
 
         kernel = self.context.modules[kernel_name]
 
+        wanted_sources = [
+            linux_utilities_modules.Modules.source_lsmod_identifier,
+            linux_utilities_modules.Modules.source_sysfs_identifier,
+            linux_utilities_modules.Modules.source_hidden_identifier,
+        ]
+
         run_results = linux_utilities_modules.Modules.run_modules_scanners(
-            self.context, kernel_name, flatten=False
+            context=self.context,
+            kernel_module_name=self.config["kernel"],
+            caller_wanted_sources=wanted_sources,
+            flatten=False,
         )
 
         aggregated_modules = {}
         # We want to be explicit on the plugins results we are interested in
-        for plugin_name in ["lsmod", "check_modules", "hidden_modules"]:
+        for plugin_name in wanted_sources:
             # Iterate over each recovered module
             for mod_info in run_results[plugin_name]:
                 # Use offsets as unique keys, whether a module
--- a/volatility3/framework/plugins/linux/modxview.py
+++ b/volatility3/framework/plugins/linux/modxview.py
@@ -50,7 +50,7 @@ spot modules presence and taints."""
     @classmethod
     @deprecation.deprecated_method(
         replacement=linux_utilities_modules.Modules.flatten_run_modules_results,
-        replacement_version=(2, 0, 0),
+        replacement_version=(3, 0, 0),
         removal_date="2025-09-25",
     )
     def flatten_run_modules_results(
@@ -73,7 +73,7 @@ spot modules presence and taints."""
     @classmethod
     @deprecation.deprecated_method(
         replacement=linux_utilities_modules.Modules.run_modules_scanners,
-        replacement_version=(2, 0, 0),
+        replacement_version=(3, 0, 0),
         removal_date="2025-09-25",
     )
     def run_modules_scanners(
--- a/volatility3/framework/plugins/linux/modxview.py
+++ b/volatility3/framework/plugins/linux/modxview.py
@@ -37,17 +37,17 @@ spot modules presence and taints."""
                 version=(3, 0, 0),
             ),
             requirements.VersionRequirement(
-                name="linux_utilities_module_gatherers",
+                name="linux_utilities_module_gatherer_lsmod",
                 component=linux_utilities_modules.ModuleGathererLsmod,
                 version=(1, 0, 0),
             ),
             requirements.VersionRequirement(
-                name="linux_utilities_module_gatherers",
+                name="linux_utilities_module_gatherer_sysfs",
                 component=linux_utilities_modules.ModuleGathererSysFs,
                 version=(1, 0, 0),
             ),
             requirements.VersionRequirement(
-                name="linux_utilities_module_gatherers",
+                name="linux_utilities_module_gatherer_scanner",
                 component=linux_utilities_modules.ModuleGathererScanner,
                 version=(1, 0, 0),
             ),
--- a/volatility3/framework/symbols/linux/utilities/modules.py
+++ b/volatility3/framework/symbols/linux/utilities/modules.py
@@ -237,14 +237,9 @@ class Modules(interfaces.configuration.VersionableInterface):
         to not operate any inter-plugin results triage.
 
         Rules for `caller_wanted_sources`:
-
-            If `ModuleGathers.all_gathers_identifier` is specified then every source will be populated
-
-            If `ModuleGathers.Scanner` is in the list, then at least one other sources must be
-            specified so a comparison will be populated
+            If `ModuleGatherers.all_gathers_identifier` is specified then every source will be populated
 
             If empty or an invalid source is specified then a ValueError is thrown
-
         Args:
             called_wanted_sources: The list of sources to gather modules.
             flatten: Whether to de-duplicate modules across sources
--- a/volatility3/framework/symbols/linux/utilities/modules.py
+++ b/volatility3/framework/symbols/linux/utilities/modules.py
@@ -1,6 +1,18 @@
 import logging
 import warnings
-from typing import Iterable, Iterator, List, Optional, Tuple, NamedTuple, Dict, Set
+from typing import (
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Tuple,
+    NamedTuple,
+    Dict,
+    Set,
+    Generator,
+    Union,
+)
+from abc import ABCMeta, abstractmethod
 
 from volatility3 import framework
 from volatility3.framework import (
@@ -10,10 +22,42 @@ from volatility3.framework import (
     exceptions,
     objects,
 )
+
 from volatility3.framework.objects import utility
 from volatility3.framework.symbols.linux import extensions
 
 vollog = logging.getLogger(__name__)
+
+
+class ModuleInfo(NamedTuple):
+    """
+    Used to track the name and boundary of a kernel module
+    """
+
+    offset: int
+    name: str
+    start: int
+    end: int
+
+
+class ModuleGathererInterface(
+    interfaces.configuration.VersionableInterface, metaclass=ABCMeta
+):
+    _version = (1, 0, 0)
+    _required_framework_version = (2, 0, 0)
+
+    framework.require_interface_version(*_required_framework_version)
+
+    gatherer_return_type = Generator[Union[ModuleInfo, "extensions.module"], None, None]
+
+    @classmethod
+    @abstractmethod
+    def gather_modules(
+        cls, context: interfaces.context.ContextInterface, kernel_module_name: str
+    ) -> gatherer_return_type:
+        """
+        This method must return a generator (yield) of each `gatherer_return_type` found from its source
+        """
 
 
 class Modules(interfaces.configuration.VersionableInterface):
@@ -23,31 +67,6 @@ class Modules(interfaces.configuration.VersionableInterface):
     _required_framework_version = (2, 0, 0)
 
     framework.require_interface_version(*_required_framework_version)
-
-    # Valid sources of kernel modules to send to `run_module_scanners`
-    source_kernel_identifier = "kernel"
-    source_lsmod_identifier = "lsmod"
-    source_sysfs_identifier = "check_modules"
-    source_hidden_identifier = "hidden_modules"
-
-    # With few exceptions, rootkit checking plugins want all sources
-    # This provides a stable identifier as new sources are added over time
-    all_sources_identifier = [
-        source_kernel_identifier,
-        source_lsmod_identifier,
-        source_sysfs_identifier,
-        source_hidden_identifier,
-    ]
-
-    class ModuleInfo(NamedTuple):
-        """
-        Used to track the name and boundary of a kernel module
-        """
-
-        offset: int
-        name: str
-        start: int
-        end: int
 
     @classmethod
     def module_lookup_by_address(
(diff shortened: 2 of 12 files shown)


ikelos:
I feel that #1699 needs fixing before this lands.  No point having to go back through everything when we know that #1699 will be done...

atcuno:
Ha I felt the opposite so the callers can all be in place, then we just change one line per place for the ones using run_module_scanner. the ones using lsmod directly won't change. It will make tsting much cleaner if this goes in first.

ikelos:
But if you're already changing the lines for this, just get that in place and then change the line once?  That also ensures that #1699 can't get pushed down the road (which I'm very cautious of happening, and is why I want that sorted before we start using the stuff).  Really, it should've been part of the first PR, but it was getting a bit big, so I agreed to split it up, but as long as there was a bug number specifically so I could point at the bug and say "this before anything else"...  5;P

atcuno:
ok @ikelos I added the #1699 work in here. `run_modules_scanner` is fully configurable to the sources, including when using hidden modules, and its trivial to add new sources. Callers converted.

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -210,77 +225,173 @@ def get_kernel_module_info(
...
 
-        # check_modules
-        run_results["check_modules"] = []
+            gatherer = module_gatherers[wanted_source]
 
-        sysfs_modules: dict = cls.get_kset_modules(context, kernel_name)
+            # process each module coming from back the current source
+            for module in gatherer(context, kernel_module_name):
+                # the kernel sends back a ModuleInfo directly
+                if wanted_source == Modules.source_kernel_identifier:
```
ikelos: I realize this will be for efficiency, but it also prevents making the sources into an interface that others can add to.

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -210,77 +225,173 @@ def get_kernel_module_info(
...
+
         run_results = {}
 
-        # the kernel module boundaries
-        run_results["kernel"] = [cls.get_kernel_module_info(context, kernel_name)]
+        run_hidden_modules = False
 
-        # lsmod
-        run_results["lsmod"] = []
+        # Special case hidden modules since it gathers modules on its own
```
ikelos: Special casing this makes it somewhat difficult for other gatherers to be written that could work the same...  5:S  Would it be better to make an *all* modules gatherer (rather than a hidden modules gatherer) and run a process over the result to identify hidden ones (basically, do the filtering after)?  That would then allow all gatherers to operate in the same way, and the results would be more consistent and all the special casing goes away (apart from a step at the end to filter one type from the other)?

ikelos (commented):
Thanks for adding #1699 to this.  The current proposal does go towards making the module gatherers more generic, but it somewhat misses the point.  They're still limited to the modules class (meaning adding new gatherers is nigh on impossible for another plugin) and several bits are special cased because of it.  The point of doing it this way was to allow others to write new different/better means of gathering modules, given that we've come up with 3 different ways ourselves already.  As such, I was hoping for a fixed function signature/interface that provided all the necessary components for something to find modules and then allow the consumer to process it (for example, seeing that some modules are hidden and others are not) however they wanted.  That would allow plugins to send in _their own_ scanners, or any combination.  Sorry if that wasn't clear from the description of #1699...

atcuno:
Ok so it seems like there are two pieces you want, one of which I would say is there already, the other not.

Request 1:

> ... then allow the consumer to process it (for example, seeing that some modules are hidden and others are not) however they wanted. 
> ...
> Would it be better to make an all modules gatherer (rather than a hidden modules gatherer) and run a process over the result to identify hidden ones (basically, do the filtering after)? 

The current code supports these already. The caller sends in the sources they want and then can parse them on their own. If they set `flatten=False`, then they get back the dictionary key by each source. Open the full link here in a seperate tab to see how `modxview` does exactly this to then display the sources per column. Its exact dynamic setup we wanted:

https://github.com/volatilityfoundation/volatility3/blob/4606495ded4f71c2bdd9066b260556947729e618/volatility3/framework/plugins/linux/modxview.py#L102

Request 2: 

> As such, I was hoping for a fixed function signature/interface that provided all the necessary components for something to find modules 

I definitely misunderstood that this was needed before (never even heard of it), but it is a really cool idea so I can add it in today.

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -10,44 +22,51 @@
...
+
 from volatility3.framework.objects import utility
 from volatility3.framework.symbols.linux import extensions
 
 vollog = logging.getLogger(__name__)
 
 
-class Modules(interfaces.configuration.VersionableInterface):
-    """Kernel modules related utilities."""
+class ModuleInfo(NamedTuple):
```
ikelos, line 33: I think this lived inside of the `Modules` class, so that it was part of the versioning of the Modules code.  If it's defined in the outside it's not tied together and has to work for all time as is...

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -204,138 +223,41 @@ def get_module_info_for_module(
...
-                f"{Modules.source_hidden_identifier} cannot be the only source or there is nothing to compare against."
+                "`caller_wanted_gatherers` must have at least one gatherer."
             )
 
-        wanted_sources = []
-
-        for source in caller_wanted_sources:
-            if source not in Modules.all_sources_identifier:
+        for gatherer in caller_wanted_gatherers:
+            if gatherer not in ModuleGatherers.all_gatherers_identifier:
```
ikelos: This prevents people from making their own gatherers, doesn't it?  Either that, or it's inviting people to monkey patch the core gatherer list?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -204,138 +223,41 @@ def get_module_info_for_module(
...
-            if modinfo:
-                hidden_results.append(modinfo)
-
-        return hidden_results
-
-    @classmethod
-    def _get_list_modules(
-        cls, context: interfaces.context.ContextInterface, kernel_module_name: str
-    ) -> List[ModuleInfo]:
+    def _validate_gatherers(cls, caller_wanted_gatherers) -> List[str]:
```
ikelos: Typing information on caller_wanted_gatherers please.  It must at least be `list[callable]` and then you can supply the function signature in the callable parameter if you want...

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -204,138 +223,41 @@ def get_module_info_for_module(
...
         """Run module scanning plugins and aggregate the results. It is designed
         to not operate any inter-plugin results triage.
 
         Rules for `caller_wanted_sources`:
 
-            If `Modules.all_sources_identifier` is specified then every source will be populated
+            If `ModuleGathers.all_gathers_identifier` is specified then every source will be populated
 
-            If `Modules.source_hidden_identifier` is in the list, then at least one other sources must be
+            If `ModuleGathers.Scanner` is in the list, then at least one other sources must be
```
ikelos: Getting bad vibes from this being conditional, I'll have to see where it's going...

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -346,52 +268,30 @@ def run_modules_scanners(
...
             Dictionary mapping each plugin to its corresponding result
         """
-
-        module_gatherers = {
-            Modules.source_kernel_identifier: cls.get_kernel_module_info,
-            Modules.source_lsmod_identifier: cls._get_list_modules,
-            Modules.source_sysfs_identifier: cls._get_sysfs_modules,
-        }
+        # Throws ValueError if invalid gatherers sent in
+        Modules._validate_gatherers(caller_wanted_gatherers)
```
ikelos: Don't be mean, let people define their own.  5:P  The API is for everyone, but then it's on them if it breaks because of what they handed it...

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -346,52 +268,30 @@ def run_modules_scanners(
...
-
-        # Special case hidden modules since it gathers modules on its own
-        if Modules.source_hidden_identifier in wanted_sources:
-            run_hidden_modules = True
-            wanted_sources.remove(Modules.source_hidden_identifier)
-
-        # Walk each source, gathering modules
-        for wanted_source in wanted_sources:
-            run_results[wanted_source] = []
+        run_results: Dict[ModuleGathererInterface, List[ModuleInfo]] = {}
```
ikelos, line 287: Might be useful for a `ModuleGathererInterface` to have a name, so that the results can be (uniquely) referred to by something other than a class...

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -346,52 +268,30 @@ def run_modules_scanners(
...
+        for gatherer in caller_wanted_gatherers:
+            run_results[gatherer] = []
 
             # process each module coming from back the current source
-            for module in gatherer(context, kernel_module_name):
+            for module in gatherer.gather_modules(context, kernel_module_name):
+
                 # the kernel sends back a ModuleInfo directly
-                if wanted_source == Modules.source_kernel_identifier:
+                if gatherer == ModuleGathererKernel:
```
ikelos: Again, we were trying to avoid the special casing?  Why do the `ModuleGathererKernel` results need handling differently?  If we're defining this method, can't we just get it to play ball like the rest of the gatherers?

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: I don't think this comment is accurate?  I can't see anything in the Scanner that requires anything else?  It looks like it always scans all modules, just as expected?

ikelos (commented):
Much happier with the layout and it feels like it should be a lot smoother and easier to work with.  There's just a couple idiosyncrasies that need looking at, but on the whole I think it's pretty much there...  5:)

atcuno:
@ikelos should be good now. much cleaner and nicer to use!

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -534,3 +539,98 @@ def validate_alignment_patterns(
...
+        start_addr = kernel.object_from_symbol("_text")
+        start_addr = start_addr.vol.offset & address_mask
+
+        end_addr = kernel.object_from_symbol("_etext")
+        end_addr = end_addr.vol.offset & address_mask
+
+        yield ModuleInfo(start_addr, constants.linux.KERNEL_NAME, start_addr, end_addr)
+
+
+class ModuleGatherers(interfaces.configuration.VersionableInterface):
```
ikelos: I'm... not sure this is useful?  I'd just keep the list versioned inside the `Modules` component?  I might also make each of the scanners live inside it, so that they inherit the versioning too?  Otherwise we're just versioning the list of all of them, but not each of the individual gatherers...

ikelos (commented):
Took another look and Just added a couple extra bits...

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: Yes I am aware. If that every changes or breaks I will have to start from scratch anyway. It is used between both Modules and ModuleGatherers, so these versions would bump too.

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: Refresh, that was my bad at first, its in the latest version to just subclass() check on the interface

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: This whole function went away just now.

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: This is gone now, only for a real check on iterable and subclass

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: hmm then we end up duplicating all the identifers.

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: This was the main reason for the entire glueing together of my old API and what was missed in the later kernel API to support tracing. The kernel doesn't have a `module` struct, so we make a modinfo directly. Its the only special needed, but every rootkit plugin will break without it (we discussed this before when the APIs were merged also).

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: But then we would need to version bump Modules on changes, which only plugins that look for rootkit signs care about that, but we would have to version bump ones that don't care about rootkits also. This breaks the versioning out so only plugins wanting the functionality have to care.

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: Nice catch, that was a stale comment. about to push a commit with it gone.

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: I think `ModuleGatherers` should live under `Modules` too?  Then we'd get them both versioned for free?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -10,37 +22,59 @@
...
+    interfaces.configuration.VersionableInterface, metaclass=ABCMeta
+):
+    _version = (1, 0, 0)
     _required_framework_version = (2, 0, 0)
 
     framework.require_interface_version(*_required_framework_version)
 
-    class ModuleInfo(NamedTuple):
+    gatherer_return_type = Generator[Union[ModuleInfo, "extensions.module"], None, None]
+
```
ikelos, line 53: Fancy adding a `name` field?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -189,98 +223,64 @@ def get_module_info_for_module(
...
+            )
 
-        kernel = context.modules[kernel_name]
+        if not isinstance(caller_wanted_gatherers, Iterable):
+            raise ValueError("`caller_wanted_gatherers` must be iterable")
 
-        address_mask = context.layers[kernel.layer_name].address_mask
+        for gatherer in caller_wanted_gatherers:
+            if not issubclass(gatherer, ModuleGathererInterface):
+                raise ValueError(
```
ikelos, line 267: Much better, thanks  5:)

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -189,98 +223,64 @@ def get_module_info_for_module(
...
+                    f"Invalid gatherer sent through `caller_wanted_gatherers`: {gatherer}"
+                )
 
-        run_results = {}
+        kernel = context.modules[kernel_module_name]
 
-        # the kernel module boundaries
-        run_results["kernel"] = [cls.get_kernel_module_info(context, kernel_name)]
+        address_mask = context.layers[kernel.layer_name].address_mask
```
ikelos, line 285: I don't think this is needed here anymore?

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Are there identifiers anywhere?  At the moment it's would just be `__class__.__name__` to figure out something you could print to show where it came from?  Seems handy to be able to say `"hidden_modules"`  rather than `"ModulesGathererHidden"` for example?

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Ok, so... would it not be better to have the interface require that a `modinfo` is returned, rather than a module-for-all-but-one-special-case?  Making specific unique things inside the generic handler breaks (because someone else might come up with a situation where they need to special case stuff, and they'll run into the same issue).  Pushing the class call into each of the yields seems a better option...

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: This would also get weird as the plugins would have to know the identiifer plus the class name unless we sent back some mixed data structures. Look at modxview, its very easy to distinguish these.

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Yeah, but just like very thing, we'd work hard to only make additive bumps (which then won't require the plugins to change).  This way, a change here isn't tracked, only a change in the default framework collection of trackers is versioned, and people will need to check the version of both the `ModuleGatherers` list *and* of the `Modules` class, when they're kinda bound together?

atcuno: Also, I really don't like everything shoved in one class. Having versioned classes as small as makes sense (grouped functionality) means plugins can make specific requirements on just what they need and we avoid the endless version bumps for plugins of unrelated work.

ikelos (commented):
The ModuleGatherer having separate versioning just feels kinda... excessive?  They're only really going to get used together in general?

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: True, but there comes a point where you're needing everyone to version everything.  At the very least the gatherers themselves need to be under *something* versioned, otherwise technically they're linked to the framework version...  5:S

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: No, no, you can index by the class, but just have the class tagged with a string identifier.  You never even have to access it if you don't want, but given you don't know exactly which gatherers may be included, you might want to let plugins get a human readable string, rather than just handing them back a class...

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: Do they not inherit the version from the interface? I was pretty certain it worked that way, so I designed it where, if the interface ever version bumps, then the gatherers have to adjust.

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: No because then everything has to be duplicated. You are only getting a return type after you specify the source you want, so a name wouldn't add anything or fit in. Look at modxview how it can just check the class identifier.

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: Then every gatherer would have do the `get_module_info_for_module` loop instead of just the generic gathering knowing the kernel doesn't have a real module.. it would make really ugly code just to avoid a check on the kernel.

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: I want plugins that do the rootkit checks to version both. Ones that don't can just do Modules and not get version bumped or changed when rootkit related stuff changes. The "everything in one module" makes versioning very painful.

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: It is used in a call a few lines later.

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Ok, but... the gatherers should be versioned under something less than the framework (at the moment they're just floating free in this module.  They could each be individually versioned, but then the "scan everything list" would need bumping every time one of them got bumped?  Happy to go that route if you think it's the right one?

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: I am resolving here as its being discussed elsewhere also and the split is confusing

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: Ok do you mean tagged like adding a property in the constructor, then callers could do `.name`? If that I don't mind that as much. Although every gatherer would have to override __init__

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: I don't think that's what I was suggesting?  Just that instead of `return thing`, they each now do `return cls.get_module_info_for_module(address_mask, thing)`?  I guess it might require feeding the `address_mask` in though...

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: Ah yea, that level of granularity is nice. I will work that in.

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Ok, sorry, I missed that.  It does mean though that the ModuleGatherers list needs to require each of the sub items (so that their versions all get verified when the ModuleGatherers list is used...

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -534,3 +534,98 @@ def validate_alignment_patterns(
...
+class ModuleGathererLsmod(ModuleGathererInterface):
+    """
+    Gathers modules from the main kernel list
+    """
+
+    @classmethod
+    def gather_modules(
+        cls, context: interfaces.context.ContextInterface, kernel_module_name: str
+    ) -> ModuleGathererInterface.gatherer_return_type:
+        yield from Modules.list_modules(context, kernel_module_name)
```
ikelos, line 572: So this one would require re-yielding after converting each module to a modinfo

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: They all are generators, so each would have to make that call to `cls.get_module_info_for_module` for each element. Check for example how lsmod looks now here, it would need to loop the calls (which would be copy/paste in every gatherer):

https://github.com/volatilityfoundation/volatility3/blob/a0f3cba6f6d71648a4feb884502491b773815255/volatility3/framework/symbols/linux/utilities/modules.py#L547

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -534,3 +534,98 @@ def validate_alignment_patterns(
...
+    @classmethod
+    def gather_modules(
+        cls, context: interfaces.context.ContextInterface, kernel_module_name: str
+    ) -> ModuleGathererInterface.gatherer_return_type:
+        kernel = context.modules[kernel_module_name]
+
+        sysfs_modules: dict = Modules.get_kset_modules(context, kernel_module_name)
+
+        for m_offset in sysfs_modules.values():
+            yield kernel.object(object_type="module", offset=m_offset, absolute=True)
```
ikelos, line 593: But otherwise it would just be one extra call here, to hand back a modinfo rather than a module...

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -10,37 +22,59 @@
...
+class ModuleGathererInterface(
+    interfaces.configuration.VersionableInterface, metaclass=ABCMeta
+):
+    _version = (1, 0, 0)
     _required_framework_version = (2, 0, 0)
 
     framework.require_interface_version(*_required_framework_version)
 
-    class ModuleInfo(NamedTuple):
+    gatherer_return_type = Generator[Union[ModuleInfo, "extensions.module"], None, None]
```
ikelos, line 52: So this shouldn't be a union...

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: yes, that makes sense, that will now be a part of the update where each gatherer gets versioned.

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: The other option is *not* to special case this, by the module, but based on the type returned in the yield?  That way you handle a gatherer that decides to just return a modinfo (as allowed by the interface) and nothing's been special cased...

ikelos: So an `isinstance(thing, Module) then`...

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: It can be now with isinstance right? Was this comment from before that?

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -89,43 +94,50 @@ def run_modules_scanners(
         )
 
     def _generator(self):
-        kernel_name = self.config["kernel"]
+        kernel = self.context.modules[self.config["kernel"]]
 
-        kernel = self.context.modules[kernel_name]
+        wanted_gatherers = [
+            linux_utilities_modules.ModuleGathererLsmod,
```
ikelos, line 110: So... I can't tell if these should be versioned individually (since they *should* be part of the main list, but...)

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Yeah, it was from when I was trying to get rid of the kernel special casing.  WIth an isinstance check instead, this can stay as it is.

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: No, just a class attribute, one per scanner only, not per instance.  I think this has been sorted, so marking as resolved.

ikelos (commented):
All looks good, nearly there.  Just one check about one that uses a specific list of gatherers, rather than the total list.  It should technically version each one it individually uses.

Also, still pondering over using the class as the dictionary key.  It's probably fine, I assume this has been tested, so it's hashable, it just means carrying something hefty around in the results dictionary and I much prefer such things to be lighter weight.  Do we ever actually use (or envisage using) the gatherer from the dictionary key, or could we just use the name field we added instead (checking that it's unique across all the gatherers used)?

volatility3/framework/plugins/linux/modxview.py
atcuno: That is a very good catch and would set a very bad precedent if we didn't do this. I will update that shortly.

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -34,7 +34,22 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
             requirements.VersionRequirement(
                 name="linux_utilities_modules",
                 component=linux_utilities_modules.Modules,
-                version=(2, 0, 0),
+                version=(3, 0, 0),
+            ),
+            requirements.VersionRequirement(
+                name="linux_utilities_module_gatherers",
```
ikelos:
```suggestion
                name="linux_utilities_module_gatherer_lsmod",
```

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -34,7 +34,22 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+                component=linux_utilities_modules.ModuleGathererLsmod,
+                version=(1, 0, 0),
+            ),
+            requirements.VersionRequirement(
+                name="linux_utilities_module_gatherers",
```
ikelos:
```suggestion
                name="linux_utilities_module_gatherer_sysfs",
```

volatility3/framework/plugins/linux/modxview.py
```diff
@@ -34,7 +34,22 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+                component=linux_utilities_modules.ModuleGathererSysFs,
+                version=(1, 0, 0),
+            ),
+            requirements.VersionRequirement(
+                name="linux_utilities_module_gatherers",
```
ikelos:
```suggestion
                name="linux_utilities_module_gatherer_scanner",
```

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -189,98 +224,80 @@ def get_module_info_for_module(
...
-        address_mask = context.layers[kernel.layer_name].address_mask
+        seen_names = set()
+
+        for gatherer in caller_wanted_gatherers:
+            if not issubclass(gatherer, ModuleGathererInterface):
+                raise ValueError(
+                    f"Invalid gatherer sent through `caller_wanted_gatherers`: {gatherer}"
+                )
+
+            if not hasattr(gatherer, "name"):
```
ikelos: Fine to do this if you really want, but I think it's part of the interface, so we've already checked that it has a name attribute above?

ikelos (commented):
Looks really good, three minor changes (config names should be unique) and then we should be all good.  5:D

volatility3/framework/symbols/linux/utilities/modules.py
atcuno: hmm that is a good point, but now I am looking and `ModuleGathererInterface` doesn't have a `name` either.. I guess we need to set a `name=None` or what is the fix?

ikelos (approved):

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Yeah, set name to `None` (and then I guess we do need to check it's set?)

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -189,98 +227,80 @@ def get_module_info_for_module(
...
+        seen_names = set()
 
-        run_results = {}
+        for gatherer in caller_wanted_gatherers:
+            if not issubclass(gatherer, ModuleGathererInterface):
+                raise ValueError(
+                    f"Invalid gatherer sent through `caller_wanted_gatherers`: {gatherer}"
+                )
+
+            if gatherer.name is None or len(gatherer.name) == 0:
```
ikelos:
```suggestion
            if not gatherer.name:
```
Think this achieves the same goal

ikelos (commented):
Sorry for the nitpick!  5:P

PR #1713, opened 2025-03-15
Windows: Refactor JSON data handling

dgmcdona:
This changes the way that JSON resources required for some plugins (excluding symbol JSON files) are handled within the volatility3 package. Instead of operating directly on paths, it uses Python's builtin `importlib.resources` module to identify a data directory with JSON files as a resource location. This greatly simplifies how resources are found within these three plugins, and conforms to best practices for inclusion of non-Python resources within a package.

The motivation for this was that I found the storage of the SIDs data in the same directory as the plugin files (and the code used to access it) to be suspect. We can consider this as a standalone issue for this particular file, or I'm happy to look into making similar changes to the way that we handle JSON files for intermediate symbol tables as well, although that would require more extensive changes than seen here and probably a major framework version bump, since I would anticipate breaking changes to the `IntermediateSymbolTable` classmethods.

References:

https://docs.python.org/3/library/importlib.resources.html#module-importlib.resources


dgmcdona:
These test failures are strange, I'll have to look into them some more...

I created a clean virtual environment and installed this branch with `pip install git+ssh://git@github.com/volatilityfoundation/volatility3.git@dgmcdona/package_data` and was able to run these plugins with no issues.

dgmcdona:
The issue with the tests ended up being differences between Python3.8 and Python3.10, and the way that `importlib.resources` handles namespace packages - it seems to be designed to work with traditional packages, so I had to add the `__init__.py` file to the data directory. I'm not 100% sure that `volatility3/data/` is where it should live, but we can't keep it alongside the plugin files anymore because of the exception raised when importing from the `volatility3.framework.plugins.windows` package. One option is to move it to `volatility3/plugins/windows/`.

ikelos:
Hiya, I'm ok with the concept and I generally like things being tidied up, but I worry that it doesn't consider external plugin developers that have data files?  Part of the reason for doing things that way was to provide people an example for including data with their plugins (like the sids plugin does), since those sets of files will work as part of the framework or not.

I realize we've mostly been taking a batteries-included approach and wrapped most every plugin we've got into core, but the framework was designed for the community to develop their own, and make them transportable.  That's why plugins can be wrapped in a zip file (although I don't know if many people realize the capability exists) and it's why we go to great lengths to merge `volatility3.framework.plugins` namespace under the `volatility3.plugins` namespace, but allows `volatility3.plugins` to take precedence.  It's all designed to allow users to tinker and replace bits of the framework as they want.  Requiring people to drop their data into a specific folder (and not separating core from non-core) might make it easier for core, but I think it'll just make it even less likely for anyone to ever make use of the capabilities we've worked to provide.

It would also seem prudent to split the data out into at least os directories, and I worry about the removal of the error handling, and logging.  What happens if the resource is missing for some reason with the new code?

I agree that having the same code in three plugins seems excessive.  We could definitely move that into a class method of one of them, with an additive bump and then keep that loading functionality all in one place.  Assuming they already import from each other, that seems like it would be a sensible move...

dgmcdona:
Thanks for giving me some background into this, I didn't realize that this way of doing things was meant to be instructive for plugin developers, but in that context things make way more sense as they are now. I experimented a bit with passing additional plugin directories as both a directory and a zip (which I didn't know you could do before), but there's no way `importlib.resources` will work in that context (at least for plugin developers trying to access their own resources) due to the `__path__` modifications, so I agree that this should remain as an example of how to access other resources within plugin the plugin directories for now.

It would probably be worthwhile to clean things up a bit by providing classmethods in `getsids` or another location that allow users to lookup SIDs on-demand without having to load the JSON file directly and inspect its internal structure, so I'll put something like that together in place of these changes.

Regarding the error handling, my assumption was that when using `importlib.resources`, it wouldn't be necessary since it would always be present in the correct location via either cloning or pip installation. I'm not sure it makes sense to protect against arbitrary deletion of files that are core to the framework.

ikelos:
Not for files that are core, but for an external plugin someone downloaded that depends on a data file, it should handle it not being there and be able to tell the user.  As you say, `importlib.resources` wouldn't work in those instances anyway, so probably best just to leave it as is, but the sid lookup centralizing in a class method seems fine.  No reason for each plugin that needs it to load it up in its own way...

dgmcdona:
Closing this for now, will add the SIDs lookup APIs in another PR later.

PR #1715, opened 2025-03-15
Fix broken truncation from Vol3 bytes to string conversion

atcuno:
@ikelos when running mass testing today on the new GUI plugin, I noticed that some of the strings it was printing were truncated or had a junk byte at the end.

I debugged it and the recent addition on how we handled encode/decode was the culprit as I was indexing back into the original string due to `decode` returning a `str`. This worked for most strings, but depending on the encoding and trailing junk bytes it broke it.

This update makes the previously problematic strings (from the first overhaul PR) still return the valid string without junk at the end while also getting the newly discovered ones.

We need to cast to bytes() here as the `decode` to transform the whole string returns a `str` as seen here:

```
>>> type(b"abc".decode(encoding="utf8"))
<class 'str'>
```

### DIFF
--- a/volatility3/framework/objects/utility.py
+++ b/volatility3/framework/objects/utility.py
@@ -188,7 +188,7 @@ def bytes_to_decoded_string(
             )
 
     # cut at terminating byte, if found
-    data = data[:idx]
+    data = bytes(full_decoded_string[:idx], encoding=encoding)
 
     # return with caller-specified encoding and errors
     return data.decode(encoding=encoding, errors=errors)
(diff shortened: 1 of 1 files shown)


ikelos (approved):
Yep, good catch, thanks

volatility3/framework/objects/utility.py
```diff
@@ -105,7 +105,7 @@ def pointer_to_string(
 
 def gather_contiguous_bytes_from_address(
```
ikelos: I think that's the wrong method to change the signature on?  I think you meant to change `bytes_to_decoded_string` which says it returns bytes...

ikelos (commented):
Think you changed the wrong signature...

ikelos (approved):

PR #1716, opened 2025-03-15
Add the windows plugin and associated extensions updates

atcuno:
This adds the `windows.windows.Windows` plugin to enumerate windows of each desktop of each window station.

This includes adding the `tagWND` extensions needed as well as the extension for a desktop's windows to be enumerated.

We also had to adjust the base types of the GUI files since the hand-added types didn't include the normal `char` but the structures referenced it.

### DIFF
--- a/volatility3/framework/plugins/windows/windows.py
+++ b/volatility3/framework/plugins/windows/windows.py
@@ -102,11 +102,13 @@ class Windows(interfaces.plugins.PluginInterface):
 
             # procedures can be empty, but if set, should be a valid pointer
             window_proc = window.get_window_procedure()
-            if window_proc is None or window_proc == 0 or window_proc > 0x1000:
+            if window_proc is None:
+                window_proc = renderers.NotAvailableValue()
+            elif window_proc == 0 or window_proc > 0x1000:
                 window_proc = format_hints.Hex(window_proc)
             else:
                 vollog.warning(
-                    f"Invalid window procedure for the window {window.vol.offset:#x}"
+                    f"Invalid window procedure {window_proc} for the window {window.vol.offset:#x}"
                 )
                 continue
 
--- a/volatility3/framework/plugins/windows/windows.py
+++ b/volatility3/framework/plugins/windows/windows.py
@@ -9,7 +9,6 @@ from volatility3.framework.objects import utility
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints
 from volatility3.plugins.windows import windowstations
-from volatility3.framework.symbols.windows.extensions import gui
 
 vollog = logging.getLogger(__name__)
 
@@ -33,9 +32,6 @@ class Windows(interfaces.plugins.PluginInterface):
                 name="windowstations",
                 component=windowstations.WindowStations,
                 version=(1, 0, 0),
-            ),
-            requirements.VersionRequirement(
-                name="GUIExtensions", component=gui.GUIExtensions, version=(1, 0, 0)
             ),
         ]
 
--- a/volatility3/framework/plugins/windows/windows.py
+++ b/volatility3/framework/plugins/windows/windows.py
@@ -9,6 +9,7 @@ from volatility3.framework.objects import utility
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints
 from volatility3.plugins.windows import windowstations
+from volatility3.framework.symbols.windows.extensions import gui
 
 vollog = logging.getLogger(__name__)
 
@@ -32,6 +33,9 @@ class Windows(interfaces.plugins.PluginInterface):
                 name="windowstations",
                 component=windowstations.WindowStations,
                 version=(1, 0, 0),
+            ),
+            requirements.VersionRequirement(
+                name="GUIExtensions", component=gui.GUIExtensions, version=(1, 0, 0)
             ),
         ]
 
@@ -88,7 +92,7 @@ class Windows(interfaces.plugins.PluginInterface):
                     )
 
             if process_name is None:
-                vollog.warning(
+                vollog.debug(
                     f"Invalid process reference for the process hosting window {window.vol.offset:#x}"
                 )
                 continue
@@ -107,7 +111,7 @@ class Windows(interfaces.plugins.PluginInterface):
             elif window_proc == 0 or window_proc > 0x1000:
                 window_proc = format_hints.Hex(window_proc)
             else:
-                vollog.warning(
+                vollog.debug(
                     f"Invalid window procedure {window_proc} for the window {window.vol.offset:#x}"
                 )
                 continue
--- a/volatility3/framework/plugins/windows/windows.py
+++ b/volatility3/framework/plugins/windows/windows.py
@@ -1,0 +1,137 @@
+# This file is Copyright 2025 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import logging
+from typing import List, Iterable
+
+from volatility3.framework import interfaces, renderers, exceptions
+from volatility3.framework.objects import utility
+from volatility3.framework.configuration import requirements
+from volatility3.framework.renderers import format_hints
+from volatility3.plugins.windows import windowstations
+
+vollog = logging.getLogger(__name__)
+
+
+class Windows(interfaces.plugins.PluginInterface):
+    """Enumerates the Windows of Desktop instances"""
+
+    _required_framework_version = (2, 0, 0)
+    _version = (1, 0, 0)
+
+    @classmethod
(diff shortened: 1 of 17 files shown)


ikelos:
Wouldn't this be better as something like windows.gui.Windows?

volatility3/framework/plugins/windows/windows.py
```diff
@@ -0,0 +1,137 @@
...
+            sess_id = window.get_session_id()
+            if sess_id is None:
+                vollog.debug(
+                    f"Unable to read session id of the process for window {window.vol.offset:#x} in process {process_name}"
+                )
+                continue
+
+            # procedures can be empty, but if set, should be a valid pointer
+            window_proc = window.get_window_procedure()
+            if window_proc is None or window_proc == 0 or window_proc > 0x1000:
```
ikelos: Think this logic is the wrong way around?  Currently says if window_proc is None windows_proc = format_hints.Hex(None)...  5:S

volatility3/framework/symbols/windows/extensions/gui.py
```diff
@@ -77,7 +81,7 @@ def desktops(self, symbol_table_name, max_desktops: int = 12):
 class tagDESKTOP(objects.StructType, pool.ExecutiveObject):
     def is_valid(self) -> bool:
         """
-        Enforce a valid sid + owning window station
+        Enforce a valid sid + name
```
ikelos: Does this validate the name?  It looks like the old comment was more accurate (although either way they're both quite brief)...

volatility3/framework/symbols/windows/extensions/gui.py
```diff
@@ -120,8 +130,193 @@ def get_threads(
...
+    def _do_get_windows(
+        self, window, max_windows
+    ) -> Generator[Tuple[interfaces.objects.ObjectInterface, str], None, None]:
+        """
+        Recusively walks and yields the adjacent and child windows
+        """
+        seen_windows = set()
+        seen_children = set()
+
+        if window.vol.offset == 0:
```
ikelos: More pythonic to do `if not window.vol.offset` but this is technically fine?  Just that `== 0` tends to trip my "not pythonic" senses  5;P

volatility3/framework/symbols/windows/extensions/gui.py
```diff
@@ -120,8 +130,193 @@ def get_threads(
...
+            return
+
+        yield window, window.get_name()
+
+        seen_windows.add(window)
+
+        # Walk adjacent windows
+        while len(seen_windows) < max_windows:
+            try:
```
ikelos: So this is kind of... isn't breadth first, or depth first?  It's kind of a hybrid?  (Since it doesn't do all nodes on a level, it does all nodes _from that child_ on that level).  Might be better to swap it to simple depth first (children then siblings)?

volatility3/framework/symbols/windows/extensions/gui.py
```diff
@@ -120,8 +130,193 @@ def get_threads(
...
+                window = window.spwndNext.dereference()
+            except exceptions.InvalidAddressException:
+                break
+
+            if window.vol.offset == 0:
```
ikelos: "not pythonic" sense tripped again...  5;P

volatility3/framework/symbols/windows/extensions/gui.py
```diff
@@ -120,8 +130,193 @@ def get_threads(
...
+        for window in seen_windows:
+            child = window
+
+            while len(seen_windows) + len(seen_children) < max_windows:
+                try:
+                    child = child.spwndChild
+                except exceptions.InvalidAddressException:
+                    break
+
+                if child.vol.offset == 0:
```
ikelos: And again

volatility3/framework/symbols/windows/extensions/gui.py
```diff
@@ -2,13 +2,17 @@
...
 from volatility3.framework import exceptions, constants, interfaces
 from volatility3.framework import objects
 from volatility3.framework.objects import utility
+from volatility3.framework.symbols.windows import extensions
 from volatility3.framework.symbols.windows.extensions import pool
 
+vollog = logging.getLogger(__name__)
+
 
 class tagWINDOWSTATION(objects.StructType, pool.ExecutiveObject):
```
ikelos: Can we put all the classes in this module inside a larger class (that can then be versioned)?  Otherwise changes to this require framework bumps.  I think it should be a catch-all, because no one's going to want to track the versions of each individual class in here.  We're trying to balance usability against upheaval from changes...

ikelos (commented):
So, a small logic flaw, a pet peeve and then a versioning decision.  Otherwise looks good, thanks!  5:)

volatility3/framework/plugins/windows/windows.py
atcuno: Changed it to the NotAvailable on None. I thought there was a recent update where format_hints returned NotAvailable when None was sent in. I guess I misunderstood that PR.

volatility3/framework/symbols/windows/extensions/gui.py
atcuno: Nice catch. Fixed and expanded.

volatility3/framework/symbols/windows/extensions/gui.py
atcuno: Fixed

volatility3/framework/symbols/windows/extensions/gui.py
atcuno: Fixed

volatility3/framework/symbols/windows/extensions/gui.py
atcuno: This is matching Vol2 directly. The first loop through `spwndNext` should catch everything on the same level. Then each window's children. Its also written this way so if we ever want to resurrect the screenshot plugin,which draws a PNG of all the windows on the desktop with their titles, then walking the tree this way preserves z-order.

volatility3/framework/symbols/windows/extensions/gui.py
atcuno: Good call. Fixed now.

atcuno:
@ikelos I put in your changes an re-ran the test plugins. For the extensions file, the diff is long, but all I did was tab every thing other, add the class, and then adjust the class_type references.

volatility3/framework/symbols/windows/extensions/gui.py
ikelos: Fair enough, but it's not strictly breadth first because children of children will get renderered before... children of sibling's (so it's a localized level, not the whole level of the tree)?  If that's how it worked and that preserves z-order then that's fine, just want to make sure we're walking the tree the right way.

ikelos:
Didn't I make a comment about it being called `windows.windows.Windows`?  I thought for sure somewhere I suggested making it `windows.gui.Gui` or `windows.gui.Windows`?

atcuno:
> Didn't I make a comment about it being called `windows.windows.Windows`? I thought for sure somewhere I suggested making it `windows.gui.Gui` or `windows.gui.Windows`?

That would be kind of confusing since already merged plugins windowstations, desktop, deskscan are all also GUI related.

ikelos (commented):
I'm not seeing where the version of the GUIExtensions is checked, or where the `class_types` are added to the symbol table?  Is that just an auxiliary class to be used in future plugins, or does the plugin in this PR need it (otherwise, why isn't in two PRs)?

atcuno:
> I'm not seeing where the version of the GUIExtensions is checked, or where the `class_types` are added to the symbol table? Is that just an auxiliary class to be used in future plugins, or does the plugin in this PR need it (otherwise, why isn't in two PRs)?

errr this was huge over sight on my part. I will add the versioning to the plugins (stations, desktops, deskscan, windows) now.

atcuno:
versioning is in now @ikelos

ikelos:
I think strictly, only `windowstations` requires it, since that's the only one that loads the class types?  Anything that depends on it should then depend on that requirement too, so... should all work together ok?

volatility3/framework/symbols/windows/extensions/gui.py
```diff
@@ -2,126 +2,333 @@
...
+                max_length=self.Length,
+                errors="replace",
+                encoding="utf16",
+            )
 
 
 class_types = {
-    "tagWINDOWSTATION": tagWINDOWSTATION,
-    "tagDESKTOP": tagDESKTOP,
+    "tagWINDOWSTATION": GUIExtensions.tagWINDOWSTATION,
```
ikelos: Might be clearer to make this dictionary part of the versioned class, so that it's easy to see where the requirement comes in (and it doesn't get missed on the plugins that use it).

ikelos (approved):
Looks good, thanks for getting it all in shape!  5:)

PR #1718, opened 2025-03-17
Fix first set of ELF parsing unhandled smear protection

atcuno:
The ELF parsing code is very, very rough when it comes to properly handling smear, returning `None` in failures of `get_*` functions and so on. These two code paths were making testing impossible.


ikelos (approved):
Seems reasonable, but at some point we probably want to move code like this into a versionable component somewhere...

PR #1719, opened 2025-03-17
Create versioned parent class for all plugins that enumerate Linux ke…

atcuno:
…rnel modules. Convert plugins to new method.

Ok @ikelos I am really happy to see how this turned out. Previously each of the enumeration plugins printed different columns and used different names. This gets us to where everything is consistent with a proper name.

Furthermore, given the maturity we now have with the versioning of modules, there is no need for a particular plugin to be the "main" one that the others inherit from (like thrdscan in Windows for threads plugins), and instead we can have a fully versioned chain of classes that plugins can use for our `self.implementation` style enumeration when the outputs will be the same.

You will see the in the comment I left, but this also sets us up to display Load Arguments, but jamming that into this PR would have spiraled it, so that will fill in here once merged.

### DIFF
--- a/volatility3/framework/plugins/linux/check_modules.py
+++ b/volatility3/framework/plugins/linux/check_modules.py
@@ -15,16 +15,33 @@ from volatility3.framework.interfaces import plugins
 vollog = logging.getLogger(__name__)
 
 
-class Check_modules(
-    linux_utilities_modules.ModuleDisplayPlugin, plugins.PluginInterface
-):
+class Check_modules(plugins.PluginInterface):
     """Compares module list to sysfs info, if available"""
 
     _version = (3, 0, 0)
     _required_framework_version = (2, 0, 0)
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(self.compare_kset_and_lsmod, *args, **kwargs)
+    @classmethod
+    def compare_kset_and_lsmod(
+        cls, context: str, vmlinux_name: str
+    ) -> Generator[extensions.module, None, None]:
+        kset_modules = linux_utilities_modules.Modules.get_kset_modules(
+            context=context, vmlinux_name=vmlinux_name
+        )
+
+        lsmod_modules = set(
+            str(utility.array_to_string(modules.name))
+            for modules in linux_utilities_modules.Modules.list_modules(
+                context=context, vmlinux_module_name=vmlinux_name
+            )
+        )
+
+        for mod_name in set(kset_modules.keys()).difference(lsmod_modules):
+            yield kset_modules[mod_name]
+
+    run = linux_utilities_modules.ModuleDisplayPlugin.run
+    _generator = linux_utilities_modules.ModuleDisplayPlugin.generator
+    implementation = compare_kset_and_lsmod
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
@@ -51,21 +68,3 @@ class Check_modules(
         cls, context: interfaces.context.ContextInterface, vmlinux_name: str
     ) -> Dict[str, extensions.module]:
         return linux_utilities_modules.Modules.get_kset_modules(context, vmlinux_name)
-
-    @classmethod
-    def compare_kset_and_lsmod(
-        cls, context: str, vmlinux_name: str
-    ) -> Generator[extensions.module, None, None]:
-        kset_modules = linux_utilities_modules.Modules.get_kset_modules(
-            context=context, vmlinux_name=vmlinux_name
-        )
-
-        lsmod_modules = set(
-            str(utility.array_to_string(modules.name))
-            for modules in linux_utilities_modules.Modules.list_modules(
-                context=context, vmlinux_module_name=vmlinux_name
-            )
-        )
-
-        for mod_name in set(kset_modules.keys()).difference(lsmod_modules):
-            yield kset_modules[mod_name]
--- a/volatility3/framework/plugins/linux/check_modules.py
+++ b/volatility3/framework/plugins/linux/check_modules.py
@@ -3,24 +3,25 @@
 #
 
 import logging
-from typing import List, Dict
+from typing import List, Dict, Generator
 
 import volatility3.framework.symbols.linux.utilities.modules as linux_utilities_modules
-from volatility3.framework import interfaces, renderers, deprecation
+from volatility3.framework import interfaces, deprecation
 from volatility3.framework.configuration import requirements
-from volatility3.framework.interfaces import plugins
 from volatility3.framework.objects import utility
-from volatility3.framework.renderers import format_hints
 from volatility3.framework.symbols.linux import extensions
 
 vollog = logging.getLogger(__name__)
 
 
-class Check_modules(plugins.PluginInterface):
+class Check_modules(linux_utilities_modules.ModuleDisplayPlugin):
     """Compares module list to sysfs info, if available"""
 
-    _version = (2, 0, 0)
+    _version = (3, 0, 0)
     _required_framework_version = (2, 0, 0)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(self.compare_kset_and_lsmod, *args, **kwargs)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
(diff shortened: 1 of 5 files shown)


atcuno:
This will need to be merged after https://github.com/volatilityfoundation/volatility3/pull/1717 is merged and then this branch rebased, but otherwise will be good.

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -684,3 +688,82 @@ def get_requirements(cls):
             )
 
         return reqs
+
+
+class ModuleDisplayPlugin(plugins.PluginInterface):
```
ikelos: This seems like the wrong place for it, and inheriting from the plugin interface means it may get listed as main plugin (our code to list plugins looks for anything that inherits from plugin).  Are we sure that's what we want?  I'd much prefer any plugins we have to live *as* a plugin in the right directory.

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -29,9 +32,9 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                 architectures=architectures.LINUX_ARCHS,
             ),
             requirements.VersionRequirement(
-                name="linux_utilities_modules",
-                component=linux_utilities_modules.Modules,
-                version=(3, 0, 0),
+                name="linux_utilities_modules_module_display_plugin",
```
ikelos, line 71: It would surely be better to define classmethod generator and run methods and then just proxy them at this point?  We're really not getting anything from inheritance other than complexity?

ikelos (commented):
This isn't passing some of the tests at the moment.

Also I really uncertain as to the effects of inheriting from a plugin that doesn't live under an os directory.  I'd really strongly advise that we only proxy methods from the class, that will allow the methods to stay shared, with minimal duplicated code and ensures we don't have to deal with any shared complexity (like a plugin inheritting from itself before it knows whether its own version passes the requirements.  I can't see a compelling reason to use inheritance for the sake of `run = ModuleDisplayPlugin.run` (although we'll need to think about binding of methods to make sure that works properly)...

atcuno:
Yes, the 1 test would break until I rebase with the netfilter PR you just merged. That is a really good point on inheriting to look like a plugin + re-using \.run. I will adjust to that.

volatility3/framework/plugins/linux/hidden_modules.py
atcuno: Can `__init__` of each plugin just set `.run` and and `._generator` to the shared function?  and Where would that shared functons best live now? I was trying to avoid one plugin becoming the container (say lsmod) that the other plugins then need to import and depend on.

volatility3/framework/plugins/linux/check_modules.py
```diff
@@ -3,25 +3,29 @@
...
 from volatility3.framework.symbols.linux import extensions
+from volatility3.framework.interfaces import plugins
 
 vollog = logging.getLogger(__name__)
 
 
-class Check_modules(plugins.PluginInterface):
+class Check_modules(
+    linux_utilities_modules.ModuleDisplayPlugin, plugins.PluginInterface
+):
```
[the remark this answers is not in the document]
ikelos: I just wouldn't import from it at all.  We're proxying, not inheritting, so just assign the run method to the run method from the other class.  Self should go in as the right thing, so you ought to get the right implementation.  In a worst case, you can just write a short few lines that calls it direct.

volatility3/framework/plugins/linux/check_modules.py
```diff
@@ -3,25 +3,29 @@
...
     """Compares module list to sysfs info, if available"""
 
-    _version = (2, 0, 0)
+    _version = (3, 0, 0)
     _required_framework_version = (2, 0, 0)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(self.compare_kset_and_lsmod, *args, **kwargs)
```
ikelos: I mean this is nice, but not worth it, just set a self variable in here, and pull it back in the proxied method.

volatility3/framework/plugins/linux/hidden_modules.py
ikelos: Yeah, that should be ok, but you can just define it straight in the class I think?  So

```
class Blah:
  run = linux_utilities_modules.ModuleDisplayPlugin.run
```
?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -684,3 +687,82 @@ def get_requirements(cls):
...
+    _required_framework_version = (2, 0, 0)
+
+    framework.require_interface_version(*_required_framework_version)
+
+    def __init__(self, implementation, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.implementation = implementation
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
```
ikelos, line 705: Add the results of this to the individual class get_requirements.

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -684,3 +687,82 @@ def get_requirements(cls):
...
+
+
+class ModuleDisplayPlugin(interfaces.configuration.VersionableInterface):
+    """
+    Plugins that enumerate kernel modules (lsmod, check_modules, etc.)
+    must inherit from this class to have unified output columns across plugins.
+    The constructor of the plugin must call super() with the `implementation` set
+    """
+
+    _version = (1, 0, 0)
```
ikelos, line 699: Don't need these local variables, since they shouldn't be being inherited anyway (you can if you really want, but the plugins should be defining their own needs

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -684,3 +687,82 @@ def get_requirements(cls):
...
+                name="linux_utilities_modules",
+                component=Modules,
+                version=(3, 0, 0),
+            ),
+            requirements.VersionRequirement(
+                name="linux-tainting", component=tainting.Tainting, version=(1, 0, 0)
+            ),
+        ]
+
+    def _generator(self):
```
ikelos: This might give a ruff warning, because it's a private member, but you can actually make it public and just assign it to a private member in the new class?

```
  _generator = modules.ModuleDisplayPlugin.generator
```

for example?

ikelos (commented):
Looking much better, but there's nothing to be gained by the multiple inheritance.  Literally just reach into the classes and assign the methods to the new classes.  It *should* all work (because `self` is included explicitly, so it'll refer to the instantiated class, not the `ModuleDisplayPlugin` class, which likely will never get instantiated ever.  It's just acting as a code wrapper to keep it all nice and warm and bundled up for the plugins that actually want to use it...  5:)

atcuno:
Nice idea! I think its sorted now. Also, if you see in the diff functions moving, its so they could be referenced from the code under them.

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Sorry, I realized we version this directly, so it's important for this to have its own version!  5:S

ikelos (approved):
Looks good, let's get it landed  5:)

ikelos:
So I said: 
>  I'd really strongly advise that we only proxy methods from the class, that will allow the methods to stay shared, with minimal duplicated code and ensures we don't have to deal with any shared complexity.

> Looking much better, but there's nothing to be gained by the multiple inheritance.

To which you eventually said:
> I think its sorted now. 

And I then merged it in, seemingly because I was getting tired of reworking it and because it was an overly large pull request.  I'm sorry, I shouldn't have accepted it, I don't think I thought this through full before landing it.  You sounded as though you'd fixed the inheritance that I objected to twice, but there's still inheritance being used and is now causing some headaches in a circular dependency because this extension module is actually partial plugins rather than discrete code blocks each doing their own thing (see #1773).  

Please can we can this refactored ASAP.  No inheritance please.  None.  Only code blocks and version checks...

ikelos:
Ok, having calmed a bit (but only a bit) and scoped out exactly what's going on, it seems we jury rigged inheritance.  It's not as bad as I thought, but trying to pull in complex functions (like run and generator) removes the ability for plugin to easily be able to override settings (or for us to separate concerns when, for example, part of the dumping functionality needs to be refactored).  I'm working up a PR that should improve this, but I suspect that the `ModuleDisplayPlugin` will not exist in its current form for long.

ikelos:
Ok, I've worked up #1801 to get this slightly more in order.

PR #1724, opened 2025-03-19
Windows: Fix raw Dpc offset calculation

dgmcdona:
The original code was still returning this as a pointer that ended up
dereferenced in later steps. However, this pointer value actually needs
to be cast to an `unsigned long long` and decoded first.

### DIFF
--- a/volatility3/framework/objects/__init__.py
+++ b/volatility3/framework/objects/__init__.py
@@ -410,6 +410,19 @@ class Pointer(Integer):
         value = int.from_bytes(data, byteorder=endian, signed=signed)
         return value & mask
 
+    def get_raw_value(self) -> int:
+        formats = {
+            4: "I",
+            8: "Q",
+        }
+        length = self.vol.data_format.length
+        endian = self.vol.data_format.byteorder
+        raw_data = self._context.layers[self.vol.layer_name].read(
+            self.vol.offset, length
+        )
+        struct_format = ("<" if endian == "little" else ">") + formats[length]
+        return struct.unpack(struct_format, raw_data)[0]
+
     def dereference(
         self, layer_name: Optional[str] = None
     ) -> interfaces.objects.ObjectInterface:
--- a/volatility3/framework/objects/__init__.py
+++ b/volatility3/framework/objects/__init__.py
@@ -402,26 +402,35 @@ class Pointer(Integer):
         pointer should be recast.  The "pointer" must always live within
         the space (even if the data provided is invalid).
         """
+        mask = context.layers[object_info.native_layer_name].address_mask
+        new = (
+            cls._get_raw_value(
+                context, data_format, object_info.layer_name, object_info.offset
+            )
+            & mask
+        )
+        return new
+
+    @classmethod
+    def _get_raw_value(
+        cls,
+        context: interfaces.context.ContextInterface,
+        data_format: DataFormatInfo,
+        layer_name: str,
+        offset: int,
+    ) -> int:
         length, endian, signed = data_format
         if signed:
             raise ValueError("Pointers cannot have signed values")
-        mask = context.layers[object_info.native_layer_name].address_mask
-        data = context.layers.read(object_info.layer_name, object_info.offset, length)
+        data = context.layers.read(layer_name, offset, length)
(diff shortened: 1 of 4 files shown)


dgmcdona:
Looking back through the history, I noticed that this originally was an `unsigned long long` in the initial PR but was changed to `pointer` due to concerns about compiler changes and the potential for that type to not exist. Is there some way that I haven’t been able to figure out to access the raw pointer without having it automatically dereference?

ikelos:
I think just `int(self.Dpc)` should do it?  Pointers are subclasses on Integers, so the `int` call should just turn it into a number?  You would then lose all the pointerishness of it though...

dgmcdona:
I thought so as well, but that doesn't seem to be the case:
```
[ins] In [5]: int(self.Dpc)
Out[5]: 56772929057211

[ins] In [6]: self.Dpc.cast("unsigned long long")
Out[6]: 3104162816094171579

[ins] In [7]: self.Dpc.dereference().vol.offset
Out[7]: 56772929057211
```

ikelos:
Errr... Is it getting masked to the address space (because pointers need to fit within the address space), so the high bits are getting masked off?  Oh, yep, that must be [what's happening ](https://github.com/volatilityfoundation/volatility3/blob/369f37ec4366bf5127aac596479b03c5b0819749/volatility3/framework/objects/__init__.py#L411)

```
>>> hex(3104162816094171579)
'0x2b1433a27a44f1bb'
>>> hex(56772929057211)
'0x33a27a44f1bb'
```

Ok, well, the difficulty we have is that clang changed from one name for `unsigned long long`, to another.  I think it only affected linux but it kinda goes to show that reliance on a specific type may one day cause us problems.  The pointer class really should be handling anything that is supposed to be a pointer (I wish OS devs would stop abusing pointers to stash extra data or obfuscate them!).  So, we've got options:

* Stick with the named types we've got and potentially have to special case things if they screw up in the future.
* Write an adaptor that can reread the data at the pointer (and the knowledge of length and endianness from the pointer) to get to the original int - possibly a little convoluted but has the least impact on the existing framework
* Add a method to a pointer to return the original data, although we'd need to be really careful of the name, because it would mask any member of the pointer's subtype.  Probably we'd call it `get_raw_value` or something? 

Which do you think we should go for?  I'm leaning towards one of the last two, but I can't decide which...

dgmcdona:
I liked option 3 so took a crack at it in 7b9fb916722a0678b666be0f02dd59b13c769951, let me know what you think.

volatility3/framework/objects/__init__.py
```diff
@@ -410,6 +410,19 @@ def _unmarshall(
...
+            4: "I",
+            8: "Q",
+        }
+        length = self.vol.data_format.length
+        endian = self.vol.data_format.byteorder
+        raw_data = self._context.layers[self.vol.layer_name].read(
+            self.vol.offset, length
+        )
+        struct_format = ("<" if endian == "little" else ">") + formats[length]
+        return struct.unpack(struct_format, raw_data)[0]
```
ikelos: I think it would be better to use `int.from_bytes` so that it's consistent with the `unmarshall` call above.  You could probably just split the unmarshall above into two functions and expose the `get_raw_value` one...

ikelos (commented):
Happy with the proposed solution, but we should keep it in sync with the unmarshalling call (and `int.from_bytes` seem better than constructing a format string and then formatting it)...

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -1268,7 +1254,7 @@ def get_dpc(self):
             )
 
             low_byte = (wait_never) & 0xFF
-            entry = utility.rol(self.get_raw_dpc() ^ wait_never, low_byte)
+            entry = utility.rol(self.Dpc.get_raw_value() ^ wait_never, low_byte)
```
ikelos, line 1257: This call now need parameters?

ikelos: Nope, scratch that, sorry, didn't see the additional method.

ikelos (approved):

PR #1727, opened 2025-03-19
Hopefully final round of kallsym fixes

atcuno:


ikelos (commented):
This is currently failing tests (but I can't immediately see why), otherwise it looks fine...

atcuno:
This is passing now @ikelos  - this was a really good catch by the tests. The `not` check was bringing in False entries for exported (its a boolean), so I broke it out more.

ikelos (approved):

PR #1728, opened 2025-03-21
Add performance event plugin to detect eBPF malware

atcuno:
This was missed by mistake when ftrace and tracepoint was added

### DIFF
--- a/volatility3/framework/plugins/linux/tracing/perf_events.py
+++ b/volatility3/framework/plugins/linux/tracing/perf_events.py
@@ -1,0 +1,138 @@
+# This file is Copyright 2025 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import logging
+from typing import List, Tuple, Generator, Optional
+
+from volatility3.framework import renderers, interfaces, constants, exceptions
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.configuration import requirements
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/linux/tracing/perf_events.py
```diff
@@ -0,0 +1,138 @@
...
+                name="pslist", plugin=pslist.PsList, version=(4, 0, 0)
+            ),
+        ]
+
+    @classmethod
+    def list_perf_events(cls, context, vmlinux_module_name: str) -> Generator[
+        Tuple[
+            interfaces.objects.ObjectInterface,
+            interfaces.objects.ObjectInterface,
+            Optional[str],
```
ikelos, line 41: I'm not sure these are just `Optional[str]`?  I think they're `Union[BaseAbsentValue, str]`?  It's probably best to use `BaseAbsentValue` (so that you can distinguish the reason the data's not available), so it's just a signature change that needs to happen...

ikelos (commented):
Seems good once the signatures fixed up...

volatility3/framework/plugins/linux/tracing/perf_events.py
atcuno: Really nice catch... this broke the rules we enforced on other plugins for the gatherers to return None and `_generator` to send the `renderers.Not...` on None. Changed these around.

ikelos (approved):

PR #1729, opened 2025-03-23
Volshell: Add byteorder argument for display_* functions

the-rectifier:
Add a byte order argument to `display_*` functions, to display data in different endianness


ikelos (approved):
Looks reasonable to me, thanks.  5:)

PR #1730, opened 2025-03-24
Add win32 start address listing. Add paths for both thread starting a…

atcuno:
…ddress types

@ikelos this is the one I mentioned some weeks ago and just got to finish and test today. Previously, we only listed the address of the `start address` thread address, which is not very helpful.

Now we also list the second thread start address, the `win32` one + we list the full path for all the threads.

### DIFF
--- a/volatility3/framework/plugins/windows/thrdscan.py
+++ b/volatility3/framework/plugins/windows/thrdscan.py
@@ -3,12 +3,12 @@
 ##
 import logging
 import datetime
-from typing import Callable, Iterable
+from typing import Callable, Iterable, Tuple, Optional, Dict
 
 from volatility3.framework import renderers, interfaces, exceptions
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints
-from volatility3.plugins.windows import poolscanner
+from volatility3.plugins.windows import poolscanner, pe_symbols
 from volatility3.plugins import timeliner
 
 vollog = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


volatility3/framework/plugins/windows/thrdscan.py
```diff
@@ -67,39 +67,109 @@ def scan_threads(
...
             return None
 
+        if vads_cache is not None:
+            vads = pe_symbols.PESymbols.get_vads_for_process_cache(
+                vads_cache, owner_proc
+            )
+            # no vads = terminated/smeared, pid 4 = kernel = don't check VADs
+            if (
+                owner_proc_pid != 4
+                and owner_proc.InheritedFromUniqueProcessId != 4
```
ikelos, line 112: Do we need to check all the parents, or will the one level of InheritedFrom be enough?

ikelos (commented):
Just one quick question, otherwise looks good, thanks.  5:)

volatility3/framework/plugins/windows/thrdscan.py
atcuno: Just one level is good. Kernel process (Registry, MemCompression, etc.) will always have a parent of 4

ikelos:
D'oh, looks like it tinkered with something another PR messed with.  It's good to go once that's fixed up though...

ikelos (approved):

atcuno:
I just resolved here through the web. It was the PR about version requirements.

PR #1731, opened 2025-03-24
Add delete on close detection to process ghosting. Update plugin to c…

atcuno:
…urrent coding flow

@ikelos This adds the Delete On Close detection plus updates all of the plugin to our current coding standards.

### DIFF
--- a/volatility3/framework/plugins/windows/processghosting.py
+++ b/volatility3/framework/plugins/windows/processghosting.py
@@ -99,7 +99,7 @@ class ProcessGhosting(interfaces.plugins.PluginInterface):
         except exceptions.InvalidAddressException:
             delete_pending = None
 
-        if delete_pending and delete_pending == 1:
+        if delete_pending == 1:
             yield file_object.vol.offset, delete_pending, None
 
     @classmethod
@@ -158,7 +158,7 @@ class ProcessGhosting(interfaces.plugins.PluginInterface):
         )
         if not has_imagefilepointer:
             vollog.warning(
-                "ImageFilePointer checks are only supported on Windows 10 builds when the ImageFilePointer member of _EPROCESS is present"
+                "ImageFilePointer checks are only supported on Windows 10+ builds when the ImageFilePointer member of _EPROCESS is present"
             )
 
         for proc in procs:
--- a/volatility3/framework/plugins/windows/processghosting.py
+++ b/volatility3/framework/plugins/windows/processghosting.py
@@ -2,21 +2,23 @@
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
 import logging
-import contextlib
+
+from typing import Optional, Tuple, Generator, Dict
 
 from volatility3.framework import interfaces, exceptions
 from volatility3.framework import renderers
 from volatility3.framework.configuration import requirements
 from volatility3.framework.objects import utility
 from volatility3.framework.renderers import format_hints
-from volatility3.plugins.windows import pslist
+from volatility3.plugins.windows import pslist, vadinfo
 
 vollog = logging.getLogger(__name__)
 
 
 class ProcessGhosting(interfaces.plugins.PluginInterface):
-    """Lists processes whose DeletePending bit is set or whose FILE_OBJECT is set to 0"""
+    """Lists processes whose DeletePending bit is set or whose FILE_OBJECT is set to 0 or Vads that are DeleteOnClose"""
 
+    _version = (1, 0, 0)
     _required_framework_version = (2, 4, 0)
 
     @classmethod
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/processghosting.py
```diff
@@ -89,7 +202,9 @@ def run(self):
                 ("PID", int),
                 ("Process", str),
                 ("FILE_OBJECT", format_hints.Hex),
-                ("DeletePending", str),
+                ("DeletePending", int),
+                ("DeleteOnClose", int),
+                ("Base", format_hints.Hex),
```
ikelos: Is this the ordering we want?  Feels like Base should be earlier (next to Process maybe?)  Entirely your call, just looks weird after two small int values?

volatility3/framework/plugins/windows/processghosting.py
```diff
@@ -33,52 +35,163 @@ def get_requirements(cls):
...
+            except exceptions.InvalidAddressException:
+                file_object = 0
+
+        # ImageFilePointer equal to 0 means process ghosting or similar techniques were used
+        else:
+            file_object = 0
+
+        # delete_pending besides 0 or 1 = smear
+        if isinstance(delete_pending, int) and delete_pending not in [0, 1]:
+            vollog.debug(
```
ikelos, line 69: Just a debug message?  Feels a bit strange if it's not an `int` (that's kind of on us, unless the typing changes upstream by MS, in which case we probably want to know)?

volatility3/framework/plugins/windows/processghosting.py
```diff
@@ -33,52 +35,163 @@ def get_requirements(cls):
...
-            # ImageFilePointer equal to 0 means process ghosting or similar techniques were used
-            else:
-                file_object = 0
+        try:
+            delete_pending = file_object.DeletePending
+        except exceptions.InvalidAddressException:
+            delete_pending = None
 
-            if isinstance(delete_pending, int) and delete_pending not in [0, 1]:
+        if delete_pending and delete_pending == 1:
```
ikelos: This seems like a different check and wouldn't it be accomplished be just `delete_pending == 1` that's kind of a more specific version of `delete_pending`?

volatility3/framework/plugins/windows/processghosting.py
```diff
@@ -33,52 +35,163 @@ def get_requirements(cls):
...
+            # these checks have no meaning for private memory areas
+            if vad.get_private_memory() == 1:
+                continue
+
+            try:
+                if vad.has_member("ControlArea"):
+                    control_area = vad.ControlArea
+                elif vad.has_member("Subsection"):
+                    control_area = vad.Subsection.ControlArea
+                # We got here from a short vad, likely smear
```
ikelos, line 137: Worth a debug message, or no?  Also feels like the comment should be under the else, not above it (even though indentation suggests where it applies).

volatility3/framework/plugins/windows/processghosting.py
```diff
@@ -33,52 +35,163 @@ def get_requirements(cls):
...
-                        format_hints.Hex(file_object),
-                        delete_pending,
-                        path,
-                    ),
+        has_imagefilepointer = kernel.get_type("_EPROCESS").has_member(
+            "ImageFilePointer"
+        )
+        if not has_imagefilepointer:
+            vollog.warning(
+                "ImageFilePointer checks are only supported on Windows 10 builds when the ImageFilePointer member of _EPROCESS is present"
```
ikelos: Is this specific to windows 10, or does it apply to 7 or 11?  Seems strange to be saying windows 10 specifically here?  Can this error never happen on a different version of windows?

ikelos (commented):
No showstoppers, just me questioning things that are probably fine.  Tell me to go away if none of them are useful/important...  5;)

volatility3/framework/plugins/windows/processghosting.py
atcuno: We set it to None on line 52 as the process might not have the member. So then we check if its an int (was set in the ImageFilePointer path). Just a debug message as its smeared out if its anything but 0 or 1.

volatility3/framework/plugins/windows/processghosting.py
atcuno: Not worth a debug message, if its non-private memory and maps a file but doesn't have a control area then its smeared out of control.

volatility3/framework/plugins/windows/processghosting.py
atcuno: Nice catch. Should be "10+" like other plugins.

volatility3/framework/plugins/windows/processghosting.py
atcuno: I will change it to be more like the other plugins.

volatility3/framework/plugins/windows/processghosting.py
atcuno: Yes will just do the == check.

volatility3/framework/plugins/windows/processghosting.py
atcuno: Done in latest commit.

volatility3/framework/plugins/windows/processghosting.py
atcuno: addressed in latest commit.

volatility3/framework/plugins/windows/processghosting.py
atcuno: addressed in latest commit.

atcuno:
ok @ikelos check my new comments and commits

ikelos (approved):

Issue #1732, opened 2025-03-24
PESymbols: Discrepancy because of symbol order


the-rectifier:
I encountered a strange bug in `pe_symbols.PESymbols._resolve_symbols_through_methods()` where the resulting symbols depend on their order in the filter function.

**Context**
Volatility Version: 2.25.0
Operating System: Linux

**To Reproduce**
I am running this volshell script using: `python volshell.py -f ../artefacts/trash.elf -w --script test.py`

```python
from volatility3.framework.symbols import intermed
from volatility3.framework.symbols.windows.extensions import pe
from volatility3.plugins.windows import pe_symbols


pe_table_name = intermed.IntermediateSymbolTable.create(
    self.context, self.config_path, "windows", "pe", class_types=pe.class_types
)

filter_modules = {
    "ntdll.dll": {
        pe_symbols.wanted_names_identifier: [
            "RtlGetUnloadEventTrace",
            "RtlpUnloadEventTrace",
        ]
    }
}

# filter_modules = {
#     "ntdll.dll": {
#         pe_symbols.wanted_names_identifier: [
#             "RtlpUnloadEventTrace",
#             "RtlGetUnloadEventTrace",
#         ]
#     }
# }

found_symbols = pe_symbols.PESymbols.addresses_for_process_symbols(
    self.context, self.config_path, self.config["kernel"], symbols=filter_modules
)

print(found_symbols)
```

Notice that the only difference between the two filter dicts is the order of the symbols. Debugging this, I was able to conclude that:

- The first (uncommented) filter resolves only `RtlGetUnloadEventTrace` using PDB and fails on `RtlpUnloadEventTrace` using both methods.
- The latter filter resolves `RtlpUnloadEventTrace` through PDB and `RtlGetUnloadEventTrace` through exports which is strange because the first filter resolved it through PDB

Unrelated to the above:

In [pe_symbols.py#L416](https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/pe_symbols.py#L416) `unresolved_symbols` is of type `dict` making the `symbol` variable the key, in this case `names`:

```
DEBUG    volatility3.framework.symbols.windows.pdbutil: Found ntdll.pdb: EE7D5AA2BC0ADFA252694A6749648950-1
DEBUG    volatility3.framework.symbols.windows.pdbutil: Using symbol library: ntdll.pdb/EE7D5AA2BC0ADFA252694A6749648950-1
[...]

PR #1733, opened 2025-03-24
Add a --script-only flag that exits after the given volshell script i…

atcuno:
…s completed.

@ikelos While testing https://github.com/volatilityfoundation/volatility3/issues/1732, I remembered that I had a local patch to my volshell for scripts that I never made into a PR. I am fixing that now. Currently you cannot sanely run volshell scripts across many samples due to volshell dropping into the shell after the script runs. This PR adds a `--script-only` flag that, when set, exits volshell after the script is completed. When `-q` is set and no `-v`, only the script output is returned, which is what is needed for quick artifact gathering. If `--script-only` is not set then volshell loads like normal after script completion.

### DIFF
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -60,7 +60,13 @@ class Volshell(interfaces.plugins.PluginInterface):
                     description="File to load and execute at start",
                     default=None,
                     optional=True,
-                )
+                ),
+                requirements.BooleanRequirement(
+                    name="script-only",
+                    description="Exit volshell after the script specified in --script completes",
+                    default=False,
+                    optional=True,
+                ),
             ]
         return reqs + [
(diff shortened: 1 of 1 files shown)


ikelos:
Should we default to this if there's a script and have a `--continue` or `--interactive` flag or something to say you want to continue using the shell afterwards?  I think an additional flag probably is the best way of handling it (even though it adds yet another option) just want to make sure we make the flag the one that'll get used least often (so the common case doesn't need an extra flag)...  What do you think?

Otherwise looks fine, happy to merge it once we decide which way round it should be.

atcuno:
I think the default of jumping into the shell is fine as its reasonable to make scripts to compute something (like a series of addresses) that you then want to explore with db/dt/dq etc. So you can have the script print them then start analysis.

PR #1734, opened 2025-03-24
Framework: Replace PluginRequirements

### LINKED ISSUE #1644: Convert `PluginRequirement`s to `VersionRequirement`s and deprecate `PluginRequirement`
`PluginRequirement` is literally an alias of `VersionRequirement`, if there's a need to differentiate between plugins and other components, `isinstance(component, PluginInterface)` could be used, so there's no real benefit to having the two names.

I was hoping one would `PluginRequirement` would die out in favour of VersionRequirement but that hasn't happen.  I propose a mass sweep of the existing code to make the change, and then deprecating PluginRequirement until the next major version bump.

dgmcdona:
This replaces all uses of `requirements.PluginRequirements` with `requirements.VersionRequirement`.

closes #1644

### DIFF
--- a/doc/source/simple-plugin.rst
+++ b/doc/source/simple-plugin.rst
@@ -53,9 +53,9 @@ to be able to run properly.  Any that are defined as optional need not necessari
                     description = "Process IDs to include (all other processes are excluded)",
                     optional = True
                 ),
-                requirements.PluginRequirement(
+                requirements.VersionRequirement(
                     name = 'pslist',
-                    plugin = pslist.PsList,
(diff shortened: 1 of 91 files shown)


atcuno:
This is meant to close https://github.com/volatilityfoundation/volatility3/issues/1644

ikelos (approved):
Hehehe, you really want to get everything in order, huh?  5;P  Well, we *could* bump the PATCH version number of every plugin we're changing but... I don't think it'd be worth it.  5;P  Looks good, thanks!  5:D

PR #1735, opened 2025-03-24
Fix issue when pe_symbols limited to searching one process. Remove ne…

### LINKED ISSUE #1732: PESymbols: Discrepancy because of symbol order

atcuno:
…ed to track symbol indexes. Provide much more useful debugging information. Fixes #1732

Closes #1732

### DIFF
--- a/volatility3/framework/plugins/windows/pe_symbols.py
+++ b/volatility3/framework/plugins/windows/pe_symbols.py
@@ -642,7 +642,7 @@ class PESymbols(interfaces.plugins.PluginInterface):
             symbol_resolver: method in a layer to resolve the symbols
 
         Returns:
-            Tuple[str, int, str, int]: the index and value of the found symbol in the wanted list, and the name and address of resolved symbol
+            Tuple[str, str, int]: the symbol identifier (key) of the found symbol in the wanted list, and the name and address of resolved symbol
         """
         if (
             wanted_names_identifier not in wanted_symbols
@@ -673,9 +673,9 @@ class PESymbols(interfaces.plugins.PluginInterface):
                     if symbol_value:
                         # yield out deleteion key, deletion index, symbol name, symbol address
                         if symbol_key == wanted_names_identifier:
-                            yield symbol_key, wanted_value, symbol_value  # type: ignore
+                            yield symbol_key, wanted_value, symbol_value
                         else:
-                            yield symbol_key, symbol_value, wanted_value  # type: ignore
+                            yield symbol_key, symbol_value, wanted_value
 
                 for value in all_wanted:
                     vollog.debug(
--- a/volatility3/framework/plugins/windows/pe_symbols.py
+++ b/volatility3/framework/plugins/windows/pe_symbols.py
@@ -229,7 +229,6 @@ class ExportSymbolFinder(PESymbolFinder):
         Returns:
             address: the address of the symbol, if found
         """
-
         for export in self._symbol_module:
             sym_name = self._get_name(export)
             if sym_name and sym_name == name:
@@ -413,8 +412,10 @@ class PESymbols(interfaces.plugins.PluginInterface):
         )
 
         for mod_name, unresolved_symbols in missing_symbols.items():
-            for symbol in unresolved_symbols:
-                vollog.debug(f"Unable to resolve symbol {symbol} in module {mod_name}")
+            for symbol_key, symbols in unresolved_symbols.items():
+                vollog.debug(
+                    f"Unable to resolve symbols {symbols} of type {symbol_key} in module {mod_name}"
+                )
 
         return found_symbols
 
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/pe_symbols.py
```diff
@@ -632,7 +633,7 @@ def _find_symbols_through_exports(
     def _get_symbol_value(
         wanted_symbols: filter_module_info,
         symbol_resolver: PESymbolFinder,
-    ) -> Generator[Tuple[str, int, str, int], None, None]:
+    ) -> Generator[Tuple[str, str, int], None, None]:
```
ikelos, line 636: The docstring needs changing too then...

volatility3/framework/plugins/windows/pe_symbols.py
```diff
@@ -661,15 +662,25 @@ def _get_symbol_value(
...
+                for wanted_value in wanted_symbols[symbol_key]:
+                    all_wanted.append(wanted_value)
+
+                for value_index, wanted_value in enumerate(all_wanted):
+                    symbol_value = symbol_getter(wanted_value)
                     if symbol_value:
                         # yield out deleteion key, deletion index, symbol name, symbol address
                         if symbol_key == wanted_names_identifier:
-                            yield symbol_key, value_index, wanted_value, symbol_value  # type: ignore
+                            yield symbol_key, wanted_value, symbol_value  # type: ignore
```
ikelos: Don't suppose we can do anything about those type: ignore things can we?  Given we're modifying the code, I'd like to just check if they're still need or not (now things are typed a little better and  the checkers are more advnaced)...

ikelos (commented):
Just a docstring change (but also I'd like to check if the type ignore instructions are really still needed or not)...

volatility3/framework/plugins/windows/pe_symbols.py
atcuno: I can remove it for now since we aren't using mypy in Actions yet. We can find a proper fix when that time comes.

volatility3/framework/plugins/windows/pe_symbols.py
atcuno: Nice catch.

atcuno:
Check now @ikelos

PR #1737, opened 2025-03-25
Windows: Additional testing

Abyss-W4tcher:
Hi,

This PR extends coverage of Windows testing with:

- 42 additional tests;
- A Windows10 Crash dump sample;
- APIs and mechanisms to check deterministic previous plugins outputs. 

"Large" testing datasets are stored in separate JSON files, but more compact ones are kept inline for better readability flow and convenience. 

To optimize runtime, most scanner plugins are run against the Windows XP sample when possible. Of course, running more plugins incidentally increases PR actions runtime by approximately 9 minutes 30. A challenge with memory forensics is that there aren't really smaller features/plugins that we can skip checks for, given that results need to be idempotent from one version of Vol3 to another.

One question that some might have, is why don't we design tests by running all plugins once, store outputs somewhere and then verify exact matches after PRs? The answer is that it'll be difficult to maintain and require constant edits since plugins updates might filter out some noise, add columns, extend fields etc. It *should* be better to pinpoint key results to look for in outputs. Although, this design can still be applied to plugins with a very limited number of outputted lines (see `test/plugins/windows/test_data/windows.info.Info.json`).

GitHub actions run with estimated runtimes:

- https://github.com/Abyss-W4tcher/volatility3/actions/runs/14057618953/job/39360533665#step:8:108


Abyss-W4tcher:
There is a traceback in `thrdscan` that needs to be addressed, it should be resolved soon however.

.github/workflows/test.yaml
```diff
@@ -47,8 +51,8 @@ jobs:
         pytest ./test/plugins/linux/linux.py --volatility=volshell.py --image-dir=./test_images -k test_linux_volshell -v
 
         # Volatility
-        pytest ./test/plugins/windows/windows.py --volatility=vol.py --image-dir=./test_images -k "test_windows and not test_windows_volshell" -v
-        pytest ./test/plugins/linux/linux.py --volatility=vol.py --image-dir=./test_images -k "test_linux and not test_linux_volshell" -v
+        pytest ./test/plugins/windows/windows.py --volatility=vol.py --image=./test_images/win-10_19041-2025_03.dmp -k "test_windows and not test_windows_volshell" -v --durations=0
```
ikelos, line 54: Won't this test fewer images, using a specific one rather than the whole directory?

ikelos (commented):
Looks ok (although I'm slightly concerned about how brittle the hexdump data in the test verification results are, since changes in formatting will break the tests).  Just one question, but otherwise good to go...

ikelos:
(And get it so the tests pass, of course).  5;P

ikelos:
`range_info_for_address`  seems to be being handed `None`...

.github/workflows/test.yaml
Abyss-W4tcher: It would be ideal to run each plugin against multiple images, but unfortunately runtime would drastically increase if we did so (especially scanner plugins)...

Abyss-W4tcher:
Another small break but https://github.com/volatilityfoundation/volatility3/issues/1741 was quickly fixed by https://github.com/volatilityfoundation/volatility3/pull/1742. I'll trigger the tests again afterwards.

ikelos (approved):

ikelos:
Is this PR dependent on another?  One of the tests is now failing, and I dunno if it's the one this changes and whether you want it merging now or not?  It may be you need to rebase it now?

Abyss-W4tcher:
The tests were failing due to https://github.com/volatilityfoundation/volatility3/issues/1741, which was introduced yesterday and fixed today

PR #1738, opened 2025-03-25
Framework: Missing `VersionRequirement`s

dgmcdona:
This PR:
- Adds the `VersionableInterface` class as a superclass of `TimeLinerInterface` to make it consistent with other versioned interfaces.
- Audits the codebase for uses of versioned components that are not listed in the consumer's `get_requirements()` classmethod, and adds them as needed.


ikelos (approved):
Good point, seems reasonable.  Just want to think through how this will fail if it's using the wrong version of the interface?  Presumably it won't provide one of the methods that the interface lists, but it should still run to the point of checking the requirements so... yeah, seems worthwhile, thanks.  5:)

PR #1739, opened 2025-03-26
Fix checks in thrdscan that broke tests

atcuno:


ikelos (approved):

atcuno:
@ikelos I pushed again to this. We want to let kernel threads show, just without paths, and then keep get paths for userland threads. The latest push covers that. Also, the test was pretty weak so I made it check more values. This will get replaced soon with AbyssWatcher's more advanced tests.

atcuno:
@Abyss-W4tcher your tests should pass after this is merged.

PR #1740, opened 2025-03-26
Add module extraction API. Add plugin to directly extract modules. Ho…

atcuno:
…ok module dumping into existing module listing plugins.

@ikelos this is the big extraction LKM PR. After this, users can extract through the existing plugins or the new, direct module extract one. I ran over 99 Linux samples and grepped for tracebacks obviously, but also debug messages we wouldn't expect to hit, and also ran readelf with -s and -S over them all to look for busted strings or sections.

`hidden_modules` is missing from this PR for the conversion as I found a bad bug in it, so that will be a separate PR right after this is merged.

### DIFF
--- a/volatility3/framework/plugins/linux/check_modules.py
+++ b/volatility3/framework/plugins/linux/check_modules.py
@@ -18,7 +18,7 @@ vollog = logging.getLogger(__name__)
 class Check_modules(plugins.PluginInterface):
     """Compares module list to sysfs info, if available"""
 
-    _version = (3, 0, 0)
+    _version = (3, 0, 1)
     _required_framework_version = (2, 0, 0)
 
     @classmethod
@@ -55,6 +55,12 @@ class Check_modules(plugins.PluginInterface):
                 name="linux_utilities_modules_module_display_plugin",
                 component=linux_utilities_modules.ModuleDisplayPlugin,
                 version=(1, 0, 0),
+            ),
+            requirements.BooleanRequirement(
+                name="dump",
+                description="Extract listed modules",
+                default=False,
+                optional=True,
             ),
         ]
 
--- a/volatility3/framework/plugins/linux/check_modules.py
+++ b/volatility3/framework/plugins/linux/check_modules.py
@@ -46,23 +46,12 @@ class Check_modules(plugins.PluginInterface):
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
         return [
-            requirements.ModuleRequirement(
-                name="kernel",
-                description="Linux kernel",
-                architectures=["Intel32", "Intel64"],
-            ),
             requirements.VersionRequirement(
                 name="linux_utilities_modules_module_display_plugin",
                 component=linux_utilities_modules.ModuleDisplayPlugin,
                 version=(1, 0, 0),
             ),
-            requirements.BooleanRequirement(
-                name="dump",
-                description="Extract listed modules",
-                default=False,
-                optional=True,
-            ),
-        ]
+        ] + linux_utilities_modules.ModuleDisplayPlugin.get_requirements()
 
     @classmethod
     @deprecation.deprecated_method(
--- a/volatility3/framework/plugins/linux/hidden_modules.py
+++ b/volatility3/framework/plugins/linux/hidden_modules.py
@@ -19,7 +19,7 @@ class Hidden_modules(plugins.PluginInterface):
     """Carves memory to find hidden kernel modules"""
 
     _required_framework_version = (2, 10, 0)
-    _version = (3, 0, 0)
+    _version = (3, 0, 1)
 
     @classmethod
     def get_hidden_modules(
@@ -71,6 +71,12 @@ class Hidden_modules(plugins.PluginInterface):
                 name="linux_utilities_modules_module_display_plugin",
                 component=linux_utilities_modules.ModuleDisplayPlugin,
                 version=(1, 0, 0),
+            ),
+            requirements.BooleanRequirement(
+                name="dump",
+                description="Extract listed modules",
+                default=False,
+                optional=True,
             ),
         ]
 
--- a/volatility3/framework/plugins/linux/hidden_modules.py
+++ b/volatility3/framework/plugins/linux/hidden_modules.py
@@ -7,7 +7,6 @@ from volatility3.framework.symbols.linux.utilities import (
     modules as linux_utilities_modules,
 )
 from volatility3.framework import interfaces, exceptions, deprecation
-from volatility3.framework.constants import architectures
 from volatility3.framework.configuration import requirements
 from volatility3.framework.symbols.linux import extensions
 from volatility3.framework.interfaces import plugins
@@ -62,23 +61,12 @@ class Hidden_modules(plugins.PluginInterface):
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
         return [
-            requirements.ModuleRequirement(
-                name="kernel",
-                description="Linux kernel",
-                architectures=architectures.LINUX_ARCHS,
-            ),
             requirements.VersionRequirement(
                 name="linux_utilities_modules_module_display_plugin",
                 component=linux_utilities_modules.ModuleDisplayPlugin,
                 version=(1, 0, 0),
             ),
-            requirements.BooleanRequirement(
-                name="dump",
-                description="Extract listed modules",
-                default=False,
-                optional=True,
-            ),
-        ]
+        ] + linux_utilities_modules.ModuleDisplayPlugin.get_requirements()
 
     @staticmethod
     @deprecation.deprecated_method(
--- a/volatility3/framework/plugins/linux/module_extract.py
+++ b/volatility3/framework/plugins/linux/module_extract.py
@@ -18,7 +18,7 @@ class ModuleExtract(interfaces.plugins.PluginInterface):
     """Recreates an ELF file from a specific address in the kernel"""
 
     _version = (1, 0, 0)
-    _required_framework_version = (2, 0, 0)
+    _required_framework_version = (2, 25, 0)
 
     framework.require_interface_version(*_required_framework_version)
 
--- a/volatility3/framework/plugins/linux/module_extract.py
+++ b/volatility3/framework/plugins/linux/module_extract.py
@@ -31,7 +31,7 @@ class ModuleExtract(interfaces.plugins.PluginInterface):
             ),
             requirements.IntRequirement(
                 name="base",
-                description="Base address to reconstruct an ELF file",
+                description="Base virtual address to reconstruct an ELF file",
                 optional=False,
             ),
         ]
--- a/volatility3/framework/plugins/linux/module_extract.py
+++ b/volatility3/framework/plugins/linux/module_extract.py
@@ -1,0 +1,85 @@
+# This file is Copyright 2025 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+import logging
+from typing import List
+
+import volatility3.framework.symbols.linux.utilities.module_extract as linux_utilities_module_extract
+from volatility3.framework import interfaces, renderers
+from volatility3.framework.configuration import requirements
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.objects import utility
+
+vollog = logging.getLogger(__name__)
+
+
+class ModuleExtract(interfaces.plugins.PluginInterface):
+    """Recreates an ELF file from a specific address in the kernel"""
+
+    _required_framework_version = (2, 0, 0)
+
+    _version = (1, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        # Since we're calling the plugin, make sure we have the plugin's requirements
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Windows kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.IntRequirement(
+                name="base",
+                description="Base address to reconstruct an ELF file",
+                optional=False,
+            ),
+        ]
+
+    def _generator(self):
+        kernel = self.context.modules[self.config["kernel"]]
+
+        base_address = self.config["base"]
+
+        kernel_layer = self.context.layers[kernel.layer_name]
+
+        if not kernel_layer.is_valid(base_address):
+            vollog.error(
+                f"Given base address ({base_address:#x}) is not valid in the kernel address space. Unable to extract file."
+            )
+            return
+
+        module = kernel.object(object_type="module", offset=base_address, absolute=True)
+
+        elf_data = linux_utilities_module_extract.ModuleExtract.extract_module(
+            self.context, self.config["kernel"], module
+        )
+        if not elf_data:
+            vollog.error(
+                f"Unable to reconstruct the ELF for module struct at {base_address:#x}"
+            )
+            return
+
+        module_name = utility.array_to_string(module.name)
+        file_name = self.open.sanitize_filename(
+            f"kernel_module.{module_name}.{base_address:#x}.elf"
+        )
+
+        with self.open(file_name) as file_handle:
+            file_handle.write(elf_data)
+
+        yield 0, (
+            format_hints.Hex(base_address),
+            len(elf_data),
+            file_handle.preferred_filename,
+        )
+
+    def run(self):
+        return renderers.TreeGrid(
+            [
+                ("Base", format_hints.Hex),
+                ("File Size", int),
+                ("File output", str),
+            ],
+            self._generator(),
+        )
--- a/volatility3/framework/plugins/linux/module_extract.py
+++ b/volatility3/framework/plugins/linux/module_extract.py
@@ -4,6 +4,7 @@
 import logging
 from typing import List
 
+from volatility3 import framework
 import volatility3.framework.symbols.linux.utilities.module_extract as linux_utilities_module_extract
 from volatility3.framework import interfaces, renderers
 from volatility3.framework.configuration import requirements
@@ -16,9 +17,10 @@ vollog = logging.getLogger(__name__)
 class ModuleExtract(interfaces.plugins.PluginInterface):
     """Recreates an ELF file from a specific address in the kernel"""
 
+    _version = (1, 0, 0)
     _required_framework_version = (2, 0, 0)
 
-    _version = (1, 0, 0)
+    framework.require_interface_version(*_required_framework_version)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
--- a/volatility3/framework/symbols/linux/__init__.py
+++ b/volatility3/framework/symbols/linux/__init__.py
@@ -52,6 +52,8 @@ class LinuxKernelIntermedSymbols(intermed.IntermediateSymbolTable):
         self.set_type_class("idr", extensions.IDR)
         self.set_type_class("address_space", extensions.address_space)
         self.set_type_class("page", extensions.page)
+        self.set_type_class("module_sect_attr", extensions.module_sect_attr)
+
         # Might not exist in the current symbols
         self.optional_set_type_class("module", extensions.module)
         self.optional_set_type_class("bpf_prog", extensions.bpf_prog)
@@ -79,6 +81,7 @@ class LinuxKernelIntermedSymbols(intermed.IntermediateSymbolTable):
         self.set_type_class("sock", extensions.network.sock)
         self.set_type_class("inet_sock", extensions.network.inet_sock)
         self.set_type_class("unix_sock", extensions.network.unix_sock)
+
         # Might not exist in older kernels or the current symbols
         self.optional_set_type_class("netlink_sock", extensions.network.netlink_sock)
         self.optional_set_type_class("vsock_sock", extensions.network.vsock_sock)
--- a/volatility3/framework/symbols/linux/utilities/module_extract.py
+++ b/volatility3/framework/symbols/linux/utilities/module_extract.py
@@ -463,7 +463,7 @@ class ModuleExtract(interfaces.configuration.VersionableInterface):
             )
             e_machine_int = 0x3E  # EM_X86_64
             e_ehsize_int = 64
-            e_shentsize_int = 52
+            e_shentsize_int = 64
             header_size = 64
 
         e_type = struct.pack("<H", 1)  # relocateble
--- a/volatility3/framework/symbols/linux/utilities/module_extract.py
+++ b/volatility3/framework/symbols/linux/utilities/module_extract.py
@@ -28,7 +28,7 @@ vollog = logging.getLogger(__name__)
 # that can be analyzed with static analysis tools
 # First, the .strtab points somewhere random and is kept off the module structure, not with the other sections
 # Second, all of the symbols (.symtab) have mangled members that we must patch for anything to make sense
-# Third, the section name string stable (.shstrtab) is not an allocated section, meaning its not in memory
+# Third, the section name string table (.shstrtab) is not an allocated section, meaning its not in memory
 # Not having the .shstrtab makes analysis impossible-to-difficult for static analysis tools. To work around this,
 # we create the .shstrtab based on the sections in memory and then glue it in as the final section
 
@@ -185,7 +185,7 @@ class ModuleExtract(interfaces.configuration.VersionableInterface):
         cls,
         context: interfaces.context.ContextInterface,
         vmlinux_name: str,
-        original_sections,
+        original_sections: Dict[int, str],
         section_sizes: Dict[int, int],
         sym_type_name: str,
         st_fmt: str,
@@ -347,7 +347,7 @@ class ModuleExtract(interfaces.configuration.VersionableInterface):
         # All others can be read with padding
 
         # get the addresses in sorted order, can index into `original_sections` for names
-        sorted_addresses = sorted(original_sections)
+        sorted_addresses = sorted(original_sections.keys())
 
         # We need to track where .symtab is for symbol name offsets
         symtab_address = None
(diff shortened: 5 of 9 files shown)


atcuno:
@ikelos I had to update hidden_modules with the dump option to make the tests happy. Still will do the bug fix as its own PR after.

volatility3/framework/plugins/linux/check_modules.py
```diff
@@ -56,6 +56,12 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                 component=linux_utilities_modules.ModuleDisplayPlugin,
                 version=(1, 0, 0),
             ),
+            requirements.BooleanRequirement(
```
ikelos: This value doesn't appear to be referenced anywhere in the plugin?  How is it used?

volatility3/framework/plugins/linux/hidden_modules.py
```diff
@@ -72,6 +72,12 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                 component=linux_utilities_modules.ModuleDisplayPlugin,
                 version=(1, 0, 0),
             ),
+            requirements.BooleanRequirement(
```
ikelos: Similarly, the dump parameter isn't used anywhere?  The config isn't passed to another plugin (which would be a problem in itself) so I don't see how this requirement is ever actually accessed?

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -52,6 +52,8 @@ def __init__(self, *args, **kwargs) -> None:
         self.set_type_class("idr", extensions.IDR)
         self.set_type_class("address_space", extensions.address_space)
         self.set_type_class("page", extensions.page)
+        self.set_type_class("module_sect_attr", extensions.module_sect_attr)
```
ikelos, line 55: This isn't optional, so just to check, this is present in every kernel from the beginning of time to the very latest one?

volatility3/framework/plugins/linux/module_extract.py
```diff
@@ -0,0 +1,85 @@
...
+        # Since we're calling the plugin, make sure we have the plugin's requirements
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Windows kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.IntRequirement(
+                name="base",
+                description="Base address to reconstruct an ELF file",
```
ikelos: Is it worth stating that this must be a virtual address?  I worry someone will try a physical address for some reason?

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+vollog = logging.getLogger(__name__)
+
+# This module is responsbile for producing an ELF file of a kernel module (LKM) loaded in memory
+# This extraction task is quite complicated as the Linux kernel discards the ELF header at load time
+# Due to this, to support static analysis, we must create an ELF header and proper file based on the sections
+# There are also several other significant complications that we must deal with when trying to extract an LKM
+# that can be analyzed with static analysis tools
+# First, the .strtab points somewhere random and is kept off the module structure, not with the other sections
+# Second, all of the symbols (.symtab) have mangled members that we must patch for anything to make sense
+# Third, the section name string stable (.shstrtab) is not an allocated section, meaning its not in memory
```
ikelos: stable or table?

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+# we create the .shstrtab based on the sections in memory and then glue it in as the final section
+
+# ModuleExtract.extract_module is the entry point and only visible method for plugins
+
+
+class ModuleExtract(interfaces.configuration.VersionableInterface):
+    """Extracts Linux kernel module structures into an analyzable ELF file"""
+
+    _version = (1, 0, 0)
+    _required_framework_version = (2, 0, 0)
```
ikelos: Again, we're quite far beyond `(2,0,0)` are you absolutely certain this could work with a version that old?  Since this is part of this release, it should probably use this release's `required_framework_version`...

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+
+    @classmethod
+    def _find_section(
+        cls, section_lookups: List[Tuple[str, int, int, int]], sym_address: int
+    ) -> Optional[Tuple[str, int, int, int]]:
+        """
+        Finds the section containing `sym_address`
+        """
+        for name, index, address, size in section_lookups:
+            if address <= sym_address < address + size:
```
ikelos, line 94: I assume the sections can't overlap, or that if they can, the first section will always be the correct one to return?

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+        cls, sym: interfaces.objects.ObjectInterface, sym_address: int, sect_name: str
+    ) -> bytes:
+        """
+        This is a helper function called from `_fix_sym_table`
+
+        Calculates the `st_info` value for the given symbol
+
+        Spec: https://refspecs.linuxbase.org/elf/gabi4+/ch4.symtab.html
+        """
+        if sym.st_name > 0:
```
ikelos, line 110: Can it ever be negative?  Otherwise `!= 0` would be better to make it clear it's a pointer...

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+            # Global symbol
+            bind = linux_constants.STB_GLOBAL
+
+            if sym_address == 0:
+                sect_type = linux_constants.STT_NOTYPE
+            elif sect_name:
+                # rela = relocations
+                if sect_name.find(".text") != -1 and sect_name.find(".rela") == -1:
```
ikelos, line 118: Should these be `startswith` rather than `find`?

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+            st_shndx = struct.pack("<H", sym.st_shndx)
+
+        return sect_name, st_value, st_info, st_shndx
+
+    @classmethod
+    def _fix_sym_table(
+        cls,
+        context: interfaces.context.ContextInterface,
+        vmlinux_name: str,
+        original_sections,
```
ikelos: Typing info for this?

ikelos: Looks like `dict[str, tuple[int, str]]`?

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+                st_fmt, sym, section_lookups
+            )
+
+            # these aren't mangled during loading
+            st_name = struct.pack("<I", sym.st_name)
+            st_other = struct.pack("B", sym.st_other)
+            st_size = struct.pack(st_fmt, sym.st_size)
+
+            # The order as in the ELF specification
+            sym_data = st_name + st_info + st_other + st_shndx + st_value + st_size
```
ikelos: Any reason to pack these one at a time and then concat, rather than doing a bigger pack statement?

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+            st_fmt = "<I"
+
+        # At this point, we have the sections starting addresses and names,
+        # but the kernel does not track the size
+        # To recover the size, we sort by address and then use the next section as the boundary to calculate size
+        # .symtab (the symbol table) and .strtab (the strings table) require special handling.
+        # All others can be read with padding
+
+        # get the addresses in sorted order, can index into `original_sections` for names
+        sorted_addresses = sorted(original_sections)
```
ikelos: This works, but it's a little confusing that it converts it to a list and then sorts the list.  Could we add `.keys()` to make it more explicit what's going on please?

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+        # original_sections does not have this information for reasons explained above
+        section_sizes: Dict[int, int] = {}
+
+        for index, address in enumerate(sorted_addresses):
+            sect_name = original_sections[address]
+
+            # Read out the string table. The full size is not kept, so we give each symbol's string up to 256 bytes
+            if sect_name == ".strtab":
+                # Read out symbol strings, giving up to 256 bytes per symbol
+                data = kernel_layer.read(
```
ikelos, line 375: Would this not work better with the string handling code we did earlier?  Just not keen on including layer read functions unless the parser can't handle it itself...

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+    def _make_elf_header_32(
+        cls, sect_hdr_offset: int, num_sections: int
+    ) -> Optional[bytes]:
+        """
+        Creates a 32 bit ELF header for the file based on recovered values
+        Called last as it needs information computed from the sections
+
+        Spec: https://refspecs.linuxfoundation.org/elf/gabi4+/ch4.eheader.html
+        """
+        e_ident = b"\x7f\x45\x4c\x46\x01\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00"
```
ikelos: I'd kinda prefer data like this were all in a constants file?  It's kinda ok here, but... it really, should be off in constants so that data's separate from code...

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+    def _make_elf_header_64(
+        cls, sect_hdr_offset: int, num_sections: int
+    ) -> Optional[bytes]:
+        """
+        Creates a 64 bit ELF header for the file based on recovered values
+        Called last as it needs information computed from the sections
+
+        Spec: https://refspecs.linuxfoundation.org/elf/gabi4+/ch4.eheader.html
+        """
+        e_ident = b"\x7f\x45\x4c\x46\x02\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00"
```
ikelos: Some of these would then (presumably) be shared with 32 bit if they were in a constants file?

ikelos: and these two could probably become a single method parameterized with a boolean to say 32/64?

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+        return header
+
+    @classmethod
+    def _calc_sect_type(cls, section_name: str) -> Optional[int]:
+        """
+        This function makes a best effort to map common section names
+        to their attributes
+        """
+        known_sections = {
+            ".note.gnu.build-id": linux_constants.SHT_NOTE,
```
ikelos, line 521: Yeah, so... we do use constants already...  5;P

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+        """
+        # looking for RELA sections
+        if name.find(".rela.") != -1:
+            return symtab_index
+
+        # per spec: "The section header index of the associated string table."
+        elif sect_type == linux_constants.SHT_SYMTAB:
+            return strtab_index
+
+        return 0
```
ikelos, line 597: Can 0 ever be mistaken for a legitimate value?  I just worry that it might be confused with a legitimate `strtab_index` or `symtab_index`?

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+            "_ftrace_events",
+            ".gnu.linkonce.this_module",
+        ]:
+            flags = flags | linux_constants.SHF_WRITE
+
+        return flags
+
+    @classmethod
+    def _calc_link(
+        cls, name: str, strtab_index: int, symtab_index: int, sect_type: int
```
ikelos, line 579: Not a big deal, but somehow feels like the distinguisher (`sect_type`) should come after name?

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+        elif sect_type == linux_constants.SHT_SYMTAB:
+            if bits == 32:
+                return 16
+            else:
+                return 24
+
+        return 0
+
+    @classmethod
+    def _make_section_header_32(
```
ikelos: These also both feel like they could be one function but parameterized?  I haven't look at how similar the structures/layouts are, but if it's just tweaking the sizes of some components, it feels like the code could be condensed to just choose on those aspects?  Not a hard and fast suggestion, just thought I'd check what we thought (which one makes the code easier to read/understand)?

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,891 @@
...
+        # Gather sections
+        updated_sections, strtab_index, symtab_index = cls._parse_sections(
+            context, vmlinux_name, module
+        )
+
+        kernel = context.modules[vmlinux_name]
+
+        # Figure out header sizes
+        if symbols.symbol_table_is_64bit(context, kernel.symbol_table_name):
+            make_elf_header = cls._make_elf_header_64
```
ikelos: This is nice, still feels like it could be a single param handed to a single function though.   Happy to go with whatever you decide, just want a written rationale for why one way over the other...

ikelos (commented):
This was a lot to review.  I think I've covered it all.  Most of the comments are just suggestions (particularly the 32/64 bit stuff) just want to get it discussed and a rationale for the choice we go for...

volatility3/framework/plugins/linux/check_modules.py
atcuno: This is used now by the ModuleDisplayInterface. Since requirement's are recursive through parent classes, each plugin must declare the option itself.

volatility3/framework/plugins/linux/hidden_modules.py
atcuno: See previous comment.

volatility3/framework/symbols/linux/__init__.py
atcuno: Yes it is always there.

volatility3/framework/plugins/linux/module_extract.py
atcuno: I will add this to be consistent. Good call.

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: Nice catch, should be table.

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: nice catch again, will update

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: yes, the first one will be correct per spec

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: No, you can have sections like .init.text, .end.text, text.rela, etc.

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: Nice catch. I was trying to make sure everything was typed.

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: It is an unsigned int for a relative offset. 0 = local symbol, > 0 = global

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: Yes as they come from different function calls and calculations, it would be a really ugly pack statement to try and fit all that into one.

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: good catch. I used in keys() for that reason in other plugins.

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: We can't use that as this is an array of strings (of unknown sizes) all jammed together in memory so like:
```
string1<null byte>string2<null byte>....
```

So we would jsut end up returning the first string. We also don't want any interpretation of the layout as the ELF format leaves that to loader.

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: hmm you would end up with the constants files having all these header strings and then the plugin just pulling them in. We also want this to be the only code for people building ELF files from scratch.

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: I normally like condensed code but it would be pretty ugly I think to split this. I could try though

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: Good catch. Will move although I am not sure where else its useful except to map sections to their section type, but its not a good pattern to encourage.

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: 0 is the value we want. It defines the size for sections we want to make sure are linked right (relocations and symtab). From the spec:

"""
sh_entsize
Some sections hold a table of fixed-size entries, such as a symbol table. For such a section, this member gives the size in bytes of each entry. The member contains 0 if the section does not hold a table of fixed-size entries
"""

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: ah I was trying to keep things consistent from other calls

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: I really think its easier to understand as separate to see the the specific fields in order. It is also already a pain to make sure we match the spec exactly.

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: This way so the specific 32/64 headers are made correctly and in line with their spec in a direct flow

volatility3/framework/plugins/linux/check_modules.py
ikelos: I'm not sure I understand that?  Could you point me at the line reference where it's required, or show a backtrace when it's not present?  Given it seems to have no impact on the function of the plugin, I can't see how it would be a requirement (unless something's doing inheritance somewhere, and then I'd want to investigate that and how it's working)...

atcuno: Ok hopefully this link works, its to where I added the ability for linux kernel module listing plugins to dump out LKMs with the new extraction module:

https://github.com/volatilityfoundation/volatility3/pull/1740/files#diff-f63002f66828a7b05783742096b71bae248320bd7d0b24a32d2b39ddb699d10dR979

That is inside the interface we merged recently that all the module listing plugins reference and set their `run` and `generator` from, so it is the place to add the dumping feature. When I added the `dump` option in that interface class, the listing plugins backtraced saying `self.config` didn't have a `dump` value. So adding in the classes themselves seemed reasonable and worked since they are the ones actually inheriting from PluginInterface.

atcuno: Ok it looks like the link worked correctly (to the specific part of the diff)

volatility3/framework/symbols/linux/utilities/module_extract.py
ikelos: I mean, the format string, might be a little ugly, but otherwise it would just be `struct.pack(f"<IB{st_fmt}", sym.st_name, sym.st_other, sym.st_size)`?  And you could change the `_get_fixed_sym_fields` to return them as ints rather than in their packed form?

Happy to leave it as is, that string on concatentation just looks excessive to me...

volatility3/framework/symbols/linux/utilities/module_extract.py
ikelos: Well, I'd be ok with the constants living where constants live and then being pulled in?  You could even put them in their own module (like `constants.linux.elf` or something?  It feels like we reuse them in this very file, and I reckon there'll be other places that want to read ELF files from elsewhere in the code, even if they don't create them?  So what's the downside of moving them to constants?

volatility3/framework/symbols/linux/utilities/module_extract.py
ikelos: It wouldn't be splitting them up, it'd be a single function to do them both, but with a parameter that clearly shows which part of the structure changes based on 32 vs 64?  What I haven't checked is if there's extra fields, or anything weird that could make it harder to understand...

volatility3/framework/symbols/linux/utilities/module_extract.py
ikelos: No, no, this is fine, it was more in relation to the fact that we using `constants` here, so we should use them for the ELF data too!  This is fine, it was just an example of us already using constants!

volatility3/framework/symbols/linux/utilities/module_extract.py
ikelos: Do other calls leave the key deciding variable to the end (after the two choices)?  It just feels to me like you'd put in the deciding variable and *then* the two options, dunno why, as I say, not a big deal, just struck me as something so I leave it up to you what to do with it.  If you think it fits better with the other calls then keep it this way, if you think it fits better with *the old code* then that's old code and not something we need to be bound to.  Final call is yours, just lemme know which you decide so we can mark it as resolved please.  5:)

volatility3/framework/symbols/linux/utilities/module_extract.py
ikelos: Ok, your call, my thinking was, since they look so similar, it would help people easily find the main structure and see exactly which bits differ based on size, rather than scrolling between the two and trying to compare them to work out what's the same/different if they needed it.  Again, your call, just lemme know which you wanna go with please...

volatility3/framework/symbols/linux/utilities/module_extract.py
ikelos: I don't see how it wouldn't be a direct flow if they both went into the same function and then the differences were decided within that one function?  Feels like it would clarify for people looking into it how the two differ and possibly why.  I'll leave it to you, but it just seems to increase the code and spread it out, rather than having it very clearly share all but the parts that differ (which would then be clearly visible by `x if 32bit else y` or similar...

volatility3/framework/plugins/linux/check_modules.py
ikelos: Ok, I see, it's because we pull the run/generator out of the other plugin.  I'd prefer that we took the `get_requirements` from the base plugin too then and just add in the requirement on the `ModuleDisplayInterface` (and any specific requirements for the implementation)?  That way if new config options are needed, they're added in one place and config options are only defined when/where they're needed...

atcuno: This worked perfect and looks really clean

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: I would like to keep this as-is

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: I want to keep this way.

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: I like the concanation this way as its a huge pain to keep the spec and this keeps it really obvious when reading the code

volatility3/framework/symbols/linux/utilities/module_extract.py
atcuno: ok I combined the making of section and elf header

atcuno:
ok @ikelos I believe I addressed all hte feedback, including condensing the functions. Also re-tested with readelf and IDA for 32 and 64 bit extraction.

volatility3/framework/symbols/linux/utilities/module_extract.py
ikelos: Fair enough, just feels a little verbose, but I can cope...  5;)

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -0,0 +1,809 @@
...
+        """
+        Creates a `bits` bit ELF header for the file based on recovered values
+        Called last as it needs information computed from the sections
+
+        Spec: https://refspecs.linuxfoundation.org/elf/gabi4+/ch4.eheader.html
+        """
+        if bits == 32:
+            fmt = "<I"
+            e_ident = (
+                b"\x7f\x45\x4c\x46\x01\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00"
```
ikelos, line 456: Pity you can't do binary format strings, but I think this makes it clearer what's change and what's the same between the two now, thanks  5:)

volatility3/framework/symbols/linux/utilities/module_extract.py
ikelos: Think this is the only outstanding commit?

volatility3/framework/plugins/linux/module_extract.py
```diff
@@ -0,0 +1,85 @@
...
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.objects import utility
+
+vollog = logging.getLogger(__name__)
+
+
+class ModuleExtract(interfaces.plugins.PluginInterface):
+    """Recreates an ELF file from a specific address in the kernel"""
+
+    _required_framework_version = (2, 0, 0)
```
ikelos: And possibly this?  I've been a bit lax in updating the API file to say what was added when...  5:S

ikelos (commented):
Nearly done, just got to confirm whether those two `_required_framework_version`s are correct, or just bump them to the latest version...

volatility3/framework/symbols/linux/utilities/module_extract.py
ikelos: Well, it's slightly less in your face now they've all been consolidated a bit.

volatility3/framework/plugins/linux/module_extract.py
atcuno: I added ` framework.require_interface_version(*_required_framework_version)` in the utilities module_extract but not the plugins by accident. I can add it to the plugins, but not sure what else you wanted in extract. Is `2, 0, 0` the wrong value?

atcuno: I did a push for the plugin `require_interface_version` call

ikelos: Well, that's the very, very first version of the framework and I haven't checked your code to see if you use any features that were only introduced recently, so the safest thing to do would be to always rely on the latest framework version?  But if you use things like `tasks.get_thread` that apparently only appeared in `(2,1,0)`...

ikelos: I'd just set it to `(2,25,0)` unless you've tested it with older versions...

ikelos: (unfortunately, we stopped documenting it accurately at 2.5.0...  5:S)
https://github.com/volatilityfoundation/volatility3/blob/develop/API_CHANGES.md

atcuno: Ok I will bump to that. We should enforce the API_CHANGES updates starting now since this will be post-parity.

ikelos: Yeah, I'll see if I can backfill it over the weekend...

atcuno: Ok just pushed with both places at 25

PR #1743, opened 2025-03-26
Testing: Verify `VersionRequirement`s

dgmcdona:
This adds a script and GitHub action to the `test` directory that dynamically imports all modules in `volatility3`, searches for usages of `VersionableInterface` objects within classes that inherit from `ConfigurableInterface` but don't enumerate the used component as a requirement in `get_requirements()`, and returns -1 if any violations are found. This a modified version of the script that was used to enumerate all of the undeclared requirements that were fixed up in #1738.

This requires the addition of the `tree-sitter` and `tree-sitter-python` to the `test` dependencies.

It works by walking the packages with `pkgutil.walk_packages`, and importing any modules where the `__file__` attribute is not `None`. For each imported module, it finds classes that are subclasses of `interfaces.configuration.ConfigurableInterface` and retrieves the AST for those class definitions via tree-sitter.

Within each class definition, it looks for identifiers or attributes matching one of three forms:
- `pslist`
- `pslist.PsList`
- `pslist.PsList.list_processes`

For each discovered item, it splits the raw text of that node on the `.`, and looks it up the first component of the resulting list in the current module namespace. If it finds that the item is an imported module, it calls `getattr` on the imported with the remaining components until it finds a `VersionableInterface` subclass. If it finds that the _first_ component was a `VersionableInterface`, it stops there.

If a `VersionableInterface` was discovered, it checks it against the `VersionRequirements` enumerated in the containing class' `get_requirements()` classmethod. If it's missing, it reports the module, class, and line number.

For now, the GitHub Action simply prints the offending lines and exits non-zero, but this could be adapted to add comments to pull requests if we decide it's worth the effort.

It is possible for a false positive to occur if someone shadows an imported module or class name within a class definition, but this is probably something we want to avoid anyway.

This PR also fixes all remaining violations that were discovered when this script was reworked to be more comprehensive.

### DIFF
--- a/test/check_configurable_requirements.py
+++ b/test/check_configurable_requirements.py
@@ -1,0 +1,300 @@
+import importlib
+import inspect
+import pkgutil
+import sys
+import traceback
+import types
+from textwrap import dedent
+from typing import Dict, Iterator, List, NamedTuple, Optional, Tuple, Type
+
+from tree_sitter import Language, Node, Parser
+from tree_sitter_python import language as python_language
+
+from volatility3.framework import configuration, interfaces
+
+
+class UnrequiredVersionableUsage(NamedTuple):
+    versionable_item_class: str
+    """
+    The name of the VersionableInterface class
+    """
+
+    consuming_class: str
+    """
+    The name of the class that is using the imported VersionableInterface class
+    """
+
+    methodname: Optional[str]
+    """
+    The name of the invoked method or attribute, if one is used or referenced
+    """
+
+    node: Node
+    """
+    The tree-sitter node encapsulating the used module component.
+    """
+
+    def __str__(self) -> str:
+        return (
+            f"Found usage of {self.versionable_item_class} "
+            f"in class {self.consuming_class} that is not declared "
+            f"in {self.consuming_class}'s `get_requirements()` classmethod"
+        )
+
+
+class RequirementValidator:
+    language = Language(python_language(), "python")
+
+    def __init__(self, plugin_module: types.ModuleType) -> None:
+        if plugin_module.__file__ is None:
+            raise ValueError("Attempting to validate a module without a file")
+
+        self._module = plugin_module
+
+        # See which classes in *this* module are configurable (can have requirements declared)
+        self._configurable_classes = get_configurable_classes(plugin_module)
+
+        # Get a mapping of class names to configurable classes that they declare in their requirements
+        self._versioned_item_mapping = get_versioned_item_mapping(
+            self._configurable_classes
+        )
+
+        # Get a mapping of module name -> versionable classes within the namespace of each module
+        self._imported_mod_classes = get_versionable_import_mapping(
+            get_imported_modules(plugin_module)
+        )
+
+        with open(plugin_module.__file__, "rb") as f:
+            source = f.read()
+
+        self._parser = Parser()
+        self._parser.set_language(self.language)
+        self._tree = self._parser.parse(source)
+
+    def enumerate_unrequired_usages(
+        self,
+        clazz: Type[interfaces.configuration.ConfigurableInterface],
+        class_node: Node,
+    ):
+
+        # This query is designed to look for three different identifier usages:
+        # simple identifiers: PsList
+        # module attrs: pslist.PsList
+        # method calls: pslist.PsList.list_processes
+        obj_query = self.language.query(
+            dedent(
+                """
+                [
+                  (identifier)
+                  (attribute
+                    object: (identifier)
+                    attribute: (identifier))
+                  (attribute
+                    object: (attribute
+                      object: (identifier)
+                      attribute: (identifier))
+                  )
+                ] @ident
+                """
+            )
+        )
+
+        containing_name = class_node.child_by_field_name("name").text.decode("utf-8")
+
+        valid_types = self._versioned_item_mapping[containing_name]
+        for _, match in obj_query.matches(class_node):
+            if "ident" not in match:
+                continue
+
+            # Get the raw text of the match. This could be something like
+            # - PsList
+            # - pslist.PsList
+            # - pslist.PsList.list_processes
+            ident_text = match["ident"].text.decode("utf-8")
+
+            # split the attributes
+            components = ident_text.split(".")
+            try:
+                # See if the first attribute is in the module namespace.
+                item = vars(self._module)[components[0]]
+            except KeyError:
+                # If it's not, it's likely a variable in a smaller scope and we
+                # can ignore it.
+                continue
+
+            # If it's in the module namespace and is a module...
+            if isinstance(item, types.ModuleType):
+                try:
+                    # We try getting attributes from it until we
+                    # find one that is a versionable class
+
+                    # Ideally, we shouldn't have to look further than
+                    # two levels
+                    item = getattr(item, components[1])
+                    if not is_versionable(item):
+                        item = getattr(item, components[2])
+                        if not is_versionable(item):
+                            continue
+
+                except (IndexError, AttributeError):
+                    # we ran out of attributes to check
+                    continue
+
+            elif is_versionable(item):
+                # The versionable thing was at the top level. This
+                # goes against our preferred style, but is possible.
+                pass
+            else:
+                # This isn't something we care about.
+                continue
+
+            if (
+                item in valid_types
+                or item is clazz
+                or inspect.isabstract(item)
+                or item
+                is interfaces.configuration.VersionableInterface  # Avoid checking the interface itself
+            ):
+                continue
+
+            yield UnrequiredVersionableUsage(
+                item,
+                containing_name,
+                components[1] if len(components) > 1 else None,
+                match["ident"],
+            )
+
+    def find_class_nodes(
+        self,
+    ) -> Iterator[Tuple[Type[interfaces.configuration.ConfigurableInterface], Node]]:
+        """
+        Yields an iterator of (classname, node) tuples, where the node is the subtree containing
+        the entire class definition.
+        """
+        class_query = self.language.query("(class_definition) @classdef")
+
+        matches = class_query.captures(self._tree.root_node)
+        for node, _ in matches:
+            classname = node.child_by_field_name("name").text.decode("utf-8")
+            if classname not in self._configurable_classes:
+                continue
+
+            yield self._configurable_classes[classname], node
+
+
+def is_versionable(var):
+    try:
+        return issubclass(var, interfaces.configuration.VersionableInterface)
+    except TypeError:
+        return False
+
+
+def is_configurable(var):
+    try:
+        return issubclass(var, interfaces.configuration.ConfigurableInterface)
+    except TypeError:
+        return False
+
+
+def get_imported_modules(
+    plugin_module: types.ModuleType,
+) -> List[Tuple[str, types.ModuleType]]:
+    return [
+        (name, var)
+        for name, var in vars(plugin_module).items()
+        if isinstance(var, types.ModuleType)
+    ]
+
+
+def get_configurable_classes(
+    plugin_module: types.ModuleType,
+) -> Dict[str, Type[interfaces.configuration.ConfigurableInterface]]:
+    return {
+        name: clazz
+        for name, clazz in vars(plugin_module).items()
+        if is_configurable(clazz)
+    }
+
+
+def get_versioned_item_mapping(
+    configurable_classes: Dict[
+        str, Type[interfaces.configuration.ConfigurableInterface]
+    ]
+) -> Dict[str, List[Type[interfaces.configuration.VersionableInterface]]]:
+    return {
+        name: [
+            req._component
+            for req in clazz.get_requirements()
+            if isinstance(req, configuration.requirements.VersionRequirement)
+        ]
+        for name, clazz in configurable_classes.items()
+    }
+
+
+def get_versionable_import_mapping(
+    imported_modules: List[Tuple[str, types.ModuleType]]
+) -> Dict[str, List[str]]:
+    return {
+        modname: [name for name, var in vars(module).items() if is_versionable(var)]
+        for modname, module in imported_modules
+    }
+
+
+def report_missing_requirements() -> Iterator[Tuple[str, UnrequiredVersionableUsage]]:
+    vol3 = importlib.import_module("volatility3")
+
+    for _, module_name, _ in pkgutil.walk_packages(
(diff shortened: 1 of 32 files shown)


dgmcdona:
@ikelos Some of the scanner classes didn't have `_version` explicitly set, so I set them to `(1, 0, 0)`.

ikelos:
So, we've generally allowed without classmethod (no exported API) or those still under development to have no `_version` (which should then default to `(0,0,0)`).  I'm ok with us taking that away, but it might be useful to get a check in for plugins that define a version of `(0,x,x)` just to see if anything' stays in development for a long time.

As to the global importing, I'm slightly concerned that may pull in plugins that aren't intended to be used directly (either abstract plugins, or something else).  It might be better to use volatility's in-built machinery to ensure consistency on how plugins are imported?  I dunno if that would be too much work, or if it could just be turned into a volshell script to make it simpler to develop?  I'll give the whole thing a review in its current state though...

ikelos:
We have machinery to explicit avoid a PluginInterface being picked up if it's only ever an interface (there's a  non-inheritable class attribute that can be specified called `hidden` that hides a class from making the plugin list...
(see https://github.com/volatilityfoundation/volatility3/blob/f33e9daa48d878696221f4debcbc403b7f5bbada/volatility3/framework/__init__.py#L80)

volatility3/cli/volshell/mac.py
```diff
@@ -25,6 +25,11 @@ def get_requirements(cls):
             requirements.IntRequirement(
                 name="pid", description="Process ID", optional=True
             ),
+            requirements.VersionRequirement(
```
ikelos, line 28: These methods are inheritted, so I'd suggest either we rework them to not inherit, or we deal with inheritance differently (possibly by adding the `super().get_requirements` on the end of this?

volatility3/cli/volshell/linux.py
```diff
@@ -36,6 +36,11 @@ def get_requirements(cls):
             requirements.IntRequirement(
                 name="pid", description="Process ID", optional=True
             ),
+            requirements.VersionRequirement(
```
ikelos, line 39: Same as below

test/check_configurable_requirements.py
```diff
@@ -0,0 +1,300 @@
...
+
+    for _, module_name, _ in pkgutil.walk_packages(
+        vol3.__path__, vol3.__name__ + ".", onerror=lambda _: None
+    ):
+        try:
+            # import the module that we want to check
+            modname = module_name.replace(
+                "volatility3.framework.plugins", "volatility3.plugins"
+            )
+            plugin_module = importlib.import_module(modname)
```
ikelos, line 296: Individually importing them should at least show up assumptions we've made about import order (which is good, not sure why I originally ended that with a question mark, sorry!)  5:)

test/check_configurable_requirements.py
```diff
@@ -0,0 +1,300 @@
...
+
+        except ImportError:
+            continue
```
ikelos, line 300: However, not even logging that an import error occurred seems like we're just letting that data slip away from us...

test/check_configurable_requirements.py
```diff
@@ -0,0 +1,300 @@
...
+        except Exception:
```
ikelos: This also seems a little open ended?  Shouldn't we know what we're expecting to happen, or we let the thing fails?  Capturing all exceptions is generally frowned upon...

test/check_configurable_requirements.py
```diff
@@ -0,0 +1,300 @@
...
+            continue
+
+        if plugin_module.__file__ is None:
```
ikelos, line 310: In what situations might this happen if we just did the import from a filename?  Be useful to add a comment for whatever legitimate reason there is...

test/check_configurable_requirements.py
```diff
@@ -0,0 +1,300 @@
...
+            continue
+
+        try:
+            # construct a validator for the module
+            try:
+                validator = RequirementValidator(plugin_module)
+            except Exception:
```
ikelos: Still a bit broad, but at least it offers a full traceback, so I'll let this one slide...  5;)

volatility3/cli/volshell/windows.py
```diff
@@ -23,6 +23,11 @@ def get_requirements(cls):
             requirements.IntRequirement(
                 name="pid", description="Process ID", optional=True
             ),
+            requirements.VersionRequirement(
+                name="generic_volshell",
+                component=generic.Volshell,
```
ikelos, line 28: Same again

volatility3/framework/automagic/pdbscan.py
```diff
@@ -17,7 +17,7 @@
 from volatility3.framework.configuration import requirements
 from volatility3.framework.layers import intel, scanners
 from volatility3.framework.symbols import native
-from volatility3.framework.symbols.windows.pdbutil import PDBUtility
+from volatility3.framework.symbols.windows import pdbutil
```
ikelos, line 20: Ah good work, thanks, I think I've seen the a couple times and then not been bothered to fix it!  5:P

volatility3/framework/automagic/pdbscan.py
```diff
@@ -50,6 +50,21 @@ class KernelPDBScanner(interfaces.automagic.AutomagicInterface):
...
 
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.VersionRequirement(
+                name="pdb_utility",
+                component=pdbutil.PDBUtility,
+                version=(1, 0, 1),
+            ),
+            requirements.VersionRequirement(
```
ikelos, line 61: I hadn't realized we even versioned the scanners but... makes sense?  5:)

volatility3/framework/plugins/vmscan.py
```diff
@@ -69,6 +71,11 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
             requirements.TranslationLayerRequirement(
                 name="primary", description="Physical base memory layer"
             ),
+            requirements.VersionRequirement(
+                name="page_start_scanner",
+                component=PageStartScanner,
```
ikelos, line 76: This is probably fine (since they're defined in the same file) but I kinda like having it explicit, so glad to have this added!

ikelos (commented):
Wow, this looks really comprehensive!  Hopefully people don't find it to onerous to version all the scanners they use, but it shouldn't be a big extra ask.  A few points around the method to import being different to the means the framework uses to do the importing.  Might be fine, just want to see if there's a way to avoid discrepancies...

ikelos:
Also (doesn't have to be in this PR) any chance of spotting whenever a `_required_framework_version` is added, and make sure it's at flagging it as potentially not having been tested (might be cases where people want to allow an old framework, but so far I can't think of any, so may be elevated to a verification fail).  Just noticed a lot of people on autopilot saying that everything works with framework `(2,0,0)` essentially making the version bumps pointless (and allowing this to break if they depend on stuff that was added later)...

volatility3/framework/automagic/pdbscan.py
dgmcdona: Happy to do it, I think I'll make another PR shortly that finds and fixes the rest of these.

volatility3/cli/volshell/mac.py
dgmcdona: Done 👍

test/check_configurable_requirements.py
dgmcdona: I'll fix these general exception handlers up and add some more useful output messages.

test/check_configurable_requirements.py
dgmcdona: Hmm not sure, it's one of those things that I added during development, but now some other change I made seems to have rendered it unnecessary - removing it doesn't seem to affect anything. I'll leave it in place and add a log message just to be safe.

volatility3/cli/volshell/linux.py
ikelos: Now we add `get_requirements` we can remove the version requirements from the class (since they get added by the below addition now)...

ikelos (commented):
Thanks, looks better but the point of adding the super requirements was so that the plugins wouldn't have requirements they don't explicitly need (because everything's now inheritted)...

volatility3/cli/volshell/linux.py
ikelos: I think that's also why it's failing the tests, because there's now two kernel module requirements (but only one gets fulfilled?).  Now we inherit the requirements, these lists should only be additional requirements needed by the code that's explicitly in these plugins and not already present in the base plugin.

ikelos:
Huh, I thought we'd ignored `G004` in the `pyproject.toml`?  The rationale is that the interpretter will stick them together even if the log never gets made (because the level is too low) so it's a waste of time, but honestly I'm not sure we care that much about optimization.  As to the other ruff errors, they should probably get resolved...  5:S

dgmcdona:
Sorry for the big rewrite, I realized after thinking about things that I was blinded by my own close familiarity with tree-sitter and that the Python `ast` module with a visitor pattern was probably a better long-term approach for this, plus it requires no additional dependencies. I'm tidying up the remaining items now

dgmcdona:
The only remaining obstacle that I'm pretty certain can't be resolved is checking configurable classes inside of factory functions. We'll just have to keep a close eye on those during code review and not assume that it's possible to check every configurable class.

volatility3/cli/volshell/linux.py
dgmcdona: Are we sure we want to remove that? We added the `TimeLinerInterface` as a requirement to lots of the plugin classes even though that's inherited, maybe because it lives in a different file?

dgmcdona: So I think I see the distinction... `TimeLinerInterface` isn't configurable, maybe the solution is to ignore checks on inherited configurables so long as the inheriting class calls `super().get_requirements()`.

ikelos: Errr, yeah, I dunno now, I'd need to think about it?  My thought was, if you're doing inheritance, it'd kinda not related/keyed into the requirements, because you have to make sure you don't break stuff *anyway*, but I guess maybe that's not true?  Eh, I guess it's ok, just seems like every instance of inheritance throughout a python program would run into this issue, so the accepted way of dealing with it is just to be careful?

ikelos:
No problem, I'm kinda happy we can do everything we need within python itself (although now I'm gonna have to go read up on the AST stuff and figure out how to use it, because it looks handy!).  5:P

Do we have any instances of Configurable classes constructed within factories or is that more of a hypothetical?

ikelos:
Also, any idea why the xp images are failing?

dgmcdona:
Now that it's calling the `generic.Volshell` class' `get_requirements()` via `super()`, it's missing the `primary` layer for some reason.

dgmcdona:
These classes are configurable and defined in a factory function:
https://github.com/volatilityfoundation/volatility3/blob/49d0404031f738407b592d632cb4463bdaa066a0/volatility3/cli/__init__.py#L763

ikelos:
> Now that it's calling the `generic.Volshell` class' `get_requirements()` via `super()`, it's missing the `primary` layer for some reason.

Oh... yeah, ok.  Generic works just on a layer, the windows/linux ones work on a Module (which includes a layer) so it won't try and fulfill two layers.  I suppose we could split the common requirements out in the Generic one, and then just add those (and have the others do the ones that they need specifically, including Generic)?  Sorry it's turning out to be more convoluted, but I still think it's probably the right way of doing things...

ikelos:
> These classes are configurable and defined in a factory function:
> 
> https://github.com/volatilityfoundation/volatility3/blob/49d0404031f738407b592d632cb4463bdaa066a0/volatility3/cli/__init__.py#L763

Ah, yeah, ok.  I think you're probably right, trying to cope with such a specific case is tricky.  As long as we document it clearly, if we run into it we should be able to determine the reason the test missed it and then fix one-off it...

dgmcdona:
> Also (doesn't have to be in this PR) any chance of spotting whenever a `_required_framework_version` is added, and make sure it's at flagging it as potentially not having been tested (might be cases where people want to allow an old framework, but so far I can't think of any, so may be elevated to a verification fail). Just noticed a lot of people on autopilot saying that everything works with framework `(2,0,0)` essentially making the version bumps pointless (and allowing this to break if they depend on stuff that was added later)...

I'm definitely guilty of this myself, I'll think about ways to do it

volatility3/cli/volshell/linux.py
dgmcdona: I'll defer to your judgement on how to handle this, I think if we want to remove this requirement while still passing the checker, we'll have to see if the component is in the class' `mro()`, or maybe just the defined base classes, since that's what the script is checking anyway?

dgmcdona:
> We have machinery to explicit avoid a PluginInterface being picked up if it's only ever an interface (there's a non-inheritable class attribute that can be specified called `hidden` that hides a class from making the plugin list... (see
> 
> https://github.com/volatilityfoundation/volatility3/blob/f33e9daa48d878696221f4debcbc403b7f5bbada/volatility3/framework/__init__.py#L80
> )

I'm not sure how to integrate this, does it need to be done as part of the `is_versionable` or `is_configurable` functions? Right now we ignore abstract classes with `inspect.isabstract`. It doesn't look like `hide_from_subclasses` is in use anywhere.

ikelos:
So where ever `inspect.isabstract` is used, we should probably also filter out if they have the class attribute `.hidden` set to True?  Is that reasonably easy to achieve?  If not, we can leave it but just add a comment saying it's not covered...

test/check_configurable_requirements.py
```diff
@@ -333,7 +344,7 @@ def perform_review():
 
     if found:
         print(
-            f"Found {found} uses of versionable components not declared in get_requirements()"
+            f"Found {found} coding standards violations"
```
ikelos: I mean, technically accurate, but violation sounds a little bit strong?  Any chance we could tone it down a little please?  Don't want to put people off of contributing!  5:D

dgmcdona:
Ok, I think this should be good now, unless you see anything else on a last look.

ikelos (commented):
Just the wording...  5;)

test/check_configurable_requirements.py
dgmcdona: For sure! Had a feeling that might be too much.

ikelos (approved):
Much better, thanks!  5:D

PR #1747, opened 2025-03-28
Update the various MINOR version bumps to the current version

ikelos:
Really sleepy so I'd appreciate someone looking this over, but I think it contains all the changes that were made between each of the MINOR version bumps of the framework...

### DIFF
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -4,6 +4,100 @@ API Changes
 When an addition to the existing API is made, the minor version is bumped.
 When an API feature or function is removed or changed, the major version is bumped.
 
+2.25.0
+======
+Pointer class now supports `get_raw_value()`.
+`KTIMER` no longer supports `get_raw_dpc()`.
+
+2.24.0
+======
+Support `encoding` parameter for `objects.utility.array_to_string`
+
+2.23.0
+======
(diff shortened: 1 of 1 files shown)


ikelos:
Just hoping to get some eyes on this please, should be very quick, just skim it and see if it all seems reasonable...

API_CHANGES.md
```diff
@@ -4,6 +4,100 @@ API Changes
...
+======
+Addition of `convert_fourcc_code` to `LinuxUtilities` class
+
+2.14.0
+======
+No significant changes (part of the 2.16.0 PR which took time in development)
+
+2.13.0
+======
+Linux `task` objectr extension addition of `getppid`
```
eve-mem: Just a tiny typo on object I think, extra r? "Linux task object extension addition of getppid"

ikelos: Yep, looks like it thanks!
```suggestion
Linux `task` object extension addition of `getppid`
```

atcuno:
I can't think of any major additions I was involved in not covered by this. I also went through some of the recent git history and we look good there too.

ikelos:
It was constructed from the git history, so I guess it's just if we made two big changes at once, but only bumped the version once?  Hopefully it's accurate enough to do the job.  Thanks for reviewing it!

PR #1748, opened 2025-03-28
Volshell: Display types pointer upgrade

### LINKED ISSUE #1705: Volshell `display_type()` issue on `.write` attribute
I encountered an issue when I was trying to display the `file_operations` of an open file descriptor in a Linux process.

The issue is that because `volobject` has a method `write()` and when in the loop in which each type is displayed, the `getattr()` function grabs the method instead of the object's address. After debugging it, I can see that in contrast with `owner`, `llseek`, etc, the value getting passed into the `_display_value()` method is not an `int` (address) but the method `write` of the `volobject`: See Example Output.

**Context**
Volatility Version: 2.23.0 
Operating System: Linux
Python Version: 3.13

**To Reproduce**

1. Open any linux memdump into volshell and isolate any process (`proc = ps()[0]`)
2. `dt(proc.files.dereference().fd_array[0].dereference().f_op.dereference())`

**Expected behavior**
The `file_operations` structure is displayed clearly

**Example output**
```
   0x0 :   owner                    symbol_table_name1!pointer               0
   0x8 :   llseek                   symbol_table_name1!pointer               281473008328944
  0x10 :   read                     symbol_table_name1!pointer               0
  File "<input>", line 13, in <module>
  File "/home/canopus/dev/diss/volatility3/volatility3/cli/volshell/linux.py", line 147, in display_type
    return super().display_type(object, offset)
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/home/canopus/dev/diss/volatility3/volatility3/cli/volshell/generic.py", line 417, in display_type
    self._display_value(getattr(volobject, member)),
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/canopus/dev/diss/volatility3/volatility3/cli/volshell/generic.py", line 437, in _display_value
    return hex(value.vol.offset)
               ^^^^^^^^^
AttributeError: 'function' object has no attribute 'vol'
```

### LINKED ISSUE #1714: volshell dt() throws exception on invalid pointer members
This is a painful bug.. if you dt(), volshell shows members until a paged out one is encountered and then throws a huge backtrace and doesn't show any further members. The broken one just needs to list as broken and then continue through the remaining members.

This should be an easy fix.

eve-mem:
Hello :wave: 

This PR updates volshell display types to follow pointers and display more information, and IMHO makes it easier to use. 

This is a replacement for PR https://github.com/volatilityfoundation/volatility3/pull/1028 as my git-fu is not strong enough to survive a rebase. 

Fixes https://github.com/volatilityfoundation/volatility3/issues/1714 
Fixes https://github.com/volatilityfoundation/volatility3/issues/1705

dt() updates
----
`dt()` will now follow pointers as needed to display information, up to a maximum of `MAX_DEREFERENCE_COUNT` times (currently set to 4).

With the a dt() of a `task_struct` you get a small hint at the changes. Here we see that `stack` and `files` are actually pointers as denoted by the `*` - we can also see what kind of struct it is, e.g. `files` is a `files_struct `. Previously it would have only displayed `pointer` so you wouldn't know what it was.  The value displayed is still the actual pointer data as before. It will show pointers in hex e.g. `0x880001f60000` rather than `149533614276608` as this feels more natural.

The `@ 0x88001d06c800` also shows where in the layer this object is. Previously you'd need to use `.vol.offset` or similar to find this out as an extra step.

```
symbol_table_name1!task_struct (1784 bytes) @ 0x88001d06c800:
    0x0 :   state                           symbol_table_name1!long int                     1
    0x8 :   stack                           *symbol_table_name1!void                        0x880001f60000
   0x10 :   usage                           symbol_table_name1!atomic_t                     0x88001d06c810
   0x14 :   flags                           symbol_table_name1!unsigned int                 4202752
   0x18 :   ptrace                          symbol_table_name1!unsigned int                 0
<snip>
  0x480 :   files                           *symbol_table_name1!files_struct                0x88001cbd59c0
<snip>
```

However things get useful when we would need to follow pointers, e.g. the `files` member of `task_struct`

Here we can see that `.files` is a pointer. It was located at offset `0x88001d06cc80` and points to `0x88001cbd59c0`. At `0x88001cbd59c0` we have the `files_struct` and dt will display the details of it. The results for `files_struct` is indented to show that it is the result of following a pointer. Previously if you did  `dt(task.files)` you would simply be told it was a pointer, you'd need to type `dt(task.files.dereference())` to get the results. I personally struggle to spell dereference correctly most of the time and these changes make my life a little easier!

```
(layer_name) >>> dt(gp(pid=8497).files)
symbol_table_name1!pointer (8 bytes) @ 0x88001d06cc80 -> 0x88001cbd59c0
    symbol_table_name1!files_struct (704 bytes) @ 0x88001cbd59c0:
       0x0 :   count                  symbol_table_name1!atomic_t            0x88001cbd59c0
       0x8 :   fdt                    *symbol_table_name1!fdtable            0x88001cbd59d0
      0x10 :   fdtab                  symbol_table_name1!fdtable             0x88001cbd59d0
      0x80 :   file_lock              symbol_table_name1!spinlock_t          0x88001cbd5a40
      0x84 :   next_fd                symbol_table_name1!int                 1
      0x88 :   close_on_exec_init     symbol_table_name1!embedded_fd_set     0x88001cbd5a48
      0x90 :   open_fds_init          symbol_table_name1!embedded_fd_set     0x88001cbd5a50
```

The changes will follow multiple pointers if needed too. If we look at the `fdt` here we can see that the `fd` member is a double pointer which eventually gets to a `file` as denoted by the `**`.

```
(layer_name) >>> dt(gp(pid=8497).files.fdt)
symbol_table_name1!pointer (8 bytes) @ 0x88001cbd59c8 -> 0x88001cbd59d0
    symbol_table_name1!fdtable (56 bytes) @ 0x88001cbd59d0:
       0x0 :   max_fds           symbol_table_name1!unsigned int         64
       0x8 :   fd                **symbol_table_name1!file               0x88001cbd5a58
      0x10 :   close_on_exec     *symbol_table_name1!__kernel_fd_set     0x88001cbd5a48
      0x18 :   open_fds          *symbol_table_name1!__kernel_fd_set     0x88001cbd5a50
      0x20 :   rcu               symbol_table_name1!rcu_head             0x88001cbd59f0
      0x30 :   next              *symbol_table_name1!fdtable             0x0 (null pointer)
```
Lastly we can see `fd` just as easily by using `dt(task.files.fdt.fd)`. Here we see the first pointer located at `0x88001cbd59d8` points to `0x88001cbd5a58`. The second pointer at `0x88001cbd5a58` then points to `0x88001d145580`. Finally at `0x88001d145580` we have the `file` itself. You still get all the information about the pointers themselves if that is what you're interested in but you quickly get to the struct information which is what I think is the more common use case. 

```
(layer_name) >>> dt(gp(pid=8497).files.fdt.fd)
symbol_table_name1!pointer (8 bytes) @ 0x88001cbd59d8 -> 0x88001cbd5a58
    symbol_table_name1!pointer (8 bytes) @ 0x88001cbd5a58 -> 0x88001d145580
        symbol_table_name1!file (208 bytes) @ 0x88001d145580:
           0x0 :   f_u               symbol_table_name1!unnamed_6733425fe3c7f4c9     0x88001d145580
          0x10 :   f_path            symbol_table_name1!path                         0x88001d145590
          0x20 :   f_op              *symbol_table_name1!file_operations             0xffff8143aa70
          0x28 :   f_lock            symbol_table_name1!spinlock_t                   0x88001d1455a8
          0x2c :   f_sb_list_cpu     symbol_table_name1!int                          0
          0x30 :   f_count           symbol_table_name1!atomic64_t                   0x88001d1455b0
          0x38 :   f_flags           symbol_table_name1!unsigned int                 32768
          0x3c :   f_mode            symbol_table_name1!unsigned int                 29
          0x40 :   f_pos             symbol_table_name1!long long int                0
          0x48 :   f_owner           symbol_table_name1!fown_struct                  0x88001d1455c8
          0x68 :   f_cred            *symbol_table_name1!cred                        0x88001ca33cc0
          0x70 :   f_ra              symbol_table_name1!file_ra_state                0x88001d1455f0
          0x90 :   f_version         symbol_table_name1!long long unsigned int       0
          0x98 :   f_security        *symbol_table_name1!void                        0x0 (null pointer)
          0xa0 :   private_data      *symbol_table_name1!void                        0x0 (null pointer)
          0xa8 :   f_ep_links        symbol_table_name1!list_head                    0x88001d145628
          0xb8 :   f_tfile_llink     symbol_table_name1!list_head                    0x88001d145638
          0xc8 :   f_mapping         *symbol_table_name1!address_space               0x88001f9b0480
```
Fixing https://github.com/volatilityfoundation/volatility3/issues/1705
---
Before the changes, you cannot `dt()` a `file_operations` struct as it includes `write` which is a python function for the vol object.

```
(layer_name) >>> dt(gp(pid=8497).files.fd_array[0].f_op.dereference())
symbol_table_name1!file_operations (208 bytes)
  0x0 :   owner                 symbol_table_name1!pointer     0
  0x8 :   llseek                symbol_table_name1!pointer     281472848377746
 0x10 :   read                  symbol_table_name1!pointer     281472848377728
  File "/home/eve/Documents/volatility3/volatility3/cli/volshell/linux.py", line 147, in display_type
    return super().display_type(object, offset)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/eve/Documents/volatility3/volatility3/cli/volshell/generic.py", line 461, in display_type
    self._display_value(getattr(volobject, member)),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/eve/Documents/volatility3/volatility3/cli/volshell/generic.py", line 481, in _display_value
    return hex(value.vol.offset)
               ^^^^^^^^^
AttributeError: 'function' object has no attribute 'vol'
```

With the change to use `.member()` as suggested by @atcuno (https://github.com/volatilityfoundation/volatility3/issues/1714#issuecomment-2731186920) it now works as expected:

```
(layer_name) >>> dt(gp(pid=8497).files.fd_array[0].f_op)
symbol_table_name1!pointer (8 bytes) @ 0x88001d1455a0 -> 0xffff8143aa70
    symbol_table_name1!file_operations (208 bytes) @ 0xffff8143aa70:
       0x0 :   owner                 *symbol_table_name1!module       0x0 (null pointer)
       0x8 :   llseek                *symbol_table_name1!function     0xffff81243792
      0x10 :   read                  *symbol_table_name1!function     0xffff81243780
      0x18 :   write                 *symbol_table_name1!function     0xffff81243783
      0x20 :   aio_read              *symbol_table_name1!function     0x0 (null pointer)
      0x28 :   aio_write             *symbol_table_name1!function     0x0 (null pointer)
      0x30 :   readdir               *symbol_table_name1!function     0x0 (null pointer)
      0x38 :   poll                  *symbol_table_name1!function     0x0 (null pointer)
      0x40 :   unlocked_ioctl        *symbol_table_name1!function     0x0 (null pointer)
      0x48 :   compat_ioctl          *symbol_table_name1!function     0x0 (null pointer)
      0x50 :   mmap                  *symbol_table_name1!function     0x0 (null pointer)
      0x58 :   open                  *symbol_table_name1!function     0x0 (null pointer)
      0x60 :   flush                 *symbol_table_name1!function     0x0 (null pointer)
      0x68 :   release               *symbol_table_name1!function     0x0 (null pointer)
      0x70 :   fsync                 *symbol_table_name1!function     0x0 (null pointer)
      0x78 :   aio_fsync             *symbol_table_name1!function     0x0 (null pointer)
      0x80 :   fasync                *symbol_table_name1!function     0x0 (null pointer)
      0x88 :   lock                  *symbol_table_name1!function     0x0 (null pointer)
      0x90 :   sendpage              *symbol_table_name1!function     0x0 (null pointer)
      0x98 :   get_unmapped_area     *symbol_table_name1!function     0x0 (null pointer)
      0xa0 :   check_flags           *symbol_table_name1!function     0x0 (null pointer)
      0xa8 :   flock                 *symbol_table_name1!function     0x0 (null pointer)
      0xb0 :   splice_write          *symbol_table_name1!function     0xffff81243a77
      0xb8 :   splice_read           *symbol_table_name1!function     0x0 (null pointer)
      0xc0 :   setlease              *symbol_table_name1!function     0x0 (null pointer)
      0xc8 :   fallocate             *symbol_table_name1!function     0x0 (null pointer)
```

Fixing https://github.com/volatilityfoundation/volatility3/issues/1714
---
Previously if you used dt() on a struct where a member was paged out it would backtrace and show no future results. As per @atcuno suggestion it will now just display `N/A` and continue. Here is a short example with a nonsensical dt at 0x0. 

```
(layer_name) >>> dt('task_struct', 0x0)
symbol_table_name1!task_struct (1784 bytes) @ 0x0:
    0x0 :   state                           symbol_table_name1!long int                   N/A
    0x8 :   stack                           symbol_table_name1!pointer                    N/A
<snip>
```

Let me know what you think! Huge thanks to @atcuno and @ikelos for the encouragement. :pray:

### DIFF
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -605,9 +605,7 @@ class Volshell(interfaces.plugins.PluginInterface):
                 # volobject branch
                 if isinstance(
                     value,
-                    Union[
-                        interfaces.objects.ObjectInterface, interfaces.objects.Template
-                    ],
+                    (interfaces.objects.ObjectInterface, interfaces.objects.Template),
                 ):
                     if isinstance(value, objects.Pointer):
                         # show pointers in hex to match output for struct addrs
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -502,11 +502,10 @@ class Volshell(interfaces.plugins.PluginInterface):
                 member_type_name = self._get_type_name_with_pointer(
                     member_type
                 )  # special case for pointers to show what they point to
-                if len(member_type_name) > MAX_TYPENAME_DISPLAY_LENGTH:
-                    member_type_name = (
-                        f"{member_type_name[:MAX_TYPENAME_DISPLAY_LENGTH - 3]}..."
-                    )
                 len_typename = len(member_type_name)
+                if len(member_type_name) > MAX_TYPENAME_DISPLAY_LENGTH:
+                    len_typename = MAX_TYPENAME_DISPLAY_LENGTH
+                    member_type_name = f"{member_type_name[:len_typename - 3]}..."
                 if isinstance(volobject, interfaces.objects.ObjectInterface):
                     # We're an instance, so also display the data
                     try:
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -547,7 +547,7 @@ class Volshell(interfaces.plugins.PluginInterface):
         ],
         include_value: bool = True,
     ) -> str:
-        # build the display_type_string based on the aviable information
+        # build the display_type_string based on the available information
 
         if hasattr(volobject.vol, "size"):
             # the most common type to display, this shows their full size, e.g.:
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -502,6 +502,10 @@ class Volshell(interfaces.plugins.PluginInterface):
                 len_typename = len(member_type_name)
                 if isinstance(volobject, interfaces.objects.ObjectInterface):
                     # We're an instance, so also display the data
+                    try:
+                        value = self._display_value(getattr(volobject, member))
+                    except exceptions.InvalidAddressException:
+                        value = self._display_value(renderers.NotAvailableValue())
                     print(
                         "    " * dereference_count,
                         " " * (longest_offset - len_offset),
@@ -513,7 +517,7 @@ class Volshell(interfaces.plugins.PluginInterface):
                         member_type_name,
                         " " * (longest_typename - len_typename),
                         "  ",
-                        self._display_value(getattr(volobject, member)),
+                        value,
                     )
                 else:
                     # not provided with an actual object, nor an offset so just display the types
@@ -585,7 +589,9 @@ class Volshell(interfaces.plugins.PluginInterface):
 
     def _display_value(self, value: Any) -> str:
         try:
-            if isinstance(value, objects.Pointer):
+            if isinstance(value, interfaces.renderers.BaseAbsentValue):
+                return "N/A"
+            elif isinstance(value, objects.Pointer):
                 # show pointers in hex to match output for struct addrs
                 # highlight null or unreadable pointers
                 if value == 0:
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -622,7 +622,12 @@ class Volshell(interfaces.plugins.PluginInterface):
                     elif isinstance(value, objects.Array):
                         return repr([self._display_value(val) for val in value])
                     else:
-                        return f"offset: {hex(value.vol.offset)}"
+                        if self.context.layers[self.current_layer].is_valid(
+                            value.vol.offset
+                        ):
+                            return f"offset: {hex(value.vol.offset)}"
+                        else:
+                            return f"offset: {hex(value.vol.offset)} (unreadable)"
                 else:
                     # non volobject
                     if value is None:
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -630,7 +630,7 @@ class Volshell(interfaces.plugins.PluginInterface):
                     if value is None:
                         return "N/A"
                     else:
-                        return value
+                        return repr(value)
 
         except exceptions.InvalidAddressException:
             # if value causes an InvalidAddressException like BaseAbsentValue then display N/A
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -399,7 +399,7 @@ class Volshell(interfaces.plugins.PluginInterface):
         a pointer otherwise it returns just the normal type name."""
         pointer_marker = "*" * depth
         try:
-            if member_type.vol.object_class == objects.Pointer:
+            if isinstance(member_type, objects.Pointer):
                 sub_member_type = member_type.vol.subtype
                 # follow at most MAX_DEREFERENCE_COUNT pointers. A guard against, hopefully unlikely, infinite loops
                 if depth < MAX_DEREFERENCE_COUNT:
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -503,7 +503,7 @@ class Volshell(interfaces.plugins.PluginInterface):
                 if isinstance(volobject, interfaces.objects.ObjectInterface):
                     # We're an instance, so also display the data
                     try:
-                        value = self._display_value(getattr(volobject, member))
+                        value = self._display_value(volobject.member(member))
                     except exceptions.InvalidAddressException:
                         value = self._display_value(renderers.NotAvailableValue())
                     print(
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -489,9 +489,12 @@ class Volshell(interfaces.plugins.PluginInterface):
                 member_type_name = self._get_type_name_with_pointer(
                     member_type
                 )  # special case for pointers to show what they point to
+
+                # find the longest typename
                 longest_typename = max(len(member_type_name), longest_typename)
-                if longest_typename > MAX_TYPENAME_DISPLAY_LENGTH:
-                    longest_typename = MAX_TYPENAME_DISPLAY_LENGTH
+
+                # if the typename is very long then limit it to MAX_TYPENAME_DISPLAY_LENGTH
+                longest_typename = min(longest_typename, MAX_TYPENAME_DISPLAY_LENGTH)
 
             for member in sorted(
                 volobject.vol.members, key=lambda x: (volobject.vol.members[x][0], x)
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -32,6 +32,8 @@ try:
 except ImportError:
     has_ipython = False
 
+MAX_DEREFERENCE_COUNT = 4  # the max number of times display_type should follow pointers
+
 
 class Volshell(interfaces.plugins.PluginInterface):
     """Shell environment to directly interact with a memory image."""
@@ -386,6 +388,30 @@ class Volshell(interfaces.plugins.PluginInterface):
                 for i in disasm_types[architecture].disasm(remaining_data, offset):
                     print(f"0x{i.address:x}:\t{i.mnemonic}\t{i.op_str}")
 
+    def _get_type_name_with_pointer(
+        self,
+        member_type: Union[
+            str, interfaces.objects.ObjectInterface, interfaces.objects.Template
+        ],
+        depth: int = 0,
+    ) -> str:
+        """Takes a member_type from and returns the subtype name with a * if the member_type is
+        a pointer otherwise it returns just the normal type name."""
+        pointer_marker = "*" * depth
+        try:
+            if member_type.vol.object_class == objects.Pointer:
+                sub_member_type = member_type.vol.subtype
+                # follow at most MAX_DEREFERENCE_COUNT pointers. A guard against, hopefully unlikely, infinite loops
+                if depth < MAX_DEREFERENCE_COUNT:
+                    return self._get_type_name_with_pointer(sub_member_type, depth + 1)
+                else:
+                    return member_type_name
+        except AttributeError:
+            pass  # not all objects get a `object_class`, and those that don't are not pointers.
+        finally:
+            member_type_name = pointer_marker + member_type.vol.type_name
+        return member_type_name
+
     def display_type(
         self,
         object: Union[
@@ -418,26 +444,51 @@ class Volshell(interfaces.plugins.PluginInterface):
                 volobject.vol.type_name, layer_name=self.current_layer, offset=offset
             )
 
-        if hasattr(volobject.vol, "size"):
-            print(f"{volobject.vol.type_name} ({volobject.vol.size} bytes)")
-        elif hasattr(volobject.vol, "data_format"):
-            data_format = volobject.vol.data_format
-            print(
-                "{} ({} bytes, {} endian, {})".format(
-                    volobject.vol.type_name,
-                    data_format.length,
-                    data_format.byteorder,
-                    "signed" if data_format.signed else "unsigned",
-                )
-            )
+        # add special case for pointer so that information about the struct the
+        # pointer is pointing to is shown rather than simply the fact this is a
+        # pointer object. The "dereference_count < MAX_DEREFERENCE_COUNT" is to
+        # guard against loops
+        dereference_count = 0
+        while (
+            isinstance(volobject, objects.Pointer)
+            and dereference_count < MAX_DEREFERENCE_COUNT
+        ):
+            # before defreerencing the pointer, show it's information
+            print(f'{"    " * dereference_count}{self._display_simple_type(volobject)}')
+
+            # check that we can follow the pointer before dereferencing and do not
+            # attempt to follow null pointers.
+            if volobject.is_readable() and volobject != 0:
+                # now deference the pointer and store this as the new volobject
+                volobject = volobject.dereference()
+                dereference_count = dereference_count + 1
+            else:
+                # if we aren't able to follow the pointers anymore then there will
+                # be no more information to display as we've already printed the
+                # details of this pointer including the fact that we're not able to
+                # follow it anywhere
+                return
 
         if hasattr(volobject.vol, "members"):
+            # display the header for this object, if the orginal object was just a type string, display the type information
+            struct_header = f'{"    " * dereference_count}{volobject.vol.type_name} ({volobject.vol.size} bytes)'
+            if isinstance(object, str) and offset is None:
+                suffix = ":"
+            else:
+                # this is an actual object or an offset was given so the offset should be displayed
+                suffix = f" @ {hex(volobject.vol.offset)}:"
+            print(struct_header + suffix)
+
+            # it is a more complex type, so all members also need information displayed
             longest_member = longest_offset = longest_typename = 0
             for member in volobject.vol.members:
                 relative_offset, member_type = volobject.vol.members[member]
                 longest_member = max(len(member), longest_member)
                 longest_offset = max(len(hex(relative_offset)), longest_offset)
-                longest_typename = max(len(member_type.vol.type_name), longest_typename)
+                member_type_name = self._get_type_name_with_pointer(
+                    member_type
+                )  # special case for pointers to show what they point to
+                longest_typename = max(len(member_type_name), longest_typename)
 
             for member in sorted(
                 volobject.vol.members, key=lambda x: (volobject.vol.members[x][0], x)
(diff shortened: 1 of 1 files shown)


volatility3/cli/volshell/generic.py
```diff
@@ -418,67 +442,171 @@ def display_type(
...
+                elif not value.is_readable():
+                    suffix = " (unreadable pointer)"
+                else:
+                    suffix = ""
+                return f"{hex(value)}{suffix}"
+            elif isinstance(value, objects.PrimitiveObject):
+                return repr(value)
+            elif isinstance(value, objects.Array):
+                return repr([self._display_value(val) for val in value])
+            else:
```
ikelos, line 615: I'm not sure there's a case for `None` here?  It looks like it would fall through to this else, and hex(None) would probably return an exception, but that exception wouldn't be caught by the below?  I guess if value weren't a volatility object, then `value.vol` would throw an attribute error which wouldn't get caught.

Essentially, give we're doing all these test already, I'd add in one to split based on it being a vol object or not, and then on the non-volobject branch, have a case for if it's handed a `None` object...

volatility3/cli/volshell/generic.py
```diff
@@ -418,67 +442,171 @@ def display_type(
...
+                print("    " * dereference_count, self._display_simple_type(volobject))
+
+    def _display_simple_type(
+        self,
+        volobject: Union[
+            interfaces.objects.ObjectInterface, interfaces.objects.Template
+        ],
+        include_value: bool = True,
+    ) -> str:
+        # build the display_type_string based on the aviable information
```
ikelos: `aviable` is probably a typo...  5;P

volatility3/cli/volshell/generic.py
```diff
@@ -418,67 +442,171 @@ def display_type(
...
             longest_member = longest_offset = longest_typename = 0
             for member in volobject.vol.members:
                 relative_offset, member_type = volobject.vol.members[member]
                 longest_member = max(len(member), longest_member)
                 longest_offset = max(len(hex(relative_offset)), longest_offset)
-                longest_typename = max(len(member_type.vol.type_name), longest_typename)
+                member_type_name = self._get_type_name_with_pointer(
+                    member_type
+                )  # special case for pointers to show what they point to
+                longest_typename = max(len(member_type_name), longest_typename)
```
ikelos, line 502: Should there be a limit to the max typename length?  We'd then have to truncate at that point (and replace the last three characters with `...`) but it might prevent someone making a ridiculously long typename that throws everything off?  Does't have to be in the this PR, and it's only an idea, so may not be useful at all...  5:)

ikelos (commented):
Looking good!  I think the last `_display_value` function is a little brittle in places, but once that's hardended up a bit, this should be good to go in!  5:D

volatility3/cli/volshell/generic.py
eve-mem: Just pushed some changes for that - what do you think?

volatility3/cli/volshell/generic.py
```diff
@@ -587,26 +587,43 @@ def _display_simple_type(
...
+                            suffix = " (unreadable pointer)"
+                        else:
+                            suffix = ""
+                        return f"{hex(value)}{suffix}"
+                    elif isinstance(value, objects.PrimitiveObject):
+                        return repr(value)
+                    elif isinstance(value, objects.Array):
+                        return repr([self._display_value(val) for val in value])
+                    else:
+                        return hex(value.vol.offset)
```
ikelos: Do we want some way of indicating that this is the offset, rather than the value of the object?  I guess this will get hit by structs and any other complex volatility types?

volatility3/cli/volshell/generic.py
```diff
@@ -587,26 +587,43 @@ def _display_simple_type(
...
+                else:
+                    # non volobject
+                    if value is None:
+                        return "N/A"
+                    else:
+                        return value
```
ikelos: Give primatives get wrapped in `repr` should be make this consistent at least?

ikelos (commented):
Thanks, feels better to have made the distinction on type early on...  5:)

volatility3/cli/volshell/generic.py
eve-mem: Neat idea. I've added that in - what do you think?

volatility3/cli/volshell/generic.py
```diff
@@ -418,67 +445,194 @@ def display_type(
...
                 relative_offset, member_type = volobject.vol.members[member]
                 longest_member = max(len(member), longest_member)
                 longest_offset = max(len(hex(relative_offset)), longest_offset)
-                longest_typename = max(len(member_type.vol.type_name), longest_typename)
+                member_type_name = self._get_type_name_with_pointer(
+                    member_type
+                )  # special case for pointers to show what they point to
+                longest_typename = max(len(member_type_name), longest_typename)
+                if longest_typename > MAX_TYPENAME_DISPLAY_LENGTH:
+                    longest_typename = MAX_TYPENAME_DISPLAY_LENGTH
```
ikelos: I think `max` can take multiple parameters, so... this should work (although it might need reformatting for ruff/black)?
```suggestion
                longest_typename = max(len(member_type_name), longest_typename, MAX_TYPENAME_DISPLAY_LENGTH)
```

volatility3/cli/volshell/generic.py
```diff
@@ -418,67 +445,194 @@ def display_type(
...
                 len_member = len(member)
-                len_typename = len(member_type.vol.type_name)
+                member_type_name = self._get_type_name_with_pointer(
+                    member_type
+                )  # special case for pointers to show what they point to
+                if len(member_type_name) > MAX_TYPENAME_DISPLAY_LENGTH:
+                    member_type_name = (
+                        f"{member_type_name[:MAX_TYPENAME_DISPLAY_LENGTH - 3]}..."
+                    )
+                len_typename = len(member_type_name)
```
ikelos: Feels marginally tidier, but not a big deal...
```suggestion
                len_typename = len(member_type_name)
                if len_typename > MAX_TYPENAME_DISPLAY_LENGTH:
                    len_typename = MAX_TYPENAME_DISPLAY_LENGTH
                    member_type_name = f"{member_type_name[:len_typename - 3]}..."
```

volatility3/cli/volshell/generic.py
eve-mem: The issue with that is then max might be 8000 for the very long type name. 

We don't actually want the max of all three, we want the biggest unless it's reached the cut off.

ikelos: Oh... hmmm... so it's almost like `min(max(len(member_type_name), longest_type_name), MAX_TYPENAME_DISPLAY_LENGTH)` then?  I guess that's getting less readable, yeah, fair enough, your call I can go with either.  Just resolve this when you're happy...  5:)

eve-mem: Yes - that's exactly right.

eve-mem: I've swapped to using min, but put it on another line with some extra comments to hopefully keep readable. It looks better than the if imo.

volatility3/cli/volshell/generic.py
eve-mem: Yes its the structs etc that hit this. 

Here is a before and after of the PR. Look at the `stack` and `usage` between the two. For me it feels natural that it's the offset, but maybe I've just been conditioned with years of volshell usage to accept it - it's what I'm used to. We could mark it with something, maybe an `@` ?


Before:
```
(layer_name) >>> dt(ps()[0])
symbol_table_name1!task_struct (1784 bytes)
   0x0 :   state                           symbol_table_name1!long int                   1
   0x8 :   stack                           symbol_table_name1!pointer                    149534111531008
  0x10 :   usage                           symbol_table_name1!atomic_t                   0x88001f994750
  0x14 :   flags                           symbol_table_name1!unsigned int               4202752
  0x18 :   ptrace                          symbol_table_name1!unsigned int               0
  0x20 :   wake_entry                      symbol_table_name1!llist_node                 0x88001f994760
```

After:
```
symbol_table_name1!task_struct (1784 bytes) @ 0x88001f994740:
    0x0 :   state                           symbol_table_name1!long int                     1
    0x8 :   stack                           *symbol_table_name1!void                        0x88001f998000
   0x10 :   usage                           symbol_table_name1!atomic_t                     0x88001f994750
   0x14 :   flags                           symbol_table_name1!unsigned int                 4202752
   0x18 :   ptrace                          symbol_table_name1!unsigned int                 0
   0x20 :   wake_entry                      symbol_table_name1!llist_node                   0x88001f994760
```

ikelos: Yeah, I'm not sure what a suitable symbol would be (and I don't use the thing very often, so I probably won't make a good choice).  Is `@` clear?  Maybe surround them in brackets or a literal `offset: {value}`?

ikelos: If someone wants to scrape the output (why would they, but... people do strange things) we probably want to make it easy to handle too...

eve-mem: Could go more extreme and expand them out? It'll be the nested structures essentially. 
```
symbol_table_name1!task_struct (1784 bytes) @ 0x88001f994740:
    0x0 :   state                           symbol_table_name1!long int                     1
    0x8 :   stack                           *symbol_table_name1!void                        0x88001f998000
   0x10 :   usage                           symbol_table_name1!atomic_t                     0x88001f994750
   0x10 :   usage.counter                   symbol_table_name1!int                          2
   0x14 :   flags                           symbol_table_name1!unsigned int                 4202752
```

This could get messy quickly with structures all the way down? Equally maybe useful to just get all the information out there. 

If this seems like a bad idea I think `offset: {value}` is probably the way forward.

ikelos: I thought one interation of this had done a tree view across lines, to show the pointers and their offsets and then finally the leaf item and its value.  Yeah, I guess we just need to make sure people know what they're seeing when they get a ***table!thing with just an offset next to it...

eve-mem: Here's a better example with two fields in the nested structure

```
symbol_table_name1!task_struct (1784 bytes) @ 0x88001f994740:
    0x0 :   state                           symbol_table_name1!long int                     1
    0x8 :   stack                           *symbol_table_name1!void                        0x88001f998000
   0x10 :   usage                           symbol_table_name1!atomic_t                     0x88001f994750
<snip>
  0x610 :   pi_state_list                   symbol_table_name1!list_head                    0x88001f994d50
  0x610 :   pi_state_list.next              *symbol_table_name1!list_head                   0x88001f994d50
  0x618 :   pi_state_list.prev              *symbol_table_name1!list_head                   0x88001f994d50
  0x620 :   pi_state_cache                  *symbol_table_name1!futex_pi_state              0x0 (null pointer)
<snip>
```

volatility3/cli/volshell/generic.py
ikelos: This doesn't include the `...` anymore, but that's ok if you prefer it that way...

eve-mem: No that's a mistake...

volatility3/cli/volshell/generic.py
```diff
@@ -418,67 +445,196 @@ def display_type(
...
                 len_offset = len(hex(relative_offset))
                 len_member = len(member)
-                len_typename = len(member_type.vol.type_name)
+                member_type_name = self._get_type_name_with_pointer(
+                    member_type
+                )  # special case for pointers to show what they point to
+                len_typename = len(member_type_name)
+                if len(member_type_name) > MAX_TYPENAME_DISPLAY_LENGTH:
+                    len_typename = MAX_TYPENAME_DISPLAY_LENGTH
+                    member_type_name = f"{member_type_name[:len_typename - 3]}..."
```
eve-mem, line 520: @ikelos This is where the type name gets truncated if it's too long.

ikelos: Ah cool, sorry, missed it in the changes.

volatility3/cli/volshell/generic.py
eve-mem: Ah I've worked it out. It's still there, this part is just working out the max size. The truncation happens a few lines down.

volatility3/cli/volshell/generic.py
eve-mem: No worries. When I went to check in vscode I'd swapped branch to develop and then couldn't find the `...` either

Took me a second to spot my mistake :smile:

volatility3/cli/volshell/generic.py
eve-mem: I think at the moment I'm inclined to leave it as it is. Display the offset with no extra context and don't expand the nested structures (let the user choose to with `dt(foo.bar)`. The changes in this PR makes it so that accessing nested structures or linked ones ends up the same `dt(foo.bar)` or `dt(foo.pointer)` - where in the past you'd have needed `dt(foo.pointer.deference())`. 

I did a quick and ugly attempt at expanding the nested structures and it gets really big. 

I can be convinced that adding `offset: {value}` is clearer - and if so I'd add the some prefix to the pointer output I think. e.g. this struct is over there in both cases. (although nested one isn't really that far away...!). I'm not bothered by it - just very used to the current output.

ikelos: Just trying to get a clear understanding of what people would see and whether they'll be able to clearly understand it.  Outputting a value but then swapping to an offset for only `pointers` strikes me as potentially confusing, particularly if we don't indicate it in the output?  I'm just wondering about a situation like a pointer as an offset, pointing to an int (that happens to be in the memory range).  Could you get to the value of the pointer (the address of the int) and the value of the `int`, and would it be clear in each circumstance?  (Sorry, I'm a bit tired, so you might have already explained it but it's still a bit fuzzy in my head)...

eve-mem: Nope all good questions. You can get all those things.

I think that this probably needs some documentation along with it, if you're asking questions now so will people in the future. I don't think i can do it tonight, but I'll do that soon.

volatility3/cli/volshell/generic.py
```diff
@@ -418,67 +445,196 @@ def display_type(
...
+            return display_type_string
+
+    def _display_value(self, value: Any) -> str:
+        try:
+            # if value is a BaseAbsentValue they display N/A
+            if isinstance(value, interfaces.renderers.BaseAbsentValue):
+                return "N/A"
+            else:
+                # volobject branch
+                if isinstance(
```
dgmcdona, line 617: This check will raise a `TypeError`:

```
TypeError: Subscripted generics cannot be used with class and instance checks
```

It's fixable by just enclosing the two types in parentheses instead of `Union[]`.

volatility3/cli/volshell/generic.py
```diff
@@ -418,67 +445,196 @@ def display_type(
...
+                    value,
+                    Union[
+                        interfaces.objects.ObjectInterface, interfaces.objects.Template
+                    ],
```
ikelos:
```suggestion
                    (interfaces.objects.ObjectInterface, interfaces.objects.Template),
```

eve-mem:
I'll try and get those merge conflicts fixed for tomorrow. Thanks for the help on those other bits @dgmcdona and @ikelos

volatility3/cli/volshell/generic.py
eve-mem: As discussed on slack ot would also be useful to flag if offsets to embedded structs are readable as per the pointers.

eve-mem: @ikelos I think I've done this now. :smile:

volatility3/cli/volshell/generic.py
```diff
@@ -418,67 +445,199 @@ def display_type(
...
+                        return f"{hex(value)}{suffix}"
+                    elif isinstance(value, objects.PrimitiveObject):
+                        return repr(value)
+                    elif isinstance(value, objects.Array):
+                        return repr([self._display_value(val) for val in value])
+                    else:
+                        if self.context.layers[self.current_layer].is_valid(
+                            value.vol.offset
+                        ):
+                            return f"offset: {hex(value.vol.offset)}"
```
ikelos: Think this can be `0x{value.vol.offset:x}` rather than `{hex(value.vol.offset)}`.  Absolutely a stylistic choice though...  5;)

volatility3/cli/volshell/generic.py
ikelos: Yep, looks good.  Just need to rebase it against the current develop and check it passes the tests!  See, easy!  5;D

volatility3/cli/volshell/generic.py
eve-mem: I don't mind either way - would you rather `0x{value.vol.offset:x}`? (can add it to the style guide PR?)

ikelos: Yeah, I think so?  Using built in format modifiers seems better than a separate function call to convert it?  Sure, I'll get it added to the coding style...

eve-mem: Makes sense - done! :smile:

PR #1749, opened 2025-03-28
Volshell: handle case where paged out member would cause backtrace for dt output. 

### LINKED ISSUE #1714: volshell dt() throws exception on invalid pointer members

eve-mem:
Rather than waiting on https://github.com/volatilityfoundation/volatility3/pull/1748 this PR would fix https://github.com/volatilityfoundation/volatility3/issues/1714

It's just the changes suggested by @atcuno.

I know it's needed for the parity work this weekend.

### DIFF
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -450,7 +450,7 @@ class Volshell(interfaces.plugins.PluginInterface):
                 if isinstance(volobject, interfaces.objects.ObjectInterface):
                     # We're an instance, so also display the data
                     try:
-                        value = self._display_value(getattr(volobject, member))
+                        value = self._display_value(volobject.member(member))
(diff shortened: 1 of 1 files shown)


ikelos (approved):

eve-mem:
Sorry - I branched from this to also fix https://github.com/volatilityfoundation/volatility3/issues/1705.

When I committed it updated this PR too - so this will now fix two things. Again using the suggestions from @atcuno.

So this is two for one... both for the parity work.

Fixes https://github.com/volatilityfoundation/volatility3/issues/1705
Fixes https://github.com/volatilityfoundation/volatility3/issues/1714

PR #1751, opened 2025-03-29
Don't delete files that are still accessible from the cache

### LINKED ISSUE #1255: Cache Update Issues
**Describe the bug**
When having and using both the latest release version of Volatility 3 and the latest development version of Volatility 3 on the same system, the  "updating caches" function has to re-update frequently.

**Context**
Volatility Version:  Volatility 3, v 2.7.0 & the latest development version
Operating System:  Windows 10
Python Version:  3.11
Suspected Operating System:  (not sure what this means... of the memory sample? If so, Windows 10 and Windows 8)
Command:  C:\Python311\python.exe "C:\Tools\Volatility\3.X\vol.py" -f "C:\Temp\memimage.mem" windows.cachedump.Cachedump

**To Reproduce**
1. Download and unzip the latest release version of Volatiltiy 3 (2.7.0, at the time of this writing) to C:\Tools\Volatility\3.x
2. Download and unzip the latest dev version of Volatility 3 to C:\Tools\Volatility\3.x-DEV
3. Download all symbol table packs and place the appropriate directories
3a. C:\Tools\Volatility\3.X\volatility3\symbols
3b. C:\Tools\Volatility\3.X-DEV\volatility3\volatility3\symbols
4. Run the windows.cachedump.Cachedump plugin with the command above using the release version, then the developement version, then the release version again.
5. You will see that 3.X release updates the caches, then 3.x-DEV updates its caches, then 3.X release has to update its caches again, if you alternate running the same command back and forth between release and development versions in the folder structure I've described.  

**Expected behavior**
I expect that each version (3.X release and 3.X DEV) will only need to update it's caches once.

**Example output**
Please copy and paste the text demonstrating the issue, ideally with verbose output turned on (`vol.py -vvv ...`).  
-The output in the CMD window doesn't keep the cache updating output text in the window when the command has completed running.

**Additional information**

ikelos:
This is a naive means to ensure we never delete files that are still valid/exist from the cache.  The downside with this is that even if a file isn't in the user's symbols path, it might still be retrieved and used (which would be confusing, so the PR can't be committed in its current form).  We'd probably need to filter the returned list of symbols to make sure they're remote or that they're in the user's current symbol directory...

Closes #1255


ikelos:
So we need to make sure the cache is filtered down to just file in the symbols path (or remote).

ikelos:
This didn't rebase well.  Rewritten at #1927.  Closing this off.

PR #1752, opened 2025-03-29
Revert "Feature/use less memory"

ikelos:
Reverts volatilityfoundation/volatility3#1736


ikelos:
This broke on a section that used the lower level maps (but didn't show up until a full test).

PR #1755, opened 2025-03-31
Fix several bugs in check_afinfo. Update through latest kernels. Matc…

atcuno:
…h current Volatility coding standards

@ikelos I would like to get this added for parity. While performing some testing today, I noticed we were missing support for kernels since 4.18 (2018). This issue was masked from previous testing in how the missing symbol was being reported, which I fixed. I also rewrote the plugin essentially to match our current coding standards.


volatility3/framework/plugins/linux/check_afinfo.py
```diff
@@ -97,39 +117,92 @@ def _generator(self):
...
+        Checks the operations structures for network protocols of >= 4.18 systems
+        """
+        vmlinux = context.modules[vmlinux_name]
+
+        ops_structs = [
+            "raw_seq_ops",
+            "udp_seq_ops",
+            "arp_seq_ops",
+            "unix_seq_ops",
+            "udp6_seq_ops", "raw6_seq_ops",
```
ikelos:
```suggestion
            "udp6_seq_ops", 
            "raw6_seq_ops",
```

ikelos (approved):
Looks good, thanks!  5:)

PR #1758, opened 2025-04-02
Windows MFTScan: Memory usage performance fixes

### LINKED ISSUE #1725: TreeGrid: Excessive memory usage for some plugins
The `windows.mftscan.MFTScan` encounters an issue with memory usage that seems to be inherent in the current implementation of `TreeGrid`. Even for renderers that stream output as it comes in like the `QuickTextRenderer` and the `JsonLinesRenderer`, `TreeGrid._children` continues to grow in size as nodes are visited and the tree is populated.

`MFTScan` routinely produces anywhere from hundreds of thousands to over one million entries, leading to thrashing. I was able to circumvent this by clearing `TreeGrid._children` every time the generator came back to a new root entry, but this seems to go against the entire existing design.

@ikelos any ideas how best to resolve this? I had considered changing the underlying storage of `TreeGrid._children` to an interface of some sort that allows for alternative storage implementations beyond just in-memory storage (so something like a temporary SQLite database), but that introduces the challenge of serializing/deserializing `TreeNode`s to disk. That can be done with the existing `BaseTypes`, but we'd be introducing a requirement for `BaseTypes` that they must be serializable/deserializable.

dgmcdona:
This fixes a major memory performance issue in the `MFTScan` plugins. We had been struggling with large memory consumption in these plugins before, and I assumed that the issue was with the `TreeGrid` implementation. There ended up being a larger issue within the actual plugin, which was unconverted `object.String` objects being stored in the `record_map` that is populated and passed around to various classmethods. This resulted in a map that was far larger than it needed to be, since those values would only ever be converted to strings for the treegrid anyway. I discovered this by accident when implementing a simple SQLite cache as a replacement for the map - watching the SQLite database, I never saw it grow to more than a few dozen megabytes. 

```
[ins] In [5]: pympler.asizeof.asizeof(rec_name)
Out[5]: 312648

[ins] In [6]: pympler.asizeof.asizeof(str(rec_name))
Out[6]: 64
```

This PR:
- Properly converts values to strings before storing them in the map
- Fixes type hints in the MFT extension classes to reflect that they return `objects.String`, not `str`.
- Updates type hints throughout the MFTScan plugin classes for clarity.
- Reimplements the map as a typed `DefaultDict` for simplicity
- Adds a simple `MFTRecord` class that uses `__slots__` in place of the lists that were being used as map values before. This should result in some more (small) performance gains, and help with readability since values are no longer accessed by index.

You can see how much better it performs on an 8GB sample here - `sys` time that was presumably spent swapping dropped from ~47 seconds to ~8 seconds:
```
~/volatility3 (fix_tests|✚2) $ time vol -f ~/memory_samples/data.lime -r jsonl windows.mftscan.ResidentData > /dev/null
Volatility 3 Framework 2.26.0
________________________________________________________
Executed in  400.14 secs    fish           external
   usr time  340.85 secs    0.32 millis  340.85 secs
   sys time   47.16 secs    1.46 millis   47.16 secs
 
~/volatility3 (fix_tests|●1) $ time vol -f ~/memory_samples/data.lime -r jsonl windows.mftscan.ResidentData > /dev/null
Volatility 3 Framework 2.26.0
________________________________________________________
Executed in  321.45 secs    fish           external
   usr time  314.03 secs    0.29 millis  314.03 secs
   sys time    8.18 secs    1.21 millis    8.18 secs
```

closes #1725

### DIFF
--- a/volatility3/framework/plugins/windows/mftscan.py
+++ b/volatility3/framework/plugins/windows/mftscan.py
@@ -20,7 +20,7 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
 
     _required_framework_version = (2, 0, 0)
 
-    _version = (2, 0, 1)
+    _version = (3, 0, 0)
 
     class MFTScanResult(NamedTuple):
         offset: format_hints.Hex
@@ -266,7 +266,7 @@ class ADS(interfaces.plugins.PluginInterface):
 
     _required_framework_version = (2, 22, 0)
 
-    _version = (1, 0, 2)
+    _version = (2, 0, 0)
 
     class ADSResult(NamedTuple):
         offset: format_hints.Hex
@@ -366,7 +366,7 @@ class ResidentData(interfaces.plugins.PluginInterface):
 
     _required_framework_version = (2, 22, 0)
 
-    _version = (1, 0, 2)
+    _version = (2, 0, 0)
 
     class ResidentDataResult(NamedTuple):
         offset: format_hints.Hex
--- a/volatility3/framework/plugins/windows/mftscan.py
+++ b/volatility3/framework/plugins/windows/mftscan.py
@@ -114,7 +114,6 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
 
         # get each of the individual Field Sets
         mft_object_type_name = symbol_table + constants.BANG + "MFT_ENTRY"
-        attribute_object_type_name = symbol_table + constants.BANG + "ATTRIBUTE"
 
         record_map: DefaultDict[str, MFTRecord] = DefaultDict(MFTRecord)
 
@@ -127,30 +126,9 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
                     mft_object_type_name, offset=offset, layer_name=layer.name
                 )
 
-                # We will update this on each pass in the next loop and use it as the new offset.
-                attr_base_offset = mft_record.FirstAttrOffset
-                attr: mft.MFTAttribute = context.object(
-                    attribute_object_type_name,
-                    offset=offset + attr_base_offset,
-                    layer_name=layer.name,
-                )
-
-                # There is no field that has a count of Attributes
-                # Keep Attempting to read attributes until we get an invalid attr_header.AttrType
-                while attr.Attr_Header.AttrType.is_valid_choice:
-                    yield from attr_callback(record_map, mft_record, attr, symbol_table)
-
-                    # If there's no advancement the loop will never end, so break it now
-                    if attr.Attr_Header.Length == 0:
-                        break
-
-                    # Update the base offset to point to the next attribute
-                    attr_base_offset += attr.Attr_Header.Length
-                    # Get the next attribute
-                    attr: mft.MFTAttribute = context.object(
-                        attribute_object_type_name,
-                        offset=offset + attr_base_offset,
-                        layer_name=layer.name,
+                for attribute in mft_record.attributes(symbol_table):
+                    yield from attr_callback(
+                        record_map, mft_record, attribute, symbol_table
                     )
 
     @classmethod
--- a/volatility3/framework/plugins/windows/mftscan.py
+++ b/volatility3/framework/plugins/windows/mftscan.py
@@ -281,7 +281,7 @@ class ADS(interfaces.plugins.PluginInterface):
     def get_requirements(cls):
         return [
             requirements.VersionRequirement(
-                name="MFTScan", component=MFTScan, version=(2, 0, 0)
+                name="MFTScan", component=MFTScan, version=(3, 0, 0)
             ),
             requirements.TranslationLayerRequirement(
                 name="primary",
@@ -380,7 +380,7 @@ class ResidentData(interfaces.plugins.PluginInterface):
     def get_requirements(cls):
         return [
             requirements.VersionRequirement(
-                name="MFTScan", component=MFTScan, version=(2, 0, 0)
+                name="MFTScan", component=MFTScan, version=(3, 0, 0)
             ),
             requirements.TranslationLayerRequirement(
                 name="primary",
--- a/volatility3/framework/plugins/windows/mftscan.py
+++ b/volatility3/framework/plugins/windows/mftscan.py
@@ -234,17 +234,24 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
             )
 
     def generate_timeline(self):
-        for row in self._generator():
-            _depth, row_data = row
+        for record in self.enumerate_mft_records(
+            self.context, self.config_path, self.config["primary"]
+        ):
+            fname = record.longest_filename()
 
-            # Only Output FN Records
-            if row_data[6] in ("FILE_NAME", "STANDARD_INFORMATION"):
-                filename = row_data[-1]
-                description = f"MFT {row_data[6]} entry for {filename}"
-                yield (description, timeliner.TimeLinerType.CREATED, row_data[7])
-                yield (description, timeliner.TimeLinerType.MODIFIED, row_data[8])
-                yield (description, timeliner.TimeLinerType.CHANGED, row_data[9])
-                yield (description, timeliner.TimeLinerType.ACCESSED, row_data[10])
+            for _, item in self.parse_standard_information_records(record):
+                description = f"MFT {item.attribute_type} entry for {fname}"
+                yield (description, timeliner.TimeLinerType.CREATED, item.created)
+                yield (description, timeliner.TimeLinerType.MODIFIED, item.modified)
+                yield (description, timeliner.TimeLinerType.CHANGED, item.updated)
+                yield (description, timeliner.TimeLinerType.ACCESSED, item.accessed)
+
+            for _, item in self.parse_filename_records(record):
+                description = f"MFT {item.attribute_type} entry for {item.filename}"
+                yield (description, timeliner.TimeLinerType.CREATED, item.created)
+                yield (description, timeliner.TimeLinerType.MODIFIED, item.modified)
+                yield (description, timeliner.TimeLinerType.CHANGED, item.updated)
+                yield (description, timeliner.TimeLinerType.ACCESSED, item.accessed)
 
     def run(self):
         return renderers.TreeGrid(
--- a/volatility3/framework/plugins/windows/mftscan.py
+++ b/volatility3/framework/plugins/windows/mftscan.py
@@ -1,12 +1,11 @@
 # This file is Copyright 2022 Volatility Foundation and licensed under the Volatility Software License 1.0
 # which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
 #
-import contextlib
 import datetime
 import logging
-from typing import Callable, DefaultDict, Iterator, Optional, Tuple
+from typing import Iterator, NamedTuple, Optional, Tuple, Union
 
-from volatility3.framework import constants, exceptions, interfaces, renderers
+from volatility3.framework import constants, exceptions, interfaces, objects, renderers
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import conversion, format_hints
 from volatility3.framework.symbols import intermed
@@ -16,28 +15,26 @@ from volatility3.plugins import timeliner, yarascan
 vollog = logging.getLogger(__name__)
 
 
-class MFTRecord:
-    # TODO: Change to dataclass with (slots=True) if/when we move minimum
-    # Python version up to 3.10
-    __slots__ = ["record_name", "data_count", "offset"]
-
-    def __init__(
-        self,
-        record_name: Optional[str] = None,
-        data_count: int = 0,
-        offset: Optional[int] = None,
-    ):
-        self.record_name = record_name
-        self.data_count = data_count
-        self.offset = offset
-
-
 class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
     """Scans for MFT FILE objects present in a particular windows memory image."""
 
     _required_framework_version = (2, 0, 0)
 
     _version = (2, 0, 1)
+
+    class MFTScanResult(NamedTuple):
+        offset: format_hints.Hex
+        record_type: str
+        record_number: int
+        link_count: int
+        mft_type: str
+        permissions: Union[str, interfaces.renderers.BaseAbsentValue]
+        attribute_type: str
+        created: Union[interfaces.renderers.BaseAbsentValue, datetime.datetime]
+        modified: Union[interfaces.renderers.BaseAbsentValue, datetime.datetime]
+        updated: Union[interfaces.renderers.BaseAbsentValue, datetime.datetime]
+        accessed: Union[interfaces.renderers.BaseAbsentValue, datetime.datetime]
+        filename: Union[interfaces.renderers.BaseAbsentValue, objects.String]
 
     @classmethod
     def get_requirements(cls):
@@ -66,16 +63,7 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
         context: interfaces.context.ContextInterface,
         config_path: str,
         primary_layer_name: str,
-        attr_callback: Callable[
-            [
-                DefaultDict[str, MFTRecord],
-                mft.MFTEntry,
-                mft.MFTAttribute,
-                str,
-            ],
-            Iterator[Tuple],
-        ],
-    ) -> Iterator[Tuple]:
+    ) -> Iterator[mft.MFTEntry]:
         try:
             primary = context.layers[primary_layer_name]
         except KeyError:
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -98,23 +120,24 @@ def enumerate_mft_records(
         )
 
         # get each of the individual Field Sets
-        mft_object = symbol_table + constants.BANG + "MFT_ENTRY"
-        attribute_object = symbol_table + constants.BANG + "ATTRIBUTE"
+        mft_object_typ_name = symbol_table + constants.BANG + "MFT_ENTRY"
```
ikelos: Any reason these are missing the `e`?

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -98,23 +120,24 @@ def enumerate_mft_records(
...
+        attribute_object_typ_name = symbol_table + constants.BANG + "ATTRIBUTE"
 
-        record_map = {}
+        record_map: DefaultDict[str, MFTRecord] = DefaultDict(MFTRecord)
```
ikelos: I'm a little concerned about the use of `DefaultDict` here?  When are we creating blank `MFTRecords`, and does that not mean we need more code to cope with them if we ever look them up?

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -17,6 +23,22 @@
 vollog = logging.getLogger(__name__)
 
 
+class MFTRecord:
```
ikelos: This feels like it could be `namedtuple`, rather than a class?  `__slots__` helps reduce its memory footprint, but not being a class would help it more I think?

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -17,6 +23,22 @@
...
+    # TODO: Change to dataclass with (slots=True) if/when we move minimum
+    # Python version up to 3.10
+    __slots__ = ["record_name", "data_count", "offset"]
+
+    def __init__(
+        self,
+        record_name: Optional[str] = None,
```
ikelos: Allowing all of these to be blank enables the use of `DefaultDict` but strikes me that it could lead to a lot of potential junk entries?

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -178,7 +201,7 @@ def parse_mft_records(
             fn_object = symbol_table_name + constants.BANG + "FILE_NAME_ENTRY"
 
             attr_data = attr.Attr_Data.cast(fn_object)
-            file_name = attr_data.get_full_name()
+            file_name = str(attr_data.get_full_name())
```
ikelos: I generally try to get people to leave string objects as strings (in case callers need to get back to that information), but sadly for memory reasons it's probably best to allow it here...  5:S  I guess we can't just do this on the generator so they go out of the plugin as strings but are otherwise carried around as the full types?  My concern would be other people calling this (although the type information doesn't imply they come with provenance information so... yeah, probably ok, I'm just trying to figure out how to generally discourage it except in situations of memory exhaustion)

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -236,53 +264,53 @@ def parse_data_record(
...
         """
         Parses DATA records while maintaining the FILE_NAME association
         from previous parsing of the record
         Suports returning the first/main $DATA as well as however many
         ADS records a file might have
         """
-        if mft_record.vol.offset not in record_map:
-            # file name, DATA count, offset
-            record_map[mft_record.vol.offset] = [renderers.NotAvailableValue(), 0, None]
+        rec = record_map[mft_record.vol.offset]
```
ikelos: Oh, now I see why you used the `DefaultDict`.  Given that `namedtuples` need to be constructed initially anyway (and then can't be changed), I'd sooner build up the component, construct the `namedtuple` and then assign it into the record_map.  I'm guessing this doesn't ever overwrite existing answers (otherwise that's also a conflict we should cope with directly, rather than just blatting over prior results without noting whether there's a discrepancy)...

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -236,53 +264,53 @@ def parse_data_record(
...
             content,
         )
 
     @classmethod
     def parse_data_records(
         cls,
-        record_map: Dict[int, Tuple[str, int, int]],
-        mft_record: interfaces.objects.ObjectInterface,
-        attr: interfaces.objects.ObjectInterface,
+        record_map: DefaultDict[str, MFTRecord],
```
ikelos:
```suggestion
        record_map: Dict[str, MFTRecord],
```
Again, pretty sure type can just be `Dict[str, MFTRecord]`

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -427,11 +455,11 @@ def get_requirements(cls):
     @classmethod
     def parse_first_data_records(
         cls,
-        record_map: Dict[int, Tuple[str, int, int]],
-        mft_record: interfaces.objects.ObjectInterface,
-        attr: interfaces.objects.ObjectInterface,
+        record_map: DefaultDict[str, MFTRecord],
```
ikelos:
```suggestion
        record_map: Dict[str, MFTRecord],
```
Again, the type doesn't need to be DefaultDict, I think just Dict should be fine here?

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -10,15 +10,15 @@
 class MFTEntry(objects.StructType):
     """This represents the base MFT Record"""
 
-    def get_signature(self) -> str:
+    def get_signature(self) -> "objects.String":
```
ikelos:
```suggestion
    def get_signature(self) -> objects.String:
```
Does importing the objects class cause a circular reference here? If not, I'd ditch the quotes, they're usually used to delay requirements to avoid needing to import classes, but it looks like the class has already been imported by now?

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -10,15 +10,15 @@
...
         signature = self.Signature.cast("string", max_length=4, encoding="latin-1")
         return signature
 
 
 class MFTFileName(objects.StructType):
     """This represents an MFT $FILE_NAME Attribute"""
 
-    def get_full_name(self) -> str:
+    def get_full_name(self) -> "objects.String":
```
ikelos:
```suggestion
    def get_full_name(self) -> objects.String:
```

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -28,7 +28,7 @@ def get_full_name(self) -> str:
 class MFTAttribute(objects.StructType):
     """This represents an MFT ATTRIBUTE"""
 
-    def get_resident_filename(self) -> Optional[str]:
+    def get_resident_filename(self) -> Optional["objects.String"]:
```
ikelos:
```suggestion
    def get_resident_filename(self) -> Optional[objects.String]:
```

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -51,7 +51,7 @@ def get_resident_filename(self) -> Optional[str]:
         except exceptions.InvalidAddressException:
             return None
 
-    def get_resident_filecontent(self) -> Optional[bytes]:
+    def get_resident_filecontent(self) -> Optional["objects.Bytes"]:
```
ikelos:
```suggestion
    def get_resident_filecontent(self) -> Optional[objects.Bytes]:
```

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -357,7 +385,7 @@ def get_requirements(cls):
     @classmethod
     def parse_ads_data_records(
         cls,
-        record_map: Dict[int, Tuple[str, int, int]],
+        record_map: DefaultDict[str, MFTRecord],
```
ikelos:
```suggestion
        record_map: Dict[str, MFTRecord],
```

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -204,11 +227,11 @@ def parse_mft_records(
     @classmethod
     def parse_data_record(
         cls,
-        mft_record: interfaces.objects.ObjectInterface,
-        attr: interfaces.objects.ObjectInterface,
-        record_map: Dict[int, Tuple[str, int, int]],
+        mft_record: mft.MFTEntry,
+        attr: mft.MFTAttribute,
+        record_map: DefaultDict[str, MFTRecord],
```
ikelos:
```suggestion
        record_map: Dict[str, MFTRecord],
```

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -131,18 +154,18 @@ def enumerate_mft_records(
...
                         layer_name=layer.name,
                     )
 
     @classmethod
     def parse_mft_records(
         cls,
-        record_map: Dict[int, Tuple[str, int, int]],
-        mft_record: interfaces.objects.ObjectInterface,
-        attr: interfaces.objects.ObjectInterface,
+        record_map: DefaultDict[str, MFTRecord],
```
ikelos:
```suggestion
        record_map: Dict[str, MFTRecord],
```
The type here could surely just be a Dict?

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -53,14 +75,14 @@ def enumerate_mft_records(
         primary_layer_name: str,
         attr_callback: Callable[
             [
-                Dict[int, Tuple[str, int, int]],
-                interfaces.objects.ObjectInterface,
-                interfaces.objects.ObjectInterface,
+                DefaultDict[str, MFTRecord],
```
ikelos:
```suggestion
                Dict[str, MFTRecord],
```

ikelos (commented):
Generally looks good thanks, a couple quibbles.  I don't really like `DefaultDict` since it leaves you with a dictionary with a whole bunch of potentially bad data (if anyone asks for a random offset from it), so I'd prefer to keep them out of the code.

I also spent a long time trying to get people to leave the volatility information in place when returning (particularly from an API call, rather than a generator), but in this case given the excessive memory it consumes, I'll allow it.  I'd quite like a comment adding next to the casts, to say that it's specifically for memory consumption reasons (to avoid people using it as excuse to do it in their code all the time)...

Lastly type hinting using quotes should only be needed when there's circular/delayed import issues, which I don't think there are here.

Otherwise looks really neat (and hopefully drops the memory usage for this plugin even further!).  5:)

volatility3/framework/plugins/windows/mftscan.py
dgmcdona: Just a reflex of leaving off the 'e' to avoid accidentally shadowing the builtin, I'll add them in here though.

volatility3/framework/plugins/windows/mftscan.py
dgmcdona: Surprisingly, a class with `__slots__` seems to use less memory than even a simple tuple:

```
[ins] In [4]: class MFTRecord:
         ...:     # TODO: Change to dataclass with (slots=True) if/when we move minimum
         ...:     # Python version up to 3.10
         ...:     __slots__ = ["record_name", "data_count", "offset"]
         ...: 
         ...:     def __init__(
         ...:         self,
         ...:         record_name: Optional[str] = None,
         ...:         data_count: int = 0,
         ...:         offset: Optional[int] = None,
         ...:     ):
         ...:         self.record_name = record_name
         ...:         self.data_count = data_count
         ...:         self.offset = offset
         ...: 

[ins] In [5]: asizeof.asizeof(MFTRecord("foo", 0, 0))
Out[5]: 136

[ins] In [6]: asizeof.asizeof(("foo", 0, 0))
Out[6]: 144
```

volatility3/framework/plugins/windows/mftscan.py
dgmcdona: The use of `DefaultDict` in this effort was just making explicit behavior that was already implicit in the original implementation, but I'll think about possible alternatives.

volatility3/framework/plugins/windows/mftscan.py
ikelos: Ok, I'm not too keen but... how do dataclasses compare?  Happy to go with whichever works out the smallest... 5:P

dgmcdona:
Converting back to draft - I think it may be possible to avoid this dict altogether, I'll push some more updates shortly.

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -153,150 +145,104 @@
...
+                    mft_flag,
+                    renderers.NotApplicableValue(),
+                    "STANDARD_INFORMATION",
+                    conversion.wintime_to_datetime(std_information.CreationTime),
+                    conversion.wintime_to_datetime(std_information.ModifiedTime),
+                    conversion.wintime_to_datetime(std_information.UpdatedTime),
+                    conversion.wintime_to_datetime(std_information.AccessedTime),
+                    renderers.NotApplicableValue(),
+                )
+        except exceptions.InvalidAddressException:
```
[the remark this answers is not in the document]
dgmcdona: fixed

dgmcdona:
Looking deeper into the way the plugins were working got me to realize that the dictionary and that class with `__slots__` were not necessary, but it took a good bit of refactoring to remove them. A lot of things have been moved into the extension classes, especially regarding attribute access on `MFTEntry` objects so it's much more convenient to work with them now, although putting that much of what used to be in the plugin logic into the extension classes kind of makes me wish they were versioned. I've heard mention of doing something like that in the past but I don't know if it's feasible right now.

The classmethods now preserve the `objects.String` objects, and they only get converted to regular `str` in the generator methods.

ikelos:
> The classmethods now preserve the `objects.String` objects, and they only get converted to regular `str` in the generator methods.

Do we still get the same memory savings now?  Also, there is now an example of versioning extensions (not extension classes per-se though).  Check out [the tainting class](https://github.com/volatilityfoundation/volatility3/blob/6ddb0fef2907364283ebc4ddf5c878a50f25fb31/volatility3/framework/symbols/linux/utilities/tainting.py#L9) as an example...

dgmcdona:
The only thing I'm really not sure about right now was if it was a good idea to override the default `objects.StructType` constructor for `MFTEntry` in order to pull `symbol_table_name` from `kwargs`. I did it to keep that parameter out of every method signature on the extension class, but that means that they'll all raise `ValueError` if it isn't constructed with it.

dgmcdona:
> > The classmethods now preserve the `objects.String` objects, and they only get converted to regular `str` in the generator methods.
> 
> Do we still get the same memory savings now? Also, there is now an example of versioning extensions (not extension classes per-se though). Check out [the tainting class](https://github.com/volatilityfoundation/volatility3/blob/6ddb0fef2907364283ebc4ddf5c878a50f25fb31/volatility3/framework/symbols/linux/utilities/tainting.py#L9) as an example...

We should get even more memory savings, since the dictionary doesn't exist anymore, and `objects.String` is still converted to `str` before being placed in the `TreeNode`.

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -2,23 +2,152 @@
...
+from volatility3.framework import constants, exceptions, interfaces, objects
+
+vollog = logging.getLogger(__name__)
 
 
 class MFTEntry(objects.StructType):
     """This represents the base MFT Record"""
 
-    def get_signature(self) -> str:
+    def __init__(
```
ikelos, line 16: I'm not sure I'm keen on this, because it then means anyone following a pointer to an MFTEntry won't actually be able to use it because it won't have a symbol_table_name (even though it will have been pulled from, presumably, the right symbol table)?

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -2,23 +2,152 @@
...
+        context: interfaces.context.ContextInterface,
+        type_name: str,
+        object_info: interfaces.objects.ObjectInformation,
+        size: int,
+        members: Dict[str, Tuple[int, interfaces.objects.Template]],
+        **kwargs,
+    ) -> None:
+        super().__init__(context, type_name, object_info, size, members)
+
+        self._symbol_table_name = kwargs.get("symbol_table_name")
```
ikelos: Assuming the needed types are in the same table as this type, you can get this information from self.vol.type_name (there's helper functions to get the symbol table part of it, I think)?  I'd be much happier with that than requiring it as an extra argument that can cause exceptions if it hasn't been initialized...

volatility3/framework/plugins/windows/mftscan.py
```diff
@@ -22,7 +20,21 @@ class MFTScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
 
     _required_framework_version = (2, 0, 0)
```
ikelos:
```suggestion
    _required_framework_version = (2, 26, 0)
```
This needs bumping because of the extension class changes you're making.

ikelos (commented):
Feels cleaner, but as you say, loads a lot into the extension class, which I'm fine with, but we don't have a way to version those yet.  I'm also not too keen on the method for getting the symbol table, I'd extract the `vol.type_name` which should include a preceding symbol table name...

ikelos:
> The only thing I'm really not sure about right now was if it was a good idea to override the default `objects.StructType` constructor for `MFTEntry` in order to pull `symbol_table_name` from `kwargs`. I did it to keep that parameter out of every method signature on the extension class, but that means that they'll all raise `ValueError` if it isn't constructed with it.

Yeah, you should be able to pull this from `self.vol.type_name` I think...

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -5,14 +5,20 @@
...
 from typing import Dict, Iterator, List, Optional, Tuple
 
+from volatility3 import framework
 from volatility3.framework import constants, exceptions, interfaces, objects
 
 vollog = logging.getLogger(__name__)
 
 
-class MFTEntry(objects.StructType):
+class MFTEntry(objects.StructType, interfaces.configuration.VersionableInterface):
```
ikelos: I'm not sure this is a good plan, since they're applied to the symbol tables by the framework, you may get handed one and have no idea which version it is (and checking it after the fact will be problematic)...  5:S

volatility3/framework/symbols/windows/extensions/mft.py
dgmcdona: Awesome thanks! I'll figure out how to get it from there instead.

volatility3/framework/symbols/windows/extensions/mft.py
dgmcdona: Ok, reverted

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -2,33 +2,180 @@
...
+
+    def get_full_name(self) -> objects.String:
         output = self.Name.cast(
             "string", encoding="utf16", max_length=self.NameLength * 2, errors="replace"
         )
         return output
 
 
-class MFTAttribute(objects.StructType):
+class MFTAttribute(objects.StructType, interfaces.configuration.VersionableInterface):
```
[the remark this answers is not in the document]
dgmcdona: reverted

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -2,33 +2,180 @@
...
 
-from volatility3.framework import objects, constants, exceptions
+from volatility3 import framework
+from volatility3.framework import constants, exceptions, interfaces, objects
 
+vollog = logging.getLogger(__name__)
 
-class MFTEntry(objects.StructType):
+
+class MFTEntry(objects.StructType, interfaces.configuration.VersionableInterface):
```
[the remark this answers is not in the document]
dgmcdona: resolved.

volatility3/framework/symbols/windows/extensions/mft.py
```diff
@@ -2,33 +2,180 @@
...
+                yield attr
 
-class MFTFileName(objects.StructType):
+    def alternate_data_streams(self) -> Iterator["MFTAttribute"]:
+        for attr in self._data_attributes():
+            if attr.Attr_Header.NameLength != 0:
+                yield attr
+
+
+class MFTFileName(objects.StructType, interfaces.configuration.VersionableInterface):
```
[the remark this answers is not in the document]
dgmcdona: resolved

ikelos (approved):
Looks good, thanks very much!  5:D

PR #1760, opened 2025-04-07
Fix traceback in volshell's `dt()`

dgmcdona:
A `SymbolError` can occur when a type contains a pointer to an opaque
type. For example, `_EPROCESS` can have a member that points to an
`_EPROCESS_QUOTA_BLOCK`, but there is no definition for that type, so
its size and readability can't be determined.

This wraps the block in a try/except, and reports that the type has an
unknown size in the suffix if a `SymbolError` occurs.

### DIFF
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -629,7 +629,7 @@ class Volshell(interfaces.plugins.PluginInterface):
                             else:
                                 suffix = ""
                         except exceptions.SymbolError as exc:
-                            suffix = f" (pointer to {exc.symbol_name} - unknown size)"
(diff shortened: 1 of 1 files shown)


dgmcdona:
This new suffix feels a little wordy, but I haven't thought of a more concise way to phrase this yet.

ikelos:
How about `(unknown sized {exc.symbol_name})` or `(unknown size for {exc.symbol_name})`?

ikelos:
All tests passed, so just shout if you're ready for it to go in with the message as is...

ikelos (approved):

eve-mem:
@dgmcdona thanks for fixing that.

PR #1763, opened 2025-04-08
InodePages: Add output column and render when dumped

dgmcdona:
Previously, the InodePages plugin wasn't rendering treegrid columns when
the `--dump` flag was passed. This fixes that, and adds an additional
`Output File` column that displays the name of the file containing the
dumped data.


ikelos (approved):
Looks good, thanks!  5:)

PR #1764, opened 2025-04-08
Documentation: Improve the readme to make quickstart the first thing …

ikelos:
…people see

Just had someone having issues with volatility because they jumped straight in, so I've moved the quick start up, made it not dependent on the lower text (so it really is only 4 steps) and should get them up and running successfully without too much fuss.  If no one shouts I'll probably merge this at the weekend...

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,32 @@ technical and performance challenges associated with the original
 code base that became apparent over the previous 10 years. Another benefit
 of the rewrite is that Volatility 3 could be released under a custom
 license that was more aligned with the goals of the Volatility community,
-the Volatility Software License (VSL). See the 
-[LICENSE](https://www.volatilityfoundation.org/license/vsl-v1.0) file for 
+the Volatility Software License (VSL). See the
+[LICENSE](https://www.volatilityfoundation.org/license/vsl-v1.0) file for
 more details.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,10 +14,33 @@ technical and performance challenges associated with the original
...
+the Volatility Software License (VSL). See the
+[LICENSE](https://www.volatilityfoundation.org/license/vsl-v1.0) file for
 more details.
 
+## Quick Start
+
+1. Install the required dependencies:
+
+    ```shell
+    pip install --user -e.
```
eve-mem: Will this install all the dependencies, e.g. including bits needed for leehcore and s3 layers? I'm less sure on how the toml files work... I think it's just pefile but don't have a laptop to test with. 

Is it worth highlighting this is the minimum requirements and not all plugins will work? I know the various yara plugins catch people out.

ikelos:
```suggestion
    pip install --user -e ".[full]"
```

You're right, that would only install pefile, it needs the [full] after it to ensure a full installation.

eve-mem:
I think doing a full install for the quick start guide is probably wise. It might be a pain for those still using air gapped systems, but hopefully they'd be more comfortable doing a minimal install themselves.

PR #1765, opened 2025-04-08
[Parity/modules] Adjust module alignment for scanners

Abyss-W4tcher:
Hi,

This PR adjusts the module alignment value for the module scanner interface. The reworked implementation does not provide a populated list of known module addresses now:

https://github.com/volatilityfoundation/volatility3/blob/842d0e77037cc26a08379145b928df98bd60fe87/volatility3/framework/symbols/linux/utilities/modules.py#L843-L849

Consequently `validate_alignment_patterns` isn't actually validating anything right now. We observed this when mass testing `modxview` on misaligned samples (32 bytes alignment). 

To solve this issue with a minimal impact on the codebase, we figured that scanning with pointer size alignment was a good compromise even if it impacts runtime a bit.

1-byte alignment was technically valid, but most compilers and allocators should naturally do allocation aligned to the object type at least. On other note, i think that even if all modules in `validate_alignment_patterns()` passed the 64 bytes check, it was not guaranteed that a one-off allocation on a 32 bytes alignment had happened (still an unlucky scenario).

### DIFF
--- a/volatility3/framework/symbols/linux/utilities/modules.py
+++ b/volatility3/framework/symbols/linux/utilities/modules.py
@@ -474,10 +474,9 @@ class Modules(interfaces.configuration.VersionableInterface):
         Returns:
             The struct module alignment
         """
-        # FIXME: When dwarf2json/ISF supports type alignments. Read it directly from the type metadata
-        # Additionally, while 'context' and 'vmlinux_module_name' are currently unused, they will be
-        # essential for retrieving type metadata in the future.
-        return 64
+        # Not L1 cache aligned, but compiler should naturally
+        # align to the referenced type as a minimum.
+        return context.modules[vmlinux_module_name].get_type("pointer").size
 
     @classmethod
     def list_modules(
--- a/volatility3/framework/symbols/linux/utilities/modules.py
+++ b/volatility3/framework/symbols/linux/utilities/modules.py
@@ -474,8 +474,6 @@ class Modules(interfaces.configuration.VersionableInterface):
         Returns:
             The struct module alignment
         """
-        # Not L1 cache aligned, but compiler should naturally
-        # align to the referenced type as a minimum.
         return context.modules[vmlinux_module_name].get_type("pointer").size
 
     @classmethod


atcuno:
@ikelos this was a deficiency found through @dgmcdona 's comparison harness

ikelos:
I assume you wanted this applying to the release branch, given you've tagged it as parity-release, so I've retargetted it and will cherrypick it merge back into the main tree after it's merged.  It seems the two have already diverged a little, so I've retargetted it back, but it should be possible to cherry pick if that's what you'd intended...

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -474,10 +474,7 @@ def get_module_address_alignment(
         Returns:
             The struct module alignment
         """
-        # FIXME: When dwarf2json/ISF supports type alignments. Read it directly from the type metadata
-        # Additionally, while 'context' and 'vmlinux_module_name' are currently unused, they will be
-        # essential for retrieving type metadata in the future.
-        return 64
+        return context.modules[vmlinux_module_name].get_type("pointer").size
```
ikelos, line 477: I believe we have code for determining if a symbol table is 32-bit or 64-bit?  This will also return its size in bytes, so either 8 or 4 but never 64.  Is that what you'd intended?

https://github.com/volatilityfoundation/volatility3/blob/842d0e77037cc26a08379145b928df98bd60fe87/volatility3/framework/symbols/__init__.py#L338

ikelos (commented):
A question about the new pointer size return value...

volatility3/framework/symbols/linux/utilities/modules.py
Abyss-W4tcher: Yes that is what we intended 👍

Abyss-W4tcher:
Yes I think this was meant to go in 2.26.0, this won't break current scanners as 8 is still a power of two but fix the non-64 aligned modules observed through regression testing (mostly 32-bytes aligned). Modxview will otherwise flag valid modules as non-present in the scanner.

ikelos (approved):
Ooookay, as long as you're happy it'll be the right value, that's ok by me...

ikelos:
Hmmm, I still think probably it should have made the decision based on `symbol_table_is_64bit` but I'll let it slide since the outcome should be the same and if we need to change it, it should be easy to find...

PR #1767, opened 2025-04-10
Layers: Fix exception message

dgmcdona:
Updates the exception message to report the correct dependency instead
of the layer name itself. Instead of reporting the actual dependency, it
was reporting, for example 'Layer layer_name is depended upon by
layer_name'.


ikelos (approved):

ikelos:
Good spot, thanks!

PR #1768, opened 2025-04-10
Extensions: Add cache decorators as appropriate

dgmcdona:
I noticed that many plugins were creating duplicate per-process translation layers - for instance, `windows.envars.Envars` was ending up with > 30 process layers per process due to repeated calls to `get_peb()`, which calls `add_process_space()` internally.

This adds `@functools.lru_cache` to the `get_peb()` and `get_peb32()` methods on the `EPROCESS` extension, since these should only need to be created once.

Also adds `@functools.lru_cache` to `add_process_space` to enable reusing the same process address space, provided the same arguments are passed to the `add_process_space()` method.

Layers after an `Envars` run:

```
[ins] In [1]: len(list(ctx.layers))
Out[1]: 4343
```

After the fix:
```
[ins] In [3]: len(list(ctx.layers))
Out[3]: 133
```

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -732,9 +732,10 @@ class EPROCESS(generic.GenericIntelProcess, pool.ExecutiveObject):
 
         return True
 
+    @functools.lru_cache
     def add_process_layer(
         self, config_prefix: Optional[str] = None, preferred_name: Optional[str] = None
-    ):
+    ) -> str:
         """Constructs a new layer based on the process's DirectoryTableBase."""
 
         parent_layer = self._context.layers[self.vol.layer_name]
@@ -761,6 +762,7 @@ class EPROCESS(generic.GenericIntelProcess, pool.ExecutiveObject):
             self._context, dtb, config_prefix, preferred_name
         )
 
+    @functools.lru_cache
     def get_peb(self) -> interfaces.objects.ObjectInterface:
         """Constructs a PEB object"""
         if constants.BANG not in self.vol.type_name:
@@ -786,6 +788,7 @@ class EPROCESS(generic.GenericIntelProcess, pool.ExecutiveObject):
         )
         return peb
 
+    @functools.lru_cache
     def get_peb32(self) -> Optional[interfaces.objects.ObjectInterface]:
         """Constructs a PEB32 object"""
         if constants.BANG not in self.vol.type_name:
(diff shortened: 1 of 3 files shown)


dgmcdona:
Still not sure about the version bump here. The API hasn't changed, and you'll still get back a process layer name, but it isn't guaranteed to be unique anymore.

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -761,6 +762,7 @@ def add_process_layer(
             self._context, dtb, config_prefix, preferred_name
         )
 
+    @functools.lru_cache
```
ikelos: Any reason to cache `get_peb` and `get_peb32` as well as add_process_layer?  Surely caching add_process_layer gives you the same benefits?

ikelos (commented):
I don't think it needs a version bump.  It just needs to provide you the name of layer for that process, this does that, so seems fine?

I'm trying to think through issues surrounding caching this.  My only worry is we're not choosing explicitly which parameters to cache on.  Self is included, but it feels like it should be more specific just as the page table offset and arch, or something like that?  Don't think that's feasible, so it's more just how this might be turned off if people wanted to analyze live memory images one day.  I don't think it should cause a problem, because the objects caching their information anyway, so you'd need to create a new object to then add the process layer on, so should be ok...

volatility3/framework/symbols/windows/extensions/__init__.py
dgmcdona: I figured it saves a little bit of time reconstructing the objects, but that's probably negligible since I don't think either of those methods gets called in a hot loop.

volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -732,9 +732,10 @@ def is_valid(self) -> bool:
 
         return True
 
+    @functools.lru_cache
     def add_process_layer(
         self, config_prefix: Optional[str] = None, preferred_name: Optional[str] = None
-    ):
+    ) -> str:
```
ikelos, line 738: Just want to check why this one is `str` and the other is `optional[str]`?  I imagine one throws an exception and the other doesn't.  We probably want to be consistent on that, but this PR isn't the place to bring that about...

ikelos (approved):
Cool, much happier with this thanks.  A slight discrepancy between `str` and `optional[str]` but not enough to block its inclusion...

PR #1769, opened 2025-04-10
Linux pagecache.Files: Memory Usage

dgmcdona:
Converts `objects.Pointer` to `int` before storing them in the set. This
should have a substantial impact on memory, similar to those in #1758

### DIFF
--- a/volatility3/framework/plugins/linux/mountinfo.py
+++ b/volatility3/framework/plugins/linux/mountinfo.py
@@ -279,7 +279,7 @@ class MountInfo(plugins.PluginInterface):
             if not (sb_ptr and sb_ptr.is_readable()):
                 continue
 
-            if int(sb_ptr) in seen_sb_ptr:
+            if sb_ptr in seen_sb_ptr:
                 continue
             seen_sb_ptr.add(int(sb_ptr))
 
--- a/volatility3/framework/plugins/linux/mountinfo.py
+++ b/volatility3/framework/plugins/linux/mountinfo.py
@@ -279,9 +279,9 @@ class MountInfo(plugins.PluginInterface):
             if not (sb_ptr and sb_ptr.is_readable()):
                 continue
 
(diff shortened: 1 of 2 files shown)


dgmcdona:
On second thought, the calls to `int()` for `seen_dentries` may not be necessary since `dentry.vol.offset` is just an `int` already, but maybe good to leave it explicit?

ikelos (commented):
Looks good, although, please could we add a couple comments to explain why we're casting in this specific instance when we normally go to great lengths to get people not to cast (so that they don't lose useful information or litter the code with unnecessary casts)...

ikelos (approved):
Out of interest, does the `if root_inode_ptr in seen_inodes:` still work, if the entries in seen_inodes are all ints?  Just curious to know whether it does object comparison or runs the equality tests on them (and whether our equality test says they're the same).  Either way I'm happy with this, but I'll hold off until we get the answer, just in case we can trim one of the casts...

dgmcdona:
Looks like the call to `__hash__` returns the same value for both, and the equality test passes: 
```
[ins] In [1]: seen = set()

[ins] In [2]: seen.add(int(root_inode_ptr))

[ins] In [3]: seen
Out[3]: {175402173655680}

[ins] In [4]: root_inode_ptr in seen
Out[4]: True

[ins] In [5]: root_inode_ptr.__class__
Out[5]: volatility3.framework.objects.Pointer

[ins] In [6]: hash(root_inode_ptr)
Out[6]: 175402173655680

[ins] In [7]: hash(int(root_inode_ptr))
Out[7]: 175402173655680

[ins] In [8]: int(root_inode_ptr) == root_inode_ptr
Out[8]: True
```

dgmcdona:
Removed the unneeded casts in https://github.com/volatilityfoundation/volatility3/pull/1769/commits/a236c0dd40a640850e6e9408c1dfbce502c4346b

PR #1770, opened 2025-04-14
ShimcacheMem: Various bugfixes

dgmcdona:
Found a few things wrong with the shimcachemem plugin and extensions:

- WinXPSP2 symbol table: Bad references to `LARGE_INTEGER` instead of `_LARGE_INTEGER`, as well as bad offsets in the `SHIM_CACHE_ENTRY` type.
- `ShimcacheMem` plugin: Updates code to use `address_to_string` in order to deal with potentially truncated reads - this matters most in the XP entries, where the path field is an array of 520 bytes, but not all of them may be readable, and the file path is typically much shorter than the full buffer length.
- `ShimcacheMem` plugin: Fixes offset tracking. The offset was being added to the tracker before the validity of the entry was being checked, so it prevents potentially valid copies of invalid entries from being yielded from other process' address spaces in XP.
- `SHIM_CACHE_ENTRY` Extension: Fixes logic error in `exec_flag`. A couple of early returns were needed when the buffer containing the flag is not valid. This was leading to `InvalidAddressException`s being raised after the validity of the memory range had been checked, instead of returning the `renderers.UnreadableValue`.

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/shimcache.py
+++ b/volatility3/framework/symbols/windows/extensions/shimcache.py
@@ -136,7 +136,7 @@ class SHIM_CACHE_ENTRY(objects.StructType):
                 self._context,
                 self.Path.vol.layer_name,
                 self.Path.vol.offset,
-                520,
+                self.Path.vol.count,
                 encoding="utf-16le",
             )
 
--- a/volatility3/framework/symbols/windows/extensions/shimcache.py
+++ b/volatility3/framework/symbols/windows/extensions/shimcache.py
@@ -57,12 +57,14 @@ class SHIM_CACHE_ENTRY(objects.StructType):
                 blob_offset, blob_size
             ):
                 self._exec_flag = renderers.UnparsableValue()
+                return self._exec_flag
 
             raw_flag = self._context.layers[self.vol.native_layer_name].read(
                 blob_offset, blob_size
             )
             if not raw_flag:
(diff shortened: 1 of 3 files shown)


volatility3/framework/symbols/windows/extensions/shimcache.py
```diff
@@ -126,8 +132,12 @@ def file_path(self) -> Union[str, interfaces.renderers.BaseAbsentValue]:
             return self._file_path
 
         if not hasattr(self.Path, "Buffer"):
-            return self.Path.cast(
-                "string", max_length=self.Path.vol.count, encoding="utf-16le"
+            return address_to_string(
+                self._context,
+                self.Path.vol.layer_name,
+                self.Path.vol.offset,
+                520,
```
ikelos: Shouldn't this be `self.Path.size` or `self.Path.vol.count`?  Hard coding 520 just kinda sticks out as potentially allowing too much to be read (or forcing too little for future versions)...

ikelos (commented):
Ok, generally happy with this save for the hardcoded maximum size.  Pretty sure that should be derived from the field itself, in case they mess with it in the future...

volatility3/framework/symbols/windows/extensions/shimcache.py
dgmcdona: Good catch, fixed

ikelos (approved):

PR #1766, opened 2025-04-10
Windows ThrdScan: Thread filtering, type-hints and tracebacks

dgmcdona:
During testing, thread enumeration with `ThrdScan`/`Threads` was resulting in tracebacks due to `AttributeError`s and `InvalidAddressExceptions`.

This PR:
1. [Typing] Adds a `typing.NamedTuple` type for encapsulating the data returned in `gather_thread_info`, which should make that API easier and more intuitive for people using that method directly.
2. [Typing] Narrows type-hints to their correct types in extension class methods. What was previously type-hinted as returning `interfaces.objects.ObjectInterface` is now typed as returning `windows_extensions.EPROCESS`. This will help with editor support for autocompletion during development, as well as type-checking.
3. [Windows Constants] Adds a constant value to the windows constants module for `MAX_PID`.
4. [ThrdScan] - Add filtering based on PID values reported in the thread. It seems reasonable to assume that an invalid `_ETHREAD.Cid.UniqueProcess` value (out of range or non-divisible by 4) means that this thread is not valid.
5. [ThrdScan] - Remove restriction on `owner_proc.InheritedFromUniqueProcessId != 4` when getting VAD mappings for thread start addresses. This change resulted in correct mappings in threads for `smss.exe`, since it is a direct child of the system process (PID 412 is `smss.exe` here):
```
149,150c149,150
< 0xe001ff810040	412	416	0x7ffe49c05a40	-	0x7ff73b721190	-	2024-10-09 20:20:28.000000 UTC	1600-11-21 06:53:40.000000 UTC
< 0xe001ffb7d080	412	440	0x7ffe49c05a40	-	0x7ffe49bd1a40	-	2024-10-09 20:20:31.000000 UTC	1600-11-21 06:53:41.000000 UTC
---
> 0xe001ff810040	412	416	0x7ffe49c05a40	\Windows\System32\ntdll.dll	0x7ff73b721190	\Windows\System32\smss.exe	2024-10-09 20:20:28.000000 UTC	1600-11-21 06:53:40.000000 UTC
> 0xe001ffb7d080	412	440	0x7ffe49c05a40	\Windows\System32\ntdll.dll	0x7ffe49bd1a40	\Windows\System32\ntdll.dll	2024-10-09 20:20:31.000000 UTC	1600-11-21 06:53:41.000000 UTC
```
[...]

Issue #1761, opened 2025-04-08
Bleeding Edge Kernel versions drop module_sect_attr


atcuno:
This is a struct we currently rely on to extract kernel modules. This structure has been there since the beginning of the kernel but was removed with this commit:

https://github.com/torvalds/linux/commit/34f5ec0f8252e2ba83fa5da05d56b6fdb0ebef59

A slack user reported the bug from a preview/bleeding edge Ubuntu kernel, so this will hit mainstream kernels soon enough. I will investigate this week so its fixed for parity.

ikelos:
I hate to say it, but this is a new feature ("support recent kernels") rather than a fix for something that currently works incorrectly, and I don't think it's enough to break the code freeze.  It doesn't yet affect mainstream kernels (you said it was bleeding edge), the existing code isn't *broken* it just can't handle all cases at the moment (which will always be the case in a moving world) and we have to cut at some point.  Feel free to get it fixed in develop, and it'll get released in July or August, whenever our next release is due to be rolled.

ikelos:
It's not a parity release issue since I'm pretty sure volatility 2 will not handle this case.

atcuno:
The updated method for module attributes (the reason for this ticket) is in released Ubuntu preview kernels now, which means over the next couple weeks they will go into stable kernels being used in production. It is also not a silent issue, it fully backtraces all of Vol3 for any Linux analysis as we set the extension for `module_sect_attr` on load. So even running something like linux.pslist on the kernel (unrelated to module extraction) on an affected symbol table will backtrace the whole framework. This is a bug fix type ticket which is supposed to be the purpose of the gap time between now and the parity release. It really does need to be fixed for parity.

ikelos:
Well, I'll take a look at the PR once it comes out to see how invasive it is, but targetting a fix for something that comes out in two weeks, is still pre-empting the future, so I'm absolutely middle of the line on this and it could go either way...

ikelos:
(It also proves we need to be careful about whether we add class overrides or optional class overrides when we add extensions).

atcuno:
Yes, but two weeks from now is during the bug release period. Also, that would put 2-3 months of Linux analysis on up to date kernels backtracking and we just be telling people to use devleop for 2 months instead of the release.

[...]

PR #1771, opened 2025-04-14
Bugfix: Linux/Windows process traversal

dgmcdona:
Both of these process listings were missing entries due to the lack of backwards traversal (`__iter__` for the `LIST_ENTRY`/`list_head` extension classes only goes forward). This updates both plugins to traverse the list backwards as well as forwards, tracking seen offsets to avoid duplicate entries.

### DIFF
--- a/volatility3/framework/plugins/linux/pslist.py
+++ b/volatility3/framework/plugins/linux/pslist.py
@@ -262,17 +262,27 @@ class PsList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
         init_task = vmlinux.object_from_symbol(symbol_name="init_task")
 
         # Note that the init_task itself is not yielded, since "ps" also never shows it.
-        for task in init_task.tasks:
-            if not task.is_valid():
-                continue
+        seen = set()
+        for forward in (True, False):
+            for task in init_task.tasks.to_list(
+                symbol_type=init_task.vol.type_name,
+                member="tasks",
+                forward=forward,
+            ):
+                if task.vol.offset in seen:
+                    continue
+                seen.add(task.vol.offset)
 
-            if filter_func(task):
-                continue
+                if not task.is_valid():
+                    continue
 
-            yield task
+                if filter_func(task):
+                    continue
 
-            if include_threads:
-                yield from task.get_threads()
+                yield task
+
+                if include_threads:
+                    yield from task.get_threads()
 
     def run(self):
         pids = self.config.get("pid")
--- a/volatility3/framework/plugins/linux/pslist.py
+++ b/volatility3/framework/plugins/linux/pslist.py
@@ -34,7 +34,7 @@ class PsList(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface):
     """Lists the processes present in a particular linux memory image."""
 
     _required_framework_version = (2, 13, 0)
-    _version = (4, 1, 0)
+    _version = (4, 1, 1)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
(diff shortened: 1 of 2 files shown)


ikelos (approved):
Looks good, thanks.  5:)

atcuno:
@ikelos is this PR and the thrdscan backtraces PR setup to be merged into the release branch? You said that bugfixes found from the harness would be put into the release branch, which both of those were/are. Also, we would be releasing really bad code in these cases (threads backtracing all over, missing processes) without it, which defeats the point of a big parity release push since we would immediately tell everyone to just use develop instead.

ikelos:
This one, no, it's an issue that's been present in the code for a while, and it's not critical it just misses a few results.

If #1761 is the threadscan one, then I was expecting a PR to land over the weekend, but we still don't have a PR for it, so... it's looking less likely for that one.  If there's another threadscan one then I haven't seen anything that looked major go past since the code freeze, lemme know which one you were thinking of...

Lastly, there's a lot of good reasons for a code freeze:
* It's a clear, well known deadline (that's been known about since January) which gives everyone lots of time to get in anything they need.
* We'll never get all the bugs out, and we'll never release if we wait until there are no bugs before we release.  Bugs are part of software
* If we do a Columbo and add "just one more thing" all the time, we'll find another thing and another.  If we don't push the release, we're adding last minute fixes which can actually break the code worse than the problem they're trying to fix.  There needs to be a period of calm where generally everything works how it should, not perfect, but ok.
* We can break the freeze period, but then we'll need that period of calm again (2-3 weeks,down from 6 weeks because of the testing harness).  If you feel that strongly about these bugs, we probably shouldn't release until May...
* If you don't think the release process is working well, we can discuss how to improve it for the next release, but this release is too close to be changing it now.

atcuno:
thrdscan, I meant this: https://github.com/volatilityfoundation/volatility3/pull/1766

Don't want to push this release cycle back.

atcuno:
Sent the comment too early by accident. This pslist ones actually make it where some samples report 5 processes or less going forward but the full set going backwards. Having this one + #1766 into the releases definitely falls under getting critical bugs found from the harness into the release branch. This doesn't change the current release cycle and these PRs are being tested under the full harness which runs all the samples per OS.

ikelos:
So... pslist returns results, accurate results even, it just misses some.  I think our definitions of critical differ.

This is the problem: pslist now returns more results, so we rerun the harness (+X days) and maybe one of the other plugins that depends on pslist is now processing processes it hadn't seen before, and now it starts to break on some of them, so we fix that (+Y days) and then rerun the test harness (+X days) and now a new bug comes out of the woodwork (+...).  Every change has the chance of making things worse in a way that we didn't see coming, and each one of those then needs fixing.  It has to stop somewhere.

From the issues you've pointed out, volatility will run fine against 95% of samples (windows, linux and mac) so there is no crisis and the issues are not critical.  The tool will work fine, and they'll already be fixed by the time the next release comes out.

The problem with a code freeze is it's like insurance.  It feels like a waste of time until the moment it was massively helpful and prevented the roll out of broken software that then needs to be fixed under rushed conditions, creating even more trouble.  These just aren't big enough issues to risk us giving up our insurance and going into panic mode trying to get everything done.

atcuno:
Ok I don't think we will agree on what should go in overall, so pslist can stay out, but can the threads one I pointed to go in since the threads related plugins were backtracing on quite a few samples?

https://github.com/volatilityfoundation/volatility3/pull/1766

ikelos:
There's a significant number of changes in #1766, if you can make it more targetted to fix *just* the problem causing the backtracking that has a chance of getting in (you can target the release branch directly if you want), but that PR by itself shifts a lot of different things at once and it's not clear how to isolate the specific thing that's supposed to be being fixed...

atcuno:
@dgmcdona - see @ikelos previous comment here. Can you make a smaller version of that PR targetted to just the fixes related to VADs filtering and PID/PPID of 4 fixes?

dgmcdona:
Made a minimal thrdscan bugfix PR in #1777

PR #1773, opened 2025-04-15
Issue #1761: handle new bin_attribute format for module sections

### LINKED ISSUE #1761: Bleeding Edge Kernel versions drop module_sect_attr

### LINKED ISSUE #1883: Linux samples without `module_sect_attr` symbol
**Describe the bug**
I have built a volatility profile for a Target NixOS 25.05 (Kernel 6.15.7) with dwarf2json.
I am able to run pslist, psscan and bash history plugins on volatility3 v2.11.0. This version is working fine, except for Malfind and PsTree.

So I have decided to switch to a newer version v2.26.2.
With this version, I have this bug message with the same symbols generated by dwarf2json: Unable to validate the plugin requirements: ['plugins.PsList.kernel.layer_name', 'plugins.PsList.kernel.symbol_table_name']

I have acquired memory dump with three different versions of AVML: 0.14.0, 0.15.0 and 0.16.0

**Context**
Volatility Version:  2.26.2
Operating System:  NixOS 25.05
Python Version:  Python 3.12.11
Suspected Operating System:  NixOS 25.05
Command:  vol --clear-cache -vvvv -s symbols/ -f memory_avml_0.dmp linux.pslist

**To Reproduce**

1. Use command : vol --clear-cache -vvvv -s symbols/ -f memory_avml_0.dmp linux.pslist
2. See error

**Expected behavior**
When I run the command pslist, I should see the process list.

**Example output**
Command: vol --clear-cache -vvvv -s symbols/ -f memory_avml_0.dmp linux.pslist

On version v2.26.2

```
Volatility 3 Framework 2.26.2
INFO     volatility3.cli: Volatility plugins path: ['/nix/store/qv0jybpalspswcl8jfk4fpx6x1z9khjx-volatility3-2.26.2/lib/python3.12/site-packages/volatility3/plugins', '/nix/store/qv0jybpalspswcl8jfk4fpx6x1z9khjx-volatility3-2.26.2/lib/python3.12/site-packages/volatility3/framework/plugins']
INFO     volatility3.cli: Volatility symbols path: ['/home/[REDACTED]/projects/memory_analysis/ram_test/symbols', '/nix/store/qv0jybpalspswcl8jfk4fpx6x1z9khjx-volatility3-2.26.2/lib/python3.12/site-packages/volatility3/symbols', '/nix/store/qv0jybpalspswcl8jfk4fpx6x1z9khjx-volatility3-2.26.2/lib/python3.12/site-packages/volatility3/framework/symbols']
DEBUG    volatility3.plugins.yarascan: Using yara-python module
INFO     volatility3.framework.automagic: Detected a linux category plugin
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList
INFO     volatility3.framework.automagic: Running automagic: SymbolCacheMagic
DETAIL 2 volatility3.framework.automagic.symbol_cache: Identified file:///home/[REDACTED]/projects/memory_analysis/ram_test/symbols/6.15.7-kernel.json as b'Linux version 6.15.7 (nixbld@localhost) (gcc (GCC) 14.2.1 20250322, GNU ld (GNU Binutils) 2.44) #1-NixOS SMP PREEMPT_DYNAMIC Thu Jul 17 16:44:05 UTC 2025\n\x00'
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using Elf64Stacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using XenCoreDumpStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using AVMLStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using QemuStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using LimeStacker
DETAIL 2 volatility3.framework.automagic.stacker: Stacked LimeLayer using LimeStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using Elf64Stacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using XenCoreDumpStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using AVMLStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using QemuStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using WindowsCrashDumpStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using VmwareStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using LinuxIntelVMCOREINFOStacker
DEBUG    volatility3.framework.automagic.linux: Identified banner: b'Linux version 6.15.7 (nixbld@localhost) (gcc (GCC) 14.2.1 20250322, GNU ld (GNU Binutils) 2.44) #1-NixOS SMP PREEMPT_DYNAMIC Thu Jul 17 16:44:05 UTC 2025\n\x00'
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using LinuxIntelStacker
DEBUG    volatility3.framework.automagic.linux: Identified banner: b'Linux version 6.15.7 (nixbld@localhost) (gcc (GCC) 14.2.1 20250322, GNU ld (GNU Binutils) 2.44) #1-NixOS SMP PREEMPT_DYNAMIC Thu Jul 17 16:44:05 UTC 2025\n\x00'
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: TypeError - Layer is not the required Architecture: LimeLayer
DETAIL 1 volatility3.framework.configuration.requirements: TypeError - Layer is not the required Architecture: FileLayer
DEBUG    volatility3.framework.automagic.stacker: physical_layer maximum_address: 33751351419
DEBUG    volatility3.framework.automagic.stacker: Stacked layers: ['LimeLayer', 'FileLayer']
INFO     volatility3.framework.automagic: Running automagic: LinuxSymbolFinder
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name

Unsatisfied requirement plugins.PsList.kernel.layer_name: 
Unsatisfied requirement plugins.PsList.kernel.symbol_table_name: 

A translation layer requirement was not fulfilled.  Please verify that:
	A file was provided to create this layer (by -f, --single-location or by config)
	The file exists and is readable
	The file is a valid memory image and was acquired cleanly

A symbol table requirement was not fulfilled.  Please verify that:
	The associated translation layer requirement was fulfilled
	You have the correct symbol file for the requirement
	The symbol file is under the correct directory or zip file
	The symbol file is named appropriately or contains the correct banner

Unable to validate the plugin requirements: ['plugins.PsList.kernel.layer_name', 'plugins.PsList.kernel.symbol_table_name']
```

On v2.11.0 (the good versionà

```
Volatility 3 Framework 2.11.0
INFO     volatility3.cli: Volatility plugins path: ['/nix/store/07drhm6d3yflz6xvszwx9q3lcqwv49c7-volatility3-2.11.0/lib/python3.12/site-packages/volatility3/plugins', '/nix/store/07drhm6d3yflz6xvszwx9q3lcqwv49c7-volatility3-2.11.0/lib/python3.12/site-packages/volatility3/framework/plugins']
INFO     volatility3.cli: Volatility symbols path: ['/home/[REDACTED]/projects/memory_analysis/ram_test/symbols', '/nix/store/07drhm6d3yflz6xvszwx9q3lcqwv49c7-volatility3-2.11.0/lib/python3.12/site-packages/volatility3/symbols', '/nix/store/07drhm6d3yflz6xvszwx9q3lcqwv49c7-volatility3-2.11.0/lib/python3.12/site-packages/volatility3/framework/symbols']
DEBUG    volatility3.plugins.yarascan: Using yara-python module
INFO     volatility3.framework.automagic: Detected a linux category plugin
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList
INFO     volatility3.framework.automagic: Running automagic: SymbolCacheMagic
DETAIL 2 volatility3.framework.automagic.symbol_cache: Identified file:///home/[REDACTED]/projects/memory_analysis/ram_test/symbols/6.15.7-kernel.json as b'Linux version 6.15.7 (nixbld@localhost) (gcc (GCC) 14.2.1 20250322, GNU ld (GNU Binutils) 2.44) #1-NixOS SMP PREEMPT_DYNAMIC Thu Jul 17 16:44:05 UTC 2025\n\x00'
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using Elf64Stacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using XenCoreDumpStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using AVMLStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using QemuStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using LimeStacker
DETAIL 2 volatility3.framework.automagic.stacker: Stacked LimeLayer using LimeStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using Elf64Stacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using XenCoreDumpStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using AVMLStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using QemuStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using WindowsCrashDumpStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using VmwareStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using LinuxIntelStacker
DEBUG    volatility3.framework.automagic.linux: Identified banner: b'Linux version 6.15.7 (nixbld@localhost) (gcc (GCC) 14.2.1 20250322, GNU ld (GNU Binutils) 2.44) #1-NixOS SMP PREEMPT_DYNAMIC Thu Jul 17 16:44:05 UTC 2025\n\x00'
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!assoc_array_ptr
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!netns_ipvs
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!mtd_info
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!can_pkg_stats
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!can_rcv_lists_stats
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!can_dev_rcv_lists
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!mpls_route
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!sctp_mib
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!smc_stats_rsn
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!smc_stats
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!mpls_dev
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!tipc_bearer
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!udp_tunnel_nic
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!phylink
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!cfg80211_conn
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!cfg80211_cached_keys
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!cfg80211_cqm_config
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!phy_led_trigger
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!pse_control
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!phy_package_shared
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!sfp
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!smc_hashinfo
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!dsa_8021q_context
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!rfkill
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!uapi_definition
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!hw_stats_device_data
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!rdma_restrack_root
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!cfg80211_internal_bss
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!ib_port
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!ib_gid_table
DEBUG    volatility3.framework.symbols: Unresolved reference: LintelStacker1!ib_pkey_cache
DEBUG    volatility3.framework.automagic.linux: Linux ASLR shift values determined: physical 5e44c3000 virtual a600000
DEBUG    volatility3.framework.automagic.linux: DTB was found at: 0x5e74e7000
DETAIL 2 volatility3.framework.automagic.stacker: Stacked IntelLayer using LinuxIntelStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using Elf64Stacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using XenCoreDumpStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using AVMLStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using QemuStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using WindowsCrashDumpStacker
DETAIL 2 volatility3.framework.automagic.stacker: Attempting to stack using VmwareStacker
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name.memory_layer
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name.memory_layer.base_layer
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList
DEBUG    volatility3.framework.automagic.stacker: physical_layer maximum_address: 33751351419
DEBUG    volatility3.framework.automagic.stacker: Stacked layers: ['IntelLayer', 'LimeLayer', 'FileLayer']
INFO     volatility3.framework.automagic: Running automagic: LinuxSymbolFinder
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DEBUG    volatility3.framework.automagic.symbol_finder: Identified banner: b'Linux version 6.15.7 (nixbld@localhost) (gcc (GCC) 14.2.1 20250322, GNU ld (GNU Binutils) 2.44) #1-NixOS SMP PREEMPT_DYNAMIC Thu Jul 17 16:44:05 UTC 2025\n\x00'
DEBUG    volatility3.framework.automagic.symbol_finder: Using symbol library: file:///home/[REDACTED]/projects/memory_analysis/ram_test/symbols/6.15.7-kernel.json

OFFSET (V)	PID	TID	PPID	COMM	CREATION TIME	File output
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!assoc_array_ptr
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!netns_ipvs
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!mtd_info
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!can_pkg_stats
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!can_rcv_lists_stats
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!can_dev_rcv_lists
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!mpls_route
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!sctp_mib
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!smc_stats_rsn
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!smc_stats
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!mpls_dev
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!tipc_bearer
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!udp_tunnel_nic
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!phylink
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!cfg80211_conn
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!cfg80211_cached_keys
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!cfg80211_cqm_config
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!phy_led_trigger
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!pse_control
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!phy_package_shared
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!sfp
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!smc_hashinfo
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!dsa_8021q_context
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!rfkill
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!uapi_definition
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!hw_stats_device_data
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!rdma_restrack_root
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!cfg80211_internal_bss
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!ib_port
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!ib_gid_table
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!ib_pkey_cache

0x8beac4b9cd00	1	1	0	systemd	2025-09-29 09:57:01.121289 UTC	Disabled
0x8beac4b98000	2	2	0	kthreadd	2025-09-29 09:57:01.121289 UTC	Disabled
0x8beac4b9e040	3	3	2	pool_workqueue_	2025-09-29 09:57:01.121289 UTC	Disabled
0x8beac4b99340	4	4	2	kworker/R-rcu_g	2025-09-29 09:57:01.121289 UTC	Disabled
0x8beac4b9a680	5	5	2	kworker/R-sync_	2025-09-29 09:57:01.121289 UTC	Disabled
0x8beac4b9b9c0	6	6	2	kworker/R-kvfre	2025-09-29 09:57:01.121289 UTC	Disabled
0x8beac4bb6040	7	7	2	kworker/R-slub_	2025-09-29 09:57:01.121289 UTC	Disabled
....
```

Abyss-W4tcher:
Hi,

This PR suggests a way to handle the removal of `module_sect_attr` structure to the benefit of `bin_attribute`. 

The `nsections` member of module's `sect_attrs` was also removed in 6.14-rc1:

https://github.com/torvalds/linux/commit/d8959b947a8dfab1047c6fd5e982808f65717bfe

The kernel-way to count the number of sections is to iterate over the NULL-terminated array of pointers pointed by `module.sect_attrs.grp.bin_attrs`. Using `module.sect_attrs.grp.attrs` was proven to dereference a NULL pointer, but I think this wasn't spotted before because:

- get_sections() wasn't called anywhere (and still isn't)
- `nsections` would have still been accessible anyways

On another note, it seems that `module_extract` and `class module` have duplicated code, which could be unified directly in `class module` and called with `my_mod.get_sections()` for example.

---

Tested on an Ubuntu 25.04 sample:

```
Linux version 6.14.0-11-generic (buildd@lcy02-amd64-043) (x86_64-linux-gnu-gcc-14 (Ubuntu 14.2.0-19ubuntu1) 14.2.0, GNU ld (GNU Binutils for Ubuntu) 2.44) #11-Ubuntu SMP PREEMPT_DYNAMIC Mon Mar 17 12:39:41 UTC 2025 (Ubuntu 6.14.0-11.11-generic 6.14.0-rc7)
```

Fixes #1761.
Fixes #1883.

### DIFF
--- a/volatility3/framework/objects/utility.py
+++ b/volatility3/framework/objects/utility.py
@@ -257,8 +257,8 @@ def array_of_pointers(
 def dynamically_sized_array_of_pointers(
     context: interfaces.context.ContextInterface,
     array: interfaces.objects.ObjectInterface,
-    iterator_guard_value: int,
     subtype: Union[str, interfaces.objects.Template],
+    iterator_guard_value: int,
     stop_value: int = 0,
     stop_on_invalid_pointers: bool = True,
 ) -> interfaces.objects.ObjectInterface:
--- a/volatility3/framework/objects/utility.py
+++ b/volatility3/framework/objects/utility.py
@@ -250,3 +250,52 @@ def array_of_pointers(
     ).clone()
     subtype_pointer.update_vol(subtype=subtype)
     return array.cast("array", count=count, subtype=subtype_pointer)
+
+
+def dynamically_sized_array_of_pointers(
+    context: interfaces.context.ContextInterface,
+    layer_name: str,
+    symbol_table_name: str,
+    array_offset: int,
+    stop_value: int = 0,
+    iterator_guard_value: int = None,
+    stop_on_invalid_pointers: bool = True,
+) -> interfaces.objects.ObjectInterface:
+    """Iterates over a dynamically sized array of pointers (e.g. NULL-terminated).
+
+        Args:
+            context: The context on which to operate.
+            layer_name: The layer on which the array should be constructed.
+            symbol_table_name: The symbol table to use to construct object types.
+            array_offset: The array offset within the layer, from which to start iterating.
+            stop_value: Stop value used to determine when to terminate iteration once it is encountered. Defaults to 0 (NULL-terminated arrays).
+            iterator_guard_value: Stop iterating when the iterator index is greater than this value. This is an extra-safety against smearing.
+            stop_on_invalid_pointers: Determines whether to stop iterating or not when an invalid pointer is encountered. This can be useful for arrays
+    that are known to have smeared entries before the end.
+    """
+    pointer_type = context.symbol_space.get_type(
+        symbol_table_name + constants.BANG + "pointer"
+    )
+    entry = context.object(
+        pointer_type,
+        layer_name=layer_name,
+        offset=array_offset,
+    )
+    i = 0
+    array = []
+    # entry and entry.vol.offset aren't the same thing, as
+    # - entry is naturally represented by the address that the pointer refers to;
+    # - entry.vol.offset is the offset at which the pointer lives.
+    while entry != stop_value:
+        if (not entry.is_readable() and stop_on_invalid_pointers) or (
+            iterator_guard_value is not None and i >= iterator_guard_value
+        ):
+            break
+        array.append(entry)
+        entry = context.object(
+            pointer_type,
+            layer_name=layer_name,
+            offset=entry.vol.offset + pointer_type.size,
+        )
+        i += 1
+    return array
--- a/volatility3/framework/objects/utility.py
+++ b/volatility3/framework/objects/utility.py
@@ -279,14 +279,27 @@ def dynamically_sized_array_of_pointers(
             An array of pointer objects
     """
     new_count = 0
-    for entry in array_of_pointers(
-        array=array, count=iterator_guard_value, subtype=subtype, context=context
-    ):
-        # "entry" is naturally represented by the address that the pointer refers to
-        if (entry == stop_value) or (
-            not entry.is_readable() and stop_on_invalid_pointers
-        ):
+    sym_table_name = array.get_symbol_table_name()
+    sym_table = context.symbol_space[sym_table_name]
+    ptr_size = sym_table.get_type("pointer").size
+    layer_name = array.vol.layer_name
+
+    offset = array.vol.offset
+    entry = None
+    while entry != stop_value and new_count < iterator_guard_value:
+        try:
+            entry = context.object(
+                sym_table_name + constants.BANG + "pointer",
+                offset=offset,
+                layer_name=layer_name,
+            )
+        except exceptions.InvalidAddressException:
             break
+
+        if not entry.is_readable() and stop_on_invalid_pointers:
+            break
+
+        offset += ptr_size
         new_count += 1
     else:
         vollog.log(
--- a/volatility3/framework/objects/utility.py
+++ b/volatility3/framework/objects/utility.py
@@ -4,7 +4,7 @@
 
 import re
 
-from typing import Optional, Union
+from typing import List, Optional, Union
 
 from volatility3.framework import interfaces, objects, constants, exceptions
 
@@ -260,7 +260,7 @@ def dynamically_sized_array_of_pointers(
     stop_value: int = 0,
     iterator_guard_value: int = None,
     stop_on_invalid_pointers: bool = True,
-) -> interfaces.objects.ObjectInterface:
+) -> List[interfaces.objects.ObjectInterface]:
     """Iterates over a dynamically sized array of pointers (e.g. NULL-terminated).
 
         Args:
@@ -272,6 +272,9 @@ def dynamically_sized_array_of_pointers(
             iterator_guard_value: Stop iterating when the iterator index is greater than this value. This is an extra-safety against smearing.
             stop_on_invalid_pointers: Determines whether to stop iterating or not when an invalid pointer is encountered. This can be useful for arrays
     that are known to have smeared entries before the end.
+
+        Returns:
+            An array of pointer objects
     """
     pointer_type = context.symbol_space.get_type(
         symbol_table_name + constants.BANG + "pointer"
--- a/volatility3/framework/objects/utility.py
+++ b/volatility3/framework/objects/utility.py
@@ -3,10 +3,12 @@
 #
 
 import re
-
-from typing import List, Optional, Union
+import logging
+from typing import Optional, Union
 
 from volatility3.framework import interfaces, objects, constants, exceptions
+
+vollog = logging.getLogger(__name__)
 
 
 def rol(value: int, count: int, max_bits: int = 64) -> int:
@@ -254,51 +256,46 @@ def array_of_pointers(
 
 def dynamically_sized_array_of_pointers(
     context: interfaces.context.ContextInterface,
-    layer_name: str,
-    symbol_table_name: str,
-    array_offset: int,
+    array: interfaces.objects.ObjectInterface,
+    iterator_guard_value: int,
+    subtype: Union[str, interfaces.objects.Template],
     stop_value: int = 0,
-    iterator_guard_value: int = None,
     stop_on_invalid_pointers: bool = True,
-) -> List[interfaces.objects.ObjectInterface]:
+) -> interfaces.objects.ObjectInterface:
     """Iterates over a dynamically sized array of pointers (e.g. NULL-terminated).
+    Array iteration should always be performed with an arbitrary guard value as maximum size,
+    to prevent running forever in case something unexpected happens.
 
         Args:
             context: The context on which to operate.
-            layer_name: The layer on which the array should be constructed.
-            symbol_table_name: The symbol table to use to construct object types.
-            array_offset: The array offset within the layer, from which to start iterating.
-            stop_value: Stop value used to determine when to terminate iteration once it is encountered. Defaults to 0 (NULL-terminated arrays).
+            array: The object to cast to an array.
             iterator_guard_value: Stop iterating when the iterator index is greater than this value. This is an extra-safety against smearing.
+            subtype: The subtype of the array's pointers.
+            stop_value: Stop value used to determine when to terminate iteration once it is encountered. Defaults to 0 (NULL-terminated arrays).
             stop_on_invalid_pointers: Determines whether to stop iterating or not when an invalid pointer is encountered. This can be useful for arrays
     that are known to have smeared entries before the end.
 
         Returns:
             An array of pointer objects
     """
-    pointer_type = context.symbol_space.get_type(
-        symbol_table_name + constants.BANG + "pointer"
-    )
-    entry = context.object(
-        pointer_type,
-        layer_name=layer_name,
-        offset=array_offset,
-    )
-    i = 0
-    array = []
-    # entry and entry.vol.offset aren't the same thing, as
-    # - entry is naturally represented by the address that the pointer refers to;
-    # - entry.vol.offset is the offset at which the pointer lives.
-    while entry != stop_value:
-        if (not entry.is_readable() and stop_on_invalid_pointers) or (
-            iterator_guard_value is not None and i >= iterator_guard_value
+    new_count = 0
+    for entry in array_of_pointers(
+        array=array, count=iterator_guard_value, subtype=subtype, context=context
+    ):
+        # "entry" is naturally represented by the address that the pointer refers to
+        if (entry == stop_value) or (
+            not entry.is_readable() and stop_on_invalid_pointers
         ):
             break
-        array.append(entry)
-        entry = context.object(
-            pointer_type,
-            layer_name=layer_name,
-            offset=entry.vol.offset + pointer_type.size,
+        new_count += 1
+    else:
+        vollog.log(
+            constants.LOGLEVEL_V,
+            f"""Iterator guard value {iterator_guard_value} reached while iterating over array at offset {array.vol.offset:#x}.\
+ This means that there is a bug (e.g. smearing) with this array, or that it may contain valid entries past the iterator guard value.""",
         )
-        i += 1
-    return array
+
+    # Leverage the "Array" object instead of returning a Python list
+    return array_of_pointers(
+        array=array, count=new_count, subtype=subtype, context=context
+    )
--- a/volatility3/framework/plugins/linux/module_extract.py
+++ b/volatility3/framework/plugins/linux/module_extract.py
@@ -4,8 +4,8 @@
 import logging
 from typing import List
 
+import volatility3.framework.symbols.linux.utilities.modules as linux_utilities_modules
 from volatility3 import framework
-import volatility3.framework.symbols.linux.utilities.module_extract as linux_utilities_module_extract
 from volatility3.framework import interfaces, renderers
 from volatility3.framework.configuration import requirements
 from volatility3.framework.renderers import format_hints
@@ -17,7 +17,7 @@ vollog = logging.getLogger(__name__)
 class ModuleExtract(interfaces.plugins.PluginInterface):
     """Recreates an ELF file from a specific address in the kernel"""
 
-    _version = (1, 0, 0)
+    _version = (1, 0, 1)
     _required_framework_version = (2, 25, 0)
 
     framework.require_interface_version(*_required_framework_version)
@@ -37,9 +37,9 @@ class ModuleExtract(interfaces.plugins.PluginInterface):
                 optional=False,
             ),
             requirements.VersionRequirement(
-                name="linux_utilities_module_extract",
-                version=(1, 0, 0),
-                component=linux_utilities_module_extract.ModuleExtract,
+                name="linux_utilities_modules_module_extract",
+                version=(1, 0, 2),
+                component=linux_utilities_modules.ModuleExtract,
             ),
         ]
 
@@ -58,7 +58,7 @@ class ModuleExtract(interfaces.plugins.PluginInterface):
 
         module = kernel.object(object_type="module", offset=base_address, absolute=True)
 
-        elf_data = linux_utilities_module_extract.ModuleExtract.extract_module(
+        elf_data = linux_utilities_modules.ModuleExtract.extract_module(
             self.context, self.config["kernel"], module
         )
         if not elf_data:
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -210,8 +210,8 @@ class module(generic.GenericIntelProcess):
         bin_attrs_list = utility.dynamically_sized_array_of_pointers(
             context=self._context,
             array=arr_offset_ptr.dereference(),
-            iterator_guard_value=100,
             subtype=self.get_symbol_table_name() + constants.BANG + arr_subtype,
+            iterator_guard_value=100,
         )
         return len(bin_attrs_list)
 
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -179,16 +179,27 @@ class module(generic.GenericIntelProcess):
             return None
 
     def _get_sect_count(self, grp: interfaces.objects.ObjectInterface) -> int:
-        """Try to determine the number of valid sections"""
+        """Try to determine the number of valid sections. Support for kernels > 6.14-rc1.
+
+        Resources:
+            - https://github.com/torvalds/linux/commit/d8959b947a8dfab1047c6fd5e982808f65717bfe
+            - https://github.com/torvalds/linux/commit/e0349c46cb4fbbb507fa34476bd70f9c82bad359
+        """
+
+        if grp.has_member("bin_attrs"):
+            arr_offset = grp.bin_attrs
+        else:
+            arr_offset = grp.attrs
+
         symbol_table_name = self.get_symbol_table_name()
         arr = self._context.object(
             symbol_table_name + constants.BANG + "array",
             layer_name=self.vol.layer_name,
-            offset=grp.attrs,
+            offset=arr_offset,
             subtype=self._context.symbol_space.get_type(
                 symbol_table_name + constants.BANG + "pointer"
             ),
-            count=25,
+            count=50,
         )
 
         idx = 0
@@ -198,6 +209,7 @@ class module(generic.GenericIntelProcess):
 
     @functools.cached_property
     def number_of_sections(self) -> int:
+        # Dropped in 6.14-rc1: d8959b947a8dfab1047c6fd5e982808f65717bfe
         if self.sect_attrs.has_member("nsections"):
             return self.sect_attrs.nsections
 
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -191,20 +191,27 @@ class module(generic.GenericIntelProcess):
         else:
             arr_offset = grp.attrs
 
-        symbol_table_name = self.get_symbol_table_name()
-        arr = self._context.object(
-            symbol_table_name + constants.BANG + "array",
-            layer_name=self.vol.layer_name,
-            offset=arr_offset,
-            subtype=self._context.symbol_space.get_type(
-                symbol_table_name + constants.BANG + "pointer"
-            ),
-            count=50,
-        )
+        if not arr_offset.is_readable():
+            vollog.log(
+                constants.LOGLEVEL_V,
+                f"Cannot dereference the pointer to the NULL-terminated list of binary attributes for module at offset {self.vol.offset:#x}",
+            )
+            return 0
 
+        entry = arr_offset.dereference()
+        symbol_table_name = self.get_symbol_table_name()
         idx = 0
-        while arr[idx] and arr[idx].is_readable():
-            idx = idx + 1
+        while entry.is_readable():
+            idx += 1
+            entry = self._context.object(
+                symbol_table_name + constants.BANG + "pointer",
+                layer_name=self.vol.layer_name,
+                offset=entry.vol.offset
+                + self._context.symbol_space.get_type(
+                    symbol_table_name + constants.BANG + "pointer"
+                ).size,
+            )
+
         return idx
 
     @functools.cached_property
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -3188,6 +3188,7 @@ class module_sect_attr(objects.StructType):
 
         return None
 
+
 class bin_attribute(objects.StructType):
     def get_name(self) -> Optional[str]:
         """
@@ -3197,7 +3198,9 @@ class bin_attribute(objects.StructType):
             try:
                 return utility.pointer_to_string(self.attr.name, count=32)
             except exceptions.InvalidAddressException:
-                vollog.debug(f"Invalid attr name for bin_attribute at {self.vol.offset:#x}")
+                vollog.debug(
+                    f"Invalid attr name for bin_attribute at {self.vol.offset:#x}"
+                )
                 return None
 
         return None
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -3221,3 +3221,10 @@ class bin_attribute(objects.StructType):
                 return None
 
         return None
+
+    @property
+    def address(self) -> int:
+        """Equivalent to module_sect_attr.address:
+        - https://github.com/torvalds/linux/commit/4b2c11e4aaf7e3d7fd9ce8e5995a32ff5e27d74f
+        """
+        return self.private
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -3172,7 +3172,9 @@ class module_sect_attr(objects.StructType):
         """
         if hasattr(self, "battr"):
             try:
-                return utility.pointer_to_string(self.battr.attr.name, count=32)
+                return utility.pointer_to_string(
+                    self.battr.attr.name, count=linux_constants.ATTRIBUTE_NAME_MAX_SIZE
+                )
             except exceptions.InvalidAddressException:
                 # if battr is present then its name attribute needs to be valid
                 vollog.debug(f"Invalid battr name for section at {self.vol.offset:#x}")
@@ -3180,14 +3182,18 @@ class module_sect_attr(objects.StructType):
 
         elif self.name.vol.type_name == "array":
             try:
-                return utility.array_to_string(self.name, count=32)
+                return utility.array_to_string(
+                    self.name, count=linux_constants.ATTRIBUTE_NAME_MAX_SIZE
+                )
             except exceptions.InvalidAddressException:
                 # specifically do not return here to give `mattr` a chance
                 vollog.debug(f"Invalid direct name for section at {self.vol.offset:#x}")
 
         elif self.name.vol.type_name == "pointer":
             try:
-                return utility.pointer_to_string(self.name, count=32)
+                return utility.pointer_to_string(
+                    self.name, count=linux_constants.ATTRIBUTE_NAME_MAX_SIZE
+                )
             except exceptions.InvalidAddressException:
                 # specifically do not return here to give `mattr` a chance
                 vollog.debug(
@@ -3197,7 +3203,9 @@ class module_sect_attr(objects.StructType):
         # if everything else failed...
         if hasattr(self, "mattr"):
             try:
-                return utility.pointer_to_string(self.mattr.attr.name, count=32)
+                return utility.pointer_to_string(
+                    self.mattr.attr.name, count=linux_constants.ATTRIBUTE_NAME_MAX_SIZE
+                )
             except exceptions.InvalidAddressException:
                 vollog.debug(
                     f"Unresolvable name for for section at {self.vol.offset:#x}"
@@ -3213,7 +3221,9 @@ class bin_attribute(objects.StructType):
         """
         if hasattr(self, "attr"):
             try:
-                return utility.pointer_to_string(self.attr.name, count=32)
+                return utility.pointer_to_string(
+                    self.attr.name, count=linux_constants.ATTRIBUTE_NAME_MAX_SIZE
+                )
             except exceptions.InvalidAddressException:
                 vollog.debug(
                     f"Invalid attr name for bin_attribute at {self.vol.offset:#x}"
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -198,21 +198,17 @@ class module(generic.GenericIntelProcess):
             )
             return 0
 
-        entry = arr_offset.dereference()
-        symbol_table_name = self.get_symbol_table_name()
-        idx = 0
-        while entry.is_readable():
-            idx += 1
-            entry = self._context.object(
-                symbol_table_name + constants.BANG + "pointer",
-                layer_name=self.vol.layer_name,
-                offset=entry.vol.offset
-                + self._context.symbol_space.get_type(
-                    symbol_table_name + constants.BANG + "pointer"
-                ).size,
-            )
-
-        return idx
+        # We chose 1000 as an arbitrary guard value against
+        # extreme cases of smearing.
+        # See PR #1773 for more information.
+        bin_attrs_list = utility.dynamically_sized_array_of_pointers(
+            context=self._context,
+            layer_name=self.vol.layer_name,
+            symbol_table_name=self.get_symbol_table_name(),
+            array_offset=arr_offset.dereference().vol.offset,
+            iterator_guard_value=1000,
+        )
+        return len(bin_attrs_list)
 
     @functools.cached_property
     def number_of_sections(self) -> int:
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -3227,18 +3227,13 @@ class bin_attribute(objects.StructType):
         """
         Performs extraction of the bin_attribute name
         """
-        if hasattr(self, "attr"):
-            try:
-                return utility.pointer_to_string(
-                    self.attr.name, count=linux_constants.ATTRIBUTE_NAME_MAX_SIZE
-                )
-            except exceptions.InvalidAddressException:
-                vollog.debug(
-                    f"Invalid attr name for bin_attribute at {self.vol.offset:#x}"
-                )
-                return None
-
-        return None
+        try:
+            return utility.pointer_to_string(
+                self.attr.name, count=linux_constants.ATTRIBUTE_NAME_MAX_SIZE
+            )
+        except exceptions.InvalidAddressException:
+            vollog.debug(f"Invalid attr name for bin_attribute at {self.vol.offset:#x}")
+            return None
 
     @property
     def address(self) -> int:
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -211,9 +211,7 @@ class module(generic.GenericIntelProcess):
             symbol_table_name + constants.BANG + "array",
             layer_name=self.vol.layer_name,
             offset=self.sect_attrs.attrs.vol.offset,
-            subtype=self._context.symbol_space.get_type(
-                symbol_table_name + constants.BANG + "module_sect_attr"
-            ),
+            subtype=self.sect_attrs.attrs.vol.subtype,
             count=self.number_of_sections,
         )
 
@@ -3187,5 +3185,19 @@ class module_sect_attr(objects.StructType):
                 vollog.debug(
                     f"Unresolvable name for for section at {self.vol.offset:#x}"
                 )
+
+        return None
+
+class bin_attribute(objects.StructType):
+    def get_name(self) -> Optional[str]:
+        """
+        Performs extraction of the bin_attribute name
+        """
+        if hasattr(self, "attr"):
+            try:
+                return utility.pointer_to_string(self.attr.name, count=32)
+            except exceptions.InvalidAddressException:
+                vollog.debug(f"Invalid attr name for bin_attribute at {self.vol.offset:#x}")
+                return None
 
         return None
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -217,6 +217,11 @@ class module(generic.GenericIntelProcess):
 
     def get_sections(self) -> Iterable[interfaces.objects.ObjectInterface]:
         """Get a list of section attributes for the given module."""
+        if self.number_of_sections == 0:
+            vollog.debug(
+                f"Invalid number of sections ({self.number_of_sections}) for module at offset {self.vol.offset:#x}"
+            )
+            return []
 
         symbol_table_name = self.get_symbol_table_name()
         arr = self._context.object(
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -187,26 +187,31 @@ class module(generic.GenericIntelProcess):
         """
 
         if grp.has_member("bin_attrs"):
-            arr_offset = grp.bin_attrs
+            arr_offset_ptr = grp.bin_attrs
+            arr_subtype = "bin_attribute"
         else:
-            arr_offset = grp.attrs
+            arr_offset_ptr = grp.attrs
+            arr_subtype = "attribute"
 
-        if not arr_offset.is_readable():
+        if not arr_offset_ptr.is_readable():
             vollog.log(
                 constants.LOGLEVEL_V,
                 f"Cannot dereference the pointer to the NULL-terminated list of binary attributes for module at offset {self.vol.offset:#x}",
             )
             return 0
 
-        # We chose 1000 as an arbitrary guard value against
-        # extreme cases of smearing.
+        # We chose 100 as an arbitrary guard value to prevent
+        # looping forever in extreme cases, and because 100 is not expected
+        # to be a valid number of sections. If that still happens,
+        # Vol3 module processing will indicate that it is missing information
+        # with the following message:
+        # "Unable to reconstruct the ELF for module struct at"
         # See PR #1773 for more information.
         bin_attrs_list = utility.dynamically_sized_array_of_pointers(
             context=self._context,
-            layer_name=self.vol.layer_name,
-            symbol_table_name=self.get_symbol_table_name(),
-            array_offset=arr_offset.dereference().vol.offset,
-            iterator_guard_value=1000,
+            array=arr_offset_ptr.dereference(),
+            iterator_guard_value=100,
+            subtype=self.get_symbol_table_name() + constants.BANG + arr_subtype,
         )
         return len(bin_attrs_list)
 
--- a/volatility3/framework/symbols/linux/utilities/module_extract.py
+++ b/volatility3/framework/symbols/linux/utilities/module_extract.py
@@ -263,9 +263,9 @@ class ModuleExtract(interfaces.configuration.VersionableInterface):
         for index, section in enumerate(module.get_sections()):
             # Extra sanity check, to prevent OOM on heavily smeared samples at line
             # "size = next_address - address"
-            if (
-                not modules_addr_min
-                <= kernel_layer.address_mask & section.address
+            if not (
+                modules_addr_min
+                <= section.address & kernel_layer.address_mask
                 < modules_addr_max
             ):
                 continue
--- a/volatility3/framework/symbols/linux/utilities/module_extract.py
+++ b/volatility3/framework/symbols/linux/utilities/module_extract.py
@@ -10,12 +10,14 @@ from typing import (
     Dict,
 )
 
+import volatility3.framework.symbols.linux.utilities.modules as linux_utilities_modules
 from volatility3 import framework
 from volatility3.framework import (
     interfaces,
     exceptions,
     symbols,
 )
+from volatility3.framework.configuration import requirements
 from volatility3.framework.constants import linux as linux_constants
 from volatility3.framework.symbols.linux import extensions
 
@@ -38,10 +40,20 @@ vollog = logging.getLogger(__name__)
 class ModuleExtract(interfaces.configuration.VersionableInterface):
     """Extracts Linux kernel module structures into an analyzable ELF file"""
 
-    _version = (1, 0, 1)
+    _version = (1, 0, 2)
     _required_framework_version = (2, 25, 0)
 
     framework.require_interface_version(*_required_framework_version)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.VersionRequirement(
+                name="linux_utilities_modules",
+                component=linux_utilities_modules.Modules,
+                version=(3, 0, 2),
+            ),
+        ]
 
     @classmethod
     def _find_section(
@@ -237,16 +249,31 @@ class ModuleExtract(interfaces.configuration.VersionableInterface):
         The data of .strtab is read directly off the module structure and not its section
         as the section from the original module has no meaning after loading as the kernel does not reference it.
         """
+        kernel = context.modules[vmlinux_name]
+        kernel_layer = context.layers[kernel.layer_name]
+        modules_addr_min, modules_addr_max = (
+            linux_utilities_modules.Modules.get_modules_memory_boundaries(
+                context, vmlinux_name
+            )
+        )
+        modules_addr_min &= kernel_layer.address_mask
+        modules_addr_max &= kernel_layer.address_mask
+
         original_sections = {}
         for index, section in enumerate(module.get_sections()):
+            # Extra sanity check, to prevent OOM on heavily smeared samples at line
+            # "size = next_address - address"
+            if (
+                not modules_addr_min
+                <= kernel_layer.address_mask & section.address
+                < modules_addr_max
+            ):
+                continue
             name = section.get_name()
             original_sections[section.address] = name
 
         if not original_sections:
             return None
-
-        kernel = context.modules[vmlinux_name]
-        kernel_layer = context.layers[kernel.layer_name]
 
         if symbols.symbol_table_is_64bit(context, kernel.symbol_table_name):
             sym_type = "Elf64_Sym"
--- a/volatility3/framework/symbols/linux/utilities/module_extract.py
+++ b/volatility3/framework/symbols/linux/utilities/module_extract.py
@@ -10,14 +10,12 @@ from typing import (
     Dict,
 )
 
-import volatility3.framework.symbols.linux.utilities.modules as linux_utilities_modules
 from volatility3 import framework
 from volatility3.framework import (
     interfaces,
     exceptions,
     symbols,
 )
-from volatility3.framework.configuration import requirements
 from volatility3.framework.constants import linux as linux_constants
 from volatility3.framework.symbols.linux import extensions
 
@@ -40,20 +38,10 @@ vollog = logging.getLogger(__name__)
 class ModuleExtract(interfaces.configuration.VersionableInterface):
     """Extracts Linux kernel module structures into an analyzable ELF file"""
 
-    _version = (1, 0, 2)
+    _version = (1, 0, 1)
     _required_framework_version = (2, 25, 0)
 
     framework.require_interface_version(*_required_framework_version)
-
-    @classmethod
-    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
-        return [
-            requirements.VersionRequirement(
-                name="linux_utilities_modules",
-                component=linux_utilities_modules.Modules,
-                version=(3, 0, 2),
-            ),
-        ]
 
     @classmethod
     def _find_section(
@@ -249,31 +237,16 @@ class ModuleExtract(interfaces.configuration.VersionableInterface):
         The data of .strtab is read directly off the module structure and not its section
         as the section from the original module has no meaning after loading as the kernel does not reference it.
         """
-        kernel = context.modules[vmlinux_name]
-        kernel_layer = context.layers[kernel.layer_name]
-        modules_addr_min, modules_addr_max = (
-            linux_utilities_modules.Modules.get_modules_memory_boundaries(
-                context, vmlinux_name
-            )
-        )
-        modules_addr_min &= kernel_layer.address_mask
-        modules_addr_max &= kernel_layer.address_mask
-
         original_sections = {}
         for index, section in enumerate(module.get_sections()):
-            # Extra sanity check, to prevent OOM on heavily smeared samples at line
-            # "size = next_address - address"
-            if not (
-                modules_addr_min
-                <= section.address & kernel_layer.address_mask
-                < modules_addr_max
-            ):
-                continue
             name = section.get_name()
             original_sections[section.address] = name
 
         if not original_sections:
             return None
+
+        kernel = context.modules[vmlinux_name]
+        kernel_layer = context.layers[kernel.layer_name]
 
         if symbols.symbol_table_is_64bit(context, kernel.symbol_table_name):
             sym_type = "Elf64_Sym"
--- a/volatility3/framework/symbols/linux/utilities/module_extract.py
+++ b/volatility3/framework/symbols/linux/utilities/module_extract.py
@@ -15,6 +15,7 @@ from volatility3.framework import (
     interfaces,
     exceptions,
     symbols,
+    deprecation
 )
 from volatility3.framework.constants import linux as linux_constants
 from volatility3.framework.symbols.linux import extensions
@@ -34,7 +35,12 @@ vollog = logging.getLogger(__name__)
 
 # ModuleExtract.extract_module is the entry point and only visible method for plugins
 
-
+# See PR #1773
+@deprecation.renamed_class(
+    deprecated_class_name="ModuleExtract",
+    removal_date="2026-06-01",
+    message="volatility3.framework.symbols.linux.utilities.module_extract.ModuleExtract is to be deprecated. Use volatility3.framework.symbols.linux.utilities.modules.ModuleExtract instead.",
+)
 class ModuleExtract(interfaces.configuration.VersionableInterface):
     """Extracts Linux kernel module structures into an analyzable ELF file"""
 
--- a/volatility3/framework/symbols/linux/utilities/module_extract.py
+++ b/volatility3/framework/symbols/linux/utilities/module_extract.py
@@ -38,7 +38,7 @@ vollog = logging.getLogger(__name__)
 class ModuleExtract(interfaces.configuration.VersionableInterface):
     """Extracts Linux kernel module structures into an analyzable ELF file"""
 
-    _version = (1, 0, 2)
+    _version = (1, 0, 1)
     _required_framework_version = (2, 25, 0)
 
     framework.require_interface_version(*_required_framework_version)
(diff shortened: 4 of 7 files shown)


volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -180,16 +180,27 @@ def get_name(self) -> Optional[str]:
...
         arr = self._context.object(
             symbol_table_name + constants.BANG + "array",
             layer_name=self.vol.layer_name,
-            offset=grp.attrs,
+            offset=arr_offset,
             subtype=self._context.symbol_space.get_type(
                 symbol_table_name + constants.BANG + "pointer"
             ),
-            count=25,
+            count=50,
```
ikelos: I'm always a little nervous about random hard coded numbers.  Is this defined anywhere we could find it, is 50 enough, does that number come from source code or how was 50 decided upon?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -3165,3 +3175,20 @@ def get_name(self) -> Optional[str]:
...
+
+
+class bin_attribute(objects.StructType):
+    def get_name(self) -> Optional[str]:
+        """
+        Performs extraction of the bin_attribute name
+        """
+        if hasattr(self, "attr"):
+            try:
+                return utility.pointer_to_string(self.attr.name, count=32)
```
ikelos: Again... how did we decide on 32?  Wouldn't this be a size of an array or a length value or something?

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -60,9 +60,14 @@ def _get_module_section_count(
         count = 0
 
         try:
+            if grp.has_member("bin_attrs"):
```
ikelos: I see what you mean about the duplication...  Can we rolls these two together somehow?  We should at least file a bug as a reminder to get this discrepancy resolved...

ikelos (commented):
Looks good, thanks.  A couple of hard coded values that I'd like to check can't be programattically determined, and a question about the code duplication, but that's not a show stopper...

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: I don't have the answer to this question, for now I just duplicated code from modules.py, given the context.

As this is a NULL terminated array of pointers, it would be optimal to iterate over it that way, or even extend current utilities for it like `array_of_pointers(null_terminated=True)`.

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: This code was duplicated from `module_sect_attr.get_name()` to make the transition smoother for now. 

I think this points out a more global issue, which is that some places are using hardcoded values without referencing kernel source, constants or at least explaining why it is done that way.

volatility3/framework/symbols/linux/utilities/module_extract.py
Abyss-W4tcher: I opened an issue:

https://github.com/volatilityfoundation/volatility3/issues/1774

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: I think we have a number of string functions that do null terminate, but they should all always have a maximum value (otherwise we could just keep running forever).  I'm just trying to find out what the appropriate maximum value should be and why (if anyone knows)

ikelos:
Ok, so... how should we proceed?  Should we just stick with the constants, open a ticket to note down where they are, or try and find out where they should come from and fix them before merging this?

Abyss-W4tcher:
Maybe this refinement can happen in the consolidation PR, as I think it'll contextualize better. Given the current schedule I think it'll also leave us more time to fix it more consistently?

ikelos:
Well, there's no rush?  This is going into develop, so if it needs more time, then by all means take it.  Just lemme know when you're happy for it to go in...

Abyss-W4tcher:
I didn't want to push too much changes into this one, given @atcuno request to include it in 2.26.1. I think the goal was to make the mandatory changes for 6.14-rc1+ analysis without an early crash, which is already distributed in Ubuntu 25.04 for example.

Happy to make the other changes we talked about in another PR if that is ok for you, while #1761's discussion is not concluded?

ikelos:
2.26.1 isn't the release branch (that's 2.26.0)?  If you meant to target the release branch, please change the PR.  There's a fair number of changes in this though, so I'd be happier just to put it in develop?

ikelos:
Again, I don't see support for only-just published kernels as needing to be rushed into this version rather than waiting for the next one in 3-4 months.  It doesn't feel critical (most of volatility works fine for the majority of images), and we're only two weeks from release (meaning if this messed things up in some way, we'd be rushing to get it fixed).  I'm open to reasons why this *has* to be in this version, but they'd have to be pretty good reasons...

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: So... rather than copying that, shouldn't we do it the right way here (and maybe add a comment from where it was copied to investigate/fix it)?  Just feels like adding more bad code because it was copied is still adding bad code...

Abyss-W4tcher: I think that is tricky, as `attribute` is not specific to module section names for example, so even if we find the maximum section name for a module there's no guarantee it will apply to other `attribute` users:

```c
struct attribute {
	const char		*name;
```


The kernel uses a NULL-terminated string check, but here to account for smear we still have to put a max value as you said. However right now I haven't found any constant in the source code.

ikelos: That's fine, then we just need to put a comment to the effect that "we've chosen 32 as the maxmium because...".  32 seems small to me, I wonder why it wouldn't be 255, but there might be a good reason for choosing that specifically?  Where did that number come from?  We can use it as long as we justify it (and saying it came from somewhere else in our code isn't a good justification).  5;P

Abyss-W4tcher: In 5.9-rc1+, the Linux kernel limits the READ size of a section bin_attribute name to MODULE_SECT_READ_SIZE:

- https://elixir.bootlin.com/linux/v6.15-rc4/source/kernel/module/sysfs.c#L106
- https://github.com/torvalds/linux/commit/11990a5bd7e558e9203c1070fc52fb6f0488e75b

However, the raw section name loaded from the .ko ELF can in theory be thousands of characters, 
and unless we do a NULL terminated search we can't set a perfect value.

We can up it to 255 if a vendor decided to make weirdly long section names for its third party module, but if a rootkit decides to mess with the section names then the fixed length method won't be of any help.

Abyss-W4tcher: Ah, we could still use the `bin_attribute.size` member of the most recent kernel, as this one tells us if a maximum size was set or not. For the previous version, we can use the 255 constant.

Abyss-W4tcher: I have made the changes, at least it should be easier to spot where to update things if needed.

volatility3/framework/symbols/linux/extensions/__init__.py
ikelos: Any progress on where this value came from?  Ideally we'd avoid this the same was as we avoid the other one (or at least comment where the size of the value comes from).

Abyss-W4tcher: I made a change that makes it a NULL-terminated iteration (just as [the structure's design](https://elixir.bootlin.com/linux/v5.7.9/source/include/linux/sysfs.h#L81)). It is safer to make a NULL-terminated array of pointers iteration, as if it encounters a value that cannot be dereferenced the loop will break immediately.

ikelos:
So... I'm not sure I'm keen on us just looping forever, and doing all that maths manually isn't a good plan.  I'd almost sooner you set a ridiculously large count than reconstruct everything the array is supposed to do just so it can be unbounded?  All I really wanted was a rationale for why that number had been picked?  We can even just say "we think this is as many as there will ever be", but we need to say why it was chosen in case someone discovers a case where it's too small and wants to know why we stop there?

Abyss-W4tcher:
I see, but using a while loop is intended by design for this NULL-terminated array. Using a for loop and constructing an array with an arbitrary size beforehand does not really make sense either. 

Still, I can add a guard value (while ... and i < GUARD_VALUE) for extreme scenarios (millions of unrelated valid pointers appended at the end of this array). I think that is more coherent this way than the other way around?

ikelos:
If you're implementing a guard statement, it has the same effect as the array max size?  The original while loop was just fine, it only needed a comment to say where the guard size came from.  Reimplementing loops is a good way to slip up (maybe not this time, but in general) and create a hard to spot corner case where something goes wrong.  Using existing machinery to do it is far better (and also means if we develop ways of doing the same task but using less memory, we only have to implement the fix in one place rather than every plugin that decides to rewrite looping methods we already have).

The old method also stopped not just on NULL, but on any unreachable pointers.  I'm not sure if that's better or worse, but it's certainly far less code to have to reason about?

Abyss-W4tcher:
This method also stops on unreachable pointers (which includes NULL of course) through `<ptr>.is_readable()` ? Here is how the kernel does it since nsections was removed:

image

As you can see it iterates over the array while a valid pointer is found. The existing machinery sometimes does not fit exactly what we want to do, which is the case here. Using a while loop also prevents instantiating an array the size of the safe guard we decided to use even if we end up iterating over 20% of it. The code size is actually equivalent as well (we only do one `context.object` call)?

We already use this kind of mechanism, for example here:

https://github.com/volatilityfoundation/volatility3/blob/57b28e5fa294330e2419539534446b4b9c3d60bb/volatility3/framework/plugins/linux/tracing/ftrace.py#L126-L128

because we cannot always come up with a maximum iteration value. In the ftrace case this could be leveraged by rootkits to hide tracing entries past our safe guard. 

In summary my goal is simply to align more closely with the kernel code, while adding our own safe guards (e.g. against smearing) 👍 

edit: Here is the guard value we could use:

```python
        # We chose 100 as an arbitrary guard value against smearing,
        # as it is highly unlikely that any module will ever have 
        # this many sections.
        while entry.is_readable() and idx < 100:
```

ikelos:
The code you've suggested to swap to is still several lines long and bogged down in object construction for something that already appears to be a pointer?

If you truly think there shouldn't be a guard statement, then I'll go along with it (the `is_readable` is a reasonable check).  Could we check whether simply doing `cast` would be better easier than all the potentially error-prone type find/maths (the repeated type finding will be extremely inefficient and could be done once before the loop at a bare minimum)?

It really doesn't feel like it needs as much code as has been used to achieve what I think you're trying to achieve and I'd prefer it be readable rather than mired in object creation code without good cause...

ikelos:
> The existing machinery sometimes does not fit exactly what we want to do, which is the case here.

I don't understand how it doesn't fit here.  Could you clarify why this deviates from simply iterating over an array (which, incidentally, *isn't* what `ftrace` does, `ftrace` follows an attribute of each item).

Abyss-W4tcher:
> > The existing machinery sometimes does not fit exactly what we want to do, which is the case here.
> 
> I don't understand how it doesn't fit here. Could you clarify why this deviates from simply iterating over an array (which, incidentally, _isn't_ what `ftrace` does, `ftrace` follows an attribute of each item).

We are still iterating over an array, the ""issue"" here is that we are not supposed to know the size of this array beforehand. The size we set is more of a "smearing guard" value, which is not exactly the same as the array size. As discussed, there is no place in the kernel where "50" is declared as this exact array size.

I'm happy with setting a guard size, even if this does not cover the case where obscure rootkits setup hundreds of sections just to mess with us and store their payloads in sections skipped because of the guard size.

> Could we check whether simply doing cast would be better easier than all the potentially error-prone type find/maths (the repeated type finding will be extremely inefficient and could be done once before the loop at a bare minimum)? 

`entry.vol.offset + ptr_size` is only an integer, not a vol object, thus I don't think we can use `cast` on it?

True, I can assign the pointer size before the loop 👍.  I think the pointer arithmetics are still performed the same by Vol3 under the hood?

By the way, I checked the Ubuntu 6.14.0 sample I have and some modules have 38 sections, which would already have made the previous `25` constant miss a lot of them. 100 should be good, but that's part of the dangers of using a guard value :/.

ikelos:
The only reason we're doing the maths to figure out the next one is because... it isn't in an array?  That's precisely what arrays are there to do!  If it's just what the guard value should be, then we can choose a ridiculously big one and say we expect the invalid pointer to happen before the guard is hit.  As long as we document it so it's not a random 25 (and someone has to scrabble through the code to figure out whether that's accurate or an invented value in the middle of a dense codebase) then it'll be fine!  5;P

It probably is what we do under the hood, but the point is to have it done by the same code so that if it needs changing/fixing/making more efficient, it can be done once for everything?  Given the old code seemed to work, and the only issue I had was where the value 25/50 came from, it seems like it's easier to understand as long as we document *why* 25/50 was chosen.  That's literally all we need to do for this.  It doesn't sounds like all the extra object creation complexity changing the outcome in any way...

Abyss-W4tcher:
If we use a ridiculously big value, then the `context.object("array")` will still instantiate a very large array in memory even if 99.9% of the time it will iterate over no more than 5% of it.

I don't think that an identical outcome always justifies keeping the core code identical, currently we are not iterating over this array the right way. 

I suggested a more accurate way to do this, that might come helpful for other use cases and is quite modular. Tell me what you think about this design, which in my opinion achieves an even better level of abstraction and readability than the previous implementation. The resources in the docstring also help a lot to see why it is done that way.

Abyss-W4tcher:
I know it might sound overkill when we could indeed just put one comment and keep the old way, but if we can achieve greater precision and coherency with the kernel I think it is worth the effort.

ikelos:
Well, ok, whatever changes you needed to achieve grater precision and coherency, go with them, but please use the array mechanism rather than manually constructing the objects?  There should be functionally no difference between a guarded version of your object construction, and a vast array, except the code that acts on it will be much shorter/easier to read and understand and that all the code we use for handling arrays would live in one place.  I don't see how your solution achieves better precision or coherency, but I feel those differences can be applied to the array creation method, rather than a strange bespoke guarded manual object iterator thing...

I feel like I'm missing something causing you to keep pushing for this technique, rather than the other one, so please try to explain it to me?  From my perspective, the code feels functionally identical (and indeed, was the original version you submitted, so you seemed happy with it too just a few days ago)?

Abyss-W4tcher:
Yeah so originally as the output was the same I didn't realize it wasn't the correct way of iterating.

When we iterate over arrays, we know their start and end beforehand. The end is most of the time directly stored in the symbols type definition or stored in another kernel symbol.

However, NULL-terminated arrays sizes cannot be predicted by the kernel at compile time. Indeed, the kernel cannot predict how many sections an arbitrary module loaded by a user will have for example (at least for > 6.14.0).

In doing so, we have to check that the first member/pointer is not NULL, and **only** then move to the next entry and repeat the process. Using already existing array mechanisms would not really apply here, hence why I added this new `dynamically_sized_array_of_pointers` helper. 

Please note that the "guard mechanism" is only a sanity mechanism inherent to memory forensics to account for smearing, but not something that we should conceptually see as a part of the original kernel iteration mechanism.

You can also notice that the original code (check the PR diff) does a while loop right after the `context.object("array")` call, which is exactly trying to find this NULL entry. One in the other, I think this new version actually makes it easier to read.

Hopefully that makes sense, sorry if it seems a bit far fetched, but we don't want to confuse arrays that have predictable sizes with ones that don't.

ikelos:
So... I think I get what you're saying, but I fundamentally disagree that:

```
        while arr[idx] and arr[idx].is_readable():
            idx = idx + 1
```
is more difficult to read than
```
        while entry.is_readable():
            idx += 1
            entry = self._context.object(
                symbol_table_name + constants.BANG + "pointer",
                layer_name=self.vol.layer_name,
                offset=entry.vol.offset
                + self._context.symbol_space.get_type(
                    symbol_table_name + constants.BANG + "pointer"
                ).size,
            )
```
?

I understand we don't know the end of the array, and I'm fine with us using a ridiculously large number just as a safety guard, as long as we comment it (so we don't end up in a pages long discussion about the number 50 again in the future).  5;P  I think the way that we deal with strings is to set a maximum limit, then read upwards to the first null and discard the rest.  Doing the same with an array seems like it would make sense too (and not confuse people).  If you'd prefer, we could add an additional method to the utility class called something like `null_terminated_array` that you hand in a starting value and it spits you out entries until it hits null.  BUT, it should do this using the existing array machinery we've already got, rather than being identical (duplicated) code that does the maths itself.  I'm open to whether it takes an array and just increases the count to a huge arbitrary number (parameterized but defaulting to something enourmous) or whether it takes in a start address and a type and builds the array itself.

Would that satisfy your concerns about people a) misunderstanding it and b) thinking it looks complicated?  At long as we're using the existing code for traversing what is an array (be it bounded or unbounded) then I'm fine with it living off in a shared location so it can be reused, we just really, really, really shouldn't be reinventing the wheel with 9 lines of duplicate custom code to calculate array members simply because they may become rubbish after a null value...

Abyss-W4tcher:
Oh, I think there was a misunderstanding as I pushed this helper and other commits this afternoon.

Please check the latest diff and tell me if it fits the framework better?

Abyss-W4tcher:
Based on that, as I said earlier even in the new helper I don't think we can rely on existing vol3 array mechanisms for unbound arrays? The current ones always expect an end value?

ikelos:
You're right, I hadn't seen it, but that's not using an array type at all, and it forces the target type to be a pointer but also doesn't allow the user to specify the subtype it points to.

I really think we can rely on the array type.  Simply create or be handed the array, then use the [count setter](https://github.com/volatilityfoundation/volatility3/blob/686057d8060d63a0be5dc30b8e33564dfaffba05/volatility3/framework/objects/__init__.py#L734) to make it a huge size, run through it until you find the null entry (or hit the guard) and then set the count to that and return the array.  It will not read entries you don't ask it to, so as long as you go through the indices in order (and stop if there's an address exception) there should be no problem at all.

Can you see a reason why it wouldn't work, or shouldn't be used?  So far you seem to suggest you don't think it's right to (although it's now hidden in a helper function) and... I think that's about it?  Is there another technical reason that using the array code wouldn't work properly?

ikelos:
If we're handed the array, it should already be primed with the correct subtype so that would be all we'd need.  We could also optionally accept a guard value if people know a specific maximum, but it's not necessary we could just choose an exceedingly high one (and log if it ever gets hit).  We probably want to allow `InvalidAddressException`s through to the caller to deal with.  The other option would be taking a start offset and a subtype and constructing an array, but in either instance the caller would have the chance to specify the type of the array (and if it's a pointer, the pointer's subtype if they wanted).  That should make the function much more applicable to all circumstances.

ikelos:
If we wanted we could instead yield as the items are traversed.  I haven't thought about which might be better?

Abyss-W4tcher:
Yeah so using `utility.array_of_pointers` could work, because I realized I was maybe wrongfully thinking it was returning a list and not an iterator.

In the second case then yes it wouldn't go past the NULL entry, but in the first case it would construct an array the size of our large value before we can iterate over it.

I can't check right now, but does your last comment implies that it isn't an iterator?

ikelos:
> In the second case then yes it wouldn't go past the NULL entry, but in the first case it would construct an array the size of our large value before we can iterate over it.

You can construct an array of any size and (just like a structure of any size) it won't actually try and read anything of any size until it's requested).  [Here](https://github.com/volatilityfoundation/volatility3/blob/686057d8060d63a0be5dc30b8e33564dfaffba05/volatility3/framework/objects/__init__.py#L800) you can see `__getitem__` when the array is indexed.  Until it's indexed, nothing is constructed, no extra memory is used.  You can have a 10000000 entry array and have it take up no more space than a 1 entry array until you access its elements...

I hadn't been intending it to be an interator, I thought it would return the array allowing the user to pick and choose the entries they wanted (in case it a was a long list and they knew they only needed the 200th entry)?  Since we have to iterate once to find the maximum, it may not be more efficient in this case, so that's a design decision I'm open to decide.  We just have to figure out what the we think the user will want most...

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -10,12 +10,14 @@
     Dict,
 )
 
+import volatility3.framework.symbols.linux.utilities.modules as linux_utilities_modules
```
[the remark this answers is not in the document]
Abyss-W4tcher: Argh, unfortunately the module address space helper is in this file

Abyss-W4tcher:
Thanks for the details, I think it basically is the same as an iterator in the end. I initially thought that a thousand Pointer objects would be created but if we iterate over an Array object it does not consume extra memory.

I made adjustments to the new helper, and also added sanity checks in the `ModuleExtract` class, as it wasn't validating section addresses. You can notice that I use `array_of_pointers` twice inside of `dynamically_sized_array_of_pointers` to fully leverage the `Array` object even in the return value. A bit of code duplication but it avoids any live re-sizing of the first array.

volatility3/framework/symbols/linux/utilities/module_extract.py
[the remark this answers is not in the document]
Abyss-W4tcher: REMINDER: Apply the patch back, similarly to the two commits [add section address sanity check](https://github.com/volatilityfoundation/volatility3/commit/4009c6b801c6ef502bb367c434290de7e7333788) and [slight readability adjustment](https://github.com/volatilityfoundation/volatility3/commit/d51981783ce7f22ea00ce57aca2041260020cca9):

https://github.com/volatilityfoundation/volatility3/commits/d51981783ce7f22ea00ce57aca2041260020cca9/volatility3/framework/symbols/linux/utilities/module_extract.py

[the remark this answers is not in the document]
Abyss-W4tcher: Please do not close this conversation yet. I'll re-open it if CodeQL resolves it.

ikelos:
Ok, so as I see it, there's two ways to (eventually) split this:

* Extraction is a feature built-on top of the general module functionality, therefore the module functionality doesn't do dumping itself and callers that want it have to do it themselves or...
* Extraction is a core part of the Module functionality and it should all live in the same class/file

Roughly happy with either.  I do get a little concerned about features that try to be all singing and dancing, but since it makes sense to do the dumping in the middle of the generation, it might be the best move is to deprecate and shift the module_extract code into the module_display_plugin code.  Having said that, module_display_plugin definitely feels like it's overreaching by doing more than just displaying...  5:S

Abyss-W4tcher:
So in the end I think you can merge #1801 and I'll handle potential conflicts here (there shouldn't be any) as well as a new design to fix the circular import?

ikelos:
@Abyss-W4tcher I saw the lack of this PR was causing a few problems, and checked that it was waiting on #1801.  That never got reviewed, so just languished there, so I've eventually pushed it (after rebasing because we moved some of the plugins to the malware subcategory) so this should be good for you to handle the conflicts (there was one) and design a fix for any circular dependencies please...

Abyss-W4tcher:
Thanks, I'll work on it soon

Abyss-W4tcher:
@ikelos I've unified the module extraction code with the general modules utilities under a common file. I've also deprecated the module_extract.py file.

volatility3/framework/objects/utility.py
```diff
@@ -250,3 +252,50 @@ def array_of_pointers(
...
+            subtype: The subtype of the array's pointers.
+            stop_value: Stop value used to determine when to terminate iteration once it is encountered. Defaults to 0 (NULL-terminated arrays).
+            stop_on_invalid_pointers: Determines whether to stop iterating or not when an invalid pointer is encountered. This can be useful for arrays
+    that are known to have smeared entries before the end.
+
+        Returns:
+            An array of pointer objects
+    """
+    new_count = 0
+    for entry in array_of_pointers(
```
ikelos: I don't know whether this is handled in `array_of_pointers`, but my worry is reading off the edge of mapped memory, because we're reading all values up to `iterator_guard_value`.  My concern is that somehow that trips an exception which we don't seem to catch in here?

I guess the iteration should stop when one of the return values isn't readable, but what about applying this to 100 bytes of mapped memory and asking for 100 4-byte pointers?  How would this handle simply not being able to read enough memory?  Presumably it'd throw an exception, and that may be what we want, because it should only throw an exception when one of the values it expected to be able to read, could be read, but I just wanted to check...

volatility3/framework/objects/utility.py
```diff
@@ -250,3 +252,50 @@ def array_of_pointers(
...
+            break
+        new_count += 1
+    else:
+        vollog.log(
+            constants.LOGLEVEL_V,
+            f"""Iterator guard value {iterator_guard_value} reached while iterating over array at offset {array.vol.offset:#x}.\
+ This means that there is a bug (e.g. smearing) with this array, or that it may contain valid entries past the iterator guard value.""",
+        )
+
+    # Leverage the "Array" object instead of returning a Python list
```
ikelos, line 311: Good work, thanks!  5:D

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -3182,10 +3211,38 @@ def get_name(self) -> Optional[str]:
...
 
         return None
+
+
+class bin_attribute(objects.StructType):
+    def get_name(self) -> Optional[str]:
+        """
+        Performs extraction of the bin_attribute name
+        """
+        if hasattr(self, "attr"):
```
ikelos: Do we know what circumstances this might *not* have an `attr` attribute?  Is there a particular kernel version, or some other indicator we could put in there for when this might happen?

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -35,54 +31,20 @@
...
 class ModuleExtract(interfaces.configuration.VersionableInterface):
     """Extracts Linux kernel module structures into an analyzable ELF file"""
 
-    _version = (1, 0, 0)
+    _version = (1, 0, 1)
     _required_framework_version = (2, 25, 0)
 
     framework.require_interface_version(*_required_framework_version)
 
-    @classmethod
```
ikelos, line 46: Rather than removing this, it should still be possible to run, but with the deprercation warning provided.  I dunno if we already have a deprecation mechanism that can just redirect stuff from one method to another, but you'd need to check whether the new method and the old one had the same signature.  Essentially, it's not deprecation if you're not leaving it intact, it's removal.  And removal would require the version number to get bumped to `(2, 0, 0)` anyway, if the public interface changes.

Since this starts with `_`, we possibly don't class it as part of the public interface, but I can't follow along with when we're duplicating linux original names (and the `_` is supposed to be used by others) and when we're not and it is just private...  If you can figure that out, then if it's private it can go, and if not it needs to come back...

volatility3/framework/symbols/linux/utilities/module_extract.py
```diff
@@ -261,54 +223,6 @@ def _fix_sym_table(
 
         return sym_table_data
 
-    @classmethod
-    def _enumerate_original_sections(
```
ikelos, line 265: Similarly with this, if it's private, it can go, if it's public but "starts with an underscore to match linux naming" then it needs to stay...  5:S

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -781,6 +785,731 @@ def get_load_parameters(
...
+        context: interfaces.context.ContextInterface,
+        vmlinux_name: str,
+        original_sections: Dict[int, str],
+        section_sizes: Dict[int, int],
+        sym_type_name: str,
+        st_fmt: str,
+        module: extensions.module,
+    ) -> Optional[bytes]:
+        """
+        This function implements the most painful part of the reconstruction
```
ikelos, line 937: Would it be possible to get Napoleon/Google style docstrings here please?  Just trying to figure out what each of the parameters does, and why the format string is being passed in as a string, rather than a flag (since it seems to be a decider for whether to do ELF32 or ELF64?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -781,6 +785,731 @@ def get_load_parameters(
...
+                st_fmt, sym, section_lookups
+            )
+
+            # these aren't mangled during loading
+            st_name = struct.pack("<I", sym.st_name)
+            st_other = struct.pack("B", sym.st_other)
+            st_size = struct.pack(st_fmt, sym.st_size)
+
+            # The order as in the ELF specification. The order is not the same between 32 and 64 bit symbols
+            if st_fmt == "<I":
```
ikelos, line 986: This feels quite brittle.  *Anything* other than `<I` hit the else clause including format strings that might essentially produce the same value?  Feels like this should be a flag or something else and then the format string determined by the flag, rather than passed as a parameter.

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -781,6 +785,731 @@ def get_load_parameters(
...
+            name = section.get_name()
+            original_sections[section.address] = name
+
+        if not original_sections:
+            return None
+
+        if symbols.symbol_table_is_64bit(context, kernel.symbol_table_name):
+            sym_type = "Elf64_Sym"
+            elf_hdr_type = "Elf64_Ehdr"
+            st_fmt = "<Q"
```
ikelos, line 1049: So, just like this the values are being set within the method based on a flag, rather than passed in a parameters.

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -781,6 +785,731 @@ def get_load_parameters(
...
+        else:
+            sym_type = "Elf32_Sym"
+            elf_hdr_type = "Elf32_Ehdr"
+            st_fmt = "<I"
+
+        # At this point, we have the sections starting addresses and names,
+        # but the kernel does not track the size
+        # To recover the size, we sort by address and then use the next section as the boundary to calculate size
```
ikelos, line 1057: Could this lead to an explosion, if they're in memory spread out and we just read from one of the next (or if we're reading a bad symbol table, and the values are just arbitrary points in memory)?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -781,6 +785,731 @@ def get_load_parameters(
...
+        """
+        Creates a `bits` bit ELF header for the file based on recovered values
+        Called last as it needs information computed from the sections
+
+        Spec: https://refspecs.linuxfoundation.org/elf/gabi4+/ch4.eheader.html
+        """
+        if bits == 32:
+            fmt = "<I"
+            e_ident = (
+                b"\x7f\x45\x4c\x46\x01\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00"
```
ikelos, line 1165: Sooooo, these look a little bit like constants...  They also seemsd the same save for one byte in the middle.  Any chance of reducing this in some way or are we happy with it just being a massive binary string in the middle of our code?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -781,6 +785,731 @@ def get_load_parameters(
...
+        and breaks initial analysis by the tool
+        """
+        # All sections in memory are allocated (`A` in readelf -S)
+        flags = linux_constants.SHF_ALLOC
+
+        if name in [".text", ".init.text", ".exit.text", ".static_call.text"]:
+            flags = flags | linux_constants.SHF_EXECINSTR
+
+        elif name in [
+            ".data",
```
ikelos, line 1273: Again, quite a specific list.  Should it be pulled out into a constants or data file somewhere (specific to the ELF format) or are we happy with it just lurking in here, hidden away?

ikelos (commented):
First off, really good work!  This feels like an extremely high quality PR and was a pleasure to review, thanks!  5:D

Most of the comments are just checking up on boundary conditions of things.  Great work on doing the docstrings for everything, really appreciated!  (There are a couple of points where Napoleon style docstrings would help describe what each of the parameters does, but just having docstrings at all is such a relief, and references for all the specifics!  Awesome!!!)  5:D

Abyss-W4tcher:
Thanks for the review @ikelos! Just FYI 95% of the module_extract code appears as new but it's just copy/pasted from the previous module_extract.py location. I'll still answer the comments for that existing code no worries 👍.

volatility3/framework/objects/utility.py
Abyss-W4tcher: I think that is `array_of_pointers`'s responsibility, and it seems to raise an exception if it goes OOB:

```python
array = context.modules["kernel"].object("array", offset=(2**64)-1)
for entry in array_of_pointers(
    array=array, count=iterator_guard_value, subtype=subtype, context=context
):
   # ...
```

```
Page error 0x329ff000 in layer layer_name (Page Fault at entry 0x0 in table page directory pointer)
```

We could add a try/except block right here:

https://github.com/volatilityfoundation/volatility3/blob/a17281a2145f5aa353fcccc35f09cfcd40ad0aa4/volatility3/framework/objects/__init__.py#L811-L818

\+ a `error_on_oob`-ish parameter to the array object builder (default to True?):

`object("array", offset=(2**64)-1), error_on_oob=False`

which would return all the entries constructed up to the breaking point, but without raising an exception. What do you think?

volatility3/framework/symbols/linux/extensions/__init__.py
Abyss-W4tcher: `bin_attribute` should always have this member, I think I'll remove the hasattr call. If it breaks in a new kernel version, at least it won't silently return empty strings and we'll be able to update the method.

volatility3/framework/symbols/linux/utilities/module_extract.py
Abyss-W4tcher: This method was private and not referenced outside of this file, it's not a duplicated Linux kernel function name either.

volatility3/framework/symbols/linux/utilities/module_extract.py
Abyss-W4tcher: This method was private and not referenced outside of this file, it's not a duplicated Linux kernel function name either.

volatility3/framework/symbols/linux/utilities/modules.py
Abyss-W4tcher: Sure, I can add the docstring!

However as I said this code was already there, and this PR starts to diverge from the original fix I think. If you decide that this code needs to be cleaned up a bit, it would be more optimal to do it in a dedicated issue/PR after this one gets merged? _This also answers the other comments of the `ModuleExtract` class._

Abyss-W4tcher:
@ikelos Could you take a look at my comments please?

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Well, I tend to allow both as long as the fixes aren't too significant (and docstring fixes don't feel that significant), but it's up to you, if you want to do a second PR I'm fine with that too...

volatility3/framework/objects/utility.py
ikelos: I think making a "give me your best attempt" flag can be dangerous because people will use it all the time without considering the consequences, so I'd rather not provide it as a matter of course, but have each caller handle and explain why they can make use of partial results.  Does that seem reasonable or am I just making everyone write checking code that could be done once because I don't trust people to code sensibly?  What do you think would be best?  5:S

Abyss-W4tcher: The use of this flag should only be justified in contexts such as the dynamically sized arrays, because on one sample the kernel might have allocated 10 pages because there are 1000 entries and on another only 1 page because there are 100 entries. To make it generic we'd have to set the `iterator_guard_value` equal to 1000, but this would of course go OOB on the latter sample without a graceful exit. 

If we don't want to add a flag to `array` because of this one instance, we can resort to the following kind of setup:

```python
    sym_table_name = array.get_symbol_table_name()
    sym_table = context.symbol_space[sym_table_name]
    ptr_size = sym_table.get_type("pointer").size
    layer_name = array.vol.layer_name

    offset = array.vol.offset
    entry = None
    while entry != stop_value and new_count < iterator_guard_value:
        try:
            entry = context.object(
                sym_table_name + constants.BANG + "pointer",
                offset=offset,
                layer_name=layer_name,
            )
            break

        if not entry.is_readable() and stop_on_invalid_pointers:
            break

        offset += ptr_size
        new_count += 1

# etc.
# still using the clean method to return data
return array_of_pointers(
    array=array, count=new_count, subtype=subtype, context=context
)
```

We're removing a layer of abstraction, but have more control over what we're doing. Not so clean as the current MR version, but this way we keep this mechanism private and possibly make our lives easier?

volatility3/framework/symbols/linux/utilities/modules.py
Abyss-W4tcher: Yes, it should be easier to track if the code is refactored in a second PR in my opinion.

volatility3/framework/objects/utility.py
ikelos: Yeah, I think I'd prefer not to modify arrays to make them never fail, because it's really difficult to tell a caller that it *would* have failed, but we've returned you as many as we could.  Seems a bad way to go.

The way this MR would do it, or the snippet you posted above would be ok, as long as it's specific to this call.

volatility3/framework/objects/utility.py
```diff
@@ -250,3 +252,50 @@ def array_of_pointers(
     ).clone()
     subtype_pointer.update_vol(subtype=subtype)
     return array.cast("array", count=count, subtype=subtype_pointer)
+
+
+def dynamically_sized_array_of_pointers(
+    context: interfaces.context.ContextInterface,
+    array: interfaces.objects.ObjectInterface,
+    iterator_guard_value: int,
```
ikelos: I think it's worth moving this later, just because I can imagine us wanting to give it a default value down the line, and we can't do that with subtype coming after it.

ikelos (commented):
A couple more additions, but I think it's pretty much there.  Thanks for musing over my considerations with me.  5:)

Abyss-W4tcher:
All done, I've pushed one additional fix that prevents the private `_parse_section` method from mixing the number of return values on failure (it was either None or (None, None, None)). I've noticed that error when lowering `iterator_guard_value` to 10 to see how the plugin would react without all the required sections.

volatility3/framework/plugins/linux/module_extract.py
```diff
@@ -37,9 +37,9 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                 optional=False,
             ),
             requirements.VersionRequirement(
-                name="linux_utilities_module_extract",
-                version=(1, 0, 0),
-                component=linux_utilities_module_extract.ModuleExtract,
+                name="linux_utilities_modules_module_extract",
+                version=(1, 0, 2),
```
ikelos, line 41: The PATCH component of a version isn't used in decisions of meeting requirements (because they're only supposed to be used for internal changes, so 1.0 should work as well as 1.0.2 as well as 1.0.1001).  It's fine to leave it here, I just want to ensure that the undrestanding of what these checks do is recorded somewhewre...  5:)

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: What was the view about getting these long lists of essentially data, moved out to the constants file?  It feels like they should at least be exposed at the class level or module level, so that other components that may need them can keep in sync?

volatility3/framework/symbols/linux/utilities/modules.py
```diff
@@ -781,6 +785,740 @@ def get_load_parameters(
...
+        # Gather sections
+        parse_sections_result = cls._parse_sections(context, vmlinux_name, module)
+        if parse_sections_result is None:
+            return None
+        updated_sections, strtab_index, symtab_index = parse_sections_result
+
+        kernel = context.modules[vmlinux_name]
+
+        # Figure out header sizes
+        if symbols.symbol_table_is_64bit(context, kernel.symbol_table_name):
```
ikelos, line 1422: I know that 32-bit support under 64-bit systems is old and likely not used much anymore, but how did it work back in the day when it was done?  Could a 32-bit binary be loaded in a 64-bit system without modification?

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Maybe a constants.linux.elf file or something?  This list clearly has a specific meaning, so I'd gather them up under that (like `constants.linux.elf.SHF_WRITABLE_SECTIONS` or something?).  Feel free to call me out if the suggestion is overkill, just want to make sure we've reached a decision on it one way or another before this lands.

ikelos (commented):
Just looked over the outstanding comments and still had a couple that seemed unanswewred...

volatility3/framework/symbols/linux/utilities/modules.py
Abyss-W4tcher: I've looked it up and it doesn't seem possible for LKMs:

- https://documentation.suse.com/sles/15-SP7/html/SLES-all/cha-64bit.html#:~:text=A%2064%2Dbit%20kernel%20can,modules%20with%2064%2Dbit%20kernels.

volatility3/framework/symbols/linux/utilities/modules.py
Abyss-W4tcher: Ok that's noted, I'll circle back to all these comments and work on them in the code cleaning MR that I'll open shortly after this one.

volatility3/framework/symbols/linux/utilities/module_extract.py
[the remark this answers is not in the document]
Abyss-W4tcher: Done

Abyss-W4tcher:
Ok, I've resolved all the conversations except the ones related to the already existing code as we discussed. There isn't a version bump though, I guess it should be MINOR as we moved one API from one place to another?

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Any update on this?

volatility3/framework/symbols/linux/utilities/modules.py
ikelos: Or were they going to be wrapped up in the cleanup?  I'd kinda prefer we just commit clean code, rather than committing known code we want to change, and then changing it later?

ikelos:
Err, for deprecated classes, leave them as is (with the old code, which will be removed as part of the deprecation).  If we're just removing part of a class that's going to continue to exist, then I'm afraid it'll be painful, but removals require a bump to the MAJOR version (or, we change the code to call the moved code instead?).  Happy with either way, but yeah, it'll need fixing.

ikelos:
Ok, looks good, let's get this in the tree.  5:)

eve-mem:
Great work getting this in, it kept coming up as a problem for me.

PR #1777, opened 2025-04-15
ThrdScan: Fix tracebacks + filtering

dgmcdona:
- ThrdScan: Fix process filtering
    
  This check was both causing an `InvalidAddressException` due to the
  member access, while at the same time not being a useful check, since it
  prevents VADs from being mapped in children of the `System` process.

- Thrdscan: Remove filtering based on VAD count
  
  This was preventing enumeration of valid processes (confirmed by
  disassembly of the start address/Win32 start address). Heuristic-based
  filtering should probably be left to consumers of the APIs.

### DIFF
--- a/volatility3/framework/plugins/windows/thrdscan.py
+++ b/volatility3/framework/plugins/windows/thrdscan.py
@@ -119,11 +119,6 @@ class ThrdScan(interfaces.plugins.PluginInterface, timeliner.TimeLinerInterface)
             vads = pe_symbols.PESymbols.get_vads_for_process_cache(
                 vads_cache, owner_proc
             )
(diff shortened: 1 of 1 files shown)


dgmcdona:
@ikelos this diverges from the fixes implemented in #1766 since `EPROCESS.is_valid()` didn't get an update - if you'd prefer to have them in sync, I can bring in those changes and retest. I was just trying to keep this one as minimal as possible while fixing the obvious tracebacks.

ikelos (approved):
I think this looks ok, we'll just need to be careful if we ever merge from release back into develop (we'd need to fix the version number too, but it might do some of that based on commit date)...

PR #1779, opened 2025-04-17
Amend SvcDiff comments at top of file

j-t-1:
Previously contained information about Skeleton_Key_Check, change this to be about SvcDiff.


ikelos (approved):
Looks good, thanks for spotting it and putting in a more suitable link!

PR #1781, opened 2025-04-17
1780 dlllist loadcount missing

### LINKED ISSUE #1780: DllList LoadCount missing
**Describe the bug**
DllList output doesn't include the LoadCount column.

**Expected output**
<pre>
$ python vol.py -q -r pretty -f data.lime windows.dlllist
Volatility 3 Framework 2.26.2
Formatting...
  |  PID |         Process |       Base |     Size |                Name |                                                                                                              Path | LoadCount | LoadTime | File output
* |  508 |        smss.exe | 0x48580000 |   0xf000 |            smss.exe |                                                                                     \SystemRoot\System32\smss.exe |        -1 |      N/A |    Disabled
* |  508 |        smss.exe | 0x7c900000 |  0xaf000 |           ntdll.dll |                                                                                     C:\WINDOWS\system32\ntdll.dll |        -1 |      N/A |    Disabled
* |  608 |       csrss.exe | 0x4a680000 |   0x5000 |           csrss.exe |                                                                                 \??\C:\WINDOWS\system32\csrss.exe |        -1 |      N/A |    Disabled
* |  608 |       csrss.exe | 0x7c900000 |  0xaf000 |           ntdll.dll |                                                                                     C:\WINDOWS\system32\ntdll.dll |        -1 |      N/A |    Disabled
* |  608 |       csrss.exe | 0x75b40000 |   0xb000 |          CSRSRV.dll |                                                                                    C:\WINDOWS\system32\CSRSRV.dll |        -1 |      N/A |    Disabled
* |  608 |       csrss.exe | 0x75b50000 |  0x10000 |         basesrv.dll |                                                                                   C:\WINDOWS\system32\basesrv.dll |         3 |      N/A |    Disabled
* |  608 |       csrss.exe | 0x75b60000 |  0x4b000 |          winsrv.dll |                                                                                    C:\WINDOWS\system32\winsrv.dll |         2 |      N/A |    Disabled
* |  608 |       csrss.exe | 0x77f10000 |  0x49000 |           GDI32.dll |                                                                                     C:\WINDOWS\system32\GDI32.dll |         5 |      N/A |    Disabled
* |  608 |       csrss.exe | 0x7c800000 |  0xf6000 |        KERNEL32.dll |                                                                                  C:\WINDOWS\system32\KERNEL32.dll |        16 |      N/A |    Disabled

</pre>

superponible:
Add the LoadCount column with a symbol extension since it depends on OS version.

@ikelos Not sure if there's a better way of printing the -1. the actual data is a signed short as 0xffff.

Also would this need a version bump and if so which number?

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -1710,3 +1710,17 @@ class SHARED_CACHE_MAP(objects.StructType):
                     )
 
         return vacb_list
+
+class LDR_DATA_TABLE_ENTRY(objects.StructType):
+    def get_load_count(self) -> Optional[int]:
+        try:
+            LoadCount = self.LoadCount
+        except:
+            try:
+                LoadCount = self.ObsoleteLoadCount
+            except:
+                LoadCount = None
+        if LoadCount == 65535:
+            LoadCount = -1
+
+        return LoadCount
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -1711,14 +1711,15 @@ class SHARED_CACHE_MAP(objects.StructType):
 
         return vacb_list
 
+
 class LDR_DATA_TABLE_ENTRY(objects.StructType):
     def get_load_count(self) -> Optional[int]:
         try:
             LoadCount = self.LoadCount
-        except:
+        except Exception:
             try:
                 LoadCount = self.ObsoleteLoadCount
-            except:
+            except Exception:
                 LoadCount = None
         if LoadCount == 65535:
             LoadCount = -1
--- a/volatility3/framework/symbols/windows/extensions/__init__.py
+++ b/volatility3/framework/symbols/windows/extensions/__init__.py
@@ -1715,13 +1715,11 @@ class SHARED_CACHE_MAP(objects.StructType):
 class LDR_DATA_TABLE_ENTRY(objects.StructType):
     def get_load_count(self) -> Optional[int]:
         try:
-            LoadCount = self.LoadCount
+            LoadCount = self.LoadCount.cast("short")
         except Exception:
             try:
-                LoadCount = self.ObsoleteLoadCount
+                LoadCount = self.ObsoleteLoadCount.cast("short")
             except Exception:
                 LoadCount = None
-        if LoadCount == 65535:
-            LoadCount = -1
 
         return LoadCount
--- a/volatility3/framework/constants/_version.py
+++ b/volatility3/framework/constants/_version.py
@@ -1,7 +1,7 @@
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 2  # Number of releases of the library with a breaking change
-VERSION_MINOR = 26  # Number of changes that only add to the interface
-VERSION_PATCH = 2  # Number of changes that do not change the interface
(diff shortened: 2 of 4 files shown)


volatility3/framework/symbols/windows/extensions/__init__.py
```diff
@@ -1710,3 +1710,18 @@ def get_available_pages(self) -> List:
...
+    def get_load_count(self) -> Optional[int]:
+        try:
+            LoadCount = self.LoadCount
+        except Exception:
+            try:
+                LoadCount = self.ObsoleteLoadCount
+            except Exception:
+                LoadCount = None
+        if LoadCount == 65535:
+            LoadCount = -1
```
ikelos: Is this a valid situation, or should this also return `None` as a means of error?

ikelos (commented):
Looks fine, just trying to figure out whether `-1` has special meaning or not?

volatility3/framework/symbols/windows/extensions/__init__.py
ikelos: If it's supposed to be signed, you could `.cast('CORRECT_TYPE')` it to the right type?

ikelos:
@superponible This has been waiting on a response for a while.  Has it gone stale or would you still like it merging in?  Due to the addition of the windows symbol is *probably* should get... I think... a framework MINOR bump, since I don't think there's a more refined way of ensuring that symbol mapping will be present otherwise...

volatility3/framework/symbols/windows/extensions/__init__.py
superponible: A cast is what I was thinking, thanks for the hint. I casted to a short and now it shows -1 like it should.

Not sure what you meant about return None as means of an error. I don't think there should be a case where both LoadCount and ObsoleteLoadCount don't exist, but just had the exception there to prevent a crash in case that ever happened.

ikelos:
Thanks for getting that sorted so quickly.  Could we just get the PATCH version of the dlllist plugin bumped (to `(3,0,1)`), and the MINOR version of the framework (to `(2,27,0)`) then I think this should be all good for inclusion...  5:)

superponible:
> Thanks for getting that sorted so quickly. Could we just get the PATCH version of the dlllist plugin bumped (to `(3,0,1)`), and the MINOR version of the framework (to `(2,27,0)`) then I think this should be all good for inclusion... 5:)

Is that changing both values in the dlllist plugin like this:

<pre>
-    _required_framework_version = (2, 0, 0)
-    _version = (3, 0, 0)
+    _required_framework_version = (2, 27, 0)
+    _version = (3, 0, 1)
</pre>

or by changing the framework version were you saying change that somewhere else. I'm not too familiar with how the versioning gets handled.

ikelos:
Nope just `_version` in `volatility3/framework/plugins/windows/dlllist.py` and `VERSION_MINOR = 27` in `volatility3/framework/constants/_version.py`

ikelos:
That'll bump the entire framework version, but we don't have finer granularity on the default windows symbol tables than that, and with semver the minor version can get pretty big, pretty quickly, it's just par for the course...

superponible:
> That'll bump the entire framework version, but we don't have finer granularity on the default windows symbol tables than that, and with semver the minor version can get pretty big, pretty quickly, it's just par for the course...

Got it.  And I assume bumping VERSION_MINOR to 27 should set VERSION_PATCH to 0 as well?

ikelos:
Yep, exactly right, thanks.  5:)

ikelos (approved):

PR #1783, opened 2025-04-22
Windows Shimcachemem: Fix tracebacks

dgmcdona:
This fixes a couple of uncaught exceptions in the shimcachemem plugin and extension classes - one when iterating over modules + checking their names, and the other when attempting to parse the exec flag. Also changes a return type from `renderers.UnparsableValue` to `renderers.UnreadableValue` which I think is more appropriate, since the underlying data could not be read from the address space.


ikelos (approved):
Thanks, looks good.  5:)

PR #1786, opened 2025-04-23
Windows Handles: Refactor plugin

dgmcdona:
A lot of functionality in the `Handles` plugin class was using instance methods instead of classmethods. This refactors the plugin to use classmethods instead of instance methods for consistency with the rest of the framework, and bumps the plugin major version to 4.0.0. It also updates plugins that depend on `Handles` with the correct required major version, and converts instances where an instance of the plugin is being constructed to uses of the new classmethods.

Also improves some docstrings and type-hinting in `handles.py`.

### DIFF
--- a/volatility3/framework/plugins/windows/handles.py
+++ b/volatility3/framework/plugins/windows/handles.py
@@ -3,7 +3,7 @@
 #
 
 import logging
-from typing import Dict, List, Optional
+from typing import Dict, Iterator, List, Optional
 
 from volatility3.framework import constants, exceptions, interfaces, renderers, symbols
 from volatility3.framework.configuration import requirements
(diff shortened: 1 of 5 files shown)


volatility3/framework/plugins/windows/handles.py
```diff
@@ -18,13 +18,9 @@ class Handles(interfaces.plugins.PluginInterface):
...
     _required_framework_version = (2, 0, 0)
-    _version = (3, 0, 0)
+    _version = (4, 0, 0)
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self._type_map = None
-        self._cookie = None
-        self._level_mask = 7
+    LEVEL_MASK = 7
```
ikelos, line 23: Can't decide whether this is a constant that other plugins might want to access, or just an internal constant only for use by the plugin?  What do you think?

ikelos (commented):
Thanks, this looks a lot cleaner.  Just mulling over whether the LEVEL_MASK should qualify as a fully fledged constant or not...  I'm not sure it needs to live elsewhere, but I don't want people seeing it and using it to justify local constants...  5:S

volatility3/framework/plugins/windows/handles.py
dgmcdona: I think it should probably be internal, I can't think of anything else that would have a use for it right now. Interestingly, it is defined as 3 instead of 7 in the reactOS source: https://github.com/reactos/reactos/blob/d722e7081796692dc2fb225517d2d9fb6283d64b/ntoskrnl/ex/handle.c#L64-L65

I tested against the regression set with both values and there were no output changes, so I guess it's fine for now regardless.

ikelos (approved):

PR #1787, opened 2025-04-24
Detect ETW Patching

JakePeralta7:
This plugin identifies ETW patching by inspecting the first opcode of the following functions:
- EtwEventWrite
- EtwEventWriteFull
- NtTraceEvent

If the initial opcode is 0xC3 (RET) or 0xE9 (JMP), the function is flagged as patched.

Below is a screenshot demonstrating when the detection is triggered:
WhatsApp Image 2025-04-24 at 20 33 46_3972c8e1

Also, I've added support for scanning specific processes.

This method only checks the first opcode. If an attacker patches elsewhere in the function in a way that corrupts it - the modification will not be detected.

Note: This is my first plugin - any feedback is welcome!

### DIFF
--- a/volatility3/framework/plugins/windows/etwpatch.py
+++ b/volatility3/framework/plugins/windows/etwpatch.py
@@ -95,8 +95,7 @@ class EtwPatch(interfaces.plugins.PluginInterface):
                     try:
                         opcode = (
                             self.context.layers[proc_layer_name]
-                            .read(func_addr, 1)
-                            .hex()
+                            .read(func_addr, 1)[0]
                         )
                         if opcode in opcode_map:
                             instruction = opcode_map[opcode]
--- a/volatility3/framework/plugins/windows/etwpatch.py
+++ b/volatility3/framework/plugins/windows/etwpatch.py
@@ -86,17 +86,16 @@ class EtwPatch(interfaces.plugins.PluginInterface):
 
             # Map of opcodes to their instruction names
             opcode_map = {
-                0xc3: "RET",
-                0xe9: "JMP",
+                0xC3: "RET",
+                0xE9: "JMP",
             }
 
             for dll_name, functions in found_symbols.items():
                 for func_name, func_addr in functions:
                     try:
-                        opcode = (
-                            self.context.layers[proc_layer_name]
-                            .read(func_addr, 1)[0]
-                        )
+                        opcode = self.context.layers[proc_layer_name].read(
+                            func_addr, 1
+                        )[0]
                         if opcode in opcode_map:
                             instruction = opcode_map[opcode]
                             yield (
--- a/volatility3/framework/plugins/windows/etwpatch.py
+++ b/volatility3/framework/plugins/windows/etwpatch.py
@@ -42,90 +42,80 @@ class EtwPatch(interfaces.plugins.PluginInterface):
             )
         ]
 
+    def _get_dll_vads(self, proc, dll_name):
+        """Retrieve VADs for a specific DLL in the process."""
+        collected_modules = pe_symbols.PESymbols.get_proc_vads_with_file_paths(proc)
+        return [
+            (vad_start, vad_size, proc.add_process_layer())
+            for vad_start, vad_size, filepath in collected_modules
+            if pe_symbols.PESymbols.filename_for_path(filepath) == dll_name
+        ]
+
+    def _find_symbols(self, dll_name, symbols, proc_layer_name, dll_vads):
+        """Find symbols for a specific DLL."""
+        filter_module = {dll_name: {"names": symbols}}
+        process_modules = {dll_name: [(proc_layer_name, vad_start, vad_size) for vad_start, vad_size, _ in dll_vads]}
+        return pe_symbols.PESymbols.find_symbols(self.context, self.config_path, filter_module, process_modules)
+
+    def _get_first_opcode(self, proc_layer_name, function_start):
+        """Check the first opcode of a function."""
+        try:
+            return self.context.layers[proc_layer_name].read(function_start, 1).hex()
+        except exceptions.InvalidAddressException:
+            return None
+
     def _generator(self):
         pid_filter = self.config.get('pid', None)
 
         for proc in pslist.PsList.list_processes(
                 context=self.context,
                 kernel_module_name=self.config['kernel']):
-            
-            # If the user passed --pid, only process those IDs
+            # Skip processes not in the PID filter
             if pid_filter and proc.UniqueProcessId not in pid_filter:
                 continue
 
             pid = int(proc.UniqueProcessId)
             proc_name = proc.ImageFileName.cast(
                 "string",
-                max_length = proc.ImageFileName.vol.count,
-                errors = 'replace'
+                max_length=proc.ImageFileName.vol.count,
+                errors='replace'
             )
 
-            # Build a per-process memory layer
             try:
                 proc_layer_name = proc.add_process_layer()
-            except Exception:
+            except exceptions.InvalidAddressException:
                 continue
 
-            proc_layer = self.context.layers[proc_layer_name]
+            dlls_to_check = {
+                "ntdll.dll": [
+                    "EtwEventWrite",
+                    "EtwEventWriteFull",
+                    "NtTraceEvent"
+                ],
+                "advapi32.dll": [
+                    "EventWrite"
+                ]
+            }
 
-            # Find ntdll.dll module
-            for module in proc.load_order_modules():
-                BaseDllName = FullDllName = renderers.UnreadableValue()
-                with contextlib.suppress(exceptions.InvalidAddressException):
-                    BaseDllName = module.BaseDllName.get_string()
-                    FullDllName = module.FullDllName.get_string()
-                
-                if BaseDllName != 'ntdll.dll':
+            for dll_name, symbols in dlls_to_check.items():
+                dll_vads = self._get_dll_vads(proc, dll_name)
+                if not dll_vads:
                     continue
 
-                base = module.DllBase
-                size = module.SizeOfImage
-
-                pe_table_name = intermed.IntermediateSymbolTable.create(
-                    self.context, self.config_path, "windows", "pe", class_types=pe.class_types
-                )
-                
-                pe_obj = pe_symbols.PESymbols.get_pefile_obj(
-                    self.context, pe_table_name, proc_layer_name, base
-                )
-                
-                try:
-                    pe_obj.parse_data_directories(
-                        directories=[pefile.DIRECTORY_ENTRY["IMAGE_DIRECTORY_ENTRY_EXPORT"]]
-                    )
-                except Exception as e:
-                    vollog.debug(f"Error parsing IMAGE_DIRECTORY_ENTRY_EXPORT with {e}")
+                found_symbols, _ = self._find_symbols(dll_name, symbols, proc_layer_name, dll_vads)
+                if dll_name not in found_symbols:
                     continue
-                
-                if not hasattr(pe_obj, "DIRECTORY_ENTRY_EXPORT"):
-                    return None
-                
-                for export in pe_obj.DIRECTORY_ENTRY_EXPORT.symbols:
-                    if export.name not in [b"EtwEventWrite", b"EtwEventWriteFull", b"NtTraceEvent"]:
-                        continue
-        
-                    function_start = base + export.address
-                    try:
-                        with contextlib.suppress(exceptions.InvalidAddressException):
-                            opcode = self.context.layers[proc_layer_name].read(
-                                function_start, 1
-                            ).hex()
-                        
-                            # 0xC3 = RET, 0xE9 = JMP (common ETW patches)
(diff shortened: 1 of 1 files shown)


atcuno:
@ikelos - I will help with this one as I am very familiar with the intent and it should go through the pe_symbols API to be complete. I will ping you again when it is ready.

JakePeralta7:
> @ikelos - I will help with this one as I am very familiar with the intent and it should go through the pe_symbols API to be complete. I will ping you again when it is ready.

What do you mean by going through the pe_symbols API? (I'm using pe_symbols to get the PE object)

JakePeralta7:
> @ikelos - I will help with this one as I am very familiar with the intent and it should go through the pe_symbols API to be complete. I will ping you again when it is ready.

Thanks for the feedback, I've extended the usage of the `pe_symbols` plugin

atcuno:
Thank you so much @JakePeralta7 - sorry for not getting back with more specifics. This weekend got lost in other work. I will leave just a few comments through the code itself then this should be good to go.

volatility3/framework/plugins/windows/etwpatch.py
```diff
@@ -0,0 +1,128 @@
+# This file is Copyright 2019 Volatility Foundation and licensed under the Volatility Software License 1.0
```
atcuno: This needs to be 2025 since it is a new plugin

volatility3/framework/plugins/windows/etwpatch.py
```diff
@@ -0,0 +1,128 @@
...
+        collected_modules = pe_symbols.PESymbols.get_proc_vads_with_file_paths(proc)
+        return [
+            (vad_start, vad_size, proc.add_process_layer())
+            for vad_start, vad_size, filepath in collected_modules
+            if pe_symbols.PESymbols.filename_for_path(filepath) == dll_name
+        ]
+
+    def _find_symbols(self, dll_name, symbols, proc_layer_name, dll_vads):
+        """Find symbols for a specific DLL."""
+        filter_module = {dll_name: {"names": symbols}}
```
atcuno: Instead of the raw "names" string use the identifier constant like unhooked system calls does here:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/unhooked_system_calls.py#L29

volatility3/framework/plugins/windows/etwpatch.py
```diff
@@ -0,0 +1,128 @@
...
+            return None
+
+    def _generator(self):
+        pid_filter = self.config.get('pid', None)
+
+        for proc in pslist.PsList.list_processes(
+                context=self.context,
+                kernel_module_name=self.config['kernel']):
+            # Skip processes not in the PID filter
+            if pid_filter and proc.UniqueProcessId not in pid_filter:
```
atcuno: Instead of manually checking the PID just use the built-in filter for this like at:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/cmdline.py#L103

volatility3/framework/plugins/windows/etwpatch.py
```diff
@@ -0,0 +1,128 @@
...
+                continue
+
+            pid = int(proc.UniqueProcessId)
```
atcuno: No need to cast int() in Volatility 3

volatility3/framework/plugins/windows/etwpatch.py
```diff
@@ -0,0 +1,128 @@
...
+            proc_name = proc.ImageFileName.cast(
```
atcuno: Use array_to_string instead of manually creating the string:

https://github.com/volatilityfoundation/volatility3/blob/0c08b94547186474518a6a2462d0216efd2642d5/volatility3/framework/plugins/windows/cmdline.py#L72

volatility3/framework/plugins/windows/etwpatch.py
```diff
@@ -0,0 +1,128 @@
...
+                for symbol_name, function_start in found_symbols[dll_name]:
+                    opcode = self._get_first_opcode(proc_layer_name, function_start)
+                    if opcode in ('c3', 'e9'):  # RET or JMP
+                        yield (0, (
+                            pid,
+                            proc_name,
+                            dll_name,
+                            symbol_name,
+                            format_hints.Hex(function_start),
+                            opcode
```
atcuno: It would be better to yield the string name of the instruction (RET or JMP) instead of the hex opcode to aid people less familiar with reversing.

JakePeralta7:
> Thank you so much @JakePeralta7 - sorry for not getting back with more specifics. This weekend got lost in other work. I will leave just a few comments through the code itself then this should be good to go.

Great comments, I think I've solved all of the problems you have raised (I have to say that the code look much more readable 🤣)

volatility3/framework/plugins/windows/etwpatch.py
```diff
@@ -0,0 +1,116 @@
...
+                                    proc_name,
+                                    dll_name,
+                                    func_name,
+                                    format_hints.Hex(func_addr),
+                                    f"{opcode} ({instruction})"
+                                ),
+                            )
+                    except exceptions.InvalidAddressException:
+                        vollog.debug(f"Invalid address when reading function {func_name} at {func_addr:#x} in process {proc_id}")
+                        continue
```
atcuno: The continue isn't needed here

volatility3/framework/plugins/windows/etwpatch.py
```diff
@@ -0,0 +1,116 @@
...
+                    except KeyError:
```
atcuno: Did you see this error in practice? Layers shouldn't be getting deleted while a plugin is running.

atcuno:
Looking very nice. Just two more comments on the latest push. After addressing those, I can run the tests plus test on my local samples with these functions hooked.

volatility3/framework/plugins/windows/etwpatch.py
JakePeralta7: Just thought handling that because I've used `layers[proc_layer_name]`, but I guess it's not necessary

volatility3/framework/plugins/windows/etwpatch.py
JakePeralta7: Yep

JakePeralta7:
> Looking very nice. Just two more comments on the latest push. After addressing those, I can run the tests plus test on my local samples with these functions hooked.

I hope I got it right this time

atcuno:
The code looks good now, but it needs to be fixed for black and ruff. You can install these through pip then do (In order):

1) black volatility3/framework/plugins/windows/etwpatch.py

^ This will lint the file

2) ruff check volatility3/framework/plugins/windows/etwpatch.py

^ This will report any formatting or other coding issues that need to be fixed.

Once these pass locally, then re-commit your code.

atcuno:
The code analysis checks failed about the requirements as it looks like you copy/pasted the pslist requirement and changed the name= but not the component=. Just changed it to be the pe_symbols class instead:

https://github.com/volatilityfoundation/volatility3/blob/c5a4b34bfa5f1466b61551768ea0ad61becbc200/volatility3/framework/plugins/windows/etwpatch.py#L53

JakePeralta7:
@atcuno
Fixed it (thanks for the patience walking me through the process)

atcuno:
Ok I just re-triggered the tests. To fix the requirements error just make the line look like this:

https://github.com/volatilityfoundation/volatility3/blob/develop/volatility3/framework/plugins/windows/unhooked_system_calls.py#L101

atcuno:
This is ready now @ikelos. Awesome work @JakePeralta7 !!

volatility3/framework/plugins/windows/etwpatch.py
```diff
@@ -0,0 +1,130 @@
...
+                proc_name = utility.array_to_string(proc.ImageFileName)
+                proc_layer_name = proc.add_process_layer()
+            except exceptions.InvalidAddressException:
+                vollog.debug(f"Unable to create process layer for PID {proc_id}")
+                continue
+
+            # Map of opcodes to their instruction names
+            opcode_map = {
+                "c3": "RET",
+                "e9": "JMP",
```
ikelos:
```suggestion
                0xc3: "RET",
                0xe9: "JMP",
```
Rather than storing these as strings, I'd just keep them as integers and then change the comparison below to just read a single byte.

volatility3/framework/plugins/windows/etwpatch.py
```diff
@@ -0,0 +1,130 @@
...
+            }
+
+            for dll_name, functions in found_symbols.items():
+                for func_name, func_addr in functions:
+                    try:
+                        opcode = (
+                            self.context.layers[proc_layer_name]
+                            .read(func_addr, 1)
+                            .hex()
+                        )
```
ikelos:
```suggestion
                        opcode = (
                            self.context.layers[proc_layer_name]
                            .read(func_addr, 1)[0]
                        )
```
This reads the single byte as a number.

ikelos (commented):
One little clear up just to avoid confusion around unnecessary hex/string conversions...  Otherwise looks good, just a couple of general questions to think about for the future.  

1) Have you considered whether the functionality used could be used by more plugins and could therefore be pulled out into an appropriately parameterized `@classmethod`?
2) This may find false positives where databytes are read as if they're an opcode, it's likely overkill, but you could pull in capstone to do disassembly and check the actual opcodes (and only the opcodes)?  For an example of this see `direct_system_calls` or `skeleton_key_check`...
3) Possibly we should gather all the malware specific plugins and put them under the malware class (so `windows.malware.etwpatch` rather than just `windows.etwpatch`...

volatility3/framework/plugins/windows/etwpatch.py
JakePeralta7: image
I don't really object this output, it's just more iconic to see the opcode in hex format

JakePeralta7:
> One little clear up just to avoid confusion around unnecessary hex/string conversions... Otherwise looks good, just a couple of general questions to think about for the future.
> 
> 1. Have you considered whether the functionality used could be used by more plugins and could therefore be pulled out into an appropriately parameterized `@classmethod`?
> 2. This may find false positives where databytes are read as if they're an opcode, it's likely overkill, but you could pull in capstone to do disassembly and check the actual opcodes (and only the opcodes)?  For an example of this see `direct_system_calls` or `skeleton_key_check`...
> 3. Possibly we should gather all the malware specific plugins and put them under the malware class (so `windows.malware.etwpatch` rather than just `windows.etwpatch`...

1. My take on this kind of approach is when something else will need it I will abstract, but guessing what will be needed is sometimes unnecessary
2. I'm reading from an offset of a dll exported function (the exact start of it), so is that really a risk?
3. Sounds great

volatility3/framework/plugins/windows/etwpatch.py
```diff
@@ -0,0 +1,130 @@
...
+                            instruction = opcode_map[opcode]
+                            yield (
+                                0,
+                                (
+                                    proc_id,
+                                    proc_name,
+                                    dll_name,
+                                    func_name,
+                                    format_hints.Hex(func_addr),
+                                    f"{opcode} ({instruction})",
```
ikelos:
```suggestion
                                    f"{opcode:02x} ({instruction})",
```
Outputting as hex from an integer is easy  5;)

volatility3/framework/plugins/windows/etwpatch.py
ikelos: Yep, knew there was something else I meant to add, changed the format string to output as a 2-digit hex value with a leading 0.

volatility3/framework/plugins/windows/etwpatch.py
```diff
@@ -0,0 +1,129 @@
...
+                0xe9: "JMP",
+            }
+
+            for dll_name, functions in found_symbols.items():
+                for func_name, func_addr in functions:
+                    try:
+                        opcode = (
+                            self.context.layers[proc_layer_name]
+                            .read(func_addr, 1)[0]
+                        )
```
ikelos:
```suggestion
                        opcode = self.context.layers[proc_layer_name].read(
                            func_addr, 1
                        )[0]                        
```
Sorry, black just being a bit specific about the formatting...  5:S

volatility3/framework/plugins/windows/etwpatch.py
```diff
@@ -0,0 +1,129 @@
...
+                proc_name = utility.array_to_string(proc.ImageFileName)
+                proc_layer_name = proc.add_process_layer()
+            except exceptions.InvalidAddressException:
+                vollog.debug(f"Unable to create process layer for PID {proc_id}")
+                continue
+
+            # Map of opcodes to their instruction names
+            opcode_map = {
+                0xc3: "RET",
+                0xe9: "JMP",
```
ikelos:
```suggestion
                0xC3: "RET",
                0xE9: "JMP",
```

ikelos (commented):
Nearly there, just black being a bit specific in its tastes...

ikelos:
Oh, beat me to it!  5;P

PR #1788, opened 2025-04-25
Windows: update psscan to fix issue #591

eve-mem:
Hello :wave: 

This is a continuation of PR https://github.com/volatilityfoundation/volatility3/pull/1215 but I completely messed up that branch with a rebase that went very wrong, and my git-fu isn't strong enough to fix it. 

This fixes an issue where if you attempt to use the `--physical` option on some windows samples psscan will crash. This PR fixes those issues and includes the changes suggested by @ikelos in the previous PR.

:fox_face:


eve-mem:
I'll get that test fixed!

volatility3/framework/plugins/windows/psscan.py
```diff
@@ -269,39 +269,60 @@ def _generator(self):
...
+                else:
+                    # the display should be virtual address, so vproc should be used
+                    # however virtual_process_from_physical is not always able to create
+                    # a vproc, in that case we need to display a UnreadableValue()
+                    if vproc is not None:
+                        offset = vproc.vol.offset
+                    else:
+                        offset = None
 
+            if self.config["dump"]:
```
ikelos, line 309: Ideally we'd guard for `if vproc is None:`.  If we hand it in, it will debug log but otherwise return None, so it's not catastrophic, but feels like an easy thing we can check (and catchall exception handlers aren't ideal, so that may go away at some point in the future)...

volatility3/framework/plugins/windows/psscan.py
```diff
@@ -269,39 +269,60 @@ def _generator(self):
...
         ):
             file_output = "Disabled"
-            if self.config["dump"]:
-                # windows 10 objects (maybe others in the future) are already in virtual memory
-                if proc.vol.layer_name == kernel.layer_name:
-                    vproc = proc
+
+            # windows 10 objects (maybe others in the future) are already in virtual memory
+            # if the proc native_layer_name and layer_name match then it is in 'virtual' memory.
+            if proc.vol.layer_name == proc.vol.native_layer_name:
```
ikelos, line 275: This isn't a robust test (it only works when `scan_processes` returns carved processes), and I don't really want people using it to determine whether an address is physical or not.  The poolscanner's been asked to scan the kernel layer, so... it should only ever return virtual results?  Have you verified that the "else" branch ever gets hit?

ikelos (commented):
So, I'm a little concerned that the support for scanning physical layers won't ever actually get used, because we only ever ask the poolscanner to scan the kernel image?

eve-mem:
@ikelos that's a really good point. I've just been focused on getting rid of the back trace for @garanews rather than looking at the plugin as a whole. 

The windows psscan is very different to the linux one. Linux will always scan the 'memory layer' and produce results from that, e.g. always physical addresses and it doesn't scan an intel layer. 

What are peoples thoughts on changing the windows one to be closer to the linux counterpart? It's a bigger change to how psscan works but to my knowledge there aren't any public plugins that rely on psscan. It would make the logic a lot simpler.

ikelos:
I think as long as we don't _lose_ functionality (ie, it can still give you virtual addresses, if that's what you expect) then it should be fine?

atcuno:
What is the latest on this @eve-mem ?

eve-mem:
No progress on this from me - thank you for the nudge.

PR #1791, opened 2025-04-27
Add in coverage recording to the test process

ikelos:
Creates an artifact for the code coverage.  I've no idea if this will incur billing or not, but added in the people most knowledgable about the test framework to check this over...


ikelos:
Ok, no comments so getting this merged and we can look through the results it creates and see if anything turns up...

PR #1793, opened 2025-04-28
cli: add debug log of plugin version when the plugin is successfully constructed

### LINKED ISSUE #1685: Debug output the plugin name and version number when run
Be good to know the exact version of a plugin being run (ideally whenever their run method is called, but if the caller has to do it then so be it).

eve-mem:
Hello :wave: 

This PR aims to help with issue https://github.com/volatilityfoundation/volatility3/issues/1685.

It logs the version of a plugin to DEBUG just before it's run.

I'm using f-strings in the debug message because it makes it easier to read as suggested by @ikelos 

Thanks
:fox_face:

### DIFF
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -505,6 +505,9 @@ class CommandLine:
         try:
             # Construct and run the plugin
             if constructed:
+                vollog.debug(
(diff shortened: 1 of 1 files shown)


ikelos (approved):
Looks good, thanks!

PR #1794, opened 2025-05-05
Volshell breakpointing

ikelos:
This is an initial implementation of breakpointing for volshell.  Breakpoints can be set on a specific layer directly, or on the lowest mapped layer of the offset (so that accesses through lower layers will still breakpoint).

Note: Volatility is very careful about only reading the data it requires, meaning that even creating an object at an offset won't necessarily trip the breakpoint.

### DIFF
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -3,6 +3,7 @@
 #
 import binascii
 import code
+import functools
 import io
 import random
 import string
@@ -187,6 +188,7 @@ class Volshell(interfaces.plugins.PluginInterface):
     def construct_locals(self) -> List[Tuple[List[str], Any]]:
         """Returns a listing of the functions to be added to the environment."""
         return [
+            (["bp", "breakpoint"], self.breakpoint),
             (["dt", "display_type"], self.display_type),
             (["db", "display_bytes"], self.display_bytes),
             (["dw", "display_words"], self.display_words),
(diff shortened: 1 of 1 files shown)


eve-mem:
Ooh wow. I would use this, very cool. I'd need to have a play.

volatility3/cli/volshell/generic.py
```diff
@@ -797,6 +798,51 @@ def create_configurable(
...
+        if not hasattr(layer.read, "breakpoints"):
+            # Layer read is not yet wrapped
+            def wrapped_read(offset: int, length: int, pad: bool = False) -> bytes:
+                original_read = getattr(wrapped_read, "original_read")
+                for breakpoint in getattr(wrapped_read, "breakpoints"):
+                    if (offset <= breakpoint) and (breakpoint < offset + length):
+                        import pdb
+
+                        pdb.set_trace()
+                        print("Hit breakpoint")
```
eve-mem: I think it's worth flipping the order of the print and the set_trace(), that way the print happens normally, then you're in the python debug.

I think it might be worth changing the print to just to save some future headaches? It makes the print quite long though? 
```
print("Hit breakpoint, entering python debugger. To continue running without the debugger use the command continue")
```

volatility3/cli/volshell/generic.py
```diff
@@ -797,6 +798,51 @@ def create_configurable(
...
+                return original_read(offset, length, pad)
+
+            setattr(wrapped_read, "breakpoints", set())
+            setattr(wrapped_read, "original_read", layer.read)
+            setattr(layer, "read", wrapped_read)
+
+        # Add the new breakpoint
+        print(f"Setting breakpoint {address:x} on {layer.name}")
```
eve-mem: Personally a fan of always including `0x` before hex numbers just so it's super clear for everyone:
e.g.
```
print(f"Setting breakpoint 0x{address:x} on {layer.name}")
```

eve-mem (commented):
Hey @ikelos I've really enjoyed this feature. I've made some tiny suggestions based on my experience using it. 

One that that I would personally find useful is being able to put a break point on a range of addresses (e.g. a whole page or two). I'm not sure how commonly that might be used. My use case is wanting to break whenever anything touches an entire structure rather than a particular address. I think adding that might be a little more complicated and I'd rather not hold up this feature if you think it'll take a while.  

It would also be useful to include a way to remove breakpoints, even if it was just a nuclear option to remove them all? Equally a way to view the existing breakpoints - again these could come later. 

I'm a huge fan of volshell so great to see new features - thank you!

eve-mem:
If it's helpful this could be used to list the breakpoints (it's what I'd been using via rs())

```python
for layer_name in self.context.layers:
    layer = self.context.layers[layer_name]
    if hasattr(layer.read, 'breakpoints'):
        for breakpoint_address in layer.read.breakpoints:
            print(f"{layer_name}: 0x{breakpoint_address:x}")
```

ikelos:
Thanks, I added in `bl` to list breakpoins and `bc` to clear breakpoints (tying to mimic windbg to a degree)...

eve-mem:
Great, that looks awesome. Thanks!

PR #1795, opened 2025-05-05
Linux: prevent "forward slashes only" dentry names

Abyss-W4tcher:
Hi,

While performing mass testing we observed that many lsof entries had "/" or "//" names. This PR fixes the issue by ensuring that at least one non-empty entry was found in the dentry walk.

### DIFF
--- a/volatility3/framework/symbols/linux/__init__.py
+++ b/volatility3/framework/symbols/linux/__init__.py
@@ -190,9 +190,13 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
 
             parent = dentry.d_parent
             dname = dentry.d_name.name_as_str()
-            path_reversed.append(dname.strip("/"))
+            dname_stripped = dname.strip("/")
+            if dname_stripped:
+                path_reversed.append(dname_stripped)
             dentry = parent
 
+        if path_reversed == []:
+            return ""
         path = "/" + "/".join(reversed(path_reversed))
         return path
 
--- a/volatility3/framework/symbols/linux/__init__.py
+++ b/volatility3/framework/symbols/linux/__init__.py
@@ -169,6 +169,7 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
             return ""
 
         path_reversed = []
+        smeared = False
         while (
             dentry
             and dentry.is_readable()
@@ -190,14 +191,16 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
 
             parent = dentry.d_parent
             dname = dentry.d_name.name_as_str()
-            dname_stripped = dname.strip("/")
-            if dname_stripped:
-                path_reversed.append(dname_stripped)
+            # empty dentry names are most likely
+            # the result of smearing
+            if not dname:
+                smeared = True
+            path_reversed.append(dname.strip("/"))
             dentry = parent
 
-        if path_reversed == []:
-            return ""
         path = "/" + "/".join(reversed(path_reversed))
+        if smeared:
+            return f"<potentially smeared> {path}"
         return path
 
     @classmethod


Abyss-W4tcher:
I'm investigating to determine if "/" stripped paths should still have been treated as "/", but that seems a bit weird.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -190,9 +190,13 @@ def do_get_path(cls, rdentry, rmnt, dentry, vfsmnt) -> Union[None, str]:
...
             parent = dentry.d_parent
             dname = dentry.d_name.name_as_str()
-            path_reversed.append(dname.strip("/"))
+            dname_stripped = dname.strip("/")
+            if dname_stripped:
+                path_reversed.append(dname_stripped)
             dentry = parent
 
+        if path_reversed == []:
+            return ""
```
ikelos: Is this the right return value?  If `path_reversed` is empty we presumably want to throw an exception and/or return `None` so that people can tell an error's occurred?

ikelos (commented):
I think one of the tests now yields different results, but it should be investigated to make sure it's returning the intended response.

volatility3/framework/symbols/linux/__init__.py
Abyss-W4tcher: Yes there is a catch to this, as it should at least return "/" if we hit the break in the global root if-block. The code made it implicit and I didn't catch that at first.

In fact, the "/" and "//" issue is most likely the result of smear, but in the end I don't think returning "" is a good idea anymore. Here is what I have in mind right now, and I prefer discussing the idea before pushing the code:

```diff
+        smeared = False
         while (
             dentry
             and dentry.is_readable()
@@ -190,10 +191,16 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
 
             parent = dentry.d_parent
             dname = dentry.d_name.name_as_str()
+            # empty dentry names are most likely
+            # to be the result of smearing
+            if not dname:
+                smeared = True
             path_reversed.append(dname.strip("/"))
             dentry = parent
 
         path = "/" + "/".join(reversed(path_reversed))
+        if smeared:
+            return f"smeared path: {path}"
         return path
```

I though that keeping those broken entries might still be of interest to the user, while adding a disclaimer before the path like we already do here:

https://github.com/volatilityfoundation/volatility3/blob/51f0a3081d5fad41efb1c09d960fe6e374ab4b2d/volatility3/framework/symbols/linux/__init__.py#L233

to make them aware that we detected something off. This would still output paths like `/my_non_smeard_entry/\xffSMEAR\xff/my_non_smeared_entry`, that can at least help identify the inode number even if a part of the name is broken. Paths are "" or start with an "/", so no risk of confusing "smeared path:" with the actual path?

ikelos: Yeah, ok.  I'm a little concerned when returning data that contains additional information (like the starting tag) because it could be used for confusion by a malicious actor (although the chances are extremely low).  I think if we already have code that prepends with `<wanring> thing` then we should do the same here, so something like `f"<potentially smeared> {path}"`.

Abyss-W4tcher:
This PR is ready for a final review, whenever you have the time 👍

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -190,10 +191,16 @@ def do_get_path(cls, rdentry, rmnt, dentry, vfsmnt) -> Union[None, str]:
 
             parent = dentry.d_parent
             dname = dentry.d_name.name_as_str()
+            # empty dentry names are most likely
+            # the result of smearing
+            if not dname:
+                smeared = True
             path_reversed.append(dname.strip("/"))
```
eve-mem, line 198: Do you still need to add the empty dname to the list? E.g. in the smeared case?

eve-mem (commented):
Just one tiny question for me about if you need to add in the empty dnames to the list.

volatility3/framework/symbols/linux/__init__.py
Abyss-W4tcher: So I thought that keeping it would output "//", and indicate the part of the path that is most likely smeared, or "missing" in that case.

Abyss-W4tcher: e.g. "/root1/root2//root4" where we can deduce root3 is smeared.

eve-mem: That makes sense. Given that `/foo//////bar` would work normally for `/foo/bar` (at least in bash) I'm not sure I'd reliease that it was the third folder in your example that was smeared. Do you think having some marker for the missing folder name would work? Or is that too messy?

Abyss-W4tcher: Technically a valid inode path with multiple "///" as a dentry name shouldn't be valid, and be understood as part of the smearing. However I get your point that most users would think it is totally normal.

Abyss-W4tcher: I like your idea, something like <empty_part>? However the point ikelos made about actors putting <empty_part> as part of their paths to confuse people should apply there.

I'm sure there is an efficient way to do this though, happy to brainstorm about it.

eve-mem: Yeah, i think most would miss that nuance. Maybe just a comment or some help text that explains what the `//` means in this context is enough? Or some marker `/<Unreadable value>/`? I really don't know what's best, or even if it's worth worrying about.

eve-mem: Yes agreed, as soon as you put `/unknown/` into the code that's exactly what the malware will start using.

ikelos: If there's a character that isn't a valid path character, we could use that instead of just `<` and `>`?  There's presumably some characters the filesystem won't accept that would work.  It just needs to be possible to distinguish between sometime trying to confuse us and an actual smear issue...

ikelos (approved):
I think as long as we're identifying it's potentially smeared should be enough (and it's before the initial slash, which means a bad actor can't cause the tool to output it without causing actual smear in some way...

volatility3/framework/symbols/linux/__init__.py
Abyss-W4tcher: I think Linux filesystems basically accept any character other than / ?

Abyss-W4tcher: Ok so I guess we keep it that way for now, maybe a bit less user friendly but without additional "attack surface".

ikelos:
Huh, I thought backtick or `*` would work, but they're both fine apparently...  5:S

Abyss-W4tcher:
By the way that's why it's sometimes possible to do command injection into poorly coded parsers through filenames, with backticks :)

PR #1802, opened 2025-05-09
Linux: Add comment for  explaining the behaviour with smear

eve-mem:
Hello :wave: 

Just a small continuation of https://github.com/volatilityfoundation/volatility3/pull/1795/ where @Abyss-W4tcher fixed an issue with smear with the linux util `do_get_path`. 

I've just added a comment to explain what the results mean as discussed in the comments on the original PR. It's the kind of thing I'd forget in the future so having a comment like this in the code is useful to me. 

However I don't want to clutter the code with unnecessary comments, so if you both @ikelos and @Abyss-W4tcher think this isn't needed that is okay with me. 

Thanks as always
:fox_face:


ikelos (approved):
Looks good, thanks  5:)

PR #1807, opened 2025-05-14
Windows Thrdscan: Fix broken tuple unpacking

dgmcdona:
Timeliner fails due to an incorrect unpacking of this tuple, which needs
3 additional dictionary items for start path, win32 start path, and
win32 start address.

### DIFF
--- a/test/plugins/windows/windows.py
+++ b/test/plugins/windows/windows.py
@@ -58,6 +58,14 @@ class TestWindowsPslist:
         }
         assert test_volatility.match_output_row(expected_row, json.loads(out))
 
+class TestWindowsTimeliner:
+    def test_windows_specific_timeliner(self, volatility, python):
+        image = WindowsSamples.WINDOWSXP_GENERIC.value.path
+        rc, out, _err = test_volatility.runvol_plugin(
+            "timeliner.Timeliner", image, volatility, python
+        )
+        assert rc == 0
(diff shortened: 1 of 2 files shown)


ikelos:
Could we please get a minimal test with the existing linux image as part of the PR to ensure this doesn't happen again please?

dgmcdona:
I added a test locally, but it took ~226 seconds to run. I'm happy to push that if you're okay with adding that much time to test runs, but what if we also/instead updated `_generator()` to return a named tuple instead, and access the values by field name in `generate_timeline()`? That would make this kind of bug a little harder to trigger.

ikelos:
Errrr, yeah, I'm happy with that?  I dunno if that should become a core coding style (in which case we should add it to the `CODING_STYLE.md` and explain the rationale so people know *why* they're doing it).  It might also bulk out a bunch of plugins unnecessarily, so we might want to figure out the cheapest way of doing it (namedtuples/dataclasses/etc).

atcuno:
Tests taking another 4 minutes seem fine with me to avoid these issues.

ikelos (approved):
Ok, looks good, thanks  5:)

ikelos:
We can still do the namedtuple change later if you think it'd be an improvement, but I'd like to keep API churn to a minimum so only if you think it'd be valuable.

PR #1810, opened 2025-05-16
Align SeCreateTokenPrivilege with the other privileges

j-t-1:


j-t-1:
Line 612 I did not change, unsure why it is highlighted on the Files changed tab.

Should Privilege Number 1 be inserted?

ikelos:
I don't know if that's a special value used in some way, but happy to accept the alignment fix, thanks.  5:)

PR #708, opened 2022-04-22
Fix: `PDBRetriever` class name.

digitalisx:
Hello, everyone in the community! 🙂
I think it should be PDB Retriever.

PR #1815, opened 2025-05-22
fix: typos

ddogfoodd:


atcuno:
Thank you for the fixes!

This part of changing a class and method name will need some versioning I assume. Will let @ikelos chime in on that:

https://github.com/volatilityfoundation/volatility3/pull/1815/files#diff-e8b8f61c12ac37e72be940fc3ca27d80fe2a966911d3ba60d8294664bf7e5a68L944

digitalisx:
About Retriever, We discussed it 2 years ago (https://github.com/volatilityfoundation/volatility3/pull/708)
Can we change at this time?

ddogfoodd:
Thanks, didn't know there was some history there.
Let me know if I should revert those.

development/compare-vol.py
```diff
@@ -339,7 +339,7 @@ def run_test(self, plugin: VolatilityPlugin, image: VolatilityImage):
         "--vol3path",
         type=str,
         default=os.path.join(os.getcwd(), "volatility3"),
-        help="Path ot the volatility 3 directory",
+        help="Path of the volatility 3 directory",
```
ikelos:
```suggestion
        help="Path to the volatility 3 directory",
```
I think this was originally supposed to say path to the directory, rather than of.

development/pdbparse-to-json.py
```diff
@@ -21,9 +21,9 @@
     logger.addHandler(console)
 
 
-class PDBRetreiver:
+class PDBRetriever:
```
ikelos: We don't know what else taps into these classes, so you'd need to add something like:

```
class PDBRetreiver(PDBRetriever):
```
And the same for methods.  We got some wrappers in `volatility3.framework.deprecation` that should allow you to deprecate a method (and one for renaming a class).

volatility3/framework/symbols/windows/pdbconv.py
```diff
@@ -941,8 +941,8 @@ def determine_extended_value(
         return name_str, value, excess
 
 
-class PdbRetreiver:
-    def retreive_pdb(
+class PdbRetriever:
```
ikelos: Same as above (and actually, these ones are more important, the development ones I'd be ok with just breaking, since they shouldn't be used directly by anyone).

ikelos (commented):
Wow!  We really typoed a lot of things over the history of the project!  (I think my favourite is "instrunction")  5:P

All the comment/docstring ones look good, any that impact that actual code (the `PDBRetriever`) will need to be renamed leaving the old version around for a couple of years in case someone's code relies on it.  Luckily we've got helper wrappers to do that in `volatility3.framework.deprecation` so should just be a case of plumbing those in and then it's good to go.  If you want the rest to go in first and leave the class names for another PR, that's fine too...

ikelos:
> This part of changing a class and method name will need some versioning I assume. Will let @ikelos chime in on that:

We haven't been versioning anything around the PDB stuff, but I think it can just be deprecated rather than versioned.  That should require anything above a PATCH bump, since all old code will still work, and if we leave it in place for a couple of years, then anyone using the old one should see the warning and we can just drop it outright...

development/pdbparse-to-json.py
ddogfoodd: I am not sure enough that I understand how to do that correctly, so I am leaving it out of this PR

ikelos (approved):
Thanks, everything in this is fine now.  If you'd like to do another PR doing the deprecation of the misspelled class, as long as we deprecate it that should be fine.  Thanks again for your eye for detail!

PR #1816, opened 2025-05-23
Add Stack Plugin

the-rectifier:
A plugin that dumps the stack for each (running) thread.

It grabs the current `RSP` from the TrapFrame `_ETHREAD->_KTHREAD->_KTRAP_FRAME.Rsp`, and the Stack base from `_ETHREAD->_KTHREAD->_TEB.NT_TIB.StackBase`

This could (and probably should) be part of another plugin like `memmap`, but I figured I would submit it for review and then change it.

Edit: No issues with the `_NT_TIB` object, see referenced 'issue'

### DIFF
--- a/volatility3/framework/plugins/windows/stack.py
+++ b/volatility3/framework/plugins/windows/stack.py
@@ -1,0 +1,118 @@
+import struct
+
+from typing import List
+
+from volatility3.framework.symbols.windows import extensions
+from volatility3.framework.configuration import requirements
+from volatility3.framework import interfaces, renderers, constants
+from volatility3.framework.renderers import format_hints
+from volatility3.plugins.windows import pslist
+
+
+class Stack(interfaces.plugins.PluginInterface):
+    """Lists the Stack boundaries and dump Stack"""
+
+    _required_framework_version = (2, 0, 0)
+    _version = (3, 0, 1)
+
+    @classmethod
+    def get_requirements(cls):
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Windows Kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.VersionRequirement(
+                name="pslist", component=pslist.PsList, version=(3, 0, 0)
+            ),
+            requirements.ListRequirement(
+                name="pid",
+                element_type=int,
+                description="Process IDs to operate on",
+                optional=False,
+            ),
+            requirements.BooleanRequirement(
+                name="dump",
+                description="Whether to dump the stack",
+                default=False,
+                optional=True,
+            ),
+        ]
+
+    def _generator(self):
+        file_output = "Disabled"
+
+        kernel = self.context.modules[self.config["kernel"]]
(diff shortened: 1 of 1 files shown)


eve-mem:
It might be worth making an issue for the dt() problem. Perhaps i broke something with https://github.com/volatilityfoundation/volatility3/pull/1748

SolitudePy:
maybe this plugin can be enhanced to list the actual callstack?

volatility3/framework/plugins/windows/stack.py
```diff
@@ -0,0 +1,118 @@
...
+
+            for thread in active_thread_list:
+                trap_frame = thread.Tcb.TrapFrame.dereference()
+                thread_rsp = trap_frame.Rsp
+
+                # first entry of TEB is NT_TIB
+                # Stack Base is at NT_TIB + 8
+                # https://github.com/wine-mirror/wine/blob/master/include/winternl.h#L494
+                # https://github.com/wine-mirror/wine/blob/master/include/winnt.h#L2445
+                stack_base = struct.unpack(
```
ikelos, line 79: Please try to avoid reading bytes directly.  Either make a struct or at least instantiate a new object of the right type for windows on it.  Having to 0 index the results of unpack basically throws away all the extra goodness we provide through the volatility model (keeping track of the type, the offset, which layer it came from, etc).

The entire volatility framework is a giant parser, so if you're resorting to struct.unpack yourself please consider if there's *no* other way of doing it (or if you're doing it so often efficiency comes into question, which ones per thread, isn't the case here).

volatility3/framework/plugins/windows/stack.py
```diff
@@ -0,0 +1,118 @@
...
+                    "Q", proc_layer.read(offset=thread.Tcb.Teb + 8, length=8)
+                )[0]
+
+                stack_size = stack_base - thread_rsp
+
+                if self.config["dump"]:
+                    fname = f"{proc.UniqueProcessId}.{thread.Cid.UniqueThread}.dmp"
+                    stack = proc_layer.read(offset=thread_rsp, length=stack_size)
```
ikelos, line 87: This is a reasonable use of read, because it's a big chunk of unstructured (or unknown structure) data, that you just want to write out as a big blob and aren't going to use again inside volatility.

ikelos (commented):
Thanks, this looks reasonable, but from a code perspective makes me cry when people use volatility to scalpel their way to some data and then use a machete to hack it raw from the layer!  5:P

the-rectifier:
Hello @ikelos, thank you for all the pointers (and apologies for the tears). I will make the code more modular and also create the `NT_TIB` object rather than reading an address of it directly from the layer.

Would migrating (a part of) this over to `extensions.ETHREAD` make sense? I was thinking of a `get_stack()` method that this plugin will call on each constructed `_ETHREAD` object:

```
for thread in active_thread_list:
	stack = thread.get_stack()
```

the-rectifier:
@SolitudePy, I think this is a nice addition, but I'm afraid I don't have the time to debug a process and find a way to follow the call stack, at least not right now. Maybe I will revisit it when my program is a bit lighter, but feel free to contribute if you want!

PR #1821, opened 2025-05-28
Add a deprecation warning for PluginRequirement

eve-mem:
There was a brief discussion on slack with @the-rectifier and @ikelos about the use of PluginRequirement or VersionRequirement. 

@ikelos detailed that PluginRequirement should be going away and VersionRequirement should be used instead. This PR uses the new deprecation  features to warn when PluginRequirement is used. 

I'm not sure that 2026-06-01 is a sensible date - but a year away felt okay? 

Thanks as always!
:fox_face:

### DIFF
--- a/volatility3/framework/configuration/requirements.py
+++ b/volatility3/framework/configuration/requirements.py
@@ -593,7 +593,7 @@ class VersionRequirement(interfaces.configuration.RequirementInterface):
     def matches_required(
         cls, required: Tuple[int, ...], version: Tuple[int, int, int]
     ) -> bool:
-        versionutils.matches_required(required, version)
+        return versionutils.matches_required(required, version)
 
 
 @deprecation.renamed_class(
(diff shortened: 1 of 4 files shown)


volatility3/framework/configuration/requirements.py
```diff
@@ -593,13 +593,14 @@ def unsatisfied(
     def matches_required(
         cls, required: Tuple[int, ...], version: Tuple[int, int, int]
     ) -> bool:
-        if len(required) > 0 and version[0] != required[0]:
-            return False
-        if len(required) > 1 and version[1] < required[1]:
-            return False
-        return True
+        versionutils.matches_required(required, version)
```
ikelos:
```suggestion
        return versionutils.matches_required(required, version)
```
Subtle but... breaks everything if you don't return it...  5;P

ikelos (commented):
Looks really good, just one minor, tiny, negligible, insignificant, hardly-worth-mentioning typo-that-breaks-everything, but very easy to fix!  5;D  5:P

volatility3/framework/configuration/requirements.py
eve-mem: Thank you! I was really scratching my head this morning.

eve-mem:
I've fixed that subtle but killer missing return! Thank you <3

ikelos (approved):

PR #1823, opened 2025-05-29
Tweak comments

j-t-1:
Also remove zeroes starting a slice.

### DIFF
--- a/volatility3/framework/plugins/windows/malfind.py
+++ b/volatility3/framework/plugins/windows/malfind.py
@@ -112,7 +112,7 @@ class Malfind(interfaces.plugins.PluginInterface):
         code.
 
         Args:
-            context: The context to retrieve required elements (layers, symbol tables)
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/malfind.py
```diff
@@ -112,13 +112,13 @@ def list_injection_sites(
         code.
 
         Args:
-            context: The context to retrieve required elements (layers, symbol tables) from
+            context: The context to retrieve required elements (layers, symbol tables)
```
ikelos:
```suggestion
            context: The context from which to retrieve required elements (layers, symbol tables)
```
Retrieve requires a location, but ending in a preposition isn't great English, so technically this include "from which" somewhere earlier in the sentence.

ikelos (commented):
One little grammatical point, but otherwise it all looks good, thanks!  5:)

PR #1825, opened 2025-06-01
pebmasquerade plugin

SolitudePy:
Hello, trying my way around os internals & memory :P

```
(venv) ubuntu@ubuntuPC:~/Dev/volatility3$ vol -f ~/dumps/peb_masq_dump.raw -r json windows.pebmasquerade | jq 'map(select(.Notes != "OK"))'
Volatility 3 Framework 2.26.2
[               
  {
    "EPROCESS_ImageFileName": "powershell_ise",
    "EPROCESS_SeAudit_ImageFileName": "\\Device\\HarddiskVolume3\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell_ise.exe",
    "Notes": "['Potential PEB.ImageFilePath Spoofing: EPROCESS=powershell_ise;PEB=explorer.exe', 'Potential PEB.CommandLine Spoofing: EPROCESS=powershell_ise;PEB=explorer.exe', 'Potential PEB.ImageFilePath Spoofing (via _EPROCESS.SeAuditProcessCreationInfo): EPROCESS=windows/system32/windowspowershell/v1.0/powershell_ise.exe;PEB=windows/explorer.exe']",
    "PEB_CommandLine_Path": "C:\\Windows\\explorer.exe",
    "PEB_ImageFilePath": "C:\\Windows\\explorer.exe",
    "PID": 11096,
    "ProcessName": "powershell_ise",
    "__children": []
  },
  {
    "EPROCESS_ImageFileName": "peb_masquerade",
    "EPROCESS_SeAudit_ImageFileName": "\\Device\\HarddiskVolume3\\Users\\Robot\\Dev\\peb_masquerade.exe",
    "Notes": "['Potential PEB.ImageFilePath Spoofing: EPROCESS=peb_masquerade;PEB=notepad.exe', 'Potential PEB.CommandLine Spoofing: EPROCESS=peb_masquerade;PEB=notepad.exe', 'Potential PEB.ImageFilePath Spoofing (via _EPROCESS.SeAuditProcessCreationInfo): EPROCESS=users/robot/dev/peb_masquerade.exe;PEB=windows/system32/notepad.exe']",
    "PEB_CommandLine_Path": "C:\\windows\\system32\\notepad.exe",
    "PEB_ImageFilePath": "C:\\windows\\system32\\notepad.exe",
    "PID": 12372,
    "ProcessName": "peb_masquerade",
    "__children": []
  }
]
```


SolitudePy:
added _UNICODE_STRING length checks. 
ProcessParameters.CommandLine introduced 2 "false positives" on my test:
```
(venv) ubuntu@ubuntuPC:~/Dev/volatility3$ vol -f ~/dumps/peb_masq_dump.raw -r json windows.pebmasquerade | jq 'map(select(.Notes != "OK"))'
Volatility 3 Framework 2.26.2
  warnings.warn(
[rogress:  100.00               PDB scanning finished                        
  {
    "EPROCESS_ImageFileName": "WmiPrvSE.exe",
    "EPROCESS_SeAudit_ImageFileName": "\\Device\\HarddiskVolume3\\Windows\\System32\\wbem\\WmiPrvSE.exe",
    "Notes": "['PEB.CommandLine Length Mismatch: Commandline=C:\\Windows\\system32\\wbem\\wmiprvse.exe, Length=48, MaximumLength=48, Actual=37']",
    "PEB_CommandLine_Path": "C:\\Windows\\system32\\wbem\\wmiprvse.exe",
    "PEB_ImageFilePath": "C:\\Windows\\system32\\wbem\\wmiprvse.exe",
    "PID": 2640,
    "ProcessName": "WmiPrvSE.exe",
    "__children": []
  },
  {
    "EPROCESS_ImageFileName": "audiodg.exe",
    "EPROCESS_SeAudit_ImageFileName": "\\Device\\HarddiskVolume3\\Windows\\System32\\audiodg.exe",
    "Notes": "['PEB.CommandLine Length Mismatch: Commandline=C:\\Windows\\system32\\AUDIODG.EXE 0x540, Length=43, MaximumLength=43, Actual=37']",
    "PEB_CommandLine_Path": "C:\\Windows\\system32\\AUDIODG.EXE",
    "PEB_ImageFilePath": "C:\\Windows\\system32\\AUDIODG.EXE",
    "PID": 11384,
    "ProcessName": "audiodg.exe",
    "__children": []
  }
]
```

volatility3/framework/plugins/windows/malware/pebmasquerade.py
```diff
@@ -0,0 +1,371 @@
...
+            ),
+            requirements.ListRequirement(
+                name="pid",
+                element_type=int,
+                description="Process ID to include (all other processes are excluded)",
+                optional=True,
+            ),
+        ]
+
+    @staticmethod
```
ikelos: Please use classmethods rather than staticmethods.  Staticmethods can't determine what class they're a part of, therefore can't access any version information about themselves, which can be important (for instance, for deprecating it in the future).

volatility3/framework/plugins/windows/malware/pebmasquerade.py
```diff
@@ -0,0 +1,371 @@
...
+    def _get_cmdline_image(cmdline: str) -> Union[str, PureWindowsPath]:
```
ikelos: Why wouldn't we just return a string in all instances?  By returning a path we're making people potentially test whether it's a path object or not?  When would we return a string?

volatility3/framework/plugins/windows/malware/pebmasquerade.py
```diff
@@ -0,0 +1,371 @@
...
+        """Extract the executable path from a command line string.
+
+        Args:
+            cmdline (str): The command line string to parse.
+
+        Returns:
+            Union[str, PureWindowsPath]: The executable path as a string or PureWindowsPath.
+        """
+        if not cmdline:
```
ikelos: Error conditions should throw an exception or return None.  An empty string is hard to differentiate from a legitimately empty string.

volatility3/framework/plugins/windows/malware/pebmasquerade.py
```diff
@@ -0,0 +1,371 @@
...
+            executable = (
+                first_token_match.group(1)
+                or first_token_match.group(2)
+                or first_token_match.group(3)
+            )
+            return PureWindowsPath(executable).name + ".exe"
+
+        return ""
+
+    @staticmethod
```
ikelos: Again, use classmethods instead of staticmethods please.

volatility3/framework/plugins/windows/malware/pebmasquerade.py
```diff
@@ -0,0 +1,371 @@
...
+            cmdline (str): The command line string to parse.
+
+        Returns:
+            Union[str, PureWindowsPath]: The executable path as a string or PureWindowsPath.
+        """
+        if not cmdline:
+            return ""
+
+        # Regex to extract first .exe ending string (handles quotes, paths, no quotes)
+        match = re.search(r'(?i)(["\']?)([^"\']*?\.exe)\1(?=\s|$)', cmdline)
```
ikelos: This feels brittle and impenetrable.  This will also prevent executables like `pat o'malley.exe` which I believe would be a legitimate path.  Let's not code all this complexity into a regex, but clarify exactly the process we're going through to find the first parameter (which would return the same results whether it ended in .exe or not, so splitting it into two cases seems pointless?).

volatility3/framework/plugins/windows/malware/pebmasquerade.py
```diff
@@ -0,0 +1,371 @@
...
+        if match:
+            exe_path = match.group(2)
+            return PureWindowsPath(exe_path)
+
+        # If no .exe found, extract the first token (handles quotes)
+        # Matches either "quoted string" or unquoted word
+        first_token_match = re.match(r'\s*(?:"([^"]+)"|\'([^\']+)\'|(\S+))', cmdline)
```
ikelos: Again, very difficult to parse and verify that no errors have snuck in.  Better to form a small loop where you can clearly keep track of levels of quotes and things like that.

volatility3/framework/plugins/windows/malware/pebmasquerade.py
```diff
@@ -0,0 +1,371 @@
...
+                first_token_match.group(1)
+                or first_token_match.group(2)
+                or first_token_match.group(3)
+            )
+            return PureWindowsPath(executable).name + ".exe"
+
+        return ""
+
+    @staticmethod
+    def _are_paths_equal(device_path: str, drive_path: str) -> Tuple[bool, str, str]:
```
ikelos: Given the return values, this feels like it's doing too much?  Either people will use the two additional results (and carrying out their own equality test is trivial) or they only care about the equality and unpacking a tuple to get to it is a bit pointless.  Consider rethinking what this is trying to achieve...

volatility3/framework/plugins/windows/malware/pebmasquerade.py
```diff
@@ -0,0 +1,371 @@
...
+
+        return (
+            eprocess_imagefilename,
+            eprocess_seaudit_imagefilename,
+            peb_imagefilepath,
+            peb_cmdline,
+        )
+
+    def _generator(self):
+        pid_filter = pslist.PsList.create_pid_filter(self.config.get("pid", None))
```
ikelos: Please parameterize config options, so the generator takes parameters, rather than needing to pull them from its own config.  It's not critical but it somewhat locks the generator to this class, which may not be what you want.

volatility3/framework/plugins/windows/malware/pebmasquerade.py
```diff
@@ -0,0 +1,371 @@
...
+
+        for proc in pslist.PsList.list_processes(
+            context=self.context,
```
ikelos: Similarly for the context and the kernel name.

volatility3/framework/plugins/windows/malware/pebmasquerade.py
```diff
@@ -0,0 +1,371 @@
...
+            if isinstance(peb_cmdline, str):
+                try:
+                    peb_cmdline_path_render = str(
+                        PebMasquerade._get_cmdline_image(peb_cmdline)
+                    )
+                except Exception as e:
+                    vollog.debug(
+                        "Error extracting command line path for PID %d: %s",
+                        proc_id,
+                        str(e)[:50],
```
ikelos: This is a debugging message, and this limits the length of the exception explanation, not the length of the filename?  Why limit it at all?  This is only likely to be seen by developers, all will frustrate them more by shortening the error message than it would by displaying the entire error message.

volatility3/framework/plugins/windows/malware/pebmasquerade.py
```diff
@@ -0,0 +1,371 @@
...
+            yield (
+                0,
+                (
+                    proc_id,
+                    proc_name_for_row,
+                    eprocess_imagefilename,
+                    eprocess_seaudit_imagefilename,
+                    peb_imagefilepath,
+                    peb_cmdline_path_render,
+                    "[" + ", ".join(notes) + "]" if notes else "OK",
```
ikelos: This doesn't allow programs consuming the results to parse the data easily.  Rather than designing the output for human consumption, each boolean should be returned as a field and the UI allowed to display the results as it sees fit.

ikelos (commented):
A number of little issues around this one, it feels like a lot of string manipulation that don't necessarily feel robust.  I'll give it another look once you've addressed the comments though...

volatility3/framework/plugins/windows/malware/pebmasquerade.py
SolitudePy: @ikelos you mean, every type of note should be a different column? then its a bit hard to supply the values because its not boolean... what I meant to achieve is to create a list as str that can easily be parsed by python and such. but I guess its not the best approach

volatility3/framework/plugins/windows/malware/pebmasquerade.py
SolitudePy: this function is not that trivial because it checks if a windows device path (i.e \Device\HardDIsk03\path) equals to its drive path (i.e C:\path), and both of the new paths are returned as well because they are then used for the note, maybe I should change the function name & description? also you are saying it is better if I simply returned the results and have the caller check if its equal? thats possible too but then _are_paths_equal need to be changed as well to something like _normalize_paths

volatility3/framework/plugins/windows/malware/pebmasquerade.py
ikelos: Well, if they're strings that need to be strings, then I guess use them, but the point is putting them in a list rather than each having its own column will make it much harder for other automated systems to make use of the output.  Ideally any that could be represented as a boolean should be to make the columns as thin as possible (for humans to read).

SolitudePy: I dont see how it can be each in its own column, but the 'OK' Note with json renderer and a tool like jq like I previewed my test, makes it quite easy to go through results, I dont think it can be converted to human readable grid as long as there are notes... I can remove the notes and then people need to check themselves if names, length etc are not equal but thats also a pain

volatility3/framework/plugins/windows/malware/pebmasquerade.py
SolitudePy: @ikelos, problem is the commandline in volatility is returned as a whole string, so if a path has spaces in it, it wont be quoted from my testing, that makes it super hard to work with all edge cases, and you are right that `pat o'malley.exe` won't work. I wanted to use the shlex module but thats also a problem because of the quoted thing I mentioned. do you have an idea how that could be fixed elegantly?

SolitudePy: Examples:
```
command = r"C:\Program Files\program.exe /arg1 Hello"
args = shlex.split(command)
print(args)
# Output: ['C:Program', 'Filesprogram.exe', '/arg1', 'Hello']
```
the file path need to be quoted additionally i.e:
```
command = r"'C:\Program Files\program.exe' /arg1 Hello"
args = shlex.split(command)
print(args)
# Output: ['C:\\Program Files\\program.exe', '/arg1', 'Hello']
```

volatility3/framework/plugins/windows/malware/pebmasquerade.py
SolitudePy: at first I wanted to render the full path, so then I would have had to turn it back to `PureWindowsPath`, but I agree we should return a string in all instances.

volatility3/framework/plugins/windows/malware/pebmasquerade.py
ikelos: Sorry it took me so long to get back to you.  Hmmm, the difficulty with the representation is that there's no way of distinguishing between:
```
['C:\some path with spaces\exe.exe', 'argument\with\slash', 'another thing']
```
and
```
['C:\some path with space\exe.exe argument\with\slash', 'another thing']
```

So either where we're getting the data from doesn't have enough information, or we're losing something somewhere along the way.  Either way, it's not a good plan for a forensics tool to choose one or the other repesentation, no matter how helpful it's trying to be, because malware could just intentionally choose the route that will get misinterpreted.

volatility3/framework/plugins/windows/malware/pebmasquerade.py
ikelos: It's a pain from the human's perspective not to have everything done by volatility, but from another tool's perspective it's a total pain to have to parse an inefficient text based "notes" format.  I realize we should be biased towards humans, but where does it end, if we keep finding new notes, the string just get longer and longer.  There's no easy way of saying "I care about these notes, but not those notes", so it doesn't scale.  It looks like there's just four fields, that have been crammed into a single text field along with extraneous words?  I'm suggesting columns for each thing you put in those strings, and then users can choose which columns to see with either the `--filter` flag or the `--hide-columns` flags.  Any that have no data should just be filled with a `BaseAbsentValue` like NotAvailable or something.  You then also aren't repeating lots and lots of pointless text with each entry...

volatility3/framework/plugins/windows/malware/pebmasquerade.py
SolitudePy: What are u suggesting then? skip this logic since neither shlex or regex are fullproof approaches.

ikelos: Yeah, if the incoming spaces aren't escaped, then there's no way of telling, and we should just return it untouched for the analyst to figure out.  Every bit of interpretation of the data that we do is a chance for us to mess it up and a forensics investigator to be given potentially bad information.  Just pass through what we get and let them figure out how windows decided to interpret it (I vaguely recall this being a notorious DLL hijack method because `C:\Program Files\somedll` would load a DLL from `C:\Program` if it existed)...

SolitudePy:
Removed Notes column & path extracted from cmdline and added 2 boolean columns for the length checks.
```
(venv) ubuntu@ubuntuPC:~/Dev/volatility3$ vol -f ~/dumps/peb_masq_dump.raw -r json windows.malware.pebmasquerade --pid 12372
Volatility 3 Framework 2.26.2
[
  {
    "EPROCESS_ImageFileName": "peb_masquerade",
    "EPROCESS_SeAudit_ImageFileName": "\\Device\\HarddiskVolume3\\Users\\Robot\\Dev\\peb_masquerade.exe",
    "PEB_CommandLine_Spoofed": true,
    "PEB_ImageFilePath": "C:\\windows\\system32\\notepad.exe",
    "PEB_ImageFilePath_Spoofed": true,
    "PID": 12372,
    "__children": []
  }
]
```

volatility3/framework/plugins/windows/malware/pebmasquerade.py
ikelos: This still needs resolving.  `Staticmethod` splits the method from the plugin, meaning we can't then use introspection effectively to determine things like which version of the plugin this was attached to.  We've run into this before, I realize it feels like it should make no difference, but please use classmethods.

ikelos (commented):
Thanks very much and sorry for the delay in getting this re-reviewed.  It's looking really good now, just that one outstanding comment that unfortunately does still need resolving, but then I think this can go in!  5:)

volatility3/framework/plugins/windows/malware/pebmasquerade.py
```diff
@@ -0,0 +1,234 @@
...
+vollog = logging.getLogger(__name__)
+
+
+# https://www.ired.team/offensive-security/defense-evasion/masquerading-processes-in-userland-through-_peb
+# https://github.com/FuzzySecurity/PowerShell-Suite/blob/master/Masquerade-PEB.ps1
+class PebMasquerade(interfaces.plugins.PluginInterface):
+    """Detects potential process name spoofing by comparing EPROCESS and PEB data."""
+
+    _version = (1, 0, 0)
+    _required_framework_version = (2, 0, 0)
```
ikelos: Does this plugin definitely not rely on any of the changes from version 2.1 to 2.27?  If you're not absolutely certain, this should be the current latest version...

ikelos (commented):
Generally looks good.  I'm a little concerned about false positives because we do our own unicode length maths, and I worry about unusual encodings or something else that throws the count off, but if you're happy this is a useful addition then I have no real problem with it going in.  Just need to check that the required_framework_version is actually accurate, and not just copied by default from other plugins...  It may well be fine, but if you aren't sure, then I'll like people to get in the habit of setting it to the version they're adding it to please...

volatility3/framework/plugins/windows/malware/pebmasquerade.py
SolitudePy: @ikelos not sure hence I set it to 2.27

ikelos (approved):
Thanks, looks ready for primetime then.  5:)

PR #1800, opened 2025-05-07
add functions to etwpatch

SolitudePy:
added functions to etwpatch.py
sources:
https://github.com/SolitudePy/Stealthy-ETW-Patch
https://github.com/mandiant/capa-rules/blob/master/anti-analysis/anti-av/patch-event-tracing-for-windows-function.yml

PR #1826, opened 2025-06-02
process_spoofing plugin

SolitudePy:
Hello, just playing with memory & OS internals. 
apparently some legitimate processes do these techniques to have enriched information in their cmdline or so. here are some such processes:
```
PID     PPID    Exe_Basename    Cmdline_Basename        Comm    Notes

966     1       platform-python3.6      platform-python firewalld       ['Potential cmdline spoofing: exe_file=platform-python3.6;cmdline=platform-python', 'Potential comm spoofing: exe_file=platform-python3.6;comm=firewalld']
991     1       platform-python3.6      platform-python tuned   ['Potential cmdline spoofing: exe_file=platform-python3.6;cmdline=platform-python', 'Potential comm spoofing: exe_file=platform-python3.6;comm=tuned']
1257    1       login   login -- root   login   ['Potential cmdline spoofing: exe_file=login;cmdline=login -- root']
1923    1903    bash    bash    entrypoint.sh   ['Potential comm spoofing: exe_file=bash;comm=entrypoint.sh']
3475    3472    systemd (sd-pam)        (sd-pam)        ['Potential cmdline spoofing: exe_file=systemd;cmdline=(sd-pam)', 'Potential comm spoofing: exe_file=syst
```
```
(venv) ubuntu@ubuntuPC:~/Dev/volatility3$ vol -f ~/dumps/procspoof_dump_lin.raw -r json linux.process_spoofing | jq 'map(select(.Notes != "OK"))'
Volatility 3 Framework 2.26.2
[
  {
    "Cmdline_Basename": "[malwareX]",
    "Comm": "change_argv",
    "Exe_Basename": "change_argv",
    "Notes": "['Potential cmdline spoofing: exe_file=change_argv;cmdline=[malwareX]']",
    "PID": 6717,
    "PPID": 3482,
    "__children": []
  },
  {
    "Cmdline_Basename": "change_comm",
    "Comm": "malwareX",
    "Exe_Basename": "change_comm",
    "Notes": "['Potential comm spoofing: exe_file=change_comm;comm=malwareX']",
    "PID": 6727,
    "PPID": 3482,
    "__children": []
  }
]
```


SolitudePy:
```
(venv) ubuntu@ubuntuPC:~/Dev/volatility3$ vol -f ~/dumps/deleted_proc_dump.raw -r json linux.process_spoofing --pid 4868
Volatility 3 Framework 2.26.2
  warnings.warn(
[
  {
    "Cmdline_Basename": "copied_bash",
    "Comm": "copied_bash",
    "Exe_Basename": "copied_bash (deleted)",
    "Notes": "['Potential Process image deletion: exe_file=copied_bash (deleted)']",
    "PID": 4868,
    "PPID": 4633,
    "__children": []
  }
]
```

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,261 @@
...
+                        ppid,
+                        exe_render,
+                        cmdline_render,
+                        comm_render,
+                        "[" + ", ".join(notes) + "]" if notes else "OK",
+                    ),
+                )
+
+            except Exception as e:
+                vollog.debug(f"Error processing task at {task.vol.offset:#x}: {e}")
```
ikelos: It might be better to make this a `vollog.warning`, since otherwise this won't show up for most users?  It's also not all that useful to user?  It's generally better to catch specific exceptions you're expecting rather than just a catchall "something went wrong" because it makes it very difficult for anyone to diagnose what went wrong (and it prevents volatility's general exception handlers from being able to step in and give the user a nice message, or a developer a full traceback of the exception).

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,261 @@
...
+            ),
+            requirements.ListRequirement(
+                name="pid",
+                description="Filter on specific process IDs",
+                element_type=int,
+                optional=True,
+            ),
+        ]
+
+    def _get_executable_path(
```
ikelos: If these functions would be useful externally by other plugins, consider making them public classmethods (using `@classmethod` and not starting the method name with an `_`)?  It looks like they only really need self to get a context, so if you add the context as the first method parameter (in keeping with all our other plugins/classmethods) then it would make them usable elsewhere.

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,261 @@
...
+            if not mm or not mm.is_readable():
+                # Kernel threads doesn't have mm
+                return None
+
+            exe_file = mm.exe_file
+
+            if not exe_file or not exe_file.is_readable():
+                return None
+
+            exe_inode = exe_file.dereference().f_path.dentry.d_inode
```
ikelos: This *shouldn't* require explicit dereferencing.  Requesting a specific member (like `f_path`) from a pointer will automatically dereference it and return the value from the dereferenced structure.  Including it explicitly perpetuates the idea that it's necessary and could lead to misunderstanding of how pointers work in volatility.

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,261 @@
...
+        Args:
+            task: task_struct object of the process
+
+        Returns:
+            Basename of the first command line argument or None if not available
+        """
+        try:
+            mm = task.mm
+            if not mm or not mm.is_readable():
+                return renderers.NotAvailableValue()
```
ikelos: The renderer values are part of the UI, and generally not all that useful to plugins wanting to process the data.  You return None almost everywhere else apart from here, it would be better to be consistent in your return values unless you're explicitly trying to differentiate something being `NotAvailable` over some other AbsentValue option.  Typically, we'd pass None through, and then just as it comes out of the generator, convert Nones to the responses that makes the most sense for a None for that field.

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,261 @@
...
+
+            proc_layer = self.context.layers[proc_layer_name]
+            start = task.mm.arg_start
+            size_to_read = task.mm.arg_end - task.mm.arg_start
+
+            if not (0 < size_to_read <= 4096):
+                return None
+
+            try:
+                argv = proc_layer.read(start, size_to_read)
```
ikelos: I'd check to see if existing argument handling isn't done somewhere else (either in another plugin or a task structure).  It seems like it might be more useful exposed that way, just to avoid people rewriting it themselves each time and possibly making slight parsing differences each time.

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,261 @@
...
+            except exceptions.InvalidAddressException:
+                return None
+
+            # Parse the arguments - they are null byte terminated
+            args_str = argv.decode(encoding="utf8", errors="replace")
```
ikelos: This feels like it should be array to string which you've used below in `_get_comm`?  You can then split it on `\x00`?

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,261 @@
...
+
+        Args:
+            task: task_struct object of the process
+
+        Returns:
+            Process name from comm field or None if not available
+        """
+        try:
+            return utility.array_to_string(task.comm)
+        except (exceptions.InvalidAddressException, AttributeError):
```
ikelos: We're definitely just swallowing a lot of errors without telling anyone we've done that?  Unless this is commonly expected, it would be better to include a `vollog.log` at a high debug level, just so that a developer *could* tell why something wasn't output when they might not be able to see why just looking at the raw bytes.

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,261 @@
...
+            # Parse the arguments - they are null byte terminated
+            args_str = argv.decode(encoding="utf8", errors="replace")
+            args_list = args_str.split("\x00")
+            if args_list and args_list[0]:
+                basename = PurePosixPath(args_list[0]).name
+                return basename
+            else:
+                return None
+
+        except (exceptions.InvalidAddressException, AttributeError):
```
ikelos: There's a lot of generic error catching without any notice to a developer that some possible output got swallowed or why.

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,261 @@
...
+            notes: List of notes indicating potential spoofing, or None if no issues found
+        """
+        notes = []
+
+        # Skip kernel threads
+        available_sources = sum(
+            1 for name in [exe_basename, cmdline_basename, comm] if name
+        )
+
+        is_deleted = exe_basename.endswith(self.deleted)
```
ikelos: This isn't a good way of identifying this, it should be a separate boolean flag handed around.  Stuffing it on the end of a string and then parsing it back out again feels very brittle.  I'd make each of your three methods return a tuple `(basename, deleted)` and then process those, the other two methods can return None but you really should be loading up the basename, and certainly not at the end, because then you can't tell the difference between `random file` that's been deleted and `random file (deleted)`, within your code the two are indistinguishable.

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,261 @@
...
+        if is_deleted:
+            notes.append(f"'Potential Process image deletion: exe_file={exe_basename}'")
+            exe_basename = exe_basename[: len(self.deleted) * -1]
+
+        if available_sources < 2:
```
ikelos: Surely you'd want to highlight that the file was only present in one source?  Seems like a sure fire sign that something unusual happened?

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,261 @@
...
+            return None
+
+        if exe_basename != cmdline_basename:
+            notes.append(
```
ikelos: This combines data fields, so creating a list of strings is not going to be easy for a volatility plugin to process (and, gets collapsed into a single string, which would need parsing again to be used by another program/a different tool).  This should probably be returning the three compared values, and then a boolean to say which tests succeeded and which failed.  Humans could read that fine (it would be less text to read) and other programs/plugins could parse the data easily.

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,261 @@
...
+            Tuple containing process information and spoofing analysis
+        """
+        for task in tasks:
+            try:
+                pid = task.pid
+                ppid = task.get_parent_pid()
+
+                exe_basename, cmdline_basename, comm = self._extract_process_names(task)
+
+                notes = self._detect_spoofing(exe_basename, cmdline_basename, comm)
```
ikelos: As mentioned above, this is a bad design pattern.  Keep the tests separate and display them each in their own column as a true/false value.

ikelos (commented):
Thanks for your submission and sorry it took so long to find time to review, this plugin looks fun!  It's mostly fine, but it smushes the results together into a string, which is longwinded for humans to read and difficult for programs/plugins to parse.  Hopefully this should be pretty straight forward to refactor into returning boolean values.  It also feels like as number of the methods here may be useful for other plugins, so first check that other plugins don't already implement them and otherwise consider converting them into classmethods so that they can be called externally...

volatility3/framework/plugins/linux/malware/process_spoofing.py
SolitudePy: @ikelos want me to use `psaux.PsAux._get_command_line_args` ? its not directly exposed as an external function but maybe I can refactor it. also in general I think its better that useful methods are exposed through a dedicated utilities class instead of other plugins?

volatility3/framework/plugins/linux/malware/process_spoofing.py
SolitudePy: currently `utility.array_to_string` expects:
```
    # TODO: Consider checking the Array's target is a native char
    if not isinstance(array, objects.Array):
        raise TypeError("Array_to_string takes an Array of char")
```
which in argv case isn't since its raw bytes, read from `mm.arg_start\arg_end` through `proc_layer.read()`

volatility3/framework/plugins/linux/malware/process_spoofing.py
SolitudePy: I agree, also I'd rather it depend on my other PR for the linuxutilities function instead of reusing the code here... I'm not sure about a boolean flag and also many gnu-utils already do that (appending (deleted) to a path).

volatility3/framework/plugins/linux/malware/process_spoofing.py
SolitudePy: kernel threads dont have `mm_struct`, that means no cmdline and exe_file

SolitudePy: I suppose kernel threads can be "flagged" in the output as such?

volatility3/framework/plugins/linux/malware/process_spoofing.py
SolitudePy: I refactored it now, tell me what you think and if we should also implement it in my lsof PR?

SolitudePy:
```
(venv) ubuntu@ubuntuPC:~/Dev/volatility3$ vol -f ~/dumps/deleted_proc_dump.raw linux.malware.process_spoofing --pid 4868 
Volatility 3 Framework 2.26.2
PID     PPID    Exe_Basename    Cmdline_Basename        Comm    Cmdline_Spoofed Comm_Spoofed    Exe_Deleted

4868    4633    copied_bash (deleted)   copied_bash     copied_bash     False   False   True
```
```
(venv) ubuntu@ubuntuPC:~/Dev/volatility3$ vol -f ~/dumps/procspoof_dump_lin.raw -r json linux.malware.process_spoofing --pid 6727
Volatility 3 Framework 2.26.2
[
  {
    "Cmdline_Basename": "change_comm",
    "Cmdline_Spoofed": false,
    "Comm": "malwareX",
    "Comm_Spoofed": true,
    "Exe_Basename": "change_comm",
    "Exe_Deleted": false,
    "PID": 6727,
    "PPID": 3482,
    "__children": []
  }
]
```

volatility3/framework/plugins/linux/malware/process_spoofing.py
ikelos: Sorry for the long delay.  Yes this should probably live in LinuxUtilities under a submodule, I guess related to the task struct, but that would need a little thinking about to make sure it's suitable (and probably some input from @atcuno or @Abyss-W4tcher to make sure it lines up with other linux general modules).

volatility3/framework/plugins/linux/malware/process_spoofing.py
ikelos: Presumably, because we're forming a string, we could cast it to an array of char first, and then convert that?  Really feels like we should avoid multiple string parsing functions, given they're where little issues tend to sneak in.

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,287 @@
...
+from volatility3.framework.configuration import requirements
+from volatility3.framework.interfaces import plugins
+from volatility3.framework.objects import utility
+from volatility3.framework.symbols import linux
+from volatility3.plugins.linux import pslist
+
+vollog = logging.getLogger(__name__)
+
+
+# https://github.com/SolitudePy/linux-mal
```
ikelos: This probably doesn't belong here if it's going in the main core?  If it's just a reference to further documentation or something then at the very least it belongs in the plugin doctstring rather than in a comment.

volatility3/framework/plugins/linux/malware/process_spoofing.py
```diff
@@ -0,0 +1,287 @@
...
+
+        proc_layer_name = task.add_process_layer()
+        if proc_layer_name is None:
+            return None
+
+        proc_layer = context.layers[proc_layer_name]
+        start = task.mm.arg_start
+        size_to_read = task.mm.arg_end - task.mm.arg_start
+
+        if not (0 < size_to_read <= 4096):
```
ikelos: Should the 4096 be page size, or is it strictly limited to 4096 as a maximum somewhere?  I prefer linux specific constants to be stored in the constants.linux file, but if this is literally a hard coded fixed size then we can leave it here (although good to linux to a kernel source file (and a kernel git commit ID for that file, so that if it moves about we can track it).

volatility3/framework/plugins/linux/malware/process_spoofing.py
ikelos: Looks like I didn't keep up with the changes, sorry!  All seems pretty good now, thanks for bearing with me!

ikelos (approved):
All looks really good, just minor quibbles, but honestly it could go in like this if you're tired of tweaking it!  Thanks very much for putting up with my schedule, it's really appreciated as are your plugins!  5:)

volatility3/framework/plugins/linux/malware/process_spoofing.py
SolitudePy: @ikelos I dont mind I did it based on an other PR I had #1800 where atcuno commented to put references there

volatility3/framework/plugins/linux/malware/process_spoofing.py
SolitudePy: a single argument is limited 4096 * 32 according to https://raw.githubusercontent.com/torvalds/linux/v6.7/include/uapi/linux/binfmts.h, I simply copied the psaux implementation so I guess its not accurate both in here and psaux because it will skip processes with multi-page arg, honestly get_cmdline_basename wants only the first argument so maybe a better approach would be to read until ```'\0'``` with a maximum of 131072 bytes?

volatility3/framework/plugins/linux/malware/process_spoofing.py
ikelos: Fair enough, I still think documentation should live in the docstring (ideally with a bit of text rather than just a dangling link), but it's not a showstopper.

volatility3/framework/plugins/linux/malware/process_spoofing.py
ikelos: Rather than returning `None`, wouldn't it make more sense to use:
```
size_to_read = min(task.mm.arg_end - task.mm.arg_start, 4096*32)
```
instead?

If you're just after the first argument, then you can use the `string` class (which psaux couldn't because it needed everything):

```
try:
  cmdline = context.object("string", proc_layer, offset=start, max_length=size_to_read, errors="replace")
except exception.InvalidAddressException as e:
  vollog.debug(
    f"Unable to read cmdline for task at {task.vol.offset:#x}: {e}"
  )
  return None

basename = PurePosixPath(cmdline).name
return basename
```

That *should* handle it all in effectively one line (and do it using the string class, which means fixes made later on will apply to this code for free)...

SolitudePy: I like that idea, I suppose checking if size_to_read > 0 is still an important guard though?

ikelos: I *think* casting to a string should handle that?  Data would come back as b"", which should then return an empty string?  If you want it to explicitly return None, then yeah you'd need an extra check...

SolitudePy: @ikelos actually im getting ```volatility3.framework.exceptions.SymbolError: Malformed name: string``` for this maybe because I'm using btf2json generated symbol table. Maybe I can use ```utility.address_to_string``` instead?

ikelos: Yeah, that's probably best.  `string` should be a builtin type, but it might needs a `symbol_table!string` name.  All symbol tables should have it defined, but `utility.address_to_string` might be easier and it has machinery to be safer I believe...

ikelos (approved):
Looks good, thanks!  5:)  We can merge as soon as the tests complete...

PR #1827, opened 2025-06-04
Lsof improvements (show deleted as in lsof output) + files_only argument

SolitudePy:
Improvements for Lsof to show deleted files as in lsof output on live system. 
its probably shouldnt be in that function directly as its no longer mimic `prepend_path` kernel function 
this PR breaks commit 68b51e873fe99ffcbf1679f25a9767774b9da3e2 but it can easily be fixed if desired. 

```
(venv) ubuntu@ubuntuPC:~/Dev/volatility3$ vol -f ~/dumps/deleted_proc_fd_dump.raw -r json linux.lsof --pid 8321
Volatility 3 Framework 2.26.2
  warnings.warn(
[
...
...
  {
    "Accessed": "2025-06-04T18:18:04.438000+00:00",
    "Changed": "2025-06-04T18:15:11.438000+00:00",
    "Device": "0:23",
    "FD": 2,
    "Inode": 4,
    "Mode": "crw--w----",
    "Modified": "2025-06-04T18:18:04.438000+00:00",
    "PID": 8321,
    "Path": "/dev/pts/1",
    "Process": "copied_bash",
    "Size": 0,
    "TID": 8321,
    "Type": "CHR",
    "__children": []
  },
  {
    "Accessed": "2025-06-04T17:49:01.296000+00:00",
    "Changed": "2025-06-04T18:17:41.693000+00:00",
    "Device": "253:0",
    "FD": 255,
    "Inode": 201866930,
    "Mode": "-rw-r--r--",
    "Modified": "2025-06-04T17:48:56.399000+00:00",
    "PID": 8321,
    "Path": "/tmp/evil.sh (deleted)",
    "Process": "copied_bash",
    "Size": 19,
    "TID": 8321,
    "Type": "REG",
    "__children": []
  }
]
```

```
(venv) ubuntu@ubuntuPC:~/Dev/volatility3$ vol -f ~/dumps/deleted_proc_fd_dump.raw linux.lsof --files-only | grep deleted
  warnings.warn(
799gress799100.0systemd-udevd   8tacking/var/lib/sss/mc/group (deleted) 253:0   201339099       REG     -rw-rw-r--      2025-06-04 17:46:02.113000 UTC  2025-06-04 17:46:02.113000 UTC   2025-06-04 17:46:01.736000 UTC  6940392
799     799     systemd-udevd   9       /var/lib/sss/mc/passwd (deleted)        253:0   201866915       REG     -rw-rw-r--      2025-06-04 17:46:02.109000 UTC  2025-06-04 17:46:02.109000 UTC   2025-06-04 17:46:01.724000 UTC  9253600
906     906     auditd  4       /var/lib/sss/mc/group (deleted) 253:0   201339099       REG     -rw-rw-r--      2025-06-04 17:46:02.113000 UTC  2025-06-04 17:46:02.113000 UTC  2025-06-04 17:46:01.736000 UTC   6940392
906     907     auditd  4       /var/lib/sss/mc/group (deleted) 253:0   201339099       REG     -rw-rw-r--      2025-06-04 17:46:02.113000 UTC  2025-06-04 17:46:02.113000 UTC  2025-06-04 17:46:01.736000 UTC   6940392
...
960     1218    gmain   9       / (deleted)     0:1     26770   REG     -rwxrwxrwx      2025-06-04 17:46:02.397000 UTC  2025-06-04 17:46:02.397000 UTC  2025-06-04 17:46:02.397000 UTC   4096
961     961     sssd_nss        6       /var/lib/sss/mc/passwd (deleted)        253:0   201866915       REG     -rw-rw-r--      2025-06-04 17:46:02.109000 UTC  2025-06-04 17:46:02.109000 UTC   2025-06-04 17:46:01.724000 UTC  9253600
```

### DIFF
--- a/volatility3/framework/plugins/linux/lsof.py
+++ b/volatility3/framework/plugins/linux/lsof.py
@@ -137,6 +137,12 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
                 element_type=int,
                 optional=True,
             ),
+            requirements.BooleanRequirement(
+                name="files_only",
+                description="Include only file descriptors of type file",
+                optional=True,
+                default=False,
+            ),
         ]
 
     @classmethod
@@ -145,6 +151,7 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
         context: interfaces.context.ContextInterface,
         vmlinux_module_name: str,
         filter_func: Callable[[int], bool] = lambda _: False,
+        include_files_only: bool = False,
     ) -> Iterable[FDInternal]:
         """Enumerates open file descriptors in tasks
 
@@ -167,7 +174,7 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
                 linuxutils_symbol_table = task.vol.type_name.split(constants.BANG)[0]
 
             fd_generator = linux.LinuxUtilities.files_descriptors_for_process(
-                context, linuxutils_symbol_table, task
+                context, linuxutils_symbol_table, task, files_only=include_files_only
             )
 
             for fd_fields in fd_generator:
@@ -175,8 +182,12 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
 
     def _generator(self, pids, vmlinux_module_name):
         filter_func = pslist.PsList.create_pid_filter(pids)
+        include_files_only = self.config.get("files_only")
         for fd_internal in self.list_fds(
-            self.context, vmlinux_module_name, filter_func=filter_func
+            self.context,
+            vmlinux_module_name,
+            filter_func=filter_func,
+            include_files_only=include_files_only,
         ):
             fd_user = fd_internal.to_user()
             yield (0, dataclasses.astuple(fd_user))
--- a/volatility3/framework/plugins/linux/lsof.py
+++ b/volatility3/framework/plugins/linux/lsof.py
@@ -180,9 +180,9 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
             for fd_fields in fd_generator:
                 yield FDInternal(task=task, fd_fields=fd_fields)
 
-    def _generator(self, pids, vmlinux_module_name):
+    def _generator(self, pids, vmlinux_module_name, include_files_only):
         filter_func = pslist.PsList.create_pid_filter(pids)
-        include_files_only = self.config.get("files_only")
+
         for fd_internal in self.list_fds(
             self.context,
             vmlinux_module_name,
@@ -195,6 +195,7 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
     def run(self):
         pids = self.config.get("pid", None)
         vmlinux_module_name = self.config["kernel"]
+        include_files_only = self.config.get("files_only")
 
         tree_grid_args = [
             ("PID", int),
@@ -212,7 +213,10 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
             ("Size", int),
         ]
         return renderers.TreeGrid(
-            tree_grid_args, self._generator(pids, vmlinux_module_name)
+            tree_grid_args,
+            self._generator(
+                pids, vmlinux_module_name, include_files_only=include_files_only
+            ),
         )
 
     def generate_timeline(self):
--- a/volatility3/framework/plugins/linux/lsof.py
+++ b/volatility3/framework/plugins/linux/lsof.py
@@ -110,7 +110,7 @@ class Lsof(plugins.PluginInterface, timeliner.TimeLinerInterface):
     """Lists open files for each processes."""
 
     _required_framework_version = (2, 0, 0)
-    _version = (2, 0, 2)
+    _version = (2, 1, 0)
 
     @classmethod
     def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
--- a/volatility3/framework/symbols/linux/__init__.py
+++ b/volatility3/framework/symbols/linux/__init__.py
@@ -101,7 +101,7 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
 
     _version = (2, 4, 0)
     _required_framework_version = (2, 0, 0)
-    deleted = "<deleted>"
+    deleted = "(deleted)"
     smear = "<potentially smeared>"
 
     framework.require_interface_version(*_required_framework_version)
@@ -210,7 +210,7 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
             return f"{LinuxUtilities.smear} {path}"
 
         if inode and inode.is_readable() and inode.is_valid() and inode.i_nlink == 0:
-            path = f"{LinuxUtilities.deleted} {path}"
+            path = f" {path} {LinuxUtilities.deleted}"
         return path
 
     @classmethod
--- a/volatility3/framework/symbols/linux/__init__.py
+++ b/volatility3/framework/symbols/linux/__init__.py
@@ -99,7 +99,7 @@ class LinuxKernelIntermedSymbols(intermed.IntermediateSymbolTable):
 class LinuxUtilities(interfaces.configuration.VersionableInterface):
     """Class with multiple useful linux functions."""
 
-    _version = (2, 3, 1)
+    _version = (2, 4, 0)
     _required_framework_version = (2, 0, 0)
     deleted = "<deleted>"
 
@@ -307,7 +307,7 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
         return f"{pre_name}:[{inode.i_ino:d}]"
 
     @classmethod
-    def path_for_file(cls, context, task, filp, files_only) -> str:
+    def path_for_file(cls, context, task, filp, files_only=False) -> str:
         """Returns a file (or sock pipe) pathname relative to the task's root directory.
 
         A 'file' structure doesn't have enough information to properly restore its
--- a/volatility3/framework/symbols/linux/__init__.py
+++ b/volatility3/framework/symbols/linux/__init__.py
@@ -101,6 +101,7 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
 
     _version = (2, 3, 1)
     _required_framework_version = (2, 0, 0)
+    deleted = " (deleted)"
 
     framework.require_interface_version(*_required_framework_version)
 
@@ -168,6 +169,7 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
             # vfsmnt can be the vfsmount object itself (>=3.3) or a vfsmount * (<3.3)
             return ""
 
+        inode = dentry.d_inode
         path_reversed = []
         smeared = False
         while (
@@ -191,6 +193,7 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
 
             parent = dentry.d_parent
             dname = dentry.d_name.name_as_str()
+
             # empty dentry names are most likely
             # the result of smearing
             if not dname:
@@ -204,6 +207,9 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
             # path would be /foo/bar/baz, but bar is missing due to smear the results
             # returned here will show /foo//baz. Note the // for the missing dname.
             return f"<potentially smeared> {path}"
+        print(path, inode.i_nlink)
+        if inode and inode.is_readable() and inode.is_valid() and inode.i_nlink == 0:
+            path += LinuxUtilities.deleted
         return path
 
     @classmethod
@@ -260,7 +266,7 @@ class LinuxUtilities(interfaces.configuration.VersionableInterface):
                     pre_name = name.dereference().cast(
                         "string", max_length=255, errors="replace"
                     )
-                    return "/" + pre_name + " (deleted)"
+                    return "/" + pre_name + LinuxUtilities.deleted
                 else:
                     pre_name = ""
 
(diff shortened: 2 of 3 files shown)


volatility3/framework/symbols/linux/__init__.py
```diff
@@ -204,6 +207,9 @@ def do_get_path(cls, rdentry, rmnt, dentry, vfsmnt) -> Union[None, str]:
             # path would be /foo/bar/baz, but bar is missing due to smear the results
             # returned here will show /foo//baz. Note the // for the missing dname.
             return f"<potentially smeared> {path}"
+
+        if inode and inode.is_readable() and inode.is_valid() and inode.i_nlink == 0:
+            path += LinuxUtilities.deleted
```
ikelos: The constant is nice, but since we return values like `<potemtially smeared> {path}` which are not done with a constant, it may be overkill here.  It would also be difficult to differentiate between a file that literally ended in ` (deleted)` and one that had been deleted.  Since all paths start with `/`, we've allowed `<string>` to identify attributes about a filename.  As such, please prepend `<deleted> ` to your path, rather than appending it to the end of the filename.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -301,7 +307,7 @@ def _get_new_sock_pipe_path(cls, context, task, filp) -> str:
         return f"{pre_name}:[{inode.i_ino:d}]"
 
     @classmethod
-    def path_for_file(cls, context, task, filp) -> str:
+    def path_for_file(cls, context, task, filp, files_only) -> str:
```
ikelos: This is an API change and because a default value hasn't been provided, it would require a bump of the MAJOR version number (so `(3,0,0)`).  If a default value were provided you could get away with a MINOR version bump (so `(2,4,0)`), since old code would work with it.

The other pain point is that this a change to the linux extensions, so this would be a bump of the whole LinuxUtilities class.  If you make it a MAJOR bump, it'll require *all* the other plugins that use it to be updated (and in turn each of their versions to get a PATCH bump, technically).  Much easier to provide a default value and just do a MINOR bump.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -353,6 +359,7 @@ def files_descriptors_for_process(
         context: interfaces.context.ContextInterface,
         symbol_table: str,
         task: interfaces.objects.ObjectInterface,
+        files_only: bool = False,
```
ikelos, line 363: This on its own would require a MINOR version bump, so the version of the extensions class needs to change one way or another (you only have to bump it once even if you're doing two changes though).

ikelos (changes requested):
Thanks for this, a couple of things fell out of it.  Firstly, messing with the core linux extensions can have knock on effects, which is why we version everything.  There *is* a way of updating it so that not everything breaks, but that one's a show stopper.

The other is just about appending to filenames to represent information about the file.  If we're going to do that, it has to be unambiguously separable from the filename (ie, you could always tell what was the tag and what was the filename).

volatility3/framework/symbols/linux/__init__.py
SolitudePy: thats true, actually lsof command does that but when you think about it this way its pretty hard to tell the difference between filenames with actual `' (deleted)'`, although I have seen it was this way in `_get_new_sock_pipe_path()`, should I change it there as well?

volatility3/framework/symbols/linux/__init__.py
SolitudePy: done

volatility3/framework/symbols/linux/__init__.py
SolitudePy: I have left the constant because its a tag that can be reused... would you want a constant for the smear tag as well?

ikelos: I'm not sure it makes sense to have them done differently.  If you really think it's valuable to keep the deleted value as a constant, then best to pull out the smear tag too.

As to changing `_get_new_sock_pipe_path()` please either file a bug or fix it in a separate PR, so it doesn't weigh down this one.

volatility3/framework/symbols/linux/__init__.py
```diff
@@ -204,6 +207,9 @@ def do_get_path(cls, rdentry, rmnt, dentry, vfsmnt) -> Union[None, str]:
             # path would be /foo/bar/baz, but bar is missing due to smear the results
             # returned here will show /foo//baz. Note the // for the missing dname.
             return f"<potentially smeared> {path}"
+
+        if inode and inode.is_readable() and inode.is_valid() and inode.i_nlink == 0:
+            path = f"{LinuxUtilities.deleted} {path}"
```
ikelos: We tend to put constants into the `constants.linux` module, as all caps?  Again, I'm not sure it's worthwhile for a tag like this?

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1253,7 +1253,9 @@ def get_page_offset(self) -> int:
 
     def _do_get_name(self, context, task) -> str:
         if self.vm_file != 0:
-            fname = linux.LinuxUtilities.path_for_file(context, task, self.vm_file)
+            fname = linux.LinuxUtilities.path_for_file(
+                context, task, self.vm_file, files_only=False
```
ikelos: Not sure this needs the `=false` here, but it's not really causing any problems...

ikelos (commented):
Much improved, last little bits and then it should be good to go.

volatility3/framework/symbols/linux/__init__.py
SolitudePy: what if the tag value will change in the future? I can see other modules rely on this tag to check specifics...

SolitudePy: not sure if it fits there, I have only seen constants from linux kernel header files

volatility3/framework/plugins/linux/lsof.py
```diff
@@ -145,6 +151,7 @@ def list_fds(
         context: interfaces.context.ContextInterface,
         vmlinux_module_name: str,
         filter_func: Callable[[int], bool] = lambda _: False,
+        include_files_only: bool = False,
```
ikelos, line 154: This is a change of a class method, but in a backwards compatible way, so it still needs the plugin MINOR version bumping (ie, `(2, 1, 0)`)

volatility3/framework/symbols/linux/__init__.py
ikelos: I'm really not keen on tacking flags at the start of the path.  It feels more and more like there should be a structure or a method that can determine this, that's separate to the filename.  I don't really like it for smear, but at least the filename can be smeared.  It's not the filename that's deleted, it's the file, this should be a property of the file object, or something you can pass the file to which tells you whether it's been deleted or not.  That then also solves the need for string based flags to say meta things about the file that shouldn't actually be part of the filename.

volatility3/framework/plugins/linux/lsof.py
```diff
@@ -167,23 +174,28 @@ def list_fds(
                 linuxutils_symbol_table = task.vol.type_name.split(constants.BANG)[0]
 
             fd_generator = linux.LinuxUtilities.files_descriptors_for_process(
-                context, linuxutils_symbol_table, task
+                context, linuxutils_symbol_table, task, files_only=include_files_only
```
ikelos, line 177: Sure this can be determined from the file desriptor, and therefore doesn't require tampering with the `files_descriptors_for_process` method?  I really can't believe that we get this structure back, but we can only separate out files from not files within that method?  I'd also prefer that we just add it to the file descriptor to say "type is file" rather than this overly specific additional parameter added to this method...

ikelos (commented):
Still not happy with tagging filesnames with a string to identify them.  It feels like a quick solution that could be setting us up for a whole heap of trouble in the future.  Hopefully @atcuno can give this a look and suggest a better means of handling this?

volatility3/framework/plugins/linux/lsof.py
SolitudePy: you mean the method `files_descriptors_for_process` should return file type as well that will then be used in `linux.lsof` to filter out non-files? right now `linux.lsof` uses `files_descriptors_for_process` directly, which uses `LinuxUtilities.path_for_file` - and that's where the filtering happens via `dname_is_valid`, so both methods need to have this argument. or I can also reuse the `dname_is_valid` logic in `linux.lsof` which seems less elegant?

volatility3/framework/symbols/linux/__init__.py
SolitudePy: how about the logic in 8f6029973c1c7a9a1a1f12dc6920354bd1202e18?

volatility3/framework/plugins/linux/lsof.py
ikelos: I meant it feels like it should be one of the `fd_fields` returned in the generator and then handled by `FDInternal`.  It might be a lot of work to thread it in that way, and I haven't even considered the version bumps that might be required to achieve it, plus I've no idea whether @atcuno would be happy doing that, but there's no indication that there aren't more fields (requiring more boolean flags to be added to the method) still to come.  As such I'd like to find an extensible way of adding those flags, and given the entire method is about returning a set of flags, it feels as though this should be added to that existing set of flags...

volatility3/framework/symbols/linux/__init__.py
ikelos: That's still fundamentally trying to cram to pieces of information into a single field.  Whether the file has been deleted or not is distinct from the filename it has.  We should be using a separate piece of information to track that, not overloading the filename field with two pieces of information that we'll later have difficult separating.

ikelos (commented):
Sorry, my view on these hasn't really changed.  Rather than designing the data flow properly, we're trying to carry the information by cramming it into other fields which is then extremely difficult to separate at a later date.  It's the wrong way of doing it, and though I don't know how much work the right way would be, this is not it.

volatility3/framework/symbols/linux/__init__.py
SolitudePy: Why? It is returned as a tuple now of (path, is_deleted) that can then be added to the path exactly like GNU utilities does (e.g lsof), also deleted flag is not really part of FD or INODE...

atcuno: The "(deleted)" part does need to be part of the file name, and to match Linux systems it needs to be after the name. You can grep across the `maps` file of processes to see how its displayed (shown last in this comment). The addition of "(deleted)" comes directly from the kernel in `d_path`, which is the function we replicate and is the one that generates file paths. It is 100% a bug if we don't show "(deleted)" on these file names, this is the kernel line of code in question:

https://github.com/torvalds/linux/blob/d7b8f8e20813f0179d8ef519541a3527e7661d3a/fs/d_path.c#L287

```
$ sudo grep delete /proc/*/maps | head -10
/proc/1198626/maps:7fa0ef35c000-7fa2fc600000 rw-s 00000000 00:01 8154                       /dev/zero (deleted)
/proc/1198626/maps:7fa30085c000-7fa30085d000 rw-s 00000000 00:01 7                          /SYSV043c0003 (deleted)
/proc/1242475/maps:7fa0ef35c000-7fa2fc600000 rw-s 00000000 00:01 8154                       /dev/zero (deleted)
/proc/1242475/maps:7fa30085c000-7fa30085d000 rw-s 00000000 00:01 7                          /SYSV043c0003 (deleted)
/proc/1242476/maps:7fa0ef35c000-7fa2fc600000 rw-s 00000000 00:01 8154                       /dev/zero (deleted)
/proc/1242476/maps:7fa30085c000-7fa30085d000 rw-s 00000000 00:01 7                          /SYSV043c0003 (deleted)
/proc/1246761/maps:7fa0ef35c000-7fa2fc600000 rw-s 00000000 00:01 8154                       /dev/zero (deleted)
/proc/1246761/maps:7fa30085c000-7fa30085d000 rw-s 00000000 00:01 7                          /SYSV043c0003 (deleted)
/proc/1246762/maps:7fa0ef35c000-7fa2fc600000 rw-s 00000000 00:01 8154                       /dev/zero (deleted)
/proc/1246762/maps:7fa30085c000-7fa30085d000 rw-s 00000000 00:01 7                          /SYSV043c0003 (deleted)

```

atcuno: As a follow up to my own comment, we already show `(deleted)` correctly, which was fixed in the parity push, in our `d_path` replication, so whatever new PRs replace/enhance the existing code need to keep it. Link here:

https://github.com/volatilityfoundation/volatility3/blob/1eb87178b8ec9f6f196590f2f024db2c702c8d31/volatility3/framework/symbols/linux/__init__.py#L263

ikelos: Ok, so my bad, apparently this is how the kernel does it (and it does it at the end).  Sorry for messing you about @SolitudePy .  I do still need @atcuno to do a review of it (because clearly I don't know linux that deeply), but he's very busy in the run up to blackhat/defcon so this may have to just wait a while I'm afraid.  In the interim, feel free to move the string back to the end of the filename, sorry for the confusion I caused!  5:S

SolitudePy: well I absolutely agree that it shouldnt be in the end for the problems you have mentioned, but thats convention. were there other issues I should fix except that?

atcuno: @SolitudePy @ikelos what is the latest on this now that the "(deleted)" string is understood?

ikelos: It's probably waiting on me.  I'm still not happy about it, but apparently this is how the kernel does it.  I absolutely wash my hands of resolving any difficulties that arise later from us doing things this way though.  It seems ill thought out by the kernel to conflate the filename and its deleted state.  I'll merge it though, thanks for your work on it (and trying your best to appease our differing opinions @SolitudePy )

PR #1828, opened 2025-06-06
linux.sockstat unix arg not used & inet arg

SolitudePy:
```
(venv) ubuntu@ubuntuPC:~/Dev/volatility3$ vol -f ~/dumps/deleted_proc_dump.raw linux.sockstat --help
Volatility 3 Framework 2.26.2
  warnings.warn(
usage: vol linux.sockstat.Sockstat [-h] [--unix] [--inet] [--pids [PIDS ...]] [--netns NETNS]

Lists all network connections for all processes.

options:
  -h, --help         show this help message and exit
  --unix             Show UNIX domain Sockets only
  --inet             Show INET/INET6 domain Sockets only
  --pids [PIDS ...]  Filter results by process IDs. It takes the root PID namespace identifiers.
  --netns NETNS      Filter results by network namespace. Otherwise, all of them are shown.
```

### DIFF
--- a/volatility3/framework/plugins/linux/sockstat.py
+++ b/volatility3/framework/plugins/linux/sockstat.py
@@ -576,7 +576,7 @@ class Sockstat(plugins.PluginInterface):
 
             if unix_only and family != "AF_UNIX":
                 continue
-            
+
             if inet_only and family not in ("AF_INET", "AF_INET6"):
                 continue
 
@@ -610,7 +610,14 @@ class Sockstat(plugins.PluginInterface):
 
         return tuple(sock_stat), protocol
 
-    def _generator(self, pids: List[int], netns_id_arg: int, kernel_module_name: str, unix_only: bool, inet_only: bool):
+    def _generator(
+        self,
+        pids: List[int],
+        netns_id_arg: int,
+        kernel_module_name: str,
+        unix_only: bool,
+        inet_only: bool,
+    ):
         """Enumerate tasks sockets. Each row represents a kernel socket.
 
         Args:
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/linux/sockstat.py
```diff
@@ -666,6 +695,8 @@ def _generator(self, pids: List[int], netns_id_arg: int, kernel_module_name: str
     def run(self):
         pids = self.config.get("pids")
         netns_id = self.config["netns"]
+        unix_only = self.config["unix"]
```
ikelos, line 698: There is no `unix` value added to the config, only `inet` got added up on line 484.

ikelos (commented):
Thanks very much for your contribution (and sorry it took so long to get reviewed).  I'm not sure of the value of the addition, since anything processing the results of this plugin could include/exclude options based on the columns returned, and CLI users could use grep?  We've even included a `--filters` option, so you could add `--filters Family,AF_INET` or `--filters Family,AF_UNIX` to limit the results to just those rows with a family value containing AF_INET or AF_UNIX.

ikelos:
Thanks for taking the time to send in this contribution.  I'm afraid I don't think this offers much additional value over functionality that's already present in the tool, so I'm going to close this off.  If you feels there's some specific value in this that's been missed, please feel free to reopen it and we can discuss its merits, and do submit any more pull requests you may have, we're always eager to improve things if we can!  5:)

volatility3/framework/plugins/linux/sockstat.py
SolitudePy:
```
            if unix_only and family != "AF_UNIX":
                continue
```
the unix argument was there before my PR, but it wasnt used... since there was a unix argument I thought adding the inet as well since its more common when vieweing sockets. I agree `--filters` can be used and I forgot about that, although adding plugin-specific arguments would be quicker I guess, want me to create a PR to remove the unix argument?

ikelos: Oh, if it was there but not doing anything then yeah, that would be useful, thanks!  5:)

PR #1829, opened 2025-06-07
Regexscan fixes 

SolitudePy:
1. use maxsize argument 
2. switched match to search to search the whole chunk
3. moved some code from _generator() to run() to reduce overhead

### DIFF
--- a/volatility3/framework/plugins/regexscan.py
+++ b/volatility3/framework/plugins/regexscan.py
@@ -48,6 +48,7 @@ class RegExScan(plugins.PluginInterface):
 
     def _generator(self, regex_pattern):
         regex_pattern = bytes(regex_pattern, "UTF-8")
+        compiled_pattern = re.compile(regex_pattern)
         vollog.debug(f"RegEx Pattern: {regex_pattern}")
         maxsize = self.config.get("maxsize", self.MAXSIZE_DEFAULT)
         layer = self.context.layers[self.config["primary"]]
@@ -57,7 +58,7 @@ class RegExScan(plugins.PluginInterface):
             result_data = layer.read(offset, maxsize, pad=True)
 
             # reapply the regex in order to extract just the match
-            regex_result = re.match(regex_pattern, result_data)
+            regex_result = compiled_pattern.search(result_data)
 
             if regex_result:
                 # the match is within the results_data (e.g. it fits within maxsize)
--- a/volatility3/framework/plugins/regexscan.py
+++ b/volatility3/framework/plugins/regexscan.py
@@ -46,14 +46,12 @@ class RegExScan(plugins.PluginInterface):
             ),
         ]
 
-    def _generator(self, regex_pattern):
-        regex_pattern = bytes(regex_pattern, "UTF-8")
-        compiled_pattern = re.compile(regex_pattern)
-        vollog.debug(f"RegEx Pattern: {regex_pattern}")
-        maxsize = self.config.get("maxsize", self.MAXSIZE_DEFAULT)
+    def _generator(self, compiled_pattern, raw_pattern, maxsize):
+        vollog.debug(f"RegEx Pattern: {raw_pattern}")
         layer = self.context.layers[self.config["primary"]]
+
         for offset in layer.scan(
-            context=self.context, scanner=scanners.RegExScanner(regex_pattern)
+            context=self.context, scanner=scanners.RegExScanner(raw_pattern)
         ):
             result_data = layer.read(offset, maxsize, pad=True)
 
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/regexscan.py
```diff
@@ -46,38 +46,63 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
             ),
         ]
 
-    def _generator(self, regex_pattern):
-        regex_pattern = bytes(regex_pattern, "UTF-8")
-        vollog.debug(f"RegEx Pattern: {regex_pattern}")
-
+    def _generator(self, compiled_pattern, raw_pattern, maxsize):
```
ikelos: The compiled pattern can be generated from the raw_pattern.  Asking for them both as parameters could lead to the two diverging (technically) and that breaking down the line.  Better to do the conversion to a compile pattern here to know that the value matches and won't change.

volatility3/framework/plugins/regexscan.py
```diff
@@ -46,38 +46,63 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
         for offset in layer.scan(
-            context=self.context, scanner=scanners.RegExScanner(regex_pattern)
+            context=self.context, scanner=scanners.RegExScanner(raw_pattern)
         ):
-            result_data = layer.read(offset, self.MAXSIZE_DEFAULT, pad=True)
+            result_data = layer.read(offset, maxsize, pad=True)
 
             # reapply the regex in order to extract just the match
-            regex_result = re.match(regex_pattern, result_data)
+            regex_result = compiled_pattern.search(result_data)
```
ikelos, line 69: The need for this is somewhat annoying, but probably can't be improved at this point without breaking the regexscanner...

volatility3/framework/plugins/regexscan.py
```diff
@@ -46,38 +46,63 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
 
             yield 0, (format_hints.Hex(offset), text_result, bytes_result)
 
     def run(self):
+        pattern = self.config.get("pattern")
+
+        # Handle pattern encoding robustly
+        if isinstance(pattern, str):
+            try:
+                raw_pattern = pattern.encode("utf-8")
```
ikelos: I'm really not keen on trying to guess the pattern encoding here.  This whole section messing with the encodings feels brittle and likely to lead to something going wrong.  If it comes in as a string, It should already be suitably encoded.  This isn't robust is asserting that it'll try utf-8 and then it'll do latin1 replacing errors but it's not possible to know beforehand under which conditions it'll do that without doing it first anyway.

volatility3/framework/plugins/regexscan.py
```diff
@@ -46,38 +46,63 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+            except UnicodeEncodeError:
+                raw_pattern = pattern.encode("latin1", errors="replace")
+        else:
+            raw_pattern = pattern
+
+        try:
+            compiled_pattern = re.compile(raw_pattern)
+        except re.error as e:
+            vollog.error(f"Invalid regex pattern: {e}")
+            return renderers.TreeGrid(
```
ikelos: Don't throw an error but then return a tree grid (even an empty one).  If it's thrown an error, stop the program or raise an exception or *something*, don't carry on like everything's fine, but silently returning no data.

ikelos (commented):
Thanks for this, shift the regex compilaytion back into the generator to avoid potential skew.  I'm also unkeen on the utf-8/latin1 munging for the sake of it.  Otherwise good catch on maxsize not being used

volatility3/framework/plugins/regexscan.py
```diff
@@ -46,38 +46,50 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
             ),
         ]
 
-    def _generator(self, regex_pattern):
-        regex_pattern = bytes(regex_pattern, "UTF-8")
-        vollog.debug(f"RegEx Pattern: {regex_pattern}")
+    def _generator(self, layer, pattern, maxsize):
```
ikelos:
```suggestion
    def _generator(self, context, layer_name, pattern, maxsize):
        layer = self.context.layers[layer_name]
```
Sorry to be a pain, but we prefer people to pass the context and then basic variables (like strings) to look up things like layers.  This makes the whole thing easier to parcel up for if we ever want to move to a multithreaded system.

volatility3/framework/plugins/regexscan.py
```diff
@@ -46,38 +46,50 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+        try:
+            compiled_pattern = re.compile(pattern_bytes)
+        except re.error as e:
+            vollog.error(f"Invalid regex pattern: {e}")
+            raise ValueError(f"Invalid regex pattern: {e}")
 
-        layer = self.context.layers[self.config["primary"]]
         for offset in layer.scan(
-            context=self.context, scanner=scanners.RegExScanner(regex_pattern)
+            context=self.context, scanner=scanners.RegExScanner(pattern_bytes)
```
ikelos:
```suggestion
            context=context, scanner=scanners.RegExScanner(pattern_bytes)
```
This should be context

volatility3/framework/plugins/regexscan.py
```diff
@@ -46,38 +46,50 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+                # the match is not with the results_data (e.g. it doesn't fit within maxsize)
                 text_result = str(result_data, encoding="UTF-8", errors="replace")
                 bytes_result = result_data
 
             yield 0, (format_hints.Hex(offset), text_result, bytes_result)
 
     def run(self):
+        pattern = self.config.get("pattern")
+        maxsize = self.config.get("maxsize", self.MAXSIZE_DEFAULT)
+        layer = self.context.layers[self.config["primary"]]
```
ikelos:
```suggestion
        layer_name = self.config["primary"]
```

volatility3/framework/plugins/regexscan.py
```diff
@@ -46,38 +46,50 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+
         return renderers.TreeGrid(
             [
                 ("Offset", format_hints.Hex),
                 ("Text", str),
                 ("Hex", bytes),
             ],
-            self._generator(self.config.get("pattern")),
+            self._generator(layer, pattern, maxsize),
```
ikelos:
```suggestion
            self._generator(self.context, layer_name, pattern, maxsize),
```

ikelos (commented):
Looks much better, thanks.  Just a couple more tweaks, nearly there  5:)

ikelos (approved):
Ok, looks reasonable, thanks.  5:)

PR #1830, opened 2025-06-07
Malware categorization: windows.malfind

SolitudePy:
Categorize windows.malfind as windows.malware.malfind
- no dep fixes applied


ikelos (approved):
Looks good, thanks  5:)

PR #1837, opened 2025-06-07
Malware categorization: windows.svcdiff

SolitudePy:
Categorize windows.svcdiff as windows.malware.svcdiff
- no dep fixes applied
- since this plugin doesnt have a run() method the wrapper must inherit from the same base class
-  `__init__` method included since its skipped when copying methods.

### DIFF
--- a/volatility3/framework/plugins/windows/malware/svcdiff.py
+++ b/volatility3/framework/plugins/windows/malware/svcdiff.py
@@ -1,0 +1,102 @@
+# This file is Copyright 2024 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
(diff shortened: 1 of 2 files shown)


ikelos (commented):
I'm not sure this is worth splitting away from SvcScan (which is what it's based on) to put it under the malware category.  One for @atcuno to speak to...

SolitudePy:
@ikelos well it is a malware plugin, and linux.svcscan isnt...

atcuno:
svcdiff only exists to detect malware (unlinked services). We don't have a linux.svcscan that I can find, and I am not sure what it would do.

SolitudePy:
@atcuno ahh I meant windows.svcscan mb😳

ikelos (approved):
Yep, looks good, thanks.

PR #1818, opened 2025-05-25
inlinehooks plugin & change to avpatch

SolitudePy:
Hello, not sure if thats good enough as a standalone plugin, but I tried to re-implement apihooks. however, since there a lot of "hooks" in legitimate dlls I couldn't get it to output less false positives, ofc I am open to ideas. 
also, some hooks are not relevant anymore e.g KiFastSystemCall. 
even the current implementation of check_inline_hooks throw false positives, anyways that was a fun experience :D
I have left some of the pseudo-code commented for now, if you think it can be improved or removed totally.

Also I have 2 questions:
I have read a paper of @atcuno about hooktracer but couldn't find code online.
second question is something I also tried asking on your slack, maybe I misunderstand something but worth a try:
`
From my understanding, unhooked_system_calls plugin will enumerate all of the processes to find differences, if every process is hooked it will not be flagged as malware. a follow-up to this, a DLL is loaded to a shared memory and if it's patched the OS creates a private copy for the process, so cant we compare the hooks against the known shared read-only memory for the DLL? if not why?
`

Anyways, about the plugin:
```
(venv) ubuntu@ubuntuPC:~/Dev/volatility3$ vol -f ~/dumps/etw_patch_dump.raw windows.inlinehooks --pid 2108
Volatility 3 Framework 2.26.2
Progress:  100.00               PDB scanning finished                        
PID     Process Hook Type       Module  Hook Address    Function        Hook Info       Hook Hexdump    Disasm

2108    patch-etw-x64.  Inline Hook     ntdll.dll       0x7ff8013bf1f0  EtwEventWrite   Early RET
c3 8b dc 48 83 ec 58 4d 89 4b e8 33 c0 45 89 43 ...H..XM.K.3.E.C
e0 45 33 c9 49 89 43 d8                         .E3.I.C.        
0x7ff8013bf1f0: ret
0x7ff8013bf1f1: mov     ebx, esp
0x7ff8013bf1f3: sub     rsp, 0x58
0x7ff8013bf1f7: mov     qword ptr [r11 - 0x18], r9
0x7ff8013bf1fb: xor     eax, eax
0x7ff8013bf1fd: mov     dword ptr [r11 - 0x20], r8d
0x7ff8013bf201: xor     r9d, r9d
0x7ff8013bf204: mov     qword ptr [r11 - 0x28], rax
```

```
[...]

PR #1838, opened 2025-06-07
Malware categorization: windows.unhooked_system_calls

SolitudePy:
Categorize windows.unhooked_system_calls as windows.malware.unhooked_system_calls
- no dep fixes applied
- class name was changed to be different from module name


SolitudePy:
btw why is this called unhooked_system_calls? these ntdll.dll stub functions are basically user-mode wrapper for the actual syscall which code is executed in kernel (hence its not hooking SSDT, or ntoskrnl syscall implementations), also unhooked refers to this plugin show the total implementations? (hence `unhooked` and not `hooked`) ?

volatility3/framework/plugins/windows/unhooked_system_calls.py
```diff
@@ -1,202 +1,20 @@
...
-from volatility3.plugins.windows import pslist, pe_symbols
+from volatility3.framework import interfaces, deprecation
+from volatility3.plugins.windows.malware import unhooked_system_calls
 
 vollog = logging.getLogger(__name__)
 
 
-class unhooked_system_calls(interfaces.plugins.PluginInterface):
-    """Looks for signs of Skeleton Key malware"""
+class UnhookedSystemCalls(
```
ikelos: This doesn't have the same name, meaning people trying to run `unhooked_system_calls.unhooked_system_calls` will fail.  Since the point of this is to provide continuity, please don't change the plugin name here.

volatility3/framework/plugins/windows/malware/unhooked_system_calls.py
```diff
@@ -0,0 +1,202 @@
...
+from volatility3.framework import interfaces, exceptions
+from volatility3.framework import renderers
+from volatility3.framework.configuration import requirements
+from volatility3.framework.objects import utility
+from volatility3.plugins.windows import pslist, pe_symbols
+
+vollog = logging.getLogger(__name__)
+
+
+class UnhookedSystemCalls(interfaces.plugins.PluginInterface):
```
ikelos, line 21: This is a change in plugin name, I'll ask @atcuno if he's happy with the name change.  I don't mind either way (and mostly I don't expect people to type it, given it contains capital letters, and volatility will let you get away with the module name) but I want to check before making a change like this.

I understand python classes typically start with a capital letter, and I'm not sure how we diverged from that, but it seems we have in places and I agree that if we want to change it, doing so during a plugin rename is probably the right time.

ikelos (commented):
There's a couple of changes snuck in there that just need to be checked first.

volatility3/framework/plugins/windows/malware/unhooked_system_calls.py
SolitudePy: also I did it since it couldnt tell the difference between the module name and class name when I tried to call import it and then use it

volatility3/framework/plugins/windows/unhooked_system_calls.py
SolitudePy: @ikelos I get the following error when I do that:

```
Issue in module volatility3.plugins.windows.unhooked_system_calls: 
line 6, col 0: Direct import of unhooked_system_calls (<class 'abc.ABCMeta'>) 
from module volatility3.plugins.windows.malware - 
change to 'from volatility3.plugins.windows import malware 
and using malware.unhooked_system_calls
Found 1 issues
```

Also, by continuity you mean for 3rd party tools that may use the framework right? because in the framework none is using it this way.

atcuno:
> btw why is this called unhooked_system_calls? these ntdll.dll stub functions are basically user-mode wrapper for the actual syscall which code is executed in kernel (hence its not hooking SSDT, or ntoskrnl syscall implementations), also unhooked refers to this plugin show the total implementations? (hence `unhooked` and not `hooked`) ?

The hooking is done by EDRs to monitor the flow. See my talk at DEF CON last year:

https://www.youtube.com/watch?v=PmqvBe1LSZc&t=7s

atcuno:
Related to this `.malware.` push, @SolitudePy if I create a ticket with the list of other "malware" plugins will you complete the set? For example on Windows:

- debugregisters.py
- direct_system_calls.py
- driverirp.py
- drivermodule.py
- etwpatch.py
- hollowprocesses.py
- ldrmodules.py
- orphan_kernel_threads.py
- processghosting.py
- suspended_threads.py
- suspicious_threads.py

There aren't that many plugins that aren't capable of showing malware in some way, so if we are going to categorize this way then we need to move them all.

SolitudePy:
> Related to this `.malware.` push, @SolitudePy if I create a ticket with the list of other "malware" plugins will you complete the set? For example on Windows:
> 
> * debugregisters.py
> * direct_system_calls.py
> * driverirp.py
> * drivermodule.py
> * etwpatch.py
> * hollowprocesses.py
> * ldrmodules.py
> * orphan_kernel_threads.py
> * processghosting.py
> * suspended_threads.py
> * suspicious_threads.py
> 
> There aren't that many plugins that aren't capable of showing malware in some way, so if we are going to categorize this way then we need to move them all.

well I have already have many other open PRs for this (including linux ones), havent finished yet, some plugins I saw less of a malware plugin since its not direct like suspicious_, malfind. E.g im not sure to categorize suspended_threads since its not a certain logic but like an indicator, btw about etwpatch if you could check my PR on it #1818

ikelos:
> There aren't that many plugins that aren't capable of showing malware in some way, so if we are going to categorize this way then we need to move them all.

The idea is not to move *all* the plugins cluttering the os namespaces into their own malware section just because they *can* be used to detect malware.  That's just shifting the problem from one bin to another.  Ideally the category would be for plugins whose only or main purpose is for spotting malware.  Anything that just provides information about a sample (that happens to be used for malware or *could* be used for malware) should stay part of the main group.  For example, the IRP stuff seems like it's just about drivers, rather than malware in particular.  Same for debugregisters.  Hollowprocesses seems like a facet specifically used for/malware primarily.

@SolitudePy I am going to get through the linux plugins too, it's just a matter of time I'm afraid.

ikelos:
@atcuno If you could draw up a list of things that are specifically or primarily only used for malware, that would help us narrow some stuff down.  If we've already moved some over, I'm happy reverting the changes, really just want to get the main os category dumping grounds for plugins a little better ordered, so people can find things they're interested in.

atcuno:
I can make a list, but:

1) suspicious_threads only prints out threads that are suspicious, otherwise the output is empty.. its 100% for malware

2) suspend_threads are only interesting in an analysis context for looking for malware and was written as part of the DEF CON talk for detecting EDR evading malware

ikelos:
That's fine, put them in the malware list, just don't want *everything* ending up in there.  Technically pslist can be used to find malware, but that's not it's primary purpose.  Happy with whatever you choose, as long as the kitchen sink's not in there...  5;)

SolitudePy:
What is the status of this PR, do I need to change anything?

ikelos:
Nope, it looks like you resolved the comments I put in place, I just need to recheck it.

volatility3/framework/plugins/windows/unhooked_system_calls.py
ikelos: Yeah, I mean other people using the library and us not pulling the rug out from under them by changing things without telling them.  It's a small point, next to no one uses the plugin class names, but still, we definitely can't change what the old one was called, because then we're just removing it with no warning...

ikelos (approved):
Looks good now, thanks  5:)

PR #1841, opened 2025-06-07
Malware categorization: linux.check_creds

SolitudePy:
Categorize linux.check_creds as linux.malware.check_creds
- Fixed 1 dep in test
- Changed to linux.malware.check_creds in doc plugins preview


ikelos:
I can't tell if this is only used for malware or is a more general plugin, so I'll wait on @atcuno 's list to decide whether to accept this or leave it where it is.

ikelos (approved):
Looks good, thanks  5:)

PR #1842, opened 2025-06-07
Malware categorization: linux.check_idt

SolitudePy:
Categorize linux.check_idt as linux.malware.check_idt
- fixed doc & test


ikelos (approved):
Looks good, I think modifications of the IDT are probably only of use for malware.

PR #1845, opened 2025-06-07
Malware categorization: linux.check_afinfo & linux.hidden_modules

SolitudePy:
Categorized 2 plugins: linux.check_afinfo & linux.hidden_modules to linux.malware.check_afinfo & linux.malware.hidden_modules
- Fixed doc & test


ikelos (approved):
These are both clearly malware plugins, but I'd much prefer individual PRs.  It'd only take a little longer and it means that if one plugin stop the PR, the other could still be moved.  It also would make reversions that might come up much easier.  Please try to ensure future PRs are per plugin...  5:S

SolitudePy:
@ikelos sorry, but that was very repetitive to do... id rather fix a PR if theres such a problem... prefer to waste time on actual volatility3 code :D

SolitudePy:
@ikelos also, I havent come across any dependency problems so it was all rather easy to do... only need to create the deprecation wrapper & fix some tests so I would say it shouldnt be so dramatic:P

PR #1846, opened 2025-06-07
Malware categorization: linux.malfind & linux.keyboard_notifiers

SolitudePy:
Categorized both plugins to linux.malware 
- including test fix


ikelos (approved):
These also appear to be predominantly used for malware analysis, but the same comment holds true about combining too many things in a single PR...

PR #1850, opened 2025-06-10
Malware categorization: linux.tty_check, linux.netfilter, linux.modxview

SolitudePy:
Categorized the above plugins as `.malware.`
- including test fixes


ikelos:
Netfilter is definitely not one I'd class as malware only.  Please separate this into individual PRs for each plugin, so they can be evaluated on their own merits.  I've no idea about tty_check, but I also need to think about whether modxview is predominantly a malware plugin or not.

SolitudePy:
@ikelos tty_check is declared to find tty hooks as in Art of Memory Forensics, modxview is meant to quickly find hidden modules that rootkit do and from my understanding of netfilter code it specifically show all filters with a "Is Hooked" column to show if the hook is from a known legitimate module, so pretty much also for malware finding?

atcuno:
- check tty is for malware
- we only enumerate netfilter hooks to find rootkits that hook them, so this would be malware category also
- modxview is to visualize finding rootkits, so this would be malware also.

SolitudePy:
@ikelos lmk if you insist on seperate PRs, I will do it

ikelos (approved):
Oookay, well if these are all malware plugins then this looks fine.

ikelos:
@atcuno I'm surprised the netfilter code doesn't display firewire rules from memory, but ok?

ikelos:
Andrew's said they're all in malware, so no need in this instance but in future, please keep each PR an individual and unique piece of work, to avoid potential issues like this one might've been.

PR #1853, opened 2025-06-17
Add dumping dirty pages to Linux Malfind

tvanegro:
This PR is for adding `--dump-page` and `--dump-size` options to `linux.malware.Malfind` plugin. 

What does it do?

When supplied to `Malfind`, the `--dump-page` argument will dump each dirty page for the given process, instead of the default behaviour which will only dump the memory region inside which the dirty page(s) is/are located. Prior to this MR, Volatility will also stop at the first hit it finds, instead of looking for all dirty pages. 

This allows to precisely identify the location and content of the dirty memory sections. This is helpful in scenarios for which there are only a few dirty pages out of the whole memory region. 

The `--dump-size` argument allows to change the length of the disassembly for each page. The default is 64 bytes. 
It can be set to 0 (no disassembly printed, useful for a summary view). 

Why not making `--dump-page` default ? 

In scenarios where the whole binary executable region has been made dirty, `--dump-page` will list 1000's of dirty pages, which is not helpful.

This PR does not change the default behaviour of the Malfind plugin: 

Summary: 

| `--dump-page` supplied ? | `--dump-size` supplied ? | Behaviour|
| --- | ---| ---| 
| N | N | Dumps the first 64 bytes of the dirty memory *region* (default behaviour prior to this PR)| 
| Y | N | Dumps the first 64 bytes of each dirty page of each dirty region | 
| Y | Y | Dumps the first X bytes of each dirty pages of each dirty region (can be set to 0 to disable disassembly printout)| 
| N | Y | Dumps the first X bytes of the dirty memory *region*|

### DIFF
--- a/volatility3/framework/plugins/linux/malware/malfind.py
+++ b/volatility3/framework/plugins/linux/malware/malfind.py
@@ -68,7 +68,7 @@ class Malfind(interfaces.plugins.PluginInterface):
         # Dumping page defaults to off, as in case a whole r-xp region is dirty
         # this would likely dump 1000's of pages which might not always be wise nor necessary
 
-        dump_page = self.config.get("dump-page") or False
+        dump_page = self.config["dump-page"]
 
         for vma in task.mm.get_vma_iter():
             vma_name = vma.get_name(self.context, task)
--- a/volatility3/framework/plugins/linux/malware/malfind.py
+++ b/volatility3/framework/plugins/linux/malware/malfind.py
@@ -38,12 +38,12 @@ class Malfind(interfaces.plugins.PluginInterface):
                 optional=True,
             ),
             requirements.IntRequirement(
-                name="dump_size",
+                name="dump-size",
                 description="Amount of bytes to dump for each dirty region/page found - Default 64 bytes",
                 optional=True,
             ),
             requirements.BooleanRequirement(
-                name="dump_page",
+                name="dump-page",
                 description="Dump each dirty page and content - Default off",
                 optional=True,
             ),
@@ -62,12 +62,12 @@ class Malfind(interfaces.plugins.PluginInterface):
         proc_layer = self.context.layers[proc_layer_name]
 
         # Allowing a dump_size of 0 (no dump)
-        dump_size = self.config.get("dump_size") if self.config.get("dump_size") is not None else 64
+        dump_size = self.config.get("dump-size") if self.config.get("dump-size") is not None else 64
 
         # Dumping page defaults to off, as in case a whole r-xp region is dirty
         # this would likely dump 1000's of pages which might not always be wise nor necessary
 
-        dump_page = self.config.get("dump_page") or False
+        dump_page = self.config.get("dump-page") or False
 
         for vma in task.mm.get_vma_iter():
             vma_name = vma.get_name(self.context, task)
--- a/volatility3/framework/plugins/linux/malware/malfind.py
+++ b/volatility3/framework/plugins/linux/malware/malfind.py
@@ -47,6 +47,7 @@ class Malfind(interfaces.plugins.PluginInterface):
                 name="dump-page",
                 description="Dump each dirty page and content - Default off",
                 optional=True,
+                default=False,
             ),
         ]
 
--- a/volatility3/framework/plugins/linux/malware/malfind.py
+++ b/volatility3/framework/plugins/linux/malware/malfind.py
@@ -61,12 +61,7 @@ class Malfind(interfaces.plugins.PluginInterface):
 
         proc_layer = self.context.layers[proc_layer_name]
 
-        # Allowing a dump_size of 0 (no dump)
-        dump_size = (
-            self.config.get("dump-size")
-            if self.config.get("dump-size") is not None
-            else 64
-        )
+        dump_size = self.config.get("dump-size", None) or 64
 
         # Dumping page defaults to off, as in case a whole r-xp region is dirty
         # this would likely dump 1000's of pages which might not always be wise nor necessary
--- a/volatility3/framework/plugins/linux/malware/malfind.py
+++ b/volatility3/framework/plugins/linux/malware/malfind.py
@@ -63,7 +63,7 @@ class Malfind(interfaces.plugins.PluginInterface):
 
         proc_layer = self.context.layers[proc_layer_name]
 
-        dump_size = self.config.get("dump-size", None) or 64
+        dump_size = self.config["dump-size"]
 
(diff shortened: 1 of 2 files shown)


ikelos (commented):
The `_list_injections` method isn't shared out as class method, so changing it here is fine, but since the functionality/output changes, this needs at least a PATCH version bump, meaning this should be `(1, 0, 4)`.

tvanegro:
Thanks for the review, I've bumped the version to `(1, 0, 4)`.

volatility3/framework/plugins/linux/malware/malfind.py
```diff
@@ -51,14 +61,41 @@ def _list_injections(
 
         proc_layer = self.context.layers[proc_layer_name]
 
+        # Allowing a dump_size of 0 (no dump)
+        dump_size = (
+            self.config.get("dump-size")
+            if self.config.get("dump-size") is not None
+            else 64
+        )
```
ikelos:
```suggestion
        dump_size = self.config.get("dump-size", None) or 64
```
This will prevent you setting a dump-size of 0, but that's actually helpful and it's a lot easier to read than the previous 4-liner...  5;)

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1273,6 +1273,32 @@ def get_name(self, context, task) -> Optional[str]:
         except exceptions.InvalidAddressException:
             return None
 
+    def get_malicious_pages(self, proclayer=None):
```
ikelos: Adding this method to the extension needs *a* version number bumping, I'm just not sure which one.  I *think* it might be the MINOR version for the whole framework unfortunately?  This one file is getting out of hand (~1300 lines of code) and we should undertake to split it which might allow us to individually version off the various components, but for now, if you want people to be able to use the `get_malicious_pages` function, you'll need to bump the closest MINOR version for the container of this code I'm afraid.

ikelos: Also, please you could type the function signature?  What is `proclayer`, what does this method return, please add a docstring and all the other good stuff like that...  5:)

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1273,6 +1273,32 @@ def get_name(self, context, task) -> Optional[str]:
...
+        """
+        This function will return a list of all malicious pages inside a given dirty region
+        """
+        malicious_pages = []
+        flags_str = self.get_protection()
+
+        if (
```
ikelos, line 1293: It seems a bit odd to allow proclayer to be None, but then fail to return anything useful if it's not set?  I'd probably not bother giving it a default value?

ikelos (commented):
Thanks, another few comments after another look at it.  We *really* need to do something about how big the linux extensions file is getting, but this probably isn't the place to do it.  It will need a version bump of the MINOR version of the whole framework sadly though.

volatility3/framework/symbols/linux/extensions/__init__.py
tvanegro: Agreed, addressed it and removed the default value.

volatility3/framework/symbols/linux/extensions/__init__.py
tvanegro: Comments addressed. Thank you :) 

For the framework version bump, would you mind pointing me to the right one to bump ? I am not exactly sure.

ikelos: Sorry for the delay.  It would be in `volatility3/framework/constants/_version.py` which would need `VERSION_MINOR` bumping to 28 at the moment...

ikelos:
I'd also like to get @atcuno or another linux dev to look over this and make sure it makes sense.  I can only really review it from a code perspective, but thank you for making the changes and for bearing with us whilst we find time to look over it and get it included...

volatility3/framework/plugins/linux/malware/malfind.py
```diff
@@ -51,14 +61,36 @@ def _list_injections(
 
         proc_layer = self.context.layers[proc_layer_name]
 
+        dump_size = self.config.get("dump-size", None) or 64
```
atcuno: this and the dump-page one should use the `default` parameter to the Int and Boolean requirements on 40 and 45 instead of this way of setting a default.

atcuno:
I added a comment about how the default values are set. After these are changed, this is good to merge @ikelos

volatility3/framework/plugins/linux/malware/malfind.py
```diff
@@ -37,6 +37,16 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                 element_type=int,
                 optional=True,
             ),
+            requirements.IntRequirement(
+                name="dump-size",
+                description="Amount of bytes to dump for each dirty region/page found - Default 64 bytes",
+                optional=True,
```
ikelos, line 43:
```suggestion
                optional=True,
                default=64,
```
As mentioned below, the default parameter will ensure the value is properly set.

volatility3/framework/plugins/linux/malware/malfind.py
```diff
@@ -37,6 +37,16 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
...
+            ),
+            requirements.BooleanRequirement(
+                name="dump-page",
+                description="Dump each dirty page and content - Default off",
+                optional=True,
```
ikelos, line 49:
```suggestion
                optional=True,
                default=False,
```

volatility3/framework/plugins/linux/malware/malfind.py
ikelos:
```suggestion
        dump_size = self.config["dump-size"]
```

volatility3/framework/plugins/linux/malware/malfind.py
```diff
@@ -51,14 +61,36 @@ def _list_injections(
 
         proc_layer = self.context.layers[proc_layer_name]
 
+        dump_size = self.config.get("dump-size", None) or 64
+
+        # Dumping page defaults to off, as in case a whole r-xp region is dirty
+        # this would likely dump 1000's of pages which might not always be wise nor necessary
+
+        dump_page = self.config.get("dump-page") or False
```
ikelos:
```suggestion
        dump_page = self.config["dump-page"]
```

ikelos (commented):
As @atcuno mentioned these are the following changes, after which we can get this merged in...  5:)

PR #1856, opened 2025-06-23
Plugins: remove unused unix argument in linux.sockstat

SolitudePy:
Remove unused unix argument in linux.sockstat


ikelos (approved):
Thanks for clearing this up.  5:)

PR #1859, opened 2025-06-27
Add convenience method for YaraScanner

JSCU-CNI:
This convenience method makes it easier to use the YaraScanner with a Python defined YARA rule. This makes one off, single file plugins easier to make because one can include the YARA rule directly in Python instead of having to include the rule in a separate YARA file. I.e.:

```python

RULE = """rule ExampleRule
{
    strings:
        $line = { 00 01 02 03 }

    condition:
        all of them
}
"""

class Plugin:
    [setup]
    
    def _generator(self, layer):
        rule = yarascan.YaraScanner.from_text(RULE)
        scanner = yarascan.YaraScanner(rules=rule)
        [...]
""""

### DIFF
--- a/volatility3/framework/plugins/yarascan.py
+++ b/volatility3/framework/plugins/yarascan.py
@@ -102,8 +102,20 @@ class YaraScanner(interfaces.layers.ScannerInterface):
             return yara.compile(file=fp)
 
     @classmethod
-    def from_text(cls, rule):
-        formatted_rule = rule.replace("\n", "")
+    def from_text(cls, rule) -> yara.Rules:
+        """Initialize a Yara Rules object from one or more rules in string format.
+
+        You can provide rules in single-line or multi-line:
+        rule = "rule dummy { condition: true }"
+        rules = '''
+            rule dummy {
+                condition: true
+            }
+            rule dummy2 {
+                condition: true
+            }
+        '''
+        """
         if USE_YARA_X:
             return yara_x.compile(source=formatted_rule)
         return yara.compile(source=formatted_rule)
--- a/volatility3/framework/plugins/yarascan.py
+++ b/volatility3/framework/plugins/yarascan.py
@@ -101,6 +101,13 @@ class YaraScanner(interfaces.layers.ScannerInterface):
                 return yara_x.compile(fp.read().decode())
             return yara.compile(file=fp)
 
+    @classmethod
+    def from_text(cls, rule):
+        formatted_rule = rule.replace("\n", "")
+        if USE_YARA_X:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: volatility3/framework/plugins/yarascan.py [hunk 1]
```diff
@@ -101,6 +101,13 @@ def from_file(cls, filepath):
                 return yara_x.compile(fp.read().decode())
             return yara.compile(file=fp)
 
+    @classmethod
+    def from_text(cls, rule):
+        formatted_rule = rule.replace("\n", "")
```
ikelos: Why are we stripping all new line characters, and is there any situation in which that could change the meaning of the rule?

ikelos: It also feels like `get_rule` could be modified to call this after wrapping its parameter into simple rule text?

ikelos (commented):
This seems fairly innocuous, but to avoid code duplication, I'd probably make use of it whenever text is being compiled into a pattern, so use it in `get_rule` as well.  You'd need to check if there's a subtle difference in output between `yara.compile(source=...)` and `yara.compile(sources=...)` but I don't think that should be too catastrophic for people to cope with.

volatility3/framework/plugins/yarascan.py
JSCU-CNI: New line stripping: good point! The [docs of yara-python](https://yara.readthedocs.io/en/latest/yarapython.html) contains an example with a single line rule hence I replaced the newlines. I tested it with ~20 rules, and the newline can be included without any problems. I removed the `replace` line of code.

The `source` vs `sources` thing: the main difference is the namespacing. Theoretically we could check whether the given rule is a string or a dict and use `source` and `sources` accordingly. Don't prefer that though because of it's implicitness. The function name `from_text` is also clear in that it requires text, and not a dict.

The `get_rule` currently uses namespaces while `from_text` does not. I don't think it'd be wise to change it to use `from_text` because that would result in a backwards incompatible change. What do you think?

ikelos: Yeah, fair enough.  Looks good then, thanks!  5:)

ikelos (approved):

ikelos:
Oh, `ruff` spotted that formatted_rule now isn't defined, we can just use rule.  5:)

ikelos:
Also, the type hinting uses `yara` which may not exist if `yara_x` is the only installed instance.  The previous methods avoided it by just not type hinting, but if you want to do it correctly, I guess assign the type to a variable after the import has been made, and then use that type in the signature...

PR #1861, opened 2025-06-27
Feature/parquet arrow renderer

### LINKED ISSUE #1860: Add Arrow and Parquet renderers for efficient data export
**Description**

While the currently implemented text-based renderers are useful for data processing, it would be even better to have the ability to export to Parquet or Arrow format, which is more amenable to further analysis using query engines like DuckDB or Polars and visualisation frameworks.

**Solution**
I want to implement Parquet and arrow renderers so that users can specify these options using the CLI.

The alternative is working with the existing JSON output, but that's very slow for large datasets.

**Additional information**
Example usage:

```sh
vol -f tmp/PE_Packing/memory.bin -r parquet windows.pslist > ~/projects/memory/pe_packing/windows2.pslist.parquet

vol -f tmp/PE_Packing/memory.bin -r arrow windows.pslist > ~/projects/memory/pe_packing/windows.pslist.arrow

```
Image

kyrre:
This adds Parquet and arrow renderers and closes #1860


ikelos (commented):
Thanks very much, this looks really nice.  I still want to check through a few more bits and pieces, but generally it looks good.

At the moment `text_renderer` is looking a little full though, I was wondering if you could test whether this still works, if all the new classes are moved into a file like `volatility3/plugins/parquet_renderer.py` or similar?  We should automatically import all python files under plugins, and I believe we do it before we choose any renderers, which means it should still show up in the available options.  We'll probably want to make a renderers directory that people can drop things into, this would just allow users to drop in their own renderers (and might therefore act as a good example of how to write new ones when people want).

Anyway, just some thoughts.  I'll give it a full review over the coming week and add in any comments of anything else, but I'm really pleased someone else has written a renderer for a very different format finally!  Thanks!  5:)

kyrre:
> At the moment `text_renderer` is looking a little full though, I was wondering if you could test whether this still works, if all the new classes are moved into a file like `volatility3/plugins/parquet_renderer.py` or similar? We should automatically import all python files under plugins, and I believe we do it before we choose any renderers, which means it should still show up in the available options. We'll probably want to make a renderers directory that people can drop things into, this would just allow users to drop in their own renderers (and might therefore act as a good example of how to write new ones when people want).

I really like the structure you're proposing here. However, it does not work at the moment because the plugins are imported after the list of available renderers is generated:

https://github.com/volatilityfoundation/volatility3/blob/997f2a4f4c32cbb268ae7236ab3a7e24a3ec9bb2/volatility3/cli/__init__.py#L109-L115

https://github.com/volatilityfoundation/volatility3/blob/997f2a4f4c32cbb268ae7236ab3a7e24a3ec9bb2/volatility3/cli/__init__.py#L196-L203

and then

https://github.com/volatilityfoundation/volatility3/blob/997f2a4f4c32cbb268ae7236ab3a7e24a3ec9bb2/volatility3/cli/__init__.py#L336-L338

ikelos:
Ok, thanks for investigating, I'll take a look into it...  5:)

kyrre:
> Ok, thanks for investigating, I'll take a look into it... 5:)

I created a new file in the plugin directory `parquet_renderer.py` and moved all the code related to the renderers in `cli/__init__.py` to after the plugins have been imported. What I moved:

- creating the list of renderers
- cli parsing and validation for the renderer arg
- display banner output redirection 

How do you feel about this @ikelos - is this something you can work with? I agree that it is cleaner to create a renderer directory, but maybe this is OK for now? 🤔

volatility3/framework/plugins/parquet_renderer.py
```diff
@@ -0,0 +1,166 @@
...
+from volatility3.cli.text_renderer import (
+    CLIRenderer,
+    display_disassembly,
+    LayerDataRenderer,
+)
+
+vollog = logging.getLogger(__name__)
+
+try:
+    ARROW_PRESENT = True
```
ikelos: I'd probably set ARROW_PRESENT to False outside/before the try, and then only when everything's imported successfully set it to true.  This way round should work, but feels like there could more easily be a failure of some kind.  It's not a show stopper though.  We *probably* should make sure this stuff is versioned though, in case someone decides they want to extend it, or a future version changes the output in some way...  5:S

ikelos (commented):
Sorry to be so indecisive (and to take so long to rereview it).  I think it's much closer and I'm pretty happy with the changes to the main CLI, which is likely the biggest hurdle.  The trick now is to get this renderer right (since it'll likely be used as the template for everybody's future renderers).  As such, could we try putting it inside of volatility3/plugins/renderers please?  I don't *think* there's any functionality that will decide it's only for `renderer` OSes, but best to check it now rather than trying to change it later.  Other than that and the other comment I posted, I think it could be ready for prime time?

volatility3/framework/plugins/renderers/parquet_renderer.py
kyrre, line 1: I moved the parquet and arrow renderers into this file as requested, and there were no issues with the cli discovering them.

volatility3/framework/plugins/parquet_renderer.py
kyrre: > I'd probably set ARROW_PRESENT to False outside/before the try, and then only when everything's imported successfully set it to true.

Yes, that definitely feels more safe. I have updated the file.

When it comes to versioning - what did you have in mind? Other files seem to use a tuple class attribute, 

`_version = (1, 0, 0)` 

I added this attribute to both he arrow and parquet renderer classes.

volatility3/framework/plugins/renderers/parquet_renderer.py
```diff
@@ -0,0 +1,173 @@
...
+            raise RuntimeError("Arrow output format requires the pyarrow package")
+
+        self._to_arrow_type = {
+            renderers.Disassembly: pa.utf8,
+            bool: pa.bool_,
+            int: pa.int64,
+            float: pa.float64,
+            str: pa.utf8,
+            datetime.datetime: lambda: pa.timestamp("ms"),
+            format_hints.Bin: pa.uint64,
```
kyrre, line 53: When running all the windows plugins locally, I noticed that there were some integer overflow issues, and so I changed the `format_hints.Bin` and `format_hints.Hex` to map to `pa.uint64` instead.

kyrre:
> Sorry to be so indecisive (and to take so long to rereview it). I think it's much closer and I'm pretty happy with the changes to the main CLI, which is likely the biggest hurdle. The trick now is to get this renderer right (since it'll likely be used as the template for everybody's future renderers). As such, could we try putting it inside of volatility3/plugins/renderers please? I don't _think_ there's any functionality that will decide it's only for `renderer` OSes, but best to check it now rather than trying to change it later. Other than that and the other comment I posted, I think it could be ready for prime time?

I think we should be set! What do you think @ikelos ? Another thing we could add is a test suite perhaps 🤔

kyrre:
I went ahead and added some tests! 

The testing strategy I choose was to port a few of the plugin tests and express the assertions using a [PyArrow table](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html) that was instantiated from both Parquet/Arrow data emitted to stdout by the plugins.

kyrre:
It didn't work correctly with plugins that created a nested output, e.g., windows.pstree. I fixed this by flattening and adding _vol_id and _vol_parent_id columns to the arrow schema.

volatility3/framework/plugins/renderers/parquet_renderer.py
```diff
@@ -0,0 +1,221 @@
...
+            fields.append(pa.field(column.name, arrow_type()))
+
+        # if the output is nested, e.g., windows.pstree
+        if self._is_tree_result:
+            fields.append(pa.field("_vol_id", pa.uint64()))
+            fields.append(pa.field("_vol_parent_id", pa.uint64()))
+
+        return pa.schema(fields)
+
+    def _flatten_tree_structure(self, nested: list[dict]) -> list[dict]:
```
dgmcdona: The minimum Python version for volatility3 is 3.8, so these hints need to use `typing.List` and `typing.Dict`.

volatility3/framework/plugins/renderers/parquet_renderer.py
```diff
@@ -0,0 +1,221 @@
...
+
+class ArrowRenderer(CLIRenderer):
+    """Renderer that outputs Arrow IPC format data."""
+
+    name = "arrow"
+    structured_output = True
+    _version = (1, 0, 0)
+
+    def __init__(
+        self, options: List[interfaces.renderers.RenderOption] | None = None
```
dgmcdona: For Python 3.8 support, we'll need this to be `typing.Optional` instead of using the pipe syntax.

volatility3/framework/plugins/renderers/parquet_renderer.py
```diff
@@ -0,0 +1,221 @@
...
+        Args:
+            nested: A list of dicts with optional `__children` lists (tree nodes).
+
+        Returns:
+            A flat list of dicts with `_vol_id` and `_vol_parent_id`.
+        """
+        rows = []
+        self._node_id_counter = 0
+
+        def _process_node(node: dict, parent_id: int | None):
```
dgmcdona: Convert to `typing.Optional`

dgmcdona (changes requested):

volatility3/framework/plugins/renderers/parquet_renderer.py
kyrre: Thank you for the review! I have updated the type hints now.

volatility3/framework/plugins/renderers/parquet_renderer.py
```diff
@@ -0,0 +1,222 @@
...
+    Tuple,
+    TextIO,
+)
+from volatility3.framework import interfaces, renderers
+from volatility3.framework.renderers import format_hints
+from volatility3.cli.text_renderer import (
+    CLIRenderer,
+    display_disassembly,
+    LayerDataRenderer,
+)
```
ikelos:
```suggestion
from volatility3.cli import text_renderer
```

The [volatility coding style](https://github.com/volatilityfoundation/volatility3/blob/develop/CODING_STYLE.md#imports) is to import only modules, so that people don't import a class directly from a module where it had been imported itself.  Please could you update these.  We allow it for typing because they're often repeated and very common.

volatility3/framework/plugins/renderers/parquet_renderer.py
```diff
@@ -0,0 +1,222 @@
...
+try:
+    import pyarrow as pa
+    import pyarrow.parquet as pq
+
+    ARROW_PRESENT = True
+except ImportError:
+    vollog.debug("Arrow/Parquet libraries not found")
+
+
+class ArrowRenderer(CLIRenderer):
```
ikelos:
```suggestion
class ArrowRenderer(text_renderer.CLIRenderer):
```

volatility3/framework/plugins/renderers/parquet_renderer.py
```diff
@@ -0,0 +1,222 @@
...
+                if column in ignore_columns:
+                    continue
+
+                data = list(node.values)[column_index]
+
+                if isinstance(data, interfaces.renderers.BaseAbsentValue):
+                    data = None
+
+                if isinstance(data, renderers.Disassembly):
+                    data = display_disassembly(data)
```
ikelos:
```suggestion
                    data = text_renderer.display_disassembly(data)
```

volatility3/framework/plugins/renderers/parquet_renderer.py
```diff
@@ -0,0 +1,222 @@
...
+
+                if isinstance(data, renderers.LayerData):
+                    data = LayerDataRenderer().render_bytes(data)[0]
```
ikelos:
```suggestion
                    data = text_renderer.LayerDataRenderer().render_bytes(data)[0]
```

ikelos (commented):
Thanks for holding with us through the delay.  This looks good now, just a couple more comments and then it should be able to land.  5:)

test/renderers/test_parquet_renderers.py
```diff
@@ -0,0 +1,108 @@
...
+from abc import ABC, abstractmethod
+from test import test_volatility
+
+HAS_PYARROW = False
+try:
+    import pyarrow as pa
+    import pyarrow.parquet as pq
+    import pyarrow.compute as pc
+    HAS_PYARROW = True
+except ImportError:
```
ikelos, line 12:
```suggestion
except ImportError:
   # The user doesn't have pyarrow installed, but HAS_PYARROW will be false so just continue
```
Last change to get it over the line...  5:)

kyrre: Thank you 😁

PR #1866, opened 2025-07-12
Hotfix/v2.26.1 - Resolve documentation not building for stable.

ikelos:
This resolves the documentation failing to build for the past five months (including the stable release).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ test = [
 docs = [
     "volatility3[dev]",
     "sphinx>=4.0.0,<9",
-    "sphinx-autodoc-typehints>=2.0.0,<3",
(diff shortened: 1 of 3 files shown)


ikelos:
@npetroni Found the culprit (capped version of `sphinx-autodoc-typehints`).  This commit resolves it and (as shown on develop) fixes the documentation, but we need to roll a new stable to keep all the version numbers matching.  Lemme know when would be a good time to get it rolled.  No signfiicant changes, so happy to just go through our checklist solo if you want?

ikelos:
Not sure if 2.26.1 is the right version number?  Didn't notice we'd already used it...  5:S  Develop is on 2.26.2 already

ikelos:
This made it in the most recent release and therefore the hotfix is no longer needed.

PR #1868, opened 2025-07-21
Linux: Fix broken mount enumeration on 6.13+

Abyss-W4tcher:
Mount enumeration is currently broken on 6.13+ kernels. This is because "rb_node" entries are cast to "mount" entries using "mount.mnt_list" member instead of "mount.mnt_node".

The kernel's node_to_mount method uses "mnt_node":

* https://elixir.bootlin.com/linux/v6.9.12/source/fs/namespace.c#L1011

Enumeration did not break on samples ranging from 6.8 to 6.13 because "mnt_node" and "mnt_list" were under a common anonymous struct (at the same struct offset then), but a recent commit moved the "mnt_list" member out of it:

* https://github.com/torvalds/linux/commit/344bac8f0d73fe970cd9f5b2f132906317d29e8b#diff-2c08c5f537fd972c88159feea1a9cd61e283d628ae62914b1b6ee524204bef27L54

### DIFF
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -1971,8 +1971,9 @@ class mnt_namespace(objects.StructType):
                 self._context, self
             )
             for node in self.mounts.get_nodes():
+                # See kernel's node_to_mount()
                 mnt = linux.LinuxUtilities.container_of(
-                    node, "mount", "mnt_list", vmlinux
+                    node, "mount", "mnt_node", vmlinux
                 )
                 yield mnt
         else:


ikelos:
Ok, one for the linux team to check and let me know if it's issue free to merge please @atcuno ...  5:)

the-rectifier:
Hey all, just my two cents here. 
I have been working with a Debian 6.12 sample where `pagecache.Files` was crashing the framework. So I checked `linux.mountinfo` and was empty as well. Applying this PR fixed both issues. So I guess some 6.12 kernels are affected as well

`Linux version 6.12.48+deb13-amd64 (debian-kernel@lists.debian.org) (x86_64-linux-gnu-gcc-14 (Debian 14.2.0-19) 14.2.0, GNU ld (GNU Binutils for Debian) 2.44) # SMP PREEMPT_DYNAMIC Debian 6.12.48-1 (2025-09-20)`

Abyss-W4tcher:
> Hey all, just my two cents here. 
> I have been working with a Debian 6.12 sample where `pagecache.Files` was crashing the framework. So I checked `linux.mountinfo` and was empty as well. Applying this PR fixed both issues. So I guess some 6.12 kernels are affected as well
> 
> `Linux version 6.12.48+deb13-amd64 (debian-kernel@lists.debian.org) (x86_64-linux-gnu-gcc-14 (Debian 14.2.0-19) 14.2.0, GNU ld (GNU Binutils for Debian) 2.44) # SMP PREEMPT_DYNAMIC Debian 6.12.48-1 (2025-09-20)`

I'm glad that it fixed the issue! Those changes can be backported to older kernel versions so that's not surprising 👍.

Abyss-W4tcher:
Hey @ikelos, I think this one should be of higher priority now that 6.13+ gets distributed more and more. Andrew should have given it a look, you can also check the description to see that's it's a really simple mistake to fix.

ikelos:
I understand it's a small change, but I'd really wanted @atcuno or just a second pair of eyes from the linux team to OK it, because I have no idea how significant that small change actually is.  Does it work for all previous versions, will it work for all future versions... why wasn't it coded that way first time around, etc, etc?  If you can answer that, and it appears as though @atcuno hasn't been able to review it since July, then I'll merge it in, I just want a second opinion that I'm not gonna break everything by merging it in...  5;)

Abyss-W4tcher:
> I understand it's a small change, but I'd really wanted @atcuno or just a second pair of eyes from the linux team to OK it, because I have no idea how significant that small change actually is. Does it work for all previous versions, will it work for all future versions... why wasn't it coded that way first time around, etc, etc? If you can answer that, and it appears as though @atcuno hasn't been able to review it since July, then I'll merge it in, I just want a second opinion that I'm not gonna break everything by merging it in... 5;)

No problem! This was likely coded with `mnt_list` because `mnt_list` and `mnt_node` were part of the same union struct, so using any of these returned the same offset from the parent struct. As no exceptions were returned I guess the original developer did not catch that small mistake, but now that they aren't in the same union struct `mnt_list` is not the offset we want. I can ask atcuno to take a look tomorrow.

The kernel really wants `mnt_node`, see https://elixir.bootlin.com/linux/v6.9.12/source/fs/namespace.c#L1011 or https://elixir.bootlin.com/linux/v6.18-rc6/source/fs/namespace.c#L1085

ikelos:
Ok, as long as someone on the linux team is happy with it and we've got references, I'll merge it.  Thanks for taking a look!  5:)

PR #1871, opened 2025-08-29
Feature: vadyarascan enrichment

blitztide:
This add contextual data to each hit with vadyarascan, saves running pslist after processing. Original didn't have imagename or PPID


ikelos:
Thanks very much for your submission.  Since you're making changes to the plugin, please could you increment the MINOR version number of the plugin, so that people will be able to determine the format of the output the plugin provides without needing to run it?

blitztide:
Added, thank you

ikelos (approved):
Sorry for the delay in reviewing, but this looks good now, thanks for bearing with us!  5:)

PR #1878, opened 2025-09-15
Update text_renderer.py

dotslashed:
Json Text renderer not properly rendering offset values.  

Command ```python .\vol.py -f memory.dmp -r json windows.registry.hivelist | ConvertFrom-Json | Format-List *```

Output:
```Volatility 3 Framework 2.26.2
File output  : Disabled
FileFullPath :
Offset       : 145708949291008
__children   : {}

File output  : Disabled
FileFullPath : \REGISTRY\MACHINE\SYSTEM
Offset       : 145708949430272
__children   : {}

File output  : Disabled
FileFullPath : \REGISTRY\MACHINE\HARDWARE
Offset       : 145708950167552
__children   : {}

File output  : Disabled
FileFullPath : \SystemRoot\System32\Config\SOFTWARE
Offset       : 145708958879744
__children   : {}
```

New output:

```Volatility 3 Framework 2.26.2
File output  : Disabled
FileFullPath :
Offset       : 0x84858229c000
__children   : {}

File output  : Disabled
FileFullPath : \REGISTRY\MACHINE\SYSTEM
Offset       : 0x8485822be000
__children   : {}

File output  : Disabled
FileFullPath : \REGISTRY\MACHINE\HARDWARE
Offset       : 0x848582372000
__children   : {}

File output  : Disabled
FileFullPath : \SystemRoot\System32\Config\SOFTWARE
Offset       : 0x848582bc1000
__children   : {}
```

### DIFF
--- a/volatility3/cli/text_renderer.py
+++ b/volatility3/cli/text_renderer.py
@@ -540,6 +540,11 @@ class JsonRenderer(CLIRenderer):
             if not isinstance(x, interfaces.renderers.BaseAbsentValue)
             else "N/A"
         ),
+        format_hints.Hex: lambda x: (
+            f"0x{x:x}"
+            if not isinstance(x, interfaces.renderers.BaseAbsentValue)
+            else None
+        ),
         renderers.Disassembly: quoted_optional(display_disassembly),
         format_hints.MultiTypeData: quoted_optional(multitypedata_as_text),
         renderers.LayerData: lambda x: (


ikelos:
The values are numbers, so the JSON should represent them as numbers.  This patch converts them into strings and formats them for humans, but that makes processing the raw number more difficult for anything that ingests JSON and acts upon the data.  I don't see a good reason to turn the values into strings, what's the value in this patch other than specifically for the use case of a human using the JSON output rather than any of the other renderers?

dotslashed:
Hi @ikelos Yes this is a drawback for calculations and for any other hex values handled as numbers. The json format is more  user-friendly in terms of handling larger registry binary data, the CSV format breaks sometimes. Sometimes if we loose track of the json decimal hive offset, we have to convert it to hex from decimal separately and use the hex offset in commands or anytime we get a new offset value during our testing we have to first convert it to hex separately.
I have a request if it is possible to get both the number form and hex form in output as separate key:value pair and the number form as usual will be used for machine and string form for human viewing?  

For example final output becomes like this:
```
File output  : Disabled
FileFullPath :
Offset       : 145708949291008
Offset_hex   : 0x84858229c000
__children   : {}
```

Thanks!

ikelos:
You said that you can only provide hex values for offsets, but I don't think that's the case?  It's been a while since I've tried it, but I believe any numerical requirement is accepted on the command line as either decimal or as hex values (this is why you can specify a pid as just `10` rather than `0xa`).  This solves all the reasons I can see that you've provided for needing this invasive change to an output format.  I think it would also be possible to use powershell code such as [format operators](https://www.byteinthesky.com/powershell/how-to-convert-decimal-to-hexadecimal/)  to convert any decimal output to hexadecimal, if you needed (although I don't know poweshell well enough to work up a solution sadly)...

The point is, changing it from a unique number, to a string format that requires parsing for any further use isn't a good plan for the tool in general, and only helps in one very specific scenario.  This also isn't good reason to duplicate all the output values (since providing the same data in two different forms could lead to discrepancies in a tool trying to import it if the two ever hold different numerical values).

Thank you for your contribution, but in this instance I don't think it's an improvement and I'm going to close this off, but do please feel free to submit other pull requests if there are specific features that would be beneficial without reducing existing functionality or being overly specific for one particular work flow...

dotslashed:
Hi! Yes I have tested, the decimal --offset value works in the same way and I have also found that at some point conversion is required when verifying the volatility results with other memory forensic tool which shows in hexadecimal. This JSONRenderer code could help in a specific scenario but not in general as you have also mentioned.
Thanks for the powershell suggestion, will try that as well.

PR #1888, opened 2025-10-14
Remove amcache.py

j-t-1:
Removal date was 2025-09-25.


j-t-1:
@ikelos also cachedump.py, hashdump.py, lsadump.py and scheduled_tasks.py can be removed.

ikelos:
Hiya, thanks for the reminder.  As it turns out, that commit was only made in March of this year, and we don't think 6 months is enough time, so we're going to extend the date to 2026 to ensure everyone's seen the message and had time to transition.  As to removing the test, these plugins have just been moved under a different name (from `windows.amcache` to `windows.registry.amcache`, for example), so the test is still valid, but you're right, it does need updating to the correct plugin name.  I've put those changes in place in #1890 so you can follow what was done, but I'm likely to close this off.  Thanks again for reminding us, and feel free to drop in a reminder when the time rolls around again next year!  5:)

PR #1907, opened 2025-11-10
Process Connection Volatility3 Plugin

malfav:
# proccon.py – Process Connectivity & Visualization (Volatility 3 Plugin)

`proccon.py` is a **Volatility 3 plugin** designed to provide **process visualization and relationship mapping** from a **Windows memory dump**.  
While it does not perform direct forensic analysis, it serves as a **critical utility** for post-processing and visualizing process hierarchy data.

The plugin extracts **Parent-Child process relationships** and outputs them as a **Graphviz DOT file**, enabling investigators to generate graphical representations of the system’s process tree at the time of capture.

---

### **Process Tree Extraction**
Efficiently iterates through the complete process list obtained via Volatility’s `windows.pslist` plugin to map:
- **PID** (Process ID) → **PPID** (Parent Process ID)

This provides a foundational dataset for visual process relationship mapping.

---

### **Graphviz DOT Output**
Automatically generates a **structured `.dot` file** compatible with Graphviz visualization tools such as:
- `dot`
- `neato`
- Online renderers and graphing tools

This output can be rendered into **network-style process trees**, allowing investigators to:
- Visualize process spawning chains  
- Detect irregular or unexpected parent-child relationships  
- Identify suspicious execution flows in malware investigations

---

### **Visualization Focus**
The plugin emphasizes **connectivity and hierarchy**, providing a clear visual representation of process structures that supports:
- **Anomaly Detection:** Spot abnormal or hidden parent-child relationships  
- **Malware Analysis:** Map execution chains to trace malicious process origins  
- **System Understanding:** Reveal the operational process tree at the time of memory capture  

---


proccon.py
```diff
@@ -0,0 +1,283 @@
...
+        dot_lines.append("    ")
+        
+        # Define node styles
+        dot_lines.append("    // Process nodes")
+        
+        # Find root processes (no parent or parent doesn't exist)
+        root_pids = []
+        for pid, proc_data in processes.items():
+            ppid = proc_data["ppid"]
+            if ppid == 0 or ppid not in processes:
```
ikelos, line 113: `ppid == 0` can be written in a more pythonic way as `not ppid`, but it's just a stylistic point, not show stopper

proccon.py
```diff
@@ -0,0 +1,283 @@
...
+            safe_path = path.replace('"', '\\"').replace('\\', '\\\\')
+            
+            # Build label with role
+            label = f"[{role}]\\n{safe_name}\\nPID: {pid} | PPID: {ppid}"
+            
+            # Add full executable path
+            if safe_path != safe_name:
+                # Extract just the directory and filename
+                if '\\\\' in safe_path:
+                    parts = safe_path.split('\\\\')
```
ikelos, line 167: Consider using python's path manipulation functions (`pathlib`), rather than manual string manipulation that may not take into account various complexities of paths for various operating systems.

proccon.py
```diff
@@ -0,0 +1,283 @@
...
+            path = proc_data["path"]
+            threads = proc_data["threads"]
+            ppid = proc_data["ppid"]
+            children = proc_data["children"]
+            
+            # Determine process role
+            depth = calculate_depth(pid)
+            if pid in root_pids:
+                role = "ROOT"
+                fillcolor = "lightcoral"
```
ikelos, line 140: Consider extracting the various colours to class variables, so they can be more easily changed and ensure they are consistent throughout the output.

proccon.py
```diff
@@ -0,0 +1,283 @@
...
+        vollog.info("Extracting process information from memory dump...")
+        processes = self._get_process_data(kernel_module_name)
+        
+        vollog.info(f"Found {len(processes)} processes")
+        vollog.info("Generating DOT file...")
+        dot_content = self._generate_dot_content(processes)
+        
+        # Write to file
+        try:
+            with open(output_file, 'w', encoding='utf-8') as f:
```
ikelos, line 247: Please don't directly write out files.  If this is running as part of a webserver, this will write the data out into the web server, rather than being able to return it to the user.  Instead use the built-in `FileHandlerInterface`, by calling `self.open`.  The parameter it takes is a suggested filename, and will work correctly however the library is called.

ikelos (commented):
Generally this looks good, but it writes files directly to the disk of whereever volatility is running, which will not work as expected when volatility is run as a library that's part of a web service, for example.  Please use `self.open` instead of `open`.  It only takes a preferred filename and will expect bytes, but otherwise the output should function similarly.

PR #1909, opened 2025-11-10
AVCHECK Volatility3 Plugin

malfav:
# avcheck.py – Antivirus and Endpoint Detection and Response (EDR) Artifact Scanner (Volatility 3 Plugin)

`avcheck.py` is a specialized **Volatility 3 plugin** designed to detect and enumerate artifacts related to **Antivirus (AV)** and **Endpoint Detection and Response (EDR)** solutions within a **Windows memory dump**.  
It enables investigators and threat hunters to identify which security products were active or installed on a system at the time of capture—critical for understanding a compromised system’s defensive posture.

---

### **Security Software Identification**
Performs broad detection of security tools across **over 80 known AV and EDR products**, including:
- Windows Defender (`MsMpEng.exe`)
- Kaspersky (`avp.exe`)
- Symantec
- AVG
- CrowdStrike (`CsAgent.exe`)
- SentinelOne
- FireEye
- And many others

This identification provides immediate situational awareness of protective technologies present on the analyzed host.

---

### **Multi-Artifact Detection**
Leverages multiple Volatility 3 subsystems to identify security software artifacts across different system components:

- **Processes:**  
  Scans active process lists (`pslist`) for executables associated with AV/EDR agents and monitoring daemons.

- **Services:**  
  Enumerates running services (`svcscan`) to locate entries related to installed security software.

- **DLLs:**  
  Inspects loaded libraries (`dlllist`) to detect known AV/EDR DLLs used for userland hooks or kernel-level integrations.

- **Files:**  
  Uses file scanning (`filescan`) to locate security software directories, databases, log files, and quarantine folders.

Each layer of detection provides corroborating evidence of installed and active endpoint protection components.

---

### **Detection History Triage**
Includes a specialized analysis routine that looks for:
- Quarantine directories and detection logs
- Temporary or deleted AV alert files
- Residual detection history (e.g., Defender history, Kaspersky reports)

These findings offer valuable context on **prior malware detections** or **automated remediation activity** performed by security software.

---

### **Detailed Summary Reporting**
Generates a structured **TreeGrid report** summarizing all detected artifacts, categorized by source type:
- **Processes Detected**
- **Services Detected**
- **DLLs Detected**
- **Files Detected**

At the end of the report, a **summary count** highlights the number of findings per category, providing a concise overview of the security landscape of the system.

Example output:

| Category | Name / Artifact | Path / Details | Source Plugin |
|-----------|----------------|----------------|----------------|
| Process   | MsMpEng.exe    | Windows Defender | pslist |
| Service   | WinDefend      | Microsoft Defender Antivirus Service | svcscan |
| DLL       | mpengine.dll   | AV core scanning module | dlllist |
| File      | DetectionHistory.db | Defender detection logs | filescan |

---


avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+
+vollog = logging.getLogger(__name__)
+
+class AVCheck(interfaces.plugins.PluginInterface):
+    """Detects installed antivirus products and their detection history from memory."""
+    _required_framework_version = (2, 0, 0)
+    _version = (1, 0, 0)
+    
+    # Comprehensive AV process names (80+ processes)
+    AV_PROCESSES = [
```
ikelos, line 16: There are a lot of fields that I'd class as data, and think it would be better to read these from a JSON file, so that the plugin can be updated without having to change the code.  Please could you include these in a JSON file, and load it from the plugin (similar to, for the example, the getsids or getservicesids plugins)...

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+                description="Windows kernel module",
+                architectures=["Intel32", "Intel64"],
+            ),
+        ]
+    
+    def _detect_av_processes(self, kernel_module_name: str) -> List[Dict]:
+        """Detect running antivirus processes."""
+        av_procs = []
+        
+        for proc in pslist.PsList.list_processes(self.context, kernel_module_name):
```
ikelos, line 191: We generally prefer methods to take a `context` parameter as the first parameter, and use that.  It helps make methods more consistent and is particularly useful if the method if found to be useful and other plugins want to use it (and so it would be turned into a classmethod).  It's not a requirement, but it would make future maintenance easier.

I think that, and the AV_PROCESSES attribute, are the only places you make use of self.  It should be fairly easy to transplant AC_PROCESSES whern the time comes, but at the moment there's no need because this isn't a classmethod (yet).

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+                    proc_ppid = proc.InheritedFromUniqueProcessId
+                    
+                    # Try to get full path
+                    try:
+                        peb = proc.get_peb()
+                        if peb:
+                            process_params = peb.ProcessParameters
+                            if process_params:
+                                image_path = process_params.ImagePathName.get_string()
+                                proc_path = image_path if image_path else "N/A"
```
ikelos, line 210: Please don't use the string `N/A` or any other string value.  Other systems (besides the CLI) may need to know that the value isn't available, so please use something derived from `BaseAbsentValue` (such as `NotAvailableValue`).  This means that a Web UI or similar could process the entry knowing that it's distinct from the actual string "N/A".

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+                        else:
+                            proc_path = "N/A"
+                    except Exception:
+                        proc_path = "N/A"
+                    
+                    # Try to get command line
+                    try:
+                        if peb and process_params:
+                            cmdline = process_params.CommandLine.get_string()
+                            cmdline_str = cmdline if cmdline else "N/A"
```
ikelos, line 222: This applies to all fields.

avcheck.py
ikelos: If necessary please return `None` and then when yielding as part of the TreeGrid, use `field or NotAvailableValue()`, just something to differentiate between a string value and something actively not being present.

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+                        'cmdline': cmdline_str
+                    })
+                    
+            except Exception as e:
+                vollog.debug(f"Error processing process: {e}")
+                continue
+        
+        return av_procs
+    
+    def _detect_av_services(self) -> List[Dict]:
```
ikelos, line 242: Again, please take in `context`, `layer_name` and `symbol_table` parameters, rather than using self (see above for the reasoning).

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+                    
+        except Exception as e:
+            vollog.debug(f"Service scan unavailable: {e}")
+        
+        return av_services
+    
+    def _detect_av_dlls(self, kernel_module_name: str) -> List[Dict]:
+        """Detect AV-related DLLs loaded in processes."""
+        av_dlls = []
+        dll_patterns = [
```
ikelos, line 281: These also should be where the others are (in a data file, if not at the top of the plugin).

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+                        errors="replace"
+                    )
+                    
+                    for dll in proc.load_order_modules():
+                        try:
+                            dll_name = dll.BaseDllName.get_string().lower()
+                            dll_path = dll.FullDllName.get_string()
+                            
+                            if any(pattern in dll_name for pattern in dll_patterns):
+                                dll_key = f"{dll_name}|{dll_path}"
```
ikelos, line 305: Why not use a tuple here, rather than combining them together into a string?  It would (theoretically) allow the following two (different) values, the same key:

```
dll_name = "A|B"
dll_path = "C"

dll_name = "A"
dll_path = "B|C"
```
By adding the tuple to the set, you can ensure certainty that the two values are actually equivalent.

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+                            
+                except Exception:
+                    continue
+                    
+        except Exception as e:
+            vollog.debug(f"Error scanning DLLs: {e}")
+        
+        return av_dlls
+    
+    def _detect_av_files(self) -> List[Dict]:
```
ikelos, line 327: Again, ideally this should take a `context` and a `kernel` module as parameters.

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+                    file_name_lower = file_name.lower()
+                    
+                    # Check if file matches AV patterns (stricter filtering)
+                    is_detection_file = any(
+                        pattern.lower() in file_name_lower 
+                        for pattern in self.DETECTION_PATTERNS
+                    )
+                    
+                    # Check for AV vendor directories (stricter)
+                    av_directories = [
```
ikelos, line 348: This also feels like data, and there's no need to redefine it within the loop, so ideally load it from a data JSON file, or put it at the top of the file, or at a bare minimum, make sure it's outside of the loop from an optimization perspective.

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+                        'windows defender', 'kaspersky', 'mcafee', 'norton', 'symantec', 
+                        'bitdefender', 'eset', 'malwarebytes', 'sophos', 'trend micro',
+                        'avast', 'avg', 'avira', 'comodo', 'fsecure', 'panda', 'gdata',
+                        'webroot', 'quarantine', 'virus', 'threat', 'infected'
+                    ]
+                    
+                    is_av_directory = any(av_dir in file_name_lower for av_dir in av_directories)
```
ikelos, line 355: This is very permissive.  Any file that contains any of these values (such as the word virus to threat, in the filename).  Does this return useful results?

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+                    
+                    # Exclude common false positives
+                    false_positives = ['windowsupdate.log', 'iconcache.db', 'thumbs.db',
```
ikelos, line 358: Also should be data, not code.

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+                    is_av_directory = any(av_dir in file_name_lower for av_dir in av_directories)
+                    
+                    # Exclude common false positives
+                    false_positives = ['windowsupdate.log', 'iconcache.db', 'thumbs.db', 
+                                     'prefetch', 'recent', 'cookies']
+                    is_false_positive = any(fp in file_name_lower for fp in false_positives)
+                    
+                    if (is_detection_file or is_av_directory) and not is_false_positive:
+                        # Determine file type with more granularity
+                        if any(p in file_name_lower for p in ['quarantine', 'infected', 'malware']):
```
ikelos, line 364: The number of items here is quite small, so these values can be left in the code (although putting it in JSON would allow it to be more configurable).

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+                        elif 'scan' in file_name_lower:
+                            file_type = "Scan Log"
+                        elif any(p in file_name_lower for p in ['mpenginedb', 'signature', 'definition']):
+                            file_type = "Signature DB"
+                        else:
+                            file_type = "AV Artifact"
+                        
+                        # Extract threat information from filename if present
+                        threat_indicators = []
+                        threat_keywords = ['trojan', 'virus', 'malware', 'ransomware',
```
ikelos, line 377: Some of these words (marking a threat) are also in the list identifying antivirus systems?  This all seems quite inaccurate.  I'd feel more comfortable if this were more specific in identifying antivirus programs, for instance verifying the code signature of the loaded exectuables (or the file paths they're allegedly loaded from).

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+        except Exception as e:
+            vollog.debug(f"Error in filescan: {e}")
+        
+        return av_files
+    
+    def _identify_av_vendor(self, proc_name: str, proc_path: str) -> str:
+        """Identify the AV vendor from process name/path."""
+        proc_lower = (proc_name + ' ' + proc_path).lower()
+        
+        vendor_checks = [
```
ikelos, line 406: Again, another data dump, please put this into a JSON file and load it at the beginning.  It also isn't clear why this is a `tuple[list[str], str]`, rather than just `dict[str, list[str]]`?

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+        av_files = self._detect_av_files()
+        
+        # Get detected vendors
+        detected_vendors = set()
+        if av_procs:
+            for proc in av_procs:
+                vendor = self._identify_av_vendor(proc['name'], proc['path'])
+                detected_vendors.add(vendor)
+        
+        # EVERY yield must have exactly 4 values
```
ikelos, line 456: Please don't include bogus lines for formatting.  If another tool is trying to ask the plugin to produce results, and can determine the types and column names there is *no* way for them to determine that these lines are unnecessary and do not contain valid data.  This is entirely about trying to format the data for a human and completely ignores the use of this plugin by another computer program.

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+        yield (0, ("=" * 80, "", "", ""))
+        yield (0, ("ANTIVIRUS DETECTION RESULTS", "", "", ""))
+        yield (0, ("=" * 80, "", "", ""))
+        yield (0, ("", "", "", ""))
+        
+        # Detected Products
+        if detected_vendors:
```
ikelos, line 463: If you're trying to categorize the results, please include an additional column explaining the type, rather than including "header" values).  This is wrong and will block this plugin from being added to the framework.

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+                    f"Process: {dll['process']} (PID: {dll['pid']})",
+                    f"Base: {dll['base']}",
+                    dll['path']
+                ))
+            if len(av_dlls) > 15:
+                yield (0, (f"  ... and {len(av_dlls) - 15} more DLLs", "", "", ""))
+            yield (0, ("", "", "", ""))
+        
+        # Detection Files (categorized) - ALWAYS SHOW IF THERE ARE FILES
+        if av_files:
```
ikelos, line 513: Almost everything in the following section is the wrong way of yielding data in volatility.  Please provide data and allow the UI (and therefore ultimately the user) the ability to sort and group the results.

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+            quarantine = [f for f in av_files if f['type'] == 'Quarantine']
+            detection = [f for f in av_files if f['type'] == 'Detection Log']
+            scan = [f for f in av_files if f['type'] == 'Scan Log']
+            signature = [f for f in av_files if f['type'] == 'Signature DB']
+            other = [f for f in av_files if f['type'] == 'AV Artifact']
+            
+            if quarantine:
+                yield (0, ("Quarantine Files:", "", "", ""))
+                for f in quarantine:
+                    threat_display = f" [Threats: {f['threats']}]" if f['threats'] != 'N/A' else ""
```
ikelos, line 523: This looks to be grouping individual data items as a single line.  This is not the point of volatility and completely breaks the use of the plugin by anything other than a human reading the CLI output.  Formatting the data before output is also unnecessary and will become problematic when the return from the data may be `None` or a `BaseAbsentValue` derivative.

avcheck.py
```diff
@@ -0,0 +1,606 @@
...
+from volatility3.framework import interfaces, renderers
+from volatility3.framework.configuration import requirements
+from volatility3.framework.renderers import format_hints
+from volatility3.plugins.windows import pslist, filescan, svcscan, dlllist
+
+vollog = logging.getLogger(__name__)
+
+class AVCheck(interfaces.plugins.PluginInterface):
+    """Detects installed antivirus products and their detection history from memory."""
+    _required_framework_version = (2, 0, 0)
```
ikelos, line 12: Unless you're certain this plugin can operate on version 2.0.0 of the framework, please use the current version of the framework (currently `(2, 27, 0)`).

ikelos (changes requested):
Thanks very much for your submission.  Currently this plugin isn't suitable for inclusion to the framework primarily because the output format is abused to produce data that will be easy for humans to use, but which will badly break consumption of the data by another program (at a minimum, specific parsing will be required).

There's also a large number of data items that are included directly in the code, which would be much better if they were loaded from a JSON file, so that the data can be changed without the plugin needing to be updated.

PR #1912, opened 2025-11-11
orphanproc Volatility3 Plugin

malfav:
# OrphanProcs Plugin — Volatility 3

**OrphanProcs** is a custom plugin for the [Volatility 3 memory forensics framework](https://github.com/volatilityfoundation/volatility3).  
Its primary function is to detect and list processes whose parent process (PPID) is no longer active — commonly referred to as *orphan processes*.

---

## Overview

In digital forensics and incident response (DFIR), orphan processes can indicate:
- Malware attempting to hide its original lineage
- Misconfigured or compromised systems
- Suspicious process spawning or persistence mechanisms

The **OrphanProcs** plugin helps identify such anomalies during memory analysis.

---

## How It Works

The plugin:
1. Scans all active processes in the memory image.
2. Compares each process's **Parent Process ID (PPID)** against the list of all active **Process IDs (PIDs)**.
3. Flags any process whose parent is missing or invalid as an **orphan**.

A process is classified as orphaned if:
- Its PPID is **not found** among active PIDs (`MISSING_PARENT`), or  
- Its PPID is **0** (System Idle Process) but the process itself is **not the System process (PID 4)** (`SYSTEM_CHILD`).

---

## Output Fields

| Field | Description |
|-------|--------------|
| **Process** | Executable name (`ImageFileName`) |
| **PID** | Process ID |
| **PPID** | Parent Process ID |
| **Status** | Classification of the orphan process:<br>• `MISSING_PARENT` – Parent terminated<br>• `SYSTEM_CHILD` – Child of System Idle Process (PPID 0)<br>• `ORPHAN` – General orphan classification |
| **Path** | Full path to the executable file |

---

## Optional Arguments

- **`include_system`** — Include the System Idle Process (PID 0) and System process (PID 4) in the results.


ikelos:
Sorry for the mis-marking, I was just trying to mark them as read since they were all submitted at once and the earlier plugins all required similar improvements.

PR #1915, opened 2025-11-11
PSDiff Volatility3 Plugin

malfav:
# ***PSDiff***

***PSDiff***  *a Volatility 3 plugin to compare process instances and detect anomalies.*

This repository contains a ***Volatility 3 plugin*** that scans processes extracted from a Windows memory image, groups processes by name, and compares specific process instances to detect suspicious differences such as differing parent processes, thread-count discrepancies, or session mismatches.

---

## ***Features***

* *Scan all processes* and report process names with multiple instances and differences.
* *Compare two specific processes* (by name or PID).
* *Analyze a single process* (list instances and compare if multiple instances exist).
* *Reports differences in parent process, thread count, and session ID.*

---

## ***Requirements***

* **Python 3.8+**
* **Volatility 3** (tested with Volatility 3.x)
* The plugin expects to be used within the Volatility 3 framework and requires a Windows kernel module when running.

---


ikelos:
Sorry for the mis-marking, I was just trying to mark them as read since they were all submitted at once and the earlier plugins all required similar improvements.

ikelos:
Sorry for the mis-marking, I was just trying to mark them as read since they were all submitted at once and the earlier plugins all required similar improvements.

PR #1917, opened 2025-11-13
vadinfo: Add ability to filter processes by physical memory offset and display physical VAD offsets.

0xHasanM:
This pull request enhances the `vadinfo` plugin in Volatility3 by adding support for filtering processes by physical memory offset and displaying physical offsets in output. It introduces new configuration options, improves process selection flexibility, and ensures offsets are translated correctly when requested.

**New Features:**

* Added `offset` (process physical address offset) and `physical-offsets` (toggle for displaying physical offsets) as new configuration requirements in `vadinfo.py`, allowing users to filter processes based on physical memory offsets and display VADs physical-offsets.

**Process Selection Improvements:**

* Integrated `psscan.PsScan` for process scanning by offset, enabling process selection either by PID (using `pslist`) or by physical offset (using `psscan`), based on user configuration. [[1]](diffhunk://#diff-e4e627a541eac27f4fcc83bf2c9ed1252753ca68ed4bc866292bb5413ca46cdfL12-R12) [[2]](diffhunk://#diff-e4e627a541eac27f4fcc83bf2c9ed1252753ca68ed4bc866292bb5413ca46cdfR303-R320)

**Offset Handling Enhancements:**

* Implemented `_translate_offset` to convert virtual offsets to physical offsets when `physical-offsets` is enabled, ensuring correct offset display and handling page faults gracefully.
* Updated the output to use the translated offset when `physical-offsets` is set, so that the displayed offset matches user expectations.

### DIFF
--- a/volatility3/framework/plugins/windows/vadinfo.py
+++ b/volatility3/framework/plugins/windows/vadinfo.py
@@ -79,6 +79,11 @@ class VadInfo(interfaces.plugins.PluginInterface):
                 default=cls.MAXSIZE_DEFAULT,
                 optional=True,
             ),
+            requirements.IntRequirement(
+                name="offset",
+                description="Process offset in the physical address space",
+                optional=True,
+            ),
         ]
 
     @classmethod
@@ -274,6 +279,24 @@ class VadInfo(interfaces.plugins.PluginInterface):
 
     def run(self) -> renderers.TreeGrid:
         filter_func = pslist.PsList.create_pid_filter(self.config.get("pid", None))
+        kernel = self.context.modules[self.config["kernel"]]
+
+        if self.config["offset"]:
+            procs = psscan.PsScan.scan_processes(
+                self.context,
+                self.config["kernel"],
+                filter_func=psscan.PsScan.create_offset_filter(
+                    self.context,
+                    kernel.layer_name,
+                    self.config["offset"],
+                ),
+            )
+        else:
+             procs = pslist.PsList.list_processes(
+                context=self.context,
+                kernel_module_name=self.config["kernel"],
+                filter_func=filter_func,
+            )
 
         return renderers.TreeGrid(
             [
@@ -290,11 +313,5 @@ class VadInfo(interfaces.plugins.PluginInterface):
                 ("File", str),
                 ("File output", str),
             ],
-            self._generator(
-                pslist.PsList.list_processes(
-                    context=self.context,
-                    kernel_module_name=self.config["kernel"],
-                    filter_func=filter_func,
-                )
-            ),
+            self._generator(procs=procs),
         )
--- a/volatility3/framework/plugins/windows/vadinfo.py
+++ b/volatility3/framework/plugins/windows/vadinfo.py
@@ -84,6 +84,11 @@ class VadInfo(interfaces.plugins.PluginInterface):
                 description="Process offset in the physical address space",
                 optional=True,
             ),
+            requirements.BooleanRequirement(
+                name="physical-offsets",
+                description="List processes with physical offsets instead of virtual offsets.",
+                optional=True,
+            ),
         ]
 
     @classmethod
@@ -203,6 +208,22 @@ class VadInfo(interfaces.plugins.PluginInterface):
 
         return file_handle
 
+    def _translate_offset(self, offset: int) -> int:
+        if not self.config["physical-offsets"]:
+            return offset
+
+        kernel = self.context.modules[self.config["kernel"]]
+        layer_name = kernel.layer_name
+
+        try:
+            _original_offset, _original_length, offset, _length, _layer_name = list(
+                self.context.layers[layer_name].mapping(offset=offset, length=0)
+            )[0]
+        except exceptions.PagedInvalidAddressException:
+            vollog.debug(f"Page fault: unable to translate {offset:0x}")
+
+        return offset
+
     def _generator(self, procs: List[interfaces.objects.ObjectInterface]) -> Generator[
         Tuple[
             int,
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/vadinfo.py
```diff
@@ -252,7 +278,7 @@ def filter_function(x: interfaces.objects.ObjectInterface) -> bool:
                     (
                         proc.UniqueProcessId,
                         process_name,
-                        format_hints.Hex(kernel_layer.canonicalize(vad.vol.offset)),
+                        format_hints.Hex(kernel_layer.canonicalize(self._translate_offset(vad.vol.offset))),
```
ikelos, line 281: This potentially returns values in different layers, but doesn't change the name of the header to indicate which layer the results come from?

volatility3/framework/plugins/windows/vadinfo.py
```diff
@@ -198,6 +208,22 @@ def vad_dump(
...
 
+    def _translate_offset(self, offset: int) -> int:
+        if not self.config["physical-offsets"]:
+            return offset
+
+        kernel = self.context.modules[self.config["kernel"]]
+        layer_name = kernel.layer_name
+
+        try:
+            _original_offset, _original_length, offset, _length, _layer_name = list(
```
ikelos, line 219: The `_layer_name` here is important, since this might point to a swap layer, or if we ever implement it in the future, a compressed memory segment.  The difficulty with "physical layer" is that everyone has an idea of what it means in general, but accurately coping with all the possibilities is complex.

volatility3/framework/plugins/windows/vadinfo.py
```diff
@@ -198,6 +208,22 @@ def vad_dump(
 
         return file_handle
 
+    def _translate_offset(self, offset: int) -> int:
```
ikelos, line 211: This should have a docstring, and when writing that docstring you'll have to explain what it returns, which could be a virtual address, or could be a physical address (ignoring the intricacies of which layer that belongs to)  and that to know which it will return depends on an arbitrary config value.

As well as the docstring, this should be parameterized to accept a flag which determines whether it will simply pass through a virtual offset, or translate it to a physical one.

volatility3/framework/plugins/windows/vadinfo.py
```diff
@@ -79,6 +79,16 @@ def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]
                 default=cls.MAXSIZE_DEFAULT,
                 optional=True,
             ),
+            requirements.IntRequirement(
```
ikelos, line 82: This value both defines an offset that acts like a flag to swap between psscan and pslist as an implmentation.  There is then *also* `physical-offsets` flag which changes only the value displayed but not the scanning method used?  Honestly, this seems quite convoluted and increases the number of options for a use case that doesn't seem very common?  I'm really not a fan of adding a million `--this` and `--that` flags to a plugin to handle every single situation that someone might want.

volatility3/framework/plugins/windows/vadinfo.py
```diff
@@ -274,6 +300,24 @@ def filter_function(x: interfaces.objects.ObjectInterface) -> bool:
 
     def run(self) -> renderers.TreeGrid:
         filter_func = pslist.PsList.create_pid_filter(self.config.get("pid", None))
+        kernel = self.context.modules[self.config["kernel"]]
+
+        if self.config["offset"]:
+            procs = psscan.PsScan.scan_processes(
+                self.context,
+                self.config["kernel"],
```
ikelos, line 308:
```suggestion
                kernel.layer_name,
```
Why do you go back to the config option here, but a couple of lines later use `kernel.layer_name`.  They *should* both be the same thing, but if they're not for some reason, this will cause problems.  Make sure they're both `kernel.layer_name`.

volatility3/framework/plugins/windows/vadinfo.py
```diff
@@ -198,6 +208,22 @@ def vad_dump(
...
+
+        kernel = self.context.modules[self.config["kernel"]]
+        layer_name = kernel.layer_name
+
+        try:
+            _original_offset, _original_length, offset, _length, _layer_name = list(
+                self.context.layers[layer_name].mapping(offset=offset, length=0)
+            )[0]
+        except exceptions.PagedInvalidAddressException:
+            vollog.debug(f"Page fault: unable to translate {offset:0x}")
```
ikelos, line 223: If it can't translate it, it returns a debugging message (which may not be visible to a normal user) and then returns the virtual offset.  This means the output could have physical and virtual offsets mixed together without the user having any way of knowing?  This should at least return a `BaseAbsentValue` like NotAvailable or similar, otherwise the results will be misleading and inaccurate.

ikelos (commented):
There are a number of inaccuracies and issues with the addition of this plugin.  Getting all the details and error conditions correct is difficult, but I'd advise that if this functionality is wanted, it be implemented in a volshell snippet to tinker with and post-process the existing output, rather than alter the main plugin itself.

Below this I'll attach a snippet that should have the same effect as this, although the parameters are defined at the top of the script, rather than requested directly as part of the plugin...

As such I'm afraid we won't be accepting this specific request, but the code you wrote was pretty solid, so if you have other enhancements, do please continue to submit them.  5:)

ikelos:
[vadinfo-physical.py](https://github.com/user-attachments/files/24191163/vadinfo-physical.py)

The above code snippet can be executed using:

```
python volshell.py -f /path/to/image -w --script-only --script /path/to/vadinfo-physical.py
```

As mentioned above, I'm going to close this off because this feature doesn't belong in the main volatility plugin unfortunately.

0xHasanM:
Thanks @ikelos for the suggestion, and comments on the update.

Will keep that in mind for upcoming contributions.

Issue #1875, opened 2025-09-09
Can't parse Win 11 Enterprise 23H2 images


Darth-Leshious:
**Describe the bug**
Volatility will not properly parse a Windows 23H2 image.

**Context**
Volatility Version:  Volatility 3 Framework 2.26.2

Operating System:  Ubuntu 22.04 (SANS SIFT WS)
Python Version: Python 3.10.12
Suspected Operating System:  Microsoft Windows 11 Enterprise (10.0.22631 N/A Build 22631) (23H2)
Command:  'python ./vol.py -f /mnt/hgfs/SecurityOperations/Incidents/03Sep2025_DC4605LP06-CS-alert/complete.dmp windows.pslist.PsList'

**To Reproduce**

1. Use command  'python ./vol.py -vvvvv -f /mnt/hgfs/SecurityOperations/Incidents/03Sep2025_DC4605LP06-CS-alert/complete.dmp windows.pslist.PsList'
3. See error:

INFO     volatility3.cli: Volatility plugins path: ['/usr/local/src/volatility3/volatility3/plugins', '/usr/local/src/volatility3/volatility3/framework/plugins']
INFO     volatility3.cli: Volatility symbols path: ['/usr/local/src/volatility3/volatility3/symbols', '/usr/local/src/volatility3/volatility3/framework/symbols']
DEBUG    volatility3.plugins.yarascan: Using yara-python module
DETAIL 3 volatility3.cli: Cache directory used: /home/sansforensics/.cache/volatility3
INFO     volatility3.framework.automagic: Detected a windows category plugin
INFO     volatility3.framework.automagic: Running automagic: ConstructionMagic
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.PsList.kernel
[...]

PR #1921, opened 2025-12-01
Improve Intel layer's address space scanning

### LINKED ISSUE #1920: Incredibly long `windows.memmap.Memmap` processing and 100% CPU load.
**Describe the bug**
Several process with `windows.memmap.Memmap` hangs with 100% CPU and never produces any rows for some specific processes in a Windows 10 x64 memory dump. The plugin:

- initializes successfully (layers + symbols are found),
- prints the table header (`Virtual Physical Size Offset in File File output`),
- then stops producing any output at all,

while the `vol-3.exe` process continues to consume CPU indefinitely (2-3 days) with almost no disk I/O.

The same memory image and the same PIDs work fine with other plugins such as `windows.vadinfo.VadInfo`. `windows.memmap.Memmap` also works correctly for many other processes in the same dump. The issue affects a subset of processes, mostly core system services (`svchost.exe`, `sihost.exe`, `wslservice.exe`, `RtkAudUService`).

I cannot share the memory image, but the behavior is fully reproducible on my dump.

**Context**
Volatility Version: 2.26.2 (PyInstaller binary `vol-3.exe` built from tag 2.26.2)  
Operating System: Windows 10. Version: 22H2. OS build: 19045.6456
Python Version: I use the compiled exe-application via vol.spec.
Suspected Operating System: the same (Windows 10. Version: 22H2. OS build: 19045.6456)

Command: 
Problematic example (one of several PIDs):
`vol-3.exe --offline -vvv --log log4860.txt -f 20251119.mem windows.memmap.Memmap --pid 4860`

Working example (many PIDs in the same image):
`vol-3.exe --offline -vvv --log log6324.txt -f 20251119.mem windows.memmap.Memmap --pid 6324`

The physical layer size (from the logs) is:
`physical_layer maximum_address: 36507222015`

**To Reproduce**
(High level, since I cannot share the image)

1. Use PyInstaller and generate `vol-3.exe` 2.26.2. Windows x64 RAM image (in my case ~34 GiB, taken from a desktop system with lots of `svchost.exe` instances, there were almost no user applications running).
2. Run `windows.pslist.PsList -r json` on the image and pick system processes
3. For any of these PIDs, run: `vol-3.exe --offline -vvv --log log_<pid>.txt -f 20251119.mem windows.memmap.Memmap --pid <pid>`
4. Observe that:
* `vol-3.exe` prints the standard log, then the memmap table header, then stops printing any rows.
* CPU usage for that `vol-3.exe` stays high (several volatility processes use processor time equally, but together they take up almost 100% of the CPU, see screenshot) for a very long time (I left it running for many hours and even days).

Image

* Disk / I/O usage is close to zero.
* The process exits after 3-4 days
* You can cancel the process in the console by pressing Ctrl+C, then you will receive the following log:
```
  File "volatility3\cli\__init__.py", line 927, in main
  File "volatility3\cli\__init__.py", line 515, in run
  File "volatility3\cli\text_renderer.py", line 330, in render
  File "volatility3\framework\renderers\__init__.py", line 317, in populate
  File "volatility3\framework\plugins\windows\memmap.py", line 67, in _generator
  File "volatility3\framework\layers\intel.py", line 307, in mapping
  File "volatility3\framework\layers\intel.py", line 363, in _mapping
  File "volatility3\framework\layers\intel.py", line 515, in _translate
  File "volatility3\framework\layers\intel.py", line 462, in _translate_swap
  File "volatility3\framework\layers\intel.py", line 162, in _translate
  File "volatility3\framework\layers\intel.py", line 242, in _translate_entry
  File "volatility3\framework\layers\intel.py", line 270, in _get_valid_table
  File "volatility3\framework\interfaces\layers.py", line 635, in read
  File "volatility3\framework\layers\physical.py", line 161, in read
  File "volatility3\framework\exceptions.py", line 52, in __init__
KeyboardInterrupt
[PYI-15492:ERROR] Failed to execute script 'vol' due to unhandled exception
```

5. For the **_same_** PIDs, run:
`vol-3.exe --offline -vvv -f 20251119.mem windows.vadinfo.VadInfo --pid <pid>`
and observe that `VadInfo` completes successfully and relatively quickly.
6. For other PIDs in the same image, `windows.memmap.Memmap --pid <pid>` completes correctly and produces output as expected.

**Expected behavior**
For all valid PIDs corresponding to real processes:
* `windows.memmap.Memmap` should either:
      * complete in a reasonable amount of time and emit rows describing the virtual→physical mapping for the process, or
      * fail with a clear error if the process layer cannot be constructed or the mapping cannot be resolved.

Hanging indefinitely with 100% CPU and no further output after the header is unexpected.

**Example output**
```
.\vol-3.exe --offline -vvv --log logShittyPid.txt -f '20251119.mem' windows.memmap.Memmap --pid 5280
Volatility 3 Framework 2.26.2
INFO     volatility3.cli: Volatility plugins path: ['C:\\me\\Repositories\\vol\\2-26-2\\plugins', 'C:\\Users\\me\\AppData\\Local\\Temp\\_MEI139162\\volatility3\\plugins', 'C:\\Users\\me\\AppData\\Local\\Temp\\_MEI139162\\volatility3\\framework\\plugins']
INFO     volatility3.cli: Volatility symbols path: ['C:\\me\\Repositories\\vol\\2-26-2\\symbols', 'C:\\Users\\me\\AppData\\Local\\Temp\\_MEI139162\\volatility3\\symbols', 'C:\\Users\\me\\AppData\\Local\\Temp\\_MEI139162\\volatility3\\framework\\symbols']
DEBUG    volatility3.plugins.yarascan: Using yara-python module
INFO     volatility3.framework.automagic: Detected a windows category plugin
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Memmap.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Memmap.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Memmap.kernel
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Memmap.kernel.layer_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Memmap.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Memmap.kernel.layer_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Memmap.kernel
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Memmap.kernel
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Memmap.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Memmap
INFO     volatility3.framework.automagic: Running automagic: SymbolCacheMagic
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Memmap.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DEBUG    volatility3.framework.automagic.windows: Detecting Self-referential pointer for recent windows
DEBUG    volatility3.framework.automagic.windows: DtbSelfRef64bit test succeeded at 0x1ad000
DEBUG    volatility3.framework.automagic.windows: DTB was found at: 0x1ad000
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Memmap.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Memmap.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Memmap.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Memmap.kernel
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Memmap.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Memmap.kernel.layer_name.memory_layer
DETAIL 1 volatility3.framework.interfaces.configuration: TypeError - kernel_virtual_offset requirements only accept int type: None
DETAIL 1 volatility3.framework.interfaces.configuration: TypeError - kernel_virtual_offset requirements only accept int type: None
DETAIL 1 volatility3.framework.interfaces.configuration: TypeError - kernel_banner requirements only accept str type: None
DETAIL 1 volatility3.framework.interfaces.configuration: TypeError - kernel_banner requirements only accept str type: None
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Memmap.kernel
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Memmap
DEBUG    volatility3.framework.automagic.stacker: physical_layer maximum_address: 36507222015
DEBUG    volatility3.framework.automagic.stacker: Stacked layers: ['IntelLayer', 'FileLayer']
INFO     volatility3.framework.automagic: Running automagic: WinSwapLayers
INFO     volatility3.framework.automagic: Running automagic: KernelPDBScanner
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Memmap.kernel.symbol_table_name
DEBUG    volatility3.framework.automagic.pdbscan: Setting kernel_virtual_offset to 0xf8035dc00000
DEBUG    volatility3.framework.symbols.windows.pdbutil: Using symbol library: ntkrnlmp.pdb\F57E740B088E5056E8AF0772F1CC5BEB-1
DEBUG    volatility3.cli: Successfully constructed windows.memmap.Memmap (0, 0, 0)

Virtual Physical        Size    Offset in File  File output
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_PO_PROCESS_ENERGY_CONTEXT
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_EPROCESS_QUOTA_BLOCK
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_PAGEFAULT_HISTORY
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_JOB_ACCESS_STATE
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_JOB_CPU_RATE_CONTROL
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_JOB_NET_RATE_CONTROL
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_JOB_NOTIFICATION_INFORMATION
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_PSP_STORAGE
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_ACTIVATION_CONTEXT_DATA
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_ASSEMBLY_STORAGE_MAP
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_EXP_LICENSE_STATE
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_DBGKP_ERROR_PORT
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_CI_NGEN_PATHS
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_EX_WNF_SUBSCRIPTION
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_ETW_EVENT_CALLBACK_CONTEXT
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_ETW_SOFT_RESTART_CONTEXT
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_ETW_STACK_CACHE
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_ETW_PERFECT_HASH_FUNCTION
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_EX_TIMER
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_HAL_PMC_COUNTERS
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_DEVICE_NODE_IOMMU_EXTENSION
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_SCSI_REQUEST_BLOCK
```

At this point no rows are ever printed under the header. The process keeps running and consuming CPU.

Good PID example (--pid 6324) in the same image (the logs are identical before the results are displayed):
```
Volatility 3 Framework 2.26.2
INFO     volatility3.cli: Volatility plugins path: [...]
INFO     volatility3.framework.automagic: Detected a windows category plugin
...
DEBUG    volatility3.cli: Successfully constructed windows.memmap.Memmap (0, 0, 0)
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_PO_PROCESS_ENERGY_CONTEXT
DEBUG    volatility3.framework.symbols: Unresolved reference: symbol_table_name1!_EPROCESS_QUOTA_BLOCK
...
```

For good PID, rows with Virtual, Physical, Size, etc. are then printed as expected and the command terminates normally.

Also notice, for the problematic PIDs, the following runs do complete successfully:
`vol-3.exe --offline -vvv -f 20251119.mem windows.vadinfo.VadInfo --pid <Probleamtic PID>`

**Additional information**
* I ran the above `vol-3.exe` commands directly in a console.
* CPU is fully utilized by `vol-3.exe` (several processes), disk I/O is minimal, and the process terminate after several days.
* For every problematic process for which `Memmap` does not work, I can find information about that process through `PsList`.
* One more screenshot (each process runs on a unique pid):

Image

* Using `Process Explorer` to inspect the hot thread’s stack while it is hung, I see it looping inside Python (e.g. `python38.dll!PyEval_EvalFrameDefault` / `PyFunction_Vectorcall`) when the memmap plugin is running for these PIDs.
* Also attached stack for threads (10 files for 10 processes)

[10.txt](https://github.com/user-attachments/files/23727995/10.txt)
[6.txt](https://github.com/user-attachments/files/23727994/6.txt)
[9.txt](https://github.com/user-attachments/files/23728000/9.txt)
[8.txt](https://github.com/user-attachments/files/23727998/8.txt)
[7.txt](https://github.com/user-attachments/files/23727997/7.txt)
[5.txt](https://github.com/user-attachments/files/23727996/5.txt)
[4.txt](https://github.com/user-attachments/files/23728001/4.txt)
[3.txt](https://github.com/user-attachments/files/23727999/3.txt)
[2.txt](https://github.com/user-attachments/files/23728002/2.txt)
[1.txt](https://github.com/user-attachments/files/23728003/1.txt)

If there is any additional logging or instrumentation you’d like me to add to a local build to help diagnose this, I can try that on my side, but I’m not able to share the memory image itself.

Regards,
Artyom

Abyss-W4tcher:
During #1920 investigation, I've noticed that the Intel layer was failing to efficiently skip large invalid blocks of memory while mapping an entire address space.

This PR makes the following changes:

- Remove the "large_page" attribute on PTEs, which was previously needed because the large page check was done in the wrong order. Now the check is performed under the correct `self._structure` member parameters.
- Decrease the page block size **after** the duplicated page table check to skip the correct amount of entries.
- Catch `InvalidAddressException` in `_get_valid_table` so that we can raise a `PagedInvalidAddressException`, allowing `_mapping` to know where the fault occured.
- Rely on virtual offset for `chunk_size` calculation:

In #1920 [I've noticed the following](https://github.com/volatilityfoundation/volatility3/issues/1920#issuecomment-3589663521):

Virtual: `0xb0240000000`, Physical: `0x1fffffffe000`, Page size: `0x40000000` (large), Chunk Size: `0x40000000 - (0x1fffffffe000 % 0x40000000) = 0x2000`

This physical address most likely corresponds to a bogus Windows regions and impacts the `chunk_size` calculation (0x2000 vs 0x40000000). I've updated the calculation to use the virtual offset instead:

Virtual: `0xb0240000000`, Physical: `0x1fffffffe000`, Page size: `0x40000000` (large), Chunk Size: `0x40000000 - (0xb0240000000 % 0x40000000) = 0x40000000`

This saves 0x40000000 / 0x2000 = 131072 iterations for this large page.

---

Benchmarks:

```sh
# Before
$ time python3 vol.py -f win-xp-laptop-2005-06-25.img windows.memmap &> /dev/null                                  
python3 vol.py -f ../test_images/win-xp-laptop-2005-06-25.img windows.memmap   82.34s user 0.67s system 99% cpu 1:23.58 total
# After
$ time python3 vol.py -f win-xp-laptop-2005-06-25.img windows.memmap &> /dev/null                                  
python3 vol.py -f ../test_images/win-xp-laptop-2005-06-25.img windows.memmap   39.37s user 0.47s system 98% cpu 40.409 total
```

Docs: https://wiki.osdev.org/Paging#:~:text=jne%20.fill_table-,64%2DBit%20Paging,-Page%20map%20table

Fixes #1920.

### DIFF
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -376,6 +376,7 @@ class Intel(linear.LinearlyMappedLayer):
             yield offset, length, mapped_offset, length, layer_name
             return None
         while length > 0:
+            skip_mask = None
             try:
                 chunk_offset, page_size, layer_name = self._translate(offset)
                 # Page align the chunk size value
@@ -386,11 +387,9 @@ class Intel(linear.LinearlyMappedLayer):
                     # Virtual -> physical is contiguous in the chunk_size range.
                     # If we fail, we can jump directly to the end as we know all bytes in between
                     # aren't mapped (virtually and) physically anyway.
-                    raise exceptions.PagedInvalidAddressException(
-                        layer_name=layer_name,
-                        invalid_address=chunk_offset,
-                        entry=0,
-                        invalid_bits=int(math.log2(chunk_size)),
+                    skip_mask = chunk_size - 1
+                    raise exceptions.InvalidAddressException(
+                        layer_name=layer_name, invalid_address=chunk_offset
                     )
             except (
                 exceptions.PagedInvalidAddressException,
@@ -398,12 +397,13 @@ class Intel(linear.LinearlyMappedLayer):
             ) as excp:
                 if not ignore_errors:
                     raise
-                # We can jump more if we know where the page fault failed
-                if isinstance(excp, exceptions.PagedInvalidAddressException):
-                    mask = (1 << excp.invalid_bits) - 1
-                else:
-                    mask = (1 << self._page_size_in_bits) - 1
-                length_diff = mask + 1 - (offset & mask)
+                if skip_mask is None:
+                    # We can jump more if we know where the page fault occured
+                    if isinstance(excp, exceptions.PagedInvalidAddressException):
+                        skip_mask = (1 << excp.invalid_bits) - 1
+                    else:
+                        skip_mask = (1 << self._page_size_in_bits) - 1
+                length_diff = skip_mask + 1 - (offset & skip_mask)
                 length -= length_diff
                 offset += length_diff
             else:
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -38,7 +38,7 @@ class Intel(linear.LinearlyMappedLayer):
     # NOTE: _maxphyaddr is MAXPHYADDR as defined in the Intel specs *NOT* the maximum physical address
     _maxphyaddr = 32
     _maxvirtaddr = _maxphyaddr
-    _structure = [("page directory", 10, False), ("page table", 10, True)]
+    _structure = [("page directory", 10, True), ("page table", 10, False)]
     _direct_metadata = collections.ChainMap(
         {"architecture": "Intel32"},
         {"mapped": True},
@@ -221,18 +221,6 @@ class Intel(linear.LinearlyMappedLayer):
                     entry,
                     "Page Fault at entry " + hex(entry) + " in table " + name,
                 )
-            # Check if we're a large page
-            if large_page and (entry & self._PAGE_PSE):
-                # Mask off the PAT bit
-                if entry & self._PAGE_PAT_LARGE:
-                    entry -= self._PAGE_PAT_LARGE
-                # We're a large page, the rest is finished below
-                # If we want to implement PSE-36, it would need to be done here
-                break
-            # Figure out how much of the offset we should be using
-            start = position
-            position -= size
-            index = self._mask(page_address, start, position + 1) >> (position + 1)
 
             # Grab the base address of the table we'll be getting the next entry from
             base_address = self._mask(
@@ -249,6 +237,11 @@ class Intel(linear.LinearlyMappedLayer):
                     "Page Fault at entry " + hex(entry) + " in table " + name,
                 )
 
+            # Figure out how much of the offset we should be using
+            start = position
+            position -= size
+            index = self._mask(page_address, start, position + 1) >> (position + 1)
+
             # Read the data for the next entry
             entry_data_start = index << self._index_shift
             entry_data = table[entry_data_start : entry_data_start + self._entry_size]
@@ -261,6 +254,15 @@ class Intel(linear.LinearlyMappedLayer):
 
             # Read out the new entry from memory
             (entry,) = struct.unpack(self._entry_format, entry_data)
+
+            # Check if we're a large page
+            if large_page and (entry & self._PAGE_PSE):
+                # Mask off the PAT bit
+                if entry & self._PAGE_PAT_LARGE:
+                    entry -= self._PAGE_PAT_LARGE
+                # We're a large page, the rest is finished below
+                # If we want to implement PSE-36, it would need to be done here
+                break
 
         return entry, position
 
@@ -429,7 +431,7 @@ class IntelPAE(Intel):
     _structure = [
         ("page directory pointer", 2, False),
         ("page directory", 9, True),
-        ("page table", 9, True),
+        ("page table", 9, False),
     ]
     _direct_metadata = collections.ChainMap({"pae": True}, Intel._direct_metadata)
 
@@ -449,7 +451,7 @@ class Intel32e(Intel):
         ("page map layer 4", 9, False),
         ("page directory pointer", 9, True),
         ("page directory", 9, True),
-        ("page table", 9, True),
+        ("page table", 9, False),
     ]
 
 
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -269,9 +269,12 @@ class Intel(linear.LinearlyMappedLayer):
     @functools.lru_cache(maxsize=1025)
     def _get_valid_table(self, base_address: int) -> Optional[bytes]:
         """Extracts the table, validates it and returns it if it's valid."""
-        table = self._context.layers.read(
-            self._base_layer, base_address, self.page_size
-        )
+        try:
+            table = self._context.layers.read(
+                self._base_layer, base_address, self.page_size
+            )
+        except exceptions.InvalidAddressException:
+            return None
 
         # If the table is entirely duplicates, then mark the whole table as bad
         if table == table[: self._entry_size] * self._entry_number:
@@ -375,12 +378,19 @@ class Intel(linear.LinearlyMappedLayer):
         while length > 0:
             try:
                 chunk_offset, page_size, layer_name = self._translate(offset)
-                chunk_size = min(page_size - (chunk_offset % page_size), length)
+                # Page align the chunk size value
+                chunk_size = min(page_size - (offset % page_size), length)
                 if not self._context.layers[layer_name].is_valid(
                     chunk_offset, chunk_size
                 ):
-                    raise exceptions.InvalidAddressException(
-                        layer_name=layer_name, invalid_address=chunk_offset
+                    # Virtual -> physical is contiguous in the chunk_size range.
+                    # If we fail, we can jump directly to the end as we know all bytes in between
+                    # aren't mapped (virtually and) physically anyway.
+                    raise exceptions.PagedInvalidAddressException(
+                        layer_name=layer_name,
+                        invalid_address=chunk_offset,
+                        entry=0,
+                        invalid_bits=int(math.log2(chunk_size)),
                     )
             except (
                 exceptions.PagedInvalidAddressException,
--- a/volatility3/framework/constants/_version.py
+++ b/volatility3/framework/constants/_version.py
@@ -1,7 +1,7 @@
 # We use the SemVer 2.0.0 versioning scheme
 VERSION_MAJOR = 2  # Number of releases of the library with a breaking change
 VERSION_MINOR = 27  # Number of changes that only add to the interface
-VERSION_PATCH = 0  # Number of changes that do not change the interface
+VERSION_PATCH = 1  # Number of changes that do not change the interface
 VERSION_SUFFIX = ""
 
 PACKAGE_VERSION = (


volatility3/framework/layers/intel.py
```diff
@@ -373,12 +378,19 @@ def _mapping(
...
                 ):
-                    raise exceptions.InvalidAddressException(
-                        layer_name=layer_name, invalid_address=chunk_offset
+                    # Virtual -> physical is contiguous in the chunk_size range.
+                    # If we fail, we can jump directly to the end as we know all bytes in between
+                    # aren't mapped (virtually and) physically anyway.
+                    raise exceptions.PagedInvalidAddressException(
+                        layer_name=layer_name,
+                        invalid_address=chunk_offset,
+                        entry=0,
```
ikelos: Ok, so... I kinda like this and I kinda hate it.  I like it because it's a smart way of using what we already have to bypass further scanning.  I don't like it, because it essentially lies about the entry value, which is bad (and sets a bad precedent).  I haven't quite worked through my thoughts entirely, but given that if `ignore_errors` is true, this might make it outside of the Intel layer, I'd really like to try and ensure the exception is valid.

The reason we seem to be reconstructing this error, is because `is_valid` swallows whatever actual error we might have expected, so I'm wondering if we shouldn't parameterize `is_valid` with a `throw_exception` flag, that does everything `is_valid` does now, but only swallows the exception (and produces `False`) when the `throw_exception` flag is set to false, and otherwise it surfaces exceptions (which is what we want in this case)?  Anything else would be inefficient (we'd be walking the tree twice) or misleading (we'd be making up a bad exception).

This all hinges on the `is_valid` swallowing a correct `PagedInvalidAddressException`, but I'd tend to trust that's the case, so if we can validate that and it's the case then that's how I'd prefer to proceed.

ikelos (commented):
Wow, really good catch!  I'm surprised that's been lurking in there the whole time!  I've run through it all and double checked the paging entries (you're right it looks like we were verifying one entry too early), so I think those changes are good.  My big issue is with that fixed `0` in the `PagedInvalidAddressException` although I think I have a reasonable way of dealing with it, please see the comment below.

Once we can validate that, and depending on the validation either tweak `is_valid` or find some other way to return something more accurate from `PagedInvalidAddressException`, then I think this is good to go in.

This is core code for volatility, so messing with it can often have unintended consequences, but thank you for braving it.  I think in this instance you've spotted a specific issue, and I appreciate you tinkering just enough to resolve that one problem, and based on the specs too and not going wild with changes.  Thanks...  5:)

ikelos:
Also, at a minimum it needs to bump the framework version.  Since it *should* be an invisible change, I'd just bump the PATCH version.

volatility3/framework/layers/intel.py
Abyss-W4tcher: Ok this makes sense, I've reverted to the previous `InvalidAddressException` version and added a dedicated variable that stores the size of the fault directly.

Also `is_valid` raises an `InvalidAddressException` because it doesn't know anything about page tables.

ikelos (approved):
Thanks, I prefer the manual skipping to trying to shove the information into an exception to transmit it.  This looks good.  I'm *just* gonna cut a release branch before this, so this won't make it into the next release given it hasn't had a lot of testing...  5;)

volatility3/framework/layers/intel.py
ikelos: Well, the point is it waits for an exception and then fails, so we could just raise if we want exceptions, something like:

```patch
@@ -276,7 +276,7 @@ class Intel(linear.LinearlyMappedLayer):
             return None
         return table
 
-    def is_valid(self, offset: int, length: int = 1) -> bool:
+    def is_valid(self, offset: int, length: int = 1, throw_errors = False) -> bool:
         """Returns whether the address offset can be translated to a valid
         address."""
         try:
@@ -286,6 +286,8 @@ class Intel(linear.LinearlyMappedLayer):
                 for _, _, mapped_offset, _, layer in self.mapping(offset, length)
             )
+            if throw_errors:
+                raise
             return False
 
     def is_dirty(self, offset: int) -> bool:
```

The thing I haven't figured out is whether that might only skip us so far ahead in the mapping (due to a physical invalid address exception), but I think that would still scan physical pages even if there's a chunk missing (for whatever reason)?

Abyss-W4tcher: `InvalidAddressException` could indeed have an optional `invalid_size` parameter added (similarly to `PagedInvalidAddressException`).

I think the `is_valid` method used is the one of the underlying layer (LiME, raw etc.) not the Intel one (that would make an infinite loop?). I'm not even sure that it can raise an `InvalidAddressException` (the try/except block you see in Intel's `is_valid` could be superfluous):

https://github.com/volatilityfoundation/volatility3/blob/493950c6346da816ef43a48afdf401cb326dc5de/volatility3/framework/layers/physical.py#L146-L153

Abyss-W4tcher: I'm not sure if it is worth updating all the `is_valid` methods for this one use case.

We can still remove the `skip_mask` variable by specifying it directly in a new `invalid_size` parameter when raising `exceptions.InvalidAddressException` in _mapping.

ikelos: You wouldn't need to update all of them, if the default value is false, then you just change the one place you want to use it to pass in True, and then underlying exception should fall through?

ikelos: Oh, I guess you meant through all the inheritance tree, yeah, fair enough.

ArtyomAfanasovBS:
@Abyss-W4tcher, @ikelos, hello! Are these changes not included in version 2.27.0?

ikelos:
They were not unfortunately.  We cut the release branch early followed by various stages of code-freeze to allow time for testing and stabilization.  In this instance, the release branch was cut on November 30th and these didn't get merged into develop until just after (December 3rd).  They will be in the next release though, which should be due in April.  Sorry for the inconvenience.

lcfut:
I've been experiencing the same issues as #1875 and came here as per the comments.  I have not tried anything on this thread yet against my RAW file.
The error message I am getting is the same.  

Unsatisfied requirement plugins.Info.kernel.symbol_table_name:

I've pulled the PDB file, updated cache, manually created the symbols JSON file and placed it into the correct folder structure of PDBName + GUID + Age
I've also confirmed the ISF is present:
b'ntkrnlmp.pdb|AF550CAA73AFB287705CC40079D786B4|1':
  file:///C:/vol3/venv/.../symbols/windows/ntoskrnl.pdb/AF55...B41/ntkrnlmp.json
  and
  file:///C:/vol3/venv/.../symbols/windows/ntkrnlmp.pdb/AF55...B41/ntkrnlmp.json
  
  Do I just need to sit tight and wait for the April release or is there anything else I can do??

PR #1927, opened 2026-01-04
Only act on local cache symbols under the symbol basepaths

### LINKED ISSUE #1255: Cache Update Issues

ikelos:
Fixes #1255

### DIFF
--- a/volatility3/framework/automagic/symbol_cache.py
+++ b/volatility3/framework/automagic/symbol_cache.py
@@ -239,7 +239,16 @@ class SqliteCache(CacheManagerInterface):
         results = self._database.cursor().execute(statement, parameters).fetchall()
         result = None
         for row in results:
-            result = row["location"]
+            local_filepath = self._get_local_filepath(row["location"])
+            if not (
+                local_filepath is None
+                or local_filepath.startswith(tuple(constants.SYMBOL_BASEPATHS))
+            ):
+                vollog.debug(
+                    f"Location {row['location']} found but outside of the registered symbol paths"
+                )
+            else:
+                result = row["location"]
         return result
 
     def get_local_locations(self) -> Generator[str, None, None]:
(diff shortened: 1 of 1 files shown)


ikelos:
Supercedes #1751

volatility3/framework/automagic/symbol_cache.py
```diff
@@ -249,7 +258,11 @@ def get_local_locations(self) -> Generator[str, None, None]:
             .fetchall()
         )
         for row in result:
-            yield row["location"]
+            local_filepath = self._get_local_filepath(row["location"])
+            if local_filepath and local_filepath.startswith(
+                tuple(constants.SYMBOL_BASEPATHS)
+            ):
+                yield row["location"]
```
ikelos, line 265: Needs an else with a debug message saying it was found but outside the appropriate path.

ikelos:
This is problematic because some local JSON files (such as those under the volatility directory) want to be loaded even though they're not under the symbol directories...

PR #1932, opened 2026-01-14
Fix format_hints.MultiTypeData representation in Parquet/Arrow renderers

### LINKED ISSUE #1931: Arrow and Parquet renderers fails on `format_hints.MultiTypeData` with binary content
The `format_hints.MultiTypeData` type is currently represented as the `pa.utf8` type in the Arrow renderer, but if binary data is present it will throw an exception, e.g.,

```plain
vol -c ../volatility_memory_forensics_course/jackal.json -r parquet printkey --offset 0x880187acb000 --recurse

<snip>

pyarrow.lib.ArrowInvalid: Could not convert b'\xf4\x00\x00\x00\x1bM@\xf3\xb7]\xd5\x01' with type MultiTypeData: was not a utf8 string
```

Two potential fixes:

1. Parse `format_hints.MultiTypeData` like the JSON renderer does (see volatility3/cli/text_renderer.py)
2. Use pa.binary instead of pa.utf8 to preserve binary data as-is and the user handle it downstream

Personally, I think option 2 is the right approach.

kyrre:
This maps `format_hints.MultiTypeData` to `pa.binary` and fixes #1931

Changes `pa.utf8` → `pa.binary` to handle binary content in registry values and other `format_hints.MultiTypeData` fields.

### DIFF
--- a/volatility3/framework/plugins/renderers/parquet_renderer.py
+++ b/volatility3/framework/plugins/renderers/parquet_renderer.py
@@ -52,7 +52,7 @@ class ArrowRenderer(text_renderer.CLIRenderer):
             datetime.datetime: lambda: pa.timestamp("ms"),
             format_hints.Bin: pa.uint64,
             format_hints.Hex: pa.uint64,
-            format_hints.MultiTypeData: pa.utf8,
+            format_hints.MultiTypeData: pa.binary,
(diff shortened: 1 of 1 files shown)


ikelos:
Looks fine, thanks for the fix.  5:)

PR #1933, opened 2026-01-21
Add kerberos plugin

citronneur:
Hello Volatility team,

We are please to propose you a plugin that is able to list Kerberos sessions and parse and dump Kerberos ticket from a windows memory dump. 

This plugin will facilitate the way track golden/silver ticket attempt, user tracking by listing service ticket in cache.

Thanks to consider my contribution.

Have a nice day

### DIFF
--- a/volatility3/framework/plugins/windows/kerberos.py
+++ b/volatility3/framework/plugins/windows/kerberos.py
@@ -1,0 +1,996 @@
+"""
+Klist plugin made with love by the Airbus CERT team
+
+The plugin intend to:
+    - Sessions      : list sessions managed by the kerberos security provider
+    - Tickets       : list ticket in cache
+    - VadTicketScan : Carve tickets from lsass process Vad
+"""
+import logging
+import datetime
+from typing import Iterable, Tuple, List, Type, Callable
+
+from volatility3.framework import interfaces, symbols, exceptions, constants, objects
+from volatility3.framework import renderers
+from volatility3.framework.configuration import requirements
+from volatility3.framework.layers import scanners
+from volatility3.framework.objects import utility
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.symbols import intermed
+from volatility3.framework.symbols.windows import pdbutil
+from volatility3.framework.symbols.windows.extensions import conversion
+from volatility3.plugins.windows import pslist, vadinfo, pe_symbols
+
+vollog = logging.getLogger(__name__)    
+
+def UnwrapMemoryOr(default):
+    """Wrapper that will check if the function could access to an invalid memory space"""
+    def _safe_memory_wrapper_1(function):
+        def _safe_memory_wrapper_2(*args, **kwargs):
+            try:
+                return function(*args, **kwargs)
+            except exceptions.InvalidAddressException as excp:
+                vollog.debug(
+                    f"Invalid address {hex(excp.invalid_address)} in layer {excp.layer_name} from {function.__qualname__}"
+                )
+                return default
+        
+        return _safe_memory_wrapper_2
+    return _safe_memory_wrapper_1
+
+def _layout_scanner(
+        source: interfaces.objects.ObjectInterface, 
+        object_type: str, 
+        condition: Callable, 
+        start=0x0,
+        increment=0x8, 
+        max_offset=512
+    ):
+    """Will scan an object layout memory by casting it into object_type and respect condition
+        source: object which want to test the layout
+        condition: function call to to know if object was found
+        
+        :return: list of offset from source where condition is valid
+    """
+    result = []
+    # align for 64 bits
+    offset = start
+    while offset < max_offset:
+        obj = source._context.object(
+            layer_name=source.vol.layer_name,
+            object_type=object_type, 
+            offset=source.vol.offset + offset
+        )
+        
+        if condition(obj):
+            result.append(offset)
+            
+        offset += increment
+    return result
+
+def _scan_for_list_header(
+        obj: interfaces.objects.ObjectInterface, 
+        max_offset=512, 
+        max_element=50,
+        inc=0x8
+    ):
+    """Will test the LIST_ENTRY by looping over all element and test if it come back to the original
+    
+    """
+    result = []
+    # align for 64 bits
+    offset = 0
+    while (offset < max_offset and inc > 0) or (offset > max_offset and inc < 0):
+        list_head = obj._context.object(
+            layer_name=obj.vol.layer_name,
+            object_type=obj._context.modules['kernel'].symbol_table_name+constants.BANG+'_LIST_ENTRY', 
+            offset=obj.vol.offset + offset
+        )
+        
+        try:
+            index_element = 0
+            element = list_head
+            while element.Flink != list_head.vol.offset:
+                element = element.Flink
+                index_element += 1
+                if index_element > max_element:
+                    raise exceptions.VolatilityException("Max element reach before looping")
+                
+            #find
+            result.append(offset)
+                
+        except:
+            """The list does not loop on self element so not a list"""
+            
+        offset += inc
+    return result
+
+def _scan_for_unicode_string(obj: interfaces.objects.ObjectInterface):
+    """Search for unicode string where pattern is MaxLength = Length + 2 (null terminator)"""
+    return _layout_scanner(
+        source=obj, 
+        object_type=obj._context.modules['kernel'].symbol_table_name+constants.BANG+'_UNICODE_STRING',
+        condition=lambda s: s.Length + 2 == s.MaximumLength and obj._context.layers[obj.vol.layer_name].is_valid(s.Buffer, s.MaximumLength)
+    )
+    
+def _scan_for_timestamp(obj, max_offset=512):
+    """ Search for timestamp integer that start with 0x01D"""
+    return _layout_scanner(
+        source=obj, 
+        object_type=obj._context.modules['kernel'].symbol_table_name+constants.BANG+'unsigned long long',
+        condition=lambda timestamp: (timestamp & 0xff00000000000000 == 0x0100000000000000) and not obj._context.layers[obj.vol.layer_name].is_valid(timestamp, 8)
+    )
+
+def _etype_str(etype:int) -> str:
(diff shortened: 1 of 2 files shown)


volatility3/framework/plugins/windows/kerberos.py
```diff
@@ -0,0 +1,1042 @@
...
+                pslist.PsList.list_processes(
+                    context=self.context,
+                    kernel_module_name=self.config["kernel"],
+                    filter_func=self._lsass_proc_filter,
+                )
+            ),
+        )
+
+
+class Dump(interfaces.plugins.PluginInterface, KlistCommand):
```
ikelos: Rather than having a whole separate dump plugin, have you considered adding a `--dump` parameter to your other plugins, to write out any tickets they find?

volatility3/framework/plugins/windows/kerberos.py
```diff
@@ -0,0 +1,1042 @@
...
+        for address in self.config["address"]:
+            try:
+                ticket = kerberos_types.object(
+                    object_type="_KERB_TICKET_INFO", 
+                    offset=address,
+                    absolute=True
+                )
+                
+                path = os.path.join(self.config["output"][0], "%s.kirbi"%hex(ticket.vol.offset))
+                with open(path, "wb") as output:
```
ikelos:
```suggestion
                with self.open(f"{ticket.vol.offset:#X}.kirbi") as output:
```

You shouldn't use open here (or request an output path) because volatility may be being used as a library, and might be accessed from a web frontend or some other mechanism that doesn't have local storage attached. Instead you could use self.open which should work in a very similar way (it's a [FileHandlerInterface](https://github.com/volatilityfoundation/volatility3/blob/fb148ba25b885c2cc41479ed6f68ea1873bc9197/volatility3/framework/interfaces/plugins.py#L22)), but you provide just the name of the file (rather than a full path) and then the tool (the CLI or web interface) decides how and where to write the file out for the user.

volatility3/framework/plugins/windows/kerberos.py
```diff
@@ -0,0 +1,1042 @@
...
+
+    def run(self):
+        return renderers.TreeGrid(
+            [
+                ("Address", format_hints.Hex),
+                ("Client", str),
+                ("Server", str),
+                ("KerbTicket Encryption Type", str),
+                ("Ticket Flags", format_hints.Hex),
+                ("Start Time", str),
```
ikelos: It might be better to convert these to `datetime.datetime` objects, so that you can provide a timeliner interface for the plugin?  Then just have the plugin inherit from `timeliner.TimeLinerInterface` and define a `generate_timeline` method, and any times it produces can be recorded into a body file along with all the other times found by the timeliner plugin.  You can look as `windows.pslist` as an example if you'd like...  5:)

volatility3/framework/plugins/windows/kerberos.py
```diff
@@ -0,0 +1,1042 @@
...
+from volatility3.framework.objects import utility
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.symbols import intermed
+from volatility3.framework.symbols.windows import pdbutil, versions
+from volatility3.framework.symbols.windows.extensions import pe, conversion
+from volatility3.plugins.windows import pslist, vadinfo, pe_symbols
+
+vollog = logging.getLogger(__name__)    
+        
+def UnwrapOr(default):
```
ikelos: This is quite nifty as a wrapper, but we wouldn't generally recommend catching *all* exceptions, and certainly not without informing the user about serious or very unexpected exceptions.  At a minimum we'd recommend parameterizing it with a vollog.debug message so that someone can determine what went wrong if something does happen.

volatility3/framework/plugins/windows/kerberos.py
```diff
@@ -0,0 +1,1042 @@
...
+    """Will scan an object layout memory by casting it into object_type and respect condition
+        source: object which want to test the layout
+        condition: function call to to know if object was found
+        
+        :return: list of offset from source where condition is valid
+    """
+    result = []
+    # align for 64 bits
+    offset = start
+    layer = source._context.layers[source.vol.layer_name]
```
ikelos: We tend to recommend parameterizing things that are relied upon, so having a parameter for the layer_name to use, as a simple string.

For scanning you may also be interested in defining something which inherits from the `ScannerInterface` class.  This then lets the memory layer figure out how to chunk up the data being inspected, and yield out results as they're found.  There's lots of examples of it in the codebase.  You can just write a scanner, then hand it to the `layer.scan` method that should start yielding out any hits that the scanner finds.  You can also specify ranges to be scanned, so it won't cover the whole layer (similar to your start parameter).

volatility3/framework/plugins/windows/kerberos.py
```diff
@@ -0,0 +1,1042 @@
...
+        
+        
+class KERB_CREDENTIAL(objects.StructType):
+    """Use to format the session id to have the same everywhere"""
+    __primary_offset__ = None
+    
+    @UnwrapOr(None)
+    def get_primary(self):
+        """ Search for primary struct whch start by two unicode string"""
+        self.__class__.__primary_offset__ = self.__class__.__primary_offset__ or _scan_for_unicode_string(self)[0]
```
ikelos: We really don't recommend diving into private members so much.  This looks like it's trying to implement a singleton design pattern.  I would definitely recommend a separate, distinct singleton class rather than trying to stash a value in a private member of the class being used.

volatility3/framework/plugins/windows/kerberos.py
```diff
@@ -0,0 +1,1042 @@
...
+        
+    def get_logon_session_id(self):
+        if self.LogonSessionId_1.HighPart == 0 and self.LogonSessionId_1.LowPart == 0:
+            return self.LogonSessionId_2
+        else:
+            return self.LogonSessionId_1
+
+
+class KERB_PRIMARY_CREDENTIAL(objects.StructType):
+    __ticket_cache_offset__ = None
```
ikelos: Again, this isn't a good design pattern, and may also affect long lived systems (such as web interfaces) where the classes aren't reloaded.  If the plugin gets run twice, then this stale state will affect the new data being analyzed.  It should definitely be parameterized and then discovered at the start of the plugin run by the plugin, and passed to anything that needs it.

volatility3/framework/plugins/windows/kerberos.py
```diff
@@ -0,0 +1,1042 @@
...
+                "_KERB_TICKET_INFO": KERB_TICKET_INFO
+            },
+            table_mapping=table_mapping,
+        )
+
+        return context.module(
+            kerberos_symbol_table, proc_layer_name, offset=kerberos_base
+        )
+        
+    def _find_lsass_proc(
```
ikelos, line 466: There may already be code to locate the lsass plugin, it might be worth looking through `lsadump` or `hashdump` to see if they expose that functionality?  If you can avoid duplicating code that may already exist, it makes it easier to only have to fix issues in a single place.

ikelos (commented):
Thanks very much for your submission, this looks like a really interesting plugin!  In general there's a lot of good stuff in here and the style is pretty good.  The main issues were:

* Mass exception catching
* Lots of singleton cases that won't work with a long lived framework
* Direct file open which won't work for a web interface, for example

And some suggestions for simple enhancements:

* Make dumping tickets a flag, rather than a separate plugin
* Output datetime.datetimes and then implement a small timeliner interface to get cheap timeliner support
* See if there's already existing code for finding the lsass process

volatility3/framework/plugins/windows/kerberos.py
citronneur: Thanks it's now done

volatility3/framework/plugins/windows/kerberos.py
citronneur: Thanks it's now done

volatility3/framework/plugins/windows/kerberos.py
citronneur: Thanks it's now done, we've added dump argument to both Tickets and VadTicketScan

volatility3/framework/plugins/windows/kerberos.py
citronneur: We remove the line because at the end we don't need it

volatility3/framework/plugins/windows/kerberos.py
citronneur: We remove all singleton pattern, and scan each time we need, it's really improving code, thanks

volatility3/framework/plugins/windows/kerberos.py
citronneur: I was not able to find the code in other plugin so I keep it like that

volatility3/framework/plugins/windows/kerberos.py
citronneur: We keep it but renamed it @UnwrapMemoryOr as it's now only specialized in invalid memory wrapper, which can happened in unstructured memory layout like in VadScan plugin

volatility3/framework/plugins/windows/kerberos.py
citronneur: We removed it

citronneur:
Really thank you for your review, we applied all the modification you requested that really improved the code !!! I will also try to implement timeliner interface in next commit

citronneur:
We added the timeliner interface for the kerberos.Ticket plugin.

citronneur:
@ikelos don't hesitate to ping if you think the plugin needs more changes ;-)

citronneur:
@ikelos @the-rectifier Don't hesitate I'm available if you need support for this PR? Thanks

the-rectifier:
@citronneur don't know why I am tagged; I'm just a user of the project, and I try to fix issues I come across. :) That being said, I have tried your plugin in a couple of projects of mine, and it worked perfectly! Really nice addition! I'm sure the maintainers will come around on this; they're probably working on other things atm

citronneur:
Sorry @the-rectifier I thought you were a volatility 3 maintainer ;-) I tak ethis opportunity to thank you for your feedback and your work around this project !!!

PR #1937, opened 2026-01-31
Issues/issue1373 add windows support to banners

### LINKED ISSUE #1373: Add windows support to the banners plugin
Now that was have a unified "identifier" cache, we should search for the windows identifier/banner, not just the linux ones (and make sure we support mac too).

This might be pulling/unifying information from the automagics or something else, but the plugin should be runnable across any kind of image and pull out what it can.

ikelos:
This finds the GUID/age combo that volatility uses to uniquely identify a kernel.  Not *that* useful to a user, but if they ever post the output of banners, it should show us what's going on.

Closes #1373.

### DIFF
--- a/volatility3/framework/plugins/banners.py
+++ b/volatility3/framework/plugins/banners.py
@@ -89,7 +89,7 @@ class Banners(interfaces.plugins.PluginInterface):
         ):
             yield (
                 format_hints.Hex(offset),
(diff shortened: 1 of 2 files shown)


eve-mem:
This is one of those things that just makes perfect sense. A very good idea.

Issue #1938, opened 2026-02-09
Add Intel 5-level paging support


amebius:
**Is your feature request related to a problem? Please describe.**

On Linux dumps where the kernel uses **5-level paging (LA57)** at runtime (e.g. Ubuntu 20.04+ with `CONFIG_X86_5LEVEL` enabled by default), Volatility3 always uses the 4-level Intel32e layer. Virtual-to-physical translation and plugins can then fail or produce wrong results because the page table walk assumes only four levels (PML4 → PDPT → PD → PT), while the dump was taken with five levels (PML5 → PML4 → PDPT → PD → PT).

I was able to reproduced the issue processing a dump from a VM instance running Ubuntu 24.04.3 LTS (6.8.0-90-generic) on Intel Xeon Processor (Icelake) with la57 feature. The dump was obtained using `avml 0.14.0`.
Without a fix I got an empty list of processes running `linux.pslist` plugin and some plugins failed with error like:
```
volatility3.framework.exceptions.PagedInvalidAddressException: Page Fault at entry 0x0 in table page table
```
Volatility3 version used: current `develop` branch
dwarf2json version: 0.9.0

**Describe the solution you'd like**

- Detect at runtime whether 5-level paging is in use (e.g. via VMCOREINFO `NUMBER(pgtable_l5_enabled)` when present).
- When 5-level is enabled, use a dedicated 64-bit translation layer that walks five levels (57-bit virtual address, PML5 as top level).
- When 4-level is in use, keep using the existing Intel32e layer so behavior is unchanged for other dumps.

**Describe alternatives you've considered**

- Relying only on `CONFIG_X86_5LEVEL` in VMCOREINFO: not reliable, as it's often omitted when 5-level is default.
- Inferring from kernel version or symbols only: does not reflect runtime state (kernel can be 5-level-capable but run in 4-level mode on older hardware).

**Additional information**

- Detection should prefer the runtime flag `NUMBER(pgtable_l5_enabled)` from VMCOREINFO when available.
- 5-level layout: `_maxvirtaddr = 57`, and `_structure` adding "page map layer 5" (9 bits, not a large page) above the existing four levels.

---

#### 1. `volatility3/framework/automagic/linux.py`

Use 5-level detection when choosing the layer class:
```python
# In LinuxIntelVMCOREINFOStacker.stack(), where layer_class is set for 64-bit:
else:
    if cls._vmcoreinfo_is_5level(vmcoreinfo):
[...]

PR #1939, opened 2026-02-09
add 5-level paging detection and support

amebius:
Related feature request: https://github.com/volatilityfoundation/volatility3/issues/1938#issue-3917556232

### DIFF
--- a/volatility3/framework/automagic/linux.py
+++ b/volatility3/framework/automagic/linux.py
@@ -298,7 +298,10 @@ class LinuxIntelVMCOREINFOStacker(interfaces.automagic.StackerLayerInterface):
             if is_32bit:
                 layer_class = intel.IntelPAE if is_pae else intel.Intel
             else:
-                layer_class = intel.Intel32e
+                if cls._vmcoreinfo_is_5level(vmcoreinfo):
+                    layer_class = intel.Intel32e5Level
+                else:
+                    layer_class = intel.Intel32e
 
             uts_release = vmcoreinfo["OSRELEASE"]
 
@@ -428,3 +431,12 @@ class LinuxIntelVMCOREINFOStacker(interfaces.automagic.StackerLayerInterface):
             is_32bit = dtb_vaddr <= 2**32
 
         return is_32bit, is_pae
+
+    @staticmethod
+    def _vmcoreinfo_is_5level(vmcoreinfo) -> bool:
+        """Returns True if 5-level paging is enabled at runtime"""
+        pgtable_l5 = vmcoreinfo.get("NUMBER(pgtable_l5_enabled)")
+        if pgtable_l5 is not None:
+            return pgtable_l5 == 1
+
+        return False
--- a/volatility3/framework/layers/intel.py
+++ b/volatility3/framework/layers/intel.py
@@ -489,6 +489,26 @@ class Intel32e(Intel):
     ]
 
 
+class Intel32e5Level(Intel):
+    """Class for handling 64-bit (32-bit extensions) for Intel
+    architectures with 5 level page tables."""
+
+    _direct_metadata = collections.ChainMap(
+        {"architecture": "Intel64"}, Intel._direct_metadata
+    )
+    _entry_format = "<Q"
+    _bits_per_register = 64
+    _maxphyaddr = 52
+    _maxvirtaddr = 57
+    _structure = [
+        ("page map layer 5", 9, False),
+        ("page map layer 4", 9, False),
+        ("page directory pointer", 9, True),
+        ("page directory", 9, True),
+        ("page table", 9, False),
+    ]
+
+
 class WindowsMixin(Intel):
     @staticmethod
     def _page_is_valid(entry: int) -> bool:


ikelos:
@atcuno, can you look this over please?  I thought we had an old branch with level-5 support, but I can't find it anywhere.  The translation code looks ok to me, although I haven't run any tests (or checked the specs yer).  The changes are minimal and only choose the layer if the vmcoreinfo is found with the appropriate flag.  The associated issue (#1938) looks well thought out and looks like it's difficult to trip accidentally.  It should be possible to extend to windows once we find a way of identifying whether it's in use...

volatility3/framework/layers/intel.py
```diff
@@ -489,6 +489,26 @@ class Intel32e(Intel):
...
+    _entry_format = "<Q"
+    _bits_per_register = 64
+    _maxphyaddr = 52
+    _maxvirtaddr = 57
+    _structure = [
+        ("page map layer 5", 9, False),
+        ("page map layer 4", 9, False),
+        ("page directory pointer", 9, True),
+        ("page directory", 9, True),
+        ("page table", 9, False),
```
dreyercito, line 508: I tested with some local memdumps but only worked after setting the large_page check to True for page tables.

AmirGhiassian, line 508: Confirming these large-page flags look correct vs Intel LA57:

- PML5 / PML4 / PT → `False` (PS reserved or PAT on PTE)
- PDPT / PD → `True` (1GiB / 2MiB)

I would **not** set the page-table level to `True`. On PTEs, bit 7 is PAT, not a page-size bit; treating it as PS can produce false large-page hits. If a dump only works with PT=`True`, that is more likely a wrong DTB/layer/class issue than a missing PS flag here.

volatility3/framework/layers/intel.py
```diff
@@ -489,6 +489,26 @@ class Intel32e(Intel):
...
+    """Class for handling 64-bit (32-bit extensions) for Intel
+    architectures with 5 level page tables."""
+
+    _direct_metadata = collections.ChainMap(
+        {"architecture": "Intel64"}, Intel._direct_metadata
+    )
+    _entry_format = "<Q"
+    _bits_per_register = 64
+    _maxphyaddr = 52
+    _maxvirtaddr = 57
```
AmirGhiassian, line 502: Geometry check LGTM: 5×9 + 12 = 57, and `Intel._initial_position = maxvirtaddr - 1` yields bit 56 → after five shifts the walk stops at the 4K page offset. Canonical high bits become 63:57 automatically via the existing `canonicalize()` path.

volatility3/framework/automagic/linux.py
```diff
@@ -428,3 +431,12 @@ def _vmcoreinfo_is_32bit(vmcoreinfo) -> Tuple[bool, bool]:
             is_32bit = dtb_vaddr <= 2**32
 
         return is_32bit, is_pae
+
+    @staticmethod
+    def _vmcoreinfo_is_5level(vmcoreinfo) -> bool:
+        """Returns True if 5-level paging is enabled at runtime"""
+        pgtable_l5 = vmcoreinfo.get("NUMBER(pgtable_l5_enabled)")
+        if pgtable_l5 is not None:
+            return pgtable_l5 == 1
```
AmirGhiassian, line 440: Good signal choice — `NUMBER(pgtable_l5_enabled)` is the runtime value Linux exports in VMCOREINFO (not `CONFIG_X86_5LEVEL`).

Minor robustness nit (non-blocking): because `VMCoreInfo` already does `int(value, 0)` for `NUMBER(...)`, `== 1` is fine. If you ever want to be defensive against unexpected non-zero encodings you could use `bool(pgtable_l5)` / `pgtable_l5 != 0`, but `== 1` matches the kernel’s boolean NUMBER today.

volatility3/framework/automagic/linux.py
```diff
@@ -298,7 +298,10 @@ def stack(
             if is_32bit:
                 layer_class = intel.IntelPAE if is_pae else intel.Intel
             else:
-                layer_class = intel.Intel32e
+                if cls._vmcoreinfo_is_5level(vmcoreinfo):
+                    layer_class = intel.Intel32e5Level
```
AmirGhiassian, line 302: Wiring only the VMCOREINFO stacker is consistent with where the runtime flag actually lives.

Follow-up (non-blocking for this PR): `LinuxIntelStacker` still always picks `LinuxIntel32e` for 64-bit. Banner-only / no-VMCOREINFO dumps on LA57 hosts will still mis-walk. Also there is no `LinuxIntel32e5Level(LinuxMixin, Intel32e5Level)` yet — only needed if the banner path (or a future switch of this stacker to `LinuxIntel*`) grows 5-level support, since `LinuxMixin` owns PROT_NONE / Linux PTE validity.

AmirGhiassian (commented):
## Code review (Hermes)

Thanks for the clean, minimal fix — this is a real gap for modern LA57 Linux dumps.

### Verdict: **Comment** (not blocking; a few follow-ups worth tracking)

I checked the change against Intel LA57 / 5-level paging and the existing Vol3 translation walk. I do **not** have a 5-level dump to re-verify end-to-end, so this is a design/correctness review, not a runtime sign-off.

### ✅ Looks good

- **Walk geometry is correct.** `_maxvirtaddr = 57` with five 9-bit levels + 12-bit offset is exactly LA57. Starting position `maxvirtaddr - 1 = 56` leaves the page-offset bits at 11 after the walk.
- **Large-page flags match the SDM.** PS is only meaningful at PDPT (1 GiB) and PD (2 MiB). PML5 / PML4 / PT correctly use `large_page=False`. Bit 7 in a PTE is PAT, not PS — so **do not** flip the page-table entry to `True` (see note on the prior review comment below).
- **Detection signal is the right one.** `NUMBER(pgtable_l5_enabled)` is the runtime truth; `CONFIG_X86_5LEVEL` alone is not. Existing VMCOREINFO parsing already coerces `NUMBER(...)` via `int(value, 0)`, so `== 1` is type-correct.
- **Canonicalization / entry masking** inherit from `Intel` and scale with `_maxvirtaddr` / `_maxphyaddr` — no special-case hacks needed.
- **Scope discipline** is good: small diff, wires only the VMCOREINFO stacker path that actually has the runtime flag.
- CI on the head SHA is green.

### ⚠️ Non-blocking gaps / follow-ups

1. **`LinuxIntelStacker` (banner / non-VMCOREINFO path) still hardcodes `LinuxIntel32e`.** Dumps that never yield usable VMCOREINFO will keep getting 4-level walks. Probably out of scope for this PR if VMCOREINFO is the intended primary path, but worth calling out in the description / a follow-up issue — pure banner ASLR scans cannot see `pgtable_l5_enabled`.
2. **No `LinuxIntel32e5Level = LinuxMixin + Intel32e5Level`.** The VMCOREINFO stacker already uses bare `Intel32e` (pre-existing), so matching that pattern is consistent. The banner stacker uses `LinuxIntel*`, and `LinuxIntel32e`’s own comment already talks about 4- and 5-level tables while still forcing `_maxphyaddr = 46`. If/when the banner path grows 5-level support, it will need a `LinuxMixin` subclass (PROT_NONE / Linux PTE validity), not the bare class.
3. **No unit coverage** of the new class (structure width sum, `_maxvirtaddr`, canonicalize for bit 56 set). Understandable without a public LA57 image; a tiny pure-Python test would still lock the geometry.
4. **PR is quite old** (opened 2025-09, issue later marked stale). Maintainer ping already happened — thanks. A quick rebase onto current `develop` before merge would be polite.

### Note on prior review comment (large_page on PT)

@dreyercito reported needing `large_page=True` on page-table entries to make a dump work. Per Intel SDM that should not be required for a correct 4K PTE walk, and enabling it treats bit 7 as PS (it is PAT on PTEs), which can mis-decode entries. If a specific image only works with that flip, that points to a different bug (wrong layer class, bad DTB, non-standard PTE layout, etc.) rather than a missing PS flag on the PT level. Happy to dig if you still have that sample.

### Nit

`Intel32e5Level` could subclass `Intel32e` and only override `_maxvirtaddr` + `_structure` to avoid duplicating the Intel64 metadata / entry format — pure style, not required.

---
*Reviewed against head `34e065bbefe6eea9f64557ea028db33eb96fa437` and Intel LA57 geometry; no LA57 memory image available on my side.*

PR #1950, opened 2026-03-05
Feature: BitLocker FVEK extraction plugin

forensicxlab:
Bitlocker plugin ported from the Volatility 2 BitLocker key extractor to recover FVEKs from Windows memory pools (Windows 7 through Windows 10/Server generations) and inspired by the latest version in memprocfs.

- https://github.com/elceef/bitlocker/blob/master/bitlocker.py
- https://github.com/ufrisk/MemProcFS/blob/master/vmm/modules/m_misc_bitlocker.c

Example output: 

```
Volatility 3 Framework 2.27.0
Address	Cipher	FVEK	TWEAK	DumpFile	DislockerFile

0xd284f3d83890	AES-XTS 128 bit (Win 10+)	929baf9supersecretbitlockerkeyredactedfordemonstration263d	Not Applicable	N/A	N/A
0xd28508294440	AES 256-bit (Win 8+)	0373a4supersecretbitlockerkeyredactedfordemonstrationb304ecdae1a2b	Not Applicable	N/A	N/A
0xd2851008f7b0	AES 128-bit (Win 8+)	98supersecretbitlockerkeyn9835b	Not Applicable	N/A	N/A
0xd2851d335000	AES 128-bit (Win 8+)	3fd7fsupersecretbitlockerkeynf7b698ffa920	Not Applicable	N/A	N/A
```

Then works to decrypt the volume: 

```
 exhume_filesystem --body /Volumes/Forensics/Suspect.aff -o 0x7500000 -s 0x743af00000  --fvek 929baf9supersecretbitlockerkeyredactedfordemonstration263d --record 5

[2026-03-05T21:32:03Z INFO  exhume_body::aff] AFF: parsed 29809 pages, pagesize=16777216, imagesize=500107862016
[2026-03-05T21:32:03Z INFO  exhume_body] Detected an AFF disk image.
[2026-03-05T21:32:03Z ERROR exhume_ntfs] The partition is BitLocker-encrypted (OEM ID: -FVE-FS-). NTFS metadata cannot be read without decryption.
[2026-03-05T21:32:03Z INFO  exhume_filesystem::detected_fs] BitLocker detected. Attempting to decrypt with provided FVEK...
[2026-03-05T21:32:04Z INFO  exhume_filesystem::detected_fs] Successfully detected BitLocker-decrypted NT filesystem.
+--------------------------+-----------------------+
| MFT Entry Header Values  |                       |
+--------------------------+-----------------------+
| Sequence                 | 5                     |
+--------------------------+-----------------------+
| $LogFile Sequence Number | 2776238966            |
+--------------------------+-----------------------+
| Flags                    | Allocated | Directory |
+--------------------------+-----------------------+
| Links                    | 1                     |
+--------------------------+-----------------------+


+--------------------------------+-----------+--------------+------+
| Attributes                     | Name      | Status       | Size |
+--------------------------------+-----------+--------------+------+
| StandardInformation (0x10‑#0)  | N/A       | Resident     | 72   |
+--------------------------------+-----------+--------------+------+
| FileName (0x30‑#1)             | N/A       | Resident     | 68   |
+--------------------------------+-----------+--------------+------+
| IndexRoot (0x90‑#6)            | $I30      | Resident     | 176  |
+--------------------------------+-----------+--------------+------+
| IndexAllocation (0xA0‑#8)      | $I30      | Non‑resident | 8192 |
+--------------------------------+-----------+--------------+------+
| Bitmap (0xB0‑#7)               | $I30      | Resident     | 8    |
+--------------------------------+-----------+--------------+------+
| LoggedUtilityStream (0x100‑#9) | $TXF_DATA | Resident     | 56   |
+--------------------------------+-----------+--------------+------+
```

Note that another bitlocker container was present on the machine. Which was also retrieved.

### DIFF
--- a/volatility3/framework/plugins/windows/bitlocker.py
+++ b/volatility3/framework/plugins/windows/bitlocker.py
@@ -5,7 +5,7 @@
 import logging
 import os
 from dataclasses import dataclass
-from typing import Dict, Iterable, List, Optional, Tuple
+from typing import Dict, Iterable, List, Optional, Tuple, Union
 
 from volatility3.framework import interfaces, renderers, symbols
 from volatility3.framework.configuration import requirements
@@ -291,13 +291,13 @@ class Bitlocker(interfaces.plugins.PluginInterface):
                 )
 
     def _dump_files(self, result: FvekResult) -> Tuple[
-        interfaces.renderers.BaseAbsentValue | str,
-        interfaces.renderers.BaseAbsentValue | str,
+        Union[interfaces.renderers.BaseAbsentValue, str],
+        Union[interfaces.renderers.BaseAbsentValue, str],
     ]:
-        dump_file: interfaces.renderers.BaseAbsentValue | str = (
+        dump_file: Union[interfaces.renderers.BaseAbsentValue, str] = (
             renderers.NotApplicableValue()
         )
-        dislocker_file: interfaces.renderers.BaseAbsentValue | str = (
+        dislocker_file: Union[interfaces.renderers.BaseAbsentValue, str] = (
             renderers.NotApplicableValue()
         )
 
--- a/volatility3/framework/plugins/windows/bitlocker.py
+++ b/volatility3/framework/plugins/windows/bitlocker.py
@@ -1,0 +1,432 @@
+# This file is Copyright 2026 Volatility Foundation and licensed under the Volatility Software License 1.0
+# which is available at https://www.volatilityfoundation.org/license/vsl-v1.0
+#
+
+import logging
+import os
+from dataclasses import dataclass
+from typing import Dict, Iterable, List, Optional, Tuple
+
+from volatility3.framework import interfaces, renderers, symbols
+from volatility3.framework.configuration import requirements
+from volatility3.framework.renderers import format_hints
+from volatility3.framework.symbols.windows import versions
+from volatility3.plugins.windows import poolscanner
+
+vollog = logging.getLogger(__name__)
+
+
+@dataclass
+class FvekResult:
+    offset: int
+    cipher: str
+    fvek: bytes
+    tweak: Optional[bytes]
+    dislocker_blob: Optional[bytes]
+
+
+class Bitlocker(interfaces.plugins.PluginInterface):
+    """Extracts BitLocker FVEK keys from Windows memory."""
+
+    _required_framework_version = (2, 0, 0)
+    _version = (1, 0, 0)
+
+    _BLMODE: Dict[str, str] = {
+        "00": "AES 128-bit with Diffuser",
+        "01": "AES 256-bit with Diffuser",
+        "02": "AES 128-bit",
+        "03": "AES 256-bit",
+        "10": "AES 128-bit (Win 8+)",
+        "20": "AES 256-bit (Win 8+)",
+        "30": "AES-XTS 128 bit (Win 10+)",
+        "40": "AES-XTS 256 bit (Win 10+)",
+    }
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Windows kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.StringRequirement(
+                name="dump-dir",
+                description="Directory in which to dump FVEK values",
+                optional=True,
+                default=None,
+            ),
+            requirements.StringRequirement(
+                name="dislocker",
+                description="Directory in which to dump FVEK values for Dislocker",
+                optional=True,
+                default=None,
+            ),
+            requirements.BooleanRequirement(
+                name="verbose",
+                description="Enable verbose logging",
+                optional=True,
+                default=False,
+            ),
+            requirements.BooleanRequirement(
+                name="debug",
+                description="Enable debug hexdump logging",
+                optional=True,
+                default=False,
+            ),
+            requirements.VersionRequirement(
+                name="poolscanner", component=poolscanner.PoolScanner, version=(3, 0, 1)
+            ),
+        ]
+
+    @staticmethod
+    def _key_hex(key: bytes) -> str:
+        return key.hex()
+
+    @staticmethod
+    def _version_tuple(
+        context: interfaces.context.ContextInterface, symbol_table_name: str
+    ) -> Tuple[int, int, int, int]:
+        metadata = context.symbol_space[symbol_table_name].metadata
+        pe_version = getattr(metadata, "pe_version", None)
+        if not pe_version:
+            return (0, 0, 0, 0)
+        normalized = list(pe_version)[:4]
+        while len(normalized) < 4:
+            normalized.append(0)
+        return tuple(int(v) for v in normalized)
+
+    def _iter_none_pools(
+        self, kernel: interfaces.context.ModuleInterface, is_64bit: bool
+    ) -> Iterable[FvekResult]:
+        if not is_64bit:
+            vollog.debug(
+                "Skipping None-tag branch on x86; offsets are only defined for x64"
+            )
+            return
+
+        if self.config.get("verbose"):
+            vollog.info("Scanning None-tag pools for Win10+ BitLocker keys")
+
+        constraints = [
+            poolscanner.PoolConstraint(
+                b"None",
+                type_name=kernel.symbol_table_name + "!_POOL_HEADER",
+                page_type=poolscanner.PoolType.NONPAGED,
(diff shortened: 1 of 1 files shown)


volatility3/framework/plugins/windows/bitlocker.py
```diff
@@ -0,0 +1,432 @@
...
+    cipher: str
+    fvek: bytes
+    tweak: Optional[bytes]
+    dislocker_blob: Optional[bytes]
+
+
+class Bitlocker(interfaces.plugins.PluginInterface):
+    """Extracts BitLocker FVEK keys from Windows memory."""
+
+    _required_framework_version = (2, 0, 0)
```
ikelos, line 31: Does this definitely work with frameworks right back to 2.0.0?  If you're not sure, then this should be the current framework version, so...
```suggestion
    _required_framework_version = (2, 28, 0)
```

volatility3/framework/plugins/windows/bitlocker.py
```diff
@@ -0,0 +1,432 @@
...
+            renderers.NotApplicableValue()
+        )
+        dislocker_file: Union[interfaces.renderers.BaseAbsentValue, str] = (
+            renderers.NotApplicableValue()
+        )
+
+        dump_dir = self.config.get("dump-dir")
+        if dump_dir:
+            os.makedirs(dump_dir, exist_ok=True)
+            dump_file = os.path.join(dump_dir, f"{result.offset:#010x}.fvek")
```
ikelos, line 307:
```suggestion
            dump_file = f"{result.offset:#010x}.fvek"
```

ikelos: We shouldn't be writing to a directory since we'll be using `self.open`...

volatility3/framework/plugins/windows/bitlocker.py
```diff
@@ -0,0 +1,432 @@
...
+            text = self._key_hex(result.fvek)
+            if result.tweak is not None:
+                text = f"{text}:{self._key_hex(result.tweak)}"
+            with open(dump_file, "w", encoding="utf-8") as fvek_file:
```
ikelos, line 311:
```suggestion
            with self.open(dump_file, "w", encoding="utf-8") as fvek_file:
```
I don't recall whether `self.open` supports modes or encodings, so worth testing, but we shouldn't be directly opening files in plugins for the reasons listed above (support as a library).

volatility3/framework/plugins/windows/bitlocker.py
```diff
@@ -0,0 +1,432 @@
...
+                fvek_file.write(text + "\n")
+
+        dislocker_dir = self.config.get("dislocker")
+        if dislocker_dir and result.dislocker_blob:
+            os.makedirs(dislocker_dir, exist_ok=True)
+            dislocker_file = os.path.join(
+                dislocker_dir, f"{result.offset:#010x}-Dislocker.fvek"
+            )
+            with open(dislocker_file, "wb") as dislocker_handle:
```
ikelos, line 320:
```suggestion
            with self.open(dislocker_file, "wb") as dislocker_handle:
```

volatility3/framework/plugins/windows/bitlocker.py
```diff
@@ -0,0 +1,432 @@
...
+                text = f"{text}:{self._key_hex(result.tweak)}"
+            with open(dump_file, "w", encoding="utf-8") as fvek_file:
+                fvek_file.write(text + "\n")
+
+        dislocker_dir = self.config.get("dislocker")
+        if dislocker_dir and result.dislocker_blob:
+            os.makedirs(dislocker_dir, exist_ok=True)
+            dislocker_file = os.path.join(
+                dislocker_dir, f"{result.offset:#010x}-Dislocker.fvek"
+            )
```
ikelos, line 319:
```suggestion
            dislocker_file = f"{result.offset:#010x}-Dislocker.fvek"
```

volatility3/framework/plugins/windows/bitlocker.py
```diff
@@ -0,0 +1,432 @@
...
+                name="dump-dir",
+                description="Directory in which to dump FVEK values",
+                optional=True,
+                default=None,
+            ),
+            requirements.StringRequirement(
+                name="dislocker",
+                description="Directory in which to dump FVEK values for Dislocker",
+                optional=True,
+                default=None,
```
ikelos, line 63:
```suggestion
                action="store_true",
```

volatility3/framework/plugins/windows/bitlocker.py
```diff
@@ -0,0 +1,432 @@
...
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Windows kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.StringRequirement(
+                name="dump-dir",
+                description="Directory in which to dump FVEK values",
+                optional=True,
+                default=None,
```
ikelos, line 57:
```suggestion
                action="store_true",
```

volatility3/framework/plugins/windows/bitlocker.py
```diff
@@ -0,0 +1,432 @@
...
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Windows kernel",
+                architectures=["Intel32", "Intel64"],
+            ),
+            requirements.StringRequirement(
```
ikelos, line 53: We don't open or dump to directories in plugins anymore, since that would break if the plugin were executed through a web API, so this should be producing files using the `self.open()` mechanism.  I've provided suggestions below but they should be tested locally to ensure they work correctly, rather than just applied.

volatility3/framework/plugins/windows/bitlocker.py
```diff
@@ -0,0 +1,432 @@
...
+        ]
+
+    @staticmethod
+    def _key_hex(key: bytes) -> str:
+        return key.hex()
+
+    @staticmethod
+    def _version_tuple(
+        context: interfaces.context.ContextInterface, symbol_table_name: str
+    ) -> Tuple[int, int, int, int]:
```
ikelos, line 89:
```suggestion
    ) -> Tuple[int, int, int, int]:
        """Extracts the kernel symbol table's windows version"""
```

volatility3/framework/plugins/windows/bitlocker.py
```diff
@@ -0,0 +1,432 @@
...
+        if not pe_version:
+            return (0, 0, 0, 0)
+        normalized = list(pe_version)[:4]
+        while len(normalized) < 4:
+            normalized.append(0)
+        return tuple(int(v) for v in normalized)
+
+    def _iter_none_pools(
+        self, kernel: interfaces.context.ModuleInterface, is_64bit: bool
+    ) -> Iterable[FvekResult]:
```
ikelos, line 101: This could do with a docstring to explain what the function's purpose is, `_iter_none_pools` isn't immediately obvious.

volatility3/framework/plugins/windows/bitlocker.py
```diff
@@ -0,0 +1,432 @@
...
+            _constraint,
+            _mem_object,
+            header,
+        ) in poolscanner.PoolScanner.generate_pool_scan(
+            self.context, self.config["kernel"], constraints
+        ):
+            layer = self.context.layers[header.vol.layer_name]
+            base_offset = header.vol.offset
+
+            f1 = layer.read(base_offset + 0x9C, 64, pad=True)
```
ikelos, line 131: These offsets seem a little brittle?  Would it not be better to store them as data, with in a JSON file, or perhaps as a data structure that can be instantiated (and therefore updated based on the version of windows, in case they ever change?).

ikelos (commented):
Thanks, this is pretty nice, but there's two major points that stick out to me.  Firstly it writes files directly, which is bad if the plugin's run via a web UI or some other mechanism that doesn't have local access, so plugins should always write out file using the `self.open` mechanism, rather than just plain `open`.  They should be similar (although not identical, so please test it) and that will allow web UIs to provide the files back to users if they wish.

The second thing is there's a lot of manual image reading and processing of raw bytes, which sometimes is an indication of things that should really be done by volatility but aren't for some reason.  A lot of hardcoded specific values is also generally bad (in case they change in the future, figuring out where those numbers came from or why they were chosen, etc).  If they're structure offsets, ideally the structure would be defined in a local JSON symbol table that would be loaded up when this plugin starts (see `windows.getsids` as an example).  Otherwise comments referencing where those values came from would be useful.  Otherwise it's all a bit black magic and difficult for people to audit/verify...

gilbutter1-coder:
any new ?

PR #1960, opened 2026-03-10
Support Cryptodome namespace when Crypto is unavailable

oh2fih:
### Problem

Some Linux distributions (notably Debian and Ubuntu) package PyCryptodome (`python3-pycryptodome`) without the legacy `Crypto` namespace in order to [avoid conflicts](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=886291) with the deprecated PyCrypto package. In those environments, the library is only available under the `Cryptodome` namespace.

As a result, imports like:
```
from Crypto.Cipher import ARC4, DES, AES
```
fail with:
```
ModuleNotFoundError: No module named 'Crypto'
```
Example error:
```
DEBUG    volatility3.framework: Failed to import module volatility3.plugins.windows.registry.hashdump based on file: /usr/lib/python3/dist-packages/volatility3/framework/plugins/windows/registry/hashdump.py
DEBUG    volatility3.framework: Traceback (most recent call last):
  . . .
  File "/usr/lib/python3/dist-packages/volatility3/framework/plugins/windows/registry/lsadump.py", line 9, in <module>
    from Crypto.Cipher import ARC4, DES, AES
ModuleNotFoundError: No module named 'Crypto'
```

### Solution

This PR adds a fallback so that if `Crypto` is unavailable, the corresponding `Cryptodome` namespace is used. As no behavior changes occur on systems where `Crypto` is already available, this keeps compatibility with both:
- upstream pip-installed PyCryptodome (`Crypto` namespace)
- Debian/Ubuntu packaged PyCryptodome (`Cryptodome` namespace)

### DIFF
--- a/volatility3/framework/plugins/windows/registry/cachedump.py
+++ b/volatility3/framework/plugins/windows/registry/cachedump.py
@@ -5,8 +5,13 @@ import logging
 from struct import unpack
 from typing import Tuple
 
-from Crypto.Cipher import ARC4, AES
-from Crypto.Hash import HMAC
+try:
+    from Crypto.Cipher import ARC4, AES
+    from Crypto.Hash import HMAC
+except ImportError:
+    # Debian/Ubuntu ship pycryptodome under Cryptodome namespace
+    from Cryptodome.Cipher import ARC4, AES
+    from Cryptodome.Hash import HMAC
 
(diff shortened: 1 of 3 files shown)


ikelos (approved):
This looks good, thanks for the improvement!  5:)

PR #1962, opened 2026-03-12
Switch to using github infrastructure for test cases...

ikelos:
Things to check are how quickly the tests can run (how long the download phases take).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -26,19 +26,19 @@ jobs:
         run: |
           mkdir test_images
           cd test_images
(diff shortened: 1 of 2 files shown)


ikelos:
All seems to work and doesn't look like the downloading added to the timing significantly, so...  I think we're good to go?  The one downside is the downloads will need to be manually updated because the point to the specific release version to get to the file (so no automatic updates).  Probably for the best, since we don't have to have to re-upload images every time...

ikelos:
Seems pulling the images now takes 53s rather than 31 on the old infrastructure?  That seems acceptable given tests regularly take many minutes to complete...

PR #1963, opened 2026-03-17
Update macOS tutorial with support warning

cpuu:
Added a warning about the discontinuation of macOS support for Volatility 3 and recommended commercial solutions for macOS memory forensics.


ikelos:
The warning would've been fine, but I don't think we can plug specific alternatives for mac analysis, particularly when they're developed by members of the team, so the last paragraph would need to go.

PR #1967, opened 2026-03-31
Creation of the Shutdown Plugin

Frascott05:
I added the Shutdown plugin from volatility2 that could be useful for some digital forensic analysis.
Hope this can be added to the official repo.

### DIFF
--- a/volatility3/framework/plugins/windows/registry/shutdown.py
+++ b/volatility3/framework/plugins/windows/registry/shutdown.py
@@ -1,6 +1,5 @@
 import logging
 import struct
-import datetime
 from typing import Iterable, Optional
 
 from volatility3.framework import interfaces, renderers, exceptions
@@ -11,8 +10,6 @@ from volatility3.plugins.windows.registry import hivelist
 from volatility3.framework.renderers import conversion
 
 vollog = logging.getLogger(__name__)
-
-
 class LastShutdown(interfaces.plugins.PluginInterface):
     """Extract last Windows shutdown time from registry"""
 
--- a/volatility3/framework/plugins/windows/registry/shutdown.py
+++ b/volatility3/framework/plugins/windows/registry/shutdown.py
@@ -76,7 +76,7 @@ class LastShutdown(interfaces.plugins.PluginInterface):
                         continue
 
                     filetime = struct.unpack("<Q", data[:8])[0]
-                    shutdown_time = self.filetime_to_dt(filetime)
+                    shutdown_time = conversion.wintime_to_datetime(filetime)
 
                     yield (0, (key_path, str(shutdown_time)))
 
--- a/volatility3/framework/plugins/windows/registry/shutdown.py
+++ b/volatility3/framework/plugins/windows/registry/shutdown.py
@@ -48,7 +48,11 @@ class LastShutdown(interfaces.plugins.PluginInterface):
         if not syshive:
             return
 
-        # try multiple controlsets (robust approach)
+        # Windows systems typically maintain a small number of ControlSets
+        # (commonly 1–3, but occasionally more in recovery scenarios).
+        # We iterate over a bounded range to ensure we can recover shutdown
+        # time even if the active ControlSet cannot be determined via
+        # Select\\Current or if registry data is partially corrupted.
         for i in range(1, 5):
 
             key_path = f"ControlSet{i:03}\\Control\\Windows"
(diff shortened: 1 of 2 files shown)


eve-mem:
Hey @Frascott05 - nice! I've tested it on a few samples and it seems to work.

The old vol plugin would find the current control set itself or default to 1 https://github.com/volatilityfoundation/volatility/blob/328a178edeecaf8d2abf96618ecfcb14a7b0c860/volatility/plugins/registry/shutdown.py#L60-L63 

It looks like your method will try them all from 1-5? I don't know much about windows, is 5 the maximum? It would be nice to explain why the magic number 5 was chosen in the code, just in case something changes in the future. 

There is also a handy function used in lots of plugins to convert the filetime in wintime_to_datetime, so you shouldn't need to write you own filetime_to_dt.  (I've attempted to use the suggestions feature of the first time to explain what I mean... hopefully it makes sense)

If you can it would be really helpful to write a test for this plugin too, that way it can automatically checked in the future.

volatility3/framework/plugins/windows/registry/shutdown.py
```diff
@@ -0,0 +1,102 @@
...
+                    try:
+                        data = syshive.read(v.Data + 4, v.DataLength)
+                    except exceptions.InvalidAddressException:
+                        continue
+
+                    if not data or len(data) < 8:
+                        continue
+
+                    filetime = struct.unpack("<Q", data[:8])[0]
+                    shutdown_time = self.filetime_to_dt(filetime)
```
eve-mem:
```suggestion
                    shutdown_time = conversion.wintime_to_datetime(filetime)
```

volatility3/framework/plugins/windows/registry/shutdown.py
```diff
@@ -0,0 +1,102 @@
...
+import struct
+import datetime
+from typing import Iterable, Optional
+
+from volatility3.framework import interfaces, renderers, exceptions
+from volatility3.framework.configuration import requirements
+from volatility3.framework.layers import registry as registry_layer
+from volatility3.framework.symbols.windows.extensions import registry
+from volatility3.plugins.windows.registry import hivelist
+
```
eve-mem, line 12:
```suggestion
from volatility3.framework.renderers import conversion

```

volatility3/framework/plugins/windows/registry/shutdown.py
```diff
@@ -0,0 +1,102 @@
...
+                version=(2, 0, 0),
+            ),
+        ]
+
+    @staticmethod
+    def filetime_to_dt(filetime: int) -> datetime.datetime:
+        return datetime.datetime(1601, 1, 1) + datetime.timedelta(
+            microseconds=filetime / 10
+        )
+
```
eve-mem:
```suggestion
```

Frascott05:
Hi @eve-mem,
I really thank you for your suggestion, I commited them to the repository.

Despite the old shutdown plugin was able to find the current control set itself or default to 1,
Windows usually maintain a small numer of ControlSets (typically 3), but in some scenarios of recovery the
number of ControlSets could be more. So I prefered to use an average number that would be able to find the 
shutdown in almost all the scenarios

Frascott05:
Also, do you have any suggestions or examples to follow for writing a test?

eve-mem:
If you look at `test/plugins/windows` then you'll see how some of the plugins are tested.


Here's an example:

https://github.com/volatilityfoundation/volatility3/blob/fc8edc4f66d6c5a1c42ce70eb5243ddda04c2e74/test/plugins/windows/windows.py#L214-L235


There are two different windows samples at the moment to test with:

https://github.com/volatilityfoundation/volatility3/blob/fc8edc4f66d6c5a1c42ce70eb5243ddda04c2e74/test/__init__.py#L13-L17

So you could write a test that runs your plugins against both samples and asserts that the output is as expected. I hope that helps.

Frascott05:
sorry, I could't do it today, I hope I'll be able to work on the test case tomorrow

eve-mem:
Don't worry, it can take weeks and months to get pull request reviewed and checked. There's no rush.

Frascott05:
I've added the test class that checks for the right output of the plugin, thank you for the help

Frascott05:
Is everything fine with the test class?

Frascott05:
I removed the unused import of datetime on the plugin, hoping it can be merged soon

Issue #1886, opened 2025-10-02
Hyper-V Save State support  (.vmrs files)


semchapeu:
When you save the state of a Hyper-V VM a VMRS file is created. From what I can gather (!) this file contains, amongst other state information, a full uncompressed memory dump of the VM. 

I like that you can use volatility directly on the .vmem files that VMware produces when saving the state, and wish the same could be done with Hyper-V save states.

I found this which may be helpful for reference, but it appears to be using a Microsoft DLL.
https://github.com/ufrisk/LeechCore-plugins/tree/master?tab=readme-ov-file#leechcore_device_hvsavedstate

PR #1969, opened 2026-04-05
Adding support for Hyper-V VMRS state files

oxnan:
### TL;DR:
Adding support for Hyper-V .vmrs files in volatiliaty3 by adding a new layer for support.

### Description
By reverse engineering the vmsavedstatedumpprovider.dll file mention in #1886 I've managed to implement a translation layer for VMRS files.

When running volatility on windows it'll attempt to use native functions from ntdll to speed up the process of decompressing memory segments, and on other host OS it will handle the decompression using native python.

The performance is much better on Windows for natural reasons, but when running on an ubuntu 24.04 host it still manages to run banners.Banners on a 2gb memory image (8gb uncompressed).

### Examples files

Following links are for example VMRS files, one from a windows 10 host, and one from an Ubuntu 24.04 host. Both can run plugins with no issues.

Ubuntu 24.04 .vmrs file: https://mscf.cab/u/159
Windows 10 .vmrs file: https://mscf.cab/u/15b

### Notes/Thoughts

For full disclosure, parts of the code was written by an LLM, however I have gone through the LLM generated sections with a fine tooth comb to try to weed out any issues. 

Most importantly, the layerwriter plugin appears to be creating consistent output to raw that matches my other local implementations along with the native functions in the windows kernel, so I would say it's ready for use by others.

If there's anything I've missed, or mistakes that I've made in submitting this PR, please let me know, and I'll do my very best to take care of it in a timely fashion :)

### DIFF
--- a/volatility3/framework/layers/vmrs.py
+++ b/volatility3/framework/layers/vmrs.py
@@ -216,6 +216,7 @@ class VMRSExportManifest:
     def from_bytes(cls, data: bytes) -> "VMRSExportManifest":
         return cls.from_dict(json.loads(data.decode("utf-8")))
 
+
 class VMRSLayer(segmented.NonLinearlySegmentedLayer):
     provides = {"type": "physical"}
     EXPORT_MANIFEST_VERSION = 1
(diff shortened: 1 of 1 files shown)


oxnan:
I saw that there was issues with running ruff for linting/formatting. Not sure if I missed a step initially, but I have run manually now @ikelos , sorry for the hassle :)

PR #1974, opened 2026-04-12
feature: added iptables (legacy) and iptables-nft plugins

Tasty-Murder:
- linux.iptables — Extracts active firewall rules from Linux memory images by walking the ipt_table / ip6t_table kernel structures via the ISF symbol table. Outputs rules for both IPv4 and IPv6 across all tables (filter, nat, mangle, raw) with their match expressions and targets, equivalent to iptables-save / ip6tables-save without requiring a live shell.                                                                    
                                                                                                                                                  
- linux.iptables_nft — Extracts firewall rules from the nf_tables subsystem, covering both native nftables rules and rules loaded through the iptables-nft compatibility layer. Decodes native nft expressions (meta, payload, cmp, nat, log, limit, range) as well as xt_compat match extensions (conntrack, addrtype, multiport, etc.), producing human-readable output similar to nft list ruleset.

Both plugins output: network namespace, address family, table, chain, default policy, rule index, decoded match string, and target. They support Docker-generated rulesets, interface negation, IPv6 addresses, and --ctstate conntrack matching.


ikelos:
Wow, thanks for submitting this!  It is huge though, so it may take us a while to go through it...  I've not seen anyone go off to the internet to pull down source code to figure out how to parse data out of the image though, so I'll need to do some thinking on that (and if it seems solid it might be something we add as support in the library itself).  I've asked @atcuno and @Abyss-W4tcher, as our resident linux gurus to look through this.  5:)

ikelos:
Github's security is a little strict, but it's generally quite helpful.  So is ruff, sadly, it helps make sure all our formatting is consistent even if no one actually likes every change it makes.

PR #1976, opened 2026-04-29
Linux: Unify banner scanning and fix memory leaks

### LINKED ISSUE #1975: linux.pagecache.RecoverFs consuming massive memory until reaped
**Describe the bug**
I recently ran linux.pagecache.RecoverFs on an 8GB memory sample. By the time Volatility was killed by the out of memory reaper it had consumed over 90GB of memory and produced 13615 lines of output, but no tar.gz file.

**Context**
Volatility Version:  2.27.1
Operating System:  Debian 12 (Bookworm)
Python Version:  3.11.2
Suspected Operating System:  Debian 13
Command:  vol -q -f memory.lime linux.pagecache.RecoverFs

**To Reproduce**

1. Download sample image from https://deerrunassoc-my.sharepoint.com/:u:/g/personal/hal_deer-run_com/IQBYc4qlNKNASKeZwreVYtcIAQnSo3yEZPiyQk8CYtTkOvI?e=Wrk496
2. Run "vol -q -f memory.lime linux.pagecache.RecoverFs"
3. Watch the memory count up using "top" or similar
4. Wait for process to be reaped

**Expected behavior**
Clearly this plugin should never consume so much RAM. 

I would further suggest that the plugin should incrementally dump files as it finds them. That way even if the plugin aborts for some reason the analyst would at least get some information.

**Example output**
The memory image is available above. Generate whatever output is most useful.

**Additional information**
N/A

Abyss-W4tcher:
Hi, this PR:

- removes `mss = scanners.MultiStringScanner([x for x in self.banners if x is not None])` as it allocates a huge trie memory object when remote ISF indexes are in use -> replaced with a regex pattern (like the `banners` plugin) for identical results;
- unifies banner scanning methods of "symbol_finder.py", "linux.py" and "banners.py" in "banners_scanners.py". This also tells the user when the framework is searching for a banner: "Scanning memory_layer using BannerScanner"

For reference I've tracked down the memory leaks using memray and scalene.

### DIFF
--- a/volatility3/framework/automagic/banner_scanners.py
+++ b/volatility3/framework/automagic/banner_scanners.py
@@ -2,7 +2,7 @@ from typing import Iterator, Optional, Tuple
 from volatility3.framework.layers import scanners
 
 VALID_BANNER_CHARSET = (
-    b" #()+,;/-.0123456789:@ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz~"
+    b" #()+,;/-.0123456789:@ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz~\n"
 )
 BANNER_READ_SIZE = 0xFFF
 
@@ -25,6 +25,7 @@ class BannerScanner(scanners.RegExScanner):
         """Gets the banner at a layer offset and validates it."""
         layer = self.context.layers[self.layer_name]
         data = layer.read(offset, BANNER_READ_SIZE, pad=True)
+        # See symbol_cache's _normalize_identifier
         data_index = data.find(b"\x00")
         if data_index <= 0:
             return None
--- a/volatility3/framework/automagic/banner_scanners.py
+++ b/volatility3/framework/automagic/banner_scanners.py
@@ -1,0 +1,51 @@
+from typing import Iterator, Optional, Tuple
+from volatility3.framework.layers import scanners
+
+VALID_BANNER_CHARSET = (
+    b" #()+,;/-.0123456789:@ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz~"
+)
+BANNER_READ_SIZE = 0xFFF
+
+
+class BannerScanner(scanners.RegExScanner):
+    """Scanner for Linux and macOS kernel version strings."""
+
+    BANNER_PATTERN = (
+        rb"(Linux version|Darwin Kernel Version) [0-9]+\.[0-9]+\.[0-9]+[^\x00]+"
+    )
+
+    _version = (1, 0, 0)
+
+    _required_framework_version = (2, 0, 0)
+
+    def __init__(self) -> None:
+        super().__init__(pattern=self.BANNER_PATTERN)
+
+    def _get_valid_banner(self, offset: int) -> Optional[bytes]:
+        """Gets the banner at a layer offset and validates it."""
+        layer = self.context.layers[self.layer_name]
+        data = layer.read(offset, BANNER_READ_SIZE, pad=True)
+        data_index = data.find(b"\x00")
+        if data_index <= 0:
+            return None
+
+        data = data[:data_index].strip()
+        failed = any(char not in VALID_BANNER_CHARSET for char in data)
+        if not failed:
+            return data
+
+        return None
+
+    def __call__(self, data: bytes, data_offset: int) -> Iterator[Tuple[int, bytes]]:
+        for off in super().__call__(data, data_offset):
+            banner = self._get_valid_banner(off)
+            if banner is not None:
+                yield off, banner
+
+
+class LinuxBannerScanner(BannerScanner):
+    BANNER_PATTERN = rb"Linux version [0-9]+\.[0-9]+\.[0-9]+[^\x00]+"
+
+
+class MacBannerScanner(BannerScanner):
+    BANNER_PATTERN = rb"Darwin Kernel Version [0-9]+\.[0-9]+\.[0-9]+[^\x00]+"
--- a/volatility3/framework/automagic/symbol_finder.py
+++ b/volatility3/framework/automagic/symbol_finder.py
@@ -4,12 +4,11 @@
 
 import logging
 import os
-from typing import Any, Callable, Iterable, List, Optional, Tuple
+from typing import Callable, List, Optional, Tuple
 
 from volatility3.framework import constants, interfaces, layers
 from volatility3.framework.automagic import symbol_cache, banner_scanners
 from volatility3.framework.configuration import requirements
-from volatility3.framework.layers import scanners
 
 vollog = logging.getLogger(__name__)
 
@@ -40,7 +39,22 @@ class SymbolFinder(interfaces.automagic.AutomagicInterface):
                 name="SQLiteCache",
                 component=symbol_cache.SqliteCache,
                 version=(1, 0, 0),
-            )
+            ),
+            requirements.VersionRequirement(
+                name="banner_scanners_bannerscanner",
+                component=banner_scanners.BannerScanner,
+                version=(1, 0, 0),
+            ),
+            requirements.VersionRequirement(
+                name="banner_scanners_linuxbannerscanner",
+                component=banner_scanners.LinuxBannerScanner,
+                version=(1, 0, 0),
+            ),
+            requirements.VersionRequirement(
+                name="banner_scanners_macbannerscanner",
+                component=banner_scanners.MacBannerScanner,
+                version=(1, 0, 0),
+            ),
         ]
 
     @property
@@ -128,7 +142,7 @@ class SymbolFinder(interfaces.automagic.AutomagicInterface):
         if layer.config.get(self.banner_config_key, None):
             banner_list = [
                 (0, bytes(layer.config[self.banner_config_key], "raw_unicode_escape"))
-            ]  # type: Iterable[Any]
+            ]
         else:
             # Swap to the physical layer for scanning
             # Only traverse down a layer if it's an intel layer
--- a/volatility3/framework/automagic/symbol_finder.py
+++ b/volatility3/framework/automagic/symbol_finder.py
@@ -7,8 +7,9 @@ import os
 from typing import Callable, List, Optional, Tuple
 
 from volatility3.framework import constants, interfaces, layers
-from volatility3.framework.automagic import symbol_cache, banner_scanners
+from volatility3.framework.automagic import symbol_cache
 from volatility3.framework.configuration import requirements
+from volatility3.framework.layers import scanners
 
 vollog = logging.getLogger(__name__)
 
@@ -41,18 +42,8 @@ class SymbolFinder(interfaces.automagic.AutomagicInterface):
                 version=(1, 0, 0),
             ),
             requirements.VersionRequirement(
-                name="banner_scanners_bannerscanner",
-                component=banner_scanners.BannerScanner,
-                version=(1, 0, 0),
-            ),
-            requirements.VersionRequirement(
-                name="banner_scanners_linuxbannerscanner",
-                component=banner_scanners.LinuxBannerScanner,
-                version=(1, 0, 0),
-            ),
-            requirements.VersionRequirement(
-                name="banner_scanners_macbannerscanner",
-                component=banner_scanners.MacBannerScanner,
+                name="multi_string_scanner",
+                component=scanners.MultiStringScanner,
                 version=(1, 0, 0),
             ),
         ]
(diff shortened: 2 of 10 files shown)


Abyss-W4tcher:
TL;DR We need to ensure all identifiers/banners have the same ending characters.

---

I've identified a mistake in my banners script that was appending "\x00\n" instead of "\n\x00" to the end of b64 encoded banners:

https://github.com/Abyss-W4tcher/volatility3-symbols/blob/master/banners/linker.py#L77

vol3 actually strips remote ISF identifiers because of that mistake:

https://github.com/volatilityfoundation/volatility3/blob/c2db6ae5dea7a0453450ba40e0b26c66a0fb8a5c/volatility3/framework/automagic/symbol_cache.py#L450-L453

The "linux_banner" suffix should be "\n\x00" (kernel source: https://elixir.bootlin.com/linux/v2.6.20/source/init/version.c#L38), so any identifier that doesn't match this format should be considered malformed. 

I've created a new branch that should be both compatible with this PR and previous vol3 versions: https://github.com/Abyss-W4tcher/volatility3-symbols/blob/linker_banner_suffix_fix/banners/banners.json. This should ensure that:

- symbol_cache -> inserts "Linux version ...\n\x00" banners (as long as the local ISF was produced by dwarf2json OR the remote ISF banners.json was correctly formatted)
- banner_scanners.py -> outputs "Linux version ...\n\x00" banners
- banner lookups should be correctly normalized: https://github.com/volatilityfoundation/volatility3/blob/c2db6ae5dea7a0453450ba40e0b26c66a0fb8a5c/volatility3/framework/automagic/linux.py#L55

---

There is this banner in the https://github.com/volatilityfoundation/volatility3-test-data/releases/download/v0.0.1/linux.zip symbol set that does not end with \x00:

> Linux version 2.6.18-8.1.15.el5 (mockbuild@builder6.centos.org) (gcc version 4.1.1 20070105 (Red Hat 4.1.1-52)) #1 SMP Mon Oct 22 08:32:28 EDT 2007\n

**So in the end we should maybe just get rid of "\n" and "\x00" at the end of banners everywhere we manipulate them. The tricky part is for existing cache files that won't be manually re-generated after this PR gets merged...**

Abyss-W4tcher:
@ikelos do you think bumping the sqlite schema version to trigger these lines:

https://github.com/volatilityfoundation/volatility3/blob/c2db6ae5dea7a0453450ba40e0b26c66a0fb8a5c/volatility3/framework/automagic/symbol_cache.py#L207-L213

would be a proper way to trigger a cache reset without any user interaction? We could also add a more explicit "cache_version" key to force cache reset whenever we need to when users switch to a newer vol3 version?

volatility3/framework/automagic/symbol_finder.py
```diff
@@ -127,23 +136,30 @@ def _banner_scan(
...
+            ]
         else:
             # Swap to the physical layer for scanning
             # Only traverse down a layer if it's an intel layer
             # TODO: Fix this so it works for layers other than just Intel
-            if isinstance(layer, layers.intel.Intel):
-                layer = context.layers[layer.config["memory_layer"]]
-            banner_list = layer.scan(
-                context=context, scanner=mss, progress_callback=progress_callback
+            if self.operating_system == "linux":
```
ikelos: I don't think we should be differentiating here, even if we know the requested OS.

ikelos:
So, there's a lot going on in here, and I'm slightly worried there's a semi-significant flaw in this approach, but I think there's a compromise that could restore accuracy and still lower memory usage.

At the moment, this tries to cram `banners` and the core linux symbol finding code together into one place.  They were in fact different, banner was quick and design to catch things that "look like" a kernel banner.  The symbol finding code needed to locate and match precise strings within the image.  The implementation here searches for known fixes prefixes, and then takes as along a string as it can get and verifies that against the possible banners (as best as I can tell).  It's very quick, but it's not very accurate.  A user could provide a custom kernel with a banner that is nothing like the expected traditional banner, and volatility wouldn't be able to find it no matter how much a user tinkered with their isf file.  That's not a worthwhile tradeoff in my view.

However, the memory usage *is* an issue, so we could add a parameter to the MultiStringScanner that caps the maximum depth of the try (something 40 deep say?).  When when it finds one it can do like this plugin does and read out a chunk of data (I'd suggest asking the symbol cache for the longest possible length, rather than a fixed 0xFFF value) and go through and verify it again all possible matching banners (again, if the cache needs modifying to support that, it's doable).  That would ensure full matching, reduce memory usage and maintain the accuracy/flexibility without having to hardcode lots of little specific values in the codebase.  I think also we could choose a more efficient data structure, dicts of dicts can probably be simplified given we know the types it will contain (at the moment, python allocates expandable dictionaries to contain any kind of object).

As to changing the cached values, they should all have an expiry date and if the symbols are updated to have a different banner they should get reindexed I believe, and if the banner hasn't change then there shouldn't be a problem?  I'm still not sure that the sanitization code we've got requires fixing?  If a symbol table has been used, it may be locally cached, but should only have been locally cached if it was loaded, meaning it matched and it should be fine to stay as is.  I think it just means that the library needs regenerating/fixing and then pushing out (at a new path perhaps, so versioning on the download)?  I don't really want to bump to version 29 just for this, or tinker with things that are deeply used and functional at the moment, so I'd need a clearer picture of what volatility is doing *wrong* (if anything) and ideally a backwards compatible way of fixing it.  If our existing ISF library is broken, we should fix the library and let the old caches expire.

Like I said, there's a lot going on in here, so let me distil down the main points:

* The PR isn't as accurate as the existing method, but there is a hybrid that could work and solve the original problem.
* There is a caching issue that affects some ISF files, somewhere, and we need a clear description of exactly which files are affected, where they might live (local cache, internet, etc) and whether they'll time out of their own accord.

halpomeranz:
It's taken me a while to get to this, and I apologize for that. I tried the current version of this PR against the memory image in my original report and it works great. The plugin runs rapidly without significant memory usage and produces a working tar.gz file. So mark this as "fixed" from my perspective. Thank you!

Abyss-W4tcher:
> A user could provide a custom kernel with a banner that is nothing like the expected traditional banner

FYI the kernel explicitly prohibits this (https://elixir.bootlin.com/linux/v2.6.20/source/init/version.c#L38):

```c
/* FIXED STRINGS! Don't touch! */
const char linux_banner[] =
	"Linux version " UTS_RELEASE " (" LINUX_COMPILE_BY "@"
	LINUX_COMPILE_HOST ") (" LINUX_COMPILER ") " UTS_VERSION "\n";
```

If the format is different, the kernel is most likely to be custom anyway and fail elsewhere (https://github.com/volatilityfoundation/volatility3/issues/215 lynxos).

> The symbol finding code needed to locate and match precise strings within the image. 

The "else" branch:

https://github.com/volatilityfoundation/volatility3/blob/6b497259a99e73b5747e03fdba40d5132e1bad8e/volatility3/framework/automagic/symbol_finder.py#L135-L139

shouldn't be taken anyway as automagics set the "banner_config_key" so we can keep the old way I think. Simply moving the "mss" declaration into this else branch will prevent it from instantiating when not needed.

> However, the memory usage is an issue, so we could add a parameter to the MultiStringScanner that caps the maximum depth of the try (something 40 deep say?).

I've tried this and it looks like a good compromise. Here's a pslist with VMCore layer disabled (memory is fine with this method) and using a remote ISF server without a cap on mss:

image

---
And now with a cap:

image

---
*The highest usage should be the in-memory ISF profile, not sure if we can do something about that.*

This surely adds memory overhead compared to scanning for a generic string and then performing a dict lookup, but it has the merit of keeping the existing scanning structure. However, as we're not matching the full banner from the ISF due to the cap we still need to ensure that trailing characters are stripped.

As I mentioned in my previous message ISF banners can either end with "\n" or "\n\x00". That's not something wrong with vol3's symbol set, but the data source I think. 

---

I think we should keep the "banner_scanners" interface as it unifies the static values and post-processing in one place, even if it's only used by the "banners" plugin for now.

Abyss-W4tcher:
I've pushed a commit that switches to a capped mss

ikelos:
Ok, 

> FYI the kernel explicitly prohibits this (https://elixir.bootlin.com/linux/v2.6.20/source/init/version.c#L38):

And yet, by kernel version 7.0, it's become [parameterized](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/init/version.c#n36).  It is more brittle, and the changes made to the version.c show that it can't really be relied upon as well as proper string matching.  That might still be ok, but it is a down side, and I just worry we're throwing out perfectly good functionality because of something inefficient in the way we're doing things.

Other options to explore include, given we cache all possible banners we know about in the cache database now, building the trie only on database update.  That should vastly reduce the time and memory of a normal run (we can either cache the trie structure to make new additions quicker, or we can just cache the resultant regex string).  That way you get the accuracy of a proper string match, with the speed and memory usage of a normal regex search.

The other changes look fine, if you put them in a different PR we can get those merged more quickly.

ikelos:
I'm still not sure I agree that banners ("find anything that looks like a banner") should use the same code as "find a banner that volatility supports".  One is designed to be fuzzy and catch unknowns and other one really shouldn't allow unknowns through.  I'm happy to have the banner scanner separate, and possibly parameterize with a mode (fuzzy/not fuzzy)?  But I don't think both types of scanning could suffice for the two different use cases we have.

volatility3/framework/layers/scanners/__init__.py
```diff
@@ -61,14 +61,14 @@ def __call__(self, data: bytes, data_offset: int) -> Generator[int, None, None]:
 class MultiStringScanner(layers.ScannerInterface):
     thread_safe = True
 
-    _version = (1, 0, 0)
+    _version = (1, 0, 1)
```
ikelos, line 64: This needs to be 1.1.0, rather than 1.0.1, since it's an addition to the way the plugin is usable.  The third digit isn't used in version checks, so one that needed the extra param couldn't specify that it required it.

volatility3/framework/automagic/banner_scanners.py
```diff
@@ -0,0 +1,52 @@
+from typing import Iterator, Optional, Tuple
+from volatility3.framework.layers import scanners
+
+VALID_BANNER_CHARSET = (
+    b" #()+,;/-.0123456789:@ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz~\n"
```
ikelos, line 5: This feels like it should be in constants.linux, rather than in this file?  I might also consider using `string.printable - ["\t\n\ra\x0b\x0c"]` or similar...

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -1615,7 +1615,9 @@ def to_list(
         vmlinux = linux.LinuxUtilities.get_module_from_volobj_type(self._context, self)
 
         current = self.first
-        while current and current.is_readable():
+        seen = set()
```
ikelos, line 1618: Sneaking other changes into a PR isn't really good practice.  It'd be much better to put this in as a small targetted PR, with its own description to describe what the change is and what it does, rather than tacking it into a large banner rework.  Fine this time, but please avoid it in the future...

volatility3/framework/automagic/banner_scanners.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __call__(self, data: bytes, data_offset: int) -> Iterator[Tuple[int, bytes]]:
+        for off in super().__call__(data, data_offset):
+            banner = self._get_valid_banner(off)
+            if banner is not None:
+                yield off, banner
+
+
+class LinuxBannerScanner(BannerScanner):
+    BANNER_PATTERN = rb"Linux version [0-9]+\.[0-9]+\.[0-9]+[^\x00]+"
```
ikelos, line 48: Reasonable to put them in here, but only because they're the only thing differentiating the class from the base class, and making the user go hunt in the constants file would be a lot.  Still concerned that the pattern isn't accurate enough, but since we're only using it for the banners plugin, I'm ok with that.

volatility3/framework/interfaces/objects.py
```diff
@@ -79,7 +79,7 @@ def __getitem__(self, key):
         raise KeyError(f"No {key} present in ObjectInformation")
 
     def __contains__(self, key):
-        return key in [field.name for field in dataclasses.fields(self)]
+        return any(field.name == key for field in dataclasses.fields(self))
```
ikelos, line 82: This change also looks like it's unrelated.  Should be fine, but why are we making it?  Does it provide a signficant optimization?  What's the value in this change specifically?  Given it's in pretty fundamental code, I'd want there to at least some small speed improvement if we're bothering to change it...

ikelos (commented):
Thanks very much for this @Abyss-W4tcher !  There's a lot going on in it, and I'd somewhat prefer if the easy one liner fixes (like the `to_list` change in `linux/extensions/__init__.py` which doesn't seem overly related to all the other changes) could be in their own tiny PRs, so they can be dealt with separately and quickly.

I much happier this doesn't seem to impact the main linux automagic so much any more (the cap-then-verify method seems a good compromise).  Thinking it over more, the banners plugin and the core symbol finding code have two related, but fundamentally different goals.  Banners is just supposed to look for anything that looks like a banner and it doesn't matter that much if it misses things or doesn't give you exactly the right banner string, it just needs to give the user a quick pass over an image to see what kind of banners might be present.  The symbol finding code has to be precise and accurate looking for matches against data it already has.  I think BannerScanner is fine for the quick pass.

I'm worried that the 7.0 kernel at least (possibly earlier) allow for non-standard prefixes, which you may want to consider for the bannerscanner, but should work ok with the cap-then-verify method.  I'd still like to investigate the "cache a capped regex value when the cache changes" idea, since we don't need to build the regex every run, if we already know what all the JSON files are?  It should then reduce it to a simple regex scan, which should be very fast (although changing symbol paths and things may cause issues since we don't change the cache in those instances anymore)...

Anyway, as I say, this version's ok with a few questions, but only because it should still be accurate for the automagic.

PR #1981, opened 2026-05-01
Rework where the banner output happens

### LINKED ISSUE #1979: Vol.py doesn't output the volatility banner/version for --help or other argparse results.
Vol should always output the volatility banner unless producing pipeable output.  Volshell does already, investigate when this went away and if there was a reason, and if not, restore it.

ikelos:
Ok, I think this sorts it, but it could do with testing to make sure pluggable renderers (such as arrow/parquet) get picked up and to make sure that structured output (such as csv format, etc) display the banner when being piped, but don't write it to the file (stderr, rather than stdout)...

Any chance you could check it out for me please @eve-mem ?

Closes #1979

### DIFF
--- a/volatility3/cli/__init__.py
+++ b/volatility3/cli/__init__.py
@@ -30,11 +30,10 @@ try:
 except ImportError:
     HAS_ARGCOMPLETE = False
 
-from volatility3.cli import text_filter
 import volatility3.plugins
 import volatility3.symbols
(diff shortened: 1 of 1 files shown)


eve-mem:
Yup that seems to work. Shows me the banner when using help and when running a plugin against a mem dump. <3

PR #1982, opened 2026-05-02
Add Windows mini kernel dump support

Kreijstal:
- New WindowsMiniKernelDump64Layer for parsing mini kernel dumps (PAGEDU64)
- ISF symbol table (crash_mini.json) for the secondary dump header
- MiniKernelDumpMixin shared by all mini-dump plugins
- Plugins: crashcontext, crashmodules, crashstack (work without kernel symbols)
- crashinfo: recognize dump type 0x4 (mini kernel dump)
- pdbscan: scan WindowsMiniKernelDump64Layer for kernel PDB signatures

### DIFF
--- a/volatility3/framework/automagic/pdbscan.py
+++ b/volatility3/framework/automagic/pdbscan.py
@@ -16,7 +16,7 @@ from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, Un
 
 from volatility3.framework import constants, exceptions, interfaces, layers
 from volatility3.framework.configuration import requirements
-from volatility3.framework.layers import intel, scanners
+from volatility3.framework.layers import crash, intel, scanners
 from volatility3.framework.symbols import native
 from volatility3.framework.symbols.windows import pdbutil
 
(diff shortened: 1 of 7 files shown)


ikelos:
Thanks very much for your contribution and sorry it's taken us so long to look into it.  This feels like quite a niche area, and if it could be totally segmented (so that crash plugins only ever ran against crash files, which the plugins ensure) and that normal plugins don't ever run against the crash dumps (which currently doesn't look like it can be ensured) then we're unlikely to ever offer support for windows mini-dumps in volatility.  The problem is the difficulty we already have in supporting people who have images that aren't detected properly, or who don't understand the error messages that volatility throws.  Given that a mini-dump won't contain all the processes on the system, explaining the difference in dump format would likely take more support than the already minimal amount we can provide...

However, you've done a lot of good work here, so have you considered either making a standalone tool that uses volatility as a library (or monkeypatches into it), or a volshell snippet or similar?  Volshell should bring you into a bootstrapped vol environment having just been given a random data blob as a layer, so you could then construct the config manually (using the file passed in as the single_location) and so sidestep the changes you had to make in the automagic?

PR #1988, opened 2026-05-25
Fix: missing f-string prefix on save-config debug log

digitalisx:
Hi team,
The debug log emitted when writing out `--save-config` previously logged the literal string "{args.save_config}" instead of the resolved filename, in both vol.py and volshell. 
Add the f-string prefix so the actual destination path appears in the log.


ikelos (approved):
Thanks very much, good spot!  5:)

PR #1991, opened 2026-06-07
Add LA57 (5-level paging) support for Linux & Windows Intel layers

ricardojrdez:
Five-level paging (LA57) widens the linear address space to 57 bits by adding a fifth page map level (PML5). The existing Intel32e layer is hardwired to 48-bit virtual addresses and a 4-entry page-table walk, so on LA57 images the 48-bit address_mask truncates canonical kernel pointers and the page walk treats the PML5 as a PML4. The visible symptom is that physical scanning (psscan) works while virtual list-following (pslist, check_syscall, ...) silently returns nothing.

Add Intel32e_LA57 with _maxvirtaddr=57 and a 5-entry _structure; all derived values (maximum_address, address_mask, canonicalization, translation walk) follow from those two constants. Add the matching WindowsIntel32e_LA57 (transition-PFN workaround) and LinuxIntel32e_LA57 (_maxphyaddr=52, as 5-level kernels always use a 52-bit physical mask).

Detect LA57 at runtime in the Linux automagic: read __pgtable_l5_enabled from the physical layer in the banner stacker, and use NUMBER(pgtable_l5_enabled) in the VMCOREINFO stacker. Detection is runtime, not build-time, since a CONFIG_X86_5LEVEL kernel still boots in 4-level mode on non-LA57 hardware.

Add synthetic-page-table unit tests in test/layers/test_intel.py covering the 5-level walk (4K/2M/1G), 57-bit canonicalization, the pointer-masking regression, and the Windows/Linux entry variants.

Validated against a real LA57 LiME capture (Ubuntu 24.04, 6.8.0-117, Diamorphine rootkit): pslist now lists processes, hidden_modules reveals the rootkit and check_syscall recovers the hijacked entries.

### DIFF
--- a/test/automagic/test_windows.py
+++ b/test/automagic/test_windows.py
@@ -12,7 +12,7 @@ import struct
 
 from volatility3.framework import constants, contexts
 from volatility3.framework.automagic import windows
-from volatility3.framework.layers import physical
+from volatility3.framework.layers import intel, physical
 
 STUB_OFFSET = 0x2000
 CR3_VALUE = 0x1AD000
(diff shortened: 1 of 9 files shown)


ikelos:
#1939 might also be related.  My concern is that it only supports linux at the moment, and it will need carefully checking to ensure it's extensible enough to support Windows if/when windows eventually moves to 5-level paging.

ricardojrdez:
I've already extended to Windows in a related commit. For testing I used a synthetic dump as Windows is not yet supporting LA57

Issue #1985, opened 2026-05-21
My memory dump parses successfully with version 2.27.0, but not with version 2.28.0


ArtyomAfanasov:
Hello

**Describe the bug**
My memory dump parses successfully with version **2.27.0**, but not with version **2.28.0**. I used both the `vol.exe` and the `vol.py` files

**Context**
Volatility Version:  2.28.0
Operating System:  Windows 11 25H2 26200.8457
Python Version:  3.13.0
Suspected Operating System:  ?
Dump size: 53680537**6** bytes
Commands:  
* python vol.py -vvv -f "path\dump.mem" windows.info
* vol.exe -vvv -f "path\dump.mem" windows.info

**To Reproduce**

1. Use commands:
`python vol.py -vvv -f "path\dump.mem" windows.info`
2. See info (kind of error)
```
INFO     volatility3.cli: Volatility plugins path: ['path\\volatility3-2.28.0\\volatility3\\plugins', 'path\\volatility3-2.28.0\\volatility3\\framework\\plugins']
INFO     volatility3.cli: Volatility symbols path: ['path\\volatility3-2.28.0\\volatility3\\symbols', 'path\\volatility3-2.28.0\\volatility3\\framework\\symbols', 'path\\volatility3\\symbols']
DEBUG    volatility3.plugins.yarascan: Using yara-python module
DEBUG    volatility3.plugins.renderers.parquet_renderer: Arrow/Parquet libraries not found
Volatility 3 Framework 2.28.0
INFO     volatility3.framework.automagic: Detected a windows category plugin
INFO     volatility3.framework.automagic: Running automagic: ConstructionMagic
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Info.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Info.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Info.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Info.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Info.kernel
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Info.kernel.layer_name
[...]

Issue #1944, opened 2026-02-24
Windows Modules Not Returning Results


jonpaulhemi:
**Describe the bug**
I was working on a CTF challenge with a Windows 10 memory dump. Some modules like windows.info worked, but most modules did not work. I checked out an older commit and it worked. The most recent working commit was on November 24, 2025 (a17281a2145f5aa353fcccc35f09cfcd40ad0aa4).

**Context**
Volatility Version:  2.27.0
Operating System:  Linux
Python Version:  3.12.3
Suspected Operating System:  Windows 10 (From windows.info: Major/Minor 15.10240)
Command:  `vol -f /tmp/memdump.mem windows.pslist`

**To Reproduce**

1. Use command `vol -f /tmp/memdump.mem windows.*`
2. Below, you can see expected behavior from November 24, 2025 commit versus example (failed) output from November 30, 2025 commit.

**Expected behavior**
Success should provide a list of processes from the memory dump, as shown here.
`git checkout a17281a2145f5aa353fcccc35f09cfcd40ad0aa4`
`vol -vvvvv -f /tmp/memdump.mem windows.pslist`
```Bash
INFO     volatility3.cli: Volatility plugins path: ['/home/remnux/python-venvs/volatility3/volatility3/plugins', '/home/remnux/python-venvs/volatility3/volatility3/framework/plugins']
INFO     volatility3.cli: Volatility symbols path: ['/home/remnux/python-venvs/volatility3/volatility3/symbols', '/home/remnux/python-venvs/volatility3/volatility3/framework/symbols']
DEBUG    volatility3.plugins.yarascan: Using yara-python module
DEBUG    volatility3.plugins.renderers.parquet_renderer: Arrow/Parquet libraries not found
Volatility 3 Framework 2.27.0
DETAIL 3 volatility3.cli: Cache directory used: /home/remnux/.cache/volatility3
INFO     volatility3.framework.automagic: Detected a windows category plugin
INFO     volatility3.framework.automagic: Running automagic: ConstructionMagic
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.PsList.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.PsList.kernel.layer_name
[...]

Issue #1925, opened 2025-12-25
Unable to validate the plugin requirements: ['plugins.Info.kernel.layer_name', 'plugins.Info.kernel.symbol_table_name']


e888llll:
When i run vol.py, it fails. I have tried copying symbols.zip to symbols diretory, but it still fails.
(python3_11) D:\downloads\volatility3-develop\volatility3-develop>python vol.py -vvv --cache-path .\1 -f M:\1.raw windows.info
INFO     volatility3.cli: Volatility plugins path: ['D:\\downloads\\volatility3-develop\\volatility3-develop\\volatility3\\plugins', 'D:\\downloads\\volatility3-develop\\volatility3-develop\\volatility3\\framework\\plugins']
INFO     volatility3.cli: Volatility symbols path: ['D:\\downloads\\volatility3-develop\\volatility3-develop\\volatility3\\symbols', 'D:\\downloads\\volatility3-develop\\volatility3-develop\\volatility3\\framework\\symbols']
DEBUG    volatility3.plugins.yarascan: Using yara-python module
DEBUG    volatility3.plugins.renderers.parquet_renderer: Arrow/Parquet libraries not found
Volatility 3 Framework 2.27.1
INFO     volatility3.framework.automagic: Detected a windows category plugin
INFO     volatility3.framework.automagic: Running automagic: ConstructionMagic
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Info.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Info.kernel.symbol_table_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Info.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: Symbol table requirement not yet fulfilled: plugins.Info.kernel.symbol_table_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Info.kernel
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Info.kernel.layer_name
DETAIL 1 volatility3.framework.automagic.construct_layers: Failed on requirement: plugins.Info.kernel.layer_name
DETAIL 1 volatility3.framework.configuration.requirements: IndexError - No configuration provided: plugins.Info.kernel.layer_name
[...]

Issue #1872, opened 2025-09-04
Windows 11


DFI-SheepDog:
Volatility 3 Framework 2.26.2
Progress:  100.00               PDB scanning finished
Unsatisfied requirement plugins.Info.kernel.layer_name:
Unsatisfied requirement plugins.Info.kernel.symbol_table_name:

A translation layer requirement was not fulfilled.  Please verify that:
        A file was provided to create this layer (by -f, --single-location or by config)
        The file exists and is readable
        The file is a valid memory image and was acquired cleanly

A symbol table requirement was not fulfilled.  Please verify that:
        The associated translation layer requirement was fulfilled
        You have the correct symbol file for the requirement
        The symbol file is under the correct directory or zip file
        The symbol file is named appropriately or contains the correct banner

Unable to validate the plugin requirements: ['plugins.Info.kernel.layer_name', 'plugins.Info.kernel.symbol_table_name']

Any Assistance would be appreciated.  Windows 11 Profile.

Carryouu:
Hello,
were you able to find a solution ? For me it worked but it might be because it wasn t the last build of windows 11 even it still a windows 11.
Let me know if it worked for you :)
thanks a lot in advance.
Best regards,
Mehdi.

Tylereck81:
Hello, 

I am currently doing a project with windows 11 live kernel dumps. I am running into the same problem. Have you found a solution? 

- Tyler

ikelos:
For people having difficulty with certain Windows 11 images, there is now a pull request in testing to try to resolve the issue.  If you know how, we'd much appreciate people testing the #1929 pull request and letting us know whether it now improves the situation.

For a more detailed reason as to why this works, a heuristic was used to find the "map" that explains how the memory is laid out.  In windows, one particular part of that map always points to itself, and we used this fact to quickly identify it.  The map also always tended to live in a particular location.  Some recent images have shown that Microsoft now stores the map in a location that can fall outside the area we were previously checking, and there can now be a false map which points to itself, but contains next to no other entries, right where we'd expect the real map to be.  For this reason, volatility was detecting that map and trying to use it, which led to it being unable to detect Windows on those images.

[...]

PR #1992, opened 2026-06-10
Windows: verify Intel layers using translation checks

Abyss-W4tcher:
Hi,

https://github.com/volatilityfoundation/volatility3/pull/1929 introduced a new DTB sanity check to filter out broken DTBs on recent Windows versions. While this fix worked in some cases, it also introduced regressions:

- refs #1944, #1925, #1872, more recently: #1985
- AOMF windows/sample_003.bin cannot be analyzed

Checking if a DTB has a certain number of valid pointers to be considered valid can lead to false positives, and is probably not the way DTBs should be validated.

I've looked at vol2 source code and found out that Windows' Intel layers were validated using translation checks:

- https://github.com/volatilityfoundation/volatility/blob/328a178edeecaf8d2abf96618ecfcb14a7b0c860/volatility/plugins/overlays/windows/windows.py#L1078

These checks rely on the historical `_KUSER_SHARED_DATA` data structure address translated both in kernel and user space.

- [X] Mass testing: on 200 windows samples, all passed the layer stacking stage.

### DIFF
--- a/volatility3/framework/automagic/windows.py
+++ b/volatility3/framework/automagic/windows.py
@@ -40,7 +40,7 @@ vollog = logging.getLogger(__name__)
 
 class Intel32LayerCheck:
     # These addresses are at a fixed location:
-    # "The KUSER_SHARED_DATA structure is a single page (4096 bytes) in memory 
+    # "The KUSER_SHARED_DATA structure is a single page (4096 bytes) in memory
     # that is mapped at a fixed, hardcoded address in both kernel and user side of VAS."
     # See: https://www.microsoft.com/en-us/msrc/blog/2022/04/randomizing-the-kuser_shared_data-structure-on-windows
     KUSER_USER_SPACE_ADDR = 0x7FFE0000
--- a/volatility3/framework/automagic/windows.py
+++ b/volatility3/framework/automagic/windows.py
@@ -56,6 +56,7 @@ class Intel32LayerCheck:
             exceptions.PagedInvalidAddressException,
             exceptions.InvalidAddressException,
         ):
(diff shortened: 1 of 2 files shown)


Abyss-W4tcher:
This is ready for review

volatility3/framework/automagic/windows.py
```diff
@@ -29,28 +29,74 @@
...
-from volatility3.framework import constants, interfaces, layers
+from volatility3.framework import constants, exceptions, interfaces, layers
 from volatility3.framework.configuration import requirements
 from volatility3.framework.layers import intel
 
 vollog = logging.getLogger(__name__)
 
 
+class Intel32LayerCheck:
+    KUSER_USER_SPACE_ADDR = 0x7FFE0000
```
ikelos, line 46: Can we document where these values come from (and why they won't change/shift between versions)?

ikelos (commented):
Ok, I'm not a fan of using fixed addresses for things, but I guess in this case it's unavailable.

Since we're only doing a single check for the NT_MAJOR field, I think reading it direct is probably ok but if our tests get more complex this will need pulling out and doing more appropriately than just direct reads.

volatility3/framework/automagic/windows.py
Abyss-W4tcher: I've added a link to a Microsoft blogpost that mentions KUSER_SHARED_DATA's purpose and history.

Abyss-W4tcher:
I've addressed the comment and bumped the framework patch version, to help debugging user issues later.

ikelos (approved):
Ok, looks good, thanks!  5:)

PR #1993, opened 2026-06-13
Fix PAE DTB detection rejecting valid 4-entry PDPTs

ricardojrdez:
On PAE 32-bit Windows images the WindowsIntelStacker discarded the correct WindowsIntelPAE DTB and fell back to a false-positive non-PAE DTB, leaving list-walking plugins (e.g. windows.pslist) empty while pool scanners still worked.

Both DTB candidates are found and the PAE one sorts first, but page_table_is_dummy() required at least 10 valid pointers. A PAE top-level table is the PDPT pointed to by CR3, which legitimately holds only four entries, so the valid PAE DTB was rejected as a dummy table and the stacker fell through to the bogus non-PAE candidate.

Make the dummy-table threshold PAE-aware (4 entries for WindowsIntelPAE, 10 otherwise). The x64/non-PAE detection path is unchanged.

Validated on a PAE XP image (windows.pslist now lists all processes with IsPAE=True) and regression-checked on an x64 image (still WindowsIntel32e).

### DIFF
--- a/volatility3/framework/automagic/windows.py
+++ b/volatility3/framework/automagic/windows.py
@@ -317,13 +317,12 @@ class WindowsIntelStacker(interfaces.automagic.StackerLayerInterface):
                     )
                 return max_ptr
 
-            def page_table_is_dummy(page_table, ptr_size: int):
-                """Verify that a page table has at least 12 valid pointers"""
+            def page_table_is_dummy(page_table, ptr_size: int, minimum: int = 10):
+                """Verify that a page table has at least `minimum` valid pointers"""
                 valid_pointers = 0
                 for _ in get_valid_page_table_pointers(page_table, ptr_size):
                     valid_pointers += 1
-                    # 10 is an arbitrary constant
-                    if valid_pointers >= 10:
+                    if valid_pointers >= minimum:
                         # Do not consume the entire generator to enhance performance
                         return False
                 vollog.debug(f"Found {valid_pointers} valid pointers")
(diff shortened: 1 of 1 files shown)


Abyss-W4tcher:
Hi @ricardojrdez, we've been working on improving this check by not relying on an arbitrary number of pointers. Could you please try out this branch: https://github.com/volatilityfoundation/volatility3/pull/1992 against your sample and tell us if it solves the issue? Thanks

ricardojrdez:
Hey @Abyss-W4tcher , thanks for the alternative approach. I tested #1992 against both of my samples and it works great.

* PAE x86 (the original failing case, a WinXP SP2, ntkrnlpa.pdb BD8F451F…-1):
  - DtbSelfRefPae now succeeds at 0x319000 (the correct PAE DTB)
  - DTB was found at: 0x319000
  - windows.pslist -> 25 processes -> OK!

  Previously this picked a false-positive non-PAE DTB at 0x39000 because the PAE top-level table (PDPT) only has 4 entries and was rejected by the old page_table_is_dummy() >= 10-pointer heuristic. Your _KUSER_SHARED_DATA translation check handles it correctly without relying on a pointer count.

 * x64 regression (Windows, ntkrnl):
   - DtbSelfRef64bit succeeds at 0x1ae000 -> WindowsIntel32e
   - windows.pslist -> 220 processes -> OK!

So no regression on the 64-bit path either. The translation-check approach is a much cleaner fix than counting pointers. I'm happy to close #1993 in favor of this.

Abyss-W4tcher:
Nice, thank you for testing it! As everything worked as expected then closing it in favor of #1993 would be much appreciated yes 👍.

PR #2001, opened 2026-06-27
Feature/windows clipboard plugin

IB-Mustafa:
## Summary

Adds a prototype for Windows clipboard support to Volatility 3, addressing issue #710.

Volatility 2 included a clipboard plugin, but there is currently no equivalent in Volatility 3. This PR adds the required GUI structures together with an initial plugin implementation.

### New file: `volatility3/plugins/windows/clipboard.py`

* Enumerates clipboard entries from `tagWINDOWSTATION.pClipBase`
* Reports clipboard formats and their associated USER handles
* Includes placeholders for future handle resolution

### `volatility3/framework/symbols/windows/extensions/gui.py`

* Added `tagCLIP` with clipboard format name lookup
* Added `tagCLIPDATA` with helper methods for reading and decoding clipboard data
* Registered both types in `class_types`

### GUI JSON files (Windows 10)

* Added the missing `tagCLIPDATA` structure.
* Updated `pClipBase` in `tagWINDOWSTATION` to point to `tagCLIP` instead of a hardcoded array definition.

## Current state

This is an initial prototype.

The plugin can enumerate clipboard entries from a WindowStation. However, it does not yet recover the actual clipboard contents because `hData` is a USER handle rather than a direct pointer. Resolving it requires walking `tagSHAREDINFO.aheList`, which is left for future work.

## Testing

* Tested on a Windows 10 (19045) memory image.
* The plugin loaded and executed successfully without errors.
* Clipboard-related data was independently verified to be present in the memory image.
* As expected, clipboard contents are not yet displayed because USER handle resolution has not been implemented.

## Not yet supported

* Clipboard content recovery (requires resolving `hData` through `tagSHAREDINFO.aheList`)
* Additional clipboard formats beyond the current prototype

### DIFF
--- a/volatility3/framework/symbols/windows/extensions/gui.py
+++ b/volatility3/framework/symbols/windows/extensions/gui.py
@@ -304,7 +304,7 @@ class GUIExtensions(interfaces.configuration.VersionableInterface):
 
     # This is copy/paste from UNICODE_STRING in `symbols/windows/extensions/__init__.py`
     # The versioning of modules would get very ugly if we let different modules share implementations
-    # across different data structure
+    # across different data structures
     class LARGE_UNICODE_STRING(objects.StructType):
         """A class for Windows unicode string structures."""
 
@@ -324,36 +324,10 @@ class GUIExtensions(interfaces.configuration.VersionableInterface):
                 encoding="utf16",
             )
 
-    class tagCLIPDATA(objects.StructType):
-        """A class for clipboard data objects stored in Windows memory"""
-
-        def get_data(self) -> Optional[bytes]:
-            """Returns the raw clipboard data as bytes"""
-            try:
-                size = self.cbData
-                if size == 0 or size > 0x100000:
-                    return None
-                return bytes(bytearray(self.abData))
-            except exceptions.InvalidAddressException:
-                return None
-
-        def get_text(self, fmt: str) -> Optional[str]:
-            """Returns clipboard data as readable text if format is text-based"""
-            data = self.get_data()
-            if data is None:
-                return None
-            try:
-                if "UNICODE" in fmt:
-                    return data.decode("utf-16-le", errors="replace").rstrip("\x00")
-                else:
-                    return data.decode("latin-1", errors="replace").rstrip("\x00")
-            except Exception:
-                return None
-
     class tagCLIP(objects.StructType):
         """A class for clipboard format entries (one per clipboard item)"""
 
-        # Common Windows clipboard format constants
+        # TODO: Extend clipboard format mapping if needed.
         CF_FORMATS = {
             1: "CF_TEXT",
             2: "CF_BITMAP",
@@ -375,12 +349,52 @@ class GUIExtensions(interfaces.configuration.VersionableInterface):
             except exceptions.InvalidAddressException:
                 return "CF_UNKNOWN"
 
+    class tagCLIPDATA(objects.StructType):
+        """A class for clipboard data objects stored in Windows memory"""
+
+        def get_data(self) -> Optional[bytes]:
+            """Returns the raw clipboard data as bytes"""
+
+            # TODO: Verify abData consistency; future Windows may change tagCLIPDATA layout.
+            try:
+                size = int(self.cbData)
+
+                # TODO: Improve clipboard size validation if needed.
+                MAX_CLIPBOARD_SIZE = 50 * 1024 * 1024  # 50 MiB
+                if size == 0 or size > MAX_CLIPBOARD_SIZE:
+                    return None
+
+                layer = self._context.layers[self.vol.layer_name]
+                return layer.read(self.abData.vol.offset, size, pad=True)
+
+            except exceptions.InvalidAddressException:
+                return None
+
+        def get_text(self, fmt: str) -> Optional[str]:
+            """Returns clipboard data as readable text if format is text-based"""
+            data = self.get_data()
+
+            if data is None:
+                return None
+
+            # TODO: Extend support for more clipboard formats as needed.
+
+            if fmt == "CF_UNICODETEXT":
+                return data.decode("utf-16-le", errors="replace").rstrip("\x00")
+            elif fmt in ("CF_TEXT", "CF_OEMTEXT"):
+                return data.decode("latin-1", errors="replace").rstrip("\x00")
+
+            elif fmt == "CF_HDROP":
+                # File paths stored as null-separated UTF-16 list
+                return data.decode("utf-16-le", errors="replace").rstrip("\x00")
+            return None
 
     class_types = {
         "tagWINDOWSTATION": tagWINDOWSTATION,
         "tagDESKTOP": tagDESKTOP,
         "tagWND": tagWND,
         "_LARGE_UNICODE_STRING": LARGE_UNICODE_STRING,
+        # Clipboard plugin extensions
         "tagCLIPDATA": tagCLIPDATA,
         "tagCLIP": tagCLIP,
     }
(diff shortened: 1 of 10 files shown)


volatility3/framework/symbols/windows/gui/gui-win10-19041-x64.json
```diff
@@ -13696,18 +13696,14 @@
     "tagWINDOWSTATION": {
       "fields": {
         "pClipBase": {
+          "offset": 96,
           "type": {
+            "kind": "pointer",
             "subtype": {
```
AmirGhiassian, line 13702: Changing this from a pointer-to-`tagCLIP[104]` into a pointer-to-one-`tagCLIP` discards the known array bound and can break consumers that dereference and iterate the symbol (the approach used by the previous implementation). The new plugin does not require this schema regression—it can dereference the existing array and iterate at most `min(cNumClipFormats, len(array))`. Please preserve the original type in all eight JSON files.

volatility3/plugins/windows/clipboard.py
```diff
@@ -0,0 +1,135 @@
...
+            except exceptions.InvalidAddressException:
+                vollog.debug(f"Cannot read clipboard fields for station {station_name}")
+                continue
+
+            vollog.debug(
+                f"Station={station_name} Session={session_id} "
+                f"cNumClipFormats={clip_count} pClipBase={clip_base_ptr:#x}"
+            )
+
+            if clip_count == 0 or clip_count > 512:
```
AmirGhiassian, line 75: The symbols describe a 104-entry clipboard array, but this accepts counts through 512 and manually walks beyond that allocation. A corrupt count between 105 and 512 can therefore turn adjacent memory into apparently valid clipboard entries. Please retain/dereference the bounded symbol array and cap iteration to its actual length.

volatility3/framework/symbols/windows/extensions/gui.py
```diff
@@ -324,9 +324,77 @@ def get_string(self) -> interfaces.objects.ObjectInterface:
                 encoding="utf16",
             )
 
+    class tagCLIP(objects.StructType):
+        """A class for clipboard format entries (one per clipboard item)"""
+
+        # TODO: Extend clipboard format mapping if needed.
+        CF_FORMATS = {
```
AmirGhiassian, line 331: This duplicates the existing `fmtEnum` but omits valid standard formats such as CF_SYLK, CF_DIF, CF_TIFF, CF_PALETTE, CF_PENDATA, CF_RIFF, CF_WAVE, and the display formats, so they are incorrectly reported as unknown. Please use the symbol enum's name with a numeric fallback. Registered formats (0xC000–0xFFFF) should ideally be resolved via the session atom table or identified as registered rather than `CF_UNKNOWN`.

volatility3/framework/symbols/windows/extensions/gui.py
```diff
@@ -324,9 +324,77 @@ def get_string(self) -> interfaces.objects.ObjectInterface:
...
+            try:
+                size = int(self.cbData)
+
+                # TODO: Improve clipboard size validation if needed.
+                MAX_CLIPBOARD_SIZE = 50 * 1024 * 1024  # 50 MiB
+                if size == 0 or size > MAX_CLIPBOARD_SIZE:
+                    return None
+
+                layer = self._context.layers[self.vol.layer_name]
+                return layer.read(self.abData.vol.offset, size, pad=True)
```
AmirGhiassian, line 368: Using `pad=True` silently fabricates zero bytes for unreadable pages, which can make incomplete forensic evidence look valid. Please use an unpadded read and explicitly return/report unreadable or truncated data. Also, these helpers are currently unreachable because the plugin never resolves `hData` to a `tagCLIPDATA` object; consider deferring them until handle resolution is implemented.

volatility3/framework/symbols/windows/extensions/gui.py
```diff
@@ -324,9 +324,77 @@ def get_string(self) -> interfaces.objects.ObjectInterface:
...
+                return None
+
+            # TODO: Extend support for more clipboard formats as needed.
+
+            if fmt == "CF_UNICODETEXT":
+                return data.decode("utf-16-le", errors="replace").rstrip("\x00")
+            elif fmt in ("CF_TEXT", "CF_OEMTEXT"):
+                return data.decode("latin-1", errors="replace").rstrip("\x00")
+
+            elif fmt == "CF_HDROP":
```
AmirGhiassian, line 387: CF_HDROP cannot be decoded as UTF-16 from byte zero. Its payload starts with a DROPFILES header; `pFiles` gives the filename-list offset and `fWide` selects ANSI versus Unicode. Please parse and validate that header, slice at `pFiles`, then decode and split the double-NUL-terminated list.

volatility3/plugins/windows/clipboard.py
```diff
@@ -0,0 +1,135 @@
...
+    _required_framework_version = (2, 0, 0)
+    _version = (1, 0, 0)
+
+    @classmethod
+    def get_requirements(cls) -> List[interfaces.configuration.RequirementInterface]:
+        return [
+            requirements.ModuleRequirement(
+                name="kernel",
+                description="Windows kernel",
+                architectures=["Intel32", "Intel64"],
```
AmirGhiassian, line 36: This advertises Intel32, but the WindowStations dependency's `create_gui_table` explicitly raises `NotImplementedError` for non-64-bit symbol tables, and this PR adds only x64 GUI symbols. Please advertise only Intel64 unless x86 GUI symbols/support are added.

volatility3/plugins/windows/clipboard.py
```diff
@@ -0,0 +1,135 @@
...
+                    fmt_name = clip.get_format_name()
+                    handle_val = int(clip.hData)
+
+                    vollog.debug(f"  clip[{i}]: fmt={fmt_name} hData={handle_val:#x}")
+
+                    # hData is a USER handle — not a direct pointer.
+                    # Resolving it requires walking tagSHAREDINFO.aheList
+                    # which is not yet implemented (see TODO above).
+                    # For now we report format and handle without data.
+                    data_display: interfaces.renderers.BaseAbsentValue | str = (
```
AmirGhiassian, line 106: Because this is unconditional, the plugin never extracts clipboard data despite its name/docstring; it only lists formats and handles. Please either resolve the USER handle through the session's shared handle table and emit object/raw data, or rename/re-scope this prototype so users are not promised extraction.

AmirGhiassian (commented):
Thanks for starting this work. The format/handle enumeration is a useful foundation, but I don't think this is ready to merge as clipboard extraction yet.

The main blockers are the shared-symbol compatibility change, incorrect/unused decoding helpers, and advertised coverage that does not match the implementation. In addition to the inline comments:

- The PR currently enumerates metadata only; every Data cell is unavailable. Either implement USER-handle resolution through each session's `tagSHAREDINFO.aheList` (filtering `TYPE_CLIPDATA`) or rename/re-scope the plugin to clipboard-format enumeration and defer `tagCLIPDATA` support.
- Please add focused synthetic tests for invalid pointers, corrupt/oversized counts, complete format naming, unreadable/truncated data, ANSI/Unicode termination, `CF_HDROP`, and registered formats. An integration test against the stated Windows 10 image would also be valuable.
- Preserve raw clipboard bytes as forensic evidence and treat decoded text as an additional presentation field rather than replacing the raw value.

Validation performed: the plugin imports and appears in `vol.py -h`; all edited JSON files parse; `git diff --check` and the configurable-class analysis pass. The full pytest run could not be completed locally because the suite requires image-specific setup and then encountered an existing abstract parquet-test collection failure.

PR #2009, opened 2026-07-28
fix: linux 6.19 taint_flag.module removal

the-rectifier:
Linux 6.19 removed the `taint_flag.module` attribute: https://lore.kernel.org/all/20251022082938.26670-1-petr.pavlu@suse.com/T/#u

As such, when running `linux.lsmod` for a snapshot of linux 6.19+ the whole framework backtraces:

```bash
  File "/usr/lib/python3.14/site-packages/volatility3/framework/symbols/linux/utilities/tainting.py", line 94, in _module_flags_taint_post_4_10_rc1
    if is_module and not taint_flag.module:
                         ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/volatility3/framework/objects/__init__.py", line 993, in __getattr__
    raise AttributeError(
        f"{agg_name} has no attribute: {self.vol.type_name}.{attr}"
    )
AttributeError: StructType has no attribute: symbol_table_name1!taint_flag.module
```

Also, update the taint table, and introduce a new taint flag `J`:
https://docs.kernel.org/admin-guide/tainted-kernels.html#table-for-decoding-tainted-state

Taints continue to appear when running `linux.lsmod` on a random snapshot:

```bash
Volatility 3 Framework 2.28.2
Offset  Module Name     Code Size       Taints  Load Arguments  File Output
[...]
0xffffc06e6bc0  vboxsf  0x17000 OOT_MODULE,UNSIGNED_MODULE      follow_symlinks=None, disabled=None     N/A
[...]
0xffffc031b040  vboxguest       0x81000 OOT_MODULE,UNSIGNED_MODULE      r3_log_to_host=None, log_dest=None, log_flags=None, log=None, disabled=None     N/A
```

### DIFF
--- a/volatility3/framework/symbols/linux/utilities/tainting.py
+++ b/volatility3/framework/symbols/linux/utilities/tainting.py
@@ -14,7 +14,7 @@ class Tainting(interfaces.configuration.VersionableInterface):
         - kernel: print_tainted
     """
 
-    _version = (1, 0, 0)
+    _version = (1, 0, 1)
     _required_framework_version = (2, 0, 0)
 
     framework.require_interface_version(*_required_framework_version)
@@ -91,7 +91,8 @@ class Tainting(interfaces.configuration.VersionableInterface):
         for taint_bit, taint_flag in enumerate(
             cls._get_kernel_taint_flags_list(context, kernel_module_name)
         ):
-            if is_module and not taint_flag.module:
+            # https://lore.kernel.org/all/20251022082938.26670-1-petr.pavlu@suse.com/T/#u: "taint/module: Remove unnecessary taint_flag.module field"
+            if is_module and taint_flag.has_member("module") and not taint_flag.module:
                 continue
 
             try:
--- a/volatility3/framework/constants/linux/__init__.py
+++ b/volatility3/framework/constants/linux/__init__.py
@@ -487,7 +487,7 @@ TAINT_FLAGS = {
     "L": TaintFlag(shift=1 << 14, desc="SOFTLOCKUP", when_present=True, module=False),
     "K": TaintFlag(shift=1 << 15, desc="LIVEPATCH", when_present=True, module=True),
     "X": TaintFlag(shift=1 << 16, desc="AUX", when_present=True, module=True),
-    "T": TaintFlag(shift=1 << 17, desc="RANDSTRUCT", when_present=True, module=True),
+    "T": TaintFlag(shift=1 << 17, desc="RANDSTRUCT", when_present=True, module=False),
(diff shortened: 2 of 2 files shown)


volatility3/framework/symbols/linux/utilities/tainting.py
```diff
@@ -91,7 +91,8 @@ def _module_flags_taint_post_4_10_rc1(
         for taint_bit, taint_flag in enumerate(
             cls._get_kernel_taint_flags_list(context, kernel_module_name)
         ):
-            if is_module and not taint_flag.module:
+            # https://lore.kernel.org/all/20251022082938.26670-1-petr.pavlu@suse.com/T/#u: "taint/module: Remove unnecessary taint_flag.module field"
+            if is_module and taint_flag.has_member("module") and not taint_flag.module:
```
ikelos, line 95: Will this flag lots of false positives now, or was the taint_flag.module just a belt and braces check to reduce a few rare false positives?

the-rectifier: In my limited understanding, and based on the lore message, no, it shouldn't produce false positives.

The commit message claims that the per-module flags are always those added to `module.taints` by calls to [`add_taint_module()`](https://github.com/torvalds/linux/blob/663385f9155f27892a97a5824006f806a32eb8dc/kernel/module/main.c#L194).

Cross-checking those calls, they (8) exist inside the following:

- [`module_augment_kernel_taints()`](https://github.com/torvalds/linux/blob/663385f9155f27892a97a5824006f806a32eb8dc/kernel/module/main.c#L2570)
- [`module_license_taint_check()`](https://github.com/torvalds/linux/blob/663385f9155f27892a97a5824006f806a32eb8dc/kernel/module/main.c#L1766)
- [`try_to_force_load()`](https://github.com/torvalds/linux/blob/663385f9155f27892a97a5824006f806a32eb8dc/kernel/module/main.c#L1107)

Along with a single direct `set_bit()` in [`inherit_taint()`](https://github.com/torvalds/linux/blob/663385f9155f27892a97a5824006f806a32eb8dc/kernel/module/main.c#L1240)

The only flags that can reach `mod->taints` are `P F O C E K N` all of them already had the `.module = true`;


Actually, the error ran the other way around; the flags `T` and `J` were marked as per-module by mistake, which prompted the removal

the-rectifier:
While we're at it, fix RANDSTRUCT as well

Silly me, marked FWCTL as a per-module... :clown_face:

ikelos (approved):
Looks good, thanks.

PR #1959, opened 2026-03-10
Attempt to fix the local filepath not being detected correctly

ikelos:
Potentially fixes #1958 .

PR #1949, opened 2026-03-04
Major Perf PoolScan physical layer

forensicxlab:
This PR addresses major performance issues I am facing in Windows 10/11 memory images for poolscan-based plugins (for example windows.filescan).

Poolscan consumers currently scan the Win10/11 virtual kernel layer by default, which can traverse very large sparse address ranges and spend most time in translation/read overhead before yielding results.

We have added an optional scan_layer_name parameter to the following:
  - `PoolScanner.generate_pool_scan_extended`
  - `PoolScanner.generate_pool_scan`
We updated poolscan-based Windows plugins to pass `memory_layer` when available (fallback to original layer if not):

- [X] `windows.filescan.FileScan`
- [X] `windows.psscan.PsScan`
- [X] `windows.thrdscan.ThrdScan`
- [X] `windows.driverscan.DriverScan`
- [X] `windows.modscan.ModScan`
- [X] `windows.mutantscan.MutantScan`
- [X] `windows.symlinkscan.SymlinkScan`
- [X] `windows.netscan.NetScan`
- [X] `windows.callbacks.Callbacks`
- [X] `windows.windowstations.WindowStations`
- [X] `windows.registry.hivescan.HiveScan` (poolscan path).

  
 - windows.filescan
   previous behavior: timed out at 3600s in control run
   patched: finished in ~68.8s, 5666 results
- `windows.psscan`
  previous behavior: timed out at 3600s, 0 results
  patched: finished in ~63.6s, 286 results
  
  I can privately share a sample image to do test. Let me know by contacting me at felix.guyard@forensicxlab.com.
  
  Kind regards.

PR #2012, opened 2026-07-30
Discard modscan results whose base is not page aligned

### LINKED ISSUE #1633: Windows modules needs to enforce page aligned base or 0 base
We get many FPs from modscan due to missing checks. I have another big push (GUI initial plugins) that touches the same files, so I will save this until after to avoid ugly merge conflicts.

Hmkz0x00:
## Summary

Fixes #1633. `ModScan.scan_modules` yielded every pool match for the `MmLd` tag
without inspecting it, so the tag appearing inside unrelated kernel data
produced a module row. Results are now discarded unless the base is page
aligned, keeping a base of zero as the issue title asks for.

This reproduces on `win-10_19041-2025_03.dmp` from `volatility3-test-data`, so
no special sample is needed to see it.

## The false positive

One row on that image:

| field | value |
| --- | --- |
| Base | `0xccccccebebc0` |
| Size | `0x83485708` (2.2 GB) |
| Name | *unreadable* |
| Offset | `0xf80436391342` |

Three independent things are wrong with it: the base is not page aligned and
carries the `0xcc` fill pattern, the size is larger than the image, and the pool
offset is not 16 byte aligned. That offset also falls inside ntoskrnl's own
mapped range, so the literal bytes `MmLd` occurred in kernel code and the pool
scanner matched them.

It was also the only row on that image whose `BaseDllName` could not be read.

## Why page alignment

A loaded module is mapped on a page boundary, so a base with any of the low bits
set cannot describe a real allocation. A base of zero is kept, because that means
the base could not be determined rather than that the entry is bogus.

This mirrors how `driverscan.scan_drivers` already treats `DriverStart`:

```python
# Many/most rootkits zero out their DriverStart member for anti-forensics
# so we accept a driver start that is either 0 or is points into kernel memory
if mem_object.DriverStart == 0 or mem_object.DriverStart > kernel_space_start:
    yield mem_object
```

Same place in the code, same tolerance of zero, so this follows the existing
idiom rather than introducing a new one.

## Where the page size comes from

`page_size` is read from the kernel's own layer rather than from the layer the
matched object was found in. The kernel layer is always the Intel virtual layer,
whereas `FileLayer` has no `page_size` at all, so taking it from the object's
layer would break if scanning moves to the physical layer. That matters for
#1949, which does exactly that.

Please note #1949 also edits `scan_modules`, so one of these will need a small
rebase. Happy to be the one that moves.

## Effect

`windows.modscan` on `win-10_19041-2025_03.dmp`:

| | before | after |
| --- | --- | --- |
| rows | 177 | 176 |
| rows with an unaligned base | 1 | 0 |
| rows with an unreadable name | 1 | 0 |

The stronger check is against the list walk. `windows.modules` reports 176
modules on that image, and after this change the set of bases reported by
`modscan` is identical to it, with no entry on either side that the other
lacks. Before the change, the unaligned row was the only difference between the
two plugins.

On `win-xp-laptop-2005-06-25.img` the output is unchanged at 150 rows, with
nothing dropped or added, so the check is inert where there is nothing wrong to
find.

## Tests

Adds `test_windows_generic_modscan_bases_are_page_aligned` to
`test/plugins/windows/windows.py`. It takes the `image` fixture, so it runs
against the images CI already supplies, and it fails on `develop` and passes
with this change. I verified that by reverting only the plugin and re-running.

`ruff format`, `ruff check` and `test/volatility3_code_analysis.py` are all
clean.

## One note on running the existing tests

`test_windows_generic_modscan` fails on my machine, before and after this
change, because `runvol_plugin` passes the image path straight to
`--single-location`, which expects a URL. On Windows that becomes
`\\\\E:\Codes\...` and every image backed test fails with an unsatisfied layer
requirement. I confirmed the failure is unrelated by reproducing it on a clean
checkout, and worked around it locally by passing a `file://` URL. It looks like
the same thing #1959 is aimed at, so I have left it alone.

## Possible follow-up

`driverscan` additionally requires the address to point into kernel space via
`Modules.get_kernel_space_start`. The equivalent check would catch this same row
and probably others, but it is a wider behaviour change than the issue asks for,
so I have not included it. Happy to add it here or separately if you would like
it.

### DIFF
--- a/test/plugins/windows/windows.py
+++ b/test/plugins/windows/windows.py
@@ -1039,6 +1039,35 @@ class TestWindowsModScan:
         for expected_row in expected_rows:
             assert test_volatility.match_output_row(expected_row, json_out)
 
+    def test_windows_generic_modscan_bases_are_page_aligned(
+        self, volatility, python, image
+    ):
+        """Pool matches whose base is not page aligned are false positives.
+
+        A loaded module is always mapped on a page boundary, so a base with
+        any of the low twelve bits set means the "MmLd" tag was found in
+        unrelated data rather than in a real pool allocation. A base of zero
+        is allowed, since that means the field could not be read.
+        """
+        rc, out, _err = test_volatility.runvol_plugin(
+            "windows.modscan.ModScan",
+            image,
+            volatility,
+            python,
+            globalargs=("-r", "json"),
+        )
+        assert rc == 0
+        json_out = json.loads(out)
+        assert len(json_out) > 0
+
+        misaligned = [row for row in json_out if row["Base"] % 0x1000 != 0]
+        assert misaligned == [], (
+            f"{len(misaligned)} module(s) reported with a base that is neither "
+            f"page aligned nor zero, e.g. "
+            f"base={misaligned[0]['Base']:#x} size={misaligned[0]['Size']:#x} "
+            f"offset={misaligned[0]['Offset']:#x}"
+        )
+
 
 class TestWindowsMutantScan:
     def test_windows_specific_mutantscan(self, volatility, python):
--- a/volatility3/framework/plugins/windows/modscan.py
+++ b/volatility3/framework/plugins/windows/modscan.py
@@ -4,7 +4,7 @@
 import logging
 from typing import Iterable
 
-from volatility3.framework import interfaces
+from volatility3.framework import exceptions, interfaces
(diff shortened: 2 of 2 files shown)


Abyss-W4tcher (approved):
LGTM, this design is something we already do in multiple places in the framework

PR #2022, opened 2026-09-05
Import `IPython.terminal` submodules explicitly in Volshell

the-rectifier:
`volshell` crashes on startup with IPython>=9.17.0:

```python
$ python -c "import IPython; print(IPython.__version__)"
$ python -c "from IPython import terminal; print(terminal.prompts)"   # fails
$ python -c "import IPython.terminal.prompts; print('ok')"            # works

$ python volshell.py -f memory.elf -w

Volshell (Volatility 3 Framework) 2.28.2
  File "/home/canopus/dev/volatility3/volshell.py", line 11, in <module>
    volshell.main()
    ~~~~~~~~~~~~~^^
  File "/home/canopus/dev/volatility3/volatility3/cli/volshell/__init__.py", line 411, in main
    VolShell().run()
    ~~~~~~~~~~~~~~^^
  File "/home/canopus/dev/volatility3/volatility3/cli/volshell/__init__.py", line 403, in run
    constructed.run()
    ~~~~~~~~~~~~~~~^^
  File "/home/canopus/dev/volatility3/volatility3/cli/volshell/generic.py", line 138, in run
    class LayerNamePrompt(terminal.prompts.Prompts):
                          ^^^^^^^^^^^^^^^^
AttributeError: module 'IPython.terminal' has no attribute 'prompts'
```

`generic.py` does `from IPython import terminal`, then uses
`terminal.prompts` and `terminal.embed`.

Importing a package doesn't bind its submodules; those only existed
because `IPython/__init__.py` did `from .terminal.embed import embed`
at module level (up until 9.16.1).

IPython 9.17.0 moved that behind a module-level
`__getattr__` for startup time, so nothing imports them now.

Afaik, this should be a no-op on lesser versions of IPython

Edit: These changes make `from IPython import terminal` obsolete

(I really need to add a pre-commit hook for `ruff check` ...)

### DIFF
--- a/volatility3/cli/volshell/generic.py
+++ b/volatility3/cli/volshell/generic.py
@@ -25,7 +25,8 @@ except ImportError:
     has_capstone = False
 
 try:
-    from IPython import terminal
+    from IPython.terminal import embed, prompts
+
     from traitlets import config as traitlets_config
 
     has_ipython = True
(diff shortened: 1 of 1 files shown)


volatility3/cli/volshell/generic.py
```diff
@@ -25,7 +25,9 @@
     has_capstone = False
 
 try:
-    from IPython import terminal
+    from IPython.terminal import embed
+    from IPython.terminal import prompts
```
ikelos:
```suggestion
    from IPython.terminal import embed, prompts
```

ikelos:
This looks fine, thanks. I think that ruff won't complain about putting them on the same line (I kinda would've expected it to do this).  If ruff doesn't like it, that's fine just lemme know, otherwise let's go for applying it and then I'll merge it...  5:)

the-rectifier:
`ruff`'s happy. For future patches, I'm guessing I should follow this type convention, correct? When importing >1 submodule, import them in a single line.

ikelos:
Yep, I don't remember what does my formatting/import organisation for me
(I'd somewhat thought it was ruff) but unless you need to import multiple
modules as different names, there's not that much point in taking up
additional lines repeating the base class if python accepts the shorter
version.  5:)

Issue #1943, opened 2026-02-23
Linux: `system_table_name1!page` is missing the `.index` attribute


the-rectifier:
I have a memory snapshot of
```
Linux version 6.17.0-14-generic (buildd@lcy02-amd64-067) (x86_64-linux-gnu-gcc-13 (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0, GNU ld (GNU Binutils for Ubuntu) 2.42) #14~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Jan 15 15:52:10 UTC 2 (Ubuntu 6.17.0-14.14~24.04.1-generic 6.17.9)
```
that backtraces the framework when I try to dump an inode:

`vol -f mem.elf -s symbols linux.pagecache.InodePages --inode 0x8bc12d61a450 --dump`:

```python
 File "/usr/lib/python3.14/site-packages/volatility3/framework/plugins/linux/pagecache.py", line 493, in write_inode_content_to_file
    cls.write_inode_content_to_stream(context, layer_name, inode, file_obj)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/volatility3/framework/plugins/linux/pagecache.py", line 526, in write_inode_content_to_stream
    for page_idx, page_content in inode.get_contents():
                                  ~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.14/site-packages/volatility3/framework/symbols/linux/extensions/__init__.py", line 2662, in get_contents
    page_index = int(page_obj.index)
                     ^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/volatility3/framework/objects/__init__.py", line 993, in __getattr__
    raise AttributeError(
        f"{agg_name} has no attribute: {self.vol.type_name}.{attr}"

AttributeError: StructType has no attribute: symbol_table_name1!page.index
```

If I run `page_obj.get_contents()` it returns correctly the first page of the cached file. Also `page_obj.index` happens after both `inode.is_valid()` and `page_obj.is_valid()` which makes me think I might've hit another "kernel too new" issue. 

As always running latest `develop` and sample is available

eve-mem:
Yes, it looks like it might be a case of being too new. 

[...]

PR #2023, opened 2026-09-09
Add `__folio_index` and fix `memdesc_flags_t` introduction

the-rectifier:
Regarding #1943, I have been rocking the `@property...` patch locally since then without noticing any issues at all. 

In addition to that, since the introduction of `memdesc_flags_t`, `get_flags_list()` incorrectly tries to calculate flags with a `StructType` rather than an int:

```
[...]
  File "/home/canopus/dev/volatility3/volatility3/framework/symbols/linux/extensions/__init__.py", line 2828, in get_flags_list
    if self.flags & (1 << value) != 0:
       ~~~~~~~~~~~^~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for &: 'StructType' and 'int'
```

- [`__folio_index` patch](https://lists.openwall.net/linux-kernel/2025/05/14/1508)
- [`memdesc_flags_t` introduction patch](https://github.com/torvalds/linux/commit/53fbef56e07df822ea3029109ffca25328c2e5ac)

Note: even without the `flags` patch, the files were dumped successfully

### DIFF
--- a/volatility3/framework/symbols/linux/extensions/__init__.py
+++ b/volatility3/framework/symbols/linux/extensions/__init__.py
@@ -2688,6 +2688,14 @@ class page(objects.StructType):
 
         return True
 
+    @property
+    def index(self) -> int:
+        if self.has_member("__folio_index"):
+            # if __folio_index exists then we are post: https://lists.openwall.net/linux-kernel/2025/05/14/1508
+            return self.__folio_index
+
+        return self.member("index")
+
     @functools.cached_property
     def pageflags_enum(self) -> Dict:
         """Returns 'pageflags' enumeration key/values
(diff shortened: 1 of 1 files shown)


the-rectifier:
Fixed recursion issue; local manual testing passes

volatility3/framework/symbols/linux/extensions/__init__.py
```diff
@@ -2688,6 +2688,14 @@ def is_valid(self) -> bool:
 
         return True
 
+    @property
+    def index(self) -> int:
+        if hasattr(self, "__folio_index"):
+            # if __folio_index exists then we are post: https://lists.openwall.net/linux-kernel/2025/05/14/1508
+            return self.__folio_index
+
+        return super().__getattr__("index")
```
Abyss-W4tcher: I think using the dedicated symbol APIs would be a bit easier to read, something like this:

```python
    @property
    def index(self) -> int:
        if self.has_member("__folio_index"):
            # if __folio_index exists then we are post: https://lists.openwall.net/linux-kernel/2025/05/14/1508
            return self.__folio_index

        return self.member("index")
```

the-rectifier:
Agreed, this is much better, thanks!

PR #2024, opened 2026-09-10
linux.pagecache.RecoverFs: Fix infinite list iteration loop and tarball in-memory usage

### LINKED ISSUE #1975: linux.pagecache.RecoverFs consuming massive memory until reaped

Abyss-W4tcher:
Hi, this PR:

- streams the content of the live tar buffer directly to disk instead of storing it all in memory before flushing to disk;
- fixes an infinite loop in the hlist_head iteration that was causing OOM in the `linux.pagecache.RecoverFs` task as a side-effect

Fixes https://github.com/volatilityfoundation/volatility3/issues/1975

### DIFF
--- a/volatility3/framework/plugins/linux/pagecache.py
+++ b/volatility3/framework/plugins/linux/pagecache.py
@@ -528,7 +528,8 @@ class InodePages(plugins.PluginInterface):
                 max_length = inode_size - current_fp
                 page_bytes_len = min(max_length, len(page_content))
                 if current_fp >= inode_size or current_fp + page_bytes_len > inode_size:
-                    vollog.error(
+                    vollog.log(
+                        constants.LOGLEVEL_VVV,
                         "Page out of file bounds: inode 0x%x, inode size %d, page index %d",
                         inode.vol.offset,
                         inode_size,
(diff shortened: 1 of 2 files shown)


Abyss-W4tcher:
@ikelos These commits were split from another PR and passed the checks previously. The checks now fail with this error:
https://github.com/volatilityfoundation/volatility3/actions/runs/34464425956/job/102829539183?pr=2024#step:8:85

Issue #1978, opened 2026-04-30
Unable to validate the plugin requirements: ['plugins.Info.kernel.layer_name', 'plugins.Info.kernel.symbol_table_name']


KolHost:
Hi, I installed volatility3 but when I want to execute this command "python3 vol.py -f ~//dump/dump.mem windows.info" then show me this:

Volatility 3 Framework 2.28.1
Progress:  100.00		PDB scanning finished                   
Unsatisfied requirement plugins.Info.kernel.layer_name: 
Unsatisfied requirement plugins.Info.kernel.symbol_table_name: 

A translation layer requirement was not fulfilled.  Please verify that:
	A file was provided to create this layer (by -f, --single-location or by config)
	The file exists and is readable
	The file is a valid memory image and was acquired cleanly

A symbol table requirement was not fulfilled.  Please verify that:
	The associated translation layer requirement was fulfilled
	You have the correct symbol file for the requirement
	The symbol file is under the correct directory or zip file
	The symbol file is named appropriately or contains the correct banner

Unable to validate the plugin requirements: ['plugins.Info.kernel.layer_name', 'plugins.Info.kernel.symbol_table_name']

ikelos:
This is why we ask people to use the bug template we've produced.  It asks you the relevant questions and explains the information we'll need to help.  There's a number of reasons you might be encountering this particular issue (many are listed in the error message you mentioned), but we ask people to provide the output of running the tool with `-vvv` at a minimum, since that will provide us more debugging information that can help pinpoint the issue.

Please can you attach the output of running `python3 vol.py -vvv -f ~//dump/dump.mem windows.info`?

KolHost:
python3 vol.py -vvv -f ~/dump/dump.mem windows.info
INFO     volatility3.cli: Volatility plugins path: ['/home/kolhost/volatility3/volatility3/plugins', '/home/kolhost/volatility3/volatility3/framework/plugins']
INFO     volatility3.cli: Volatility symbols path: ['/home/kolhost/volatility3/volatility3/symbols', '/home/kolhost/volatility3/volatility3/framework/symbols', '/home/kolhost/.cache/volatility3/symbols']
DEBUG    volatility3.plugins.yarascan: Using yara-python module
Traceback (most recent call last):
  File "/home/kolhost/volatility3/vol.py", line 16, in <module>
    volatility3.cli.main()
    ~~~~~~~~~~~~~~~~~~~~^^
[...]

Issue #1970, opened 2026-04-06
Unable to scan or run modules on  windows 11 memory dump image


anisharma-debug:
**Describe the bug**
While running the windows.info module on a windows 11 memory dump image taken via multiple tools like Belkasoft RamCapturer, Magnet Response, Magnet Response, i unable to read the layers

**Context**
Volatility Version:  latest
Operating System:  Windows 11 WSL and Ubuntu installation (Error is same on both)
Python Version:  3.12.3
Suspected Operating System:  Windows 11
Command:  python vol.py -vvv -f redacted.mem windows.info

**To Reproduce**

1. Use command '...'
2. See error:

**Expected behavior**
Windows version was expected

**Example output**
DEBUG    volatility3.framework: Failed to import module volatility3.plugins.windows.registry.lsadump based on file: /mnt/c/Users/anisharma/Desktop/testing memory/volatility3/volatility3/framework/plugins/windows/registry/lsadump.py
[...]

PR #2033, opened 2026-09-14
automagic/windows: merge DTB fast-path ranges into one contiguous span

brian-mckinney:
## Problem

`WindowsIntelStacker.test_sets` has three separate physical address ranges for the 64-bit DTB search, leaving two gaps:

| Gap | Size |
| --- | --- |
| `0x2a0000` – `0x550000` | 2.7 MiB |
| `0x6f0000` – `0x900000` | 2.1 MiB |

On one 25H2 image in our environment (build 26200.8037, 8 GiB Hyper-V Gen2), the System DTB landed at `0x7b5000`, inside the second gap. Every plugin fails with a translation error on that image; `--dtb 0x7b5000` by hand makes them work. Because the DTB address varies per boot, not all images of this build fail, only those whose DTB happens to land in a gap.

The DTB sits at the high-water mark of physical memory the boot loader claims before building the page tables. As `winload.efi` has grown (1.52 MiB at build 14393, 3.62 MiB at build 28000), the DTB has moved to higher addresses and the current ranges haven't kept up.

Related: 
- #1872 
- #1970 
- #1978

## Fix

Merge the three ranges into one contiguous span:

```diff
            [
-               (0x150000, 0x150000),
-               (0x550000, 0x1A0000),
-               (0x900000, 0x100000),
+               (0x150000, 0x8B0000),
            ],
```

The merged range ends at `0xa00000`, where the wider fallback scan begins, so there are no gaps between the two.

Widening is safe because #1992 added translation-based validation: any candidate DTB is verified by translating a known kernel address before being accepted. Extra candidates cost scan time only.

## Timing

Two images: the failing 25H2 dump and the Win XP integration test image (DTB at `0x39000`, well within the existing ranges and unaffected by this change).

| Image | develop | this patch |
| --- | --- | --- |
| Win XP 5.1, 512 MiB | ~2.5 s, DTB `0x39000` | ~2.5 s, DTB `0x39000` |
| Win 11 25H2, 11.8 GiB | 65 s, DTB `0x5507000` (process table, wrong) | 3.5 s, DTB `0x7b5000` (System, correct) |

On `develop`, the 25H2 image exhausts the fast-path ranges without a hit, then finds a process page table in the 80 MiB fallback scan. Plugins appear to work but are operating on the wrong address space. The patch finds the correct System DTB in the fast path and avoids the fallback entirely.

## Checklist

- [x] Win XP test image timing unchanged
- [x] Two stale docstring references to specific range values corrected
- [x] PR #1971's `(0xa00000, 0x5000000)` range is untouched


brian-mckinney:
Hello, would it be possible to get some feedback on this PR, at least to know if it has a chance at making it through?

ikelos:
Sorry, the team's time is quite limited these days, so I apologize for the delay in reviewing this.

The reason the code was done in gaps was to ensure that the right areas were hit quickly without a long delay.  Skipping to where we'd empirically seen actual examples saved a fair amount of time scanning pages that we'd never seen DTBs turn up in.

So the timings you've given are useful, but ideally it would test how long it takes to find an image that's at the very end of the ranges we look in.

Probably this will get included, but it may take us a bit more time to verify the trade offs for various images.

brian-mckinney:
Would you be open to adding the 2 gap ranges to the slow path instead?

```python
# "Very large memory with high DTBs (slow)"
[
    (0x2A0000, 0x2B0000),   # gap 1 — was 0x2a0000–0x550000
    (0x6F0000, 0x210000),   # gap 2 — was 0x6f0000–0x900000
    (0xA00000, 0x5000000),  # original
]
```

This would preserve the fast path, and also find the DTBs we are seeing in our images.  I ran that change against the two images referenced in the PR body and they came out essentially the same, while at the same time preserving existing behavior.

| Image | develop | this PR | workaround |
| --- | --- | --- | --- |
| Win XP 5.1, 512 MiB (DTB `0x39000`) | ~2.5 s, correct | ~2.5 s, correct | ~2.5 s, correct |
| Win 11 25H2, 11.8 GiB (DTB `0x7b5000`) | wrong DTB (`0x5507000`) | 3.5 s, correct | 3.6 s, correct |

